PR #115, opened 2020-03-10
fix gcp-sql-user.yaml password

MingyanZhao:
According to the specification of [SQLUser](https://cloud.google.com/config-connector/docs/reference/resources#spec_75), the `password` field is a `oneof` field. A `value` field should be enough for the sample app use case.


MingyanZhao:
> @googlebot I signed it!

MingyanZhao:
@googlebot I signed it!

MingyanZhao:
Obsolete

PR #169, opened 2020-05-08
Fix region tags (1 of 2)

jmdobry:
1.  Add corrected region tags (this PR).
2. Point docs to corrected region tags (pending).
3. Remove incorrect region tags (will send PR after No. 2 is done).


jmdobry:
I can't add you as a reviewer, so pinging via comment @spew @kibbles-n-bytes

jmdobry:
Making change elsewhere...

Issue #78, opened 2020-01-16
Support referencing IAMCustomRole from IAMPolicyMember and IAMPolicy


ppennanen:
There is a CRD for `IAMCustomRole` however it appears impossible to refer to a custom role from `IAMPolicy` and `IAMPolicyMember`, since the pattern for the role field is specified as `^roles/[\w\.]+$`.

It would be useful to allow binding custom roles with paths like `projects/<project id>/roles/<role id>` or `organization/<organization id>/roles/<role id>` .

xiaobaitusi:
Hi ppennanen@, thanks for your feedback. I'll mark it as enhancement for this feature request. Ack that  we should support binding custom roles.

kibbles-n-bytes:
@ppennanen As a temporary workaround, if you are blocked, we have confirmed that removing the `pattern: ^roles/[\w\.]+$` restriction within the JSON schema for the `IAMPolicy` and `IAMPolicyMember` CRDs and using a relative resource name like `projects/[PROJECT]/roles/[CUSTOM_ROLE]` works.

You can do so via the following:
* Before installing, remove the validation string in `install-bundle/crds.yaml` with:  `sed -i '/pattern: \^roles/d' install-bundle/crds.yaml`

OR

* After installing, remove the validation string from the CRDs by:
  * Run `kubectl edit iampolicies.iam.cnrm.cloud.google.com`
  * Find the line that says `pattern: ^roles/[\w\.]+$` and delete it
  * Repeat the above two steps, but for `iampolicymembers.iam.cnrm.cloud.google.com` instead of `iampolicies.iam.cnrm.cloud.google.com`

Now something like this should work:
```yaml
apiVersion: iam.cnrm.cloud.google.com/v1beta1
kind: IAMPolicyMember
metadata:
  name: iampolicymember-sample
spec:
    member: user:[USER]@google.com
    resourceRef:
      kind: Project
    role: projects/[PROJECT]/roles/iamcustomrolesample
```

Please note that our official support may look a bit different as we will want to be able to reference `IAMCustomRole` objects within the cluster as well.

ppennanen:
Cool. Thank you!

dan-j:
If you want to maintain validation, you could also edit the regex like so:

```
pattern: ^(roles|projects)/[\w-/\.]+$
```

This isn't the most restrictive but gets the job done, it allows for `-` in project names too.

govinda-attal:
[...]

PR #245, opened 2020-07-17
Update pattern to support organization level roles

regadas:
Related to #78


jcanseco:
Hi @regadas, we appreciate the effort and initiative put towards making this PR, though we're not quite yet ready to accept PRs in general as we're not an open source project. So if it's ok with you, I'll be closing this PR and will let you know in #78 when we've added support for referencing org custom roles. Thank.

PR #615, opened 2022-02-24
fix prometheus scraping

jnodorp-jaconi:
This fixes a regression introduced with 1.46.0. Prometheus uses services to discover pods, but does still scrape the pods directly, not the service. Therefore the annotation has to contain the target port.


xiaobaitusi:
Hi @jnodorp-jaconi, thanks for offering the PR. Unfortunately, Config Connector at this moment doesn't accept contribution via Github because the source of truth code lives google internal. 

I've submitted a change and the fix will be released within next few versions. Thanks.

rasmus:
@xiaobaitusi please reverse this as mentioned in https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/618#issuecomment-2007175644 port `8888` is the valid port on the service and `48797` is the valid port on the pod.

PR #674, opened 2022-06-27
Enable using .spec.patches on ConfigConnector

tomasaschan:
At Spotify we need to bump the resource limits of some of the resources created by activating Config Connector. Currently, we do this by applying a patched version of this operator that has the changes we need, but that requires manual work from us every time we update KCC.

This PR enables the `.spec.patches` addon on the `ConfigConnector` resource type, so that we can make the changes we need that way instead, and run the operator itself without modifications.

I also drive-by updated the Declarative Pattern version used in this repo; let me know if you'd like to have that excluded here.


kaovilai:
Hey @tomasaschan did you sign the CLA?

tomasaschan:
It should be automatically handled by virtue of me working at Spotify, but as I'm on parental leave since December (and will be until after summer) I might have fallen out of the rolls due to inactivity probably something.

PR #703, opened 2022-09-06
Add least privilege role file

shannonxtreme:
We're updating the hardening guide to use the newly added `roles/container.nodeServiceAccount` least privilege IAM role. This file binds that role to the new service account


diviner524:
Worked with Shannon offline and the change was submitted through internal repo.

PR #804, opened 2023-04-25
Fix ResourceReference fields that are not ResourceReferences

zealws:
This PR modifies the type generate code to fix an assumption relating to ResourceReference fields.

Fixes #598 

The PR has 3 commits:
- ResourceReference must have the Ref name and expected children.
  - This commit modifies the `isResourceReference` code so that a ResourceReference field must have the expected `Ref` name suffix, and the expected children. Both checks must be true for the field to be a resource reference.
- Add test for isResourceReference
  - This commit adds a unit-test file that tests the behavior of `isResourceReference`.
- Generated changes from 'make test'
  - This commit includes the generated changes from running `make test`.

Note that I couldn't actually run the tests locally since they assume `kubebuilder` is installed in `/usr/local/kubebuilder`:
```
❯ go test -v ./scripts/generate-go-crd-clients -run TestIsResourceField
{"severity":"error","timestamp":"2023-04-25T12:33:09.738-0400","logger":"controller-runtime.test-env","msg":"unable to start the controlplane","tries":0,"error":"fork/exec /usr/local/kubebuilder/bin/etcd: no such file or directory"}
{"severity":"error","timestamp":"2023-04-25T12:33:09.738-0400","logger":"controller-runtime.test-env","msg":"unable to start the controlplane","tries":1,"error":"fork/exec /usr/local/kubebuilder/bin/etcd: no such file or directory"}
{"severity":"error","timestamp":"2023-04-25T12:33:09.739-0400","logger":"controller-runtime.test-env","msg":"unable to start the controlplane","tries":2,"error":"fork/exec /usr/local/kubebuilder/bin/etcd: no such file or directory"}
{"severity":"error","timestamp":"2023-04-25T12:33:09.740-0400","logger":"controller-runtime.test-env","msg":"unable to start the controlplane","tries":3,"error":"fork/exec /usr/local/kubebuilder/bin/etcd: no such file or directory"}
{"severity":"error","timestamp":"2023-04-25T12:33:09.741-0400","logger":"controller-runtime.test-env","msg":"unable to start the controlplane","tries":4,"error":"fork/exec /usr/local/kubebuilder/bin/etcd: no such file or directory"}
FAIL    github.com/GoogleCloudPlatform/k8s-config-connector/scripts/generate-go-crd-clients     0.183s
FAIL
```

I have `kubebuilder` installed, but my installation did not include an `etcd` binary.
```
❯ command -v kubebuilder
/usr/bin/kubebuilder

❯ kubebuilder version
[...]

PR #848, opened 2023-08-11
Fix ResourceReference fields that are not ResourceReferences (cont.)

joshheinrichs-shopify:
This is a continuation of https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/804 with the go client generated.


scripts/generate-go-crd-clients/client_test.go
```diff
@@ -151,9 +152,9 @@ func TestComputeInstanceGoClient(t *testing.T) {
 					SubnetworkRef: &v1alpha1.ResourceRef{
 						External: "default",
 					},
-					NetworkIpRef: &v1alpha1.ResourceRef{
-						Kind: "ComputeAddress",
-						Name: computeAddressName,
+					NetworkIpRef: &computev1beta1.InstanceNetworkIpRef{
+						Kind: &networkIpKind,
+						Name: &computeAddressName,
```
joshheinrichs, line 157: according to the schema in https://cloud.google.com/config-connector/docs/reference/resource-docs/compute/computeinstance

* NetworkIpRef should have these fields (all strings are optional)
  ```
  networkIpRef:
    external: string
    kind: string
    name: string
    namespace: string
   ```
* SubnetworkRef should have these fields (all strings are optional)
  ```
  subnetworkRef:
    external: string
    name: string
    namespace: string
  ```

so it's a bit odd that:
* SubnetworkRef is a ResourceRef which has a Kind field that isn't supported in the schema
* Fields are `string`s in ResourceRef, and `*string`s in InstanceNetworkIpRef. I'm not sure which is preferable but it'd be good to be consistent.

joshheinrichs-shopify, line 157: according to the schema in https://cloud.google.com/config-connector/docs/reference/resource-docs/compute/computeinstance#schema

* NetworkIpRef should have these fields (all strings are optional)
  ```
  networkIpRef:
    external: string
    kind: string
    name: string
    namespace: string
  ```
* SubnetworkRef should have these fields (all strings are optional)
  ```
  subnetworkRef:
    external: string
    name: string
    namespace: string
  ```

so it's a bit odd that:
1. SubnetworkRef is a ResourceRef which has a Kind field that isn't supported in the schema
2. Fields are `string`s in ResourceRef, and `*string`s in InstanceNetworkIpRef. I'm not sure which is preferable but it'd be good to be consistent.

i'll look into cleaning up these issues next week.

joshheinrichs-shopify:
we're not using config connector anymore so i unfortunately won't get back to this any time soon 😞

PR #859, opened 2023-09-26
mockgcp support for CertificateManager CertificateMap and dnsAuthorizations

justinsb:
Add mockgcp support for CertificateManager CertificateMap and dnsAuthorizations

### DIFF
--- a/mockgcp/mockcertificatemanager/v1.go
+++ b/mockgcp/mockcertificatemanager/v1.go
@@ -16,6 +16,7 @@ package mockcertificatemanager
 
 import (
 	"context"
+	"time"
 
 	pb "google.golang.org/genproto/googleapis/cloud/certificatemanager/v1"
 	"google.golang.org/genproto/googleapis/longrunning"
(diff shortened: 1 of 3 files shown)


justinsb:
(WIP, this obviously can't work until I actually implement the functions, but hopefully we can see good results from our CI suite)

justinsb:
The tests do indeed fail (with no implementation) - good.  Somewhat difficult to see, but the failing test is `TestAllInSeries/samples/certificatemanagercertificatemap` and so

Setup kubebuilder:
```
go run  sigs.k8s.io/controller-runtime/tools/setup-envtest@latest use -p env
```

Run tests:
```
E2E_KUBE_TARGET=envtest RUN_E2E=1 E2E_GCP_TARGET=mock  go test ./tests/e2e/ -test.count=1 -v -run TestAllInSeries/samples/certificatemanagercertificatemap
```

Then fails locally with:

```
2023/09/26 17:08:13 request: GET https://certificatemanager.googleapis.com/v1/projects/mock-project-1695762479/locations/global/certificateMaps/certificatemanagercertificatemap-sample?alt=json
2023/09/26 17:08:13 [DEBUG] Retry Transport: Stopping retries, last request failed with non-retryable error: googleapi: got HTTP response code 501 with body: HTTP/0.0 501 Not Implemented

{"code":12,"message":"method GetCertificateMap not implemented","details":[]}
```

Clearly indicating that we need to support GetCertificateMap, which is reasonably clear.

justinsb:
And indeed, implementing that method and the create and delete (which I was similarly prompted to implemented when running the `go test` locally):

`2023-09-26T21:40:25.4928233Z         --- PASS: TestAllInSeries/samples/certificatemanagercertificatemap (69.69s)`

mockgcp/mockcertificatemanager/v1.go
```diff
@@ -88,3 +89,125 @@ func (s *CertificateManagerV1) DeleteCertificate(ctx context.Context, req *pb.De
...
+		} else {
+			return nil, status.Errorf(codes.Internal, "error reading dns authorization: %v", err)
+		}
+	}
+
+	return obj, nil
+}
+
+func (s *CertificateManagerV1) CreateDnsAuthorization(ctx context.Context, req *pb.CreateDnsAuthorizationRequest) (*longrunning.Operation, error) {
+	reqName := req.Parent + "/dnsAuthorizations/" + req.DnsAuthorizationId
```
jingyih, line 177: For my own education, how is the format for `reqName` determined? I am looking at Terraform implementation, the create URL looks like this `{{CertificateManagerBasePath}}projects/{{project}}/locations/global/dnsAuthorizations?dnsAuthorizationId={{name}}` [1]. I did not look into the actual GCP implementation on how the url is parsed.


[1] https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/44bfcef1694935b669a6b1e72f4fc54f7784eb71/third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/certificatemanager/resource_certificate_manager_dns_authorization.go#L143C49-L143C164

justinsb: So this is probably the hardest bit of the mock :-)  What we're trying to do is establish the "fully qualified name" of the resource.  The only reason we actually need that in most of our mocks is for a key in our map of objects, so we create it / delete it / get it with the same key.  Arguably we could use anything - for example here we could have a type like `type Key struct { Project string; DNSAuthorizationID string; }`.  But I think we will want to populate the selfLink in at least some APIs, so using the same key that GCP uses externally seems to be an easier choice.

In a get or a delete, we normally get this key directly because grpc-gateway parses the URL into a string field in the request object.  In the create, there's a sort of pattern that we get the _parent_ "key", and then we have to add on the child name etc.  My guess is that we have some AIP somewhere that recommends this, as most of our recent APIs seem to follow this pattern!

jingyih: Thank you for the explanation, it makes sense. I was just making sure I did not miss anything.

PR #861, opened 2023-09-26
mockgcp: Support returning values from LROs

justinsb:
Some clients rely on the value in the LRO; by supporting this we can
also clean up the API for delete.

### DIFF
--- a/mockgcp/common/operations/operations.go
+++ b/mockgcp/common/operations/operations.go
@@ -20,9 +20,12 @@ import (
 	"time"
 
 	pb "google.golang.org/genproto/googleapis/longrunning"
+	rpcstatus "google.golang.org/genproto/googleapis/rpc/status"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/encoding/prototext"
+	"google.golang.org/protobuf/proto"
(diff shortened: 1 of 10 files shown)


justinsb:
Rebase

justinsb:
Rebased

mockgcp/common/operations/operations.go
```diff
@@ -60,6 +63,57 @@ func (s *Operations) NewLRO(ctx context.Context) (*pb.Operation, error) {
...
 }
 
+func (s *Operations) StartLRO(ctx context.Context, callback func() (proto.Message, error)) (*pb.Operation, error) {
+	now := time.Now()
+	millis := now.UnixMilli()
+	id := uuid.NewUUID()
+
+	op := &pb.Operation{}
+
+	op.Name = fmt.Sprintf("operations/operation-%d-%s", millis, id)
```
cheftako, line 73: YYYY-MM-DDTHH:mm:ss.sssZ maintains time sort order but has the advantage of being human readable. (Sorry I can't convert epoch millis to something more readable in my head) Might be worth using this. You can always use '-' if the ':' are a problem.

justinsb: Good point.  We're actually going to clean this up in a few commits, because we want to look as close to GCP as possible (so we can do golden testing of requests/responses)

PR #862, opened 2023-09-27
Skipping problematic reinstall of gcloud if setup.

cheftako:
I see an error on the apt install if I already have the package installed.
Added a check to skip the relevant apt install if the package is already installed.


justinsb:
Thanks @cheftako 

/approve
/lgtm

fmichaelobrien (commented):
LGTM

PR #863, opened 2023-09-27
mockgcp: generate isolated proto & grpc & grpc-gateway

justinsb:
Because of surprising proto limitations we have to rewrite the proto
files into a different path.  But the advantage here is that we are
self-contained, and don't have any version skew with other libraries.


justinsb:
See https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/866#issuecomment-1737386998 for an example of the problem

justinsb:
I had to rebase, and lost lgtm... mind reapplying @gemmahou ?

PR #866, opened 2023-09-27
WIP: APIKeys Key resource

justinsb:
Manually scaffold the APIKeys Key resource (which is an oddity in terraform, implemented via DCL).

Also implement a mock.


justinsb:
This requires isolated imports in mockgcp, as implemented in #863

```
panic: proto: file "google/api/apikeys/v2/resources.proto" is already registered
	previously from: "cloud.google.com/go/apikeys/apiv2/apikeyspb"
	currently from:  "github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/generated/google/api/apikeys/v2"
See https://protobuf.dev/reference/go/faq#namespace-conflict
```

diviner524:
@justinsb I may have missed something obvious. This looks similar to how we add a TF based resource. What is the main difference when you mentioned "Manually scaffold"? For example do you plan to manually write the controller for this resource?

justinsb:
@diviner524 - sorry, by "manually scaffold" I mean adding support _not_ via the TF autogen approach.  I believe that TF autogen doesn't work for this resource because it is DCL based?  I should double-check our instructions for how to add a TF resource though, it might answer some of my todos :-)

I do also intend to try implementing this one manually, but first I want to show that the terraform implementation has some problems.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_apikeyskeys.apikeys.cnrm.cloud.google.com.yaml
```diff
@@ -0,0 +1,249 @@
...
+      type: string
+    - description: The reason for the value in 'Ready'
+      jsonPath: .status.conditions[?(@.type=='Ready')].reason
+      name: Status
+      type: string
+    - description: The last transition time for the value in 'Status'
+      jsonPath: .status.conditions[?(@.type=='Ready')].lastTransitionTime
+      name: Status Age
+      type: date
+    name: v1alpha1
```
diviner524, line 43: Since we are already manually adding it... Why not just mark the CRD as v1beta1? Do you see failing sample tests/integration tests when trying them locally? Or there are some fields with uncertainty?

IIUC this PR is basically doing the same as:

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/README.NewResourceFromTerraform.md

diviner524:
> @diviner524 - sorry, by "manually scaffold" I mean adding support _not_ via the TF autogen approach. I believe that TF autogen doesn't work for this resource because it is DCL based? I should double-check our instructions for how to add a TF resource though, it might answer some of my todos :-)
> 
> I do also intend to try implementing this one manually, but first I want to show that the terraform implementation has some problems.

I see, thanks for sharing!

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_apikeyskeys.apikeys.cnrm.cloud.google.com.yaml
justinsb: This resource will only be as good as terraform, I think we have a higher bar for v1beta1?  (Maybe we didn't enforce this bar before we introduced v1alpha1?)

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_apikeyskeys.apikeys.cnrm.cloud.google.com.yaml
```diff
@@ -0,0 +1,249 @@
+apiVersion: apiextensions.k8s.io/v1
```
jingyih, line 1: Hi Justin, is the generation of this CRD purely manual? Or is there some automation involved?

justinsb: It uses the "classic" generation mechanism (i.e. not automated generation) from a terraform resource

justinsb:
Removing WIP.  I believe as long as it's alpha it will be opt-in only.

justinsb:
Rebased; I plan on doing the generation again though in case anything changed in terraform in the past 4 months.  cc @cheftako

justinsb:
Ran through generation again, I propose we look at #1168 instead, and then I'll rebase this one and see if there's anything left.

/hold

yuwenma:
Shall we close this PR?

PR #867, opened 2023-09-27
Generate clients without requiring GOPATH

justinsb:
Some of the kubernetes generators still work better with GOPATH; we
can create a temporary directory structure to keep them happy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -145,3 +145,4 @@ Session.vim
 
 # always skip vendored files
 /vendor
+.build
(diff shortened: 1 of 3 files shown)


justinsb:
Removing WIP; verified this works for #887 (and without it, we can't do `make generate-go-clients`)

jingyih (approved):
Thank you for fixing this. I remember I had to manually set `GOPATH` to bypass some issues.

/lgtm
/approve

PR #868, opened 2023-09-27
Add updates to tests/e2e

justinsb:
This ensures that we get a bit more test coverage here.

### DIFF
--- a/config/tests/samples/create/samples.go
+++ b/config/tests/samples/create/samples.go
@@ -15,7 +15,6 @@
 package create
 
 import (
-	"context"
 	"fmt"
 	"io/ioutil"
 	"path"
@@ -41,6 +40,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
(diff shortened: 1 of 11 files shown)


config/tests/samples/create/samples.go
```diff
@@ -90,17 +90,44 @@ func getNamespaces(samples []Sample) []string {
...
+	waitForReady(t, opt.Create)
+
+	if len(opt.Updates) != 0 {
+		// treat as a patch
+		for _, updateUnstruct := range opt.Updates {
+			if err := t.GetClient().Patch(ctx, updateUnstruct, client.Apply, client.FieldOwner("kcc-tests"), client.ForceOwnership); err != nil {
+				t.Fatalf("error updating resource: %v", err)
+			}
+		}
+		time.Sleep(5 * time.Second)
```
cheftako, line 123: :( This seems like a flake waiting to happen. Can we switch to something which stops waiting early when success is reached but has a longer timeout for failure? I know K8s has utilities like that. Not saying you have to use those utilities but ....

justinsb: Fair - it's not ideal.  We need status.observedGeneration to do this "right" (that's why we're waiting here).  I'll check out if we have that field today....

justinsb: Good news, we do have observedGeneration and it looks easy to improve our wait-for-ready logic to incorporate it: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/871

justinsb: I've removed this sleep.  It's not 100% safe until we get #871 in, but it should "fail open" I believe

justinsb:
@diviner524 would be great to get this one in, without it I'm reluctant to mark other PRs that extend mockgcp in (because we aren't testing the update method without this PR)

config/tests/samples/create/samples.go
```diff
@@ -90,17 +90,42 @@ func getNamespaces(samples []Sample) []string {
 	return results
 }
 
-func RunCreateDeleteTest(t *Harness, unstructs []*unstructured.Unstructured, cleanupResources bool) {
+type CreateDeleteTestOptions struct {
+	// Create is the set of objects to create
+	Create []*unstructured.Unstructured
+
+	// Updates is the set of objects to update (after all objects have been created)
+	Updates []*unstructured.Unstructured
```
diviner524, line 98: * Where do we plan to store these updates (assuming they are going to be another set of YAML files with slight changes)? If they are going to be under `samples/resources/` we need to make sure they don't get rendered in the resource reference page.

* Also we are already testing updates using artifacts under `pkg/test/resourcefixture/testdata/basic`.
For example: https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/master/pkg/test/resourcefixture/testdata/basic/certificatemanager/v1beta1/certificatemanagercertificate/certificatemanagercertificatemanageddns

Would it be easier to directly enable mock for these test cases?

justinsb: Yes, this is leveraging the existing fixture/sample data.  Where those have updates, we will now do those updates.

I don't mind us trying to enable mocks with the "old" tests, they're just a little more complicated to deal with, so I found it easier to just build up a new set of tests for the mock tests.

diviner524: > Where those have updates, we will now do those updates.

Are you referring to data under `config/samples/resources` or other test data? I don't think we have any updates under `config/samples/resources`.

justinsb: You're right, looks like we only set Update on the fixtures test data, not the samples test data.

justinsb:
Rebased

justinsb:
cc @diviner524 this one is holding up the rest of the parade, because without it we don't require the update method (so it has to keep growing as we add more resources)

gemmahou (approved):

PR #869, opened 2023-09-27
mockgcp: Add basic mock for cloudfunctions

justinsb:
Working on boosting our (fast-path) test coverage.


mockgcp/mockcloudfunctions/v1.go
```diff
@@ -0,0 +1,89 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mockcloudfunctions
+
+import (
+	"context"
+
+	pb "google.golang.org/genproto/googleapis/cloud/functions/v1"
```
jingyih, line 20: We need to migrate away from the deprecated proto?

justinsb: Now that we've merged #863 we shouldn't need to

jingyih (commented):
/lgtm

PR #870, opened 2023-09-27
Add tests for exporting resources

justinsb:
We need to plumb through the context so we can use this in our mocks, and then we add a test for exporting resources.

Currently just one test (for serviceusage), because it requires some mapping to the export "URI".


justinsb:
cc @diviner524 this blocks the golden request tests

diviner524 (commented):
/lgtm

PR #871, opened 2023-09-28
tests: Verify the object observedGeneration

justinsb:
This is more accurate, and means we don't have to rely on waiting and
hoping for a reconcile.


jingyih (commented):
/lgtm

PR #872, opened 2023-09-28
github CI: upload artifacts directory

justinsb:
This lets us see richer and more structured output than we get with
just the logfile.


diviner524 (approved):
I don't see anything gets uploaded, I assume it is expected as we haven't enabled artifacts for mock tests?

https://github.com/GoogleCloudPlatform/k8s-config-connector/actions/runs/6338592433/job/17216027901#step:6:10

jingyih:
On a side note, how do we apply "approved" and "lgtm" labels to make tide happy?

PR #873, opened 2023-09-28
github ci: mock tests should not need gcloud CLI

justinsb:
Stop installing the gcloud CLI, so we can validate that it is not a
requirement for the mock tests.


jingyih (commented):
/lgtm

PR #874, opened 2023-09-28
tests: set the test into the context

justinsb:
This lets us include the test name in the artifact path.


acpana:
[triage] @justinsb do we want to keep this PR around?

PR #878, opened 2023-10-02
Add a few rules in the operator Makefile.

jingyih:
Add "install", "uninstall", "deploy" and "destroy" rules to the operator Makefile.

### DIFF
--- a/operator/Makefile
+++ b/operator/Makefile
@@ -72,6 +72,15 @@ ifneq (Version: v$(CONTROLLER_GEN_VERSION), $(shell $(CONTROLLER_GEN) --version)
 	GOFLAGS='' go install sigs.k8s.io/controller-tools/cmd/controller-gen@v$(CONTROLLER_GEN_VERSION)
 endif
 
+# Install CRDs into a cluster
+.PHONY: install
+install: manifests
+	kustomize build config/crd | kubectl apply -f -
+
+# Uninstall CRDs from a cluster
+.PHONY: uninstall
+uninstall: manifests
+	kustomize build config/crd | kubectl delete -f -
 
 # Build the docker image
 .PHONY: docker-build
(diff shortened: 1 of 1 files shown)


operator/Makefile
```diff
@@ -72,6 +72,15 @@ ifneq (Version: v$(CONTROLLER_GEN_VERSION), $(shell $(CONTROLLER_GEN) --version)
 	GOFLAGS='' go install sigs.k8s.io/controller-tools/cmd/controller-gen@v$(CONTROLLER_GEN_VERSION)
 endif
 
+# Install CRDs into a cluster
+.PHONY: install
+install: manifests
```
justinsb, line 77: I think these are existing from our internal repo (?), so we probably want to add them as-is and then make them better.

Some ideas for how to make them better:

* Use server-side apply
* Maybe rename install to install-crds? (and same for uninstall)
* If deploy relies on docker-build, maybe we should make it a makefile dependency :-)

But these are nits and I vote to merge and iterate!

justinsb:
Thanks - I think this is pulling these across from our internal repo (RIP!), so I think a merge-and-iterate strategy is the right one here.

/lgtm

(Not sure the lgtm-bot is fully working yet, but worth a shot...)

justinsb:
It may be on now... maybe

/lgtm

operator/Makefile
jingyih: Yeah these rules are from internal Makefile which I used often when working on the operator code.

jingyih:
> It may be on now... maybe
> 
> /lgtm

Looks like it is not... Shall we merge a PR after getting one lgtm? Not sure if there is specific guidelines for accepting and merging PRs. In etcd repo we used to get 2 LGTM (ideally from different org), but the situation is different for each repo.

diviner524 (approved):
LGTM

PR #887, opened 2023-10-04
Fix "certificate" typo in description

justinsb:
This makes it into the docs.


justinsb:
(I do expect this will fail because I need to generate docs / crds etc.  I want to verify that is the case)

jingyih:
Test failure says `make generate-go-client` is needed. Did you use `make ready-pr` to generate the files for the second commit?

justinsb:
> Did you use make ready-pr to generate the files for the second commit?

Not yet, but running it now :-)  I'm sort of going through the contributor experience here.  Obviously it would be better if we didn't require contributors to generate code, but as long as the tests catch it that is okay IMO.

PR #891, opened 2023-10-05
generate-google3-docs: stop using sprig

justinsb:
It doesn't appear that we were using it, and this avoids GO-2022-0411


justinsb:
Looks like we aren't using this, and then we fix a vuln report.  cc @diviner524

diviner524 (approved):
/lgtm

PR #892, opened 2023-10-05
Bump golang version to latest 1.21

justinsb:
So we get the latest and greatest security fixes.


justinsb:
/hold

Apparently 1.21.2 is coming today, so we might as well wait: https://groups.google.com/g/golang-announce/c/2dWHvJVFA9s

justinsb:
Updated to 1.21.2.  Note that we're using a floating `golang:1.21` tag in our Dockerfiles, not sure if that's a better approach (less reproducible vs easier updates).

/hold cancel

jingyih (commented):
/lgtm

PR #894, opened 2023-10-05
Check for duplicate webhooks in customization CR

### LINKED ISSUE #893: Customizing the same webhook multiple times in a custom resource should be considered an error
### Checklist

`ValidatingWebhookConfigurationCustomization` and `MutatingWebhookConfigurationCustomization` CRs could contain duplicate webhooks within the `.spec.webhooks` list. In such cases, the final entry will prevail. Customizing the same webhook multiple times in a custom resource should be considered an error, which can be surfaced through the custom resource's status field.

### Additional information

_No response_

### Importance

_No response_

jingyih:
### Change description

Check for duplicate webhooks in `.spec.webhooks` list of webhook configuration customization CRs.

Fixes #893, b/296513545

### Tests you have done

### DIFF
--- a/operator/pkg/controllers/configconnector/configconnector_controller.go
+++ b/operator/pkg/controllers/configconnector/configconnector_controller.go
@@ -691,6 +691,9 @@ func (r *ConfigConnectorReconciler) fetchAndApplyAllWebhookConfigurationCustomiz
 
 // applyValidatingWebhookConfigurationCustomizationCR applies customizations specified in a ValidatingWebhookConfigurationCustomization CR.
 func (r *ConfigConnectorReconciler) applyValidatingWebhookConfigurationCustomizationCR(ctx context.Context, cr *customizev1alpha1.ValidatingWebhookConfigurationCustomization) error {
+	if err := checkForDuplicateWebhooks(cr.Spec.Webhooks); err != nil {
+		return r.handleApplyValidatingWebhookConfigurationCustomizationCRFailed(ctx, cr, fmt.Sprintf("invalid webhook configuration customization: %v", err))
+	}
 	// 1. get the webhook configuration.
 	whCfg := &admissionregistration.ValidatingWebhookConfiguration{}
 	targetWebhookConfigurationName := fmt.Sprintf("%s.cnrm.cloud.google.com", cr.Name)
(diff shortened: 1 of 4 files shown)


operator/pkg/test/controller/customization.go
```diff
@@ -224,6 +224,46 @@ var NamespacedControllerResourceCRWrongNamespace = &customizev1alpha1.Namespaced
...
+				{
+					Name: "resource-validation", // another valid webhook name
+				},
+				{
+					Name: "deny-immutable-field-updates", // a valid webhook name but duplicated
+				},
+			},
+		},
+	}
+	MutatingWebhookCRForDuplicatedWebhook = &customizev1alpha1.MutatingWebhookConfigurationCustomization{
```
diviner524, line 246: Is MutatingWebhookCRForDuplicatedWebhook used somewhere? I cannot find it.

jingyih: Good catch! I forgot to add a unit test case.

operator/pkg/controllers/configconnector/configconnector_controller.go
```diff
@@ -691,6 +691,9 @@ func (r *ConfigConnectorReconciler) fetchAndApplyAllWebhookConfigurationCustomiz
 
 // applyValidatingWebhookConfigurationCustomizationCR applies customizations specified in a ValidatingWebhookConfigurationCustomization CR.
 func (r *ConfigConnectorReconciler) applyValidatingWebhookConfigurationCustomizationCR(ctx context.Context, cr *customizev1alpha1.ValidatingWebhookConfigurationCustomization) error {
+	if err := checkForDuplicateWebhooks(cr.Spec.Webhooks); err != nil {
```
justinsb, line 694: Nit: we may want to use the "apimachinery" validation libraries - gathering all the errors.  It means the errors are presented more consistently.  Example PR that makes this change in another project: https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3026/files#diff-93dbac8d674a0494b5f802a4b57b0a42b92e25196b8b31c2fc54db8ceaf22562

operator/pkg/controllers/configconnector/configconnector_controller.go
```diff
@@ -720,6 +723,9 @@ func (r *ConfigConnectorReconciler) applyValidatingWebhookConfigurationCustomiza
 
 // applyMutatingWebhookConfigurationCustomizationCR applies customizations specified in a MutatingWebhookConfigurationCustomization CR.
 func (r *ConfigConnectorReconciler) applyMutatingWebhookConfigurationCustomizationCR(ctx context.Context, cr *customizev1alpha1.MutatingWebhookConfigurationCustomization) error {
+	if err := checkForDuplicateWebhooks(cr.Spec.Webhooks); err != nil {
```
justinsb, line 726: IIUC, a bad webhook will "stop the world" e.g. we won't upgrade KCC.  I think that could be a valid decision, but it feels like a big blast radius.  Is my understanding correct?

jingyih: No it will not "stop the world". A bad webhook customization CR means it won't be applied by the operator, and the operator reports an error in the CR status.

operator/pkg/controllers/configconnector/configconnector_controller.go
jingyih: Good to know! Thank you! Although for this case there is only one error. If there are duplicated webhooks in the CR, I don't think we should continue to apply it. So the code returns on the first error. Do you suggest that we should use `field.ErrorList{}` even if there is only one error?

PR #897, opened 2023-10-06
Graduate customization CRDs to v1beta1

jingyih:
### Change description

 - Add v1beta1 API, which is the same as the v1alpha1.
 - Mark v1beta1 as the storage version.
 - Mark v1alpha1 as not served version.

### Tests you have done

1. A fresh install of KCC v1.1110.0 (which has v1alpha1 customziation CRD)

```
$ git checkout v1.110.0

$ cd operator

$ git checkout upstream/master Makefile
(For convenience, we need newly added Makefile rules from HEAD)

$ make uninstall destroy

$ make install docker-build docker-push deploy
```

2. Create a CC object
```
$ k apply -f cc.yaml 
configconnector.core.cnrm.cloud.google.com/configconnector.core.cnrm.cloud.google.com created
```

3. Create v1alpha1 customization CRs
```
$ k apply -f config/samples/controller_resource_customization_sample.yaml 
controllerresource.customize.core.cnrm.cloud.google.com/cnrm-controller-manager created
controllerresource.customize.core.cnrm.cloud.google.com/cnrm-webhook-manager created

$ k get controllerresources
NAME                      AGE
cnrm-controller-manager   9s
cnrm-webhook-manager      9s
```

4. Verify the customization is applied
```
$ k get deployment cnrm-webhook-manager -n cnrm-system -o jsonpath='{range .spec.template.spec.containers[*]}[{.name}, {.resources}]{"\n"}{end}' 
[webhook, map[limits:map[memory:512Mi] requests:map[cpu:250m memory:256Mi]]]

$ k get pods -n cnrm-system (4 replicas for webhook manager pod)
NAME                                           READY   STATUS    RESTARTS   AGE
cnrm-controller-manager-0                      2/2     Running   0          20s
cnrm-deletiondefender-0                        1/1     Running   0          107s
cnrm-resource-stats-recorder-b766c4fd6-nskvr   2/2     Running   0          108s
cnrm-webhook-manager-d5887dc87-6npkj           1/1     Running   0          19s
cnrm-webhook-manager-d5887dc87-7wb4b           1/1     Running   0          11s
cnrm-webhook-manager-d5887dc87-t5gwz           1/1     Running   0          31s
cnrm-webhook-manager-d5887dc87-wpk92           1/1     Running   0          31s
```

5. Install operator and CRDs from this PR (This step mimics the KCC version upgrade)
```
$ git checkout v1beta1_customization_CRDs

$ make install docker-build docker-push deploy
```

6. Verify the previously created customization CRs still exist
```
$ k get controllerresources
Error from server (NotFound): Unable to list "customize.core.cnrm.cloud.google.com/v1alpha1, Resource=controllerresources": the server could not find the requested resource (get controllerresources.customize.core.cnrm.cloud.google.com)

$ k get controllerresources.customize.core.cnrm.cloud.google.com
NAME                      AGE
cnrm-controller-manager   4m46s
cnrm-webhook-manager      4m46s
```

7. Verify the customization is still effective
```
$ k get deployment cnrm-webhook-manager -n cnrm-system -o jsonpath='{range .spec.template.spec.containers[*]}[{.name}, {.resources}]{"\n"}{end}' 
[webhook, map[limits:map[memory:512Mi] requests:map[cpu:250m memory:256Mi]]]

$ k get pods -n cnrm-system
NAME                                           READY   STATUS    RESTARTS   AGE
cnrm-controller-manager-0                      2/2     Running   0          4m41s
cnrm-deletiondefender-0                        1/1     Running   0          6m8s
cnrm-resource-stats-recorder-b766c4fd6-nskvr   2/2     Running   0          6m9s
cnrm-webhook-manager-d5887dc87-6npkj           1/1     Running   0          4m40s
cnrm-webhook-manager-d5887dc87-7wb4b           1/1     Running   0          4m32s
cnrm-webhook-manager-d5887dc87-t5gwz           1/1     Running   0          4m52s
cnrm-webhook-manager-d5887dc87-wpk92           1/1     Running   0          4m52s
``` 

8. The existing CRs can be read via v1beta1
```
$ k get -f config/samples/controller_resource_customization_sample.yaml
NAME                      AGE
cnrm-controller-manager   7m7s
cnrm-webhook-manager      7m7s
```

9. v1alpha1 is no longer served
```
$ git checkout v1.110.0

$ k get -f config/samples/controller_resource_customization_sample.yaml
unable to recognize "config/samples/controller_resource_customization_sample.yaml": no matches for kind "ControllerResource" in version "customize.core.cnrm.cloud.google.com/v1alpha1"
unable to recognize "config/samples/controller_resource_customization_sample.yaml": no matches for kind "ControllerResource" in version "customize.core.cnrm.cloud.google.com/v1alpha1"
```


operator/config/crd/bases/customize.core.cnrm.cloud.google.com_controllerresources.yaml
```diff
@@ -16,6 +16,107 @@ spec:
   scope: Cluster
   versions:
   - name: v1alpha1
+    schema:
```
justinsb, line 19: Nit (no need to fix, and certainly not in this PR): I like to either split out the generated files into a separate commit, or mark them as generated using the approach in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/864

operator/pkg/apis/core/customize/v1beta1/controllerresource_types.go
```diff
@@ -0,0 +1,81 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   ControllerResourceSpec   `json:"spec"`
+	Status ControllerResourceStatus `json:"status,omitempty"`
+}
+
+// ControllerResourceSpec is the specification of the resource customization for containers of
+// a config connector controller.
+type ControllerResourceSpec struct {
```
justinsb, line 39: Nit (more of a general thought than a blocker): We don't have any way of enforcing that v1alpha1 == v1beta1 or that it round-trips (e.g. when we add a field here, will we remember to add it to v1alpha1?).  Perhaps we don't care about adding fields to v1alpha1 if it is unserved.

But should we include a round-trip test or another trick?  cc @apelisse

justinsb:
I know this is WIP, but it LGTM :-)

operator/pkg/apis/core/customize/v1beta1/controllerresource_types.go
apelisse: I think you're supposed to do a round-trip test, but I don't know how easy we make it for people to do that. I always felt like if that's a requirement of the api, controller-runtime/kubebuilder should generate a test for you and run it.

jingyih: Thank you @apelisse, is there an example round-trip test you can point me to for reference?

maqiuyujoyce:
Thanks a lot for the PR, @jingyih ! LGTM! Not approving yet as it seems that there is on-going discussion about round-trip test.

diviner524 (commented):
Looks good, except one question, I noticed in the manual testing, we first did:

`$ make uninstall destroy`.

I think we cannot assume our manual installation users (or even Config Controller) will do that when they upgrade KCC version?

If they don't have this step, will they run into any issues? For example a mismatch of operator version versus the CRD version? Maybe the operator hasn't been updated, while the CRD is no longer serving v1alpha1.

jingyih:
> Looks good, except one question, I noticed in the manual testing, we first did:
> 
> `$ make uninstall destroy`.
> 
> I think we cannot assume our manual installation users (or even Config Controller) will do that when they upgrade KCC version?
> 
> If they don't have this step, will they run into any issues? For example a mismatch of operator version versus the CRD version? Maybe the operator hasn't been updated, while the CRD is no longer serving v1alpha1.

The step is to mimic a fresh install of upstream/master code (which has v1alpha1 CRD). It is technically not part of the upgrade? (my understanding of the upgrade starts with step `Build operator and CRDs from PR`)

diviner524:
> > Looks good, except one question, I noticed in the manual testing, we first did:
> > `$ make uninstall destroy`.
> > I think we cannot assume our manual installation users (or even Config Controller) will do that when they upgrade KCC version?
> > If they don't have this step, will they run into any issues? For example a mismatch of operator version versus the CRD version? Maybe the operator hasn't been updated, while the CRD is no longer serving v1alpha1.
> 
> The step is to mimic a fresh install of upstream/master code (which has v1alpha1 CRD). It is technically not part of the upgrade? (my understanding of the upgrade starts with step `Build operator and CRDs from PR`)

You are right! I misunderstood the intent of this step. Then I think we should be good.

jingyih:
/hold

Need to rebase the PR to resolve conflict. I am also looking into the possibility of writing a roundtrip test for "not served" CRD version.

jingyih:
/hold cancel

Rebased PR.

Not easy to do roundtrip test in unit test, since the serialization and deserialization happens in apiserver.

jingyih:
> > > Looks good, except one question, I noticed in the manual testing, we first did:
> > > `$ make uninstall destroy`.
> > > I think we cannot assume our manual installation users (or even Config Controller) will do that when they upgrade KCC version?
> > > If they don't have this step, will they run into any issues? For example a mismatch of operator version versus the CRD version? Maybe the operator hasn't been updated, while the CRD is no longer serving v1alpha1.
> > 
> > 
> > The step is to mimic a fresh install of upstream/master code (which has v1alpha1 CRD). It is technically not part of the upgrade? (my understanding of the upgrade starts with step `Build operator and CRDs from PR`)
> 
> You are right! I misunderstood the intent of this step. Then I think we should be good.

I updated the description to make the test steps clearer.

PR #898, opened 2023-10-06
Mark CertificateManager group as supported in mockgcp

justinsb:
We can start to mark whole groups as being supported now in mockgcp, rather than doing it for each type.

### DIFF
--- a/config/tests/samples/create/harness.go
+++ b/config/tests/samples/create/harness.go
@@ -291,38 +291,30 @@ func MaybeSkip(t *testing.T, name string, resources []*unstructured.Unstructured
 	if os.Getenv("E2E_GCP_TARGET") == "mock" {
 		for _, resource := range resources {
 			gvk := resource.GroupVersionKind()
-			switch gvk.GroupKind() {
-			case schema.GroupKind{Group: "certificatemanager.cnrm.cloud.google.com", Kind: "CertificateManagerCertificate"}:
-			case schema.GroupKind{Group: "certificatemanager.cnrm.cloud.google.com", Kind: "CertificateManagerCertificateMap"}:
-			case schema.GroupKind{Group: "certificatemanager.cnrm.cloud.google.com", Kind: "CertificateManagerDNSAuthorization"}:
-				// ok
 
+			switch gvk.Group {
+			case "certificatemanager.cnrm.cloud.google.com":
+				continue
+			}
+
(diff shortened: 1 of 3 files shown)


justinsb:
Builds on #859.  I think they're all supported, but we can let CI check while I do other things :-)

justinsb:
Now that #859 has merged, I think this is no longer WIP

config/tests/samples/create/harness.go
```diff
@@ -278,34 +278,37 @@ func MaybeSkip(t *testing.T, name string, resources []*unstructured.Unstructured
...
-				// ok
 			case schema.GroupKind{Group: "secretmanager.cnrm.cloud.google.com", Kind: "SecretManagerSecretVersion"}:
-				// ok
+				supported = true
 
 			case schema.GroupKind{Group: "", Kind: "Secret"}:
-				// ok
+				supported = true
 
 			case schema.GroupKind{Group: "serviceusage.cnrm.cloud.google.com", Kind: "Service"}:
```
cheftako, line 306: If we're removing default, why bother keeping in this case with no code in it? Seems less confusing to remove the "Service" case.

config/tests/samples/create/harness.go
```diff
@@ -278,34 +278,37 @@ func MaybeSkip(t *testing.T, name string, resources []*unstructured.Unstructured
...
+
+			switch gvk.Group {
+			case "certificatemanager.cnrm.cloud.google.com":
+				supported = true
+			}
+
+			switch gvk.GroupKind() {
 			case schema.GroupKind{Group: "iam.cnrm.cloud.google.com", Kind: "IAMServiceAccount"}:
-				// ok
+				supported = true
```
cheftako, line 291: Seems easier to use [continue](https://go.dev/play/p/KlMkFH_wtpF) rather than futzing with the supported flag.

config/tests/samples/create/harness.go
justinsb: Not sure I follow what you're saying here

justinsb: Oh I get it - I was thinking about fallthrough, but we don't have fallthrogh

justinsb: I think removing the ok comments makes the logic clearer, without implying fallthrough

PR #901, opened 2023-10-11
mockgcp: update README to define "good enough" criteria

justinsb:
Tackling the question of how accurate we want the mock to be, focusing
on the journey over perfection from the outset.


jingyih (commented):
/lgtm
/approve

PR #911, opened 2023-10-12
mockgcp: support for IAM policies

justinsb:
Very minimal support, aiming to unblock testing.


mockgcp/mock_http_roundtrip.go
```diff
@@ -184,10 +187,60 @@ func (m *mockRoundTripper) modifyUpdateMask(s string) (string, error) {
...
+	requestPath := req.URL.Path
+
+	lastColon := strings.LastIndex(requestPath, ":")
+	verb := requestPath[lastColon+1:]
+
+	requestPath = strings.TrimSuffix(requestPath, ":"+verb)
+
+	switch verb {
+	case "getIamPolicy":
+		if req.Method == "GET" || req.Method == "POST" {
```
gemmahou, line 203: Just curious why we support "POST" for getIamPolicy? I think it should be "GET" only

justinsb: Good question.  Looks like it might not be very consistent across services.

I'm starting with cloudresourcemanager (because I'm trying to unblock projects, as that unblocks a whole bunch of other tests).  That one uses POST: https://cloud.google.com/resource-manager/reference/rest/v3/projects/getIamPolicy

Billing accounts (for example) use GET: https://cloud.google.com/billing/docs/reference/rest/v1/billingAccounts/getIamPolicy

PR #915, opened 2023-10-14
e2e: speed up tests

justinsb:
We share a kube-apiserver instance rather than bringing it up for each
test, and we reduce some critical polling intervals.


justinsb:
Given the large improvements in speed with the mocks here, removing WIP

PR #916, opened 2023-10-15
Add mockGCP support for container attached cluster resource

gemmahou:
Add mockGCP support for container attached cluster resource
b/297913069


mockgcp/mock_http_roundtrip.go
```diff
@@ -77,6 +78,7 @@ func NewMockRoundTripper(t *testing.T, k8sClient client.Client, storage storage.
 	var services []MockService
 
 	services = append(services, mockcertificatemanager.New(env, storage))
+	services = append(services, mockcontainerattached.New(env, storage))
```
justinsb, line 81: Nit: I've been naming the package after the service name, i.e. in this case `gkemulticloud` ... it doesn't really matter, it just stops us having to think of our own names :-)

gemmahou: Yes it's a bit confusing, the cnrm resource name is [ContainerAttached](https://source.corp.google.com/h/cnrm/cnrm/+/main:config/servicemappings/containerattached.yaml;l=21), while the api host name is [gkemuticloud](https://cloud.google.com/anthos/clusters/docs/multi-cloud/reference/rest).

PR #928, opened 2023-10-16
mockgcp: replace uuid with apimachinery uuid

justinsb:
It's a wrapper around the same library, but it avoids a go.mod
dependency - and we're already using apimachinery.


jingyih (commented):
/lgtm

PR #929, opened 2023-10-16
Enable dependabot for operator Dockerfile

jingyih:
Follow up of #910. Enable dependabot on operator/Dockerfile as well.


justinsb:
Good spot - thanks!

/approve
/lgtm

PR #930, opened 2023-10-16
Tell dependabot to ignore some k8s releases

justinsb:
This will let us pick up patch releases and remind us when 1.29 is
released.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,10 +18,22 @@ updates:
     directory: /
     schedule:
       interval: daily
+    ignore:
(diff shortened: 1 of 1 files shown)


justinsb:
I can't see a good way to test this before merging.  I took the approach from the kueue project: https://github.com/kubernetes-sigs/kueue/blob/5ab997d26a1d0a1d433327357eb95a8a0f85a492/.github/dependabot.yml#L5

cheftako:
YOLO
/lgtm
/approve

PR #935, opened 2023-10-16
tests: update urlMap tests

justinsb:
Some samples/fixtures were using a field that isn't valid against the
schema.


justinsb:
cc @diviner524 I'm not entirely sure how this happened - maybe we changed the schema at some stage from alpha -> beta?

jingyih (commented):
/lgtm

PR #937, opened 2023-10-17
Minor fix: add debug message into mock_http_roundtrip

gemmahou:
Minor fix: add clear debug message into mock_http_roundtrip when provided host name is invalid.


justinsb:
Thanks @gemmahou  - I like the hint for the developer here also

/approve
/lgtm

PR #943, opened 2023-10-18
Experiments: Implement basic reconciler for github

justinsb:
Starting with just github repos, and just managing tag protections.

The intent is to explore what we can do for things beyond just GCP
APIs, and whether there are patterns that make that much easier.

Building up the functionality gradually here, rather than trying to do
everything all at once.


cheftako:
A bit short on tests and docs but it seems like a great start.

justinsb:
> A bit short on tests

Agreed.  I'm torn here - I think our mockgcp approach works well, but do we want to force people to create a mock for each API they are targeting?  It would be great if we could figure out a good approach here.

justinsb:
For testing, in #957 I started working on recording the actual requests made in tests.  I think that could be a reasonable option, it works with a mock, but it also works with the real APIs when people don't want to implement mocks.

cc @diviner524

justinsb:
I think we can close this one now

/close

PR #951, opened 2023-10-19
tests: if we run the controller once then the object does not converge

justinsb:
Because we only run the reconcile method of the controller once in
some tests, we can't expect that the object has converged fully
(observedGeneration == generation); that needs another reconciliation.
In these tests, we wait only for the observedGeneration to include the
generation before reconciliation.


diviner524:
Verified with one failing test case and it is now passing:

--- PASS: TestCreateNoChangeUpdateDelete (1.33s)
    --- PASS: TestCreateNoChangeUpdateDelete/basic-basicalloydbinstance (1975.80s)

PR #955, opened 2023-10-20
vet tags

justinsb:
- tests: wire in context to the next "wave" of functions
- WIP: Verify code that runs with tags


justinsb:
OK, I was worried that we couldn't detect failures in errors with build tags, but looks like we can :-)

PR #957, opened 2023-10-20
Record and verify HTTP requests

justinsb:
~WIP: Builds on #870.~

Demonstrates how we can export and verify the HTTP requests using golden-testing.

One gotcha is that sometimes we poll for resources, so we need to implement some basic request normalization (e.g. collapsing redundant queries for the same object)

### DIFF
--- a/mockgcp/mockiam/serviceaccounts.go
+++ b/mockgcp/mockiam/serviceaccounts.go
@@ -16,6 +16,7 @@ package mockiam
 
 import (
 	"context"
+	"crypto/md5"
 	"regexp"
 	"strconv"
 	"time"
@@ -25,6 +26,7 @@ import (
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/emptypb"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/klog/v2"
 
 	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/common/projects"
 	pb "github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/generated/mockgcp/iam/admin/v1"
@@ -59,7 +61,7 @@ func (s *ServerV1) GetServiceAccount(ctx context.Context, req *pb.GetServiceAcco
 		}
 
 		if found == nil {
-			return nil, status.Errorf(codes.NotFound, "serviceaccount %q not found", req.Name)
+			return nil, status.Errorf(codes.NotFound, "Service account %q not found", req.Name)
 		}
 
 		return found, nil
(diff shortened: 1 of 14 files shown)


justinsb:
@gemmahou this might be of interest to you

justinsb:
Next chunk split out into #1098, then I think this one can be rebased and ready-for-review.

justinsb:
We should review/merge #1098 first, but I think we can review this one concurrently now that most of the commits have been merged separately.

/assign @gemmahou

mockgcp/mockiam/serviceaccounts.go
```diff
@@ -59,7 +61,8 @@ func (s *ServerV1) GetServiceAccount(ctx context.Context, req *pb.GetServiceAcco
 		}
 
 		if found == nil {
-			return nil, status.Errorf(codes.NotFound, "serviceaccount %q not found", req.Name)
+			return nil, status.Errorf(codes.NotFound, "Service account %q not found", req.Name)
+			// return nil, status.Errorf(codes.NotFound, "Unknown service account")
```
gemmahou, line 65: remove line

mockgcp/mockiam/serviceaccounts.go
```diff
@@ -69,7 +72,8 @@ func (s *ServerV1) GetServiceAccount(ctx context.Context, req *pb.GetServiceAcco
 	fqn := name.String()
 	if err := s.storage.Get(ctx, fqn, sa); err != nil {
 		if apierrors.IsNotFound(err) {
-			return nil, status.Errorf(codes.NotFound, "serviceaccount %q not found", req.Name)
+			return nil, status.Errorf(codes.NotFound, "Service account %q not found", req.Name)
+			// return nil, status.Errorf(codes.NotFound, "Unknown service account")
```
gemmahou, line 76: remove line

tests/e2e/unified_test.go
```diff
@@ -186,7 +163,58 @@ func TestAllInSeries(t *testing.T) {
 					h.CompareGoldenFile(expectedPath, string(output), h.IgnoreComments, h.ReplaceString(project.ProjectID, "example-project-id"))
 				}
 
-				create.DeleteResources(h, s.Resources)
+				create.DeleteResources(h, opt.Create)
+
+				// Verify events against golden file
+				if os.Getenv("GOLDEN_REQUEST_CHECKS") != "" {
```
gemmahou, line 169: nit: just a thought no need to make changes rn. I feel instead of saying `env != ""`, it's better to say `env == true`.

justinsb: The gotcha is that it's a string, but I agree it could be confusing, I'm just not sure what we should accept.  Maybe "0" and "1", and then being strict and failing the test if the value is specified but not 0 or 1?  We could wrap that logic into a helper function and evolve it.

This is "only test code", it probably is more important to think through the ergonomics if we support "feature flags" in the production codepaths.  But yes, I don't think this is perfect, I just don't know what's better!

mockgcp/mockiam/serviceaccounts.go
justinsb: Done!

mockgcp/mockiam/serviceaccounts.go
justinsb: Done!

justinsb:
Resolved merge conflict, PTAL @gemmahou

gemmahou:
Still need to resolve merge conflict since we have https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1190 merged just now.

gemmahou (approved):

PR #958, opened 2023-10-23
refactor: stop using inject framework in controller-runtime

justinsb:
The inject framework has been removed, so we need this to continue to
update controller-runtime.


diviner524:
> The inject framework has been removed, so we need this to continue to update controller-runtime.

Thank you for making the change @justinsb! Just for self-education, could you also share a link with more details on the removal of inject framework?

justinsb:
I think the canonical reference for inject deprecation is https://github.com/kubernetes-sigs/controller-runtime/issues/1433

What really breaks us though is that https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/runtime/inject is removed after 0.14.6, that stops (future) updates

PR #959, opened 2023-10-23
tests: Support mock-kubeapiserver instead of envtest

justinsb:
This is much faster than envtest, at the expense of being less accurate (most immediately relevant to us: webhooks are not supported).

Builds on #958


justinsb:
/close

We merged this elsewhere

PR #960, opened 2023-10-23
tests: pass project explicitly to test methods

justinsb:
Embedding into SystemContext did not ensure it was always initialized.


justinsb:
This aims to fix the breakages on the "real" runs we observed.  (And #962 is working towards being able to run some of those tests against mocks)

PR #966, opened 2023-10-25
versionbump: bump github.com/google/uuid

justinsb:
Dependabot identified this, but only bumped mockgcp.  Trying to bump
both at once, I think the problem is that we are at the update limit
(maybe?)


justinsb:
This was identified by dependabot in #923

PR #975, opened 2023-10-27
mockgcp: support for projects

justinsb:
Add support for projects.  Some complexity because the version of projects terraform uses doesn't have a published proto.  Also implement a bit of the billing because projects try to set billing.


justinsb:
Likely blocked on #911 (includes it as the first 2 commits!)

justinsb:
No longer blocked on #911, as that was merged (thanks @gemmahou )

Marking as no longer WIP

pkg/test/gcp/gcp.go
```diff
@@ -52,15 +52,17 @@ func (v *EnvVar) Set(s string) {
 }
 
 var (
+	TestFolderID                        = EnvVar{Key: "TEST_FOLDER_ID"}
+	TestFolder2ID                       = EnvVar{Key: "TEST_FOLDER_2_ID"}
 	TestOrgID                           = EnvVar{Key: "TEST_ORG_ID"}
 	TestBillingAccountID                = EnvVar{Key: "TEST_BILLING_ACCOUNT_ID"}
 	IAMIntegrationTestsOrganizationID   = EnvVar{Key: "IAM_INTEGRATION_TESTS_ORGANIZATION_ID"}
 	IAMIntegrationTestsBillingAccountID = EnvVar{Key: "IAM_INTEGRATION_TESTS_BILLING_ACCOUNT_ID"}
+	TestAttachedClusterName             = EnvVar{Key: "TEST_ATTACHED_CLUSTER_NAME"}
```
gemmahou, line 61: I think I forgot to delete this variable. This is no longer used anywhere.

tests/e2e/unified_test.go
```diff
@@ -41,31 +40,50 @@ func TestAllInSeries(t *testing.T) {
...
-		}
 		// Some fixed-value fake org-ids for testing.
 		// We used fixed values so that the output is predictable (for golden testing)
+		testgcp.TestFolderID.Set("123451001")
+		testgcp.TestFolder2ID.Set("123451002")
 		testgcp.TestOrgID.Set("123450001")
 		testgcp.TestBillingAccountID.Set("123456-777777-000001")
 		testgcp.IAMIntegrationTestsOrganizationID.Set("123450002")
 		testgcp.IAMIntegrationTestsBillingAccountID.Set("123456-777777-000002")
+		testgcp.TestAttachedClusterName.Set("xks-cluster")
```
gemmahou, line 61: We can remove this too.

mockgcp/mockbilling/names.go
```diff
@@ -0,0 +1,77 @@
...
+
+		name := &projectName{
+			Project: project,
+		}
+
+		return name, nil
+	} else {
+		return nil, status.Errorf(codes.InvalidArgument, "name %q is not valid", name)
+	}
+}
```
gemmahou, line 53: Maybe it's better to move above functions into mockresourcemanager/names.go?

pkg/test/gcp/gcp.go
justinsb: Thanks - removed!

tests/e2e/unified_test.go
justinsb: Done!

pkg/test/gcp/gcp.go
justinsb: Actually, looks like we still use ATTACHED_CLUSTER_NAME in the samples (which we also run / can run in the unified tests)

mockgcp/mockbilling/names.go
justinsb: Good call - done!

pkg/test/gcp/gcp.go
gemmahou: Samples test for this resource is disabled: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/config/tests/samples/create/samples_test.go#L69-L73.

gemmahou (approved):

PR #976, opened 2023-10-27
tests: reduce log verbosity

justinsb:
Set some of the informational messages that are repeated a lot to V(2).


jingyih (commented):
/lgtm

PR #977, opened 2023-10-27
Experiment: Implement trivial composite operator

justinsb:
KCC goes well with compositions!


experiments/composite/pkg/commonoperator/entrypoint.go
```diff
@@ -0,0 +1,130 @@
...
+		}
+	}
+
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		Scheme:                 scheme,
+		MetricsBindAddress:     metricsAddr,
+		Port:                   9443,
+		HealthProbeBindAddress: probeAddr,
+		LeaderElection:         enableLeaderElection,
+		LeaderElectionID:       "3d22f779.k8s.io",
```
cheftako, line 101: This ID seems a bit unfriendly. Why not use something like "experiments.cnrm.cloud.google.com"?

experiments/composite/pkg/controllers/future.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func (f *Future[T]) Wait() T {
+	f.mutex.Lock()
+	for {
+		if !f.done {
+			f.cond.Wait()
+		} else {
+			result := f.result
+			f.mutex.Unlock()
```
cheftako, line 57: Seems safer to defer unlock right after the lock.

experiments/composite/pkg/watchset/dependencies.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+}
+
+var _ fmt.Stringer = dependencySet{}
+
+func (x *DependencySet) String() string {
+	var sb strings.Builder
+	for gvk, objects := range x.objects {
+		if sb.Len() != 0 {
+			fmt.Fprintf(&sb, ",")
```
cheftako, line 51: Why Fprintf rather than sb.WriteString?

experiments/composite/pkg/watchset/manager.go
```diff
@@ -0,0 +1,225 @@
...
+
+	// gvk schema.GroupVersionKind
+	gvr schema.GroupVersionResource
+
+	cancel func()
+}
+
+func (w *gvrWatcher) watchForever(ctx context.Context) error {
+	for {
+		if ctx.Err() != nil {
```
cheftako, line 147: Might be nice to also be able to break out on ctx.Done so we can actually implement a clean shut down. (Yeah I know we're currently using leader election but . . . )

experiments/composite/pkg/controllers/future.go
justinsb: Oooh - nice simplification, you're right - it reads much better now!

experiments/composite/pkg/watchset/dependencies.go
justinsb: So in other places we do this:

`fmt.Fprintf(&sb, "%v@%d", nn.String(), rv)`

Sadly there's some sb.WriteStringF, so we would have to do `sb.WriteString(fmt.Sprintf("%v@%d", nn.String(), rv))`

But then this was trying to be consistent and not alternate between sb.WriteString and fmt.Fprintf.

I'm honestly not sure which is better... I can see a case for any of mixing or WriteString everywhere or Fprintf everywhere.

experiments/composite/pkg/watchset/manager.go
justinsb: We do :-)  If the ctx is done, ctx.Err() is guaranteed to return non-nil.

I did replace time.Sleep(5 * time.Second) with a sleep-with-context function.

experiments/composite/pkg/commonoperator/entrypoint.go
justinsb: Good point - this was automatically generated by controller-runtime and I left it in accidentally!  Moved it and switched to a sensible default.

experiments/composite/pkg/watchset/dependencies.go
cheftako: That makes sense. Given the API you have to work with.

PR #979, opened 2023-10-28
Add EdgenetworkNetwork and EdgenetworkSubnet resources

f1urps:
### Change description

Adds resources for Edge Network API.
- EdgenetworkNetwork
- EdgenetworkSubnet

### Tests you have done

Edge Network API has unique requirements such that integration tests are not possible without a mock.


jingyih (commented):
May I suggest that for better readability, we could:
 - split the first commit into multiple smaller commits. For example, 1 commit for modifying servicemappings and rbacs, etc; 1 commit for the generated files; 1 commit for adding tests and samples.
 - instead of having 2 separate commits for "fixing merge conflicts", directly amend the previous commit(s) and do a git force push.

f1urps:
@jingyih I've split up the commit history as you requested. Please review.

config/servicemappings/edgenetwork.yaml
```diff
@@ -0,0 +1,75 @@
...
+        - tfField: project
+          key: projectRef
+          description: |-
+            The project that this resource belongs to.
+          gvk:
+            kind: Project
+            version: v1beta1
+            group: resourcemanager.cnrm.cloud.google.com
+    - name: google_edgenetwork_subnet
+      kind: EdgenetworkSubnet
```
jingyih, line 48: Same here. `EdgeNetworkSubnet`

config/servicemappings/edgenetwork.yaml
```diff
@@ -0,0 +1,75 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: core.cnrm.cloud.google.com/v1alpha1
+kind: ServiceMapping
+metadata:
+  name: edgenetwork.cnrm.cloud.google.com
+  namespace: cnrm-system
+spec:
+  name: Edgenetwork
```
jingyih, line 21: I think this should be `EdgeNetwork` based on my readings on other servicemapping files. Although I did not find a rule specific for this field. cc @maqiuyujoyce

config/tests/samples/create/samples_test.go
```diff
@@ -64,6 +64,8 @@ var testDisabledList = map[string]bool{
 	// The resources below have special test requirements
 	"computeinterconnectattachment": true,
 	"firestoreindex":                true,
+	"edgenetworknetwork":            true,
```
jingyih, line 67: I am missing the context. Why are these disabled?

config/samples/resources/edgenetworksubnet/edgenetwork_v1beta1_edgenetworksubnet.yaml
```diff
@@ -0,0 +1,28 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: edgenetwork.cnrm.cloud.google.com/v1beta1
+kind: EdgenetworkSubnet
+metadata:
+  name: edgenetworknetwork-sample
```
jingyih, line 18: `edgenetworksubnet-sample`

config/samples/resources/edgenetworksubnet/edgenetwork_v1beta1_edgenetworknetwork.yaml
```diff
@@ -0,0 +1,24 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: edgenetwork.cnrm.cloud.google.com/v1beta1
+kind: EdgenetworkNetwork
+metadata:
+  name: edgenetworknetwork-dep
```
jingyih, line 18: `edgenetworksubnet-dep`

config/samples/resources/edgenetworknetwork/edgenetwork_v1beta1_edgenetworknetwork.yaml
```diff
@@ -0,0 +1,24 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: edgenetwork.cnrm.cloud.google.com/v1beta1
+kind: EdgenetworkNetwork
+metadata:
+  name: edgenetworknetwork-sample
+spec:
+  description: "A sample edgenetwork network"
+  location: sample-location
```
jingyih, line 21: Could you use real values for location and zone?

config/samples/resources/edgenetworksubnet/edgenetwork_v1beta1_edgenetworksubnet.yaml
```diff
@@ -0,0 +1,28 @@
...
+spec:
+  description: "A sample edgenetwork subnet"
+  location: sample-location
+  zone: sample-zone
+  networkRef:
+    name: edgenetworknetwork-dep
```
jingyih, line 24: `edgenetworksubnet-dep`

jingyih (commented):
Thank you for your contribution. Overall looks good! I added a few comments.

config/servicemappings/edgenetwork.yaml
```diff
@@ -0,0 +1,75 @@
...
+  version: v1beta1
+  serviceHostName: autopush-edgenetwork.sandbox.googleapis.com
```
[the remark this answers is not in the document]
maqiuyujoyce: We suggest using the prod hostname of the API. Note that this is not to specify which endpoint it talks to when making an HTTP request.

config/servicemappings/edgenetwork.yaml
maqiuyujoyce: We suggest the service name be the same as the hostname of the prod endpoint of the API following the Pascal Case, and the split of the words (either it's "Edge Network" or "Edgenetwork") should probably follow the branding of the product (and the readability of the name). According to my brief research, there is no occurrences of Edgenetwork in the public doc, so I'd suggest it should be `EdgeNetwork` for service name instead.

config/servicemappings/edgenetwork.yaml
```diff
@@ -0,0 +1,75 @@
...
+  resources:
+    - name: google_edgenetwork_network
+      kind: EdgenetworkNetwork
```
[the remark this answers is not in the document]
maqiuyujoyce: Please find my comment above. @jingyih could you add an example / more details to the doc if users find it confusing?

config/servicemappings/edgenetwork.yaml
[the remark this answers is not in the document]
f1urps: Oops, I was trying to do some integration testing on a different endpoint and forgot to change this back. Fixed.

Sidenote -- Is there a better way to override the API endpoint that the config controller uses?

config/tests/samples/create/samples_test.go
f1urps: Edge Network API is a special case among google cloud APIs in that it must connect to a special hardware deployment, an "edge zone", to create resources. (See: https://cloud.google.com/distributed-cloud/edge/latest/docs/how-it-works) Internal to Google, we can provision edge zones for testing purposes, but they can't be exposed here. Therefore the only possible integration test for Edge Network API is with a mock.

In magic-modules, it was decided to disable the integration tests for edgenetwork because they have no mocking framework. (see comments here: https://github.com/GoogleCloudPlatform/magic-modules/pull/8905)

config/servicemappings/edgenetwork.yaml
f1urps: Done, renamed everything from Edgenetwork -> EdgeNetwork.

config/samples/resources/edgenetworknetwork/edgenetwork_v1beta1_edgenetworknetwork.yaml
f1urps: Cloud locations are well-known, but edge zone names are private. I've replaced these with a real location and a fake (but realistic) zone name.

jingyih (commented):
Could you fix the mock test failures? Thanks!

Otherwise looks good to me.

jingyih (commented):
/lgtm

PR #980, opened 2023-10-30
Add one more test case for Attached Cluster and fix link in doc

gemmahou:
### Change description

1. Added one more test case with test EKS cluster k8s version 1.27(The highest version supported by Multiclould API). AWS will stop supporting version 1.25 on May 2024. After this PR checked in, we should consider remove test version 1.25. 

2. Fixed the link in our documentation.

Fixes b/307542957

### Tests you have done
`go test -v -tags=integration ./pkg/controller/dynamic/ -test.run TestCreateNoChangeUpdateDelete -run-tests containerattached -timeout 900s`

Dynamic tests passed locally: https://screenshot.googleplex.com/47B4iYbTH9LiY8q.png


maqiuyujoyce:
Thank you for the change, @gemmahou ! Though I am curious - why don't we replace the existing test case with this one? Is there any reason we want to keep the 1.25 one?

gemmahou:
> Thank you for the change, @gemmahou ! Though I am curious - why don't we replace the existing test case with this one? Is there any reason we want to keep the 1.25 one?

The 1.25 one can be removed. But I want to hold until new test submitted, to avoid potential test failures in the Prow job.    Maybe it's fine to remove the old test here as the local test passed.

maqiuyujoyce:
Yes, I believe it's totally fine to just replace it. We can always make further changes if needed.

maqiuyujoyce (approved):

gemmahou:
> Yes, I believe it's totally fine to just replace it. We can always make further changes if needed.

Updated the existing test. 
Note that I created a new test EKS cluster to be used for testing, instead of upgrading the existing one. The reason is, we are still running the old test configuration. I'm afraid if I upgrade the existing EKS cluster directly, current test would fail until this PR takes effect.

PR #983, opened 2023-10-31
AlloyDB: Update/Add test cases and samples to reflect modifications as per TF Provider 4.84.0

nitinchugh-google:
### Change description

Addresses the following for AlloyDB Resources
1. Use networkConfig field instead of networkRef(deprecated) for AlloyDBCluster
2. Add initialUser field to "mutableButNotReadable" section in the servicemapping config for AllouDBCluster
3. Add test case and sample for AlloyDBCluster which is restored from backup
4. Add test case and sample for zonal AlloyDBInstance

Fixes b/308577459

### Tests you have done

Executed make ready-pr

Ran tests for all test cases
go test -v -tags=integration ./pkg/controller/dynamic/ -test.run TestCreateNoChangeUpdateDelete -run-tests alloydb -timeout 7200s
Output Summary of test run at https://paste.googleplex.com/5879544559960064

- [Yes] Run `make ready-pr` to ensure this PR is ready for review.
- [Yes] Perform necessary E2E testing for changed resources.

### DIFF
--- a/config/samples/resources/alloydbcluster/regular-cluster/kms_v1beta1_kmskeyring.yaml
+++ b/config/samples/resources/alloydbcluster/regular-cluster/kms_v1beta1_kmskeyring.yaml
@@ -17,4 +17,4 @@ kind: KMSKeyRing
 metadata:
   name: alloydbcluster-dep-regular
 spec:
-  location: us-central1
+  location: us-east1
--- a/config/samples/resources/alloydbcluster/regular-cluster/kms_v1beta1_kmskeyring.yaml
+++ b/config/samples/resources/alloydbcluster/regular-cluster/kms_v1beta1_kmskeyring.yaml
@@ -4,23 +4,17 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#      http://www.apache.org/licenses/LICENSE-2.0
+#     http://www.apache.org/licenses/LICENSE-2.0
 #
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-# external: projects/${projectId}/locations/us-west1/clusters/alloydbcluster-${uniqueId}
 
-apiVersion: alloydb.cnrm.cloud.google.com/v1beta1
-kind: AlloyDBInstance
+apiVersion: kms.cnrm.cloud.google.com/v1beta1
+kind: KMSKeyRing
 metadata:
-  name: alloydbinstance-${uniqueId}
+  name: alloydbcluster-dep-regular
 spec:
-  clusterRef:
-    name: alloydbcluster-${uniqueId}
-  availabilityType: REGIONAL
-  instanceType: PRIMARY
-  machineConfig:
-    cpuCount: 2
+  location: us-central1
--- a/config/servicemappings/alloydb.yaml
+++ b/config/servicemappings/alloydb.yaml
@@ -67,6 +67,8 @@ spec:
         labels: labels
       resourceID:
         targetField: cluster_id
+      mutableButUnreadableFields:
+        - initial_user
       hierarchicalReferences:
         - type: project
           key: projectRef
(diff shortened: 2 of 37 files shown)


config/samples/resources/alloydbcluster/regular-cluster/kms_v1beta1_kmskeyring.yaml
```diff
@@ -15,6 +15,6 @@
 apiVersion: kms.cnrm.cloud.google.com/v1beta1
 kind: KMSKeyRing
 metadata:
-  name: alloydbcluster-dep
+  name: alloydbcluster-dep-regular
 spec:
-  location: us-central1
\ No newline at end of file
+  location: us-east1
```
diviner524, line 20: Out of curiosity, why do we want to change the location? Does us-east1 offer better capacity for AlloyDB compared to us-central1?

config/servicemappings/alloydb.yaml
```diff
@@ -67,6 +67,8 @@ spec:
         labels: labels
       resourceID:
         targetField: cluster_id
+      mutableButUnreadableFields:
+        - initial_user
```
diviner524, line 71: Is this field supposed to be mutable but unreadable? 

i.e. It is not a bug in the GCP API and we don't have plan to return it in the API response in future?

pkg/test/resourcefixture/contexts/alloydb_context.go
```diff
@@ -25,4 +25,14 @@ func init() {
...
 	}
+
+	resourceContextMap["restorebackupalloydbcluster"] = ResourceContext{
+		ResourceKind: "AlloyDBCluster",
+		SkipUpdate:   true,
+	}
+
+	resourceContextMap["zonalalloydbinstance"] = ResourceContext{
+		ResourceKind: "AlloyDBInstance",
+		SkipUpdate:   true,
```
diviner524, line 36: Are we skipping update because we want to accelerate the test run? Or because update is not supported?

pkg/test/resourcefixture/testdata/basic/alloydb/v1beta1/alloydbcluster/basicalloydbcluster/create.yaml
```diff
@@ -18,8 +18,16 @@ metadata:
   name: alloydbcluster-${uniqueId}
 spec:
   location: us-central1
-  networkRef: 
-    external: projects/${projectId}/global/networks/default
+  networkConfig:
+    networkRef: 
+      name: default
```
diviner524, line 23: Shall we use a non-default network to avoid test failures we have seen before?

diviner524 (commented):
@nitinchugh-google: thanks for the awesome update on the test cases and samples!

Sorry if I missed it somewhere, I see the integration test logs, but could you also share a link to the sample tests result from your local run?

config/samples/resources/alloydbcluster/regular-cluster/kms_v1beta1_kmskeyring.yaml
nitinchugh-google: This is mostly to ensure that while running tests on samples in parallel, we do not run out of quota. As there is a quota on number of clusters per region per project

config/servicemappings/alloydb.yaml
nitinchugh-google: We do not plan to return it in response, as it contains password as well. Also the API documentation clearly mentions that this is an input_only field

pkg/test/resourcefixture/contexts/alloydb_context.go
nitinchugh-google: Because we want to accelerate the test run, as we already have tried updates for other test cases for AlloyDBCluster

pkg/test/resourcefixture/testdata/basic/alloydb/v1beta1/alloydbcluster/basicalloydbcluster/create.yaml
nitinchugh-google: We can, but since this is a quick test (couple of minutes) it has not been seen to get into any contentions so far. the issues have been seen with long running tests

diviner524 (approved):

PR #988, opened 2023-11-02
docs: Start removing verbatim tags

justinsb:
The verbatim tags are not valid HTML/markdown, and we don't really need them.

This PR removes a first (simple) case


justinsb:
Opened #991 for the mockgcp test

/retest

PR #989, opened 2023-11-02
docs: Fix line breaks in inline html

justinsb:
Because we're generating HTML inline with the markdown, line breaks confuse at least the github markdown parser.


justinsb:
To see the problem, search for "External Reference Formats" on https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/7c0ef5f4238b837f08b4521d8f828ed90d7261f5/scripts/generate-google3-docs/resource-reference/generated/resource-docs/iam/iampolicy.md

diviner524 (commented):
/lgtm

diviner524 (commented):
Thank you for the fix @justinsb !

PR #999, opened 2023-11-07
Enable observed state in CRD generation via allowlist

maqiuyujoyce:
### Change description

Fixes b/309651922

### Tests you have done

### DIFF
--- a/pkg/k8s/constants.go
+++ b/pkg/k8s/constants.go
@@ -157,6 +157,14 @@ var (
 		// KCC no longer supports GameServicesRealm CRD as of v1.101.0.
 		"GameServicesRealm": true,
 	}
+
+	// ObservedStateAllowlist is a map containing a list of resources that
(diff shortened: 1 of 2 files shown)


pkg/k8s/constants.go
```diff
@@ -157,6 +157,14 @@ var (
 		// KCC no longer supports GameServicesRealm CRD as of v1.101.0.
 		"GameServicesRealm": true,
 	}
+
+	// ObservedStateAllowlist is a map containing a list of resources that
```
diviner524, line 161: Could you add a comment to explain this is only introduced for TF based resources?

maqiuyujoyce: Good point! I actually forgot about DCL-based resources, so I made the allowlist available for DCL-based resources as well.

maqiuyujoyce:
We decided to go with a different approach so this PR is no longer valid.

PR #1016, opened 2023-11-16
Adding support for google_tags_location_tag_binding

higef:
Change Description:
Using `Kind: TagsTagBinding` for resources such as storage buckets throws an error described in the issue https://github.com/hashicorp/terraform-provider-google/issues/11448

In response to that, a new resource was introduced [google_tags_location_tag_binding](https://registry.terraform.io/providers/hashicorp/google-beta/latest/docs/resources/google_tags_location_tag_binding) which allows to tag location-based resources.

Tests:
Due to different reasons, the only test I've done was on cloud environment:
- `make manifests` and applied generated CRD manifest
- `make docker-build` and pushed the manager artifact
- applied artifact to the controller's stateful set
- used the following resource manifest for tests
```
apiVersion: tags.cnrm.cloud.google.com/v1beta1
kind: TagsLocationTagBinding
metadata:
  annotations:
    cnrm.cloud.google.com/project-id: ******
  name: ******
  namespace: ******
spec:
  parentRef:
    external: //storage.googleapis.com/projects/_/buckets/<bucket name>
  tagValueRef:
    external: tagValues/<numerical value>
  location: eu
```

The expected tag was added with the resource, its status was properly set to "UpToDate", and the tag was removed after the resource was deleted.

Assistance with proper tests is highly appreciated.


config/servicemappings/tags.yaml
```diff
@@ -0,0 +1,51 @@
...
+            kind: Project
+            version: v1beta1
+            group: resourcemanager.cnrm.cloud.google.com
+          targetField: number
+          valueTemplate: "//cloudresourcemanager.googleapis.com/projects/{{value}}"
+          parent: true
+        - key: tagValueRef
+          tfField: tag_value
+          gvk:
+            kind: TagsTagValue
```
gemmahou, line 46: I do not think TagsTagValue is a resource KCC currently support, so we cannot reference to it. 
You may want to add TagsTagValue to service mapping as well, similar to EdgeContainerCluster: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/config/servicemappings/edgecontainer.yaml#L95

config/samples/resources/tagslocationtagbinding/tags_v1beta1_tagslocationtagbinding.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+apiVersion: tags.cnrm.cloud.google.com/v1beta1
+kind: TagsLocationTagBinding
+metadata:
+  annotations:
+    cnrm.cloud.google.com/project-id: ${PROJECT_ID?}
+  name: location-tag-binding-sample
+spec:
+  parentRef:
+    name: ${PROJECT_ID?}-tag-dep
```
gemmahou, line 23: I do not think I understand this. ${PROJECT_ID?}-tag-dep is a StorageBucket resource(defined in above yaml), why we use it as parent of resource TagsLocationTagBinding?

config/samples/resources/tagslocationtagbinding/storage_v1beta1_storagebucket.yaml
gemmahou, line 1: Not sure if why we need to define StorageBucket? it's not a dependency for the resources we are adding right?
I do not see StorageBucket is mentioned in the TF doc: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_tags_location_tag_binding

config/samples/resources/tagslocationtagbinding/tags_v1beta1_tagslocationtagbinding.yaml
higef: `parent` for the binding resource is the target, where the binding will attach the TagValue. Any [tag-compatible](https://cloud.google.com/resource-manager/docs/tags/tags-supported-services) resource can be defined as a parent. Instead of using `ref`, I can define the binding target through the `external` parameter, but I thought defining the whole chain through KCC resources would be more useful.

config/samples/resources/tagslocationtagbinding/storage_v1beta1_storagebucket.yaml
higef: In the TF example in `parent` definition there's `${google_cloud_run_service.default.name}` used as the binding target. For this particular example, a Storage Bucket is easier to declare. Can you give a hint on how can I declare sample resources more acceptably?

config/samples/resources/tagslocationtagbinding/tags_v1beta1_tagstagvalues.yaml
```diff
@@ -0,0 +1,24 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: tags.cnrm.cloud.google.com/v1beta1
+kind: TagsTagValue
+metadata:
+  annotations:
+    cnrm.cloud.google.com/project-id: ${PROJECT_ID?}
+  name: tag-key-sample
```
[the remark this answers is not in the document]
higef: fixed

config/servicemappings/tags.yaml
higef: [CRD](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/crds/tags_v1beta1_tagstagvalue.yaml) is provisioned with KCC and [service mapping](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/scripts/resource-autogen/generated/servicemappings/tags.yaml) is also there. I tested [TagsTagBinding](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/crds/tags_v1beta1_tagstagbinding.yaml) before working on LocationTagBinding for KCC and it was returning the expected error, so the assumption was TagValues and TagKeys are also supported.

Please confirm that they should be re-declared

gemmahou: Thanks for the pointer! Yes they are auto-generated resources that's why I did not aware of it.

config/samples/resources/tagslocationtagbinding/storage_v1beta1_storagebucket.yaml
gemmahou: I do not have much context about Storage Bucket, but by reading the TF doc, the binding target should be either  [compute_instance](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance) or  [cloud_run_service](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloud_run_service)?

gemmahou: By reading [tag-compatible](https://cloud.google.com/resource-manager/docs/tags/tags-supported-services), I understand there are more resources can be used as parent. But to keep consistency, I recommend use the resources in the examples in TF doc, i.e. use [compute_instance](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance) or [cloud_run_service](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloud_run_service) as parent(They are supported in KCC). Otherwise, people will get confused as I did. 

Also in our samples test, we want to cover all cases listed in the TF doc. That's say, you may want to add two cases here: one use compute_instance as parent, the other use cloud_run_service.

config/samples/resources/tagslocationtagbinding/tags_v1beta1_tagslocationtagbinding.yaml
gemmahou: Discussed in another comment, mark this as resolved.

gemmahou:
@higef I discussed this PR with my teammate, there are two concerns:
1. TagsTagKey and TagsTagValue are Alpha resource as you can see in [CRD](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/crds/tags_v1beta1_tagstagvalue.yaml) line23, currently we do not support reference to an Alpha resource. We will need to graduate Alpha resource to Beta, or we will need to make this field external only. Both ways take time to implement. Can you clarify the importance and urgency of this resource? Any business impact this resource has?
2. The parent field is actually a multi-kind reference field, it can reference to multiple resource kinds as listed in [tag-compatible](https://cloud.google.com/resource-manager/docs/tags/tags-supported-services), i.e. Storage Bucket, Cloud run Service, or Compute instance. We will need special handling on this multi-kind reference field.

higef:
@gemmahou 

1. There is a business impact. Recently, there was a new org. policy enforced on all the environments (including production and two production-like), and while working on a new environment (google project per environment), I noticed that we were no longer able to bind public IAM policies (e.g. `allUsers` for Cloud Run invoker) to some of our resources. Resource Tags are the only workaround for us at the moment. And there's no easy way for us to migrate back from KCC to TF. If it's easier and faster, `external` field will work just fine, but I understand that it's not an ideal approach for you.
2. I understand, thank you for the clarification. 
I'd appreciate it if you keep me posted or tag me wherever the development/discussion is happening. Thank you in advance

maqiuyujoyce:
Hi @higef are you open to add this resource as a v1alpha1 resource for now so that you can have a CRD to experiment with soon?

The disadvantage of `v1alpha1` is that we don't guarantee backwards compatibility. But that is why it is probably suitable for this resource, as the two friction points @gemmahou mentioned above may cause breaking changes if they are not solved properly.

The current way to add a v1alpha1 CRD is through the autogen channel. Here is an example about how to add the service mapping under autogen folder and enable it via the allowlist: https://github.com/GoogleCloudPlatform/k8s-config-connector/compare/master...maqiuyujoyce:202311-v1alpha1-crd.

higef:
Resource's been moved to `v1alpha1`. Should the samples under `config/samples/resources/` be updated to use respective API or removed completely until promoted to `v1beta1`? Same question regarding documentation.

config/servicemappings/tags.yaml
gemmahou, line 1: I think we can remove this file s we have the servicemapping file in generated/servicemappings

config/samples/resources/tagslocationtagbinding/storage_v1beta1_storagebucket.yaml
gemmahou, line 1: Docs and tests are not required for alpha resources. Feel free to remove them in this PR, and test it locally if needed.

scripts/resource-autogen/generated/servicemappings/tags.yaml
```diff
@@ -91,3 +91,30 @@ spec:
...
+      idTemplateCanBeUsedToMatchResourceName: false
+      resourceAvailableInAssetInventory: false
+      serverGeneratedIDField: name
+      resourceID:
+        targetField: name
+      resourceReferences:
+        - key: parentRef
+          tfField: parent
+          gvk:
+            kind: Project
```
gemmahou, line 107: Per the [list](https://cloud.google.com/resource-manager/docs/tags/tags-supported-services), Project is one of the supported resource, so it's fine to have it here. Just a side note that you are able to add multiple resource kinds like [example code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/config/servicemappings/compute.yaml#L1402-L1432). Considering there're numerous supported resources, I won't suggest add all of them. 

I would recommend always using "external" instead of "name" for parentRef field, in case you want to reference to other kinds of resources that not listed in the servicemapping. i.e. in your example, 
```
parentRef:
    **external**: //storage.googleapis.com/projects/_/buckets/<bucket name>
```

Another alternative would be simply remove the reference of parent field, and use plain text directly. i.e.
```
parent: "your_value"
```
You'll need to remove parentRef from servicemapping file of course. Since this is Alpha, I think we can add the reference back when needed. However, I haven't tested this and I'm unsure if this is going to work as expected, if you feel this is clearer and easier, feel free to try this way.

gemmahou:
> Resource's been moved to `v1alpha1`. Should the samples under `config/samples/resources/` be updated to use respective API or removed completely until promoted to `v1beta1`? Same question regarding documentation.

Docs and tests(samples, dynamic) are not required for an Alpha resource. Please see my comments and lmk if you have other questions.

scripts/resource-autogen/generated/servicemappings/tags.yaml
gemmahou: I'm fine to approve this PR at this point. Make sure that always use "external" for parentRef:https://cloud.google.com/config-connector/docs/how-to/creating-resource-references#external_references

gemmahou (approved):

gemmahou:
Got `ERROR: Generated code out-of-date. Please run 'make generate' and update your PR.`, please run `make generate`

gemmahou (approved):

PR #1020, opened 2023-11-20
Patch google_alloydb_user terraform resource to configure password as sensitive field

nitinchugh-google:
Patch google_alloydb_user terraform resource to configure password as sensitive field
Updating and renaming the existing alloydb-user-tf-5.4.0.patch to alloydb_user.patch

Fixes b/309020083

Tested locally by generating CRD for google_alloydb_user, and run unit tests with the updated CRD

- [NA] Run `make ready-pr` to ensure this PR is ready for review.
- [NA] Perform necessary E2E testing for changed resources.


diviner524:
Looks good, thank you @nitinchugh-google!

I think you need to resolve the conflict first, will LGTM after that.

PR #1025, opened 2023-11-22
fixes SecretKeyRef on Go client and CRD definition

### LINKED ISSUE #598: Go client and CRD definition of SQLUser are inconsistent
## Bug Description

SQLUser CRD definiton states: 

```yaml
password:
description: |-
  The password for the user. Can be updated. For Postgres instances this is a Required field, unless type is set to
                  either CLOUD_IAM_USER or CLOUD_IAM_SERVICE_ACCOUNT.
oneOf:
- not:
    required:
    - valueFrom
  required:
  - value
- not:
    required:
    - value
  required:
  - valueFrom
properties:
  value:
    description: Value of the field. Cannot be used if 'valueFrom'
      is specified.
    type: string
  valueFrom:
    description: Source for the field's value. Cannot be used if 'value'
      is specified.
    properties:
      secretKeyRef:
        description: Reference to a value with the given key in the
          given Secret in the resource's namespace.
        properties:
          key:
            description: Key that identifies the value to be extracted.
            type: string
          name:
            description: Name of the Secret to extract a value from.
            type: string
        required:
        - name
        - key
        type: object
```
Notice, that the `secretKeyRef ` actually is a object with `key` and `name`.
The Go struct generated for the SQLUser  boils down to

```go
type UserValueFrom struct {
	/* Reference to a value with the given key in the given Secret in the resource's namespace. */
	// +optional
	SecretKeyRef *v1alpha1.ResourceRef `json:"secretKeyRef,omitempty"`
}

```
and the `ResourceRef`

```go
type ResourceRef struct {
	/* The external name of the referenced resource */
	/* Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names */
	Name string `json:"name,omitempty"`
	/* Namespace of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/ */
}
```
So when I try to create a `SQLUser` resource and attempt to reference the password from `Secret` I get an error.

```bash
❯ kubectl version --short
Client Version: v1.21.2
Server Version: v1.22.4
```

### Config Connector Version

k8s-config-connector v1.71.0

### Config Connector Mode

cluster

## Log Output
```log
SQLUser.sql.cnrm.cloud.google.com \"alpha\" is invalid: spec.password.valueFrom.secretKeyRef.key: Required value", "errorVerbose": "SQLUser.sql.cnrm.cloud.google.com \"XXXX\" is invalid: spec.password.valueFrom.secretKeyRef.key: Required value
```

jijinp:
### Change description

Fixes #598 

### Tests you have done


pkg/clients/generated/apis/run/v1beta1/runservice_types.go
```diff
@@ -388,7 +388,7 @@ type ServiceTraffic struct {
 type ServiceValueSource struct {
 	/* Selects a secret and a specific version from Cloud Secret Manager. */
 	// +optional
-	SecretKeyRef *v1alpha1.ResourceRef `json:"secretKeyRef,omitempty"`
+	SecretKeyRef *v1alpha1.SecretKeyRef `json:"secretKeyRef,omitempty"`
```
diviner524, line 391: This is actually neither a ResourceRef, nor a SecretKeyRef.

It has two properties which are both of type ResourceRef.

diviner524: I think we can address this issue separately in future, as long as we don't assume all fields named secretKeyRef are of type "SecretKeyRef".

scripts/generate-go-crd-clients/generate-types-file.go
```diff
@@ -344,18 +348,22 @@ func isResourceReference(d fielddesc.FieldDescription) bool {
 	return false
 }
 
+func isSecretReference(d fielddesc.FieldDescription) bool {
+	return d.ShortName == "secretKeyRef"
```
diviner524, line 352: Could you add a bit more check here on `d.Children`? ex: If there are two children with name "name" and "key", then we know for sure it is a secretReference. This avoids us mistakenly treat cases like below as secretReference.

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/crds/run_v1beta1_runservice.yaml#L240

diviner524 (commented):
Looks good overall except a suggestion on `isSecretReference()`. Happy to LGTM after that.

Thank you @jijinp for the fix!

scripts/generate-go-crd-clients/generate-types-file.go
jijinp: @diviner524 Updated the function. I assume this addresses all scenarios.

PR #1029, opened 2023-11-22
mockgcp: Implement Network resource

justinsb:


justinsb:
@gemmahou  mind looking at this one?  It's been sitting for a while!

mockgcp/mockcompute/networksv1.go
```diff
@@ -0,0 +1,129 @@
...
+	fqn := name.String()
+
+	id := s.generateID()
+
+	obj := proto.Clone(req.GetNetworkResource()).(*pb.Network)
+	obj.SelfLink = PtrTo("https://compute.googleapis.com/compute/v1/" + name.String())
+	obj.CreationTimestamp = PtrTo(s.nowString())
+	obj.Id = &id
+	obj.SelfLinkWithId = PtrTo(fmt.Sprintf("https://compute.googleapis.com/compute/v1/projects/%s/global/networks/%d", name.Project.ID, id))
+	obj.Kind = PtrTo("compute#network")
```
gemmahou, line 69: Wondering why those values are needed? Per the [api doc](https://cloud.google.com/compute/docs/reference/rest/v1/networks/insert) only name is required

mockgcp/mockcompute/service.go
```diff
@@ -0,0 +1,94 @@
...
+
+func (s *MockService) Register(grpcServer *grpc.Server) {
+	pb.RegisterNetworksServer(grpcServer, s.networksv1)
+}
+
+func (s *MockService) NewHTTPMux(ctx context.Context, conn *grpc.ClientConn) (*runtime.ServeMux, error) {
+	mux := runtime.NewServeMux()
+
+	// TODO: Is all of beta a direct mapping to v1?
+	rewriteBetaToV1 := func(w http.ResponseWriter, r *http.Request, pathParams map[string]string) {
```
gemmahou, line 68: Will this beta to v1 mapping apply to other resources? Do you think it worth to add some doc to explain how do we determine if a resource has beta version and needs to rewrite to v1? As well as how to determine the correct mapping from beta to v1(similar to your TODO)
No need to be done in this PR.

mockgcp/mockcompute/networksv1.go
justinsb: We're acting as the service here, so we are populating these values.

If we do https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/957 then we can actually verify the mock against the real cloud APIs... I need to get that PR cleaned up!

mockgcp/mockcompute/service.go
justinsb: This mapping only applies to compute, it's a separate "mux".  We switch on the host header to pick which mux to use.

Good point on the docs/comments; I reworked the comment so it is hopefully better:

```
	// Terraform uses the /beta/ endpoints, but we have protos only for v1.
	// Also, we probably want to be implementing the newer versions
	// as that makes it easier to move KCC to newer API versions.
	// So far, it seems that all of beta is a direct mapping to v1 - though
	// I'm sure eventually we'll find something that needs special handling.
```

PR #1033, opened 2023-11-23
mockgcp: initial support for StorageBucket

justinsb:
A little unusual because we don't have the v1 service proto (though we
have the request/response definitions)


justinsb:
Oh actually this one is sort of tricky because we're also generating the proto from the OpenAPI (though this is all only for mocks at the end of the day, so low-consequence).  So I think it would be good if @cheftako could also take a look.

/assign @cheftako

mockgcp/mockstorage/v1.go
```diff
@@ -0,0 +1,427 @@
...
+	// for _, path := range paths {
+	// 	switch path {
+	// 	case "description":
+	// 		obj.Description = req.GetCertificate().GetDescription()
+	// 	case "labels":
+	// 		obj.Labels = req.GetCertificate().GetLabels()
+	// 	default:
+	// 		return nil, status.Errorf(codes.InvalidArgument, "update_mask path %q not valid", path)
+	// 	}
+	// }
```
gemmahou, line 104: nit: remove commented lines

justinsb: Good catch - done (and there were more later on also!)

justinsb:
Rebased to pick up test fixes

gemmahou:
This PR need a rebase, and refactor the error handling for CRUD functions.

justinsb:
Thanks @gemmahou  ... rebased and cleaned up!

PR #1038, opened 2023-11-28
Upgrade kdp to v0.15.2

maqiuyujoyce:
### Change description

Fixes b/309175495

* Made a local copy of K8s core `ResourceRequirements` type to avoid adding new fields into  ControllerResource and NamespacedControllerResource.
* Replaced `Watches()` function with `WatchesRawSource()` function in controllers. Updated the parameter for `Watches()` function.
* Added a non-nil cache to instantiate source.Kind (https://github.com/kubernetes-sigs/controller-runtime/pull/2120).  
* Updated the test kubeclient in namecheck_test.go and upgrade_test.go.
* Updated `singleResourceClient` and `errorClient` to match the latest interface of the controller-runtime client.
* Removed the configuration of `GoHeaderFilePath` from deepcopy-gen and client-gen. The line of code turns out to be unnecessary.
* Updated the client builders for the test clients.
* Updated the registeration of the webhook server. 
* The following generated files are also changed:
  * The clientset/doc.go is removed (https://github.com/kubernetes/code-generator/commit/8b5f5878f4a9c9cc0c385732aabe0b2e78872807).
  * The instantiation of the resource and kind variables in the generated files under clientset is changed.

### Tests you have done

* `make test` passed.
* Deployed the controllers locally and successfully created a PubSubTopic.


justinsb:
LGTM except for the resources.claims change.  I'm *guessing* this was introduced upstream in k8s and we're reusing the k8s type here.  I'm also guessing we would have problems if someone actually set this new value (?).  Maybe we want to define our own version of resources so we don't pick up the new "claims" field?

maqiuyujoyce:
Hi @justinsb , thank you for reviewing!

> I'm guessing this was introduced upstream in k8s and we're reusing the k8s type here.

Yes, `claims` field was introduced because the upstream type in K8s added it: https://github.com/kubernetes/api/compare/v0.25.0...v0.27.2#diff-04e78d7bfddcb8b0b0fb2b950dbefcceb0af25bd827bfb41721f1280810a160aR2345.

> I'm also guessing we would have problems if someone actually set this new value (?). Maybe we want to define our own version of resources so we don't pick up the new "claims" field?

I'm not familiar with this feature enough to determine if it'll be problematic or not off the top of my head, so agreed we should play it safe. I made a local copy of this type without the `claims` field.

justinsb:
Thanks @maqiuyujoyce - lgtm with the embedding of the resource type!

/approve
/lgtm

Looks like it'll need a rebase on go.mod though

PR #1039, opened 2023-11-29
tests: refactor sample loading to defer substitution

justinsb:
This lets us load the samples before creating a test project, which in
turn lets us run samples in isolated projects, particularly in mocks.

### DIFF
--- a/config/tests/samples/create/contents_test.go
+++ b/config/tests/samples/create/contents_test.go
@@ -1,6 +1,3 @@
-//go:build integration
-// +build integration
-
 package create
 
 import (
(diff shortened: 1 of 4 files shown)


config/tests/samples/create/contents_test.go
```diff
@@ -15,11 +12,12 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
-var matchEverythingRegex = regexp.MustCompile(".*")
```
maqiuyujoyce, line 18: Maybe it's still reasonable to keep a global variable representing this regex? But maybe in samples.go instead?

maqiuyujoyce (approved):
Thank you for the refactoring! It is more readable now!

config/tests/samples/create/contents_test.go
justinsb: My 2c is that we should add it when/if we need it.  I don't think it will be very hard to recreate if we do need it!

maqiuyujoyce:
Looks like there was a force push; was there any change?

justinsb:
> Looks like there was a force push; was there any change?
No, just a rebase

config/tests/samples/create/contents_test.go
```diff
@@ -15,11 +12,12 @@ import (
...
-
 func TestNames(t *testing.T) {
-	project := testgcp.GetDefaultProject(t)
-	samples := loadSamplesOntoUnstructs(t, matchEverythingRegex, project)
+	project := testgcp.GCPProject{
+		ProjectID:     "tests-testnames",
```
cheftako, line 17: Not a big deal but it might be nice if the fake project name contained "project"?
Makes it easier to find/understand where it came from if it shows up somewhere unexpected.

PR #1042, opened 2023-11-29
tests: use conventional MustReadFile function

justinsb:
We had a common test helper function that is normally called
MustReadFile, and had some potentially surprising behaviours when the
absolute path could not be determined.

Replace with the more common definition.


diviner524 (commented):
/lgtm

PR #1049, opened 2023-12-01
Update README.ChangingTerraform.md to be based on TF Git Subtree

diviner524:
This is required as we have now removed all the previous Terraform "patches" and abandoned the patch approach.

Fixes b/314326104


justinsb:
This is cool, a big advantage of the subtree approach that I don't think I had grokked is that we can now just fix terraform-provider issues following (roughly) the normal development flow (albeit with separate commits)

/approve
/lgtm

PR #1052, opened 2023-12-04
Update DCL to version 1.58.0

diviner524:
### Change description

* Resource GKEHubFeatureMembership(v1beta1):
  * Added `spec.configmanagement.configSync.metricsGcpServiceAccountRef` field.
  * Added `spec.membershipLocation` field.
  * Deprecated `spec.configmanagement.binauthz` field.

* Resource RecaptchaEnterpriseKey(v1beta1):
  * Added `spec.wafSettings` field.

* Adding instructions on how to perform DCL upgrade, as we still need to upgrade our DCL library periodically to pick up bug fixes and important new fields.

Fixes b/314794142

### Tests you have done


README.UpdatingDCL.md
```diff
@@ -0,0 +1,11 @@
+# Updating the DCL (declarative-resource-client-library)
```
maqiuyujoyce, line 1: In the Makefile, we call it "upgrade-dcl", and in the readme, we call it "updating the DCL". Should it be upgrade or update? IMO "upgrade" is specific for version updates, and "update" then can be used for non-version updates.

Not a blocker, though I think using the consistent wording can help reduce confusion.

maqiuyujoyce (approved):

README.UpdatingDCL.md
diviner524: That makes sense, thank you @maqiuyujoyce!

I have thought about using "Upgrading" as well, but then I switched to "Updating" since we were using the same term on the Terraform Provider:

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/README.UpdatingTerraformProvider.md

We can go back and change this if it turns out to be confusing.

PR #1053, opened 2023-12-05
Resolve the observed state for TF-based resources

maqiuyujoyce:
### Change description

Fixes b/309651934

* Supported the logic to resolve the observed state for TF-based resources during a successful reconciliation.
  * Reference fields, labels field, name field, sensitive fields, and fields under an array are currently not supported in the observed state.
  * Added unit tests.
* Added more unit tests for service mappings to ensure the same limitations are enforced.

### Tests you have done

Added the following configurations in the resource config for StorageBucket and generated the CRD.

```
observedFields:
  - versioning.enabled
  - location
```

Ran the integration test with the following YAML:

```
apiVersion: storage.cnrm.cloud.google.com/v1beta1
kind: StorageBucket
metadata:
  name: storagebucket-sample-${uniqueId}
spec:
  versioning:
    enabled: true
  location: US-WEST1
  lifecycleRule:
    - action:
        type: Delete
      condition:
        age: 7
```

The created StorageBucket contains the observed state in `status`:

```
         "observedState": {
            "location": "US-WEST1",
            "versioning": {
                "enabled": true
            }
        },
```

### DIFF
--- a/config/tests/servicemapping/servicemapping_test.go
+++ b/config/tests/servicemapping/servicemapping_test.go
@@ -1278,11 +1278,13 @@ func TestStorageVersionIsSetAndValidIFFV1alpha1ToV1beta1IsSet(t *testing.T) {
 func TestObservedFields(t *testing.T) {
 	t.Parallel()
 	serviceMappings := testservicemappingloader.New(t).GetServiceMappings()
+	provider := tfprovider.NewOrLogFatal(tfprovider.UnitTestConfig())
 	for _, sm := range serviceMappings {
 		sm := sm
 		t.Run(sm.Name, func(t *testing.T) {
 			t.Parallel()
 			for _, rc := range sm.Spec.Resources {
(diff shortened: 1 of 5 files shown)


diviner524 (commented):
Mock test looks to be flaky recently. There is a timeout error:

    harness.go:153: error stopping envtest environment: timeout waiting for process kube-apiserver to stop


I think you can retry it.

maqiuyujoyce:
Found a related topic about the error message: https://github.com/kubernetes-sigs/controller-runtime/issues/1571 though we may have already done the workaround for it: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/1e21f6a20f1fbd3c01f2b9791091d3fb67f34f20/config/tests/samples/create/harness.go#L104 I wondered if context cancel() is indeed called before env.Stop(), or if there are any other issues.

PR #1059, opened 2023-12-07
Support the partial observed state in ContainerCluster, ContainerNodePool and RedisInstance

maqiuyujoyce:
### Change description

Fixes b/315227901

* Supported the following observed fields:

  * `status.observedState.masterAuth.clientCertificate` in ContainerCluster
  * `status.observedState.version` in ContainerNodePool
  * `status.authString` in RedisInstance

* Fixed the issue that when the parent of the observed field exists but the observed field itself doesn't, there is still an empty observed state. Added test cases to cover this scenario.

### Tests you have done

Did local integration test using the following test cases.

* ContainerCluster and ContainerNodePool

  1. Applied the following YAML:
      
      ```
      apiVersion: container.cnrm.cloud.google.com/v1beta1
      kind: ContainerCluster
      metadata:
        name: test-observed-state-1
        namespace: test
      spec:
        location: us-central1-a
        initialNodeCount: 1
        loggingService: logging.googleapis.com/kubernetes
        monitoringService: monitoring.googleapis.com/kubernetes
      ---
      apiVersion: container.cnrm.cloud.google.com/v1beta1
      kind: ContainerNodePool
      metadata:
        name: test-observed-state-1
        namespace: test
      spec:
        location: us-central1-a
        autoscaling:
          minNodeCount: 1
          maxNodeCount: 3
        clusterRef:
          name: test-observed-state-1
      ---
      apiVersion: container.cnrm.cloud.google.com/v1beta1
      kind: ContainerCluster
      metadata:
        name: test-observed-state-2
        namespace: test
      spec:
        location: us-central1-a
        initialNodeCount: 1
        loggingService: logging.googleapis.com/kubernetes
        monitoringService: monitoring.googleapis.com/kubernetes
        masterAuth:
          clientCertificateConfig:
            issueClientCertificate: true
      ```

  2. Verified that after the reconciliation completed, in the output of `kubectl describe`, the values under `status.observedState` in the CRs matched the corresponding fields under `spec`.

     ```
      # No `status.observedState` in ContainerCluster test-observed-state-1.

      # Had `status.observedState` that matched the spec in ContainerNodePool test-observed-state-1.
      Status:
        ...
        Observed State:
          Version:  1.27.3-gke.100
      
      # Had `status.observedState` that matched the spec in ContainerCluster test-observed-state-2.
      Status:
        ...
        Observed State:
          Master Auth:
            Client Certificate: [truncated]
     ```

* RedisInstance

   1. Applied the following YAML:

      ```
      apiVersion: redis.cnrm.cloud.google.com/v1beta1
      kind: RedisInstance
      metadata:
        name: test-observed-state-1
        namespace: test
      spec:
        authEnabled: true
        displayName: Sample Redis Instance
        region: us-central1
        tier: BASIC
        memorySizeGb: 16
        maintenanceSchedule:
        - endTime: 2023-08-01T16:29:22.045123456Z
          scheduleDeadlineTime: 2023-08-01T17:29:22.045123456Z
          startTime: 2023-08-01T15:29:22.045123456Z
      ```

   2. Verified that after the reconciliation completed, in the output of `kubectl describe`, `status.observedState.authString` can be found.

   3. Updated the resource with the following YAML:

      ```
      apiVersion: redis.cnrm.cloud.google.com/v1beta1
      kind: RedisInstance
      metadata:
        name: test-observed-state-1
        namespace: test
      spec:
        authEnabled: false
        displayName: Sample Redis Instance
        region: us-central1
        tier: BASIC
        memorySizeGb: 16
        maintenanceSchedule:
        - endTime: 2023-08-01T16:29:22.045123456Z
          scheduleDeadlineTime: 2023-08-01T17:29:22.045123456Z
          startTime: 2023-08-01T15:29:22.045123456Z
      ```

   4. Verified that after the reconciliation completed, in the output of `kubectl describe`, the entire `status.observedState` struct is gone.


pkg/krmtotf/tftokrm_test.go
```diff
@@ -2138,7 +2201,58 @@ func TestResolveSpecAndStatusContainingObservedState(t *testing.T) {
 			},
 		},
 		{
-			name: "with partial string observed fields in observed state",
+			name: "nested observed field not exist in the returned state but " +
```
diviner524, line 2204: I think I understand the test case, but can the name be more descriptive? The parent does exist, not because it must exist, but because the parent has another child field, which has  an actual value being returned by the underlying API.

scripts/generate-google3-docs/resource-reference/generated/resource-docs/redis/redisinstance.md
```diff
@@ -896,6 +898,20 @@ resolution and up to nine fractional digits.{% endverbatim %}</p>
             <p>{% verbatim %}ObservedGeneration is the generation of the resource that was most recently observed by the Config Connector controller. If this is equal to metadata.generation, then that means that the current reported status reflects the most recent desired state of the resource.{% endverbatim %}</p>
         </td>
     </tr>
+    <tr>
+        <td><code>observedState</code></td>
+        <td>
+            <p><code class="apitype">object</code></p>
+            <p>{% verbatim %}The observed state of the underlying GCP resource.{% endverbatim %}</p>
```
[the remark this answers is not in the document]
maqiuyujoyce: Good point. Created b/316017785.

pkg/krmtotf/tftokrm_test.go
maqiuyujoyce: Sorry for the confusion! What about the new name?

pkg/test/resourcefixture/testdata/basic/container/v1beta1/maqiuyu/create.yaml
```diff
@@ -0,0 +1,26 @@
+# Copyright 2022 Google LLC
```
diviner524, line 1: It seems you have added four test YAMLs including this one by mistake?

maqiuyujoyce: Yep, that was by mistake. Removed now!

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_containerclusters.container.cnrm.cloud.google.com.yaml
```diff
@@ -1808,6 +1808,24 @@ spec:
                   current reported status reflects the most recent desired state of
                   the resource.
                 type: integer
+              observedState:
```
justinsb, line 1811: So given that e.g. masterVersion is status.masterVersion, do we want to introduce status.observedState?  i.e. should it be status.observedState.masterAuth.clientCertificate vs status.masterAuth.clientCertificate?

The way I see it:
1. If we introduce status.observedState we have to deal with the status.masterVersion vs status.observedState.masterVersion inconsistency at some time in the future
2. If we don't introduce status.observedState we (maybe) have a naming conflict if there's a status field in the API called "conditions"

Personally, I'd rather deal with the hypothetical work of (2) than the known work of (1), but that's just my intuition

maqiuyujoyce: Thank you for the suggestions, @justinsb !

> 1. If we introduce status.observedState we have to deal with the `status.masterVersion` vs `status.observedState.masterVersion` inconsistency at some time in the future
>
> 2. If we don't introduce `status.observedState` we (maybe) have a naming conflict if there's a status field in the API called "conditions"

IMO, if we introduce `status.observedState.masterVersion` one day, then `status.masterVersion` and `status.observedState.masterVersion` will always share the same value. I'd consider it a redundancy but not inconsistency. Or are you referring to some other scenarios?

If we need to deal with naming conflict, then it seems to be a true inconsistency: possibly we either need to rename the observed field (inconsistent with its original field name), or rename the existing status field (may impact all CRDs).

And in general I find redundancy safer to the users.

cheftako: My 2 cents. I think grouping the observed state has additional benefits in that we're grouping it. We can tell customers (and future devs) that all cloud response state can be found under .spec.observedState. More importantly I think we should insist that only cloud response state is allowed into .spec.observedState. That allows anyone looking at a value under .spec.observedState to trust its a value from the cloud provider and not say an artifact from the controller. I think it will make a docs easier and clearer. I think the fact that it essentially name spaces the response data and hence provides protection from future conflicts (either from new fields, newly cared about response fields or other non cloud response status fields which happen to have a name collision) is valuable but possibly not the most valuable.

diviner524: I agree with @cheftako, my view on a dedicated `status.observedState` is it gives our users a new mental model on how to identify status changes caused by the external cloud provider, but not KCC.

> if there's a status field in the API called "conditions"

Also I want to mention this is not a hypothetical scenario. We have already seen it in a few CRDs. See the following two examples in our servicemapping config. We had to ignore a bunch of fields because they have conflict naming with the reserved K8s status name fields.

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/config/servicemappings/run.yaml#L27

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/config/servicemappings/run.yaml#L147

Should customers require support for these ignored fields in the future, we have the flexibility to add them to `status.observedState`.

maqiuyujoyce: Thank you for the input, @cheftako and @diviner524 !

@justinsb I'll probably merge the implementation as is for now. If you think not having `status.observedState` is important, then let's discuss offline about a revised design.

PR #1065, opened 2023-12-11
mockgcp: Initial support for container.google.apis (GKE)

justinsb:
- mockgcp: generate bindings for containers.googleapis.com v1
- mockgcp: Initial mock for container.googleapis.com


mockgcp/mockcontainer/lro.go
```diff
@@ -0,0 +1,49 @@
...
+	pb "github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/generated/mockgcp/container/v1beta1"
+	lropb "google.golang.org/genproto/googleapis/longrunning"
+)
+
+func (s *ClusterManagerV1) newLRO(ctx context.Context, projectID string) (*pb.Operation, error) {
+	lro, err := s.operations.NewLRO(ctx)
+	if err != nil {
+		return nil, err
+	}
+	selfLinkPrefix := "https://container.googleapis.com/v1beta1/projects/" + projectID + "/global/operations/"
```
cheftako, line 30: Should we be using the v1beta1 API here? I thought GKE was pushing for everyone to use the V1 container.googleapi?

mockgcp/mockcontainer/names.go
```diff
@@ -0,0 +1,65 @@
...
+
+// parseClusterName parses a string into a clusterName.
+// The expected form is `projects/*/locations/*/clusters/*`.
+func (s *MockService) parseClusterName(name string) (*clusterName, error) {
+	tokens := strings.Split(name, "/")
+
+	if len(tokens) == 6 && tokens[0] == "projects" && tokens[2] == "locations" && tokens[4] == "clusters" {
+		project, err := s.projects.GetProjectByID(tokens[1])
+		if err != nil {
+			return nil, err
```
cheftako, line 52: Do we want to wrap this error in something which says could not get project from ID $tokens[1]?

mockgcp/mockcontainer/nodepool.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+}
+
+// parseNodePoolName parses a string into a nodePoolName.
+// The expected form is `projects/*/locations/*/clusters/*`.
+func (s *MockService) parseNodePoolName(name string) (*nodePoolName, error) {
+	tokens := strings.Split(name, "/")
+
+	if len(tokens) == 8 && tokens[0] == "projects" && tokens[2] == "locations" && tokens[4] == "clusters" && tokens[6] == "nodePools" {
+		project, err := s.projects.GetProjectByID(tokens[1])
```
cheftako, line 57: Would it make sense to extract a utility for getting the project from a string with "projects/ID" ?

mockgcp/mockcontainer/nodepool.go
```diff
@@ -0,0 +1,209 @@
...
+
+	fqn := name.String()
+
+	oldObj := &pb.NodePool{}
+	if err := s.storage.Delete(ctx, fqn, oldObj); err != nil {
+		klog.Infof("deleting nodePool %q: %v", fqn, err)
+		if apierrors.IsNotFound(err) {
+			return nil, status.Errorf(codes.NotFound, "nodePool %q not found", name)
+		} else {
+			return nil, status.Errorf(codes.Internal, "error deleting nodePool: %v", err)
```
cheftako, line 204: Was there a reason to not use %w?

cheftako:
Very nice :D Will happily lgtm.

mockgcp/mockcontainer/lro.go
justinsb: Our terraform provider is using v1beta1, so this is what we have to mock today.

We could also support v1 (and we'll likely have to eventually), but I figured we should start with the codepath we're using today.

mockgcp/mockcontainer/names.go
justinsb: So my rule of thumb is that it's a call to our own code, I normally prefer to have it return a reasonable error message.  It knows the projectID, so I don't have any additional context to add to the error message.

In this case, it actually breaks terraform if we don't return the exact error message terraform expects, it does naive string matching (!!!), so we really don't want to wrap it here :-)

mockgcp/mockcontainer/nodepool.go
justinsb: GetProjectByID is already returning a project object.  This code is pretty boilerplate, and I have been thinking it would be nice to be able to pass in a "template" like `projects/:project/locations/:location/clusters/:cluster/nodePools/:nodePool` and have it automatically parsed.  I haven't done that yet though, in practice Duet just does most of the work for these methods for us automatically

mockgcp/mockcontainer/nodepool.go
justinsb: IIRC status.Errorf doesn't allow it.  I'll verify.

justinsb: Confirmed, it's just a wrapper on fmt.Sprintf: https://github.com/grpc/grpc-go/blob/v1.60.0/status/status.go#L64

Maybe I should be using `status.FromError`, but I think the problem is that we need to return a particular status code at the http layer, which I don't think `status.FromError` would do

justinsb:
> Very nice :D Will happily lgtm.
OK, thank you - removed WIP then!

I do think we want to get more of the test infrastructure in place (e.g. golden request capturing), but the sequencing doesn't _really_ matter.

mockgcp/mockcontainer/nodepool.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+	if obj.Config.DiskType == "" {
+		obj.Config.DiskType = "pd-balanced"
+	}
+	if obj.Config.ImageType == "" {
+		obj.Config.ImageType = "COS_CONTAINERD"
+	}
+
+	if obj.Config.MachineType == "" {
+		obj.Config.MachineType = "e2-standard-4"
```
cheftako, line 93: How do we keep these up to date?

PR #1069, opened 2023-12-14
Disable containerattachedcluster integration test

gemmahou:
### Change description

Fixes b/315013168

### Tests you have done


maqiuyujoyce:
/lgtm

Though currently if we want to run the tests defined in the lists, they'll just be silently skipped and may cause confusions. Can we make it more verbose by adding some log [here](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/test/resourcefixture/test_runner.go#L47) like [this](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/config/tests/samples/create/harness.go#L365) (no need to address it now and feel free to just create a ticket to track)?

PR #1079, opened 2023-12-26
Add the GitHub action to upgrade the operator when a new version tag is created

maqiuyujoyce:
### Change description

Fixes b/315219970

### Tests you have done

Tested the GitHub action against my fork repo with the following commands (to create a new test tag):

```
git tag -a "v1.200.9" [LAST_COMMIT_SHA_IN_HISTORY] -m "test"
git push origin v1.200.9
```

The action was triggered here: https://github.com/maqiuyujoyce/k8s-config-connector/actions/runs/7332214200/job/19966023037

And it created a PR containing all the changes in my branch: https://github.com/maqiuyujoyce/k8s-config-connector/pull/13

(This PR contains the changes in one commit so it doesn't have the same commit history as the PR.)

We should verify it works fine against the product repo in the next release.

### DIFF
--- a/.github/workflows/upgrade-operator.yaml
+++ b/.github/workflows/upgrade-operator.yaml
@@ -53,14 +53,16 @@ jobs:
           echo "Running 'make upgrade-kcc'..."
           make -C operator upgrade-kcc
           echo "Ensuring there is diff and committing the change..."
-          git diff
+          if [[ -z `git status --porcelain` ]]; then
+            echo "Exiting the workflow as there is no change to generate a PR."
+            exit 1
+          fi
           git add .
           git commit -m "Upgrade KCC operator to version ${{ env.RELEASE_VERSION }}"
       - name: "Create PR"
         run: |
           echo "Adding a remote pointing to GoogleCloudPlatform/k8s-config-connector..."
           git remote add target https://github.com/GoogleCloudPlatform/k8s-config-connector.git
-          git remote -vv
           export BRANCH=operator-upgrade-$(date +"%Y%m%d%H%M%S")
           echo "Setting up branch '${BRANCH}'..."
           git checkout -b ${BRANCH}
--- a/.github/workflows/upgrade-operator.yaml
+++ b/.github/workflows/upgrade-operator.yaml
@@ -1,0 +1,72 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Upgrade Operator
+
+on:
+  push:
+    # The action is only triggered when a new tag following the current
+    # convention is created.
+    tags: [ "v1.[0-9]+.[0-9]+" ]
+
+jobs:
+  upgrade:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          # This is to get all the commits.
+          fetch-depth: 0
+      - uses: actions/setup-go@v4
+        with:
+          go-version: "1.21.5"
+      - name: "Set env"
+        run: |
+          echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+      # ${{ env.RELEASE_VERSION }} is not accessible in the same step when it is
+      # set.
+      - name: "Check current state"
+        run: |
+          git config user.name "GitHub Actions Bot"
+          git config user.email "<>"
+          echo "Checking the current state..."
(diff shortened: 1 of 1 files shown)


.github/workflows/upgrade-operator.yaml
```diff
@@ -0,0 +1,72 @@
...
+          make -C operator upgrade-kcc
+          echo "Ensuring there is diff and committing the change..."
+          git diff
+          git add .
+          git commit -m "Upgrade KCC operator to version ${{ env.RELEASE_VERSION }}"
+      - name: "Create PR"
+        run: |
+          echo "Adding a remote pointing to GoogleCloudPlatform/k8s-config-connector..."
+          git remote add target https://github.com/GoogleCloudPlatform/k8s-config-connector.git
+          git remote -vv
```
diviner524, line 63: This is for debugging and can be removed?

.github/workflows/upgrade-operator.yaml
```diff
@@ -0,0 +1,72 @@
...
+          export BRANCH=operator-upgrade-$(date +"%Y%m%d%H%M%S")
+          echo "Setting up branch '${BRANCH}'..."
+          git checkout -b ${BRANCH}
+          echo "Creating the PR against master branch of GoogleCloudPlatform/k8s-config-connector..."
+          git push -u target ${BRANCH}
+          gh repo set-default GoogleCloudPlatform/k8s-config-connector
+          gh pr create --head ${BRANCH} --title 'Upgrade KCC operator to version ${{ env.RELEASE_VERSION }}' --body 'Upgrade KCC operator to version ${{ env.RELEASE_VERSION }}.'
```
diviner524, line 70: If this auto-generated PR is not getting merged timely, I think there will be a chance of conflicting changes merged, and we can do nothing to resolve the conflict.

As an example if someone happens to have updated `operator/config/default/kustomization.yaml` after we create the release tag and before we merge this PR, the PR will then become not mergeable. Also there will be no way for us to re-generate this PR, as IIUC, it will always be generated based on v1.X.X tag.

This was not a problem previously because we manually run make upgrade-kcc against the HEAD of master branch.

Having said that, I think the chance of the above happening is low so I think we don't need to worry about it now.

.github/workflows/upgrade-operator.yaml
```diff
@@ -0,0 +1,72 @@
...
+      # ${{ env.RELEASE_VERSION }} is not accessible in the same step when it is
+      # set.
+      - name: "Check current state"
+        run: |
+          git config user.name "GitHub Actions Bot"
+          git config user.email "<>"
+          echo "Checking the current state..."
+          echo "Release version: ${{ env.RELEASE_VERSION }}"
+          echo "Location: $(pwd)"
+          echo "Commit history:"
```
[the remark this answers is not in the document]
maqiuyujoyce: I consider them part of the logs, not necessarily only for debugging purpose. For example, we print out the version, and other necessary information when we run the weekly release scripts. I think this is similar. Do you have any concerns?

.github/workflows/upgrade-operator.yaml
```diff
@@ -0,0 +1,72 @@
...
+          git log --oneline -10
+          echo "Current branch:"
+          git branch
+      - name: "Upgrade KCC operator"
+        run: |
+          echo "Running 'make upgrade-kcc'..."
+          make -C operator upgrade-kcc
+          echo "Ensuring there is diff and committing the change..."
```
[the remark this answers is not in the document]
maqiuyujoyce: Good point. Added a if condition to exit the workflow when there is no diff.

.github/workflows/upgrade-operator.yaml
maqiuyujoyce: Done.

.github/workflows/upgrade-operator.yaml
maqiuyujoyce: > As an example if someone happens to have updated operator/config/default/kustomization.yaml after we create the release tag and before we merge this PR, the PR will then become not mergeable. Also there will be no way for us to re-generate this PR, as IIUC, it will always be generated based on v1.X.X tag.

Agreed that this is solid concern. Let's keep track of the occurrences of such edge cases and see if we should introduce another trigger for this GitHub action.

.github/workflows/upgrade-operator.yaml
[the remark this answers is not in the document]
maqiuyujoyce: Updated the log a bit.

.github/workflows/upgrade-operator.yaml
```diff
@@ -0,0 +1,74 @@
...
+          if [[ -z `git status --porcelain` ]]; then
+            echo "Exiting the workflow as there is no change to generate a PR."
+            exit 1
+          fi
+          git add .
```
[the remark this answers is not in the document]
maqiuyujoyce: Good point! Done.

.github/workflows/upgrade-operator.yaml
```diff
@@ -57,7 +57,7 @@ jobs:
             echo "Exiting the workflow as there is no change to generate a PR."
             exit 1
           fi
-          git add .
+          git add -A .
```
diviner524, line 60: nit: `git add -A` is the same as `git add -A .`, so the `.` is redundant.

maqiuyujoyce: Thanks for pointing it out! I also checked the latest manual and experimented with the command locally, and found that `git add .` should already add all the modified, added and deleted files: https://git-scm.com/docs/git-add#Documentation/git-add.txt-ltpathspecgt82308203

PR #1040, opened 2023-11-29
Remove last uses of go-datastructures library

justinsb:
We weren't using it significantly.

PR #1093, opened 2024-01-04
Remove last uses of go-datastructures library

justinsb:
We weren't using it signficantly.


justinsb:
Actually we already have #1040

Issue #991, opened 2023-11-02
MockGCP: Flake related to resource deletion ordering


justinsb:
### Checklist

- [X] I did not find a related open issue.
- [X] I did not find a solution in the troubleshooting guide: (https://cloud.google.com/config-connector/docs/troubleshooting)
- [X] If this issue is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Bug Description

We saw a flake in the mockgcp tests, it looks like the certificatemanagercertificatemapentry was deleted out-of-order and this was not tolerated:

```
2023-11-02T12:23:57.6002721Z {"severity":"error","timestamp":"2023-11-02T12:23:57.599Z","msg":"Reconciler error","controller":"certificatemanagercertificatemapentry-controller","controllerGroup":"certificatemanager.cnrm.cloud.google.com","controllerKind":"CertificateManagerCertificateMapEntry","CertificateManagerCertificateMapEntry":{"name":"certificatemanagercertificatemapentry-sample","namespace":"mock-project-1698926825"},"namespace":"mock-project-1698926825","name":"certificatemanagercertificatemapentry-sample","reconcileID":"8238424e-a254-434e-9699-905bce0aee2b","error":"Delete call failed: error fetching live state: error getting ID for resource: error getting value from reference: could not parse reference resolution value '<nil>' as string"}
```

### Additional Diagnostic Information

-

### Kubernetes Cluster Version

kubetest

### Config Connector Version

main

### Config Connector Mode

namespaced mode (default)

### Steps to reproduce the issue

It was seen here: https://github.com/GoogleCloudPlatform/k8s-config-connector/actions/runs/6732286379/job/18298694061?pr=988

gemmahou:
I checked our "real" samples test and did not see flakiness in the certificatemanagercertificatemapentry: [test grid](https://testgrid.corp.google.com/release%20branch%20-%20last%20known%20good#periodic-cnrm-samples-test). This test is green all the time. I suspect the failure is related to how the mock is configured. I created a buganizer ticket to track this issue. b/321582240

gemmahou:
Seems like this has been fixed, I don't see the test failures any more.

PR #1098, opened 2024-01-05
tests: run each mock test in a different project

justinsb:
This also allows for context isolation and (in theory) parallel execution.

### DIFF
--- a/config/tests/samples/create/harness.go
+++ b/config/tests/samples/create/harness.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	exportparameters "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/cli/cmd/export/parameters"
+	"google.golang.org/api/cloudresourcemanager/v1"
 	cloudresourcemanagerv1 "google.golang.org/api/cloudresourcemanager/v1"
 	"google.golang.org/api/option"
 
@@ -37,6 +38,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/logging"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test"
 	testenvironment "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test/environment"
+	testgcp "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test/gcp"
 	testwebhook "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test/webhook"
 	cnrmwebhook "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/webhook"
 	"golang.org/x/oauth2"
@@ -62,6 +64,8 @@ import (
 type Harness struct {
 	*testing.T
 	Ctx context.Context
+
+	Project testgcp.GCPProject
 
 	client     client.Client
 	restConfig *rest.Config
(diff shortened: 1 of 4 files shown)


justinsb:
Rebased; I think this one is next up on the road to recording requests (#957)

cc @gemmahou

gemmahou:
Noticed that the presubmit/mock-test is failing. I retried but it still failed due to:
```
2024/01/22 21:29:08 request: GET https://secretmanager.googleapis.com/v1/projects/mock-project/secrets/secretmanagersecretversion-dep?alt=json
2024/01/22 21:29:08 [DEBUG] Retry Transport: Stopping retries, last request failed with non-retryable error: googleapi: got HTTP response code 404 with body: HTTP/0.0 404 Not Found

{"code":5, "message":"secret \"projects/mock-project/secrets/secretmanagersecretversion-dep\" not found", "details":[]}
```

gemmahou:
I do not think I have comments on this PR, as long as the mock test passed. Just curious if this change would lead to a longer execution time on mocks?

justinsb:
> Just curious if this change would lead to a longer execution time on mocks?

Yes.  But engineer time is much more expensive than computer time, and this is on the road to better test coverage.  If we're waiting on the mocks, we can split them into multiple github action jobs or pay for faster github action runners.

The reason it's slower though is primarily because we're spinning up a new kube-apiserver for each harness (with E2E_KUBE_TARGET=envtest).  https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1133 uses a faster mock kube-apiserver which is _much_ faster (but doesn't support webhooks, which brings its own complications)

justinsb:
On the test failure / flake, we wouldn't expect to see test flakes on the mocks.  I retriggered the job so we'll see.

I looked in the history though, and I think the error is:

```
2024-01-20T15:32:03.1449687Z {"severity":"error","timestamp":"2024-01-20T15:32:03.144Z","msg":"Reconciler error","controller":"certificatemanagercertificatemapentry-controller","controllerGroup":"certificatemanager.cnrm.cloud.google.com","controllerKind":"CertificateManagerCertificateMapEntry","CertificateManagerCertificateMapEntry":{"name":"certificatemanagercertificatemapentry-sample","namespace":"mock-project"},"namespace":"mock-project","name":"certificatemanagercertificatemapentry-sample","reconcileID":"2e1ea373-f34d-4b58-9d24-083fe1756293","error":"Delete call failed: error fetching live state: error getting ID for resource: error getting value from reference: could not parse reference resolution value '<nil>' as string"}
2024-01-20T15:34:00.8544958Z     samples.go:229: error while polling for resource cleanup on CertificateManagerCertificateMapEntry with name 'certificatemanagercertificatemapentry-sample': timed out waiting for the condition; last seen status: map[conditions:[map[lastTransitionTime:2024-01-20T15:04:00Z message:Delete call failed: error fetching live state: error getting ID for resource: error getting value from reference: could not parse reference resolution value '<nil>' as string reason:DeleteFailed status:False type:Ready]] observedGeneration:3 state:SERVING_STATE_UNSPECIFIED]
2024-01-20T15:34:00.8549457Z 2024/01/20 15:34:00 [DEBUG] parent context canceled, cleaning up batcher batches
```

i.e. it is a deletion ordering problem on CertificateMapCertificateMapEntry, which I think is https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/991

`2024/01/22 21:29:08 [DEBUG] Retry Transport: Stopping retries, last request failed with non-retryable error: googleapi: got HTTP response code 404 with body: HTTP/0.0 404 Not Found` is a message from one of our libraries saying that it isn't going to retry the particular request.  Normally the client libraries retry if they get a transient error e.g. a network timeout, though I would have expected logged in the error case (retry) rather than the normal case (no retry needed) - I guess it is debug level though!

KCC will continue the reconciliation - i.e. I don't think this is actually causing a test failure.  I actually think we expect a 404 on deletion, i.e. we are confirming the object is gone.

justinsb:
So I think the flakes were deletion ordering and were fixed by your recent PR @gemmahou

gemmahou:
> So I think the flakes were deletion ordering and were fixed by your recent PR @gemmahou

In my recent PR, I noticed that the mock test for CertificateMapCertificateMapEntry still fails. I didn't include this resource in my deletion allowlist, I need to take a look at it.
So let's keep https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/991 open.

PR #1100, opened 2024-01-06
WIP: Implement mustache-style templating in composition

justinsb:
A simple "starter" templating language inspired by mustache


justinsb:
We're doing this in KRO now

/close

PR #1103, opened 2024-01-08
Create very simple test for operators

justinsb:
- Add e2e test for ConfigConnector operator
- operator: build scheme once and reuse in tests

### DIFF
--- a/operator/cmd/manager/main.go
+++ b/operator/cmd/manager/main.go
@@ -106,7 +106,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err = configconnector.Add(mgr, repoPath); err != nil {
+	if err := configconnector.Add(mgr, repoPath); err != nil {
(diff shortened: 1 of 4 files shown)


justinsb:
We do need to make sure this fails tests, at first, given that we think that we need https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/368

justinsb:
Failing as expected:

```
=== NAME  TestE2E
    e2e_test.go:75: unexpected value for status.healthy: got 'false', want 'true'
    e2e_test.go:78: unexpected number of errors in status.errors: got 1, want 0. Got errors: [error during reconciliation: no matches for kind "" in version ""]
--- FAIL: TestE2E (25.25s)
```

justinsb:
Should now be passing after kdp upgrade :crossed_fingers:

maqiuyujoyce (approved):
LGTM except for a few unblocking questions!

PR #1105, opened 2024-01-08
operator: create smoke test for ConfigConnectorContext

justinsb:


justinsb:
This likely won't pass until we have #1106 (that's why this is split out)

Also likely won't pass until we update kdp with https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/368

justinsb:
We merged #1106 which includes this

/close

PR #1114, opened 2024-01-10
Tests: close manager context before envtest context

justinsb:
As reported in
https://github.com/kubernetes-sigs/controller-runtime/issues/1571, not
doing this can cause timeout errors waiting for kube-apiserver to
stop.


jingyih (commented):
/lgtm

PR #1115, opened 2024-01-10
Add support for NetworkSecurityAddressGroup

katrielt:
### Change description

This PR adds support for NetworkSecurity Address Groups

Fixes b/319419968

### Tests you have done

go test -v -tags=integration ./pkg/controller/dynamic/ -test.run TestCreateNoChangeUpdateDelete -run-tests addressgroup-parent-project -timeout 900s
go test -v -tags=integration ./config/tests/samples/create -test.run TestAll -run-tests project-addressgroup


katrielt:
@justinsb presubmit failures seem to be unrelated. Appreciate a review when you have time.

katrielt:
> I reviewed the resulting schema and it looks good. But I do see this is v1beta1... if we could make this v1alpha1 I would be much comfortable, particularly with our plans around direct actuation
> 
> 1. @katrielt would starting with v1alpha1 work for your use case?
> 2. @maqiuyujoyce can we generate v1alpha1?

From my end I would rather have it be beta, so I don't have to do a future change from alpha to beta. This is a pretty standard resource, supported by the latest terraform 4, no gotchas or weird stuff while adding it.

maqiuyujoyce:
@justinsb yes, it can be auto-generated: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1169/files#diff-f860bab8090d5f92b1c9c3f0c938e5a54488f92b3414a2c18659101c36ca5b74

#1169 actually adds many auto-generated resource configs and we can easily turn them into v1alpha1 CRDs if we want.

PR #1116, opened 2024-01-12
Update kubebuilder-declarative-pattern

justinsb:
We are aiming to pick up https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/368


justinsb:
Done via dependabot

/close

PR #1133, opened 2024-01-19
tests: support mockkubeapiserver as E2E_KUBE_TARGET

justinsb:
This is much faster than envtest, at the expense of being less
accurate (most immediately relevant to us: webhooks are not supported).


gemmahou:
> 🤦🏼ah I was trying to give others a chance to see the comments and explicitly didn't use the "slash" (words of approval). I think I need to read more about the bot's scraping. It may also have been me clicking approve in github.
> 
> Re the comments highlighted I can take them on if you agree w them justin

No worries, I do think we should be able to merge the PR by ourselves, or maybe we can configure the bot to have at least 2 approvals to merge a PR?

justinsb:
I am surprised that it merged based on the "four letter word" without a slash.  My guess is that it was the github PR "approve" review.  If it happens again, we can dig into why...

PR #1135, opened 2024-01-22
mockgcp: Implement subnetworks

justinsb:


mockgcp/mockcompute/networksv1.go
```diff
@@ -127,3 +129,34 @@ func (s *NetworksV1) Delete(ctx context.Context, req *pb.DeleteNetworkRequest) (
...
+		}
+
+		name := &networkName{
+			Project: project,
+			Name:    tokens[4],
+		}
+
+		return name, nil
+	} else {
+		return nil, status.Errorf(codes.InvalidArgument, "name %q is not valid", name)
```
cheftako, line 160: Can we add either the token length or expected name format to the error message?

mockgcp/mockcompute/subnetsv1.go
```diff
@@ -0,0 +1,153 @@
...
+	pb "github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/generated/mockgcp/cloud/compute/v1"
+)
+
+type SubnetsV1 struct {
+	*MockService
+	pb.UnimplementedSubnetworksServer
+}
+
+func (s *SubnetsV1) Get(ctx context.Context, req *pb.GetSubnetworkRequest) (*pb.Subnetwork, error) {
+	name, err := s.parseSubnetName("projects/" + req.GetProject() + "/regions/" + req.GetRegion() + "/subnetworks/" + req.GetSubnetwork())
```
cheftako, line 36: Silly question but why are we assembling a subnet name here from the 3 component parts, just to parse out those same 3 pieces? I'd get it if this were a public API but its private and all calling sites have the 3 relevant fields...

mockgcp/mockcompute/networksv1.go
```diff
@@ -127,3 +129,34 @@ func (s *NetworksV1) Delete(ctx context.Context, req *pb.DeleteNetworkRequest) (
...
+	Name    string
+}
+
+func (n *networkName) String() string {
+	return "projects/" + n.Project.ID + "/global" + "/networks/" + n.Name
+}
+
+// parseNetworkName parses a string into a networkName.
+// The expected form is `projects/*/global/networks/*`.
+func (s *MockService) parseNetworkName(name string) (*networkName, error) {
```
cheftako, line 144: Wondering why we want to intermediary name field for an internal function. Not seeing any signs of having it in single string form. All calling sites seem to already have the 2 fields already broken out and then we assemble the string so we can parse out those 2 fields.

mockgcp/mockcompute/subnetsv1.go
justinsb: So there is some logic in parseSubnetName around the project.  I'm also trying to normalize the functions across the APIs, hopefully to play into AI.  We are using the normalized name (which we get from `String`) as a key into our "database", so we do want something to normalize the names.

The name parsing also resolves the project (and converts it to an ID for the canonical key), and it could also do things like normalize regions etc in future.

mockgcp/mockcompute/networksv1.go
justinsb: Similar to the other question, we do resolve the project and normalize it.

I am wondering if we should generate the parse and String functions, e.g. using tags on the struct, but that feels like (1) something AI could do and (2) "pre AI" thinking, i.e. it's better to have concrete code so the AI can operate on it (even if there's more of it), than to have complex meta programming that the AI has no idea about.

mockgcp/mockcompute/networksv1.go
justinsb: I'm not sure it gets us anything?  We would not expect this error path to be hit, because we wouldn't expect terraform or the go SDK to construct invalid URLs - and even if they did, to reach this function the URLs have to be valid enough to get through the `google.api.http` grpc-gateway matching, e.g. here the proto matching is this:

```
  // Creates a network in the specified project using the data included in the request.
  rpc Insert(InsertNetworkRequest) returns (Operation) {
    option (google.api.http) = {
      body: "network_resource"
      post: "/compute/v1/projects/{project}/global/networks"
    };
    option (google.api.method_signature) = "project,network_resource";
    option (mockgcp.cloud.operation_service) = "GlobalOperations";
  }
```

So it has to be of the form `/compute/v1/projects/{project}/global/networks` anyway

And compute is actually the least strict here, most other protos look more like this:

```
  // Creates a new Network in a given project and location.
  rpc CreateNetwork(CreateNetworkRequest)
      returns (google.longrunning.Operation) {
    option (google.api.http) = {
      post: "/v1/{parent=projects/*/locations/*/zones/*}/networks"
      body: "network"
    };
    option (google.api.method_signature) = "parent,network,network_id";
    option (google.longrunning.operation_info) = {
      response_type: "Network"
      metadata_type: "OperationMetadata"
    };
  }
```

mockgcp/mockcompute/subnetsv1.go
cheftako: Personally I think I would rather have a struct which is the normal way of passing the data around. Internally we should always use the struct and leave the parse function at our API boundary where we can A) easily give an error to the correct caller if there is a problem with the input string and b) not have to worry about parsing and serializing throughout the code base as well as dealing with lack of typing and partial changes to this code.

cheftako:
/lgtm
/approve

I think we should chat about moving the parse field (project, location, zone) to our API boundary. However I don't want to block this change on that, so I am going ahead and approving.

PR #1140, opened 2024-01-23
Move the state-into-spec defaulter from webhook to controllers

maqiuyujoyce:
### Change description

Fixes part of b/320784855

* Moved the defaulter for the `state-into-spec` annotation from the webhook to the controllers.
  * The webhook is at the cluster level and the controllers are at the namespace level. In order to support the default value overrides for the `state-into-spec` annotation at the namespace level, we decided that the defaulter should be moved to the controllers.
  * To distinguish between the following cases, a `default-state-into-spec` annotation is added:
    * New resource creation without defaulting: Validation is done for the `state-into-spec` annotation.
    * New resource creation with defaulting: Defaulting is done with the default value the controller chooses. 
    * Resource upgrade: Resources is ensured to have `state-into-spec: merge` for backwards compatibility purpose.

### Tests you have done

Ran integration tests locally to verify the DCL-based resources, TF-based resources, and IAMPolicy resources are still working.

### DIFF
--- a/pkg/k8s/constants.go
+++ b/pkg/k8s/constants.go
@@ -80,6 +80,9 @@ const (
 	StateMergeIntoSpec = "merge"
 	StateAbsentInSpec  = "absent"
 
+	// Default state into spec indicator value
+	DefaultStateIntoSpecIndicator = "true"
+
 	// Core kubernetes constants
 	LastAppliedConfigurationAnnotation = "kubectl.kubernetes.io/last-applied-configuration"
 	ManagedFieldsTypeFieldsV1          = "FieldsV1"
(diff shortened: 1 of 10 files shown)


pkg/k8s/constants.go
```diff
@@ -137,6 +140,11 @@ var (
 		StateMergeIntoSpec,
 		StateAbsentInSpec,
 	}
+	DefaultStateIntoSpecIndicatorAnnotation = FormatAnnotation("default-state-into-spec")
+	DefaultStateIntoSpecIndicatorValues     = []string{
```
diviner524, line 144: I noticed the design doc is already updated, are you going to update this PR and simplify the implementation? (removing `default-state-into-spec` annotation).

maqiuyujoyce: Hi @diviner524 sorry for the confusion. The updated change is in #1149 . Will abandon this PR.

maqiuyujoyce:
Replaced by #1149 .

Issue #1144, opened 2024-01-23
github-action robot account has not signed CLA so we can't merge auto-generate PRs


justinsb:
### Checklist

- [X] I did not find a related open enhancement request.
- [X] I understand that enhancement requests filed in the GitHub repository are by default low priority.
- [X] If this request is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Describe the feature or resource

e.g. https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1142 could not be merged

justinsb:
I'll take a look at this one as it may require changes in https://github.com/GoogleCloudPlatform/oss-test-infra/tree/master

ldanielmadariaga:
This item has been open for >12 months. As part of scheduled repository maintenance, this Issue/PR will be closed. If this is still a critical ask, please re-open and work with your Google Cloud Account team to prioritize it with GCP KCC Product Management.

PR #1146, opened 2024-01-24
build: set an email for github actions bot

justinsb:
The customary email is noreply@github.com.

Without setting an email, we can't pass the CLA.

Issue #1144

### DIFF
--- a/.github/workflows/upgrade-operator.yaml
+++ b/.github/workflows/upgrade-operator.yaml
@@ -40,7 +40,7 @@ jobs:
       - name: "Check current state"
         run: |
           git config user.name "GitHub Actions Bot"
-          git config user.email "<>"
+          git config user.email "noreply@github.com"
           echo "Checking the current state..."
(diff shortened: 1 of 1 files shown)


.github/workflows/upgrade-operator.yaml
```diff
@@ -40,7 +40,7 @@ jobs:
       - name: "Check current state"
         run: |
           git config user.name "GitHub Actions Bot"
-          git config user.email "<>"
+          git config user.email "noreply@github.com"
```
diviner524, line 43: Is this an actual noreply email address provided by Github, or it is just a placeholder?

diviner524 (commented):
/lgtm

.github/workflows/upgrade-operator.yaml
justinsb: It's the one that people seem to use.  I've discovered that we aren't going to be able to get the CLA to pass with this robot account anyway, so there's still more to do here

PR #1147, opened 2024-01-24
add annotation to sample; troubleshooting instructions

acpana:
Some small changes to the getting started guide that I'd have found useful going thru and a small fix potentially to the sample from the getting started guide.

### DIFF
--- a/config/samples/resources/artifactregistryrepository/artifactregistry_v1beta1_artifactregistryrepository.yaml
+++ b/config/samples/resources/artifactregistryrepository/artifactregistry_v1beta1_artifactregistryrepository.yaml
@@ -18,6 +18,9 @@ metadata:
   name: artifactregistryrepository-sample
   labels:
     label-one: "value-one"
+  annotations:
+    # Replace ${PROJECT_ID?} with your project ID.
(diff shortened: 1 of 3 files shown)


jingyih (commented):
/lgtm

Thank you for improving the doc!

config/samples/resources/artifactregistryrepository/artifactregistry_v1beta1_artifactregistryrepository.yaml
```diff
@@ -18,6 +18,9 @@ metadata:
   name: artifactregistryrepository-sample
   labels:
     label-one: "value-one"
+  annotations:
```
justinsb, line 21: Just FYI these samples do make it into our docs (and tests), so we have to be a little careful.  That said this is done in a bunch of other places so should be fine.  _Maybe_ we shouldn't be referencing the sample here, but I'm not sure I feel too strongly about that.

justinsb:
Looks good but please fix that typo.

Also the test is failing because:
`ERROR: Resource docs must be regenerated. Please run 'make ready-pr' or 'make resource-docs' and update your PR.`

So we are checking that the generated docs (which do include the sample file you changed) are up-to-date - that's a good thing, I think!

PR #1149, opened 2024-01-25
Move the state-into-spec defaulter from webhook to controllers

maqiuyujoyce:
### Change description

Fixes b/323999277

* Moved the defaulter for the `state-into-spec` annotation from the webhook to the controllers.
  * The webhook is at the cluster level and the controllers are at the namespace level. In order to support the default value overrides for the `state-into-spec` annotation at the namespace level, we decided that the defaulter should be moved to the controllers.
  * We assume that backwards compatibility is not a concern because users will first upgrade Config Connector to the new version, and after the upgrade is completed (all the existing CRs have been reconciled and all the unset `state-into-spec` annotations have been defaulted to `merge`), they will leverage the new feature to override the namespace-level `state-into-spec` annotation.
* Fixed the EnsureFinalizers function as it would wipe out the defaulted annotations including the `state-into-spec` annotation.

This is a improved and simplified solution compared with #1140 .

### Tests you have done

Ran integration tests locally to verify the DCL-based resources, TF-based resources, and IAMPolicy resources are still working.

### DIFF
--- a/pkg/controller/iam/partialpolicy/iampartialpolicy_controller.go
+++ b/pkg/controller/iam/partialpolicy/iampartialpolicy_controller.go
@@ -152,7 +152,7 @@ func (r *ReconcileIAMPartialPolicy) Reconcile(ctx context.Context, request recon
 		return reconcile.Result{}, err
 	}
 	if err := r.handleDefaultStateIntoSpecValue(policy); err != nil {
-		return reconcile.Result{}, fmt.Errorf("error handling default values for IAM parital policy '%v': %v", k8s.GetNamespacedName(policy), err)
+		return reconcile.Result{}, fmt.Errorf("error handling default values for IAM partial policy '%v': %w", k8s.GetNamespacedName(policy), err)
 	}
 	runCtx := &reconcileContext{
 		Reconciler:     r,
@@ -178,7 +178,7 @@ func (r *ReconcileIAMPartialPolicy) handleDefaultStateIntoSpecValue(pp *iamv1bet
 	// Validate or set the default value (cluster-level or namespace-level) for
 	// the 'state-into-spec' annotation.
 	if err := k8s.ValidateOrDefaultStateIntoSpecAnnotation(pp, k8s.StateMergeIntoSpec); err != nil {
-		return fmt.Errorf("error validating or defaulting the '%v' annotation for resource '%v': %v", k8s.StateIntoSpecAnnotation, k8s.GetNamespacedName(pp), err)
+		return fmt.Errorf("error validating or defaulting the '%v' annotation for resource '%v': %w", k8s.StateIntoSpecAnnotation, k8s.GetNamespacedName(pp), err)
 	}
 	return nil
 }
--- a/pkg/controller/iam/partialpolicy/iampartialpolicy_controller.go
+++ b/pkg/controller/iam/partialpolicy/iampartialpolicy_controller.go
@@ -151,7 +151,7 @@ func (r *ReconcileIAMPartialPolicy) Reconcile(ctx context.Context, request recon
 		// Error reading the object - requeue the request.
 		return reconcile.Result{}, err
 	}
-	if err := r.handleDefault(policy); err != nil {
+	if err := r.handleDefaultStateIntoSpecValue(policy); err != nil {
 		return reconcile.Result{}, fmt.Errorf("error handling default values for IAM parital policy '%v': %v", k8s.GetNamespacedName(policy), err)
 	}
 	runCtx := &reconcileContext{
@@ -174,7 +174,7 @@ func (r *ReconcileIAMPartialPolicy) Reconcile(ctx context.Context, request recon
 	return reconcile.Result{RequeueAfter: jitteredPeriod}, nil
 }
 
-func (r *ReconcileIAMPartialPolicy) handleDefault(pp *iamv1beta1.IAMPartialPolicy) error {
+func (r *ReconcileIAMPartialPolicy) handleDefaultStateIntoSpecValue(pp *iamv1beta1.IAMPartialPolicy) error {
 	// Validate or set the default value (cluster-level or namespace-level) for
 	// the 'state-into-spec' annotation.
 	if err := k8s.ValidateOrDefaultStateIntoSpecAnnotation(pp, k8s.StateMergeIntoSpec); err != nil {
--- a/pkg/controller/iam/partialpolicy/iampartialpolicy_controller.go
+++ b/pkg/controller/iam/partialpolicy/iampartialpolicy_controller.go
@@ -151,6 +151,9 @@ func (r *ReconcileIAMPartialPolicy) Reconcile(ctx context.Context, request recon
 		// Error reading the object - requeue the request.
 		return reconcile.Result{}, err
 	}
+	if err := r.handleDefault(policy); err != nil {
+		return reconcile.Result{}, fmt.Errorf("error handling default values for IAM parital policy '%v': %v", k8s.GetNamespacedName(policy), err)
+	}
 	runCtx := &reconcileContext{
 		Reconciler:     r,
 		Ctx:            ctx,
@@ -169,6 +172,15 @@ func (r *ReconcileIAMPartialPolicy) Reconcile(ctx context.Context, request recon
 	}
 	logger.Info("successfully finished reconcile", "resource", request.NamespacedName, "time to next reconciliation", jitteredPeriod)
 	return reconcile.Result{RequeueAfter: jitteredPeriod}, nil
+}
+
+func (r *ReconcileIAMPartialPolicy) handleDefault(pp *iamv1beta1.IAMPartialPolicy) error {
+	// Validate or set the default value (cluster-level or namespace-level) for
+	// the 'state-into-spec' annotation.
+	if err := k8s.ValidateOrDefaultStateIntoSpecAnnotation(pp, k8s.StateMergeIntoSpec); err != nil {
+		return fmt.Errorf("error validating or defaulting the '%v' annotation for resource '%v': %v", k8s.StateIntoSpecAnnotation, k8s.GetNamespacedName(pp), err)
+	}
+	return nil
 }
 
 func (r *reconcileContext) doReconcile(pp *iamv1beta1.IAMPartialPolicy) (requeue bool, err error) {
--- a/pkg/controller/lifecyclehandler/handler.go
+++ b/pkg/controller/lifecyclehandler/handler.go
@@ -197,20 +197,22 @@ func reasonForUnresolvableDeps(err error) (string, error) {
 
 func (r *LifecycleHandler) EnsureFinalizers(ctx context.Context, original, resource *k8s.Resource, finalizers ...string) error {
 	if !k8s.EnsureFinalizers(resource, finalizers...) {
-		u, err := original.MarshalAsUnstructured()
+		uo, err := original.MarshalAsUnstructured()
 		if err != nil {
 			return err
 		}
-		copy, err := k8s.NewResource(u)
+		originalCopy, err := k8s.NewResource(uo)
 		if err != nil {
 			return err
 		}
-		if !k8s.EnsureFinalizers(copy, finalizers...) {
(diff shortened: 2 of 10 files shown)


pkg/controller/lifecyclehandler/handler.go
```diff
@@ -197,20 +197,22 @@ func reasonForUnresolvableDeps(err error) (string, error) {
...
-			if err := r.updateAPIServer(ctx, copy); err != nil {
+		if !k8s.EnsureFinalizers(originalCopy, finalizers...) {
+			if err := r.updateAPIServer(ctx, originalCopy); err != nil {
 				return err
 			}
-			// sync the resource up with the updated metadata.
-			resource.ObjectMeta = copy.ObjectMeta
+			// Sync the resource up with the updated metadata except for the
+			// defaulted / pre-processed annotations.
+			originalCopy.ObjectMeta.Annotations = resource.ObjectMeta.Annotations
```
diviner524, line 214: Besides `state-into-spec`, are there any other annotations that might have been updated in `resource.ObjectMeta.Annotations` and needs to be reflected in `originalCopy.ObjectMeta.Annotations`?

pkg/controller/tf/controller.go
```diff
@@ -174,6 +174,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (res
 	if err != nil {
 		return reconcile.Result{}, fmt.Errorf("could not parse resource %s: %v", req.NamespacedName.String(), err)
 	}
+	if err := r.handleDefault(resource); err != nil {
```
diviner524, line 177: IIUC `handleDefault` is only handling default for state-into-spec. So shall we name it `handleDefaultStateIntoSpecValue` instead?

pkg/k8s/state_into_spec_annotation.go
```diff
@@ -19,24 +19,28 @@ import (
...
-	return validateStateIntoSpecAnnotation(obj, obj.GroupVersionKind())
+	defaultStateIntoSpecAnnotation(obj, gvk, defaultValue)
+	return nil
 }
 
-func EnsureSpecIntoSateAnnotation(obj *Resource) error {
-	_, found := GetAnnotation(StateIntoSpecAnnotation, obj)
-	if !found {
+func defaultStateIntoSpecAnnotation(obj metav1.Object, gvk schema.GroupVersionKind, defaultValue string) {
+	if defaultValue == StateAbsentInSpec && !ResourceSupportsStateAbsentInSpec(gvk.Kind) {
```
diviner524, line 38: I suggest we remove `ResourceSupportsStateAbsentInSpec` now that we support state-into-status.

IIRC the only reason this method exist is because we were not able to populate the ip address of ComputeAddress resource into status.

IMO we should update ComputeAddress CRD to surface that info in status so all resources will support state absent.

pkg/controller/lifecyclehandler/handler.go
maqiuyujoyce: There is a rare case that [`management-conflict-prevention-policy` annotation is added for backwards compatibility in the controller](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/tf/controller.go#L407) and got wiped out here. Whenever an annotation is defaulted by the webhook, there shouldn't be any issue.

pkg/controller/tf/controller.go
maqiuyujoyce: Done.

pkg/k8s/state_into_spec_annotation.go
maqiuyujoyce: Good point! Skipped the special handling here, and created b/322836859 and b/322839663 to remove the special handling completely.

pkg/controller/tf/controller.go
justinsb: I quite liked handleDefault because it hinted that we can use this approach to eliminate all the mutating webhooks, but ... not a blocker.

pkg/controller/iam/partialpolicy/iampartialpolicy_controller.go
```diff
@@ -151,6 +151,9 @@ func (r *ReconcileIAMPartialPolicy) Reconcile(ctx context.Context, request recon
 		// Error reading the object - requeue the request.
 		return reconcile.Result{}, err
 	}
+	if err := r.handleDefaultStateIntoSpecValue(policy); err != nil {
+		return reconcile.Result{}, fmt.Errorf("error handling default values for IAM parital policy '%v': %v", k8s.GetNamespacedName(policy), err)
```
justinsb, line 155: Nit: typo in parital.

Nit: In general, we should use %w with fmt.Errorf.  (But not in klog.Warningf, which causes me great frustration!)

pkg/controller/iam/partialpolicy/iampartialpolicy_controller.go
```diff
@@ -171,6 +174,15 @@ func (r *ReconcileIAMPartialPolicy) Reconcile(ctx context.Context, request recon
 	return reconcile.Result{RequeueAfter: jitteredPeriod}, nil
 }
 
+func (r *ReconcileIAMPartialPolicy) handleDefaultStateIntoSpecValue(pp *iamv1beta1.IAMPartialPolicy) error {
```
justinsb, line 177: I note that we don't have the call to `EnsureSpecIntoStateAnnotation` here in these reconcilers.  Should that have been here previously?

pkg/controller/lifecyclehandler/handler.go
```diff
@@ -197,20 +197,22 @@ func reasonForUnresolvableDeps(err error) (string, error) {
...
 			return err
 		}
-		copy, err := k8s.NewResource(u)
+		originalCopy, err := k8s.NewResource(uo)
 		if err != nil {
 			return err
 		}
-		if !k8s.EnsureFinalizers(copy, finalizers...) {
-			if err := r.updateAPIServer(ctx, copy); err != nil {
+		if !k8s.EnsureFinalizers(originalCopy, finalizers...) {
```
justinsb, line 208: I'm confused by what's going on here (or why we have to change it).  Maybe the comment should explain what happens if we don't do this i.e. why are we doing this syncing.

(I suspect the "full" answer is that we should avoid having multiple copies of the Resource, and we should just use `client.Object` - often *unstructured.Unstructured.  But I do think that's a bigger change so I understand not making it here.  I think if we applied the finalizers first - which is more conventional - we could make this change safely)

pkg/k8s/state_into_spec_annotation.go
```diff
@@ -19,24 +19,19 @@ import (
...
 
-func ValidateOrDefaultStateIntoSpecAnnotation(obj *unstructured.Unstructured) error {
+// ValidateOrDefaultStateIntoSpecAnnotation validates the value of the
+// 'state-into-spec' annotation if it is set and defaults the annotation to the
+// passed in defaultValue if it is unset.
+func ValidateOrDefaultStateIntoSpecAnnotation(obj metav1.Object, gvk schema.GroupVersionKind, defaultValue string) error {
 	_, found := GetAnnotation(StateIntoSpecAnnotation, obj)
 	if !found {
-		SetAnnotation(StateIntoSpecAnnotation, StateMergeIntoSpec, obj)
+		// TODO(b/322836859): Ensure ComputeAddress doesn't need special handling.
```
justinsb, line 31: I think we can drop this todo, because #1162 has merged?

pkg/k8s/state_into_spec_annotation.go
```diff
@@ -19,24 +19,19 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
-func ValidateOrDefaultStateIntoSpecAnnotation(obj *unstructured.Unstructured) error {
+// ValidateOrDefaultStateIntoSpecAnnotation validates the value of the
+// 'state-into-spec' annotation if it is set and defaults the annotation to the
+// passed in defaultValue if it is unset.
+func ValidateOrDefaultStateIntoSpecAnnotation(obj metav1.Object, gvk schema.GroupVersionKind, defaultValue string) error {
```
justinsb, line 28: Nit: If we pass `obj client.Object`, we can avoid passing the GVK also I believe.  I'm not sure we even need the GVK anyway!

pkg/webhook/generic_defaulter.go
```diff
@@ -44,16 +38,5 @@ func NewGenericDefaulter() HandlerFunc {
...
-		klog.Error(err)
-		return admission.Errored(http.StatusBadRequest,
-			fmt.Errorf("error decoding object: %v", err))
-	}
-	newObj := obj.DeepCopy()
-	if err := k8s.ValidateOrDefaultStateIntoSpecAnnotation(newObj); err != nil {
-		return admission.Errored(http.StatusBadRequest, fmt.Errorf("error validating or defaulting '%v' annotation: %v", k8s.StateIntoSpecAnnotation, err))
-	}
-	return constructPatchResponse(obj, newObj)
+	return admission.ValidationResponse(true, "no-op: this webhook is deprecated")
```
justinsb, line 41: Hooray!

justinsb:
Generally LGTM.  Things marked as Nit are nits and don't need to addressed, but the questions about missing EnsureSpecIntoStateAnnotation calls and what's going on in that finalizer update seem more important.

Meta: I feel like I reviewed this PR previously with similar comments.  Did you see similar comments from me somewhere previously?

pkg/controller/iam/partialpolicy/iampartialpolicy_controller.go
justinsb: So in #1151 I think you suggested we just forgot; so that's fine.  I was just worried there was some reason...

maqiuyujoyce: @diviner524 do you happen to have any context about it? I personally don't know why the backwards compatible logic is not there for IAM resources.

pkg/controller/tf/controller.go
maqiuyujoyce: We can always generalize the function name. And as you suggested in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1151#discussion_r1472951138, I'll do some refactoring around the defaulting logic.

So I'll keep the code as is in this PR.

pkg/k8s/state_into_spec_annotation.go
maqiuyujoyce: Actually b/322839663 is also necessary. But will move it anyways since https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1178 is out.

pkg/controller/iam/partialpolicy/iampartialpolicy_controller.go
maqiuyujoyce: Done.

pkg/k8s/state_into_spec_annotation.go
maqiuyujoyce: Just for self-education purpose, are `metav1.Object`, `client.Object`, and `unstructured.Unstructured` just interfaces for the KRM object of different granularities?

Also with https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1178, a GVK is no longer needed. I'll need to change it after a rebase.

maqiuyujoyce: Rebased.

Didn't change it to `client.Object` as further refactoring will be needed.

pkg/controller/lifecyclehandler/handler.go
maqiuyujoyce: @justinsb do you consider EnsureFinalizers() a blocker? I think this may not be a super quick fix and can be addressed separately so created b/323998236 to track it. Right now I'd prefer prioritizing the `state-into-spec` namespace-level override feature.

justinsb:
Thanks @maqiuyujoyce 

/approve
/lgtm

PR #1151, opened 2024-01-26
Enable the configuration of state-into-spec default value for TF controller in kccmanager

maqiuyujoyce:
### Change description

Fixes part of b/320784855

* Supported struct `StateIntoSpecDefaulter` to validate, persist, determine and apply the default value of the `state-into-spec` annotation.
* Supported new fields `StateIntoSpecDefaultValue` and `StateIntoSpecUserOverride` in `kccmanager.Config` struct to pass the configuration of the `state-into-spec` annotation to the registration controller and then to the other controllers.
* Used the configuration passed from the kccmanager to determine the default value of the `state-into-spec` annotation in the TF controller.

### Tests you have done

Ran `go test -v -tags=integration ./pkg/controller/dynamic/ -test.run TestCreateNoChangeUpdateDelete -run-tests pubsubtopic -timeout 900s` and the created PubSubTopic has `state-into-spec: merge`.

Built and installed the operator locally, and the `state-into-spec` annotation is defaulted to `merge` for a PubSubTopic.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -26,6 +26,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/kccmanager"
 	controllermetrics "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/metrics"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/gcp/profiler"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/krmtotf"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/logging"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/metrics"
@@ -143,6 +144,8 @@ func newManager(ctx context.Context, restCfg *rest.Config, scopedNamespace strin
 
 	controllersCfg.UserProjectOverride = userProjectOverride
 	controllersCfg.BillingProject = billingProject
+	// TODO(b/320784855): StateIntoSpecDefaultValue and StateIntoSpecUserOverride values should come from the flags.
+	controllersCfg.StateIntoSpecDefaultValue = k8s.StateIntoSpecDefaultValueV1Beta1
 	mgr, err := kccmanager.New(ctx, restCfg, controllersCfg)
 	if err != nil {
 		return nil, fmt.Errorf("error creating manager: %w", err)
--- a/pkg/controller/registration/registration_controller.go
+++ b/pkg/controller/registration/registration_controller.go
@@ -56,16 +56,17 @@ var logger = klog.Log.WithName(controllerName)
 
 // Add creates a new registration Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
-func Add(mgr manager.Manager, p *tfschema.Provider, smLoader *servicemappingloader.ServiceMappingLoader, dclConfig *dcl.Config, dclConverter *conversion.Converter, regFunc registrationFunc) error {
+func Add(mgr manager.Manager, p *tfschema.Provider, smLoader *servicemappingloader.ServiceMappingLoader, dclConfig *dcl.Config, dclConverter *conversion.Converter, regFunc registrationFunc, stateIntoSpecValue *k8s.StateIntoSpecValue) error {
 	r := &ReconcileRegistration{
-		Client:           mgr.GetClient(),
-		provider:         p,
-		smLoader:         smLoader,
-		dclConfig:        dclConfig,
-		dclConverter:     dclConverter,
-		mgr:              mgr,
-		controllers:      make(map[string]map[string]controllerContext),
-		registrationFunc: regFunc,
+		Client:             mgr.GetClient(),
+		provider:           p,
+		smLoader:           smLoader,
+		dclConfig:          dclConfig,
+		dclConverter:       dclConverter,
+		mgr:                mgr,
+		controllers:        make(map[string]map[string]controllerContext),
+		registrationFunc:   regFunc,
+		stateIntoSpecValue: stateIntoSpecValue,
 	}
 	c, err := controller.New(controllerName, mgr,
 		controller.Options{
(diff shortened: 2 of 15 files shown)


pkg/k8s/constants.go
```diff
@@ -77,8 +77,9 @@ const (
 	ManagementConflictPreventionPolicyResource = "resource"
 
 	// State into spec annotation values
-	StateMergeIntoSpec = "merge"
-	StateAbsentInSpec  = "absent"
+	StateMergeIntoSpec               = "merge"
```
justinsb, line 80: Nit: More go-idiomatic would be `StateIntoSpecMerge`.  Sometimes with a `type StateIntoSpec string`.  Not important to fix here though, as it is pre-existing.

pkg/k8s/state_into_spec_annotation.go
```diff
@@ -66,8 +96,8 @@ func validateStateIntoSpecAnnotation(obj metav1.Object, gvk schema.GroupVersionK
 	return nil
 }
 
-func isAcceptedValue(val string) bool {
-	for _, v := range StateIntoSpecAnnotationValues {
+func isAcceptedValue(val string, acceptedValues []string) bool {
```
justinsb, line 99: Any reason to make this a parameter?

pkg/k8s/state_into_spec_annotation.go
```diff
@@ -19,24 +19,54 @@ import (
...
+// 'state-into-spec' annotation.
+func ValidateOrDefaultStateIntoSpecAnnotation(obj metav1.Object, gvk schema.GroupVersionKind, defaultValue string) error {
 	_, found := GetAnnotation(StateIntoSpecAnnotation, obj)
 	if !found {
+		defaultStateIntoSpecAnnotation(obj, gvk, defaultValue)
+	}
+	return validateStateIntoSpecAnnotation(obj, gvk)
+}
+
+func defaultStateIntoSpecAnnotation(obj metav1.Object, gvk schema.GroupVersionKind, defaultValue string) {
```
justinsb, line 63: Nit: We could write a function `GetGVKForKCCType(obj metav1.Object)` or use `runtime.Object`.  controller-runtime has https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client/apiutil#GVKForObject , though that takes a scheme and we could either have a global scheme instance or we could simply hard-code our own logic here.   That saves having to pass the GVK around.

cmd/manager/main.go
```diff
@@ -143,6 +144,8 @@ func newManager(ctx context.Context, restCfg *rest.Config, scopedNamespace strin
 
 	controllersCfg.UserProjectOverride = userProjectOverride
 	controllersCfg.BillingProject = billingProject
+	// TODO(b/320784855): StateIntoSpecDefaultValue and StateIntoSpecUserOverride values should come from the flags.
```
justinsb, line 147: At some point we are going to read the CCC directly in the reconciler rather than relaunching.  So we don't need to plumb this through the operator etc.

pkg/k8s/state_into_spec_annotation.go
```diff
@@ -19,24 +19,54 @@ import (
...
+	if defaultValue == StateAbsentInSpec && !ResourceSupportsStateAbsentInSpec(gvk.Kind) {
```
justinsb, line 64: Maybe we add address to status.observedState , so we can delete ResourceSupportsStateAbsentInSpec?

justinsb:
Generally LGTM, a few nits/comments/questions (nits don't generally need a reply, more for the future) but I don't think any of them are going to be blockers

maqiuyujoyce (commented):
Will need to do a force push before experimenting with the defaulter idea.

pkg/controller/registration/registration_controller.go
```diff
@@ -56,16 +56,17 @@ var logger = klog.Log.WithName(controllerName)
 
 // Add creates a new registration Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
-func Add(mgr manager.Manager, p *tfschema.Provider, smLoader *servicemappingloader.ServiceMappingLoader, dclConfig *dcl.Config, dclConverter *conversion.Converter, regFunc registrationFunc) error {
+func Add(mgr manager.Manager, p *tfschema.Provider, smLoader *servicemappingloader.ServiceMappingLoader, dclConfig *dcl.Config, dclConverter *conversion.Converter, regFunc registrationFunc, stateIntoSpecValue *k8s.StateIntoSpecValue) error {
```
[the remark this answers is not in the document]
maqiuyujoyce: Added the `Defaulter` interface and changed `StateIntoSpecValue` into `StateIntoSpecDefaulter`. If we are going to support more defaulters in the future, we probably can turn it into a defaulter map / struct (instead of appending more parameters).

We'll probably improve the generic interface over time when we support more defaulters.

Also we seem to not really do nil checks for pointers passed over function calls into the reconcilers. Are you suggesting that we should do it?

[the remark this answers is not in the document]
justinsb: Sorry, I should have been more verbose:

By defaulter I meant a function or interface:
```
type Defaulter interface {
  ApplyDefaults(ctx context.Context, obj client.Object) (changed bool, err error)
}
```

Then you do this:
```
type Reconciler struct {
...
  defaulter Defaulter
...
}

func (r *Reconciler) Reconcile(ctx context.Context...
...
  if defaulter != nil {
    _, err := r.defaulter.ApplyDefaults(ctx, obj); err != nil {
      return nil, err
      } 
  }
...
```

If you're going to do that, then you can avoid the nil check with this:

```
type NoopDefaulter struct {
}

var _ Defaulter = &NoopDefaulter{}

func (d *NoopDefaults) ApplyDefaults(ctx context.Context, obj client.Object) (bool, error) {
  return false, nil
}
```

But easier still may be to do this:
```
type Reconciler struct {
...
  defaulters []Defaulter
...
}

func (r *Reconciler) Reconcile(ctx context.Context...
...
  for _, defaulter := range r.defaulters {
    _, err := defaulter.ApplyDefaults(ctx, obj); err != nil {
      return nil, err
      } 
  }
...
```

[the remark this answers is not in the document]
maqiuyujoyce: Thank you for the detailed suggestions! I like the idea of using a list of defaulters. I think it's good to make the interface as generic as possible so I kept `ctx context.Context` in the parameters and `changed bool` in the returned values for `ApplyDefaults()`.

PR #1161, opened 2024-01-30
chore: updated smd, fix: use typed scalar

acpana:
~testing out proposed fix from~ context: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1122#issuecomment-1915758071

In lieu of https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1122

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -56,7 +56,7 @@ require (
 	sigs.k8s.io/controller-tools v0.6.2
 	sigs.k8s.io/kubebuilder-declarative-pattern v0.15.0-beta.2
 	sigs.k8s.io/kubebuilder-declarative-pattern/mockkubeapiserver v0.0.0-20230303024857-d1f76c15e05b
-	sigs.k8s.io/structured-merge-diff/v4 v4.2.3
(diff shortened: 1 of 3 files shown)


acpana:
@maqiuyujoyce / @diviner524 / @justinsb fyi keep me honest here:

my intuition is that this oneliner change is safe and will be transparent to anyone using kcc; this is isolated to how we "communicate" to smd.

justinsb:
Ah - I think I saw this the other day when looking at the OpenAPI, IIRC it's the `labels map[string]string` field where the type is object and then there's only one type (so I have no idea what happens if someone does map[string]int, maybe we need OpenAPI v4!).

I think this fix is correct, but maybe we could do some defensive coding to validate our assumptions here.  e.g. we could keep track of the current field path and klog.Warningf if it isn't `metadata.labels` (likely compare against a list/map of "well known" field paths?).

justinsb:
Thanks @acpana, this lgtm particularly with #1167

/approve
/lgtm

PR #1168, opened 2024-01-30
Generate APIKeys Key from Terraform

justinsb:
Following the existing (legacy) procedure to generate an APIKeys Key resource from terraform.

Because it's coming from terraform and we intend to rework it, mark it as alpha quality until we have reworked it.


justinsb:
/assign @maqiuyujoyce 

This is just going through the "legacy" terraform-backed generation process at this point.  Targeting v1alpha1 because it won't yet be the quality we want.

My plan is to then implement this directly using the API, to verify that it is worthwhile in terms of quality / efficient, that we can migrate existing TF resources etc etc.
 
I do wonder if I should put this CRD into a separate directory, just so that we can delay adoption for a little longer.  It is alpha though.

config/servicemappings/apikeys.yaml
```diff
@@ -0,0 +1,46 @@
...
+spec:
+  name: APIKeys
+  version: v1alpha1
+  serviceHostName: "apikeys.googleapis.com"
+  resources:
+    - name: google_apikeys_key
+      kind: APIKeysKey
+      metadataMapping:
+        name: name
+        #labels: labels
```
maqiuyujoyce, line 29: Remove?

mockgcp/fixup-third-party.sh
```diff
@@ -25,13 +25,17 @@ rm -rf mockgcp/ && mkdir -p mockgcp/api/
...
 mv google/iam/ mockgcp/
+mv google/api/apikeys/ mockgcp/api/
 mv google/api/serviceusage/ mockgcp/api/
 
 cd mockgcp
 
 find . -type f -print0 | xargs -0 sed -i -e "s@google/iam/@mockgcp/iam/@g"
 find . -type f -print0 | xargs -0 sed -i -e "s@google\.iam@mockgcp.iam@g"
 
+find . -type f -print0 | xargs -0 sed -i -e "s@google/api/apikeys/@mockgcp/api/apikeys/@g"
```
maqiuyujoyce, line 36: Just for self-education: Is it a required step when adding a mock API? I'm not sure if I've missed anything, but I didn't find them in the [README](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/mockgcp/README.md).

mockgcp/Makefile
```diff
@@ -26,6 +26,7 @@ gen-proto:
 		--grpc-gateway_opt paths=source_relative \
 		--experimental_allow_proto3_optional \
 		./apis/mockgcp/cloud/resourcemanager/v1/*.proto \
+		./third_party/googleapis/mockgcp/api/apikeys/v2/*.proto \
```
maqiuyujoyce, line 29: Nit: Maybe we can add a lint/a unit test to ensure the paths are ordered alphabetically?

config/servicemappings/apikeys.yaml
```diff
@@ -0,0 +1,46 @@
...
+# limitations under the License.
+
+apiVersion: core.cnrm.cloud.google.com/v1alpha1
+kind: ServiceMapping
+metadata:
+  name: apikeys.cnrm.cloud.google.com
+  namespace: cnrm-system
+spec:
+  name: APIKeys
+  version: v1alpha1
```
maqiuyujoyce, line 22: We don't have a service level version set to `v1alpha1` so there might be some rare feature gaps somewhere.

Meanwhile, we support all the other resources via autogen. So APIKeysKey configured here, I think there'll be two routes to promote a resource from alpha to beta. Feels like our code will get even more complicated.

mockgcp/mockapikeys/apikeysv2.go
```diff
@@ -0,0 +1,191 @@
+// Copyright 2023 Google LLC
```
maqiuyujoyce, line 1: Nit: 2024

maqiuyujoyce, line 1: Also this filename is apikeysv2.go, which seems to be different from the naming conventions you mentioned in the README: `Implement the core CRUD methods in <resourcename>.go.` or the [example mock resource PR](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/869/files#diff-2563d51fd13526bf66666826de6474615077cd2d9073b6f01aa340e24fff2897). Should the naming convention be adjusted or something? I'd lean towards an option that may help achieve consistency and enable the automation of this step in the future.

mockgcp/mockapikeys/apikeysv2.go
```diff
@@ -0,0 +1,191 @@
...
+			return nil, status.Errorf(codes.NotFound, "key %q not found", name)
+		} else {
+			return nil, status.Errorf(codes.Internal, "error reading key: %v", err)
+		}
+	}
+
+	return obj, nil
+}
+
+func (s *APIKeysV2) GetKeyString(ctx context.Context, req *pb.GetKeyStringRequest) (*pb.GetKeyStringResponse, error) {
```
maqiuyujoyce, line 57: Just for self-education purpose: How do we know that this additional function is needed? Is it only detectable by test failures?

mockgcp/mockapikeys/apikeysv2.go
```diff
@@ -0,0 +1,191 @@
...
+	obj := &pb.Key{}
+	if err := s.storage.Get(ctx, fqn, obj); err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, status.Errorf(codes.NotFound, "key %q not found", name)
+		} else {
+			return nil, status.Errorf(codes.Internal, "error reading key: %v", err)
+		}
+	}
+
+	// TODO: Do we need a real value here?
```
maqiuyujoyce, line 74: Could you make the TODO more actionable?

mockgcp/mockapikeys/apikeysv2.go
```diff
@@ -0,0 +1,191 @@
...
+		} else {
+			return nil, status.Errorf(codes.Internal, "error reading key: %v", err)
+		}
+	}
+
+	// From the proto:
+	// You can update only the `display_name`, `restrictions`, and `annotations` fields.
+
+	// See docs for UpdateMask
+	updateMask := req.GetUpdateMask()
```
maqiuyujoyce, line 127: Again just for self-education purpose: How do we know that an update mask is needed if we happen to update the right field in the mock test (i.e. didn't trigger any test failure)?

mockgcp/mockapikeys/names.go
```diff
@@ -0,0 +1,56 @@
...
+	"strings"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/common/projects"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type apiKeyName struct {
+	Project  *projects.ProjectData
+	Location string
```
maqiuyujoyce, line 27: Do we always make location value configurable even if the resource is global and has only one fixed location value, `global`?

mockgcp/mockapikeys/names.go
```diff
@@ -0,0 +1,56 @@
...
+	KeyName  string
+}
+
+func (n *apiKeyName) String() string {
+	return "projects/" + n.Project.ID + "/locations/" + n.Location + "/keys/" + n.KeyName
+}
+
+// parseAPIKeyName parses a string into a apiKeyName.
+// The expected form is projects/<projectID>/locations/<region>/keys/<keyName>
```
maqiuyujoyce, line 36: I'm guessing the expected form is `projects/<projectID>/locations/global/keys/<keyName>`?

pkg/resourceskeleton/testdata/asset-skeleton.yaml
```diff
@@ -26,8 +26,29 @@
...
+    metadata:
+      name: examplekey
+    spec:
+      projectRef:
+        external: "1234567890"
+
+# Service: artifactregistry.googleapis.com
+
+- resourceConfigId: ArtifactRegistryRepository # not supported by asset inventory
+
```
maqiuyujoyce, line 50: Nit: You seem to have added a few newlines, which most of this yaml file doesn't have. Is there any best practice we should follow?

pkg/test/constants/presubmitconstants.go
```diff
@@ -32,6 +32,7 @@ var (
 		"accesscontextmanager": {"accesscontextmanagerserviceperimeter"},
 		"alloydb":              {"fullalloydbcluster"},
 		"apigee":               {"apigeeenvironment"},
+		"apikeys":              {"apikeyskeybasic"},
```
maqiuyujoyce, line 35: I think we may not want to add v1alpha1 resources into presubmit (and get distracted by its failures)?

pkg/test/resourcefixture/testdata/basic/apikeys/v1alpha1/apikeyskey/apikeyskeybasic/create.yaml
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: apikeys.cnrm.cloud.google.com/v1alpha1
+kind: APIKeysKey
+metadata:
+  name: apikeyskeybasic-${uniqueId}
```
maqiuyujoyce, line 18: Should be `name: aapikeyskey-${uniqueId}`. The naming convention is `[kindinlowercase]-${uniqueId}`.

pkg/test/resourcefixture/testdata/basic/apikeys/v1alpha1/apikeyskey/apikeyskeybasic/update.yaml
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: apikeys.cnrm.cloud.google.com/v1alpha1
+kind: APIKeysKey
+metadata:
+  name: apikeyskeybasic-${uniqueId}
```
maqiuyujoyce, line 18: ditto

scripts/generate-google3-docs/resource-reference/templates/apikeys_apikeyskey.tmpl
```diff
@@ -0,0 +1,58 @@
+{{template "headercomment.tmpl" .}}
```
maqiuyujoyce, line 1: Mhmm, maybe remove the resource doc?
We didn't add any resource doc for any autogen / v1alpha1 resources because we believe additional information about no resource quality guaranteed, no backwards compatibility support, and etc should be provided on the page, but we never get the chance to prioritize working with PM to finalize the idea.

pkg/test/resourcefixture/sets.go
```diff
@@ -54,8 +54,16 @@ func GetBasicTypeSetCover(t *testing.T) []ResourceFixture {
 		return testType == Basic
 	}
 	heavyFilter := func(fixture ResourceFixture) bool {
+
 		// Skip v1alpha1 CRDs when testing set cover as they may not yet be
 		// correctly supported.
+
+		// We do allowlist a few that we are hand-coding / checking.
+		// (Needed to get TestGetBasicTypeSetCover to pass)
```
maqiuyujoyce, line 62: Hmmm, can we make TestGetBasicTypeSetCover() check for v1beta1 services only?

maqiuyujoyce (commented):
Hi @justinsb , thank you for adding the resource!

Most of the comments are questions, though I want to make sure we are on the same page before we merge the change.

There are still some process gaps to support a v1alpha1 resource, so I think we might want to make v1alpha1 resources less noticeable by customers before graduating them to beta/GA.

mockgcp/fixup-third-party.sh
justinsb: This is just when we add a new API outside of the normal google/cloud group.

However, I'll update the README!

mockgcp/Makefile
justinsb: We should aim to but I don't think we need to enforce it.  (It's also tricky to enforce)

config/servicemappings/apikeys.yaml
justinsb: Well we're going to switch this one to direct actuation pretty fast, it looks like!  There are some significant gaps in the terraform implementation.

mockgcp/mockapikeys/apikeysv2.go
justinsb: I wish :-)

justinsb: Renaming to match the name of the implemented proto service.

mockgcp/mockapikeys/apikeysv2.go
justinsb: Right - we implement the methods we need to pass the tests.  When we're missing one, we see an error like `"method GetKeyString not implemented"` in the logs, coming from pb.UnimplementedApiKeysServer

mockgcp/mockapikeys/apikeysv2.go
justinsb: Removed todo

mockgcp/mockapikeys/apikeysv2.go
justinsb: This is just a first-pass implementation based on what the docs on the proto describe.  There was actually some fun here about whether it is `display_name` or `displayName`, which I hope we'll be able to validate with the work in #957

mockgcp/mockapikeys/names.go
justinsb: Yes, we're parsing the value provided.  We aren't currently validating the location in any of our mocks (though we do save the object under a key which includes the location, in case we do support regions in future).

mockgcp/mockapikeys/names.go
justinsb: Fixed!

pkg/resourceskeleton/testdata/asset-skeleton.yaml
justinsb: I'm just trying to make it more human-readable (it's a human-generated/edited file I believe), so I put in white space and section headers to break up the text around my edit point.

pkg/test/constants/presubmitconstants.go
justinsb: So this is needed because of TestAllServicesInMap.

I do think we want to test it because we're actively developing it, and it's an important evaluation of a new approach.  The API itself is relatively old I believe, so I don't think we'll see too many service flakes.

pkg/test/resourcefixture/testdata/basic/apikeys/v1alpha1/apikeyskey/apikeyskeybasic/create.yaml
justinsb: Not sure I agree with this convention (we know the kind from the kind, so we get more information from [test name]-$uniqueId), but ... fixed

pkg/test/resourcefixture/testdata/basic/apikeys/v1alpha1/apikeyskey/apikeyskeybasic/update.yaml
justinsb: Done

pkg/test/resourcefixture/sets.go
justinsb: Changed `getResourceConfigIdSet` to ignore all v1alpha1 resources.

config/servicemappings/apikeys.yaml
justinsb: Done!

scripts/generate-google3-docs/resource-reference/templates/apikeys_apikeyskey.tmpl
justinsb: Removed the alpha docs.  We can add them back in a separate PR, we should also make sure they are readable on github for those alpha resources - a good opportunity.

pkg/test/constants/presubmitconstants.go
maqiuyujoyce: Sure, I'm okay with us keeping it as is. Though we may want to re-evaluate it if it adds more maintenance burden than expected.

maqiuyujoyce (approved):
LGTM! Comments are non-blockers.

mockgcp/mockapikeys/apikeysv2.go
justinsb: Right now it is "make the samples/fixtures tests pass".

Once we get #957, I think we'll be able to define a higher bar: capture the HTTP from real-gcp, make the mockgcp produce the same HTTP output (and make the samples/fixtures pass).  That is trial & error, but I believe we now call that "test driven development" ;-)

Yes, this requires the contributor to understand the API, but only the methods we are actually using.

PR #1183, opened 2024-02-02
WIP: Add finalizers before copying the unstructured.Unstructured.

justinsb:
client.Client is supposed to write-back to the object, allowing
further updates.

### DIFF
--- a/pkg/controller/tf/controller.go
+++ b/pkg/controller/tf/controller.go
@@ -154,6 +154,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (res 
 		}
 		return reconcile.Result{}, err
 	}
+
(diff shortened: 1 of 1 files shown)


justinsb:
cc @maqiuyujoyce  ... as long as we do it early enough, controller-runtime should write-back the new resourceVersion and allow further object updates.

maqiuyujoyce:
Thank you for the suggestion, @justinsb !

I took a deep look at [updateAPIServer()](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/lifecyclehandler/handler.go#L84) - I think the steps to preserve the status and the observedGeneration, and the step to check failure status after the update should necessary. The update to add the finalizers should be transparent to the users. What do you think?

ldanielmadariaga:

PR #1186, opened 2024-02-03
ci: add golangci-lint

acpana:
- adds golangci-lint gh action and config with 3 core linters: `revive` in lieu of golint, `errcheck` and `errorlint`
- make target runs in a docker container

review instructions:
- best reviewed by commit, in reverse chronological order.
- one can also use [`act`](https://github.com/nektos/act) to test the GH action e2e


justinsb:
Cool.  One trick I've seen that may be easier than fixing all 237 errors at once is to exempt certain directories, and then roll through them with smaller PRs.  Does that work here / is there another trick that's more fluent here?  Otherwise it looks like many of them aren't _too_ bad (e.g. strings.ReplaceAll should be easy), though I do worry that there will be a long tail of ones that may be more tedious to fix

justinsb:
Can we get the core linter change merged, and then I'm happy to review lots of smaller PRs each morning!

If it helps, I don't mind us checking in a "known bad" file with the things to ignore (assuming we can auto-generate that!), even if it starts large

config/tests/samples/create/samples.go
```diff
@@ -15,6 +15,7 @@
 package create
 
 import (
+	goerrors "errors"
```
justinsb, line 18: Nit (No need to fix), but I normally prefer to import apimachinery's errors as apierrors

operator/pkg/controllers/configconnector/configconnector_controller.go
```diff
@@ -373,38 +370,35 @@ func (r *ConfigConnectorReconciler) removeClusterModeOnlySharedComponents(ctx co
 	sc := &corev1.ServiceAccount{}
 	sc.Namespace = k8s.CNRMSystemNamespace
 	sc.Name = k8s.KCCControllerManagerComponent
-	if err := controllers.DeleteObject(ctx, c, sc); err != nil {
```
justinsb, line 376: Nit:This one is sort of debateable (like maybe we should wrap the error instead?), hopefully it doesn't come up _too_ often!

operator/pkg/preflight/upgrade_test.go
```diff
@@ -51,15 +51,15 @@ type FakeRepo struct {
 
 var _ manifest.Repository = &FakeRepo{}
 
-func (r *FakeRepo) LoadManifest(ctx context.Context, component string, id string, o declarative.DeclarativeObject) (map[string]string, error) {
+func (r *FakeRepo) LoadManifest(_ context.Context, _ string, _ string, _ declarative.DeclarativeObject) (map[string]string, error) {
```
justinsb, line 54: Nit: These seem a little overzealous when we're implementing an interface, but /shrug

operator/tests/e2e/e2e_test.go
```diff
@@ -137,16 +139,16 @@ func TestMain(m *testing.M) {
 
 	log, err := newLogger("TestMain")
 	if err != nil {
-		fmt.Printf("error creating logger: %v", err)
+		fmt.Print(fmt.Errorf("error creating logger: %w", err))
```
justinsb, line 142: Nit: I assume you did this for the linter, but it seems like the original version was fine

operator/tests/e2e/e2e_test.go
```diff
@@ -159,124 +161,124 @@ func TestMain(m *testing.M) {
...
+	testID, log, cluster, teardown := setup(t, testOptions)
 	if f.cleanup {
 		defer teardown()
 	}
 
-	manifestsDir, sample, err := getOperatorReleaseAssetsForVersion(f.version, testOptions.ServiceAccountID, testOptions.ProjectID, log)
+	manifestsDir, sample, err := getOperatorReleaseAssetsForVersion(f.version, testOptions.serviceAccountID, testOptions.ProjectID, log)
 	if err != nil {
-		t.Fatalf("error getting operator release assets for version '%v': %v", f.version, err)
+		t.Fatal(fmt.Errorf("error getting operator release assets for version '%v': %w", f.version, err))
```
justinsb, line 172: I like the %v / %w in general, but the linter seems to be just wrong here.

operator/tests/e2e/e2e_test.go
```diff
@@ -89,7 +91,7 @@ var (
 type TestOptions struct {
 	OrganizationID     string
 	BillingAccountID   string
-	ServiceAccountID   string
+	serviceAccountID   string
```
justinsb, line 94: I assume this is because the variable is not set anywhere?  That seems like an over-zealous linter...

acpana:
This PR was split into a 3/4 smaller ones each enabling errcheck, errorlint and revive respectively. Closing.

PR #1188, opened 2024-02-03
tests: even more careful cleanup

justinsb:
When we close the controller-manager runtime, wait for it to actually
shutdown.

The shutdown is fairly slow, and otherwise it may reconnect to
kube-apiserver during shutdown.

(Aside: not sure if this kube-apiserver approach to shutdown is sound; it seems likely to
cause production issues)


cheftako:
/lgtm
/approve

Seems a bit odd having the wait call before you call add. Having said that even on an early exit it looks ok.

PR #1190, opened 2024-02-04
mockgcp: services should return http.Handler

justinsb:
This avoids requiring grpc-gateway, and means we can more cleanly
inject advanced behaviours.

### DIFF
--- a/mockgcp/mock_http_roundtrip.go
+++ b/mockgcp/mock_http_roundtrip.go
@@ -27,7 +27,6 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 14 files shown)


cheftako:
`
Error: ./mock_http_roundtrip.go:91:30: cannot use mockcompute.New(env, storage) (value of type *mockcompute.MockService) as MockService value in argument to append: *mockcompute.MockService does not implement MockService (wrong type for method NewHTTPMux)
		have NewHTTPMux(context.Context, *grpc.ClientConn) (*"github.com/grpc-ecosystem/grpc-gateway/v2/runtime".ServeMux, error)
		want NewHTTPMux(context.Context, *grpc.ClientConn) (http.Handler, error)
`

cheftako:
I like the change but I think we missed a couple of instances.

justinsb:
/assign @cheftako 

Thanks @cheftako - I think I got them now!

hkundag:
I think this missed mockapikeys? https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/mockgcp/mockapikeys/service.go#L62

hkundag:
Actually this mock service didn't exist when this commit was made. Either way it needs to be updated.

hkundag:
Fixed in #1194

PR #1191, opened 2024-02-05
mockgcp: special case compute API operations

justinsb:
These operations don't follow the normal LRO pattern, so we just
create our own operations store rather than using the common
implementation.


gemmahou:
This PR need to be rebased with master.

mockgcp/mockcompute/operations.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// Gets the latest state of a long-running operation.  Clients can use this
+// method to poll the operation result at intervals as recommended by the API
+// service.
+func (s *computeOperations) getOperation(ctx context.Context, fqn string) (*pb.Operation, error) {
+	op := &pb.Operation{}
+	if err := s.storage.Get(ctx, fqn, op); err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.Infof("LRO not found for %v", fqn)
```
gemmahou, line 165: nit: I do not think the info adds any value here, since we have the error message below

mockgcp/mockcompute/regionaloperationsv1.go
```diff
@@ -0,0 +1,40 @@
...
+	pb.UnimplementedRegionOperationsServer
+}
+
+func (s *RegionalOperationsV1) Get(ctx context.Context, req *pb.GetRegionOperationRequest) (*pb.Operation, error) {
+	log := klog.FromContext(ctx)
+
+	fqn := s.regionalOperationFQN(req.Project, req.Region, req.Operation)
+	lro, err := s.getOperation(ctx, fqn)
+	if err != nil {
+		log.Info("operation not found", "fqn", fqn)
```
gemmahou, line 35: nit: same I do not think we need this info message

mockgcp/mockcompute/operations.go
justinsb: Good point; removed

mockgcp/mockcompute/regionaloperationsv1.go
justinsb: Thanks - removed!

justinsb:
Thanks @gemmahou - rebased and cleaned up those log statements!

justinsb:
@gemmahou I think it needs /approve, not just /lgtm

gemmahou (approved):

PR #1192, opened 2024-02-05
mockgcp: add mock for TagsTagKey

justinsb:
mockgcp: add mock for TagsTagKey

Also improve our golden testing framework to support this use case.

Builds on #957

### DIFF
--- a/tests/e2e/unified_test.go
+++ b/tests/e2e/unified_test.go
@@ -16,8 +16,10 @@ package e2e
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/GoogleCloudPlatform/k8s-config-connector/config/tests/samples/create"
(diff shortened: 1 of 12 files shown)


justinsb:
/assign @gemmahou 

With #957, we can start mocking these out more quickly by capturing the GCP results and then making the mock do the same thing.

justinsb:
Rebased now that #957 has merged (thank you @gemmahou !)

This now produces a stable _http.log, which is the same for both real gcp & mockgcp.

tests/e2e/unified_test.go
gemmahou, line 1: I would suggest to move the changes to a separate PR. The code change you have here actually targets a boarder problem: improve golden http log validation.

tests/e2e/unified_test.go
```diff
@@ -173,42 +222,61 @@ func TestAllInSeries(t *testing.T) {
...
+						tokens := strings.Split(path, ".")
+						jsonMutators = append(jsonMutators, func(obj map[string]any) {
+							_, found, _ := unstructured.NestedString(obj, tokens...)
+							if found {
+								unstructured.SetNestedField(obj, newValue, tokens...)
+							}
+						})
+					}
+
+					addReplacement("uniqueId", "111111111111111111111")
```
gemmahou, line 249: This method is already super long. Can we turn these "normalizations" to a helper function, like `[IgnoreComments](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/config/tests/samples/create/harness.go#L529)`?

gemmahou: Mark as resolve. As long as it's working properly, we can plan to do a cleanup later

tests/e2e/unified_test.go
gemmahou: Depends on you if we want to move to a separate PR or not. I won't worry too much about this now since we can do a cleanup later.

PR #1134, opened 2024-01-19
WIP: mockgcp: Support DNSPolicy

justinsb:
WIP because this is an "integration PR", I intend to cherry-pick pieces of it so that they are reviewable, but I also want to show the big picture.

- mockgcp: tool to convert OpenAPI to proto
- Generate with http annotations
- Fixup singular/plural naming
- Generated dns code
- mockgcp: support for dns policy
- Update generated for latest googleapis

PR #1193, opened 2024-02-05
mockgcp: Support for DNSPolicy

justinsb:
Splitting out from #1134.  In this PR, we magically assume we have a proto (there's some wiggle room about whether it should be proto2 vs proto3).

(The proto is produced in #1134, from the OpenAPI, but that mechanism is somewhat independent of the mockgcp implementation)


mockgcp/mockdns/names.go
```diff
@@ -0,0 +1,55 @@
+// Copyright 2023 Google LLC
```
cheftako, line 1: 2024?

mockgcp/mockdns/v1.go
```diff
@@ -0,0 +1,142 @@
...
+	update := req.GetPolicy()
+	if update.AlternativeNameServerConfig != nil {
+		obj.AlternativeNameServerConfig = update.AlternativeNameServerConfig
+	}
+
+	if update.Description != "" {
+		obj.Description = update.Description
+	}
+
+	// TODO: Do we need to use proto2?
```
cheftako, line 97: Do we want to keep the commented out code here?

mockgcp/mockdns/names.go
justinsb: Good call - this one is 2024 :-)

mockgcp/mockdns/v1.go
justinsb: Let's get #957 in and then we can make the call that way

ldanielmadariaga:

Issue #1199, opened 2024-02-06
e2e tests use the folder search method, which is eventually consistent


justinsb:
### Checklist

- [X] I did not find a related open enhancement request.
- [X] I understand that enhancement requests filed in the GitHub repository are by default low priority.
- [X] If this request is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Describe the feature or resource

Eventual consistency is not consistent with correctness :-)  And indeed, we are seeing lots of flakes.  If we're going to use an eventually consistent method, we should verify the results with a strongly consistent method.

justinsb:
We merged #1200 a while ago, hopefully we now have enough signal as to whether that was good enough to consider this bug fixed/mitigated

ldanielmadariaga:
This item has been open for >12 months. As part of scheduled repository maintenance, this Issue/PR will be closed. If this is still a critical ask, please re-open and work with your Google Cloud Account team to prioritize it with GCP KCC Product Management.

PR #1200, opened 2024-02-06
tests: when creating a folder, inject a wait afterwards

justinsb:
When we create a folder, we then try to reacquire it immediately
afterwards using the eventually-consistent search method.  Eventual
consistency is not good enough, and we regularly see flakes here.

This is a workaround to (hopefully) verify the root cause of these
flakes; we can then move to a strongly-consistent approach in a future
change.

Issue #1199

### DIFF
--- a/pkg/controller/dynamic/dynamic_controller_integration_test.go
+++ b/pkg/controller/dynamic/dynamic_controller_integration_test.go
@@ -591,6 +591,11 @@ func testReconcileAcquire(ctx context.Context, t *testing.T, testContext testrun
 		if gcpUnstruct, err = resourceContext.Create(ctx, t, unstructToCreate, systemContext.TFProvider, kubeClient, systemContext.SMLoader, systemContext.DCLConfig, systemContext.DCLConverter); err != nil {
 			t.Fatalf("unexpected error when creating GCP resource '%v': %v", unstructToCreate.GetName(), err)
 		}
+		if unstructToCreate.GroupVersionKind().Kind == "Folder" {
+			// We should not be using the search method, it is only eventually consistent.
+			t.Logf("created GCP Folder; waiting 60 seconds for eventual consistency to catch up")
+			time.Sleep(time.Minute)
+		}
 	}
 
 	// Acquire the resource using the original unstruct.


pkg/controller/dynamic/dynamic_controller_integration_test.go
```diff
@@ -591,6 +591,11 @@ func testReconcileAcquire(ctx context.Context, t *testing.T, testContext testrun
 		if gcpUnstruct, err = resourceContext.Create(ctx, t, unstructToCreate, systemContext.TFProvider, kubeClient, systemContext.SMLoader, systemContext.DCLConfig, systemContext.DCLConverter); err != nil {
 			t.Fatalf("unexpected error when creating GCP resource '%v': %v", unstructToCreate.GetName(), err)
 		}
+		if unstructToCreate.GroupVersionKind().Kind == "Folder" {
```
gemmahou, line 594: Can we handle this before the error occurs? i.e. move this block to after initialUnstruct(line 580)

justinsb: Not sure I follow - can you explain why?

gemmahou: Ah sorry, I misunderstood the logic. The resource gets created when we call `resourceContext.Create` and we want to hold a bit if it's Folder kind.

I just feel it's a bit hard to read to put this special handling of resource creation inside resourceContext.Get `err != nil` block.
Maybe like:
```
gcpUnstruct, err = resourceContext.Get(ctx, t, unstructToCreate, systemContext.TFProvider, kubeClient, systemContext.SMLoader, systemContext.DCLConfig, systemContext.DCLConverter)
if err != nil {
		if !strings.Contains(err.Error(), "not found") {
			t.Fatalf("unexpected error when GETting '%v': %v", unstructToCreate.GetName(), err)
		}
                if gcpUnstruct, err = resourceContext.Create(ctx, t, unstructToCreate, systemContext.TFProvider, kubeClient, systemContext.SMLoader, systemContext.DCLConfig, systemContext.DCLConverter); err != nil {
			                t.Fatalf("unexpected error when creating GCP resource '%v': %v", unstructToCreate.GetName(), err)
		}
}

// Hold for eventual consistency
if unstructToCreate.GroupVersionKind().Kind == "Folder" {
         // some code...
}
```

gemmahou: /lgtm

PR #1203, opened 2024-02-06
tests: Add more logging in e2e tests

justinsb:
In particular, we want to know if a reconcile operation is taking a
long time.

### DIFF
--- a/pkg/test/controller/reconcile.go
+++ b/pkg/test/controller/reconcile.go
@@ -193,15 +193,19 @@ func isTransientError(t *testing.T, err error) bool {
 }
 
 // RunReconcilerAssertResults asserts the expected state of the reconciler run.
-func RunReconcilerAssertResults(ctx context.Context, t *testing.T, reconciler reconcile.Reconciler, objectMeta v1.ObjectMeta,
+func RunReconcilerAssertResults(ctx context.Context, t *testing.T, reconciler reconcile.Reconciler,
+	kind string, objectMeta v1.ObjectMeta,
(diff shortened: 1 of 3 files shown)


pkg/test/controller/reconcile.go
```diff
@@ -193,15 +193,19 @@ func isTransientError(t *testing.T, err error) bool {
 }
 
 // RunReconcilerAssertResults asserts the expected state of the reconciler run.
-func RunReconcilerAssertResults(ctx context.Context, t *testing.T, reconciler reconcile.Reconciler, objectMeta v1.ObjectMeta,
+func RunReconcilerAssertResults(ctx context.Context, t *testing.T, reconciler reconcile.Reconciler,
+	kind string, objectMeta v1.ObjectMeta,
 	expectedResult reconcile.Result, expectedErrorRegex *regexp.Regexp) {
```
acpana, line 198: would it be worth creating some wrapper struct for all these parameters? something like

```golang
type testResources struct {
  reconciler reconcile.Reconciler,
  ...
}
```

i don't think this needs to happen in this PR just something for the future or maybe an issue in our backlog.

justinsb: Yes.  But these are the "old" tests - the new ones are in tests/e2e.  I'd rather we focused on making sure that the new tests cover anything the old tests did.  The "old" tests are quite hard to understand, hard to repoint to mockgcp, hard to add recording to etc etc

justinsb:
Thanks for looking @acpana !  I think you need to write "/lgtm" on its own line (?)

It's a small tweak, so

/hold cancel

PR #1204, opened 2024-02-06
tests: run tests grouped by GVK group

justinsb:
This will enable us to split the execution by service.

### DIFF
--- a/pkg/test/resourcefixture/test_runner.go
+++ b/pkg/test/resourcefixture/test_runner.go
@@ -17,10 +17,12 @@ package resourcefixture
 import (
 	"context"
 	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test"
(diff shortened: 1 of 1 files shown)


pkg/test/resourcefixture/test_runner.go
```diff
@@ -17,22 +17,46 @@ package resourcefixture
...
+		groups.Insert(tc.GVK.Group)
+	}
+
+	for _, group := range groups.List() {
+		group := group
+		groupTestName := strings.TrimSuffix(group, ".cnrm.cloud.google.com")
+		t.Run(groupTestName, func(t *testing.T) {
+			t.Parallel()
+			for _, tc := range filtered {
+				if tc.GVK.Group != group {
```
acpana, line 54:
```suggestion
			t.Parallel()
			for _, tc := range filtered {
			        tc := tc
				if tc.GVK.Group != group {
```

do we need to make sure tc doesn't change from for loop iteration to iteration?

acpana (approved):
/lgtm

PR #1205, opened 2024-02-07
script: cleanup temp generation folder

acpana:
I think this dir held the temp gen folder but now it's `github.com/...` so let's clean it up.


acpana:
cc @maqiuyujoyce in case you have any context and can confirm my assumption

maqiuyujoyce:
/lgtm 
@acpana Thank you for the fix!

PR #1209, opened 2024-02-08
mockgcp: Support LoggingLogMetric

justinsb:
Add mock support for LoggingLogMetric (our first in the logs API)


mockgcp/mocklogging/logmetric.go
```diff
@@ -0,0 +1,179 @@
...
+			return nil, status.Errorf(codes.Internal, "error reading metric: %v", err)
+		}
+	}
+
+	return obj, nil
+}
+
+func (s *metricsService) CreateLogMetric(ctx context.Context, req *pb.CreateLogMetricRequest) (*pb.LogMetric, error) {
+	reqName := req.Parent + "/metrics/" + req.GetMetric().GetName()
+	name, err := s.parseLogMetricName(reqName)
```
cheftako, line 61: I get the need for parseLogMetricName at line 40. I still think its better to just directly construct the logMetricName here. Heck have a helper function which takes the project id and name and returns the logMetricName, then both this spot and the parse method can call that.

mockgcp/mocklogging/logmetric.go
```diff
@@ -0,0 +1,179 @@
...
+	fqn := name.String()
+
+	now := time.Now()
+
+	obj := proto.Clone(req.GetMetric()).(*pb.LogMetric)
+	obj.Name = name.MetricName
+	obj.CreateTime = timestamppb.New(now)
+	obj.UpdateTime = timestamppb.New(now)
+
+	// Not clear why, but we return the values before populating the MetricDescriptor Name/Type
```
cheftako, line 75: /we return the values/we construct the return values/

cheftako: That is weird, especially for a mock. Timestamps can sometimes be sensitive for security.

mockgcp/mocklogging/logmetric.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	now := time.Now()
+
+	updated := req.GetMetric()
+	updated.Name = name.MetricName
+	updated.CreateTime = existing.CreateTime
+	updated.UpdateTime = timestamppb.New(now)
+
+	// Not clear why, but we return the values before populating the MetricDescriptor Name/Type
```
cheftako, line 119: Ditto

mockgcp/mocklogging/logmetric.go
```diff
@@ -0,0 +1,179 @@
...
+	retval := proto.Clone(updated).(*pb.LogMetric)
+
+	s.populateDefaultsForLogMetric(name, updated)
+
+	if err := s.storage.Update(ctx, fqn, updated); err != nil {
+		return nil, status.Errorf(codes.Internal, "error updating logmetric: %v", err)
```
cheftako, line 125: I believe even Update can return a not found error which should be returned as such and not an internal.

mockgcp/mocklogging/logmetric.go
```diff
@@ -0,0 +1,179 @@
...
+	obj.CreateTime = timestamppb.New(now)
+	obj.UpdateTime = timestamppb.New(now)
+
+	// Not clear why, but we return the values before populating the MetricDescriptor Name/Type
+	retval := proto.Clone(obj).(*pb.LogMetric)
+
+	s.populateDefaultsForLogMetric(name, obj)
+
+	if err := s.storage.Create(ctx, fqn, obj); err != nil {
+		return nil, status.Errorf(codes.Internal, "error creating metric: %v", err)
```
cheftako, line 81: Do we want to treat already exists errors similar to NotFound but return an AlreadyExists error?

mockgcp/mocklogging/logmetric.go
justinsb: I'll rephrase, but what I wrote is correct.  The value we return should not have these default values set.  When we get it immediately afterwards, the values are there.  (And there's no LRO in this API either, which is also odd)

mockgcp/mocklogging/logmetric.go
justinsb: I mean we technically should be running this in a transaction (or with optimistic concurrency), but we just fetched the object and got it.  I guess it's technically possible, but I'd rather figure out how to get OCC going.

mockgcp/mocklogging/logmetric.go
justinsb: Yeah, maybe what I'll do is look into OCC and stop using apierrors and start using status.Errorf with codes (so we can just `return nil, err` (and check the code first if we need to)

mockgcp/mocklogging/logmetric.go
justinsb: I did suspect you would flag this one.  The challenge is that now we need a method to parse the parent (which in this case is a project, so we do have that method), but now we have a second code path.  I don't think it's worth the extra line count, when the APIs are following the "modern" parent/name pattern.

mockgcp/mocklogging/logmetric.go
justinsb: I might do that in a separate PR though, as it will break all the open PRs and the ones that have yet to be opened.

mockgcp/mocklogging/logmetric.go
cheftako: *Pedant alert*
I agree your return value does not have the values. However your not returning before you populate the values. Your constructing the return value before you populate.

mockgcp/mocklogging/logmetric.go
cheftako: Fair enough, it just seems odd that we care about the distinction of resource existence errors from internal errors on some of these APIs but not others.

mockgcp/mocklogging/logmetric.go
cheftako: Saying "we return ... before" makes me think its talking about a return statement before some other action. 
As that isn't what its talking about I think its confusing.

mockgcp/mocklogging/logmetric.go
```diff
@@ -0,0 +1,167 @@
...
+
+	now := time.Now()
+
+	obj := proto.Clone(req.GetMetric()).(*pb.LogMetric)
+	obj.Name = name.MetricName
+	obj.CreateTime = timestamppb.New(now)
+	obj.UpdateTime = timestamppb.New(now)
+
+	// Not clear why, but the response from this method does not have the default MetricDescriptor Name/Type populated.
+	retval := proto.Clone(obj).(*pb.LogMetric)
```
acpana, line 71: question: here and and in UpdateLogMetric

does it make sense to guard the casting? or is the invariant that we know that kind of object we gave it so it must be the same kind?

acpana:
/hold for author to close on items (also because I am coming in late on this PR)
/lgtm
/approve

thanks justin for putting this together 💯

hack/compare-mock
acpana, line 1: nice thanks for adding those I was going to ask about them!

it's odd that `make fmt` doesn't complain about the missing header here

acpana:
fyi I've noticed that the "pause" tests flake (with what seems to be a race with the informer cache at first glance) and will investigate;

thanks for rebasing your changes in the meantime

/lgtm
/approve

PR #1210, opened 2024-02-08
mockgcp: Refactor errors from storage.Get

justinsb:
Going through one method at a time, but the code is much simplified by
having storage.Get return an object that follows the GRPC design
patterns, where codes.NotFound is returned when the object is not
found.

### DIFF
--- a/mockgcp/pkg/storage/memory.go
+++ b/mockgcp/pkg/storage/memory.go
@@ -19,6 +19,8 @@ import (
 	"strings"
 	"sync"
 
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/reflect/protoreflect"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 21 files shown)


mockgcp/pkg/storage/memory.go
```diff
@@ -125,7 +131,7 @@ func (s *typeStorage) Get(ctx context.Context, fqn string, dest proto.Message) e
 
 	existing, found := s.byKey[fqn]
 	if !found {
-		return apierrors.NewNotFound(schema.GroupResource{}, fqn)
```
cheftako, line 128: I like concentrating these concerns in the storage layer. However I think we still have 1 of the issues I originally brought up here. Why return an explicit code.NotFound error for Get() when the key is not found but not for Update() or Delete() when the key is not found (lines 98 and 111). Why not codes.AlreadyExists for Create() at line 74?

justinsb: Ah, because I want to save tasty PR treats for tomorrow/Monday :-)  But now you've gone and ruined the surprise...

justinsb:
Merged in #1221 (thanks!), closing

/close

PR #1214, opened 2024-02-08
mockgcp: Support for MonitoringDashboard

justinsb:
This one is a bit complicated in terms of structure, but relatively
simple in that it doesn't use LROs etc.


justinsb:
/assign @gemmahou 

Another mock :-)

gemmahou:
Looks good to me as the presubmit tests passed
/lgtm
/approve

PR #1215, opened 2024-02-08
change ComputeForwardingRule subnetworkRef to use self_link

hkundag:
### Change description

The sample added in this PR will fail when run with the current controller version with the following error:

```
Invalid value for subnetwork: Cannot determine region: set in this resource, or set provider-level 'region' or 'zone'.
```

This is because KCC currently passes the subnetwork `name` to Terraform, with no region information. This is not a problem for regional forwarding rules, since Terraform can use the forwarding rule’s region to infer the subnetwork’s region. However, since global forwarding rules don’t have a region, there is no backup value so we encounter this error.

The fix is to pass the subnetwork’s `self_link` instead, since that additionally includes the subnetwork’s region. This change should be invisible to current users since, for regional forwarding rules, Terraform will still pass the same subnetwork value in its API calls (it will just get the region info in a different way).

### Tests you have done

```
go test -v -timeout 600s -tags=integration ./config/tests/samples/create/ \
  -run-tests global-internal-forwarding-rule-with-target-http-proxy
```

### DIFF
--- a/config/samples/resources/computeforwardingrule/global-internal-forwarding-rule-with-target-http-proxy/compute_v1beta1_computeaddress.yaml
+++ b/config/samples/resources/computeforwardingrule/global-internal-forwarding-rule-with-target-http-proxy/compute_v1beta1_computeaddress.yaml
@@ -1,0 +1,25 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: compute.cnrm.cloud.google.com/v1beta1
+kind: ComputeAddress
+metadata:
+  name: computeforwardingrule-dep-global-internal-http-proxy
+spec:
+  addressType: INTERNAL
+  location: us-central1
+  ipVersion: IPV4
+  purpose: SHARED_LOADBALANCER_VIP
+  subnetworkRef:
+    name: computeforwardingrule-dep-global-internal-http-proxy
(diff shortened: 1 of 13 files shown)


hkundag:
the mock-tests testcase that failed was TestAllInSeries/fixtures/certificatemanagercertificatemapentry - seems unrelated.

config/samples/resources/computeforwardingrule/global-internal-forwarding-rule-with-target-http-proxy/compute_v1beta1_computeaddress.yaml
```diff
@@ -0,0 +1,25 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: compute.cnrm.cloud.google.com/v1beta1
+kind: ComputeAddress
+metadata:
+  name: computeforwardingrule-dep-global-internal-app
```
jingyih, line 18: Should this be `computeforwardingrule-dep-global-internal-forwarding-rule-with-target-http-proxy`? Same question for the other sample files in this PR.

hkundag:
> Another question.
> 
> PR description mentioned that you've tested
> 
> ```
> go test -v -timeout 600s -tags=integration ./config/tests/samples/create/ \
>   -run-tests global-forwarding-rule-internal-with-target-http-proxy
> ```
> 
> The sample added in this PR is under `global-internal-forwarding-rule-with-target-http-proxy`, did you mean you tested
> 
> ```
> go test -v -timeout 600s -tags=integration ./config/tests/samples/create/ \
>   -run-tests global-internal-forwarding-rule-with-target-http-proxy
> ```

That's right, the folder was originally named `global-internal-forwarding-rule-with-target-http-proxy` (when I ran the test), but it got renamed to `global-internal-forwarding-rule-with-target-http-proxy`. Will update the description.

config/samples/resources/computeforwardingrule/global-internal-forwarding-rule-with-target-http-proxy/compute_v1beta1_computeaddress.yaml
hkundag: changed to `computeforwardingrule-dep-global-internal-http-proxy` to stay within the name character limit.

jingyih (commented):
/lgtm
/approve

PR #1217, opened 2024-02-08
mockgcp: support MonitoringUptimeCheckConfig

justinsb:
Introduce a fieldmask helper also.

### DIFF
--- a/mockgcp/mockmonitoring/fieldmaskutils.go
+++ b/mockgcp/mockmonitoring/fieldmaskutils.go
@@ -1,0 +1,91 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mockmonitoring
+
+import (
+	"fmt"
+	"strings"
+
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+func setField(dest proto.Message, src proto.Message, path string) error {
+	tokens := strings.Split(path, ".")
+	if len(tokens) == 0 {
+		return fmt.Errorf("path is not valid: %q", path)
+	}
+
+	val, err := getFieldValue(src.ProtoReflect(), tokens)
+	if err != nil {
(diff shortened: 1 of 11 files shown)


tests/e2e/normalize.go
```diff
@@ -129,6 +131,13 @@ func normalizeKRMObject(u *unstructured.Unstructured, project testgcp.GCPProject
 				})
 			}
 		}
+
+		switch u.GroupVersionKind() {
```
yuwenma, line 135: Do you have suggestions on how we want to maintain a custom list in the long term? Personally I like listing individual kinds in tests like this one and use generic methods in the code. But then I realize recently that it may be uneasy to discover for contributors.

mockgcp/mockmonitoring/uptimecheck.go
```diff
@@ -0,0 +1,194 @@
...
+			return nil, status.Errorf(codes.NotFound, "Config not found for check %s in project %s", name.Name, name.Project.ID)
+		}
+		return nil, err
+	}
+
+	return redactUptimeCheckConfig(obj), nil
+}
+
+func redactUptimeCheckConfig(obj *pb.UptimeCheckConfig) *pb.UptimeCheckConfig {
+	// Fields containing sensitive information like authentication tokens or contact info are only partially populated on retrieval.
```
yuwenma, line 57: ah!

mockgcp/mockmonitoring/fieldmaskutils.go
```diff
@@ -0,0 +1,91 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+func setField(dest proto.Message, src proto.Message, path string) error {
```
yuwenma, line 25: oh, I need to borrow this. Thank you!!

yuwenma (commented):
/lgtm
/approve

This looks great!
/hold I have a question but no blockers.

tests/e2e/normalize.go
justinsb: Interesting question.  One idea is that we could just replace spec.resourceID with `"${" + Kind + "ResourceID}"` - that would probably give good results.  It's hard to know while we have only have one example here though.

tests/e2e/unified_test.go
```diff
@@ -349,6 +349,8 @@ func runScenario(ctx context.Context, t *testing.T, testPause bool, fixture reso
 								pathIDs[id] = "${alertPolicyID}"
 							case "conditions":
 								pathIDs[id] = "${conditionID}"
+							case "uptimeCheckConfigs":
+								pathIDs[id] = "${uptimeCheckConfigId}"
```
justinsb, line 353: We could use the same trick here though (dropping the s, adding ID)

justinsb:
I'll send another PR to try the generic rule for unified_test (where we have multiple examples), and we can go from there...

/hold cancel

PR #1221, opened 2024-02-09
mockgcp: Refactor errors from storage.Update and storage.Delete also

justinsb:
- mockgcp: Refactor errors from storage.Get
- mockgcp: Refactor errors from storage.Update
- mockgcp: Refactor errors from storage.Delete

Builds on #1210, I broke it up to be easier to review.


cheftako:
/lgtm
/approve
This is clearly an improvement over where we are. Still curious about the Create and the already exists case.

PR #1222, opened 2024-02-09
tests: use full path in samples

justinsb:
Otherwise we were skipping tests with non-unique names.


justinsb:
So I think we were actually previously skipping a test that was broken on mockgcp, #1228 adds the functionality.

ldanielmadariaga:

PR #1223, opened 2024-02-09
Refactor golden test

gemmahou:
### Change description
Refactor golden test

Goals:
1. Use helper function.
    I assume we will introduce more format rules in the future to support different resources, I recommend that we relocate them to a helper function. By doing so, we can ensure that our primary test function remains manageable, avoiding unnecessary complexity.
2. Prevent from overriding the golden file
    After the golden file is generated, we might want to prevent external contributors from accidentally overriding it. To achieve this, I've adjusted the logic a bit:
a) If the golden file exists, perform a comparison regardless of whether the "WRITE_GOLDEN_OUTPUT" variable is specified.
b) If the golden file does not exist and the "WRITE_GOLDEN_OUTPUT" variable is specified, write the golden file.
c) If the golden file does not exist and the "WRITE_GOLDEN_OUTPUT" variable is not specified, log an error indicating that the variable must be provided to write golden file.
People can still update the golden file by deleting it and rewriting it.
3. Add more comments

### Tests you have done

### DIFF
--- a/config/tests/samples/create/harness.go
+++ b/config/tests/samples/create/harness.go
@@ -484,45 +484,33 @@ func (t *Harness) waitForCRDReady(obj client.Object) {
 	}
 }
 
-func (h *Harness) CompareGoldenFile(p string, got string, normalizers ...func(s string) string) {
-	if os.Getenv("WRITE_GOLDEN_OUTPUT") != "" {
-		// Short-circuit when the output is correct
-		b, err := os.ReadFile(p)
-		if err == nil {
-			want := string(b)
-			for _, normalizer := range normalizers {
-				got = normalizer(got)
-				want = normalizer(want)
-			}
-			if want == got {
-				return
-			}
-		}
-
-		if err := os.WriteFile(p, []byte(got), 0644); err != nil {
-			h.Fatalf("failed to write golden output %s: %v", p, err)
-		}
-		h.Errorf("wrote output to %s", p)
-	} else {
-		want := string(h.MustReadFile(p))
-
-		for _, normalizer := range normalizers {
-			got = normalizer(got)
-			want = normalizer(want)
-		}
-
+func (h *Harness) CompareGoldenFile(p string, got string) {
+	b, err := os.ReadFile(p)
+	if err == nil {
+		h.Logf("golden file %s exist, comparing with it", p)
+		want := string(b)
 		if diff := cmp.Diff(want, got); diff != "" {
 			h.Errorf("unexpected diff in %s: %s", p, diff)
 		}
+	} else if os.IsNotExist(err) {
+		if os.Getenv("WRITE_GOLDEN_OUTPUT") != "" {
+			// golden http log does not exist, write golden file
+			if err := os.WriteFile(p, []byte(got), 0644); err != nil {
+				h.Fatalf("failed to write http output %s: %v", p, err)
+			}
+			h.Logf("wrote output to %s", p)
+		} else {
+			h.Fatalf("golden file does not exist, please specify WRITE_GOLDEN_OUTPUT")
+		}
+	} else {
+		h.Fatalf("error reading golden file(%q): %v", p, err)
 	}
 }
 
-func (h *Harness) MustReadFile(p string) []byte {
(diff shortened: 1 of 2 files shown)


config/tests/samples/create/harness.go
```diff
@@ -484,45 +484,33 @@ func (t *Harness) waitForCRDReady(obj client.Object) {
...
-			want = normalizer(want)
-		}
-
+func (h *Harness) CompareGoldenFile(p string, got string) {
+	b, err := os.ReadFile(p)
+	if err == nil {
+		h.Logf("golden file %s exist, comparing with it", p)
+		want := string(b)
 		if diff := cmp.Diff(want, got); diff != "" {
 			h.Errorf("unexpected diff in %s: %s", p, diff)
```
justinsb, line 493: We do want to write the golden file here also

config/tests/samples/create/harness.go
```diff
@@ -484,45 +484,33 @@ func (t *Harness) waitForCRDReady(obj client.Object) {
...
 		}
+	} else if os.IsNotExist(err) {
+		if os.Getenv("WRITE_GOLDEN_OUTPUT") != "" {
+			// golden http log does not exist, write golden file
+			if err := os.WriteFile(p, []byte(got), 0644); err != nil {
+				h.Fatalf("failed to write http output %s: %v", p, err)
+			}
+			h.Logf("wrote output to %s", p)
```
justinsb, line 501: I personally have found that I normally want this to be an error, because the tests didn't technically pass.  We can try it this way for a bit though if you prefer

config/tests/samples/create/harness.go
```diff
@@ -484,45 +484,33 @@ func (t *Harness) waitForCRDReady(obj client.Object) {
...
+	} else {
+		h.Fatalf("error reading golden file(%q): %v", p, err)
 	}
 }
 
-func (h *Harness) MustReadFile(p string) []byte {
-	b, err := os.ReadFile(p)
-	if err != nil {
-		h.Fatalf("error from ReadFile(%q): %v", p, err)
+func (h *Harness) NormalizeLog(input string, normalizers ...func(s string) string) {
```
justinsb, line 510: Nit: s/NormalizeLog/Normalize/g - because it isn't specific to logs any more ?  Unless you're planning another change to this method

tests/e2e/unified_test.go
```diff
@@ -155,66 +155,34 @@ func TestAllInSeries(t *testing.T) {
 					}
 
 					expectedPath := filepath.Join(fixture.SourceDir, "export.yaml")
-					output := h.MustReadFile(outputPath)
-					h.CompareGoldenFile(expectedPath, string(output), h.IgnoreComments, h.ReplaceString(project.ProjectID, "example-project-id"))
+					output, err := os.ReadFile(outputPath)
```
justinsb, line 158: I think this is what MustReadFile did ... any reason to remove it?

config/tests/samples/create/harness.go
gemmahou: Just my personal preference: people may not 100% understand the command they use and simply copy&paste. My idea was once we have a golden file exists, we do not want users to override it. While, if we feel we need to update the golden file, we can always remove it and regenerate. Maybe I can make the log message more clear: `golden file exists, comparing with it. If you want to update the golden file, please remove the old one and regenerate it by specifying "WRITE_GOLDEN_OUTPUT"`

config/tests/samples/create/harness.go
gemmahou: Could you explain when you say `the tests didn't technically pass`?  I feel we can consider test as pass because: 1) The CRUD function does not return errors 2) the http log successfully written to file. I was confused when I got "FAIL" and the only error message "wrote  output to file" did not sound like an error 😅

tests/e2e/unified_test.go
gemmahou: No big difference, but I do not think it's necessary to turn that into a new function? Handle the error after `ReadFile` is called seems a normal pattern and more straightforward.

gemmahou:
Closing. This will be addressed in b/332777103

PR #1224, opened 2024-02-09
typed client generation: parameterize version

justinsb:
We were previously assuming everything was v1beta1.


maqiuyujoyce (approved):
Good catch! Thank you for the fix!

PR #1228, opened 2024-02-10
mockgcp: Support workflows

justinsb:
Some actions are cross-service, for example a "default" network is created in new projects.


justinsb:
FYI @anhdle-sso I _think_ this will unblock the project-policy flakes (the existence of the network _seems_ to be the non-determinism there)

PR #1230, opened 2024-02-13
mockgcp: Support for bigquery DataSet

justinsb:
Initial support for BigQueryDataSet in mockgcp


mockgcp/common/httpmux/filterbody.go
```diff
@@ -31,9 +31,6 @@ func FilterBodyOn204(inner http.Handler) (http.Handler, error) {
 		if brw.statusCode == 204 && brw.body.String() == "{}" {
 			brw.statusCode = 204
 			brw.body.Reset()
-
-			// May not belong here, TBD what other services do...
-			brw.Header().Set("Content-Type", "application/json")
```
cheftako, line 36: Doesn't matter now but in general I think header operations should be before body operations.

PR #1235, opened 2024-02-13
tests: Fix loading samples to avoid mixing samples with the same name

justinsb:
Otherwise foo/test and bar/test both get mixed into "test"


jingyih (commented):
/lgtm

justinsb:
Test failures fixed by #1236

justinsb:
(Maybe we should enable the merge queue, as that is github's built-in solution for PRs that "cross")

PR #1236, opened 2024-02-13
mockgcp: Don't wrap errors in the storage layer

justinsb:
It is unnecessary now that we return "grpc ready" errors from the
storage layer, and may also break HTTP status codes (e.g. 404 on not
found).

### DIFF
--- a/mockgcp/mockcompute/disksv1.go
+++ b/mockgcp/mockcompute/disksv1.go
@@ -22,7 +22,6 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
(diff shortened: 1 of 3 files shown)


justinsb:
/assign @gemmahou 

(I think our tests are broken without this, if they're not ... we may have bigger problems!)

gemmahou:
> /assign @gemmahou
> 
> (I think our tests are broken without this, if they're not ... we may have bigger problems!)

I ran the test locally. storage.Get and storage.Delete are broken without the change, but seems like storage.Create and storage.Update are not causing the test to fail. 

Update: After checking #1210, and mockgcp/pkg/storage/memory.go, I understand that `apierrors.IsNotFound` won't throw, instead we log an error message. Create and Update will not fail, but error messages will be redundant with out the code change.

PR #1241, opened 2024-02-14
tests: Filter log messages

justinsb:
This avoids a lot of logspam and lets us better see what is actually
happening.

### DIFF
--- a/config/tests/samples/create/harness.go
+++ b/config/tests/samples/create/harness.go
@@ -24,6 +24,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


gemmahou:
Thanks for the code change! It's very useful for debugging and identifying important information in the log. Do you think it's possible to also filter out "Registered controller" as well? https://screenshot.googleplex.com/3wU2ouGbMFBbFed.png

acpana:
I can buy log filtering but it may confuse people that don't know it's happening. If we run into that down the road, it may make sense to make this configurable so folks can choose which log lines get filtered, etc.

/lgtm
/approve

thanks for working on this 💯

PR #1242, opened 2024-02-14
Release notes for 1.113.0

justinsb:
I figure we can stage these in our repo, and also collaboratively build them before the actual release (e.g. in PRs)

### DIFF
--- a/docs/releasenotes/release-1.113.0.md
+++ b/docs/releasenotes/release-1.113.0.md
@@ -1,0 +1,24 @@
+* Initial support for status.observedState in ContainerCluster,
+  ContainerNodePool and RedisInstance; we are trying to encourage use of
+  `cnrm.cloud.google.com/state-into-spec: absent` and eventually
+  make it the default.  Some important resource information (such as the
+  certificate for connecting to a GKE cluster) is only currently available
+  in spec, and we recommend instead reading it from observedState
+  where this is available.  We expect to add more fields to observedState
+  in the future.
+
+* Isolate terraform provider into a git subtree so we can more directly fix
+  problems.
+
+* Special shout-outs to svwijk@, katrielt@, sofam@, higef@ for their
+  contributions to this release.
+
+## New Resources:
+
+* Added support for `ComputeNetworkFirewallPolicy` (v1beta1) resource.
+* Added support for `TagsLocationTagBinding` (v1alpha1) resource.
+
+## New Fields:
+
+* RunJob (CloudRun Job)
+  * Added `spec.template.vpcAccess.connectorRef` field.


docs/releasenotes/release-1.113.0.md
```diff
@@ -0,0 +1,25 @@
...
+  contributions to this release.
+
+## New Resources:
+
+* ComputeNetworkFirewallPolicy
+* TagsLocationTagBinding
+
+## New Fields:
+
+* connectorRef in RunJob (CloudRun Job)
```
jingyih, line 25: I forgot about the process, are we generating the notes by manually going over all the new commits?

https://github.com/GoogleCloudPlatform/k8s-config-connector/compare/v1.112.0...v1.113.0

`TagsLocationTagBinding` is an alpha resource. We could probably just following the existing release note format to avoid confusion?

```
* Added support for `ComputeNetworkFirewallPolicy` (v1beta1) resource.
* Added support for `TagsLocationTagBinding` (v1alpha1) resource.
```

justinsb: Used the language - thanks!   Were you thinking of _not_ splitting out the New Resources / New Fields headings.  I introduced this, thinking that this helps people read the notes better - do you think we should try this structure or go with a flat structure?

gemmahou:
What is the primary difference between this release note and the ones that were officially published(https://github.com/GoogleCloudPlatform/k8s-config-connector/releases)? Was the intention of this release note to provide more internal information, such as special thanks and additional details that we prefer not to share in the official release note?

justinsb:
> What is the primary difference between this release note and the ones that were officially published(https://github.com/GoogleCloudPlatform/k8s-config-connector/releases)?

This is the release note that _will_ be published for 1.113.0 on that page.  But by doing this in a PR, we can collaborate and it will also live in our code repo e.g. if we wanted to reuse this on a github pages website or similar.

docs/releasenotes/release-1.113.0.md
jingyih: Splitting out the New Resources / New Fields is more clear to me. 

I could be wrong, but I guess part of the old format came from running the script `scripts/generate-field-diffs-tf-upgrade/generate_field_changes.py` when updating Terraform prior to v4.84.0. e.g. https://cnrm-review.git.corp.google.com/c/cnrm/+/65900

gemmahou: nit: For consistency, I would suggest using the same format for new fields.

Resource RunJob(v1beta1):
    Added spec.template.template.vpcAccess.connectorRef field.

jingyih (commented):
/lgtm
/approve

PR #1244, opened 2024-02-15
tests: use golden testing framework for hcl tests

justinsb:
Our existing testing was not very strict.

### DIFF
--- a/config/tests/samples/create/harness.go
+++ b/config/tests/samples/create/harness.go
@@ -25,7 +25,6 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
-	"github.com/google/go-cmp/cmp"
 	transport_tpg "github.com/hashicorp/terraform-provider-google-beta/google-beta/transport"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
@@ -497,44 +496,11 @@ func (h *Harness) waitForCRDReady(obj client.Object) {
 }
 
 func (h *Harness) CompareGoldenFile(p string, got string, normalizers ...func(s string) string) {
-	if os.Getenv("WRITE_GOLDEN_OUTPUT") != "" {
-		// Short-circuit when the output is correct
-		b, err := os.ReadFile(p)
-		if err == nil {
-			want := string(b)
-			for _, normalizer := range normalizers {
-				got = normalizer(got)
-				want = normalizer(want)
-			}
-			if want == got {
-				return
-			}
-		}
-
-		if err := os.WriteFile(p, []byte(got), 0644); err != nil {
-			h.Fatalf("failed to write golden output %s: %v", p, err)
-		}
-		h.Errorf("wrote output to %s", p)
-	} else {
-		want := string(h.MustReadFile(p))
-
-		for _, normalizer := range normalizers {
-			got = normalizer(got)
-			want = normalizer(want)
-		}
-
-		if diff := cmp.Diff(want, got); diff != "" {
-			h.Errorf("unexpected diff in %s: %s", p, diff)
-		}
-	}
+	test.CompareGoldenFile(h.T, p, got, normalizers...)
 }
 
 func (h *Harness) MustReadFile(p string) []byte {
-	b, err := os.ReadFile(p)
-	if err != nil {
-		h.Fatalf("error from ReadFile(%q): %v", p, err)
-	}
-	return b
+	return test.MustReadFile(h.T, p)
 }
 
 // IgnoreComments is a normalization function that strips comments.
--- a/pkg/test/utils.go
+++ b/pkg/test/utils.go
@@ -18,7 +18,10 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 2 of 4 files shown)


justinsb:
You were right @gemmahou  - the golden file method needed a refactor.  I _still_ don't think we're done here, because it would be nice if we added the copyright automatically when we write the golden file.   But this came up for me because of the hcl v1 dependency...

/assign @gemmahou

config/tests/samples/create/harness.go
```diff
@@ -491,44 +490,11 @@ func (t *Harness) waitForCRDReady(obj client.Object) {
...
-		for _, normalizer := range normalizers {
-			got = normalizer(got)
-			want = normalizer(want)
-		}
-
-		if diff := cmp.Diff(want, got); diff != "" {
-			h.Errorf("unexpected diff in %s: %s", p, diff)
-		}
-	}
+	test.CompareGoldenFile(h.T, p, got, normalizers...)
```
gemmahou, line 493: Hmmm can we remove CompareGoldenFile and MustReadFile from harness? I do not see any reason we need them here, we can call the function via test utils.go

pkg/tf/serialization/testdata/cluster.golden.tf
```diff
@@ -94,10 +94,6 @@ resource "google_container_cluster" "foo" {
   project         = "my-project"
   release_channel = {}
 
-  resource_labels = {
```
gemmahou, line 97: why removing?

pkg/test/utils.go
```diff
@@ -60,3 +63,78 @@ func MustReadFile(t *testing.T, p string) []byte {
...
+
+	wantBytes, err := os.ReadFile(p)
+	if err != nil {
+		if writeGoldenOutput && os.IsNotExist(err) {
+			// expected when creating output for the first time
+		} else {
+			t.Fatalf("failed to read golden file %q: %v", p, err)
+		}
+	}
+	want := string(wantBytes)
```
gemmahou, line 83: I might not familiar with golang string cast and I do not test the code locally. I'm just curious if `wantBytes` is nil(assuming golden file does not exist for the first time), is string(nil) a valid expression and what's the return value?

pkg/tf/serialization/tf_resource_test.go
```diff
@@ -34,46 +31,21 @@ func TestSerializeCluster(t *testing.T) {
...
 }
 
-func testSerialize(t *testing.T, instanceStateFile, tfType, goldenFile string) string {
+func testSerialize(t *testing.T, instanceStateFile, tfType, goldenFile string) {
 	provider := tfprovider.NewOrLogFatal(tfprovider.UnitTestConfig())
-	b, err := ioutil.ReadFile(instanceStateFile)
-	if err != nil {
-		t.Fatalf("failed to load instance state for instance, %s", err.Error())
-	}
+	b := test.MustReadFile(t, instanceStateFile)
```
gemmahou, line 36: I'm unsure if turn error handling to an util function(MustReadFile) is best practice. My thought is handling errors after calling ReadFile is more straightforward. This also allows us to customize the error message to include more details, i.e. specific files being read, which is useful for debugging purposes.

config/tests/samples/create/harness.go
justinsb: I think it's better to keep it from the developer ergonomics point of view - easier discovery and one less argument to pass.

pkg/test/utils.go
justinsb: https://go.dev/play/p/UJppOPhVWw5

pkg/tf/serialization/testdata/cluster.golden.tf
justinsb: So I believe we aren't removing them, they were never generated.  Our golden test was way too flexible..

I'll send a separate PR to split it out!

pkg/tf/serialization/tf_resource_test.go
justinsb: This only works in tests because of t.Fatalf being allowed in tests.  I don't know what we would print that isn't "error reading file %q: %v" though?  The contents of the file don't really matter IMO

pkg/test/utils.go
gemmahou: Cool, and the playground tool is awesome!

gemmahou:
Need to rebase and resolve merge conflict. 

/lgtm

pkg/test/utils.go
gemmahou: Hmm correct me if I'm wrong, in your example, var b is not nil, it's an empty variable right? https://go.dev/play/p/DnUOhQ2aAuN

justinsb: But ... given revive is complaining about it also, I might as well make it more self-documenting :-)

gemmahou: Thanks for the explanation! This might be where Golang is different from other languages. LGTM!

gemmahou (approved):

PR #1256, opened 2024-02-16
ci: add errorlint linter

acpana:
Branching off from https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1186

Commits mostly cherry-picked for errorlint with some fixup.

This is mostly replacing a %v to %w directive where supported and getting around that where not.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -22,7 +22,7 @@ linters:
   disable-all: true
   enable:
     - errcheck
-    # - errorlint
+    - errorlint
     # - exportloopref
     # - forcetypeassert
     # - gci
--- a/cmd/recorder/main.go
+++ b/cmd/recorder/main.go
@@ -163,7 +163,7 @@ func forEach(c client.Client, gvk schema.GroupVersionKind, listOptions *client.L
 		list.SetGroupVersionKind(gvk)
 		err := c.List(context.Background(), &list, listOptions)
 		if err != nil {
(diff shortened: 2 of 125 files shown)


justinsb:
LGTM - thanks for breaking this down :-)

I reviewed with `gh pr checkout 1256 && git diff --word-diff upstream/master`.  It helped that I didn't have any quibbles :-)

`go run github.com/polyfloyd/go-errorlint@v1.4.8 -errorf -fix ./...` shows there are still some more %v, but I think that can be done in another PR.  More concerning I guess is that the linter is passing, though I thought golang-ci was supposed to be a wrapper over go-errorlint.  They are in the scripts & operator directory, maybe these are excluded somewhere.

We might want to include go-errorlint in our linting, not least because of that magic `-fix` command.  I think it would be very cool if we had an "autofix" where possible command - in general we have a lot of commands that are of this type, where we can run the tool and if there are any diffs then we consider that a failure.  Our golden tests are an example of this.  The really nice thing about this is that by running the tool locally, we can then just review and apply the changes (and `git-autofixup` is magical!).  This also leads to another thing I like, where we make sure that our CI can all be run locally, so github actions just invokes scripts.  I personally like e.g. `dev/ci/lint` that invokes `dev/lint-go`, but the downside of _that_ is that we tend to duplicate the Makefile.

In any case...

/approve
/lgtm

PR #1257, opened 2024-02-16
chore:lint: support revive part1

acpana:
branching off from https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1186

Commits mostly cherry-picked for revive with some fixup. This is only part 1; there's another commit that will come as a separate PR.

- machine replacements for *Id* -> *ID*, *json* -> *JSON*
- some stuttering removal: eg. TestMainFoo to Foo
- some more redundant type definitions were taken out
- some easy unused variables and extra indentations and return nils

All of this was revive suggested.


operator/tests/e2e/e2e_test.go
```diff
@@ -1755,7 +1751,7 @@ func getAnnotationsForNS(ns *v1.Namespace) map[string]string {
 	return annotations
 }
 
-func newUniqueTestId() string {
+func newUniquetestID() string {
```
justinsb, line 1754: Nit: Probably newUniqueTestID ... I will continue to try to nerd-snipe you into writing a linter that can autofix for this rule :-)

justinsb:
Thanks - you snuck in under the github size limit for number of files in a PR before they start being collapsed by default :-)

/approve
/lgtm

operator/tests/e2e/e2e_test.go
acpana: 🤦🏼 actually the "t" should be capitalized; will fix forward in the next PR

PR #1264, opened 2024-02-20
chore:lint: second round of revive

acpana:
Last PR to be branched off from https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1186 . Specifically, this is part two from https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1257 .

- mostly if blocks, unused vars
- some light refactors

All of modifications were revive suggested.

Also, this turns on the revive linter


operator/pkg/preflight/upgrade_test.go
```diff
@@ -29,15 +29,10 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes/scheme"
-	"k8s.io/client-go/rest"
 	"sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/addon/pkg/loaders"
 	"sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/declarative"
 )
 
-var (
-	cfg *rest.Config
```
justinsb, line 38: Nice - one less global variable :tada:

pkg/controller/iam/partialpolicy/bindings_test.go
```diff
@@ -28,7 +28,7 @@ import (
 // mockIdentityResolver helps to resolve referenced member identity
 type mockIdentityResolver struct{}
 
-func (t *mockIdentityResolver) Resolve(member iamv1beta1.Member, memberFrom *iamv1beta1.MemberSource, defaultNamespace string) (string, error) {
+func (t *mockIdentityResolver) Resolve(member iamv1beta1.Member, memberFrom *iamv1beta1.MemberSource, _ string) (string, error) {
```
justinsb, line 31: This is an example of where it's tricky to drop the parameter names.  I can guess that `_ context.Context` is a ctx.  But I can't guess what `_ string` is.  And because Go interfaces are so implicit, it's as trivial to figure it out as it would be in e.g. Java...

pkg/resourceoverrides/overrides.go
```diff
@@ -70,19 +70,19 @@ type ResourceOverrides struct {
 	Overrides []ResourceOverride
 }
 
-type ResourceOverridesHandler struct {
+type ROHandler struct {
```
justinsb, line 73: Handler was not available? :-)

pkg/webhook/register.go
```diff
@@ -39,7 +39,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
-	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
```
justinsb, line 42: FYI this allows for authentication using GCP, it isn't safe to remove this in general.  However, in this case yeah shouldn't be needed!

justinsb:
A few nits but nothing blocking

/approve
/lgtm

pkg/resourceoverrides/overrides.go
acpana: Yeah there's a Handler variable here:

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/09c6b365bb1e416e555f86f27011047e449e70dc/pkg/resourceoverrides/overrides.go#L83

It was either this or changing all the references to that variables and since this struct was only used in this file I decided to go with renaming it here.

pkg/controller/iam/partialpolicy/bindings_test.go
acpana: yeah agreed on interface implementations for `_ string ` types it's hard to know what the parameters would be.

I had reverted some here:

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/09c6b365bb1e416e555f86f27011047e449e70dc/operator/pkg/preflight/upgrade_test.go#L43-L59

But missed these. I'll create an issue for the handful of these; 

```
grep -r --include "*.go" --exclude-dir="third_party" --exclude-dir=".build" -nRH "_ string" .
```

PR #1265, opened 2024-02-21
chore:lint: enable go-vet linter

acpana:
Enables go-vet for golang-ci.

This should be free for us since we already run vet as part of make test:

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/09c6b365bb1e416e555f86f27011047e449e70dc/Makefile#L42

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -34,7 +34,7 @@ linters:
     # - goimports
     # - gosec
     # - gosimple
-    # - govet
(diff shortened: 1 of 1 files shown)


justinsb:
Thanks - maybe we can pull it out of some of our e2e test targets also - I think right now our integration tests run e.g. the fmt test!

/approve
/lgtm

PR #1266, opened 2024-02-21
Support state-into-spec user override in ConfigConnectorContext

maqiuyujoyce:
### Change description

Fixes b/320784855

* Added the optional `spec.stateIntoSpec` field in ConfigConnectorContext for namespace-level default value override.
* Added `--state-into-spec-default-value` and `--state-into-spec-user-override` flags to the controller manager CLI to pass the default value and user override value.
* Updated the CCC controller logic to transform the manifest to pass the value specified by `spec.stateIntoSpec` to the controller manager CLI via the new flags.
* Updated the unit test to verify the transformed manifest contains the new flags.

### Tests you have done

Deployed the operator locally using the updated images, verified that
1. when `spec.stateIntoSpec` is unset in CCC, the created KCC resource has `state-into-spec: merge` by default;
2. when `spec.stateIntoSpec` is set to `absent` in CCC, the create KCC resource has `state-into-spec:absent` by default;
3. when `spec.stateIntoSpec` is set to `invalid` in CCC, apply the CCC CR returns an error:
   ```
   The ConfigConnectorContext "configconnectorcontext.core.cnrm.cloud.google.com" is invalid: spec.stateIntoSpec: Unsupported value: "invalid": supported values: "absent", "merge"
   ```
Details can be found in #1267 .


cmd/manager/main.go
```diff
@@ -63,6 +65,9 @@ func main() {
 	flag.StringVar(&scopedNamespace, "scoped-namespace", "", "scope controllers to only watch resources in the specified namespace; if unspecified, controllers will run in cluster scope")
 	flag.BoolVar(&enablePprof, "enable-pprof", false, "Enable the pprof server.")
 	flag.IntVar(&pprofPort, "pprof-port", 6060, "The port that the pprof server binds to if enabled.")
+	flag.StringVar(&stateIntoSpecDefaultValue, "state-into-spec-default-value", k8s.StateIntoSpecDefaultValueV1Beta1, "default value for the cnrm.cloud.google.com/state-into-spec annotation if the annotation is unset in a resource; if --state-into-spec-user-override is set, then the user override is used instead")
```
justinsb, line 68: I think we want to read the CCC directly.  @acpana will be setting us up for this with his "pause" code.

maqiuyujoyce: Is @acpana 's change a blocker to this PR? If the answer to question 2 (see below) is yes, then I guess it is not a blocker?

A couple of design related questions regarding the dynamic reading of CCC:
1. If the only usage of the fields (e.g. the field used to override state-into-spec) in CCC is to be read and used by the controller manager (but not CCC controller itself), then they seem to be more of environment variables. Should they be put under the `spec.env` field to make the purpose more clear?
2. Will we change the behavior of existing fields (e.g. billingProject, and etc)? Making the controller adopt live changes of an existing field sounds like a backwards-compatible behavior. But the higher flexibility also makes it easier to make a mistake because we can't do validation for all such fields (e.g. billingProject). Is it something we're okay with?
3. Using the flags in the CLI makes it easier for us to test core KCC and the operator separately. If the controller manager is reading the live state of the CCC, have we designed the test framework/test cases to validate if this will work as expected e2e?

Also, reading the live value looks like a new "feature" for KCC operator, I assume we'll update the public doc to reflect it?

justinsb: I don't know whether it's a blocker, I think we should figure out what this code looks like if we assume we are reading the CCC in the reconciler anyway.  Maybe a separate WIP PR so we can take a look, branched from this one?

For the questions:
1) I'll take a look, I don't intuitively think it should be nested under env, but ... I'll take a look
2) Not immediately, but maybe in future?  The good news is that they are much less likely to change.  If we want to run "multi-tenant" where one KCC pod serves multiple CCCs, we would have to.  But I don't know if we want to do that.
3) We discussed, probably we want to create the CCC or change the defaulter.

And yes, we should document this, though I think it's an implementation detail rather than something we expect users will need to be particularly mindful of (?).

operator/pkg/apis/core/v1beta1/configconnectorcontext_types.go
```diff
@@ -42,6 +42,18 @@ type ConfigConnectorContextSpec struct {
...
+	// cnrm.cloud.google.com/state-into-spec annotation if the annotation is
+	// unset in a resource.
+	// 'absent' means unspecified fields in the resource spec stay unspecified
+	// after the successful reconciliations.
+	// 'merge' means the unspecified fields in the resource spec are populated
+	// after the successful reconciliations if the fields are computed/defaulted
+	// by the API.
+	//+kubebuilder:validation:Enum=absent;merge
+	//+kubebuilder:validation:Optional
+	StateIntoSpec *string `json:"stateIntoSpec,omitempty"`
```
justinsb, line 56: This looks right, it's important that it's optional so :+1: .  Name ("stateIntoSpec") seems logical and fine, not sure if anyone else on the team has a better suggestion but barring that no objections to the name!

cmd/manager/main.go
acpana: still working on a PoC for reading CCC in the controller mgr; will update when ready

but if I am reading this thread right, this work shouldn't be blocked on it

operator/pkg/apis/core/v1beta1/configconnectorcontext_types.go
acpana: Would it be better to call this `StateIntoSpecOverride` ? I know it stutters a bit but I still find the `*IntoSpec` confusing as a newish comer here.

maqiuyujoyce: I'm a bit hesitant to call it `StateIntoSpecOverride` because it sounds like this value will override the resource-level `state-into-spec` configuration.

The override hierarchy is:
resource-level `state-into-spec` annotation > 
CCC/namespace-level `stateIntoSpec` configuration > 
CCC/cluster-level `stateIntoSpec` configuration.

I don't know if I have any other good ideas so I'm calling for some more feedback @jingyih @gemmahou !

jingyih: I think we probably don't want yet another level of `state-into-spec` per resource Kind. /cc @justinsb

maqiuyujoyce:
Replaced by #1280 .

PR #1270, opened 2024-02-21
tests: remove extra lines from our golden cluster test

justinsb:
This golden test was ignoring extra lines in the golden output.  We
want to move to a stricter comparison model; start by removing these
lines so that it can be clearer.


justinsb:
/assign @gemmahou 

Our golden tests seem to ignore extra lines

gemmahou (approved):

PR #1271, opened 2024-02-21
Promote VertexAIDataset v1alpha1 to v1beta1

jingyih:
### Change description

Fixes b/325303120

### Tests you have done

```
go test -timeout 1200s -v -tags=integration ./pkg/controller/dynamic/ -test.run TestCreateNoChangeUpdateDelete -run-tests vertexaidatasetbasic

go test -timeout 1200s -v -tags=integration ./pkg/controller/dynamic/ -test.run TestCreateNoChangeUpdateDelete -run-tests vertexaidatasetencryptionkey

go test -v -tags=integration ./config/tests/samples/create/ -run-tests vertexai-dataset-encryptionkey

go test -v -tags=integration ./config/tests/samples/create/ -run-tests vertexai-dataset-basic
```


justinsb:
Looking good ... thank you for doing the refs.

I am mindful of moving the goalposts, but a few things I think we should consider with this breaking change:

* I'm not the biggest fan of projectRef, but I guess the path forwards here would be to make projectRef optional and take the value from the namespace, so not a breaking change.
* We should try to set ourselves up for state-into-spec absent.  Are there any important fields that will be dropped with state-into-spec absent?  Are there any input/output fields in spec at all?
* Can we / should we just ignore state-into-spec merge?

Maybe we need to rethink the procedure here, like adding refs etc in alpha so that the promotion to beta is just a new version.

cc @cheftako

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_vertexaidatasets.vertexai.cnrm.cloud.google.com.yaml
```diff
@@ -67,11 +67,215 @@ spec:
                   Dataset. If set, this Dataset and all sub-resources of this Dataset
```
maqiuyujoyce, line 67: I'm curious - can `encryptionSpec` be used by itself (as an empty struct)? The reason I asked is because the field description looks a bit confusing. I want to avoid we run into a similar issue like b/325359109.

config/servicemappings/vertexai.yaml
```diff
@@ -0,0 +1,61 @@
...
+      idTemplate: "{{name}}"
+      idTemplateCanBeUsedToMatchResourceName: false
+      resourceAvailableInAssetInventory: true
+      v1alpha1ToV1beta1: true
+      storageVersion: v1alpha1
+      metadataMapping:
+        labels: labels
+      serverGeneratedIDField: name
+      resourceID:
+        targetField: name
```
maqiuyujoyce, line 38: Just to confirm, the name field doesn't have a format (e.g. "projects/[PROJECT_ID]")?

scripts/generate-google3-docs/resource-reference/templates/vertexai_vertexaidataset.tmpl
```diff
@@ -0,0 +1,54 @@
...
+<td>{{"{{gcp_name_short}}"}} Service Name</td>
+<td>Vertex AI</td>
+</tr>
+<tr>
+<td>{{"{{gcp_name_short}}"}} Service Documentation</td>
+<td><a href="/vertex-ai/docs">/vertex-ai/docs</a></td>
+</tr>
+<tr>
+<td>{{"{{gcp_name_short}}"}} REST Resource Name</td>
+<td>v1/projects.locations.datasets</td>
```
maqiuyujoyce, line 26: Nit: We probably want to use v1beta1 as KCC probably talks to their beta endpoint (it depends on the configuration in the TF lib).

maqiuyujoyce:
> * I'm not the biggest fan of projectRef, but I guess the path forwards here would be to make projectRef optional and take the value from the namespace, so not a breaking change.

@justinsb Could you share more rationale about this suggestion? `projectRef`, `folderRef`, `organizationRef`, and `billingAccountRef` are hierarchical references we added to make sure single step application is possible (folder ID, organization ID and billingAccount ID are computed values) and we support consistency among these containers. Make namespace-level project ID override a preference feels like we become less concerned about the consistency and prefer customized optimizations for different type of containers.

> * We should try to set ourselves up for state-into-spec absent. Are there any important fields that will be dropped with state-into-spec absent? Are there any input/output fields in spec at all?

@jingyih is there any field in the `spec` that will be computed by the API if unset by the user? If so, those fields are potential candidates to be added into the observed state.

> * Can we / should we just ignore state-into-spec merge?

I'm inclined to keep `state-into-spec` `merge`.

`state-into-spec` with options `merge` or `absent` is a non-resource-specific feature based on my understanding. If we want to change it to be a resource-specific feature, I'd double check with the product team to confirm this is a backwards compatible feature change.

Besides it'll make the defaulting/user-override logic more complicated / inconsistent to users: Should they expect we silently default the behavior of `state-into-spec` into `absent` for this resource? Or should they get a warning to explain that the default behavior of `state-into-spec` is different for this resource? What if the users explicitly set `merge` to override the default behavior at the namespace-level or the cluster-level?

Maybe we should use a different thread to discuss more details how we can get closer to this direction in an agreed backwards compatible way?

> Maybe we need to rethink the procedure here, like adding refs etc in alpha so that the promotion to beta is just a new version.

This sounds to me that we'll want to invest more time to increase the quality of the v1alpha1 resources? I think this is nice-to-have, but may not be a priority at our current capacity?

maqiuyujoyce (commented):
LGTM! I only have a few questions.

config/servicemappings/vertexai.yaml
jingyih: This is what I got when playing with TF directly. It looks to me that the TF field `name` already has all the information?

```
$ terraform state show google_vertex_ai_dataset.dataset
# google_vertex_ai_dataset.dataset:
resource "google_vertex_ai_dataset" "dataset" {
    create_time         = "2024-02-16T23:56:54.194318Z"
    display_name        = "terraform"
    id                  = "projects/674741657116/locations/us-central1/datasets/5201786212473372672"
    labels              = {
        "env" = "test"
    }
    metadata_schema_uri = "gs://google-cloud-aiplatform/schema/dataset/metadata/image_1.0.0.yaml"
    name                = "projects/674741657116/locations/us-central1/datasets/5201786212473372672"
    project             = "cnrm-jingyih"
    region              = "us-central1"
    update_time         = "2024-02-16T23:57:43.624186Z"
}
```

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_vertexaidatasets.vertexai.cnrm.cloud.google.com.yaml
jingyih: https://cloud.google.com/vertex-ai/docs/reference/rest/v1/EncryptionSpec
Says this field "Represents a customer-managed encryption key spec that can be applied to a top-level resource." I don't think an empty struct is meaningful for this field.

scripts/generate-google3-docs/resource-reference/templates/vertexai_vertexaidataset.tmpl
jingyih: I think you are right but how do I confirm which API endpoint it is using?

justinsb: You could write a mock :-)

justinsb: Or you can just record the http traffic using something like:

```
WRITE_GOLDEN_OUTPUT=1 ARTIFACTS=$(pwd)/artifactz/gcp/ \
E2E_KUBE_TARGET=envtest RUN_E2E=1 E2E_GCP_TARGET=real \
   go test -test.count=1 -timeout 3600s -v ./tests/e2e -run "TestAllInSeries/${FIXTURE}" 2>&1 | tee log
```

justinsb:
> [projectRef optional] ... Could you share more rationale about this suggestion? 

Simply that it makes it harder to create a resource in "the current project".  I need to inject my `gcloud config get project` into every resource, instead of annotating the namespace once.  I guess I could create a project in the namespace, and I don't think making projectRef optional would be blocking, so I don't think this is a blocker for this PR.

> Maybe we should use a different thread to discuss more details how we can get closer to this direction in an agreed backwards compatible way?

I'll start an issue.  Ideally there would be no spec fields that we would write back anyway, so the point would be moot.

> This sounds to me that we'll want to invest more time to increase the quality of the v1alpha1 resources? I

The idea was to make the changes more gradually in v1alpha1, and then promote to v1beta1 as a non-breaking change.  It's about structuring the work into small chunks (e.g. one ref at a time), rather than one big promotion with fields changing to refs etc

scripts/generate-google3-docs/resource-reference/templates/vertexai_vertexaidataset.tmpl
jingyih: Thank you! Updated to use `v1beta1` in docs.

jingyih:
```
--- FAIL: TestAllInSeries/samples/project-level-policy#01 (88.81s)
```

@gemmahou @justinsb Do we know if this test is flaky? Does not looks like the failure is related to this PR.

justinsb:
> --- FAIL: TestAllInSeries/samples/project-level-policy#01 (88.81s)
> @gemmahou @justinsb Do we know if this test is flaky? Does not looks like the failure is related to this PR.

That's actually seems to be because the project isn't going ready, because the default network isn't created, fixed in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1228 .  I'll rebase that one.  The tricky question is why this test ever works :-)

scripts/generate-google3-docs/resource-reference/templates/vertexai_vertexaidataset.tmpl
maqiuyujoyce: > I think you are right but how do I confirm which API endpoint it is using?

It can also be found in code: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/transport/config.go#L565

config/servicemappings/vertexai.yaml
maqiuyujoyce: In that case a valueTemplate will be needed for resourceID configuration.
Example: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/config/servicemappings/monitoring.yaml#L30
The format of the target field of the example configuration can be found in the TF field description: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/monitoring_alert_policy#attributes-reference

jingyih:
> > --- FAIL: TestAllInSeries/samples/project-level-policy#01 (88.81s)
> > @gemmahou @justinsb Do we know if this test is flaky? Does not looks like the failure is related to this PR.
> 
> That's actually seems to be because the project isn't going ready, because the default network isn't created, fixed in #1228 . I'll rebase that one. The tricky question is why this test ever works :-)

The test is passing

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_vertexaidatasets.vertexai.cnrm.cloud.google.com.yaml
```diff
@@ -67,11 +67,215 @@ spec:
...
+                    status:
+                      description: Status is the status of the condition. Can be True,
+                        False, Unknown.
+                      type: string
+                    type:
+                      description: Type is the type of the condition.
+                      type: string
+                  type: object
+                type: array
+              createTime:
```
justinsb, line 176: Should we put this under observedState?  (While we're making breaking changes)

(It's also relatively unimportant I think, so I'm OK if we want to ignore it alogether also)

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_vertexaidatasets.vertexai.cnrm.cloud.google.com.yaml
```diff
@@ -67,11 +67,215 @@ spec:
...
+                description: The timestamp of when the dataset was created in RFC3339
+                  UTC "Zulu" format, with nanosecond resolution and up to nine fractional
+                  digits.
+                type: string
+              name:
```
justinsb, line 181: -> observedState.name ?

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_vertexaidatasets.vertexai.cnrm.cloud.google.com.yaml
```diff
@@ -67,11 +67,215 @@ spec:
...
+                  Google.
+                type: string
+              observedGeneration:
+                description: ObservedGeneration is the generation of the resource
+                  that was most recently observed by the Config Connector controller.
+                  If this is equal to metadata.generation, then that means that the
+                  current reported status reflects the most recent desired state of
+                  the resource.
+                type: integer
+              updateTime:
```
justinsb, line 192: I propose ignoring this field for now; I think it will likely be out of date whenever a change is made externally.

justinsb:
1) Proposed moving some fields to observedState (while we're making breaking changes)
2) Proposed dropping the last-updated timestamp field (until we need it)
3) I think we should ignore state-into-spec on this resource: observedState is the replacement, I don't see any reason to support reading from spec if we're already making breaking changes.  "Ignoring state into spec" could mean your saying that there are no state-into-spec relevant fields, but if we added an exception into the state-into-spec logic for this Kind that would make me very happy!

config/servicemappings/vertexai.yaml
jingyih: I see. I realized that I had misunderstood the mapping direction; it was actually the opposite of what I initially thought... 
But for this particular resource, both `resourceID` and `name` has the same format and value. Why is a valueTemplate still needed?

jingyih:
> 1. Proposed moving some fields to observedState (while we're making breaking changes)
> 2. Proposed dropping the last-updated timestamp field (until we need it)
> 3. I think we should ignore state-into-spec on this resource: observedState is the replacement, I don't see any reason to support reading from spec if we're already making breaking changes.  "Ignoring state into spec" could mean your saying that there are no state-into-spec relevant fields, but if we added an exception into the state-into-spec logic for this Kind that would make me very happy!

These fields are already under KRM `status`. Do we want to put all GCP status fields into observedState, and leave only KRM related status fields directly under `status`? If so, should we start with this resource (which means this resource will be inconsistent with other existing resources)?

config/servicemappings/vertexai.yaml
jingyih: Discussed offline with Joyce. The valueTemplate is indeed necessary. Without valueTemplate, the `spec.resourceID` will look like `projects/674741657116/locations/us-central1/datasets/7410406005512601600`, which is not what we want. The correct resourceID value should only contain the last part of the string. e.g. `7410406005512601600`.

maqiuyujoyce:
@justinsb thank you for sharing your thoughts!

> Simply that it makes it harder to create a resource in "the current project". I need to inject my gcloud config get project into every resource, instead of annotating the namespace once. I guess I could create a project in the namespace, and I don't think making projectRef optional would be blocking, so I don't think this is a blocker for this PR.

Do we have any use cases/feedback from customers about the `projectRef` field? If we can track them in a GitHub issue, then it will help us make decisions in the future (if we indeed want to drop `projectRef`). Also, it seems that you value convenience over consistency for hierarchical references in general? How will it impact users' GitOps story then?

We can also start with making `projectRef` optional (right now it is not possible, b/202396707 and b/192596388).

> The idea was to make the changes more gradually in v1alpha1, and then promote to v1beta1 as a non-breaking change. It's about structuring the work into small chunks (e.g. one ref at a time), rather than one big promotion with fields changing to refs etc

I think it will be impacted by our contributing model and the users' maintenance style.

Incremental changes makes a lot of sense if the changes are driven by external contributors, e.g. whenever they find a field needs to be a reference, they make the change, and we don't ask them to configure all the reference fields correctly. Though if the API team/customer just want the resource to be 99.99% correct/promoted to v1beta1, then naturally, they'll fix everything at once, right?

It may also only make a difference to users who prefer continuous maintenance of resources. If a user prefers to make multiple upgrades, and each time deal with a couple of breaking schema changes, then they'll benefit from the incremental changes. If they prefer to make a single upgrade to v1beta1 and solve all the breaking schema changes at once, then they will not benefit from the incremental changes.

> Proposed moving some fields to observedState (while we're making breaking changes)

Could you share more rationale behind this proposal? It looks like you proposed we move all the previous `status` fields coming from TF schema to be under `status.observedState`. We don't have a prior art for it and I wondered if it is a bit out of scope of this PR.

> Proposed dropping the last-updated timestamp field (until we need it)

`status.updateTime` seems to be a field coming from TF schema (so ultimately from the GCP resource). Historically we support all the status fields unless it caused testing issues. So I'm curious: Why do you think we don't want this field? Is "not supporting status.updateTime" a new guideline for how we add a resource in general?

> I think we should ignore state-into-spec on this resource: observedState is the replacement, I don't see any reason to support reading from spec if we're already making breaking changes. "Ignoring state into spec" could mean your saying that there are no state-into-spec relevant fields, but if we added an exception into the state-into-spec logic for this Kind that would make me very happy!

According to the discussion with PM, it looks like we're okay with only supporting `state-into-spec: absent` option for new resources as long as we document it properly.
I'll start a new proposal and work with @jingyih offline to have a quick workaround for it in this resource.
Agreed that `state-into-spec: merge` is not really necessary as there is no optional and computed fields in this resource. Meanwhile, `observedState` doesn't seem to be required field either for the same reason.

justinsb:
> Do we want to put all GCP status fields into observedState, and leave only KRM related status fields directly under status? 

I believe so.   When we decided to nest these under observedState, we knew we were creating tech debt for every other GCP fields under "status".  As we're making a breaking change to the resource, the time has come for this particular resource.

I'd be open to duplicating the fields if we're starting with v1beta1, but ... we're starting from v1alpha1 here.

@cheftako was this your interpretation also?

> If so, should we start with this resource (which means this resource will be inconsistent with other existing resources)?

I believe so.

justinsb:
> projectRef field... can track them in a GitHub issue

Agreed - I think we've decided that we could make projectRef optional in future, so not a blocker for this PR.

> Why do you think we don't want updateTime?

I think the value in KRM will likely be out of date, and I think it likely has little value when you're using KRM.  If we have a use-case, great.  But once we add it, it's very hard to take away.  OTOH, it's very easy to add if someone asks for it.  So I propose not adding it.

> observedState doesn't seem to be required field either for the same reason.

My interpretation was that status.observedState would be where we would put the output fields from GCP.  Do you want to upload your design here?

jingyih:
Moving GCP resource fields under `status` to `status.observedState` requires #1296.

jingyih:
Removing WIP label. Rebased to master, and regenerated everything.

jingyih:
```
--- FAIL: TestAllInSeries/samples/project-level-policy#01 (95.36s)
```

https://github.com/GoogleCloudPlatform/k8s-config-connector/actions/runs/8429510627/job/23085935625?pr=1271#step:4:25425

config/servicemappings/vertexai.yaml
```diff
@@ -0,0 +1,64 @@
...
+      kind: VertexAIDataset
+      # This resource does not support import.
+      skipImport: true
+      idTemplate: "{{name}}"
+      idTemplateCanBeUsedToMatchResourceName: false
+      resourceAvailableInAssetInventory: true
+      v1alpha1ToV1beta1: true
+      storageVersion: v1alpha1
+      metadataMapping:
+        labels: labels
```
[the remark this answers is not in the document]
jingyih: > Was the "merge" of the labels in the [current schema](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/vertexai/resource_vertex_ai_dataset.go#L76) intentional in KCC?

I don't understand the "merge" part. Good findings on the Terraform HEAD. I did a quick research, it seems that splitting the "labels" into "optional" and "computed" was a breaking change they introduced in TF 5.0.0. I assume we want to backport the change.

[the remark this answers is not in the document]
justinsb: Let's not backport terraform's breaking changes :-)

I'm going to continue this as a top-level comment on the PR so it doesn't get hidden-by-default in the UI.

maqiuyujoyce:
>  Is there any way in the service mapping today to say "don't write back labels"? If not, maybe we just don't support labels at all initially.

Unfortunately, not writing back labels is not a feature we support. We can ignore the `labels` field for now to unblock this resource (and potentially other VertexAI resources if they all support both configurable and computed fields in the same `labels` list).

maqiuyujoyce:
> Here's what I think the ideal behaviour should be:
> 
> * The set of all labels on the GCP object (what terraform calls effective_labels) should map our status / observedState. I think it's status.observedState.labels .

Agreed.

> * Today in KCC our specified labels come from metadata.labels. I think that's actually not ideal (it means users can't use kube labels for things other than GCP labels, and the set of allowed values is not identical). But that's what we have today, so I imagine we don't want this resource to be inconsistent. We should figure out the right path forwards for labels though.

Agreed that we should by default separate KRM labels and GCP labels in the future. In general I don't recall getting much feedback on it so I guess we may have the flexibility to be really opinionated. Meanwhile we probably still need to decide if we want to keep the feature of the `managed-by-cnrm` system label: https://cloud.google.com/config-connector/docs/concepts/resources#object_metadata. It's actually a useful metric for usage tracking/monitoring.

> * Whether the desired labels are in metadata.labels or spec.labels, we should not write back to the spec desired labels. Those should go into the status.

Agreed for `spec.labels`, but it's tricky for `metadata.labels` because I don't think within KRM there is distinction between desired labels and observed labels as it's the metadata.

jingyih:
So my understanding is that we have two options.

Option 1. Ignore Terraform "labels" field. This means users cannot configure labels. This also means we are not surfacing the labels generated by Vertex GCP API in KRM object.

Option 2. backport the breaking change in TF. This means users can configure labels, and the labels generated by Vertex GCP API can be surfaced under `status.ObservedState.Effective_labels`.

maqiuyujoyce:
I'm leaning towards option 1 unless there is an explicit feature request about this field.

jingyih:
> I'm leaning towards option 1 unless there is an explicit feature request about this field.

Sounds good to me. Ignoring the "labels" field for now. We could do option 2 in the future if needed.

jingyih:
Rebased to pick up a presubmit failure fix.

maqiuyujoyce:
Looks like the unit tests failed? Let me know when it's ready for review.

jingyih:
```
=== RUN   TestOutputOnlyFieldsAreUnderObservedState/vertexai.cnrm.cloud.google.com/v1alpha1/VertexAIDataset
    allowlist_test.go:121: 'status.observedState' exists, but it shouldn't
```

Interesting. It is expected that the observedState should exist for `VertexAIDataset`. I will take a look at the unit test.

maqiuyujoyce:
It's possible that the test logic needs to be updated and/or the allowlist should be updated.

jingyih:
Rebased to pick up #1550.

pkg/resourceskeleton/testdata/uri-skeleton.yaml
```diff
@@ -1108,3 +1108,12 @@
...
   URI: "https://edgecontainer.googleapis.com/v1/projects/kcc-test/locations/us-central1/vpnConnections/edgecontainervpnconnection-test"
+- ExpectedSkeleton:
+    apiVersion: vertexai.cnrm.cloud.google.com/v1beta1
+    kind: VertexAIDataset
+    spec:
+      projectRef:
+    external: kcc-test
+    location: us-central1
+  ResourceConfigId: VertexAIDataset
+  URI: "https://us-central1-aiplatform.googleapis.com/v1beta1/projects/kcc-test/locations/us-central1/datasets/7662677953389527040?alt=json"
```
maqiuyujoyce, line 1119: Nit: Can we use a fake dataset ID?

scripts/generate-google3-docs/resource-reference/_toc.yaml
```diff
@@ -525,3 +525,7 @@ toc:
   section:
   - title: "VPCAccessConnector"
     path: /config-connector/docs/reference/resource-docs/vpcaccess/vpcaccessconnector.md
+- title: "Vertex AI"
```
maqiuyujoyce, line 528: Could you move the section up to be before "VPC Access" section? I think we prefer to for the alphabetical order but not the ASCII char order.

scripts/generate-google3-docs/resource-reference/overview.md
```diff
@@ -830,6 +830,10 @@ issues for {{product_name_short}}.
   <td><a href="/vpc/docs/">{{serverless_vpc_access_name}}</a></td>
   <td><a href="/config-connector/docs/reference/resource-docs/vpcaccess/vpcaccessconnector.md">VPCAccessConnector</a></td>
 </tr>
+<tr>
```
maqiuyujoyce, line 833: Similarly, could you move this section up?

maqiuyujoyce (commented):
/lgtm except for a couple of doc ordering issue.

pkg/resourceskeleton/testdata/uri-skeleton.yaml
jingyih: Done.

scripts/generate-google3-docs/resource-reference/_toc.yaml
jingyih: Done. Is the ordering enforced? I don't remember seeing it in the new resource guide.

scripts/generate-google3-docs/resource-reference/overview.md
jingyih: Done

scripts/generate-google3-docs/resource-reference/_toc.yaml
maqiuyujoyce: Good catch! Could you also do a quick update of the doc? Would be nice if we have tests to automate it, but that's not a priority (b/315184931).

maqiuyujoyce:
/lgtm
/approve

Thank you so much for supporting the resource, @jingyih !

PR #1272, opened 2024-02-21
Refactor: rewrite SplitYAML to be more conventional

justinsb:
I think the behaviour here changes in the lint-fixes; refactor to
follow the normal pattern so this is more self-evident code.

### DIFF
--- a/operator/scripts/utils/yaml.go
+++ b/operator/scripts/utils/yaml.go
@@ -66,11 +66,16 @@ func SplitYAML(yamlBytes []byte) ([][]byte, error) {
 	r := bytes.NewReader(yamlBytes)
 	dec := goyaml.NewDecoder(r)
 	results := make([][]byte, 0)
-	var value map[string]interface{}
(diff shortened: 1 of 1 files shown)


operator/scripts/utils/yaml.go
```diff
@@ -66,11 +66,16 @@ func SplitYAML(yamlBytes []byte) ([][]byte, error) {
 	r := bytes.NewReader(yamlBytes)
 	dec := goyaml.NewDecoder(r)
 	results := make([][]byte, 0)
-	var value map[string]interface{}
-	for eof := dec.Decode(&value); errors.Is(eof, io.EOF); eof = dec.Decode(&value) {
```
acpana, line 70: oh it was a double Decode here that's the issue, right ?

acpana:
thanks justin !! 🥇 

/lgtm
/approve

operator/scripts/utils/yaml.go
justinsb: It should be `...; !errors.Is(...); ...` the negation was missing

But yeah, the code is too clever for me :-)

PR #1275, opened 2024-02-21
Powertool to force-set a field

justinsb:
Create a "powertools" section of the CLI for our dangerous and experimental commands. 

Our first powertool, force-set will override the immutable field webhook and change the value of a field.


pkg/cli/powertools/cmd.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/cli/powertools/forcesetfield"
+	"github.com/spf13/cobra"
+)
+
+func AddCommands(parent *cobra.Command) {
+	powertoolsCmd := &cobra.Command{
+		Use:   "powertools",
+		Short: "Powertools holds our experimental / dangerous tools",
```
cheftako, line 25: Some sort of description of what it will do would be nice.

pkg/cli/powertools/forcesetfield/cmd.go
```diff
@@ -0,0 +1,173 @@
...
+	o.FieldManager = "change-state-into-spec"
+	o.DryRun = true // While we are developing this
+}
+
+func AddCommand(parent *cobra.Command) {
+	var options Options
+	options.PopulateDefaults()
+
+	cmd := &cobra.Command{
+		Use:   "force-set-field KIND NAME FIELD.PATH=VALUE",
```
cheftako, line 56: Ahh, sorry I should have read through the whole PR before my initial comment.

cheftako: If the primary use case is setting state-into-spec: absent then maybe have that as an example? I assume the KIND and NAME here are for the object being modified but I think using a 'random' KCC type for that would still be helpfull. Also what about namespace for this?

pkg/cli/powertools/kubecli/options.go
```diff
@@ -0,0 +1,50 @@
...
+	Namespace string
+}
+
+func (o *ObjectOptions) PopulateDefaults() {
+
+}
+
+func (o *ObjectOptions) AddFlags(cmd *cobra.Command) {
+	// cmd.Flags().StringVar(&o.Name, "name", o.Name, "Name of the object to change")
+	cmd.Flags().StringVarP(&o.Namespace, "namespace", "n", o.Namespace, "Namespace of the object")
```
cheftako, line 49: I assume this works for force-set-field but why is this option here and not with the rest of that command.

pkg/cli/powertools/forcesetfield/cmd.go
justinsb: Actually that's a separate tool.  This one is for setting a specific field like location on a bucket.

I figured this tool is much more straightforward than the state-into-spec tool, so we can deal with that one separately

pkg/cli/powertools/kubecli/options.go
justinsb: You caught me!  It's because I originally did state-into-spec, then did this tool, they both need to read an object and so they both have e.g. the Namespace field.  So Namespace is refactored into its own type for commands that need to read/write objects.  But of course, now that this PR only has one command, it looks like YAGNI :-)

pkg/cli/powertools/forcesetfield/cmd.go
```diff
@@ -0,0 +1,173 @@
...
+
+// Options configures the behaviour of the ChangeStateIntoSpec operation.
+type Options struct {
+	kubecli.ClusterOptions
+	kubecli.ObjectOptions
+
+	// FieldManager is the field-manager owner value to use when making changes
+	FieldManager string
+
+	// DryRun is true if we should not actually make changes, just print the changes we would make
```
cheftako, line 39: If we ship it true I think its dangerous to ever undo that. I'm fine with that but I want to be clear I don't think we should change this after we ship it.

justinsb: Yes, I see your point, I'll fix it up!

justinsb:
Thanks @cheftako 

/hold cancel

Issue #16, opened 2019-07-24
Error creating sql replicated primary-secondary instances: Operation failed because another operation was already in progress


AlexBulankou:
**TLDR**: This is expected that secondary instance cannot be created before primary, but also expected that the controller will not attempt to create the instance, until primary creation is complete and not issue calls that will result in errors.

Try to create 2 SQL instances:
```yaml
apiVersion: sql.cnrm.cloud.google.com/v1alpha2
kind: SQLInstance
metadata:
  labels:
    label-one: "value-one"
  name: mysql-instance-sample-with-replication-primary
spec:
  databaseVersion: MYSQL_5_7
  region: us-central1
  settings:
    tier: db-f1-micro
    backupConfiguration:
        binaryLogEnabled: true
        enabled: true
        startTime: "18:00"
    ipConfiguration:
      requireSsl: true
    locationPreference:
        zone: us-central1-b
---
apiVersion: sql.cnrm.cloud.google.com/v1alpha2
kind: SQLInstance
metadata:
  labels:
    label-one: "value-one"
  name: mysql-instance-sample-with-replication-secondary
spec:
  databaseVersion: MYSQL_5_7
  region: us-central1
  masterInstanceName: mysql-instance-sample-with-replication-primary
  mysqlReplicaConfiguration:
    connectRetryInterval: 30
  settings:
    tier: db-f1-micro
    ipConfiguration:
      requireSsl: true
    locationPreference:
        zone: us-central1-c
```

The first instance is created successfully, the second instance is not created and in the event log I can see the following errors, first this one: `Could not read response as json for google api sqladmin-v1beta4: reading the error from the request is not implemented`, which seems to be common and then another one with more details:

```
 jsonPayload: {
  controller:  "sqlinstance-controller"   
  error:  "error syncing resource: Create call failed: googleapi: Error 409: Operation failed because another operation was already in progress., operationInProgress"   
  level:  "error"   
  logger:  "kubebuilder.controller"   
  msg:  "Reconciler error"   
  request:  "default/mysql-instance-sample-with-replication-secondary"   
  stacktrace:  "cnrm.googlesource.com/cnrm/vendor/github.com/go-logr/zapr.(*zapLogger).Error
	/go/src/cnrm.googlesource.com/cnrm/vendor/github.com/go-logr/zapr/zapr.go:128
[...]

PR #1280, opened 2024-02-23
Support state-into-spec user override in ConfigConnectorContext

maqiuyujoyce:
### Change description

Fixes b/320784855

* Added the optional `spec.stateIntoSpec` field in ConfigConnectorContext for namespace-level default value override.
* Updated the StateIntoSpecDefaulter to fetch the live value of the `spec.stateIntoSpec` field in ConfigConnectorContext to determine the default `state-into-spec` value on the fly.
* Updated the test environment to ensure integration test passes.
* Updated test harness and test setups to fix mock tests.

### Tests you have done

`go test -v -tags=integration ./config/tests/samples/create -test.run TestAll -run-tests pubsubtopic` passed and the created resource has `state-into-spec` annotation set to `merge`.


pkg/k8s/defaulter.go
```diff
@@ -29,35 +34,59 @@ type Defaulter interface {
...
+		Name:      operatork8s.ConfigConnectorContextAllowedName,
+	}
+	ccc, err := v.getConfigConnectorContext(ctx, cccNamespacedName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("error getting ConfigConnectorContext object %v/%v: %w", cccNamespacedName.Namespace, cccNamespacedName.Name, err)
+	}
+	if ccc.APIVersion != "v1beta1" {
```
justinsb, line 58: Harmless, but you don't need to check this, you requested v1beta1 so you'll always get v1beta1

pkg/test/controller/reconcile.go
```diff
@@ -63,6 +66,28 @@ func StartTestManagerInstance(env *envtest.Environment, testType test.Type, whCf
 	return mgr, stopFunc
 }
 
+// Without the scheme, dynamic test returns `no kind is registered for the type
```
justinsb, line 69: Right - if you're not using unstructured, you need to be sure that the type is registered in the scheme.  As we want to read a particular field and we own the CRD, using types is a good plan.

pkg/test/controller/reconcile.go
```diff
@@ -63,6 +66,28 @@ func StartTestManagerInstance(env *envtest.Environment, testType test.Type, whCf
...
+// XXX in scheme` error.
+// Even with the scheme, dynamic test still returns `the server could not find
```
justinsb, line 71: Odd - I'm guessing the CCC CRD isn't registered, as it looks like you handle the case where the CCC object does not exist.   Do we know where the error is coming from / if the CRD is registered?

pkg/k8s/defaulter.go
maqiuyujoyce: Good point! Removed.

pkg/test/controller/reconcile.go
maqiuyujoyce: Moved the scheme registration to the test environment.

pkg/test/controller/reconcile.go
maqiuyujoyce: I debugged for while today and finally realized the CRD installation and scheme registration need to be done separately. With the CRD, unstructured can be created/retrieved; with the scheme, the typed object can be created/retrieved.

maqiuyujoyce:
Snapshot the original PR description for posterity:

WIP experiment based on [this comment](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1266/files#r1497912625).

Current status:
* Dynamic test failed with the following error: `failed to get API group resources: unable to retrieve the complete list of server APIs: core.cnrm.cloud.google.com/v1beta1: the server could not find the requested resource`
   ```
   reconcile.go:215: error was not considered transient; chain is [[*fmt.wrapError: error handling default values for resource 'crxm2f26xxwugsrtpq6a/pubsubschema-crxm2f26xxwugsrtpq6a': error getting ConfigConnectorContext object crxm2f26xxwugsrtpq6a/configconnectorcontext.core.cnrm.cloud.google.com: failed to get API group resources: unable to retrieve the complete list of server APIs: core.cnrm.cloud.google.com/v1beta1: the server could not find the requested resource] [*fmt.wrapError: error getting ConfigConnectorContext object crxm2f26xxwugsrtpq6a/configconnectorcontext.core.cnrm.cloud.google.com: failed to get API group resources: unable to retrieve the complete list of server APIs: core.cnrm.cloud.google.com/v1beta1: the server could not find the requested resource] [*fmt.wrapError: failed to get API group resources: unable to retrieve the complete list of server APIs: core.cnrm.cloud.google.com/v1beta1: the server could not find the requested resource] [*discovery.ErrGroupDiscoveryFailed: unable to retrieve the complete list of server APIs: core.cnrm.cloud.google.com/v1beta1: the server could not find the requested resource]]
    testreconciler.go:110: reconcile returned unexpected error: error handling default values for resource 'crxm2f26xxwugsrtpq6a/pubsubschema-crxm2f26xxwugsrtpq6a': error getting ConfigConnectorContext object crxm2f26xxwugsrtpq6a/configconnectorcontext.core.cnrm.cloud.google.com: failed to get API group resources: unable to retrieve the complete list of server APIs: core.cnrm.cloud.google.com/v1beta1: the server could not find the requested resource
   ```
* `make docker-build` failed without a clear error message:
  ```
  #16 20.21 go: downloading github.com/census-instrumentation/opencensus-proto v0.4.1
  #16 21.02 pkg/k8s/defaulter.go:26:2: no required module provides package github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/apis/core/v1beta1; to add it:
  #16 21.02 	go get github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/apis/core/v1beta1
  #16 21.02 pkg/k8s/defaulter.go:27:2: no required module provides package github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/k8s; to add it:
  #16 21.02 	go get github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/k8s
  ------
  executor failed running [/bin/sh -c CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags "-s -w" -o deletiondefender github.com/GoogleCloudPlatform/k8s-config-connector/cmd/deletiondefender]: exit code: 1
  make: *** [Makefile:126: docker-build-builder] Error 1
  ```

Suspecting the test manager is not set up correctly and potentially some imports are problematic.

maqiuyujoyce:
/do-not-merge/hold

`make docker-build` still fails as commented above. Needs further investigation.

maqiuyujoyce:
`make docker-build` now finishes successfully with the latest commit.

pkg/controller/mocktests/secretmanager_secret_test.go
```diff
@@ -35,6 +31,10 @@ import (
 
 	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/pkg/storage"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/dcl/clientconfig"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test"
+	testreconciler "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test/controller/reconciler"
+	tfprovider "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/tf/provider"
```
acpana, line 37: nit: turning on the `gci` linter will help us w these down the road; you don't have to, but you could revert changes to this file.

operator/pkg/apis/core/v1beta1/configconnectorcontext_types.go
```diff
@@ -42,6 +42,18 @@ type ConfigConnectorContextSpec struct {
...
+	// cnrm.cloud.google.com/state-into-spec annotation if the annotation is
+	// unset in a resource.
+	// 'absent' means unspecified fields in the resource spec stay unspecified
+	// after the successful reconciliations.
+	// 'merge' means the unspecified fields in the resource spec are populated
+	// after the successful reconciliations if the fields are computed/defaulted
+	// by the API.
+	//+kubebuilder:validation:Enum=absent;merge
+	//+kubebuilder:validation:Optional
+	StateIntoSpec *string `json:"stateIntoSpec,omitempty"`
```
acpana, line 56: unless I misunderstood, Justin mentioned we may want to use upper case here  for the values themselves (Absent, Merge)

xref https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1300#discussion_r1505217530

If you define a default along the type, then maybe we don't need the SystemDefaultValue validation?

pkg/k8s/defaulter.go
```diff
@@ -29,35 +34,58 @@ type Defaulter interface {
...
 	}
 	return true, nil
 }
 
-func (v *StateIntoSpecDefaulter) getValue() string {
-	if v.userOverride == nil {
-		return v.defaultValue
+func (v *StateIntoSpecDefaulter) getDefaultValue(userOverride *string, systemDefaultValue string) (string, error) {
+	if !isAcceptedValue(systemDefaultValue, StateIntoSpecAnnotationValues) {
+		return "", fmt.Errorf("invalid system default value '%v' for '%v' annotation, need to be one of {%v}", systemDefaultValue, StateIntoSpecAnnotation, strings.Join(StateIntoSpecAnnotationValues, ", "))
```
acpana, line 74: IIUC, we don't need to check that the systemDefaultValue is allowed every time we apply a default right ? Since that is a static value, it may not change, unlike the userOverride

operator/pkg/apis/core/v1beta1/configconnectorcontext_types.go
```diff
@@ -42,6 +42,18 @@ type ConfigConnectorContextSpec struct {
...
 	BillingProject string `json:"billingProject,omitempty"`
+
+	// StateIntoSpec is the user override of the default value for the
+	// cnrm.cloud.google.com/state-into-spec annotation if the annotation is
+	// unset in a resource.
+	// 'absent' means unspecified fields in the resource spec stay unspecified
+	// after the successful reconciliations.
+	// 'merge' means the unspecified fields in the resource spec are populated
+	// after the successful reconciliations if the fields are computed/defaulted
+	// by the API.
```
acpana, line 53: smoll nit:

```suggestion
	// StateIntoSpec is the user override of the default value for the
	// 'cnrm.cloud.google.com/state-into-spec' annotation if the annotation is
	// unset for a resource.
	// 'absent' means that unspecified fields in the resource spec stay unspecified
	// after successful reconciliation.
	// 'merge' means that unspecified fields in the resource spec are populated
	// after a successful reconciliation if those unspecified fields are computed/defaulted
	// by the API.
```

acpana (commented):
mostly non blocking and nits

/hold for author visibility

acpana:
also, assuming lower case naming is fine

/lgtm
/approve

I trust you to close on the items above as appropriate; Otherwise, happy to have another look asap after changes

operator/pkg/apis/core/v1beta1/configconnectorcontext_types.go
maqiuyujoyce: Thank you! Always glad to improve in tech doc writing and English!

pkg/controller/mocktests/secretmanager_secret_test.go
acpana: Yes `gci` will sort them for us but it may not follow necessarily any predefined patterns we want to enforce.

pkg/crd/crdloader/crdloader.go
```diff
@@ -155,7 +156,27 @@ func formatGVK(group, version, kind string) string {
...
+	for _, loadFunc := range loadFuncs {
+		crds, err := loadFunc()
+		if err != nil {
+			return nil, err
+		}
+		results = append(results, crds...)
+	}
+	return results, nil
+}
+
```
gemmahou, line 174: nit: It would be beneficial to have comments explaining the differences between LoadCRDs and LoadAllCRDs, as their names are similar and may lead to confusion.

IIUC LoadAllCRDs includes CRDs + OperatorCRDs

operator/pkg/apis/core/v1beta1/configconnectorcontext_types.go
maqiuyujoyce: @gemmahou @jingyih do you have any thoughts on the enum values we should use here?

gemmahou: I might miss something but I'm wondering why the values are different at namespace/cluster level and resource level? Can we make them all upper case or lower case? While, I feel for resource annotations(like deletion-policy, etc) we always use lower case.

maqiuyujoyce: > If you define a default along the type, then maybe we don't need the SystemDefaultValue validation?

@acpana sorry I forgot this question. What I want to achieve here is to make it clear to KCC whether the field is unset. Defining the default value in the type will make userOverride always populated in CCC when we get it in the controllers so the user intention can be vague. Knowing whether the field is set or not may help us determine proper warnings/instructions when we gradually "turn off" state-into-spec.

> I might miss something but I'm wondering why the values are different at namespace/cluster level and resource level? Can we make them all upper case or lower case?

@gemmahou  Currently the values at namespace/cluster level and resource level are not different, they are all lowercase values. We should not use the values with the first letter capitalized for resource-level configuration because this is an existing feature, and changing the available values is kinda expensive. If consistency is more important, then we should make all of them (namespace/cluster level and resource level) in lowercase.

pkg/crd/crdloader/crdloader.go
maqiuyujoyce: Done.

acpana:
/lgtm
/approve

thanks for addressing comments/ questions 🥇 💯

operator/pkg/apis/core/v1beta1/configconnectorcontext_types.go
justinsb: KRM Enums are upper-case: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#constants

We can support both values if we think it will cause confusion, I don't think it will.

pkg/crd/crdloader/crdloader.go
```diff
@@ -154,8 +155,33 @@ func formatGVK(group, version, kind string) string {
...
 }
 
+// LoadCRDs returns the list of CRDs Config Connector manages that map to GCP
+// resources.
 func LoadCRDs() ([]apiextensions.CustomResourceDefinition, error) {
-	crdsRoot := repo.GetCRDsPath()
+	return loadCRDs(repo.GetCRDsPath())
+}
+
+// LoadAllCRDs returns the list of all the CRDs Config Connector manages,
```
justinsb, line 164: Is there any reason to call LoadCRDs instead of LoadAllCRDs?

operator/pkg/apis/core/v1beta1/configconnectorcontext_types.go
maqiuyujoyce: SG. We should keep an eye on the customer feedback about it.

pkg/crd/crdloader/crdloader.go
maqiuyujoyce: Looks like LoadCRDs() is called in a few tests in which the test scenarios are probably for KCC resource (non core/operator ones) only.

PR #1282, opened 2024-02-24
mockgcp: support VertexAIDataset

jingyih:
### Change description

Follow up of #1271.

Diffbase: #1271 

### Tests you have done

```
$ E2E_KUBE_TARGET=envtest RUN_E2E=1 E2E_GCP_TARGET=mock go test -test.count=1 -timeout 600s -v ./tests/e2e -run TestAllInSeries/fixtures/vertexaidataset  2>&1 | tee log                                                           
```


mockgcp/mockaiplatform/dataset.go
```diff
@@ -0,0 +1,201 @@
...
+func (s *datasetService) GetDataset(ctx context.Context, req *pb.GetDatasetRequest) (*pb.Dataset, error) {
+	name, err := s.parseDatasetName(req.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	fqn := name.String()
+
+	obj := &pb.Dataset{}
+	fmt.Printf("[debug] getting %s\n", fqn)
```
justinsb, line 48: Nit: probably better to `log := klog.FromContext(ctx)` and then `log.Info`, or just `klog.Infof` , but certainly not a blocker, just one for the future!

(We also have reasonable logging of the requests etc anyway in our tests, so you may not need your own logging.  But it might be that this is just left over for dev, I'm always doing that :-) )

justinsb:
mockgcp commit looks good.  I'll look at the service itself in #1271.  One tip for writing mocks, I don't know if you did this, but it can be helpful to capture the HTTP logs against the real service first, then try to make the results identical (after replacing generated values / timestamps).  I should try to write up the flow / create scripts / record a video!

mockgcp/mockaiplatform/dataset.go
jingyih: Sorry these are just something I randomly added during my debugging. I'll remove them.

jingyih:
> mockgcp commit looks good. I'll look at the service itself in #1271. One tip for writing mocks, I don't know if you did this, but it can be helpful to capture the HTTP logs against the real service first, then try to make the results identical (after replacing generated values / timestamps). I should try to write up the flow / create scripts / record a video!

No I didn't do it. Just followed the mockgcp guide. Having a script or video would be super cool!

gemmahou:
> > mockgcp commit looks good. I'll look at the service itself in #1271. One tip for writing mocks, I don't know if you did this, but it can be helpful to capture the HTTP logs against the real service first, then try to make the results identical (after replacing generated values / timestamps). I should try to write up the flow / create scripts / record a video!
> 
> No I didn't do it. I just followed the mockgcp guide. Having a script or video would be super cool!

To get the http log, you can specify two more variables `GOLDEN_REQUEST_CHECKS=1 WRITE_GOLDEN_OUTPUT=1` in the command you run the mock test.

jingyih:
PR is rebased.

jingyih:
Mock tests failed 3 times in a row. Did not find a specific test that was failed.
Found the following at the end of the log
```
Error: The operation was canceled.
```

PR #1287, opened 2024-02-26
wip:feat: watch cc,ccc in core controller

acpana:
- add a simple controller to observe CC/CCC in core kcc.
- add a live config singleton to capture config from CC/ CCC
    
todo
- smoke controller test for cc/ccc; e2e for actions
- add in crds CC/ CCCs
- more tests! units! decide on LC home

sharing this early for collaboration

### DIFF
--- a/pkg/controller/corecc/controller.go
+++ b/pkg/controller/corecc/controller.go
@@ -1,0 +1,104 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// logic for reading ConfigConnector (CC) and ConfigConnectorContext (CCC) CRs.
+package corecc
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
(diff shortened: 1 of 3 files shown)


acpana:
cc @justinsb @maqiuyujoyce for awareness

not quite ready for review as I wanna add some smoke tests and wire through the CRDs

pkg/controller/corecc/controller.go
```diff
@@ -0,0 +1,104 @@
...
+// onto the operator's controllers for CC & CCC.
+type CoreReconciler struct {
+	mgr                     manager.Manager
+	lc                      *LiveConfig
+	gvk                     schema.GroupVersionKind
+}
+
+var _ reconcile.Reconciler = &CoreReconciler{}
+
+func (r *CoreReconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
```
justinsb, line 47: I don't think you need to do this; you should just call this at the top of each reconcile operation (probably in a helper function!)

```
var cc kcc.ConfigConnector
if err := r.client.Get(ctx, types.NamespacedName{Name: foo}, &cc); err != nil {
return fmt.Errorf("getting configconnector: %w", err)
}
if cc.Spec.Mode == "Namespaced" { // We probably have this anyway, but this is where it comes from
var ccc kcc.ConfigConnectorContext
if err := r.client.Get(ctx, types.NamespacedName{Name: foo, Namespace: req.GetNamespace()}, &ccc); err != nil {
return fmt.Errorf("getting configconnectorcontext: %w", err)
}
}
```

In a normal controller-runtime operator, that will be cached.  We turned off caching in KCC, so you'll want to turn it back on for these kinds, it should be in the manager.Options.

acpana: thanks for the guidance! opened https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1300

the ideas in these PR are self contained enough that it doesn't make sense to me to continue on it so will close it in favor of the one above

PR #1291, opened 2024-02-27
chore: don't use caching for go task in lint github action

justinsb:
This seems to cause double caching.

https://github.com/golangci/golangci-lint-action/issues/135


acpana:
thanks for patching this 💯 

/lgtm
/approve

PR #1292, opened 2024-02-27
wip:refactor: move core apis out of operator

acpana:
If we use the core apis for cc/ccc outside of the operator package maybe it makes sense to move them out?

If I get buy in for this, I'll rework the wiring for crd generation (ie make manifests) since I bet this breaks it :) and any other refactoring needed. I'm also game to wait out any PRs that touch CC/ CCCs and then do the refactor work so folks don't get massive merge conflicts because of this.


acpana:
I think now that we have an `apis/` package for our direct resources, we can make another argument to add the core types to the `apis/` package.

@justinsb lmk if you think it'd be worth finishing the intent of the change here and am happy to re-open or fix these merge conflicts.

PR #1293, opened 2024-02-27
chore: better error messages in servicemappingloader

justinsb:
Avoids printing a nil error, gives a better contextual hint.


jingyih (commented):
/lgtm
/approve

PR #1295, opened 2024-02-27
tests: try introducing a subtest timeout to unified_test

justinsb:
We would like to fail fast on individual tests; with our mocks no test
should take very long.

### DIFF
--- a/tests/e2e/unified_test.go
+++ b/tests/e2e/unified_test.go
@@ -19,7 +19,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	testvariable "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test/resourcefixture/variable"
 	"hash/fnv"
 	"io"
 	"io/ioutil"
@@ -30,6 +29,7 @@ import (
 	"strconv"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/GoogleCloudPlatform/k8s-config-connector/config/tests/samples/create"
 	opcorev1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/apis/core/v1beta1"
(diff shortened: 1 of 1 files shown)


justinsb:
OK, that was a lucky test failure.  It would have been a timeout on certificatemanagercertificatemapentry, but the timeout kicked in.  I would like to see something clearer than "parent context cancelled" though, so going to give it another try...

justinsb:
Hooray - a fairly rapid timeout:


```
2024-02-28T02:10:05.8240593Z     samples.go:230: error while polling for resource cleanup on CertificateManagerCertificateMapEntry with name 'certificatemanagercertificatemapentry-sample': context deadline exceeded; last seen status: map[conditions:[map[lastTransitionTime:2024-02-28T02:09:25Z message:Delete call failed: error fetching live state: error getting ID for resource: error getting value from reference: could not parse reference resolution value '<nil>' as string reason:DeleteFailed status:False type:Ready]] observedGeneration:3 state:SERVING_STATE_UNSPECIFIED]
2024-02-28T02:10:05.8245645Z     harness.go:347: waiting for controller-runtime manager shutdown
2024-02-28T02:10:05.8246786Z     harness.go:349: controller-runtime manager is shutdown
2024-02-28T02:10:06.9805038Z     unified_test.go:365: subtest timeout after 1m0s
....
 --- FAIL: TestAllInSeries/samples/certificatemanagercertificatemapentry (61.83s)
 ```

So that is WAI - timing out after about a minute.

justinsb:
/assign @gemmahou 

This is a simple idea, but annoyingly tricky to implement (though not too many lines of code!).  There's a proposal to just have a SetTimeout method on the subtest but no real progress yet.  

But I think it's good because it should stop our mocks timing out in github, we'll force-fail any individual test after a minute, so we can see the test failures clearly.  In contrast, when the whole test times out we normally just get a panic and we don't get the `PASS:` / `FAIL:` block

We might raise the time limit e.g. to two minutes in future, but TBH they're already taking longer than I think they should anyway, so I'm hoping we could look at why they are slow (it might be the terraform reconciliation logic having built-in delays)

gemmahou:
Thanks for introducing timeout for individual tests!

/lgtm
/approve

justinsb:
Thanks @gemmahou ... needs another lgtm after rebase I'm afraid!

gemmahou:
The presubmit tests start failing after rebase...

justinsb:
> The presubmit tests start failing after rebase...
Yeah ... we could really use a submit queue :-)

gemmahou:
/lgtm
/approve

thanks! i think the timeout will be useful for vcr testing as well.

PR #1179, opened 2024-02-01
Try moving EnsureFinalizers() up in the TF controller

maqiuyujoyce:
Commit [6869e87](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1179/commits/6869e875d689ec1fddc310ef714261014adb7f11) simply moved the `EnsureFinalizers()` up to the beginning of the reconciler and used the unstructured to do the update. And this resulted in the following error in the dynamic integration test:

```
reconcile.go:191: error was not considered transient; chain is [[*errors.errorString: couldn't update the API server due to 
conflict. Re-enqueue the request for another reconciliation attempt: Operation cannot be fulfilled on 
pubsubschemas.pubsub.cnrm.cloud.google.com "pubsubschema-m3d3aukju4za5uklrxtq": the object has been modified; 
please apply your changes to the latest version and try again]]

testreconciler.go:104: reconcile returned unexpected error: couldn't update the API server due to conflict. Re-enqueue the 
request for another reconciliation attempt: Operation cannot be fulfilled on pubsubschemas.pubsub.cnrm.cloud.google.com 
"pubsubschema-m3d3aukju4za5uklrxtq": the object has been modified; please apply your changes to the latest version and 
try again
```

Looks like the root cause was that unstructured contains generated fields that shouldn't be updated/maintained by the user.

Commit [b53356d](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1179/commits/b53356dd1226d1943031d1781d907c37413802f6) used the parsed Resource instead and it worked.

PR #1300, opened 2024-02-28
feat: pause resource actuation

acpana:
This PR proposes a way to support a `pause` mode for KCC. In particular, it adds new fields in the CC/ CCC specs to support an `actuationMode`. These values are then consumed before we actuate the GCP resources: see the tf controller.

Work that will come as separate PR(s)
- dcl (or direct) controller support for pausing actuation
- docs changes
- sone higher level e2e tests using golden logs

Abandoned ideas
- having a fully fledged simple reconciler for CC/ CCC to watch the resources and keep track of the `actuationMode` values

Renaming
- the `actuationMode` war originally `actuationAction` then `reconciliationMode`.

### DIFF
--- a/operator/config/crd/bases/core.cnrm.cloud.google.com_configconnectorcontexts.yaml
+++ b/operator/config/crd/bases/core.cnrm.cloud.google.com_configconnectorcontexts.yaml
@@ -45,6 +45,16 @@ spec:
           spec:
             description: ConfigConnectorContextSpec defines the desired state of ConfigConnectorContext
             properties:
+              actuationMode:
+                description: The actuation mode of Config Connector controls how resources
+                  are actuated onto the cloud provider. This can be either 'Reconciling'
+                  or 'Paused'. The default is 'Reconciling' where resources get actuated.
+                  In 'Paused', k8s resources are still reconciled with the api server
+                  but not actuated onto the cloud provider.
+                enum:
+                - Reconciling
+                - Paused
+                type: string
               billingProject:
                 description: Specifies the project to use for preconditions, quota
                   and billing. Should only be used when requestProjectPolicy is set
--- a/operator/pkg/apis/core/v1beta1/configconnector_types.go
+++ b/operator/pkg/apis/core/v1beta1/configconnector_types.go
@@ -40,6 +40,16 @@ type ConfigConnectorSpec struct {
 	// When in namespaced mode, you must create a ConfigConnectorContext object per namespace that you want to enable Config Connector in, and each must set `googleServiceAccount` to specify the Google Service Account to be used to authenticate with Google Cloud APIs for the namespace.
 	//+kubebuilder:validation:Enum=cluster;namespaced
 	Mode string `json:"mode,omitempty"`
+
+	// The actuation mode of Config Connector controls how resources are actuated onto the cloud provider.
+	// This can be either 'Reconciling' or 'Paused'.
+	// In 'Paused', k8s resources are still reconciled with the api server but not actuated onto the cloud provider.
+	// If Config Connector is running in 'namespaced' mode, then the value in ConfigConnectorContext (CCC) takes precedence.
+	// If CCC doesn't define a value but ConfigConnecor (CC) does, we defer to that value. Otherwise,
+	// the default is 'Reconciling' where resources get actuated.
+	//+kubebuilder:validation:Enum=Reconciling;Paused
+	//+kubebuilder:validation:Optional
+	Actuation ActuationMode `json:"actuationMode,omitempty"`
 }
 
 // ConfigConnectorStatus defines the observed state of ConfigConnector
--- a/operator/pkg/apis/core/v1beta1/configconnectorcontext_types.go
+++ b/operator/pkg/apis/core/v1beta1/configconnectorcontext_types.go
@@ -54,6 +54,12 @@ type ConfigConnectorContextSpec struct {
 	//+kubebuilder:validation:Enum=absent;merge
 	//+kubebuilder:validation:Optional
 	StateIntoSpec *string `json:"stateIntoSpec,omitempty"`
+	// The actuation mode of Config Connector controls how resources are actuated onto the cloud provider.
+	// This can be either 'Reconciling' or 'Paused'. The default is 'Reconciling' where resources get actuated.
+	// In 'Paused', k8s resources are still reconciled with the api server but not actuated onto the cloud provider.
+	//+kubebuilder:validation:Enum=Reconciling;Paused
+	//+kubebuilder:validation:Optional
+	Actuation ActuationMode `json:"actuationMode,omitempty"`
 }
 
 // ConfigConnectorContextStatus defines the observed state of ConfigConnectorContext
--- a/pkg/controller/kccmanager/kccmanager.go
+++ b/pkg/controller/kccmanager/kccmanager.go
@@ -31,6 +31,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/gcp"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/servicemapping/servicemappingloader"
+
 	tfprovider "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/tf/provider"
 
 	corev1 "k8s.io/api/core/v1"
--- a/pkg/controller/kccmanager/kccmanager.go
+++ b/pkg/controller/kccmanager/kccmanager.go
@@ -31,7 +31,6 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/gcp"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/servicemapping/servicemappingloader"
-
 	tfprovider "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/tf/provider"
 
 	corev1 "k8s.io/api/core/v1"
@@ -85,18 +84,18 @@ func New(ctx context.Context, restConfig *rest.Config, config Config) (manager.M
 		// the calls to AddToScheme(..) will modify the internal maps
 		opts.Scheme = runtime.NewScheme()
 	}
-	// Disable the cache. The cache causes problems in namespaced mode when trying
-	// to read resources in our system namespace.
-	opts.NewClient = nocache.NoCacheClientFunc
 	opts.BaseContext = func() context.Context {
 		return ctx
 	}
+	if err := addSchemes(opts.Scheme); err != nil {
+		return nil, fmt.Errorf("error adding schemes: %w", err)
+	}
+
+	// only cache CC and CCC resources
+	opts.Cache.ByObject = nocache.ByCCandCCC
 	mgr, err := manager.New(restConfig, opts)
 	if err != nil {
 		return nil, fmt.Errorf("error creating new manager: %w", err)
-	}
-	if err := addSchemes(mgr); err != nil {
-		return nil, err
 	}
 
 	// Bootstrap the Google Terraform provider
@@ -147,8 +146,7 @@ func New(ctx context.Context, restConfig *rest.Config, config Config) (manager.M
 	return mgr, nil
 }
 
-func addSchemes(mgr manager.Manager) error {
-	scheme := mgr.GetScheme()
+func addSchemes(scheme *runtime.Scheme) error {
 	if err := corev1.AddToScheme(scheme); err != nil {
 		return fmt.Errorf("error adding 'corev1' resources to the scheme: %w", err)
 	}
--- a/pkg/controller/resourceactuation/resourceactuation.go
+++ b/pkg/controller/resourceactuation/resourceactuation.go
@@ -15,13 +15,73 @@
 package resourceactuation
 
 import (
+	"context"
 	"fmt"
 
+	opv1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/apis/core/v1beta1"
+	opk8s "github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/k8s"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/reconciliationinterval"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+// FetchLiveKCCState tries to fetch the ConfigConnector (CC) resource and the ConfigConnectorContext (CCC)
+// for the resource's namespace if running in Namespaced mode. It ignores not found errors for CC fetching
+// but errors out if KCC is running in Namespaced mode and no CCC is found for the namespace of the resource.
+func FetchLiveKCCState(ctx context.Context, c client.Client, resourceNN types.NamespacedName) (opv1beta1.ConfigConnector, opv1beta1.ConfigConnectorContext, error) {
+	var cc opv1beta1.ConfigConnector
+	if err := c.Get(ctx, types.NamespacedName{
+		Name:      opk8s.ConfigConnectorAllowedName,
+		Namespace: k8s.SystemNamespace,
+	}, &cc); err != nil {
+		if apierrors.IsNotFound(err) {
+			// if no CC exists, then by definition, KCC cannot be running in namespaced mode;
+			return opv1beta1.ConfigConnector{}, opv1beta1.ConfigConnectorContext{}, nil
+		}
+		return opv1beta1.ConfigConnector{}, opv1beta1.ConfigConnectorContext{}, err
+	}
+
+	if cc.Spec.Mode == opk8s.NamespacedMode {
+		var ccc opv1beta1.ConfigConnectorContext
+		if err := c.Get(ctx, types.NamespacedName{
+			Name:      opk8s.ConfigConnectorContextAllowedName,
+			Namespace: resourceNN.Namespace,
+		}, &ccc); err != nil {
+
+			// this should not happen but if we attempt to actuate a resource
+			// AND we are running in namespaced mode, not finding a CCC in that namespace
+			// is an error in the assumptions that KCC has (i.e. that there is a CCC defined
+			// that actively manages resources in that namespace).
+			return cc, opv1beta1.ConfigConnectorContext{}, err
+		}
+		return cc, ccc, nil
+	}
+
+	return cc, opv1beta1.ConfigConnectorContext{}, nil
+}
+
+// DecideActuationMode looks at CC and CCC to see if they specify an actuationMode.
+// - If both CC & CCC specify a actuationMode in Namespaced mode, we defer to the CCC's value.
+//   - If only CC specifies a actuationMode in Namespaced mode, we defer to the CC's value.
+//
+// - If both CC & CCC specify an actuationMode in cluster mode, the CCC specification is irrelevant.
+// - If neither CC nor CCC specify a actuationMode, we defer to the default value defined in apis.
+func DecideActuationMode(cc opv1beta1.ConfigConnector, ccc opv1beta1.ConfigConnectorContext) opv1beta1.ActuationMode {
+	if ccc.Spec.Actuation != "" && cc.Spec.Mode == opk8s.NamespacedMode {
+		return ccc.Spec.Actuation
+	}
+
+	// if no CCC exists or doesn't define a value, defer to the CC's value.
+	if cc.Spec.Actuation != "" {
+		return cc.Spec.Actuation
+	}
+
+	return opv1beta1.DefaultActuationMode()
+}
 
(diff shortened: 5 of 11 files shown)


pkg/controller/resourceactuation/resourceactuation.go
```diff
@@ -15,14 +15,86 @@
...
+	return DefaultActuationAction == "reconciling"
+}
+
+// ShouldActuate checks if the resource should be actuated. It errors out if it cannot fetch the
+// ConfigConnector (CC) resource or the ConfigConnectorContext (CCC) resource for the resource's namespace if running
+// in namespaced mode.
+// - If both CC & CCC specify an actuationAction in namespaced mode, we defer to the CCC's value.
+//   - If only CC specifies an actuationAction in namespaced mode, we defer to the CC's value.
+// - If both CC & CCC specify an actuationAction in cluster mode, the CCC specification is irrelevant.
+func ShouldActuate(ctx context.Context, c client.Client, resourceNN types.NamespacedName) (bool, error) {
```
acpana, line 61: could probably further break this up into a helper to fetch the CC, CCC instances and one to decide on the actuation value but leaving it as is for now unless someone objects

operator/config/crd/bases/core.cnrm.cloud.google.com_configconnectorcontexts.yaml
```diff
@@ -45,6 +45,12 @@ spec:
           spec:
             description: ConfigConnectorContextSpec defines the desired state of ConfigConnectorContext
             properties:
+              actuationAction:
+                description: todo acpana docstring
```
justinsb, line 49: todo(acpana) install Duet

pkg/apis/addtoscheme_cc_ccc_v1beta.go
```diff
@@ -0,0 +1,24 @@
...
+
+package apis
+
+import (
+	opcorev1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/apis/core/v1beta1"
+)
+
+func init() {
+	// register CC, CCC kinds for resourceactuation.ShouldActuate
+	AddToSchemes = append(AddToSchemes, opcorev1beta1.SchemeBuilder.AddToScheme)
```
justinsb, line 23: This is not the normal way, and I worry that it's non-deterministic.  Are we doing this elsewhere?

pkg/controller/resourceactuation/resourceactuation.go
justinsb: No objection to only doing the refactoring when needed!

pkg/controller/resourceactuation/resourceactuation.go
```diff
@@ -15,14 +15,86 @@
...
+	}
+
+	if cc.Spec.Mode == opk8s.NamespacedMode {
+		var ccc opv1beta1.ConfigConnectorContext
+		if err := c.Get(ctx, types.NamespacedName{
+			Name:      opk8s.ConfigConnectorContextAllowedName,
+			Namespace: resourceNN.Namespace,
+		}, &ccc); err != nil {
+			if apierrors.IsNotFound(err) {
+				// if no CCC is defined in this namespace, defer to the "globally"
```
justinsb, line 81: It's probably going to be a lot more readable (and reusable) to first collect the `var cc *CC` and `var ccc *CCC`, then check them in the order of precedence, rather than mixing it into the fetching logic here

pkg/controller/tf/controller.go
```diff
@@ -164,6 +164,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (res
 		r.logger.Info("Skipping reconcile as nothing has changed and 0 reconcile period is set", "resource", req.NamespacedName)
 		return reconcile.Result{}, nil
 	}
+
+	actuate, err := resourceactuation.ShouldActuate(ctx, r.mgr.GetClient(), req.NamespacedName)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	if !actuate {
+		r.logger.Info("Skipping reconcile as actuation action is \"paused\"", "resource", req.NamespacedName)
```
justinsb, line 173: Might be easier to return an ActuationMode than a boolean, and switch on it

operator/pkg/apis/core/v1beta1/configconnectorcontext_types.go
```diff
@@ -42,6 +42,11 @@ type ConfigConnectorContextSpec struct {
 	// Specifies the project to use for preconditions, quota and billing.
 	// Should only be used when requestProjectPolicy is set to BILLING_PROJECT.
 	BillingProject string `json:"billingProject,omitempty"`
+
+	// todo acpana docstring
+	//+kubebuilder:validation:Enum=reconciling;paused
+	//+kubebuilder:validation:Optional
+	ActuationAction string `json:"actuationAction,omitempty"`
```
justinsb, line 49: I think we would normally suggest defining an "enum type" here (as much as  go allows)

KRM suggests upper-case constants also.

For example: https://github.com/kubernetes/kubernetes/blob/cb989b84c82ab8c695d02c9a1866bc27de37caec/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/types.go#L252-L258

Technically you can have `ActuationAction ActuationAction json:"actuationAction,omitempty"`, though we normally do something like `Actuation ActuationMode json:"actutation,omitempty"` just to avoid overloading our own mental lexers/parsers

pkg/controller/resourceactuation/resourceactuation.go
```diff
@@ -15,14 +15,86 @@
...
+	opv1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/apis/core/v1beta1"
+	opk8s "github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/k8s"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// todo acpana: revisit
```
justinsb, line 34: Yeah, I don't think you need any of these here, I would just return the enum values and switch on them.  It's fine to be case-sensitive for enum values.

acpana: n/a after adding enum values to types

pkg/apis/addtoscheme_cc_ccc_v1beta.go
acpana: I think this is the current pattern for the `iam` schemes in the same folder;

however, with https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1280/files, I think I will undo this change and just rebase onto that since it already solves for adding the CC/ CCCs to the mgr's scheme in kccmanager

pkg/controller/tf/controller.go
```diff
@@ -164,6 +164,28 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (res
 		r.logger.Info("Skipping reconcile as nothing has changed and 0 reconcile period is set", "resource", req.NamespacedName)
 		return reconcile.Result{}, nil
 	}
+
+	cc, ccc, err := resourceactuation.FetchLiveKCCState(ctx, r.mgr.GetClient(), req.NamespacedName)
```
maqiuyujoyce, line 168: We are doing some generic processing of the KRM object through out the reconciliation. Such processing includes applying the default values for KCC owned annotation fields, ensuring the finalizers are properly set, etc. I'm not concerned about the defaulting logic, but I think we'll probably want to ensure the finalizers are set properly: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/tf/controller.go#L311. Finalizers we have are used to help manage the deletion: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/deletiondefender/controller.go#L99

This reminds me of another tricky edge case(s) we need to deal with: What if the defaulting logic changes over versions? This can be between KCC versions (should be rare because it is backwards incompatible) or CRD versions (safer, but still will cause confusions).

E.g. in CC/CCC v1beta1, the default value for `state-into-spec` is `merge`, and in CC/CCC v1beta2, the default value for `state-into-spec` will be `absent`. If the users installs KCC via CC/CCC v1beta1 in two clusters, one reconciling and the other paused, upgrades CC/CCC to v1beta2 at some points, and then needs to unpause the paused cluster, they will suddenly see some behavioral changes.

This use case is something I'm concerned about in the unpausing scenario that we will probably need to think carefully about later.

pkg/controller/resourceactuation/resourceactuation.go
```diff
@@ -15,14 +15,71 @@
...
+}
+
+// DecideReconciliationMode looks at CC and CCC to see if they specify a reconciliationMode
+// - If both CC & CCC specify a reconciliationMode in Namespaced mode, we defer to the CCC's value.
+//   - If only CC specifies a reconciliationMode in Namespaced mode, we defer to the CC's value.
+//
+// - If both CC & CCC specify an reconciliationMode in cluster mode, the CCC specification is irrelevant.
+// - If neither CC nor CCC specify a reconciliationMode, we defer to the default value defined in apis.
+func DecideReconciliationMode(cc opv1beta1.ConfigConnector, ccc opv1beta1.ConfigConnectorContext) opv1beta1.ReconciliationMode {
+	if ccc.Spec.Reconciliation != "" && cc.Spec.Mode == opk8s.NamespacedMode {
```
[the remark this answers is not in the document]
acpana: I did spend some time prior to writing the code thinking about set/ unset values and whether they matter. Here's where I arrived:

Isn't the optional intention encoded as "" ? if a user tries to specify anything other than the allowed Enum values, the api server will reject them. IOW, a user cannot unset this value on purpose. And for all intents and purposes, unset means "" means use default for us here.

pkg/controller/kccmanager/kccmanager_test.go
```diff
@@ -44,7 +43,7 @@ import (
 
 var (
 	// These managers are used to just get the rest.Config since our testmain package's methods are not easily changed
-	// to return a rest.Config
+	// to return a rest.Config.
```
acpana, line 46: I think I gofumpt'd this file; will revert to reduce diff

pkg/controller/resourceactuation/resourceactuation.go
```diff
@@ -15,14 +15,71 @@
...
+		return ccc.Spec.Reconciliation
+	}
+
+	// if no CCC exists, defer to the CC's value.
```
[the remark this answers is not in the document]
acpana: > KCC should probably error out under such an use case. WDYT?

I think we can guard against this in the FetchLiveKCCState and have a separation of responsibility between catching bad KCC states and deciding on the actuationMode/ reconcileMode

pkg/controller/kccmanager/kccmanager.go
```diff
@@ -83,19 +82,19 @@ func New(ctx context.Context, restConfig *rest.Config, config Config) (manager.M
...
-	opts.NewClient = nocache.NoCacheClientFunc
 	opts.BaseContext = func() context.Context {
 		return ctx
 	}
+	if err := addSchemes(opts.Scheme); err != nil {
+		return nil, fmt.Errorf("error adding schemes: %w", err)
+	}
+
+	// only cache CC and CCC resources
+	opts.Cache.ByObject = nocache.ByCCandCCC
```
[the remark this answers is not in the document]
acpana: IIUC, we don't want to slam the api server for these resources, especially since we will attempt to fetch CC/CCCs for every CR's reconciliation that is managed by kcc.

pkg/controller/kccmanager/kccmanager.go
```diff
@@ -31,7 +31,6 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/servicemapping/servicemappingloader"
 	tfprovider "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/tf/provider"
-
```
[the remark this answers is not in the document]
acpana: thanks for the context! I'll try to format these files accordingly. I'd love to use `gci` for this. Maybe that means using it outside the `golangci-lint` uber linter if the default format in gci (which is not quite what we have here) is unacceptable: https://github.com/daixiang0/gci?tab=readme-ov-file#old-style

pkg/controller/tf/controller.go
acpana: Thanks for highlighting this! I think the concerns are well founded.

I moved this code block around and now call `EnsureFinalizers` as part of handling the Paused scenario. 

I think the general idea is that we want to keep reconciling the KRM resource but not actually actuate it onto the cloud provider. So by that logic, all other KRM specific that would happen as part of `sync()`  should probably be factored out so we can apply them in the Paused case too. At first glance, EnsureFinalizers is the only thing that we should definitely include. I'm open to revisit that and keep factoring things out as necessary.

operator/pkg/apis/core/v1beta1/configconnector_types.go
```diff
@@ -40,6 +40,14 @@ type ConfigConnectorSpec struct {
 	// When in namespaced mode, you must create a ConfigConnectorContext object per namespace that you want to enable Config Connector in, and each must set `googleServiceAccount` to specify the Google Service Account to be used to authenticate with Google Cloud APIs for the namespace.
 	//+kubebuilder:validation:Enum=cluster;namespaced
 	Mode string `json:"mode,omitempty"`
+
+	// The reconciliation mode of Config Connector controls how resources are actuated onto the cloud provider.
+	// This can be either 'Reconciling' or 'Paused'. The default is 'Reconciling' where resources get actuated.
+	// In 'Paused', k8s resources are still reconciled with the api server but not actuated onto the cloud provider.
+	// If Config Connector is running in 'namespaced' mode, then the value in ConfigConnectorContext takes precedence.
```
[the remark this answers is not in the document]
maqiuyujoyce: Sounds good. The reason I was asking for confirmation is that when I saw `If Config Connector is running in 'namespaced' mode, then the value in ConfigConnectorContext takes precedence.` It is unclear to me what if value is unset in ConfigConnectorContext when KCC is running in `namespaced` mode. Maybe explicitly say something like "Otherwise, the value in ConfigConnector is used"?

Not a blocking issue.

pkg/controller/kccmanager/kccmanager.go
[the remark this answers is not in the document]
maqiuyujoyce: Gotcha, thank you!

pkg/controller/resourceactuation/resourceactuation.go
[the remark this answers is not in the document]
maqiuyujoyce: Makes sense. Thank you for sharing your thoughts!

I'm not strongly opinionated. Especially this is an enum that `""` is not an available option and we don't plan to allow users to unset this field (though why not?) 
Another use case I found using the nil value to represent the unset state annoying/problematic was when we need to use those types for export. After unmarshaling and marshaling, (I think?) the yaml will be turned from  [1] to [2], and [2] is not an accurate intention of the user, and can't be directly leveraged by the users.

[1]
```
kind: ConfigConnector
metadata:
spec:
 mode: cluster
 googleServiceAccount: "SERVICE_ACCOUNT_NAME@PROJECT_ID.iam.gserviceaccount.com"
```

[2]
```
kind: ConfigConnector
metadata:
spec:
 mode: cluster
 googleServiceAccount: "SERVICE_ACCOUNT_NAME@PROJECT_ID.iam.gserviceaccount.com"
 actuationAction: ""
```

But overall using `""` to represent unset is kinda common in our codebase already, and I'm not very concerned about. I'll let you decide.

pkg/controller/tf/controller.go
maqiuyujoyce: No problem! Most of them can be addressed/improved in the future.

Have you verified that EnsureFinalizers runs properly? I had some problems testing in #1179 so just want to confirm it is indeed working.

maqiuyujoyce (commented):
Overall LGTM! We probably want to rename the field and ensure the controller works locally with the newly added step.

pkg/controller/resourceactuation/resourceactuation.go
```diff
@@ -15,14 +15,71 @@
...
+// FetchLiveKCCState tries to fetch the ConfigConnector (CC) resource or the ConfigConnectorContext (CCC)
+// for the resource's namespace if running in Namespaced mode. It ignores not found errors for CC and CCC when fetching.
+func FetchLiveKCCState(ctx context.Context, c client.Client, resourceNN types.NamespacedName) (opv1beta1.ConfigConnector, opv1beta1.ConfigConnectorContext, error) {
+	var cc opv1beta1.ConfigConnector
+	if err := c.Get(ctx, types.NamespacedName{
+		Name: opk8s.ConfigConnectorAllowedName,
+		Namespace: k8s.SystemNamespace,
+	}, &cc); err != nil {
+		if apierrors.IsNotFound(err) {
+			// if no CC exists, then by definition, KCC cannot be running in namespaced mode;
```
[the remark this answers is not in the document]
acpana: Chatted offline; I think we addressed this in this comment thread too https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1300#discussion_r1512033410

operator/config/crd/bases/core.cnrm.cloud.google.com_configconnectorcontexts.yaml
```diff
@@ -54,6 +54,16 @@ spec:
                 description: The Google Service Account to be used by Config Connector
                   to authenticate with Google Cloud APIs in the associated namespace.
                 type: string
+              reconciliationMode:
```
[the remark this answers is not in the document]
acpana: Renamed as per the latest push

operator/pkg/apis/core/v1beta1/configconnector_types.go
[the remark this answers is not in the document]
acpana: With the most recent rename, I also changed the description to spell out the precedence. Please let me know if it's still not super clear from the api.

pkg/controller/resourceactuation/resourceactuation.go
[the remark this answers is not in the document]
acpana: Let's stick with `""` to mean unset for now and I can come back and change this before the release if there's a strong need for representing/ encoding unset for actuationMode.

pkg/controller/resourceactuation/resourceactuation_test.go
```diff
@@ -18,12 +18,81 @@ import (
...
+			ccc:                   opv1beta1.ConfigConnectorContext{},
+			expectedActuationMode: opv1beta1.Reconciling,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			actualMode := resourceactuation.DecideActuationMode(test.cc, test.ccc)
+			if test.expectedActuationMode != actualMode {
+				t.Errorf("DecideReconciliationMod failed; got %v, want %v", actualMode, test.expectedActuationMode)
```
maqiuyujoyce, line 90: `Decide ActuationMode failed`?

maqiuyujoyce (commented):
LGTM! Will follow up offline to determine the merge order.

acpana:
Immortalizing on the PR some of my manual e2e testing here. With a live GCP project connected to a GKE cluster, I can throw the following on the k8s api server:

* a CC with `actuationMode: Paused`:

```
$ kubectl --namespace=cnrm-system get configconnectors configconnector.core.cnrm.cloud.google.com -oyaml
kind: ConfigConnector
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"core.cnrm.cloud.google.com/v1beta1","kind":"ConfigConnector","metadata":{"annotations":{},"name":"configconnector.core.cnrm.cloud.google.com"},"spec":{"actuationMode":"Paused","mode":"cluster"}}
  creationTimestamp: "2024-03-01T00:47:51Z"
  generation: 7
  managedFields:
  - apiVersion: core.cnrm.cloud.google.com/v1beta1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          .: {}
          f:kubectl.kubernetes.io/last-applied-configuration: {}
      f:spec:
        .: {}
        f:actuationMode: {}
        f:mode: {}
        f:reconciliationMode: {}
    manager: kubectl
    operation: Update
    time: "2024-03-07T19:22:34Z"
  resourceVersion: "30348073"
  uid: 09ea32a7-6c10-4fc1-92ae-31596e832bf7
spec:
  actuationMode: Paused
  mode: cluster
```

* an `ArtifactRegistryRepository` CR:

```
apiVersion: artifactregistry.cnrm.cloud.google.com/v1beta1
kind: ArtifactRegistryRepository
metadata:
  name: artifactregistryrepository-sample-2
  labels:
    label-one: "value-one"
    label-two: value-two
  annotations:
    cnrm.cloud.google.com/project-id: "acpana-kcc"
spec:
  format: DOCKER
  location: us-west1
  description: "this is a description"
```

Running kccmanager via `SERVICE_MAPPING_DIR=config/servicemappings/ go run ./cmd/manager/main.go` :

```
...
{"severity":"info","timestamp":"2024-03-07T19:23:11.486Z","logger":"artifactregistryrepository-controller","msg":"Skipping actuation of resource as actuation mode is \"Paused\"","resource":{"name":"artifactregistryrepository-sample-2","namespace":"default"},"time to next reconciliation":"9m19.447199865s"}
...
```

and importantly `gcloud artifacts repositories list` does not have the new artifact registry. I've been manually testing this flow as well as "updates paused, deletion paused, unpause and see KRM reconciliation reflect through the actuation on resources, etc". Hopefully, all of these cases will be programmatically added via my next PR for golden log manipulation

PR #1302, opened 2024-02-28
chore:ci: bump golangci-lint and add make target

acpana:
It seems that 1.54.1 had one issue with handling non existing VCS information for binaries. 1.54.2 fixes that. I'm also adding the `lint` target as a required for `ready-pr` so devs get an early signal if there's an issue.


justinsb:
Thanks - lgtm

/approve
/lgtm

PR #1303, opened 2024-02-28
logging: log to configured logger when registering controller

justinsb:
This should hopefully allow us to filter these messages in testing,
and it is also a step towards using the more coherent context-aware
logging introduced over the years.


justinsb:
/assign @gemmahou 

Hopefully this gets rid of the "registering controller" messages :-)

gemmahou (approved):

PR #1304, opened 2024-02-28
mockgcp: fix a few more extra-wrappings of errors

justinsb:
Errors from storage Create/Update should not be wrapped.


justinsb:
A few more I spotted.  Hopefully as we get the backlog merged there will be fewer examples of the old code.

/assign @gemmahou

gemmahou (approved):

PR #1306, opened 2024-02-28
chore: avoid errors.Is, replace with IsFoo

justinsb:
errors.Is performs an equality check, which is not normally what we want.

### DIFF
--- a/pkg/controller/lifecyclehandler/handler.go
+++ b/pkg/controller/lifecyclehandler/handler.go
@@ -16,7 +16,6 @@ package lifecyclehandler
 
 import (
 	"context"
-	"errors"
 	"fmt"
 
 	corekccv1alpha1 "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/apis/core/v1alpha1"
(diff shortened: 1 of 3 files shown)


justinsb:
/assign @acpana 

1) I think this is right
2) I think we should avoid errors.Is with anything that isn't a reference to a global variable.  A quick scan didn't find anything outside this function, but maybe double-check me?
3) Is there a linter to detect suspicious use of errors.Is?

justinsb:
(And BTW, duet wrote the majority of the unit test.  I added the special cases for calling e.g. NewTransitiveDependencyNotFoundError

acpana:
/lgtm
/approve

thanks for recognizing and patching this 🥇 ; I think I messed up here 🤦🏼 

2. I checked and this seemed to be isolated in the lifecyclehandler:

```
$ grep -r "errors.Is" . --exclude-dir=third_party --exclude-dir=.tmp --exclude-dir=.build | grep "{}"

./pkg/controller/lifecyclehandler/handler.go:   case errors.Is(err, &k8s.ReferenceNotReadyError{}) || errors.Is(err, &k8s.TransitiveDependencyNotReadyError{}):
./pkg/controller/lifecyclehandler/handler.go:   case errors.Is(err, &k8s.ReferenceNotFoundError{}) || errors.Is(err, &k8s.SecretNotFoundError{}) || errors.Is(err, &k8s.TransitiveDependencyNotFoundError{}):
./pkg/controller/lifecyclehandler/handler.go:   case errors.Is(err, &k8s.KeyInSecretNotFoundError{}):
```

> 3. Is there a linter to detect suspicious use of errors.Is?

3. Hm, not off the top of my head but I will have a look! 🔍  (or maybe I will write one hehe ⌨️ )

PR #1313, opened 2024-02-29
Add http log instructions in mockgcp readme

gemmahou:
### Change description

Upon team's request, I added some instructions for obtaining golden HTTP logs in the mockGCP readme.

### Tests you have done

### DIFF
--- a/mockgcp/README.md
+++ b/mockgcp/README.md
@@ -66,3 +66,24 @@ You can also use `ARTIFACTS=artifacts` to get detailed HTTP logs of the traffic,
 If you also use `E2E_GCP_TARGET=real` you can run against the real (non-mocked) GCP, and easily see what the actual behaviour should be.
 Usually however, this is not necessary; the most common failure mode is that terraform or KCC expects a field to be automatically populated,
 and it normally logs an error like "foo not set" (in this case, simply add that to your mock implementation.)
+
+## Capture HTTP golden logs
+
+Capture HTTP logs against real GCP:
+
+Example command: `E2E_KUBE_TARGET=envtest RUN_E2E=1 E2E_GCP_TARGET=real GOLDEN_REQUEST_CHECKS=1 WRITE_GOLDEN_OUTPUT=1 go test -test.count=1 -timeout 3600s -v ./tests/e2e -run TestAllInSeries/fixtures/test-name`.
+
+Capture HTTP logs against mock GCP:
+
+We can consider getting HTTP logs against mockGCP after implementing the mockGCP service.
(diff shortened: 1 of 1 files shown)


maqiuyujoyce (approved):
Thank you for adding the doc, @gemmahou ! We probably should add a linter for the .md files in the future.

mockgcp/README.md
```diff
@@ -65,4 +65,25 @@ If something is not behaving as you would expect, you should be able to launch a
...
+it's acceptable to capture the golden HTTP logs from the mock service.
+
+Example command: `E2E_KUBE_TARGET=envtest RUN_E2E=1 E2E_GCP_TARGET=mock GOLDEN_REQUEST_CHECKS=1 WRITE_GOLDEN_OUTPUT=1 go test -test.count=1 -timeout 3600s -v ./tests/e2e -run TestAllInSeries/fixtures/test-name`.
+
+A new file named "_http.log" will be created in the test folder during the first time the command is executed, that's our golden HTTP log.
+We will compare new HTTP logs with the golden HTTP log as a part of the mockGCP test if GOLDEN_REQUEST_CHECKS is specified.
+
+To prevent any problems when comparing with the golden logs, it is necessary to replace all generated values in the HTTP log with identical values.
+
+See [here](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/tests/e2e/unified_test.go#L167-L329) to get some ideas.
```
jingyih, line 89: Do we have to manually replace all the generated values in the HTTP log before uploading to Github?
If so, have we considered doing it programmatically either during HTTP log generation or when it gets compared?

gemmahou: The goal is to replace the values during log generation. 
We have already done that in the code I mentioned above. However, resources are different, for a new resource like vertexai, there might be fields that we haven't covered yet. So after we get the http log, we have to check the log and if there's no new random values, then we are good; if there's any, we need to add a new logic to replace them and re-capture http logs.

gemmahou: For example, this [logic](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/tests/e2e/unified_test.go#L287-L289) is added for vertexai.

maqiuyujoyce: @gemmahou @jingyih do you mind making incremental doc changes to reduce confusions?

The eventual goal of the instructions is that users can follow it in a no-brainer way. E.g. `To prevent any problems when comparing with the golden logs, it is necessary to replace all generated values in the HTTP log with identical values.` can be turned into something like
1. Look for generated values [1] in the HTTP log.
2. Manually replace all generated values in the HTTP log with identical values [2].
3. ...

[1] Would be great if there is definition of `generated values` because it is not mentioned anywhere in the doc.
[2] Similarly, would be great if there is definition of `identical values`.

PR #1314, opened 2024-02-29
enh: log err to record event

acpana:
Small follow up from the lint changes to log errors from recording events.


justinsb:
LGTM

/approve
/lgtm

(I assume it won't merge while a draft, if it does .. that'll be interesting too!)

PR #1315, opened 2024-02-29
chore:ci: further lint errors.is

acpana:
Inspired from https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1306, I hand rolled a check to make sure we don't use `errors.Is` on a "empty" error struct.

### DIFF
--- a/scripts/linters/lint_errors_is.sh
+++ b/scripts/linters/lint_errors_is.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+set -o errexit
+set -o nounset
+set -o pipefail
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+cd "${REPO_ROOT}"
+
+if grep -rInH "errors.Is" --include=\*.go --exclude-dir={third_party,.tmp,.build,bin} . | grep "{}"; then
+  printf "\n\nERROR: errors.Is is not used correctly\n"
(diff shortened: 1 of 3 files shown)


scripts/linters/lint_errors_is.sh
```diff
@@ -0,0 +1,26 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+set -o errexit
+set -o nounset
+set -o pipefail
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+cd "${REPO_ROOT}"
+
+if grep -rInH "errors.Is" --include=\*.go --exclude-dir={third_party,.tmp,.build,bin} . | grep "{}"; then
```
acpana, line 21: this is a bit of a hack as it's overly sensitive and will trigger on a comment line for instance; that said, if people run into that as an issue, I will take on upleveling this script;

acpana:
/assign @gemmahou 

Can you PTAL when you have some time 🙏🏼 ?

justinsb:
This is a reasonable approach to detecting the problem, but I think as a project we have a deeper interest in static analysis: https://pkg.go.dev/golang.org/x/tools/go/analysis  So it might be worth writing this in that framework, as a means of evaluating its suitability for "future problems".

By "future problems" , I mean in the direct-implementation layer, we want to be able to reason about the mappings from proto <-> KRM.  Today the PoC does that by constructing a mapping object, so it's all driven by data, we can introspect the data and compare it to reflection etc.

However, this has obvious similarities to the "magic machine" we are trying to get away from.  How do we stop the new "magic machine" from growing indefinitely?

Another approach could be to write go functions for the mappings we need, and then we could statically analyze the go code.  For example, today we might have this:

`Map(&pb.Foo{}, &krm.Foo{}, Spec("f1"), Spec("f2"), Status("s1"), Status("s2"), Ignore("resourceID")`

We could also have this:

```
func MapFooToCloud(in *pb.Foo, out *krm.Foo) {
out.Spec.F1 = in.F1
out.Spec.F2 = in.F2
out.Status.S1 = in.S1
out.Status.S2 = in.S2
map.Ignore(in.ResourceID)
}
```

Maybe we can statically analyze the code and determine "hey, you're not mapping or ignoring this new field F3".  I don't think we need to be perfect, for example I am willing to accept being fooled by `klog.Infof("we should handle %v", in.F3)`

cc @cheftako

ldanielmadariaga:

PR #1316, opened 2024-02-29
Don't use other reconcilers if direct reconciler was requested

justinsb:
This would be surprising to users!


justinsb:
As suggested in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1177#discussion_r1507096336

/assign @cheftako

PR #1323, opened 2024-03-03
chore: Rework deepcopy generators to not require gopath

justinsb:
We can use the upstream deepcopy implementation, eliminate our fork, and move this out of the section that requires GOPATH.

This is more maintainable and feels faster.

- chore: Rework deepcopy generators to not require gopath
- Remove not-generated deepcopy
- Update generated code


cheftako:
Looks good but we need to fix the conflict.

acpana:
Just for my own gratification, I rebased and integrated the changes locally and unfortunately the error in https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/1427 is still there 😢 . Probably have to dig deeper to understand why the new resources (from dcl 1.63 on) are causing this.

pkg/apis/core/v1alpha1/zz_generated.deepcopy.go
yuwenma, line 1: Just out of curiosity: how did you find out this file?

yuwenma (commented):
/lgtm
/approve

PR #1324, opened 2024-03-04
Promote VertexAIEndpoint v1alpha1 to v1beta1 

jingyih:
### Change description

Fixes b/325303446

### Tests you have done

```
$ go test -v -tags=integration ./config/tests/samples/create/ -run-tests vertexai-endpoint-network
$ go test -v -tags=integration ./config/tests/samples/create/ -run-tests vertexai-endpoint-encryptionkey
$ go test -timeout 1200s -v -tags=integration ./pkg/controller/dynamic/ -test.run TestCreateNoChangeUpdateDelete -run-tests vertexaiendpointencryptionkey
$ go test -timeout 1200s -v -tags=integration ./pkg/controller/dynamic/ -test.run TestCreateNoChangeUpdateDelete -run-tests vertexaiendpointnetwork
$ go test -v ./pkg/resourceskeleton
```


justinsb:
IIUC this one is waiting on the observed state changes also?  If so, mind adding WIP?  If not, please let me know!

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_vertexaiendpoints.vertexai.cnrm.cloud.google.com.yaml
```diff
@@ -162,233 +207,669 @@ spec:
...
+                description: ObservedGeneration is the generation of the resource
+                  that was most recently observed by the Config Connector controller.
+                  If this is equal to metadata.generation, then that means that the
+                  current reported status reflects the most recent desired state of
+                  the resource.
+                type: integer
+              observedState:
+                description: The observed state of the underlying GCP resource.
+                properties:
+                  createTime:
```
justinsb, line 220: Mentioned this on the other PR, but if it's easy we should consider not including createTime / etags in observedState _yet_.  Etags in particular I worry is going to be laggy and not very useful.  createTime I guess I don't really have a concrete objection to, just a question of whether we should default-yes or default-no.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_vertexaiendpoints.vertexai.cnrm.cloud.google.com.yaml
```diff
@@ -162,233 +207,669 @@ spec:
...
+                          type: string
+                        enableAccessLogging:
+                          description: These logs are like standard server access
+                            logs, containing information like timestamp and latency
+                            for each prediction request. Note that Stackdriver logs
+                            may incur a cost, especially if your project receives
+                            prediction requests at a high queries per second rate
+                            (QPS). Estimate your costs before enabling this option.
+                          type: boolean
+                        enableContainerLogging:
```
justinsb, line 374: Similarly, I think this field (and a few others) is input-only.  Do we want to wait till we understand it better before exposing it here?  There is some use-case for the actual state of the object if we were unable to apply changes, but I'm not sure.

justinsb:
LGTM overall, some questions about whether we should expose _all_ the observedState immediately (thereby committing us), vs exposing the clearer ones first, and following up to add more if needed (keeping our options open)

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_vertexaiendpoints.vertexai.cnrm.cloud.google.com.yaml
jingyih: Removing "etag" sounds good to me.
I think we shipped `VertexAIDataset` with "createTime", we could do the same for this resource.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_vertexaiendpoints.vertexai.cnrm.cloud.google.com.yaml
jingyih: Do you mean the field is output-only? The API generates them but users cannot configure them. I am not sure I fully understood your concern.

maqiuyujoyce: I dug into the API a bit, and looks like while [`deployedModels` is an "output-only" field](https://cloud.google.com/vertex-ai/docs/reference/rest/v1beta1/projects.locations.endpoints#resource:-endpoint), there is a method to ["mutate" each individual deployed model](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.endpoints/mutateDeployedModel). There are also methods to [add](https://cloud.google.com/vertex-ai/docs/reference/rest/v1beta1/projects.locations.endpoints/deployModel#google.cloud.aiplatform.v1beta1.EndpointService.DeployModel) or [remove](https://cloud.google.com/vertex-ai/docs/reference/rest/v1beta1/projects.locations.endpoints/undeployModel) deployed models in this list.

Looking at the basic use case here: https://cloud.google.com/vertex-ai/docs/general/deployment#aiplatform_deploy_model_custom_trained_model_sample-drest, the endpoints are used to deploy the models, and I think the deployed model should be a separate resource instead of being captured here.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/vertexai/resource_vertex_ai_endpoint.go
```diff
@@ -1004,5 +1010,91 @@ func expandVertexAIEndpointEncryptionSpecKmsKeyName(v interface{}, d tpgresource
...
+	projectID, err := getProjectID(config, projectNumber, userAgent)
+	if err != nil {
+		return nil, err
+	}
+	transformed := fmt.Sprintf("projects/%s/global/networks/%s", projectID, networkName)
+	return transformed, nil
+}
+
+func getProjectNumber(config *transport_tpg.Config, project, userAgent string) (string, error) {
+	getProjectCall := config.NewResourceManagerClient(userAgent).Projects.Get(project)
```
maqiuyujoyce, line 1073: Do you know if it is possible to reference a network in a different project? I'm curious if we should add warnings about permissions. (Though we can always improve the doc/workaround the issue incrementally.)

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/vertexai/resource_vertex_ai_endpoint.go
```diff
@@ -954,6 +955,11 @@ func flattenVertexAIEndpointEncryptionSpecKmsKeyName(v interface{}, d *schema.Re
 }
 
 func flattenVertexAIEndpointNetwork(v interface{}, d *schema.ResourceData, config *transport_tpg.Config) interface{} {
+	// This API uses a unique formatting of network different from the standard self_link URI.
+	// We need to replace the project_number with project_id when flattening the value of network field.
+	if transformed, err := transformProjectNumberToProjectIDInNetwork(v, d, config); err == nil {
```
maqiuyujoyce, line 960: Thank you for customizing the underlying logic!

maqiuyujoyce (commented):
Overall it looks good, but I wondered if we should dive a bit deeper about the deploy model scenarios.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/vertexai/resource_vertex_ai_endpoint.go
jingyih: In theory it is possible as long as the GSA has the `resourcemanager.project.get` permission. If users gets confused by it, we could add a warning about permission later.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_vertexaiendpoints.vertexai.cnrm.cloud.google.com.yaml
jingyih: > and I think the deployed model should be a separate resource instead of being captured here.

Could you clarify? Are you suggesting that we should remove `deployedModels` from the `status.observedState`?

jingyih:
> Overall it looks good, but I wondered if we should dive a bit deeper about the deploy model scenarios.

`https://{service-endpoint}/v1beta1/{endpoint}:deployModel` was not implemented in Terraform(v4 or v5). Users need to call the API directly. It is something we can improve when we migrate this resource implementation to SciFi in the future.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_vertexaiendpoints.vertexai.cnrm.cloud.google.com.yaml
justinsb: The concern is if anyone else writes to the object (possibly including internal clients), etags will be out of date until our next reconcile.  We don't want to encourage people to write to the object directly, and that's the primary use for etag.  etag also is intended to just be an opaque "cookie" for optimistic concurrency.

OTOH, if/when we implement the "LIST" optimization over polling each object every 10 minutes, maybe this field will be useful.

But ... it seems like we don't need it today, and we only have speculative uses for it.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_vertexaiendpoints.vertexai.cnrm.cloud.google.com.yaml
justinsb: Ah, yes, I think I understand what's going on now.  So I did mean that `enableAccessLogging` and `enableContainerLogging` looked like input-only fields.   It's because the deployed models are more list a subresource, and this resource isn't following our normal API design.

I don't know if we should actually split these into their own resources (as we normally would for subresources), but they are _not_ really output-only fields, so they likely don't belong in status.

I also don't think this resource is useful without being able to deploy Models, so I don't think it makes sense to merge this resource until we've figured this out... which is a bummer in terms of timing.

maqiuyujoyce: > Are you suggesting that we should remove deployedModels from the status.observedState?

Yes, I think it's a fine-grained resource that should not be represented here.

> I don't know if we should actually split these into their own resources (as we normally would for subresources), but they are not really output-only fields, so they likely don't belong in status.

I consider "resource deployment" a reasonable use case to be represented individually and declaratively. We should reach out to the API team to understand more details. We probably have to "redo the API design" for them in the KCC resource.

> I also don't think this resource is useful without being able to deploy Models, so I don't think it makes sense to merge this resource until we've figured this out... 

Agreed. But it's always doable to release the resource without the uncertain fields (i.e. we can ignore the tricky deployedModels field) and get feedback from users while we are finalizing the design.

maqiuyujoyce:
> > Overall it looks good, but I wondered if we should dive a bit deeper about the deploy model scenarios.
> 
> `https://{service-endpoint}/v1beta1/{endpoint}:deployModel` was not implemented in Terraform(v4 or v5). Users need to call the Vertex AI API directly. It is something we can improve when we migrate this resource implementation to SciFi in the future.

I think we should get clarifications from our users to understand whether they expect to call the API directly to deploy the models or not.

jingyih:
Added a commit to ignore "deployedModels" and "etag". Rebased PR to resolve conflict.

jingyih:
Added a commit to ignore "deployedModels" and "etag". Rebased PR to resolve conflict.

maqiuyujoyce (commented):
/lgtm

@justinsb do you have any other feedback?

@jingyih let me know if you need the approval.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_vertexaiendpoints.vertexai.cnrm.cloud.google.com.yaml
maqiuyujoyce: #1637

config/servicemappings/vertexai.yaml
```diff
@@ -61,3 +61,54 @@ spec:
             kind: KMSCryptoKey
             version: v1beta1
             group: kms.cnrm.cloud.google.com
+    - name: google_vertex_ai_endpoint
+      kind: VertexAIEndpoint
+      idTemplate: "projects/{{project}}/locations/{{location}}/endpoints/{{name}}"
+      idTemplateCanBeUsedToMatchResourceName: false
+      resourceAvailableInAssetInventory: true
+      v1alpha1ToV1beta1: true
+      storageVersion: v1alpha1
```
justinsb, line 70: Nit: maybe should be v1beta1?  We should figure out how we eventually stop shipping v1alpha1 CRD versions!

pkg/test/resourcefixture/testdata/basic/vertexai/v1beta1/vertexaiendpoint/vertexaiendpointencryptionkey/update.yaml
```diff
@@ -0,0 +1,30 @@
...
+kind: VertexAIEndpoint
+metadata:
+  name: vertexaiendpoint-${uniqueId}
+  labels:
+    label-one: "value-one"
+spec:
+  displayName: "vertexai-endpoint-encryptionkey-updated"
+  description: "example vertex AI endpoint with encryption key"
+  location: us-central1
+  region: us-central1
```
justinsb, line 25: Missed this before, but why do we have location & region?

justinsb:
Looks good, except I think having both location & region is a blocker to merge unless there's a reason the two fields are different.

config/servicemappings/vertexai.yaml
jingyih: When we promote a resource from alpha to beta, we set `v1alpha1ToV1beta1` to `true` and `storageVersion` to `v1alpha1` according to the [New Resource Guide](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/README.NewResourceFromTerraform.md#add-the-resource-config-to-service-mappings).
@maqiuyujoyce Do we have a process of eventually stop shipping v1alpha1 CRDs?

pkg/test/resourcefixture/testdata/basic/vertexai/v1beta1/vertexaiendpoint/vertexaiendpointencryptionkey/update.yaml
jingyih: Now I remembered, I tried to ignore the field before. But it looks like it is required by the Terraform (although not directly used in the `resource_vertex_ai_endpoint.go`). Ignoring the field would lead to the following error.

Maybe we need to patch the Terraform code to use "region" instead of "location".

```
{"severity":"info","timestamp":"2024-04-29T18:17:30.668Z","msg":"resource is not ready","kind":"VertexAIEndpoint","name":"sdliqpjzcxxdcneitslqvrjxztykvlg","conditions":[{"lastTransitionTime":"2024-04-29T18:10:22Z",
"message":"Update call failed: error fetching live state: error reading underlying resource: summary: Cannot determine region: set in this resource, or set provider-level 'region' or 'zone'.","reason":"UpdateFailed
","status":"False","type":"Ready"}]}  
```

jingyih:
> Looks good, except I think having both location & region is a blocker to merge unless there's a reason the two fields are different.

@justinsb Thank you for catching the location vs. region issue. I found the issue 2 months ago when I was working on the initial implementation, but forgot about it later. I should have added a TODO for myself.

Uploaded a commit to patch Terraform to use "region", so we can ignore the "location" in KCC.

Using "region" seems to be consistent with Vertex AI [dataset](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/0119895cdb7525ecc5d815062a45d4aef0b4bab6/third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/vertexai/resource_vertex_ai_dataset.go#L83) and [index](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/0119895cdb7525ecc5d815062a45d4aef0b4bab6/third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/vertexai/resource_vertex_ai_index.go#L202).

@maqiuyujoyce Could you please take a look if this makes sense? Thank you!

pkg/resourceskeleton/testdata/asset-skeleton.yaml
```diff
@@ -1184,3 +1184,19 @@
...
+      - organizations/1234567890
+    name: //aiplatform.googleapis.com/projects/kcc-test/locations/us-central1/endpoints/vertexaiendpoints-test
+    asset_type: aiplatform.googleapis.com/Endpoint
+  expectedSkeleton:
+    apiVersion: vertexai.cnrm.cloud.google.com/v1beta1
+    kind: VertexAIEndpoint
+    metadata:
+      name: vertexaiendpoints-test
+    spec:
+      location: us-central1
```
maqiuyujoyce, line 1199: Clean up?

pkg/resourceskeleton/testdata/uri-skeleton.yaml
```diff
@@ -1117,3 +1117,14 @@
...
   URI: "https://us-central1-aiplatform.googleapis.com/v1beta1/projects/kcc-test/locations/us-central1/datasets/1234567890?alt=json"
+- ExpectedSkeleton:
+    apiVersion: vertexai.cnrm.cloud.google.com/v1beta1
+    kind: VertexAIEndpoint
+    metadata:
+      name: vertexaiendpoint-test
+    spec:
+      projectRef:
+    external: kcc-test
+    location: us-central1
```
maqiuyujoyce, line 1128: ditto

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/vertexai/resource_vertex_ai_endpoint.go
```diff
@@ -97,7 +98,7 @@ func ResourceVertexAIEndpoint() *schema.Resource {
...
+				Description: `The full name of the Google Compute Engine [network](https://cloud.google.com//compute/docs/networks-and-firewalls#networks) to which the Endpoint should be peered. Private services access must already be configured for the network. If left unspecified, the Endpoint is not peered with any network. Only one of the fields, network or enable_private_service_connect, can be set. [Format](https://cloud.google.com/compute/docs/reference/rest/v1/networks/insert): 'projects/{project_id}/global/networks/{network}'.`,
 			},
 			"region": {
 				Type:        schema.TypeString,
```
maqiuyujoyce, line 104: Nit: Maybe explicitly mark this field required?

config/servicemappings/vertexai.yaml
maqiuyujoyce: We do have a proposal but further work is needed to implement it. Will follow up offline about it.

maqiuyujoyce (commented):
/lgtm

except for a couple of small issues.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/vertexai/resource_vertex_ai_endpoint.go
jingyih: Sorry which field are you referring to?

pkg/resourceskeleton/testdata/uri-skeleton.yaml
jingyih: Done

pkg/resourceskeleton/testdata/asset-skeleton.yaml
jingyih: Done

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/vertexai/resource_vertex_ai_endpoint.go
maqiuyujoyce: Gotcha. Those fields are marked as required in KCC due to our own schema generation logic: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/crd/crdgeneration/tf2crdgeneration.go#L289

In practice they are required in KCC, but TF can have the global `region` configured, so they are optional within the resource. https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/provider_reference.html

maqiuyujoyce:
/lgtm
/approve

/hold
in case @justinsb has any follow up comments.

justinsb:
LGTM, thank you!

/hold cancel

PR #1335, opened 2024-03-08
feat: introduce paused event

acpana:
todo


acpana:
@justinsb PTAL when you have a chance

justinsb:
I feel like we don't understand pause enough yet to make a very-likely-to-be-correct call on how to surface this in each object (if at all)

acpana:
> I feel like we don't understand pause enough yet to make a very-likely-to-be-correct call on how to surface this in each object (if at all)

That's fair! Without writing some event type in the case of pause, I worried that this would be more confusing to customers:

```
$ kubectl get ArtifactRegistryRepository
NAME                                  AGE     READY   STATUS   STATUS AGE
artifactregistryrepository-sample     48d                        48d
artifactregistryrepository-sample-2   4d18h                      4d17h
```

Ok to park this for now.

justinsb:
> I worried that this would be more confusing to customers...

I agree, and see the value, but I'd rather get some customer feedback first (and try it out myself in a real environment).

pkg/k8s/constants.go
```diff
@@ -35,6 +35,8 @@ const (
 	TimeToLeaseRenewal                   = 20 * time.Minute
 	MeanReconcileReenqueuePeriod         = 10 * time.Minute
 	JitterFactor                         = 2.0
+	Paused                               = "Paused"
+	PausedMessage                        = "The resource is up to date but actuation onto the cloud provider is Paused"
```
maqiuyujoyce, line 39: I personally think this paused message may cause some potential confusion ("actuation onto the cloud provider" may not be a very common concept) though I don't have a better suggestion. Though we can/should always explain it in our public doc.

@jingyih @gemmahou do you have any thoughts on it?

pkg/controller/tf/controller.go
```diff
@@ -199,11 +199,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (res
...
 		}
 
-		// add finalizers for deletion defender
 		if resource.GetDeletionTimestamp().IsZero() {
+			// add finalizers for deletion defender
 			if err := r.EnsureFinalizers(ctx, resource.Original, &resource.Resource, k8s.ControllerFinalizerName, k8s.DeletionDefenderFinalizerName); err != nil {
 				return reconcile.Result{}, err
 			}
+
+			if err := r.HandlePaused(ctx, resource.Original); err != nil {
```
maqiuyujoyce, line 208: As you put HandlePaused() in the if branch to check whether this is not for deletion - how will deletion work when it is paused? Taking a brief look at the code, iiuc, it just logs `Skipping actuation ...` and continues the re-reconciliation?

This is orthogonal to this PR though I wondered if I've missed anything about the deletion story.

acpana: This is a good point -- if we bill pause as "reconciliation of k8s resources still happens BUT not actuation" then there's more work needed here to implement that promise.

acpana: ( deleted previous comment as I had misunderstood the situation here )

sorry @maqiuyujoyce I may be missing something obvious but what is the concern for deletion when the resource is paused?

acpana:
This is on the backburner for now. We can revisit if needed.

PR #1341, opened 2024-03-09
Increase the timeout for golangci-lint to 10m

maqiuyujoyce:
### Change description

`make lint` kept timing out when I ran it locally so I increased the timeout to 10m.

### Tests you have done

`make lint` finished without timing out.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,7 +98,7 @@ lint:
 	docker run --rm -v $(shell pwd):/app \
 		-v ${GOLANGCI_LINT_CACHE}:/root/.cache/golangci-lint \
 		-w /app golangci/golangci-lint:${GOLANGCI_LINT_VERSION}-alpine \
-		golangci-lint run -v
+		golangci-lint run -v --timeout=10m
 
 # Run go vet against code
 .PHONY: vet


Makefile
```diff
@@ -98,7 +98,7 @@ lint:
 	docker run --rm -v $(shell pwd):/app \
 		-v ${GOLANGCI_LINT_CACHE}:/root/.cache/golangci-lint \
 		-w /app golangci/golangci-lint:${GOLANGCI_LINT_VERSION}-alpine \
-		golangci-lint run -v
```
acpana, line 101: hmm, I think walter already raised made this exact change byt curious why GitHub is not displaying it 🤔 

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/e7735cc9b7e35671a2aac7f862de23b67e95fcee/Makefile#L101

Anyhow, the better way for us to do this is probably modify this line instead:

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/e7735cc9b7e35671a2aac7f862de23b67e95fcee/.golangci.yaml#L16

maqiuyujoyce: Perfect! I'll abandon this change!

> but curious why GitHub is not displaying it

Probably because I'm not synced to HEAD locally and edited a different file.

PR #1342, opened 2024-03-09
Stop merging sensitive fields in SQLInstance and ComputeBackendService

maqiuyujoyce:
### Change description

Fixes b/328854103

### Tests you have done

- [N/A] Run `make ready-pr` to ensure this PR is ready for review.
- [X] Perform necessary E2E testing for changed resources.

Test details can be found at #1343 .

PR #1343, opened 2024-03-09
[WIP] Test SQLInstance after sensitive fields are not populated.

maqiuyujoyce:
This is to test #1342 only. Should be abandoned after #1342 is merged.
Ran the following command and verified in the updated unstructured that the rootPassword no longer existed.
```
go test -timeout 1200s -v -tags=integration ./pkg/controller/dynamic/ -test.run TestCreateNoChangeUpdateDelete -run-tests mysqlinstance
```


maqiuyujoyce:
Abandoning because this PR is for temporary test only.

To support a reusable test scenario, we should add a test step to verify sensitive fields are not leaked in the mutable-but-unreadable annotation.

PR #1290, opened 2024-02-27
Create step-wise test for CertificateMapEntry

justinsb:
The step-wise test runner runs more complicated operations in a linear order.  We use it to verify that the real GCP certificatemanager implementation prevents out-of-order deletion of resources.  And then we enhance the mock to do the same thing.  This should mean we don't have flakes on certificatemapentry any more.

- tests: create step-by-step test runner
- tests: create stepwise test for CertificateMapEntry
- mockgcp: update certificatemanager mock to be more accurate

PR #1347, opened 2024-03-12
test: basic golden test for pause

acpana:
basic smoke test for pause; this will fail until DCL controller also starts using the `actuationMode` field

### DIFF
--- a/config/tests/samples/create/samples.go
+++ b/config/tests/samples/create/samples.go
@@ -94,6 +94,10 @@ type CreateDeleteTestOptions struct { //nolint:revive
 
 	// CleanupResources is true if we should delete resources when we are done
 	CleanupResources bool
+
+	// SoftDelete true means check that the delete timestamp is set for a resource
+	// and considered it deleted.
+	SoftDelete bool
 }
 
 func RunCreateDeleteTest(t *Harness, opt CreateDeleteTestOptions) {
@@ -120,7 +124,7 @@ func RunCreateDeleteTest(t *Harness, opt CreateDeleteTestOptions) {
 
 	// Clean up resources on success if CleanupResources flag is true
 	if opt.CleanupResources {
-		DeleteResources(t, opt.Create)
+		DeleteResources(t, opt)
 	}
 }
 
@@ -152,9 +156,13 @@ func waitForReadySingleResource(t *Harness, wg *sync.WaitGroup, u *unstructured.
 		if u.GetKind() == "Secret" { // If unstruct is a Secret and it is found on the API server, then the Secret is ready
 			return true, nil
 		}
-		if u.Object["status"] == nil ||
-			u.Object["status"].(map[string]interface{})["conditions"] == nil { // status not ready
-			logger.Info("resource does not yet have status or conditions", "kind", u.GetKind(), "name", u.GetName())
+		if u.Object["status"] == nil {
+			logger.Info("resource does not yet have status", "kind", u.GetKind(), "name", u.GetName())
+			return false, nil
+		}
+
+		if u.Object["status"].(map[string]interface{})["conditions"] == nil {
+			logger.Info("resource does not yet have conditions", "kind", u.GetKind(), "name", u.GetName())
 			return false, nil
 		}
 		objectStatus := dynamic.GetObjectStatus(t.T, u)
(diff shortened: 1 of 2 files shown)


tests/e2e/unified_test.go
```diff
@@ -81,7 +83,33 @@ func TestAllInSeries(t *testing.T) {
 		}
 	})
 
-	t.Run("fixtures", func(t *testing.T) {
+	testFixturesInSeries(ctx, t, "AllInSeries", false, cancel)
+}
+
+// TestPauseInSeries is a basic smoke test to prove that if CC pauses actuation of resources
+// via the actuationMode field, then resources are not actuated onto the cloud provider.
+// The current test is to make sure that POST requests are not recorded as HTTP events.
```
acpana, line 91: More complex scenarios like "pause then unpause" or "pause" in a specific namespace should follow. They can use some of the step ideas from #1290

tests/e2e/unified_test.go
```diff
@@ -337,8 +378,35 @@ func TestAllInSeries(t *testing.T) {
 	cancel()
 }
 
+
+func mustNotPost(t *testing.T, got string, normalizers ...func(s string) string) {
```
justinsb, line 382: Q: Why only POST?  Do the controllers make any requests to GCP when paused?

config/tests/samples/create/samples.go
```diff
@@ -94,6 +94,10 @@ type CreateDeleteTestOptions struct { //nolint:revive
 
 	// CleanupResources is true if we should delete resources when we are done
 	CleanupResources bool
+
+	// SoftDelete true means check that the delete timestamp is set for a resource
+	// and considered it deleted.
+	SoftDelete bool
```
justinsb, line 100: Maybe "NoWaitForDelete" ?

config/tests/samples/create/samples.go
```diff
@@ -207,12 +216,13 @@ func DeleteResources(t *Harness, unstructs []*unstructured.Unstructured) {
 	var wg sync.WaitGroup
```
justinsb, line 216: In the SoftDeleteCase, I think the deletion timestamp will always be set, should we just skip `waitForDeleteToComplete`?  (This also motivates the potentially simpler name `WaitForDelete: false`)

tests/e2e/unified_test.go
```diff
@@ -81,7 +83,33 @@ func TestAllInSeries(t *testing.T) {
 		}
 	})
 
-	t.Run("fixtures", func(t *testing.T) {
+	testFixturesInSeries(ctx, t, "AllInSeries", false, cancel)
```
justinsb, line 86: Not a blocker, but I think it's an error that we think about samples & fixtures so differently.  It didn't matter when we were just calling RunCreateDeleteTest, but over time we've added to fixtures but not to samples (OK, it was me :-) )!  Maybe we preload the samples and fixtures into a common data structure, and then have a func to load the full data if needed in each test.  (parameterizing the project-id requires starting mockgcp, that's what we're trying to avoid doing at top level)

tests/e2e/unified_test.go
```diff
@@ -81,7 +83,33 @@ func TestAllInSeries(t *testing.T) {
...
+}
+
+func testFixturesInSeries(ctx context.Context, t *testing.T, testName string, testPause bool, cancel context.CancelFunc) {
+	t.Helper()
+
+	if os.Getenv("RUN_E2E") == "" {
+		t.Skip("RUN_E2E not set; skipping")
+	}
+	if testPause && os.Getenv("GOLDEN_REQUEST_CHECKS") == "" {
+		t.Skip("GOLDEN_REQUEST_CHECKS not set; skipping as this test relies on the golden files.")
```
justinsb, line 109: Fine for now, but maybe we should always record the HTTP traffic, and use GOLDEN_REQUEST_CHECKS only to control if we do the CompareGoldenFile call.

tests/e2e/unified_test.go
acpana: that's a good point; I figured we may fetch the live state in places but let me explore more and get back to you.

acpana: In my further testing I didn't see any issues with GETs

tests/e2e/unified_test.go
acpana: I think that may require some refactoring below but I can do so in a follow up -- want to focus more on the main PR for pause for now.

config/tests/samples/create/samples.go
```diff
@@ -94,6 +94,9 @@ type CreateDeleteTestOptions struct { //nolint:revive
 
 	// CleanupResources is true if we should delete resources when we are done
 	CleanupResources bool
+
+	// SkipWaitForDelete true means that we don't wait to query that a resource has been deleted.
+	SkipWaitForDelete bool
```
justinsb, line 99: I like this name 👍

tests/e2e/unified_test.go
justinsb: Cool - I needed the step test in #1360 (and that one is a 1.114 blocker I believe), so hopefully we'll some of the stepwise test infrastructure in today!

justinsb (commented):
Looks good!

justinsb:
LGTM

/approve
/lgtm

/hold just in case you're not ready for it to merge, but I think it looks good!

acpana:
/hold cancel

thanks for reviewing Justin! 🥇; the Pause test is not run as part of CI yet but I will merge this in and carry over that in #1348

PR #1348, opened 2024-03-12
feat: more controllers supporting pause

acpana:
This patch extends the Pause feature to all other controllers that are responsible for actuating resources onto a cloud provider.

We probably don't want to run a `pause` test for every PR so just including this commit for now here. Finally, there's a `cherry-picked` commit for fixing an issue w a missing golden log file.

### DIFF
--- a/pkg/controller/dcl/controller.go
+++ b/pkg/controller/dcl/controller.go
@@ -223,7 +223,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (res 
 		}
 
 		if resource.GetDeletionTimestamp().IsZero() {
-			// add finalizers for deletion defender
+			// add finalizers for deletion defender to make sure we don't delete cloud provider resources when uninstalling
 			if err := r.EnsureFinalizers(ctx, resource.Original, &resource.Resource, k8s.ControllerFinalizerName, k8s.DeletionDefenderFinalizerName); err != nil {
 				return reconcile.Result{}, err
 			}
--- a/pkg/controller/dcl/controller.go
+++ b/pkg/controller/dcl/controller.go
@@ -21,6 +21,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/apis/core/v1beta1"
 	corekccv1alpha1 "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/apis/core/v1alpha1"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/jitter"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/lifecyclehandler"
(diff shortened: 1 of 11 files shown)


acpana:
`test: run the paused test` is actually green except for the `serviceusage.cnrm.cloud.google.com/service` resource fixture which doesn't actually support golden log checking. Will propose a patch fix for that in a separate PR and cherry pick it here to increase confidence.

pkg/controller/dcl/controller.go
```diff
@@ -205,6 +206,39 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (res
 	if err := r.applyChangesForBackwardsCompatibility(ctx, resource); err != nil {
 		return reconcile.Result{}, fmt.Errorf("error applying changes to resource '%v' for backwards compatibility: %w", k8s.GetNamespacedName(resource), err)
 	}
+
+	cc, ccc, err := resourceactuation.FetchLiveKCCState(ctx, r.mgr.GetClient(), req.NamespacedName)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	am := resourceactuation.DecideActuationMode(cc, ccc)
```
justinsb, line 215: Nit (for the future) - it may be easier to return a single object with the CC and CCC as member fields, and then you can easily have that object support `DecideActuationMode()`

pkg/controller/dcl/controller.go
```diff
@@ -205,6 +206,39 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (res
...
+			return reconcile.Result{}, err
+		}
+
+		if resource.GetDeletionTimestamp().IsZero() {
+			// add finalizers for deletion defender
+			if err := r.EnsureFinalizers(ctx, resource.Original, &resource.Resource, k8s.ControllerFinalizerName, k8s.DeletionDefenderFinalizerName); err != nil {
+				return reconcile.Result{}, err
+			}
+
+			if err := r.HandlePaused(ctx, resource.Original); err != nil {
```
justinsb, line 231: Yeah, this case is going to be particularly surprising when we're paused

pkg/controller/dcl/controller.go
```diff
@@ -205,6 +206,39 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (res
...
+	case v1beta1.Reconciling:
+		r.logger.V(2).Info("Actuating a resource as actuation mode is \"Reconciling\"", "resource", req.NamespacedName)
+	case v1beta1.Paused:
+		jitteredPeriod, err := jitter.GenerateJitteredReenqueuePeriod(r.schemaRef.GVK, nil, r.converter.MetadataLoader, u)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		if resource.GetDeletionTimestamp().IsZero() {
+			// add finalizers for deletion defender
```
justinsb, line 226: I'm confused as to why we need to touch finalizers here - can you explain?  (Ideally by expanding the comment)

pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -223,7 +247,6 @@ func (r *reconcileContext) doReconcile(ctx context.Context, u *unstructured.Unst
 		}
 		return false, r.handleUpdateFailed(ctx, u, err)
 	}
-	k8s.EnsureFinalizers(u, k8s.ControllerFinalizerName, k8s.DeletionDefenderFinalizerName)
```
justinsb, line 226: I'm now even more confused :-)  I agree this seems awfully late to be adding finalizers, but better late than never?

justinsb:
A few nits / questions, but generally LGTM

acpana:
/hold

want to make sure commits are well organized

pkg/controller/dcl/controller.go
acpana: I *think* the issue here would be in an uninstallation for KCC case where we don't want the underlying resources on GCP / the cloud provider to go away. So I updated that comment to reflect my understanding.

justinsb:
> We probably don't want to run a pause test for every PR 

I think if we had a scripted test that covers pause in CC and pause in CCC that would be sufficient and we would want to run that on every PR; agree we don't want to run it for every resource on every PR (this is what I tried to do in #1366).  We should run it in periodics though :-)

pkg/controller/direct/directbase/directbase_controller.go
justinsb: I don't think we should be removing this line, but not a blocker as this controller is feature-flagged out

justinsb:
LGTM, two follow-on items but neither is a blocker:

1. I think we need to re-add the k8s.EnsureFinalizers line in the direct controller
2. I think we need to add a watch on the CC / CCC or otherwise make sure we reconcile all objects again.  But this is pretty interesting, as we likely want to be strategic e.g. soft-start,prioritize resources that have never been reconciled.

Thanks for getting this done!

/approve
/lgtm

justinsb:
/approve
/lgtm

So just the second follow-on AI :-)

PR #1349, opened 2024-03-12
vcr: Apply hook in  vcr recorder

gemmahou:
### Change description

ref: https://github.com/dnaeon/go-vcr?tab=readme-ov-file#hooks

The hook will replace values in request/response pairs before we save them to file. 

Note: Applying hook to replace data that we don't want to expose to public. Some data, like timestamp, or id, does not contain sensitive information, so it's fine to leave them for now.

If we can migrate VCR testing as a part of prow job, and save cassette file to GCS bucket, then we may not need hooks since it's a private storage.

### Tests you have done

### DIFF
--- a/config/tests/samples/create/harness.go
+++ b/config/tests/samples/create/harness.go
@@ -24,14 +24,13 @@ import (
 	"testing"
 	"time"
 
-	"gopkg.in/dnaeon/go-vcr.v3/recorder"
-
 	"github.com/go-logr/logr"
 	transport_tpg "github.com/hashicorp/terraform-provider-google-beta/google-beta/transport"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 	cloudresourcemanagerv1 "google.golang.org/api/cloudresourcemanager/v1"
 	"google.golang.org/api/option"
+	"gopkg.in/dnaeon/go-vcr.v3/recorder"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
@@ -326,17 +325,10 @@ func NewHarness(ctx context.Context, t *testing.T) *Harness {
 			}
 			r, err := recorder.NewWithOptions(opts)
 			if err != nil {
-<<<<<<< HEAD
 				t.Fatalf("[VCR] Failed create vcr recorder: %v", err)
 			}
 			h.VCRRecorder = r
 			ret = &http.Client{Transport: h.VCRRecorder}
-=======
-				t.Fatalf("creat vcr recorder failed: %v", err)
-			}
-			h.Rec = r
-			ret = &http.Client{Transport: h.Rec}
->>>>>>> e848f5336 (add hook)
 			return ret
 		}
 	} else {
--- a/config/tests/samples/create/harness.go
+++ b/config/tests/samples/create/harness.go
@@ -326,10 +326,17 @@ func NewHarness(ctx context.Context, t *testing.T) *Harness {
 			}
 			r, err := recorder.NewWithOptions(opts)
 			if err != nil {
+<<<<<<< HEAD
 				t.Fatalf("[VCR] Failed create vcr recorder: %v", err)
 			}
 			h.VCRRecorder = r
 			ret = &http.Client{Transport: h.VCRRecorder}
+=======
+				t.Fatalf("creat vcr recorder failed: %v", err)
+			}
+			h.Rec = r
+			ret = &http.Client{Transport: h.Rec}
+>>>>>>> e848f5336 (add hook)
 			return ret
 		}
 	} else {
--- a/tests/e2e/unified_test.go
+++ b/tests/e2e/unified_test.go
@@ -16,7 +16,11 @@ package e2e
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"gopkg.in/dnaeon/go-vcr.v3/cassette"
+	"gopkg.in/dnaeon/go-vcr.v3/recorder"
+	"k8s.io/klog/v2"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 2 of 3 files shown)


config/tests/samples/create/harness.go
```diff
@@ -325,7 +325,10 @@ func NewHarness(ctx context.Context, t *testing.T) *Harness {
 				Mode:          recorder.ModeRecordOnly,
 				RealTransport: ret.Transport,
 			}
-			r, _ := recorder.NewWithOptions(opts)
+			r, err := recorder.NewWithOptions(opts)
+			if err != nil {
+				t.Fatalf("creat vcr recorder failed: %v", err)
```
justinsb, line 330: Nit: typo in create

tests/e2e/unified_test.go
```diff
@@ -122,12 +126,60 @@ func TestAllInSeries(t *testing.T) {
...
+					vcrRecorder := h.Rec
 					defer func() {
-						err := h.Rec.Stop()
+						err := vcrRecorder.Stop()
 						if err != nil {
 							t.Errorf("[VCR] Failed stop vcr recorder: %v", err)
 						}
 					}()
+
+					hook := func(i *cassette.Interaction) error {
```
justinsb, line 137: Cool - this looks like what I was looking for in my comments on the other PR.  Can you reverse these also?  The other option is to "fix" {uniqueID} when running in VCR mode, I.e. always have the unique ID be 11111111111 or something like that

config/tests/samples/create/harness.go
```diff
@@ -24,6 +24,8 @@ import (
 	"testing"
 	"time"
 
+	"gopkg.in/dnaeon/go-vcr.v3/recorder"
+
```
maqiuyujoyce, line 28: Nit: Remove empty line 28?

config/tests/samples/create/harness.go
```diff
@@ -236,6 +240,8 @@ func NewHarness(ctx context.Context, t *testing.T) *Harness {
 		kccConfig.GCPAccessToken = h.gcpAccessToken
 	} else if targetGCP := os.Getenv("E2E_GCP_TARGET"); targetGCP == "real" {
 		t.Logf("targeting real GCP")
+	} else if targetGCP := os.Getenv("E2E_GCP_TARGET"); targetGCP == "vcr" {
```
maqiuyujoyce, line 243: Non-blocking future improvement idea: Can we have all the environment variables and valid values used by integration tests documented and maintained somewhere? As someone who rarely touches the code here, it's hard for me to figure out what environment variables I should set and why.

config/tests/samples/create/harness.go
```diff
@@ -305,17 +311,40 @@ func NewHarness(ctx context.Context, t *testing.T) *Harness {
 		kccConfig.HTTPClient = &http.Client{Transport: t}
 	}
 
-	// Intercept (and log) TF requests
-	transport_tpg.DefaultHTTPClientTransformer = func(ctx context.Context, inner *http.Client) *http.Client {
-		ret := inner
-		if t := ctx.Value(httpRoundTripperKey); t != nil {
-			ret = &http.Client{Transport: t.(http.RoundTripper)}
+	if targetGCP := os.Getenv("E2E_GCP_TARGET"); targetGCP == "vcr" {
```
maqiuyujoyce, line 314: Non-blocking refactoring idea:
Can we do one `targetGCP := os.Getenv("E2E_GCP_TARGET")` in the beginning of the function, and then do switch case/if else for different conditions in line233-247, line314 and one, etc?

Repetitively retrieving the same environment seems to reduce the readability.

config/tests/samples/create/harness.go
```diff
@@ -305,17 +311,40 @@ func NewHarness(ctx context.Context, t *testing.T) *Harness {
...
+		transport_tpg.DefaultHTTPClientTransformer = func(ctx context.Context, inner *http.Client) *http.Client {
+			ret := inner
+			if t := ctx.Value(httpRoundTripperKey); t != nil {
+				ret = &http.Client{Transport: t.(http.RoundTripper)}
+			}
+			dir := "pkg/test/resourcefixture/testdata/vcr/cassette/"
```
maqiuyujoyce, line 320: Just curious: Is `cassette` a special code name? I personally don't know much about VCR and its concepts so would like to learn more context.

tests/e2e/unified_test.go
```diff
@@ -16,7 +16,11 @@ package e2e
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"gopkg.in/dnaeon/go-vcr.v3/cassette"
```
maqiuyujoyce, line 21: Line 21-23 should be moved down to group into line39-41.

tests/e2e/unified_test.go
```diff
@@ -153,6 +157,78 @@ func testFixturesInSeries(ctx context.Context, t *testing.T, testPause bool, can
...
+						obj := make(map[string]any)
+						if err := json.Unmarshal([]byte(s), &obj); err != nil {
+							klog.Fatalf("error from json.Unmarshal(%q): %v", s, err)
+						}
+						toReplace, _, _ := unstructured.NestedString(obj, "user")
+						if len(toReplace) != 0 {
+							s = strings.Replace(s, toReplace, "user@google.com", -1)
+						}
+
+						// A very hacky way to replace actual debug info from log.
```
maqiuyujoyce, line 212: Nit: Would be great if the hacky way is explained (why hacky, any future TODO, etc).

maqiuyujoyce (commented):
/lgtm

All comments are non-blocking. Would be great to see the demo!

tests/e2e/unified_test.go
gemmahou: Actually I plan to remove this code block in PR #1393 and re-generate the log file. 
To be able to speed up test, I'd like to remove unnecessary interactions from the recorder log. For example, failed HTTP logs with error code 404/400/403 and repeated logs come from the retry transport.

config/tests/samples/create/harness.go
gemmahou: `recorder` and `cassette` are both VCR concepts. Similar to the VCR recorder in the real world, this test package simulates the workflow, recording HTTP interactions instead of audio (😂), and saving them in a file, which is used to replay them later.
TBH, the folder name can be anything (`tape`, `episode`, etc., 😂), I just borrowed the name  as it's a known concept for storing recorded logs.

justinsb:
Thanks @gemmahou 

/approve
/lgtm

config/tests/samples/create/harness.go
gemmahou: I created b/332719149 to address this comment.

config/tests/samples/create/harness.go
gemmahou: this will be done in b/332719149 as well.

PR #1352, opened 2024-03-13
tests: log if no samples

acpana:
While running the samples tests it may be helpful to know if no samples load and exist early.

### DIFF
--- a/config/tests/samples/create/samples_test.go
+++ b/config/tests/samples/create/samples_test.go
@@ -234,8 +234,7 @@ func TestAll(t *testing.T) {
 	setup()
 	samples := LoadMatchingSamples(t, regexp.MustCompile(runTestsRegex), project)
 	if len(samples) == 0 {
-		t.Logf("No tests to run for pattern %s", runTestsRegex)
-		return
+		t.Fatalf("No tests to run for pattern %s", runTestsRegex)
 	}
 
 	// Sort the samples in descending order by number of resources. This is an attempt to start the samples that use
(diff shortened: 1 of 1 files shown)


config/tests/samples/create/samples_test.go
```diff
@@ -233,6 +233,11 @@ func TestAll(t *testing.T) {
 
 	setup()
 	samples := LoadMatchingSamples(t, regexp.MustCompile(runTestsRegex), project)
+	if len(samples) == 0 {
+		t.Logf("No tests to run for pattern %s", runTestsRegex)
```
justinsb, line 237: Nit: Does this happen?  Should this be t.Errorf?   It's an odd edge case ... t.Skipf might be the most appropriate!  (This is why we use nested tests elsewhere!)

acpana: Actually I quite like erroring out as it will be a better signal that something is wrong.

Yeah, with the tests as is, it happened to me when I didn't figure out how to use `run-tests` correctly 😅 . So:

```
# misspel
$ go test -v -tags=integration ./config/tests/samples/create -timeout 1600s -test.count=1 -test.run TestAll -run-tests gfehub 2>&1

# not a thing / i'm holding it wrong
$ go test -v -tags=integration ./config/tests/samples/create -timeout 1600s -test.count=1 -test.run TestAll -run-tests notathing 2>&1
```

come out as 

```
...
ok      github.com/GoogleCloudPlatform/k8s-config-connector/config/tests/samples/create 11.430s
```

which can offer a false sense of security hehe.

maqiuyujoyce: I think error out gives a more clear signal. We can always make improvements in the future.

PR #1353, opened 2024-03-13
Move computed fields to status.observedState for new CRDs

maqiuyujoyce:
### Change description

Fixes b/327226936

* Reused the allowlist to block the `state-into-spec: merge` feature to identify whether computed fields should be under `status.observedState` in a resource. Added util functions.
* Updated the logic when fetching resource-specific fields under `status` in TF-based IAM client, and the reconciliation flow for TF-based resources.
* Updated the logic when populating resource-specific fields under `status` for TF-based resources.
* Updated the CRD generation for newly added TF-based resources to move computed fields to status.observedState.
* Added unit tests to cover the changes.

### Tests you have done

Downloaded #1271, rebased on top of this PR and ran the following command:

```
go test -timeout 1200s -v -tags=integration ./pkg/controller/dynamic/ -test.run TestCreateNoChangeUpdateDelete -run-tests vertexaidatasetbasic
```

Verified that the created resource has the following status:

```
status:map[conditions:[map[lastTransitionTime:2024-03-13T20:02:10Z message:The resource is up to date reason:UpToDate status:True type:Ready]] observedGeneration:1 observedState:map[createTime:2024-03-13T20:02:00.322965Z name:projects/[REDACT]/locations/us-central1/datasets/1730698372328718336 updateTime:2024-03-13T20:02:01.420379Z]]
```


pkg/controller/iam/iamclient/tfiamclient.go
```diff
@@ -685,7 +685,7 @@ func resolveTargetFieldValue(r *krmtotf.Resource, targetField string) (string, e
 		if val, exists, _ := unstructured.NestedString(r.Spec, strings.Split(key, ".")...); exists {
 			return val, nil
 		}
-		if val, exists, _ := unstructured.NestedString(r.Status, strings.Split(key, ".")...); exists {
+		if val, exists, _ := unstructured.NestedString(r.GetStatusOrObservedState(), strings.Split(key, ".")...); exists {
```
justinsb, line 688: This is high risk (touches many code paths in the magic machine), so I'm going to assign to 1.115

pkg/k8s/observedstate.go
```diff
@@ -0,0 +1,22 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package k8s
+
+func HasComputedFieldsUnderObservedState(kind, version string) bool {
+	if version == KCCAPIVersion && !SupportsComputedFieldsUnderStatusInKind(kind) {
```
justinsb, line 18: Nit: We need to change KCCAPIVersion to something that includes the word "beta" :-)

pkg/k8s/allowlist.go
```diff
@@ -18,223 +18,237 @@ import (
...
+	// Any newly supported v1beta1 kinds should NOT have computed fields
+	// directly under 'status' in the schema.
+	v1beta1KindsWithComputedFieldsUnderStatus = v1beta1KindsWithStateIntoSpecMergeSupport
+)
 
 func SupportsStateIntoSpecMergeInKind(kind string) bool {
 	return isValueInAllowlist(kind, v1beta1KindsWithStateIntoSpecMergeSupport)
 }
 
+func SupportsComputedFieldsUnderStatusInKind(kind string) bool {
```
justinsb, line 248: IIUC callers should call `HasComputedFieldsUnderObservedState` instead ... Make this private or inline?

pkg/k8s/observedstate.go
```diff
@@ -0,0 +1,22 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package k8s
+
+func HasComputedFieldsUnderObservedState(kind, version string) bool {
```
justinsb, line 17: Nit: This method needs a comment.  I think "Computed Field" is a Terraform thing, maybe this should  "OutputFieldsAreUnderObservedState" or maybe just "GetOutputField(u *unstructured.Unstructured, fieldPath string)"?

pkg/krmtotf/resource.go
```diff
@@ -322,6 +326,21 @@ func (r *Resource) AllTopLevelFieldsAreImmutableOrComputed() bool {
...
+	observedState, exists, _ := unstructured.NestedMap(status, k8s.ObservedStateFieldName)
+	if !exists {
+		observedState = make(map[string]interface{})
+	}
+	return observedState
+}
+
+func (r *Resource) GetStatusOrObservedState() map[string]interface{} {
+	if k8s.HasComputedFieldsUnderObservedState(r.Kind, r.GroupVersionKind().Version) {
+		return getObservedStateFromStatus(r.Status)
```
justinsb, line 339: One way to reduce risk in this PR is to rename r.Status to r.status (just the field, not a refactor).  Then we can be sure that we have looked at every code point that reads r.Status / r.Status.ObservedState

pkg/crd/crdgeneration/tf2crdgeneration.go
```diff
@@ -67,12 +68,17 @@ func GenerateTF2CRD(sm *corekccv1alpha1.ServiceMapping, resourceConfig *corekccv
...
+		moveComputedFieldsToObservedState(statusOrObservedStateJSONSchema)
+	} else {
+		statusOrObservedStateJSONSchema, err = k8s.RenameStatusFieldsWithReservedNamesIfResourceNotExcluded(resource, statusOrObservedStateJSONSchema)
+		if err != nil {
+			return nil, fmt.Errorf("error renaming status fields with reserved names for %#v: %w", statusOrObservedStateJSONSchema, err)
+		}
 	}
-	populateObservedState(resourceConfig, specJSONSchema, statusJSONSchema)
-	for k, v := range statusJSONSchema.Properties {
+	addObservedFieldsToObservedState(resourceConfig, specJSONSchema, statusOrObservedStateJSONSchema)
```
jingyih, line 80: Does this mean we should NOT add a "computed field" to  `observedFields` in servicemapping.

pkg/controller/iam/iamclient/tfiamclient.go
maqiuyujoyce: Since this PR can't make it to 1.114, I'm going to allowlist ComputeNetworkFirewallPolicyAssociation.

pkg/krmtotf/resource.go
maqiuyujoyce: I love this trick. Will do it and see if there is anything missing.

pkg/crd/crdgeneration/tf2crdgeneration.go
maqiuyujoyce: Yes. This is indicated in the [observedState field comment](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/apis/core/v1alpha1/servicemapping_types.go#L185) and executed in the [CRD generation logic](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/crd/crdgeneration/tf2crdgeneration.go#L464) but I think we should add a unit test to protect it.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_computenetworkfirewallpolicyassociations.compute.cnrm.cloud.google.com.yaml
```diff
@@ -186,9 +186,13 @@ spec:
                   current reported status reflects the most recent desired state of
                   the resource.
                 type: integer
-              shortName:
-                description: The short name of the firewall policy of the association.
-                type: string
+              observedState:
```
jingyih, line 189: I don't quite understand why this field got moved?

jingyih: This was discussed offline in the group chat.

pkg/k8s/allowlist.go
maqiuyujoyce: Done.

pkg/k8s/observedstate.go
maqiuyujoyce: Changed the function name to `OutputOnlyFieldsAreUnderObservedState`. It needs to take the smallest set of params because it is also called by CRD generation code (which doesn't have unstructured).

pkg/crd/crdgeneration/tf2crdgeneration.go
maqiuyujoyce: #1401

pkg/k8s/observedstate.go
maqiuyujoyce: Agreed that having the version in the name will make it more readable. Updated it in a separate PR to avoid "contaminating" this PR: #1402 .

pkg/krmtotf/resource.go
maqiuyujoyce: Verified that it the current PR already covers everything we can identify: #1403.

jingyih (commented):
/lgtm
/approve


/hold for @justinsb to take another look

jingyih:
/hold cancel
to unblock other PRs.

PR #1355, opened 2024-03-13
fix: skip for golden check

acpana:
During the `pause` work, I discovered that there's not a golden log for `serviceusage.cnrm.cloud.google.com/Service` just yet. I can capture that later but skipping it if `GOLDEN_REQUEST_CHECKS` is set for now.

### DIFF
--- a/config/tests/samples/create/harness.go
+++ b/config/tests/samples/create/harness.go
@@ -448,6 +448,10 @@ func MaybeSkip(t *testing.T, name string, resources []*unstructured.Unstructured
 			case schema.GroupKind{Group: "", Kind: "Secret"}:
 
 			case schema.GroupKind{Group: "serviceusage.cnrm.cloud.google.com", Kind: "Service"}:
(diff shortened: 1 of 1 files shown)


config/tests/samples/create/harness.go
```diff
@@ -448,6 +448,10 @@ func MaybeSkip(t *testing.T, name string, resources []*unstructured.Unstructured
 			case schema.GroupKind{Group: "", Kind: "Secret"}:
 
 			case schema.GroupKind{Group: "serviceusage.cnrm.cloud.google.com", Kind: "Service"}:
+				if os.Getenv("GOLDEN_REQUEST_CHECKS") != "" {
+					// no golden log for this resource  yet
```
gemmahou, line 452: Nit: Can we add a TODO here and create a ticket to capture logs later and remove this special handling?

acpana: Yes currently in the process of running a capture for this.

PR #1364, opened 2024-03-14
chore: Fix some linter issues

justinsb:
These merged because the PR was super-old (and maybe a prow bug also?)


justinsb:
(This is in 1.114 on the assumption that we're going to start hitting this in other PRs)

justinsb:
This is not a blocker for 1.114, and it doesn't seem to be causing problems, so I guess I will punt to 1.115.  I don't _think_ I imagined the failure, so I am a little confused!

yuwenma:
/lgtm
/approve

(The review is a little bit late) Thank you for this!

PR #1360, opened 2024-03-14
CCC: Switch Enum values for spec.stateIntoSpec to upper camel case

justinsb:
This follows the KRM conventions.

PR #1366, opened 2024-03-14
WIP: chore: Add tests for CCC stateIntoSpec

justinsb:
Builds on #1360 (the first commit).

### DIFF
--- a/config/tests/samples/create/harness.go
+++ b/config/tests/samples/create/harness.go
@@ -406,6 +406,8 @@ func MaybeSkip(t *testing.T, name string, resources []*unstructured.Unstructured
 			gvk := resource.GroupVersionKind()
 
 			switch gvk.Group {
+			case "core.cnrm.cloud.google.com":
+				continue
 			case "certificatemanager.cnrm.cloud.google.com":
 				continue
 			}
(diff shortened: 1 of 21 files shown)


justinsb:
/wip

/assign @acpana 

Putting in milestone 1.115 as I think we want tests, but it doesn't block the release
/milestone 1.115

acpana:
this LGTM to me overall but I think we raced on some of the changes; I did a rebase and a small change in #1394

maqiuyujoyce (commented):
Thank you for setting up the test framework, @justinsb !

I have a couple of high level questions:

1. What is the benefit of this new test framework compared with our existing test frameworks? Is it a correct understanding that the benefit is the new feature to control steps for the test scenarios? Why not merge it into the unified tests? E.g. for all the existing test (samples / mock), they do two steps by default: APPLY and DELETE.
3. Looks like with the new framework, normalization of each kind needs to be done separately. I assume having a simpler approach for resource contributors to add resource-specific test (e.g. via dynamic tests and samples test) is still preferred before we can provide a simpler and more systematic approach to use the scenario test?

acpana:
hey @maqiuyujoyce thanks for all the feedback here. I'm afraid I raced you here to integrate this change in #1394. 

Let me address your feedback in a follow up PR 🙏🏼  and take this off Justin's plate.

PR #1377, opened 2024-03-15
Modify google_alloydb_cluster and google_alloydb_instance resources t…

199201shubhamsahu:
### Change description

This PR only includes the third party terraform-provider-google-beta resource file changes.
Implementation PR in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1279

Add support for AlloyDBB Cross Region Replication by following changes

1. cluster.cluster_type : possible values [default=PRIMARY, SECONDARY]
2. cluster.secondary_config : For secondary clusters to reference their associated primary cluster
3. instance.instance_type : new value SECONDARY to support secondary instances

b/323793486


config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_alloydbinstances.alloydb.cnrm.cloud.google.com.yaml
```diff
@@ -121,10 +121,6 @@ spec:
                   instance, instance is created in a random zone with available capacity.
                 type: string
               instanceType:
-                description: 'Immutable. The type of the instance. If the instance
```
justinsb, line 124: This field is still immutable, I'm guessing, we just didn't want the `depends_on` terraform-ism in the docs (I'm guessing)?

(If it's not immutable, I have questions about what happens if you go from READ_POOL -> PRIMARY)

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_alloydbclusters.alloydb.cnrm.cloud.google.com.yaml
```diff
@@ -178,6 +178,10 @@ spec:
                     - startTimes
                     type: object
                 type: object
+              clusterType:
+                description: 'The type of cluster. If not set, defaults to PRIMARY.
```
justinsb, line 182: We discussed in chat.  AIUI:

PRIMARY -> SECONDARY is not allowed.

SECONDARY -> PRIMARY is allowed, but it does not "replace" the existing PRIMARY.  Rather, we start a "fork" of the database, with this cluster as the PRIMARY.  So the data from the old PRIMARY and this new PRIMARY would be expected to diverge.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_alloydbclusters.alloydb.cnrm.cloud.google.com.yaml
```diff
@@ -231,6 +235,12 @@ spec:
                       If not set, defaults to 14 days.
                     type: integer
                 type: object
+              deletionPolicy:
+                description: |-
+                  Policy to determine if the cluster should be deleted forcefully.
```
justinsb, line 240: Ideally deletion would be deletion and this wouldn't be needed.  KCC also want to mirror the underlying GCP APIs, and this isn't technically a field on the resource.

It sounds like the big reason this is needed is that we basically require force to be set for secondaries with instances (?).

Another option is that we automatically set force=true when trying to delete a secondary cluster, but maybe that gets more complicated as the API evolves (e.g. maybe at some point it will be possible to delete a secondary instance?)

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_alloydbclusters.alloydb.cnrm.cloud.google.com.yaml
```diff
@@ -505,6 +515,19 @@ spec:
                 - clusterRef
                 - pointInTime
                 type: object
+              secondaryConfig:
+                description: Configuration of the secondary cluster for Cross Region
+                  Replication. This should be set if and only if the cluster is of
+                  type SECONDARY.
+                properties:
+                  primaryClusterName:
```
justinsb, line 523: From a usability perspective, this should be a ref I think.  We should be able to configure that in the service mapping.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_alloydbclusters.alloydb.cnrm.cloud.google.com.yaml
justinsb: (If so, I think this is fine - that behaviour shouldn't cause any problems for usage with e.g. gitops)

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_alloydbinstances.alloydb.cnrm.cloud.google.com.yaml
199201shubhamsahu: Yes, the field is still immutable. So modifying the instanceType deletes and recreates the instance.
However, there is no use case where a user needs to modify the instanceType directly.

The READ_POOL do not change their type, and for PRIMARY and SECONDARY instances it always reflects the cluster type of the cluster they are associated with.

We do have `promotion` implementation on the secondary clusters.
For promotion we modify the cluster_type of secondary cluster from SECONDARY to PRIMARY, which internally also modifies the secondary instance to be a primary instance.
And for the same reason we are using instanceTypeRef implemented in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1279/files#diff-ea2ee6386ff7f479f831a2b6e3fad19501c9873cb8a9127c33d94e957f3a5ed7R194-R200

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_alloydbclusters.alloydb.cnrm.cloud.google.com.yaml
199201shubhamsahu: Yes. Thanks for adding the discussion here

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_alloydbclusters.alloydb.cnrm.cloud.google.com.yaml
199201shubhamsahu: Yes, we do this because there is no deletion operation for secondary instances. And the only way to delete a secondary instance is by deleting the associated secondary cluster.

We do have the `--force` in the deletion API https://cloud.google.com/sdk/gcloud/reference/alloydb/clusters/delete and the `deletion_policy` set to `FORCE` does the same thing.

Yes, we can automatically set force=true for secondary clusters, but by allowing the user to set the deletion_policy field, it is more clear to the users that their operation will forcefully delete the associated secondary instance as well

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_alloydbclusters.alloydb.cnrm.cloud.google.com.yaml
199201shubhamsahu: Yes, this is already part of the service mapping in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1279

The presubmit validation expects the third-party terraform-provider-google-beta update in a different PR, hence this new PR

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_alloydbinstances.alloydb.cnrm.cloud.google.com.yaml
maqiuyujoyce: (Just some non-blocking thoughts.)

Based on the use cases it supports, the field's mutability is kinda tricky: Users are not allowed to update it, but the API will update it when needed, and the users need to keep up with the updated, new value in their gitops flow. Not having control over the fields they set previously seems to be not very well compatible with the declarative style, because it means ordering of imperative operations (who sets the value first, second, third) matters a lot.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_alloydbclusters.alloydb.cnrm.cloud.google.com.yaml
maqiuyujoyce: A couple of thoughts (non-blocking):

1. We should add instructions about how to actually trigger the force deletion because we also has a general `deletion-policy` annotation, and it must be set "correctly" before the `deletionPolicy` option can be properly executed.
2. As this is a cascading deletion, the users will need to be aware of the following facts and clean up their resources properly:
    1. When the parent AlloyDBCluster is force-deleted correctly, the AlloyDBInstance resources should be deleted, otherwise it'll cause reconciliation failures (and potential causing alerts and consuming error budget).
    2. When the secondary AlloyDBInstance is deleted by itself, the only thing happening will be the corresponding GCP resource is no longer controlled by Config Connector.

maqiuyujoyce (commented):
/lgtm

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_alloydbinstances.alloydb.cnrm.cloud.google.com.yaml
justinsb: Just to check - can the user ever set this?  It sounds like it is determined by the cluster?

justinsb:
/lgtm also

(I had one question about whether instanceType is settable on the instance, vs being derived from the cluster, I think a mock would help me at least to better understand this API)

199201shubhamsahu:
> /lgtm also
> 
> (I had one question about whether instanceType is settable on the instance, vs being derived from the cluster, I think a mock would help me at least to better understand this API)

During creation the instanceType is settable by the user and is a required field.
For instance created with clusterRef pointing to PRIMARY clusters, the instanceType can be only set to PRIMARY or READ_POOL
For instance created with clusterRef pointing to SECONDARY clusters, the instanceType can be only set to SECONDARY or READ_POOL
It throws a validation error otherwise

However, we support an feature operation called PROMOTION. This operation can convert a SECONDARY cluster to PRIMARY, this operation also converts the associated SECONDARY instance to PRIMARY internally and requires no change on the READ_POOL instances. Hence, we introduced a instanceTypeRef field in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1279 so that the instanceType is directly tied to the clusterType of the associated cluster for PRIMARY & SECONDARY instances.

justinsb:
Thanks @199201shubhamsahu - that makes sense!

/approve
/lgtm

maqiuyujoyce:
@199201shubhamsahu fyi what @justinsb means about mock is: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/mockgcp/README.md

@justinsb do you consider a mock is a blocker for #1279 ?

PR #1384, opened 2024-03-16
tests: revert addition of PROJECT_ID to sample

justinsb:
This sample is shared by our operator smoke-tests (!!!), and those
don't perform substitution.


justinsb:
@acpana I think you did this in #1147 .. unfortunately it looks like we _also_ (re)used the same sample in our operator tests (and they only run after release, sigh....).  "Thankfully" I think we can fix the tests and unblock 1.114

#1385 is probably the better fix - we should be testing our code, not some artifacts we download IMO.  But that'll likely be a lot more involved to merge.

/assign @acpana

acpana:
thanks for spotting and fixing this Justin! LGTM

I do think that adding the sample annotation is required as part of the CONTRIBUTING.md flow to work but I can add an instruction in the that doc instead of adding the annotation in the sample here.

/lgtm
/approve

maqiuyujoyce:
> I do think that adding the sample annotation is required as part of the CONTRIBUTING.md flow to work but I can add an instruction in the that doc instead of adding the annotation in the sample here.

I think if we don't reuse the sample (like what Justin did in #1385 ) then the exception in the instructions is not necessary?

PR #1385, opened 2024-03-16
WIP: chore: Refactor operator e2e tests

justinsb:
Loading the ArtifactRegistry sample from a distribution artifact is
brittle, and makes it harder to test the code we are running.

### DIFF
--- a/operator/tests/e2e/e2e_test.go
+++ b/operator/tests/e2e/e2e_test.go
@@ -47,10 +47,13 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


justinsb:
I think we want to do #1384 to unblock the release.

In general, we should not be downloading artifacts, we should be testing the code we have.  I think we can improve that as part of 1.116

operator/tests/e2e/e2e_test.go
```diff
@@ -927,126 +856,73 @@ func setupWorkloadIdentityForNamespace(namespace, serviceAccEmail, projectID str
...
-		}
-	}
-	switch len(repoNames) {
-	case 0:
-		return "", fmt.Errorf("no ArtifactRegistryRepository found in directory '%v'", repoYAMLDir)
-	case 1:
-		return repoNames[0], nil
-	default:
-		return "", fmt.Errorf("multiple ArtifactRegistryRepositories found in directory '%v'", repoYAMLDir)
+func getArtifactRegistryRepository(namespace string, uniqueID string) (*unstructured.Unstructured, error) {
```
maqiuyujoyce, line 859: Some non-blocking thoughts:

We may want to somehow match this sample with the latest working sample. In the future we may introduce breaking changes in the CRDs (unlikely) or update CRD version (v1beta1 -> v1), and this hard-coded, secret yaml may only be detected "after release" just like this time.

ldanielmadariaga:

PR #1387, opened 2024-03-16
WIP: genapi: create direct IAM ServiceAccount controller

justinsb:
This API is not the most complicated, but it is also unusual in that
the resource identifier (primary key) is variable.

### DIFF
--- a/pkg/test/utils.go
+++ b/pkg/test/utils.go
@@ -142,18 +142,18 @@ func IgnoreLeadingComments(s string) string {
 	return strings.TrimSpace(strings.Join(out, "\n")) + "\n"
 }
 
-func CompareGoldenObject(t *testing.T, path string, got []byte) {
+func CompareGoldenObject(t *testing.T, p string, got []byte) {
 	writeGoldenOutput := os.Getenv("WRITE_GOLDEN_OUTPUT") != ""
-	want, err := os.ReadFile(path)
+	want, err := os.ReadFile(p)
 	if err != nil {
 		if writeGoldenOutput && os.IsNotExist(err) {
 			// Expected when creating output for the first time
-			if err := os.WriteFile(path, got, 0644); err != nil {
-				t.Errorf("failed to write golden output %s: %v", path, err)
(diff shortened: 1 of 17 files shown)


justinsb:
/assign @cheftako 

Experimenting with direct implementation to figure out the mapping functions we need.  Service account is perhaps a bad choice in that the API is so small and so irregular, so did a more realistic API in #1389 (though really I should try an even more modern API that follows all the latest AIPs)

justinsb:
@cheftako any thoughts on this, otherwise I say we ask @acpana and @anhdle-sso to take a look and see if this is something they have stronger opinions on :-)

justinsb:
This is the updated "big picture", and includes the other PRs.

/assign @gemmahou

pkg/test/utils.go
```diff
@@ -165,7 +165,19 @@ func CompareGoldenObject(t *testing.T, path string, got []byte) {
...
+	if diff == "" {
+		return
+	}
+
+	if writeGoldenOutput {
+		// Write the output to the golden file
+		if err := os.WriteFile(p, []byte(got), 0644); err != nil {
+			t.Fatalf("failed to write golden output %s: %v", p, err)
+		}
+		t.Errorf("wrote updated golden output to %s", p)
```
gemmahou, line 179: Since `writeGoldenOutput ` is true, it's expected we are writing/updating the golden output. So use `Logf` here is clearer than `Errorf`. Otherwise, I think we want to make the message more specific about why we consider this an error. For example: "diff detected between incoming log and golden log. Wrote updated golden log to %s"

tests/e2e/testdata/scenarios/walk_iamserviceaccount/_http00.log
gemmahou, line 1: Not a blocker but the filename _http00/_object00 is unclear what exactly the file does. 
I'm thinking about adding a `title` field in `Script` struct and in script.yaml we specify a title for each scenario, when loading script.yaml, we retrieve the title and append it to the file name. So we will have a better context for those logs.

ldanielmadariaga:

PR #1388, opened 2024-03-16
chore: avoid changing name of tests

justinsb:
I think the test method name (the first component) is sufficient here,
and then we don't need to change the test paths for "normal"
TestAllInSeries.

### DIFF
--- a/tests/e2e/unified_test.go
+++ b/tests/e2e/unified_test.go
@@ -83,7 +83,7 @@ func TestAllInSeries(t *testing.T) {
 		}
 	})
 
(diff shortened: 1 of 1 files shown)


justinsb:
/assign @acpana 

I don't think we need to change the test name here.  I'm not opposed to changing the test path (indeed, I'm proposing we include the GCP service), but it is a bit of work, so I'd rather not change it unless we need to - is there any reason to change the test path here?

acpana:
Ah no I goofed here, thanks for simplifying this. A format of `TestPauseInSeries/PauseInSeries/fixtures/fullalloydbinstance` is redundant 😝 .

/lgtm
/approve

PR #1397, opened 2024-03-19
chore: replace project id placeholder in operator tests

justinsb:
This should allow us to use the sample with the placeholder.


justinsb:
/assign @acpana 

My previous attempt did not work, I think because we're still loading the released version of the manifest in the tests (brain fail!).  Hopefully this will let us deal with that manifest, and if it does we can potentially re-add the annotation if it helps for the samples...

acpana:
/lgtm
/approve

> Hopefully this will let us deal with that manifest, and if it does we can potentially re-add the annotation if it helps for the samples...

Yeah no worries I can add that back in if you think it's worth it: I mostly added it for the guide in CONTRIBUTING.MD to work and was going to call out adding the annotation as an instruction in the guide instead.

PR #1400, opened 2024-03-19
docs: clarify field usage

acpana:
Clarifies field `spec.resourcePolicies` only applies for zonal ComputeDisk resources.

internal: b/324925137


maqiuyujoyce:
/lgtm
/approve

Looks like the Go client doesn't have proper comments for the `resourcePolicies` field: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/clients/generated/apis/compute/v1beta1/computedisk_types.go#L244 I'm guessing this is because this field is a list of references, and our template doesn't support it properly. @acpana could you file a issue about it?

PR #1401, opened 2024-03-20
Ensure observed fields are 'spec' fields

maqiuyujoyce:
### Change description

Updated the servicemapping unit tests to ensure observed fields are 'spec' fields.

Also flattened some assertions to avoid additional checks.

### Tests you have done


jingyih (commented):
/lgtm
/approve

maqiuyujoyce:
`TestAllInSeries/samples/project-level-policy` in the mock test is quite flaky, I personally ran into it a couple of times a week. We should root-cause it and decide if we should take it out of the mock test suite.

PR #1403, opened 2024-03-20
[WIP] Unexport 'status' field in K8s.Resource

maqiuyujoyce:
### Change description

Unexported `status` field in K8s.Resource and updated the marshal/unmarshal logic to ensure the unexported `status` field is still passed along.

### Tests you have done

go test -timeout 1200s -v -tags=integration ./pkg/controller/dynamic/ -test.run TestCreateNoChangeUpdateDelete -run-tests pubsubtopic

### DIFF
--- a/pkg/k8s/resource_test.go
+++ b/pkg/k8s/resource_test.go
@@ -88,7 +88,9 @@ func TestResource_IsSpecOrStatusUpdateRequired(t *testing.T) {
 	tests := []struct {
 		name           string
 		resource       *k8s.Resource
+		resourceStatus map[string]interface{}
 		original       *k8s.Resource
+		originalStatus map[string]interface{}
 		expectedResult bool
 	}{
 		{
@@ -100,18 +102,18 @@ func TestResource_IsSpecOrStatusUpdateRequired(t *testing.T) {
 				Spec: map[string]interface{}{
 					"foo": "someValue",
 				},
-				Status: map[string]interface{}{
-					"observedGeneration": float64(1),
-				},
+			},
+			resourceStatus: map[string]interface{}{
+				"observedGeneration": float64(1),
 			},
 			original: &k8s.Resource{
 				ObjectMeta: metav1.ObjectMeta{
 					Generation: 1,
 				},
 				Spec: map[string]interface{}{},
-				Status: map[string]interface{}{
-					"observedGeneration": float64(1),
-				},
+			},
+			originalStatus: map[string]interface{}{
+				"observedGeneration": float64(1),
 			},
 			expectedResult: true,
 		},
@@ -124,9 +126,9 @@ func TestResource_IsSpecOrStatusUpdateRequired(t *testing.T) {
 				Spec: map[string]interface{}{
 					"foo": "someValue",
 				},
-				Status: map[string]interface{}{
-					"bar": "someValue",
-				},
+			},
+			resourceStatus: map[string]interface{}{
+				"bar": "someValue",
 			},
 			original: &k8s.Resource{
 				ObjectMeta: metav1.ObjectMeta{
@@ -135,8 +137,8 @@ func TestResource_IsSpecOrStatusUpdateRequired(t *testing.T) {
 				Spec: map[string]interface{}{
 					"foo": "someValue",
 				},
-				Status: map[string]interface{}{},
 			},
+			originalStatus: map[string]interface{}{},
 			expectedResult: true,
 		},
 		{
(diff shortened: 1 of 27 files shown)


justinsb:
FYI, I don't think there's any reason to abandon this.  Using methods is generally much better at preventing errors than directly accessing fields.

pkg/k8s/resource_test.go
```diff
@@ -83,147 +83,147 @@ func TestResource_IsResourceIDConfigured(t *testing.T) {
 	}
 }
 
-func TestResource_IsSpecOrStatusUpdateRequired(t *testing.T) {
```
justinsb, line 86: What broke here?

pkg/krmtotf/resource.go
```diff
@@ -322,6 +326,21 @@ func (r *Resource) AllTopLevelFieldsAreImmutableOrComputed() bool {
 	return true
 }
 
+func getObservedStateFromStatus(status map[string]interface{}) map[string]interface{} {
+	observedState, exists, _ := unstructured.NestedMap(status, k8s.ObservedStateFieldName)
+	if !exists {
+		observedState = make(map[string]interface{})
```
justinsb, line 332: Probably safer to just return nil, lest someone try to write to it

pkg/krmtotf/tftokrm.go
```diff
@@ -80,26 +80,51 @@ func GetSpecAndStatusFromState(resource *Resource, state *terraform.InstanceStat
...
+	if observedGeneration, ok := resource.GetStatus()["observedGeneration"]; ok {
 		status["observedGeneration"] = deepcopy.DeepCopy(observedGeneration)
 	}
 	if resource.ResourceConfig.ObservedFields != nil {
-		observedState := resolveObservedState(resource, krmState)
-		if len(observedState) > 0 {
-			status["observedState"] = observedState
+		observedFields := resolveObservedFields(resource, krmState)
+		if len(observedFields) > 0 {
+			// Merge the observed fields into the observed state.
```
justinsb, line 112: Should we merge or just build it anew?

pkg/krmtotf/fetchlivestate_test.go
```diff
@@ -30,666 +30,768 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
-func TestWithFieldsPresetForRead(t *testing.T) {
```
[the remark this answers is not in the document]
maqiuyujoyce: Nope, see the answer above.

maqiuyujoyce:
> FYI, I don't think there's any reason to abandon this. Using methods is generally much better at preventing errors than directly accessing fields.

In order to make it work E2E, marshaling/unmarshaling logic for k8s.Resource, krmtotf.Resource, and dcl.Resource need to be updated. This impacts every single resource in admission and reconciliation.

I'm pretty confident about #1353 as the scope of it is quite limited (TF-based resources only, `status` field only, will likely only impact new resources), and we have decent unit tests, mock tests, and integration tests to cover most use cases. But for this PR, the scope of the change is way larger, and I'm quite concerned about merging the change before we have healthy integration tests with good coverage.

I'm proposing we don't include it for 1.115 milestone and only merge the refactored code after we have the existing integration test back to healthy / set up a new integration test flow with good coverage.

pkg/krmtotf/tftokrm.go
maqiuyujoyce: Just to clarify, what is "build it anew" in this context? Creating a new map?

Observed fields are also part of the observed state but manually configured for `spec` fields in service mappings (#1023, #1059) so I think merging should be the right behavior. Glad to adjust if you are suggesting a simpler implementation.

pkg/krmtotf/resource.go
maqiuyujoyce: #1430

justinsb:
I propose we get the observedStatus work into 1.116.

Then we can cut the 1.115 branch this week (and give it a few days to stabilize).  Big change there will be AlloyDB, it looks like.

This does presuppose that we 


>marshaling/unmarshaling logic for k8s.Resource, krmtotf.Resource, dcl.Resource, and unstructured.Unstructured need to be updated. 

What does this look like?  Have you tried implementing json.Unmarshaler / Json.Marshaler?  The trick is normally to assemble whatever you want the object to look like (often in a `map[string]any`), and then call json.Marshal on it.

Anyway, I popped this back into the 1.116 milestone, but mostly because I hadn't seen your comments when going through.  I don't mind how we tackle the safety from a code-perspective, we could add tests as I started doing in #1438

maqiuyujoyce:
> What does this look like?

Here is one example: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1403/files#diff-565e4cccca460065f11c5e41024f882b8ea7a8cb9eeff6f0e0959b96b75dfd63R414

> Have you tried implementing json.Unmarshaler / Json.Marshaler?

In our codebase, a lot of the marshaling/unmarshaling logic are done by calling util.Marshal() (code here:https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/util/util.go#L22). I did a search, and found it is called at 45 places, and not every time it is used to unmarshall to a k8s.Resource.

I tried to add special logic to handle k8s.Resource in util.Marshal(), but it gives me a circular dependency issue. Also the raw object is not always a map, but can be any inherited type of k8s.Resource. I consider this task a quick win, so worked around it by manually identifying all the places where k8s.Resource is the target type to unmarshall.

I'm okay with merge the change as is and let our internal integration tests to verify whether the change makes sense.

PR #1319, opened 2024-03-01
Enable IAM conditions in SpannerDatabase

maqiuyujoyce:
### Change description

Fixes #1317

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [X] Perform necessary E2E testing for changed resources.

Refactored the test framework a bit to enable the local testing in #1406 .

PR #1406, opened 2024-03-21
Enable force-running of test cases in IAM client integration tests

maqiuyujoyce:
### Change description

Enable force-running of test cases in IAM client integration tests to support local testing for IAM configurations.
Diffbase: #1319

### Tests you have done

The following command triggered the integration test to test IAM conditions for SpannerDatabase and it passed.
```
go test -v -tags=integration ./pkg/controller/iam/iamclient/ -test.run TestAllGetSetDeletePolicyMemberWithIAMCondition -must-run-tests spannerdatabase -timeout 900s
```


config/servicemappings/spanner.yaml
```diff
@@ -54,6 +54,7 @@ spec:
         referenceField:
           name: database
           type: name
+        supportsConditions: true
```
justinsb, line 57: Can you rebase so that #1319 drops out of the diff (I assume that's what is going wrong here)

pkg/controller/iam/iamclient/iamclient_integration_test.go
```diff
@@ -53,14 +53,19 @@ func init() {
 	// run-tests and skip-tests allows you to limit the tests that are run by
 	// specifying regexes to be used to match test names. See the
 	// formatTestName() function to see what test names look like.
-	flag.StringVar(&runTestsRegex, "run-tests", "", "run only the tests whose names match the given regex")
+	flag.StringVar(&runTestsRegex, "run-tests", "", "run the tests whose names match the given regex if the tests are allowed")
 	flag.StringVar(&skipTestsRegex, "skip-tests", "", "skip the tests whose names match the given regex, even those that match the run-tests regex")
+
+	// must-run-tests triggers the tests that match the given regex regardless
+	// of other "shouldRun" conditions.
+	flag.StringVar(&mustRunTestsRegex, "must-run-tests", "", "run and only run the tests whose names match the given regex if the regex is specified")
```
justinsb, line 61: What problem are we solving?  Can we use the unified tests instead?

pkg/controller/iam/iamclient/iamclient_integration_test.go
```diff
@@ -94,7 +114,10 @@ func TestAllGetSetDeletePolicy(t *testing.T) {
 	}
 	testName := getCurrentFuncName()
 	shouldRun := func(fixture resourcefixture.ResourceFixture, mgr manager.Manager) bool {
-		return shouldRunBasedOnRunAndSkipRegexes(testName, fixture) && fixture.Type == resourcefixture.Basic && testiam.FixtureSupportsIAMPolicy(t, smLoader, serviceMetadataLoader, dclSchemaLoader, fixture)
+		return shouldRun(testName, fixture,
```
justinsb, line 117: Let's not reuse the `shouldRun` name

pkg/controller/iam/iamclient/iamclient_integration_test.go
```diff
@@ -117,7 +140,11 @@ func TestAllGetSetDeletePolicyWithExternalRef(t *testing.T) {
 	}
 	testName := getCurrentFuncName()
 	shouldRun := func(fixture resourcefixture.ResourceFixture, mgr manager.Manager) bool {
-		return shouldRunBasedOnRunAndSkipRegexes(testName, fixture) && fixture.Type == resourcefixture.Basic && testiam.ShouldRunWithExternalRef(fixture) && testiam.FixtureSupportsIAMPolicy(t, smLoader, serviceMetadataLoader, dclSchemaLoader, fixture)
+		return shouldRun(testName, fixture,
```
justinsb, line 143: Would this be clearer if we have this:

```
if mustRun(test name, fixture) { return true }
...
```

config/servicemappings/spanner.yaml
maqiuyujoyce: Done.

pkg/controller/iam/iamclient/iamclient_integration_test.go
maqiuyujoyce: Without the new flag, we can't trigger a specific test case if it is not within the scope of the "shouldRun". Adding this flag to add more flexibility for local testing.

pkg/controller/iam/iamclient/iamclient_integration_test.go
maqiuyujoyce: Removed.

pkg/controller/iam/iamclient/iamclient_integration_test.go
maqiuyujoyce: Because the new flag is for "must-only-run" scenario, so refactored it a bit differently.

maqiuyujoyce:
@justinsb could you take another look when you get a chance?

maqiuyujoyce:
Will get to it when we need this feature.

PR #1407, opened 2024-03-21
Backporting https://github.com/GoogleCloudPlatform/magic-modules/pull/9463

cheftako:
### Change description

splits off the default API endpoint half of HandleSDKDefaults func into its own function and calls that new function after TPC universe domain is set. This allows the basepath defaults to use the new domain if set

The API custom/default endpoints are not used between when HandleSDKDefaults was called to after the TPC universe logic, so this reorder should be safe. HandleSDKDefaults is also not called anywhere else outside of the one use.

provider: fixed an issue where universe domains would not overwrite API endpoints

### DIFF
--- a/third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/provider/provider.go
+++ b/third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/provider/provider.go
@@ -1945,7 +1945,6 @@ func ProviderConfigure(ctx context.Context, d *schema.ResourceData, p *schema.Pr
 	if err != nil {
 		return nil, diag.FromErr(err)
 	}
-	transport_tpg.HandleDCLCustomEndpointDefaults(d)
 
 	config := transport_tpg.Config{
 		Project:             d.Get("project").(string),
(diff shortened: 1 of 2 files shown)


third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/provider/provider.go
```diff
@@ -2035,6 +2034,12 @@ func ProviderConfigure(ctx context.Context, d *schema.ResourceData, p *schema.Pr
 		}
 	}
 
+	err = transport_tpg.SetEndpointDefaults(d)
```
justinsb, line 2037: I assume this is upstream code?  We would normally try to keep err local, e.g.

```
if err := transport...; err != nil {
```

cheftako: https://github.com/GoogleCloudPlatform/magic-modules/pull/9463/files#diff-4f9f71ccc5b601b48f04a9252d6c74e40b081c78c30265c1cc0277e29f689ed8R757 I'm ok patching this but this is C&P from the listed PR.

PR #1410, opened 2024-03-21
fix:"fix:" move cache location

acpana:
As suggested by Anh, we can move the golangci-lint cache location outside of our repo for now. After developers clean up their `REPO_ROOT/.tmp/` folder  then our script tooling won't be literred by log lines like:

```
...
tmp/golangci-lint/ff/ffbd2d9bad66f8ad457fbd8f0cb7f56a0758e1bdd37547ebfa306386f3704753-a error: lstat .tmp/golangci-lint/ff/ffbd2d9bad66f8ad457fbd8f0cb7f56a0758e1bdd37547ebfa306386f3704753-a: permission denied
...
```

I probably want to fix the core permissions issue in a follow up PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ WEBHOOK_IMG ?= gcr.io/${PROJECT_ID}/webhook:${SHORT_SHA}
 DELETION_DEFENDER_IMG ?= gcr.io/${PROJECT_ID}/deletiondefender:${SHORT_SHA}
 UNMANAGED_DETECTOR_IMG ?= gcr.io/${PROJECT_ID}/unmanageddetector:${SHORT_SHA}
 # Detects the location of the user golangci-lint cache.
-GOLANGCI_LINT_CACHE := $(shell pwd)/.tmp/golangci-lint
+GOLANGCI_LINT_CACHE := /tmp/golangci-lint
 # When updating this, make sure to update the corresponding action in
 # ./github/workflows/lint.yaml
 GOLANGCI_LINT_VERSION := v1.56.2


Makefile
```diff
@@ -21,7 +21,7 @@ WEBHOOK_IMG ?= gcr.io/${PROJECT_ID}/webhook:${SHORT_SHA}
 DELETION_DEFENDER_IMG ?= gcr.io/${PROJECT_ID}/deletiondefender:${SHORT_SHA}
 UNMANAGED_DETECTOR_IMG ?= gcr.io/${PROJECT_ID}/unmanageddetector:${SHORT_SHA}
 # Detects the location of the user golangci-lint cache.
-GOLANGCI_LINT_CACHE := $(shell pwd)/.tmp/golangci-lint
+GOLANGCI_LINT_CACHE := /tmp/golangci-lint
```
anhdle-sso, line 24: We can use /tmp/k8s-config-connector/golangci-lint  in case other repo also use golangci-lint

anhdle-sso: NIT: I am not sure if we need to do mkdir -p for the cache folder under make lint section at line 98.

acpana: "__technically__", the golangi-ci lint cache can be shared across repoes and the official docs don't point out a need to separate on a per repo basis. If there's strong opinions I can rename it.

Also re the nit, I don't think we need to mkdir and when I tried it out before the PR that statement was validated so I think we are good.

anhdle-sso: We can change to folder if needed later.

maqiuyujoyce:
/lgtm
/approve

Thank you for the change. Will try it out after it's merged!

PR #1414, opened 2024-03-21
chore: upgrade dcl to v1.62

acpana:
Upgrade to DCL v1.62; [release notes](https://github.com/GoogleCloudPlatform/declarative-resource-client-library/releases/tag/v1.62.0)

Instructions as per https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/README.UpdatingDCL.md where I pinned the v1.62 version.


pkg/dcl/metadata/metadata.go
```diff
@@ -232,6 +232,11 @@ var serviceList = []ServiceMetadata{
 				// This resource is implemented through Terraform/KCC bridge, only load its DCL OpenAPI schema for resource references
 				Releasable: false,
 			},
+			{
+				Kind:       "ComputeNetworkAttachment",
+				DCLType:    "NetworkAttachment",
+				Releasable: false,
+			},
```
acpana, line 239: This missing entry was causing the CRD generation before.

acpana:
Tests ran locally:

`eventarc`:

```
# samples
$ go test -v -tags=integration ./config/tests/samples/create -timeout 1600s -test.count=1 -test.run TestAll -run-tests eventarc 2>&1
...
--- PASS: TestAll (1.19s)
    --- PASS: TestAll/eventarctrigger (645.36s)
PASS
...
ok      github.com/GoogleCloudPlatform/k8s-config-connector/config/tests/samples/create 714.968s

# dynamic
$ go test -v -tags=integration ./pkg/controller/dynamic/ -timeout 1600s -test.count=1 -test.run TestCreateNoChangeUpdateDelete -run-tests eventarc 2>&1
...
--- PASS: TestCreateNoChangeUpdateDelete (0.08s)
    --- PASS: TestCreateNoChangeUpdateDelete/eventarc (0.00s)
        --- PASS: TestCreateNoChangeUpdateDelete/eventarc/basic-eventarctrigger (172.16s)
PASS
...
ok      github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/dynamic      181.516s
```

`logginglogbucket`:

TODO needs some TLC.

pkg/dcl/metadata/metadata.go
acpana: @justinsb this missing entry was actually causing the CRD generation failure. AFAIU now that failure was to be expected 😅 . I can add this gotcha to our docs for upgrading the lib if that helps.

Also note, this does not upgrade us to latest DCL lib, which last I checked was 1.63 which fails as w a deepcopy issue. I filed #1427 to track that

acpana:
@justinsb I was able to run most dynamic and samples [caveat below] tests for the resources that are changing with 1.62 dcl upgrade. I was missing setting the right vars initially for the `logginglogbucket` to run locally and jingyi pointed that out to me.

---

caveat: one samples scenario is not run under `samples/resources/logginglogbucket/billing-account-log-bucket` due to a permissions issues on the billing account. However, I think we have enough confidence to say that changes to the `logginglogbucket` are not causing regressions. Wdyt?

maqiuyujoyce:
Thank you for verifying using the integration tests, @acpana ! /lgtm

> caveat: one samples scenario is not run under samples/resources/logginglogbucket/billing-account-log-bucket due to a permissions issues on the billing account. However, I think we have enough confidence to say that changes to the logginglogbucket are not causing regressions. Wdyt?

Agreed it's likely restricted by your account's permissions (though I still suggest you try to take a look what's missing when you don't have other priorities).

Though I am curious: What's the reason (what features to support/bugs to fix) we upgrade to DCL 1.62? Have you verified the feature is properly supported or bug is fixed in the latest version? Just want to double check because I didn't find any context in the comments / changes in testdata. If there is no specific feature/bugfix to verify, then it's ready to be merged. Otherwise, we probably want to cover the new feature with some testdata/sample changes.

acpana:
> What's the reason (what features to support/bugs to fix) we upgrade to DCL 1.62? [...] If there is no specific feature/bugfix to verify, then it's ready to be merged.

AFAICT/ as far as i'm aware, there is no specific reason to upgrade to 1.62 other than trying to get the `make upgrade-dcl && ready-pr` flow unfixed. This PR upgrades us to 1.62. Then I have #1427 to track 1.63 upgrade failure.

Leaving things as is today (i.e. with `make upgrade-dcl && ready-pr` broken) can impact contributors who want to add/ modify DCL based resources.

PR #1422, opened 2024-03-25
genapi: Add support for TagsTagKey

justinsb:
Also fix up Update method to match Create method.

### DIFF
--- a/pkg/controller/direct/directbase/interfaces.go
+++ b/pkg/controller/direct/directbase/interfaces.go
@@ -42,9 +42,11 @@ type Adapter interface {
 
 	// Create creates a new GCP object.
 	// This should only be called when Find has previously returned false.
+	// The implementation should write the updated status into `u`.
 	Create(ctx context.Context, u *unstructured.Unstructured) error
 
 	// Update updates an existing GCP object.
 	// This should only be called when Find has previously returned true.
-	Update(ctx context.Context) (*unstructured.Unstructured, error)
+	// The implementation should write the updated status into `u`.
+	Update(ctx context.Context, u *unstructured.Unstructured) error
 }
--- a/pkg/controller/direct/resourcemanager/tagkey_controller.go
+++ b/pkg/controller/direct/resourcemanager/tagkey_controller.go
@@ -1,0 +1,252 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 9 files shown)


justinsb:
I'm going to rebase this, but 

/assign @acpana

pkg/controller/direct/directbase/interfaces.go
```diff
@@ -46,5 +46,5 @@ type Adapter interface {
 
 	// Update updates an existing GCP object.
 	// This should only be called when Find has previously returned true.
-	Update(ctx context.Context) (*unstructured.Unstructured, error)
+	Update(ctx context.Context, u *unstructured.Unstructured) error
```
acpana, line 49: just to highlight, we are expecting that implementors will use `u` to write the status to it, right?

pkg/controller/direct/resourcemanager/tagkey_controller.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	a.actual = tagKey
+
+	return true, nil
+}
+
+// Delete implements the Adapter interface.
+func (a *tagKeyAdapter) Delete(ctx context.Context) (bool, error) {
+	// Already deletd
```
acpana, line 122:
```suggestion
	// Already deleted
```

pkg/controller/direct/resourcemanager/tagkey_controller.go
```diff
@@ -0,0 +1,252 @@
...
+	req := &pb.DeleteTagKeyRequest{
+		Name: a.fullyQualifiedName(),
+	}
+
+	op, err := a.tagKeysClient.DeleteTagKey(ctx, req)
+	if err != nil {
+		if IsNotFound(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("deleting tagKey: %w", err)
```
acpana, line 137: nit: here and elsewhere where it makes sense, do we want to also include the `tagKey` name? Assuming it doesn't bubble up in the direct controller.
```suggestion
		return false, fmt.Errorf("deleting tagKey %s: %w", a.fullyQualifiedName(), err)
```

acpana (commented):
/hold for author to merge
/lgtm
/approve

Thanks for rebasing this 🙏🏼 💯 

I'd ignore the `license-lint` failure:

1) Fixed it in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1534
2) Verified it's not a real issue locally (ran w the fix).

pkg/controller/direct/directbase/interfaces.go
justinsb: Right.  Specifically we write to the Status.  I think in another PR I added more docs, I'll try to track those down!

acpana:
/lgtm
/approve

thanks for addressing comments

PR #1424, opened 2024-03-25
fix:test: regex against dir

acpana:
It seems like the intent in https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/README.NewResourceFromTerraform.md#run-tests is that we can run tests under a folder. For instance:

```
pkg/test/resourcefixture/testdata/basic
├── pubsub
|   ├── v1beta1
|       ├── pubsubsubscription
|           ├── basicpubsubsubscription
|               └── create.yaml
|               └── dependencies.yaml
|               └── update.yaml
|           ├── bigquerypubsubsubscription
|               └── create.yaml
|               └── dependencies.yaml
|               └── update.yaml
```

It seems like we should be able to run `pubsub` but `run-tests` actually checks the NAME of the test. So if the test does not contain `pubsub` in it, it will not match the regex. This change mostly switches the matching to happen against the source directory (which contains the name of the test) as opposed to just the name.

### DIFF
--- a/pkg/controller/dynamic/dynamic_controller_integration_test.go
+++ b/pkg/controller/dynamic/dynamic_controller_integration_test.go
@@ -116,7 +116,7 @@ func shouldRunBasedOnRunAndSkipRegexes(parentTestName string, fixture resourcefi
 
 	// If a run-tests regex has been provided and it doesn't match the test name, skip the test.
 	if runTestsRegex != "" {
-		if !regexp.MustCompile(runTestsRegex).MatchString(testName) {
+		if !regexp.MustCompile(runTestsRegex).MatchString(fixture.SourceDir) {
 			return false
 		}
 	}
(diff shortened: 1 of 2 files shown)


justinsb:
I think @maqiuyujoyce has some in flight PRs grappling with these regexes also.

Should we just recommend using the unified tests?  Also, hopefully we aren't writing too many more new TF resources?

maqiuyujoyce:
`-run-tests` flag is widely used in our integration tests configuration, and I'd not recommend changing it's functionality unless we are sure the change doesn't impact the behavior of all the integration tests (and other places where it is called). @acpana let me know if you need links to those configurations, they are on GoB.

The current behavior of `-run-tests` is to match test cases by the test name, i.e. the last section of the path to the test case. 
I don't recall whether we've intended to invoke a suite of integration tests by API/high-level folder name, but feel free to support it as a separate feature if needed. Or I can update the doc to reflect the current situation.

In my opinion, we either should have a better way to trigger tests by test name, target Kind, or target Service separately; or we should have a unit test to regulate the naming convention for the testdata folder names so that our current regex match works as expected.

acpana:
Might still be something worth exploring but closing this out as I am not working on this actively

PR #1426, opened 2024-03-25
Control caching by kind using exclude list

justinsb:
I don't think the other approach was actually configuring caching,  although unstructured objects are never cached (by default).


justinsb:
Merged as part of a larger PR

/close

PR #1429, opened 2024-03-26
WIP: tests: Verify the kubernetes watches we open

justinsb:
We want to check that we aren't opening full watches on all objects,
also that we aren't opening a full and a metadata-only watch.


justinsb:
Merged as part of a larger PR

/close

PR #1438, opened 2024-03-26
tests: add tests for OutputOnlyFieldsAreUnderObservedState

justinsb:
- chore: refactor OutputOnlyFieldsAreUnderObservedState and supportsStateIntoSpecMerge
- tests: add tests for OutputOnlyFieldsAreUnderObservedState


pkg/k8s/allowlist_test.go
```diff
@@ -15,35 +15,86 @@
...
+				Version: version.Name,
+				Kind:    crd.Spec.Names.Kind,
+			}
+			t.Run(fmt.Sprintf("%s/%s/%s", gvk.Group, gvk.Version, gvk.Kind), func(t *testing.T) {
+				openAPISchema := version.Schema.OpenAPIV3Schema
+				prop := findOpenAPIProperty(openAPISchema, "status", "observedState")
+				hasObservedState := prop != nil
+				got := OutputOnlyFieldsAreUnderObservedState(gvk)
+
+				if hasObservedState != got {
```
maqiuyujoyce, line 82: The test case fails because observedFields configuration needs to be taken into consideration as well: e.g. https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/config/servicemappings/compute.yaml#L36 .

pkg/crd/crdloader/crdloader.go
```diff
@@ -138,7 +137,13 @@ func isMatch(group, version, kind string, crd apiextensions.CustomResourceDefini
 		}
 	}
 	if version != "" {
-		if k8s.GetVersionFromCRD(&crd) != version {
+		foundVersion := false
```
maqiuyujoyce, line 140: Nit: Probably easier for readability and reusability if it is a util function (e.g. `ContainsVersionInCRD(crd, version)`) under k8s package.

pkg/k8s/allowlist_test.go
```diff
@@ -15,35 +15,86 @@
...
-			actualResult := supportsStateIntoSpecMergeInKind(tc.kind)
+			actualResult := supportsStateIntoSpecMerge(tc.gvk)
 			if actualResult != tc.expectedResult {
 				t.Fatalf("got %v, want %v", actualResult, tc.expectedResult)
 			}
 		})
 	}
 }
+
+func TestOutputOnlyFieldsAreUnderObservedState(t *testing.T) {
```
maqiuyujoyce, line 64: Hmmm, after some thinking, I propose this test should test "output-only fields are under status" situation.

CRD existing in the allowlist means that:
(1) currently all its output-only fields are directly under `status`,
(2) but it's still possible for it to have `status.observedState` with the observed fields configuration.
-> CRD existing in the allowlist should not have `status.observedState` unless it supports observed fields.

CRD not existing in the allowlist means that:
(1) currently all its output-only fields are under `status.observedState`,
(2) but it's still possible for it to not have any output-only fields, thus not have `status.observedState`.
-> It's valid for CRD not existing in the allowlist to have or not have `status.observedState`, and this is not easily detectable with our current code logic.

acpana:
@justinsb would you mind rebasing ?

ldanielmadariaga:

PR #1440, opened 2024-03-26
chore: vendor in deps

acpana:
> reviewer notes: I'll use this WIP PR as a tracking ticket

* this is mostly a `go mod tidy && go mod vendor`
* also don't gitignore the vendor directory

TODO:

* [ ] solve any issues


justinsb:
I for one was not sad to see vendor gone :-)

acpana:
I think we are going to punt/ wait on this to see if we actually need to vendor our deps back in.

The good news is that it seems this is not too bad.

PR #1434, opened 2024-03-26
mockgcp: Add support for TagsTagValue

justinsb:
- mockgcp: Implement TagsTagValue
- tests: run deletion in order when we're doing golden testing

Note that we also enforce strict deletion ordering when doing golden testing, so that the http log is consistent.

PR #1446, opened 2024-03-27
tests: run create/delete in order when we're doing golden testing

justinsb:
This does mean that we have to have the objects pre-ordered correctly,
but it means that these requests will be made in a consistent order,
making our golden testing easier.


acpana:
I think this work went in with https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1434/commits right ?

justinsb:
> I think this work went in with https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1434/commits right ?

Partially, I think we still need creation ordering if we want test consistency.  I'll rebase and we can see what's left.

justinsb:
Rebased & cleaned this up, it's much shorter now!

acpana:
/lgtm
/approve

thanks for sticking with it 💯 !!

PR #1447, opened 2024-03-27
tests: Use golden test to validate KRM object fields

gemmahou:
Add test to cover code changes made in #1353 

1. yaml files in this PR were generated. 
2. Currently to save time, I run the golden object test against mock GCP. However, some allowlisted resources([ref](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/525a99fb313fe78ce84a1b8e3798060bfab97a93/pkg/k8s/allowlist.go)) are not supported in mockGCP yet.
3. We'll verify the created object vs golden file when "GOLDEN_OBJECT_CHECKS" is specified.

Next step:
I plan to set up a Prow job to run the golden object test. So we can run it as part of our CI workflow and against real GCP to get more test coverage on resources.

created b/332777103 to improve our unified_test

Test command:
E2E_KUBE_TARGET=envtest RUN_E2E=1 E2E_GCP_TARGET=mock WRITE_GOLDEN_OUTPUT=1 GOLDEN_OBJECT_CHECKS=1 go test -test.count=1 -timeout 3600s -v ./tests/e2e -run TestAllInSeries/fixtures


justinsb:
I like this output, this will really give us a lot of confidence - thank you!

I _think_ object_test is a copy of unified_test?  (Is that right?)  If we just added to unified_test, what is missing there today?

I do think it's a good idea to use the occasion to refactor out pieces of unified_test at the same time, though I think export is already factored into its own method (if not file)

gemmahou:
> I like this output, this will really give us a lot of confidence - thank you!
> 
> I _think_ object_test is a copy of unified_test? (Is that right?) If we just added to unified_test, what is missing there today?
> 
> I do think it's a good idea to use the occasion to refactor out pieces of unified_test at the same time, though I think export is already factored into its own method (if not file)

Yes, most of the test code is copy-pasted from the unified test to make it clear when I test it locally. Let me see if it looks better to put it back.

I re-wrote the "compareGoldenFile" method for this specific test. I think we care more about the schema rather than the field values, so I only check the fields in the yaml match those in the golden file. This saves us some effort in modifying unique values.

To avoid duplicated code, I'm thinking we can create a test_util or something similar to help us reuse code. I don't have a complete plan for doing that yet.

gemmahou:
> > To avoid duplicated code, I'm thinking we can create a test_util or something similar to help us reuse code. I don't have a complete plan for doing that yet.
> 
> We do now have more files in tests/e2e. Because the whole package is for testing, we can just add files to the package.
> 
> One big reason to prefer putting it into unified_tests that occurred to me: we don't have to run them twice. It doesn't matter with mockgcp or VCR, but if we're running for real it is an advantage to just have one test.
> 
> To reduce the size of the method, I suspect a good approach might be to pull more logic into separate methods as we did with `exportResource` https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/tests/e2e/unified_test.go#L280
> 
> (BTW I realized that I've been using "export" to mean two things, both "read GCP and create a KCC object", which is the traditional meaning of export, but also "kubectl get and write to disk". We might need a need name for the 2nd to disambiguate - maybe "dump object"?)

`One big reason to prefer putting it into unified_tests that occurred to me: we don't have to run them twice.  It doesn't matter with mockgcp or VCR, but if we're running for real it is an advantage to just have one test.`

It makes sense. For now, the unified test, or TestAllInSeries, includes mockGCP test, VCR test, pause test, golden KRM object test, and golden HTTP test. If one of them fails, we need to provide clear information for engineers to understand which resource fails on which test. We have a good error message, though.

Regarding setup GH action for golden object test, I feel that's covered in presubmit/[mock-test](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/22bfdf2344878f9b811be46df9a4739d708c9965/scripts/github-actions/ga-mock-test.sh#L35) job, we don't need to run it twice. However, the name `mock-test` might be confusing, it contains both mockGCP test and golden KRM object test.

justinsb:
>  it contains both mockGCP test and golden KRM object test.

"mock" is supposed to mean mock-gcp instead of real-gcp.

I see golden tests primarily as an implementation of our tests - we could do the same thing by checking every field :-)

Maybe we rename `mock-test` ->  `mockgcp-tests` ?

pkg/test/utils.go
```diff
@@ -140,3 +143,63 @@ func IgnoreLeadingComments(s string) string {
...
+	if err != nil {
+		return err
+	}
+	if ok, err := compareNestedFields(wantMap, gotMap); !ok {
+		return err
+	}
+	return nil
+}
+
+func compareNestedFields(wantMap, gotMap map[string]interface{}) (bool, error) {
```
justinsb, line 177: You can probably use cmp.Diff for this, but ... is the output any better than using cmp.Diff on the file (and then `git diff` if we write the file)?

tests/e2e/unified_test.go
```diff
@@ -273,11 +274,42 @@ func testFixturesInSeries(ctx context.Context, t *testing.T, testPause bool, can
...
+						var testName string
+						if len(pieces) > 0 {
+							testName = pieces[len(pieces)-1]
+						} else {
+							t.Errorf("failed to get test name")
+						}
+						// Golden test exported GCP object
+						got := exportResource(h, obj)
+						if got != "" {
+							expectedPath := filepath.Join(dir, fmt.Sprintf("export/_%v.yaml", testName))
```
justinsb, line 291: Shall we just write it to the fixture.SourceDir, so it lives with the test inputs ?

tests/e2e/unified_test.go
```diff
@@ -273,11 +274,42 @@ func testFixturesInSeries(ctx context.Context, t *testing.T, testPause bool, can
...
+						} else {
+							if err := normalizeObject(u, project, uniqueID); err != nil {
+								t.Fatalf("error from normalizeObject: %v", err)
+							}
+							got, err := yaml.Marshal(u)
+							if err != nil {
+								t.Errorf("failed to convert KRM object to yaml: %v", err)
+							}
+							// File name looks like _testname.yaml
+							expectedPath := filepath.Join(dir, fmt.Sprintf("object/_%v.yaml", testName))
```
justinsb, line 307: Ditto here, fixture.SourceDir feels more natural to me - WDYT?

tests/e2e/unified_test.go
```diff
@@ -273,11 +274,42 @@ func testFixturesInSeries(ctx context.Context, t *testing.T, testPause bool, can
...
+							err = test.CompareGoldenObject(expectedPath, got)
```
justinsb, line 308: Nit: personally I think it's fine to pass `t` in: it makes it clear this is a test method, and it means we can e.g. call t.Errorf or t.Fatalf

tests/e2e/export.go
```diff
@@ -15,6 +15,7 @@
 package e2e
 
 import (
+	"k8s.io/apimachinery/pkg/types"
```
maqiuyujoyce, line 18: Could you run `goimports -w tests/` to fix the import grouping?

I also created #1497 .

tests/e2e/normalize.go
```diff
@@ -44,6 +44,7 @@ func normalizeObject(u *unstructured.Unstructured, project testgcp.GCPProject, u
 
 	visitor.replacePaths = map[string]any{}
 	visitor.replacePaths[".metadata.deletionTimestamp"] = "1970-01-01T00:00:00Z"
+	visitor.replacePaths[".status.creationTimestamp"] = "1970-01-01T00:00:00Z"
```
maqiuyujoyce, line 47: I'm not very familiar with this part of the code, but this seems to be reused by a few test flows. Have you verified that other test flows are not impacted?

tests/e2e/unified_test.go
```diff
@@ -25,6 +25,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"sigs.k8s.io/yaml"
```
maqiuyujoyce, line 28: Similarly, could you run `goimports -w tests/`?

tests/e2e/unified_test.go
maqiuyujoyce: I'm leaning towards we separate human-crafted files vs. generated-only files as it could distract the contributors. Though I do think having test data for the same resource to be centralized at one place makes sense. We could start with a more distinguishable name like `_generated_[testname].golden.yaml`. And in the long term, maybe creating a subfolder (we need to figure out a better regex to trigger individual test cases) called `generated` under each test case folder to hold all the generated test data.

tests/e2e/unified_test.go
maqiuyujoyce: As this is the second PR I reviewed for unified_test.go, I'm still getting used to the coding style here.

IMO `CompareGoldenFile()` or `ComputeGoldenObject()` don't look like a test util function due to their vague names. It's reasonable or even preferred that we pass in `t` in a test util function, but I think the name should be more specific and test-related, e.g. `AssertRetrievedYAMLMatchesGoldenFile`, `AssertKRMObjectMatchesGoldenObject`.

tests/e2e/unified_test.go
maqiuyujoyce: I'm curious, why is it `natural`? Is there any design pattern or best practice related to it? My intuition is actually more aligned with @gemmahou 's design choice so would like to learn more about it.

maqiuyujoyce (commented):
@justinsb just to clarify, this PR is blocking #1465 , right?

pkg/test/utils.go
gemmahou: Thanks, I took a look at cmp.Diff method. I think what I’m trying to do here is to verify the existence of fields in the object; I don’t think we really care about the actual value in each field?

The values might be generated (like date or timestamp), we need to replace them with a fixed value for golden comparison. But for this case, I think we only want to verify that all fields are the same when compared to the golden file, please correct me if I’m wrong...

tests/e2e/unified_test.go
gemmahou: My intention was to put all generated files used for tests (i.e. vcr cassettes, golden files) together under a separate folder. I agree that the folder names and file names can be clearer so that we know they are generated files and used for vcr/golden testing.

gemmahou:
> @justinsb just to clarify, this PR is blocking #1465 , right?

I would like to learn more about the acceptance criteria to unblock #1465. Currently, we don't run golden tests in GH presubmit or test probes, and we need golden file coverage for all test cases, I guess (I only covered a few tests that support mockGCP in this PR). I'm thinking we can check in #1452, so at least we have a certain level of testing to unblock #1465.

tests/e2e/unified_test.go
justinsb: I like the underscore convention - that works for me!

justinsb:
>  Is there any design pattern or best practice related to it?

Yeah, it's called golden testing.  Most of the docs are paywalled but this is a reasonable intro: https://ro-che.info/articles/2017-12-04-golden-tests

Putting the golden files alongside the tests works well e.g. in PR reviews, or because you will often look at the http output alongside the inputs.  A subdirectory works fine, but it probably overkill compared to just the underscore convention - I like that suggestion!

tests/e2e/normalize.go
gemmahou: This function replaces random values with a fixed string so that the values are identical and predictable during a golden comparison. This function is only used for golden testing and would not affect anything:
- If the item to be normalized does not have `status. creationTimestamp`  field, nothing changes.
- If the item to be normalized has `status. creationTimestamp` field, it's necessary to replace it or the golden test might fail.

tests/e2e/export.go
```diff
@@ -95,3 +96,13 @@ func exportResourceAsUnstructured(h *create.Harness, obj *unstructured.Unstructu
 	}
 	return u
 }
+
+func getKRMObject(h *create.Harness, obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
```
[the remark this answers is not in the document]
gemmahou: The goal here is to move some logic out of unified_test so that it doesn't become larger and larger, and to improve readability. I believe it is acceptable to put this brief logic back into unified_test. I borrowed the code from [script_test.go](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/tests/e2e/script_test.go#L149)

I think we'll need to refactor the unified_test:
1. reorganize and document the usage of env variables(b/332719149)
2. move reusable code to util files

tests/e2e/unified_test.go
gemmahou: As mentioned in [here](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1447#discussion_r1550645435), I think it's easier to address Joyce's comment in another task when we do a refactor of unified_test. I created a ticket b/332777103.

pkg/test/utils.go
maqiuyujoyce: I thought we do care both the fields the values? Eventually the replay tests will be applied to all the test cases to verify if the KCC logic is executed as expected or not, and being able to verify the status fields are more of a "side effect" that we can leverage immediately.

maqiuyujoyce:
/lgtm

I think we haven't reached an agreement on the compareNestedFields() function though I think it is not blocking. @justinsb would like your input!

tests/e2e/unified_test.go
```diff
@@ -285,11 +288,41 @@ func testFixturesInSeries(ctx context.Context, t *testing.T, testPause bool, can
...
+						var testName string
+						if len(pieces) > 0 {
+							testName = pieces[len(pieces)-1]
+						} else {
+							t.Errorf("failed to get test name")
+						}
+						// Golden test exported GCP object
+						got := exportResource(h, obj)
+						if got != "" {
+							expectedPath := filepath.Join(fixture.SourceDir, fmt.Sprintf("_generated_export_%v.golden", testName))
```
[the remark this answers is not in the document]
justinsb: Actually, never mind - this was already the case!

justinsb:
Looks good, it's a test so we can deal with my nits if we encounter them in practice (e.g. multiple objects, and we should also try just using the normal golden file function without special casing for an object)

/approve
/lgtm

PR #1449, opened 2024-03-27
chore: create build scripts

justinsb:
These are (partially) simplified forms of our internal build scripts,
trying to bring them to OSS.

Note that we will treat building separately from pushing to a
registry/GCS.


gemmahou:
/lgtm
/approve
/hold
Feel free to remove the hold when we are ready to merge this.

PR #1451, opened 2024-03-28
tests: add step wise for CCC Pause

acpana:
Adds a first step wise based test for pause/ unpause with a CCC.

generated artifacts by running:

```bash
$ WRITE_GOLDEN_OUTPUT=1 GOLDEN_REQUEST_CHECKS=1 E2E_KUBE_TARGET=envtest E2E_GCP_TARGET=real RUN_E2E=1 \
go test -test.count=1 -timeout 360s -v ./tests/e2e -run TestE2EScript 2>&1
```

### DIFF
--- a/tests/e2e/script_test.go
+++ b/tests/e2e/script_test.go
@@ -20,6 +20,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 	"time"
 
@@ -33,6 +34,7 @@ import (
 	"sigs.k8s.io/yaml"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 )
 
@@ -49,6 +51,12 @@ func TestE2EScript(t *testing.T) {
 		cancel()
 	})
 
+	type gvkNN struct {
+		gvk schema.GroupVersionKind
+		nn  types.NamespacedName
+	}
+
+	logCheckTimeout := 10 * time.Second
 	t.Run("scenarios", func(t *testing.T) {
 		scenarioDir := "testdata/scenarios"
 		scenarioPaths := findScripts(t, scenarioDir)
@@ -78,8 +86,10 @@ func TestE2EScript(t *testing.T) {
 				var eventsByStep [][]*test.LogEntry
 
 				eventsBefore := h.Events.HTTPEvents
-				for i, obj := range script.Objects {
+				// tracks the set of all applied objects as keyed by a gvk, namespaced name tuple
+				appliedObjects := map[gvkNN]*unstructured.Unstructured{}
 
+				for i, obj := range script.Objects {
 					testCommand := ""
 					v, ok := obj.Object["TEST"]
 					if ok {
@@ -91,12 +101,25 @@ func TestE2EScript(t *testing.T) {
 
 					exportResource := obj.DeepCopy()
 					shouldGetKubeObject := true
+					k := gvkNN{
+						gvk: obj.GroupVersionKind(),
+						nn: types.NamespacedName{
+							Name:      obj.GetName(),
+							Namespace: obj.GetNamespace(),
+						},
+					}
 
 					switch testCommand {
 					case "APPLY":
 						applyObject(h, obj)
 						create.WaitForReady(h, obj)
 
+						appliedObjects[k] = obj
+					case "APPLY-NO-WAIT":
+						applyObject(h, obj)
+						appliedObjects[k] = obj
+						exportResource = nil
+						shouldGetKubeObject = false
 					case "DELETE":
 						create.DeleteResources(h, create.CreateDeleteTestOptions{Create: []*unstructured.Unstructured{obj}})
 						exportResource = nil
@@ -112,13 +135,41 @@ func TestE2EScript(t *testing.T) {
 						// The object probably still exists (that's probably
 						// why we're using DELETE-NO-WAIT), so export the
 						// resource and the kube object.
+					case "CHECK-LOG":
+						applyObject(h, obj)
+
+						val, ok := obj.Object["VALUE_PRESENT"]
+						if !ok {
+							t.Fatalf("did not find key VALUE_PRESENT in step")
+						}
+						sval := val.(string)
+
+						ticker := time.NewTicker(1 * time.Second)
+						for {
+							stopWaiting := false
+							select {
+							case <-time.After(logCheckTimeout):
+								t.Fatalf("timed out looking for value %s in http log", sval)
+								stopWaiting = true
+							case <-ticker.C:
+								// todo(acpana): find better asympotatic approach
+								for _, l := range h.Events.HTTPEvents {
+									if strings.Contains(l.Response.Body, sval) {
+										stopWaiting = true
+										break
+									}
+								}
+							}
+							if stopWaiting {
+								break
+							}
+						}
 
 					case "ABANDON":
 						setAnnotation(h, obj, "cnrm.cloud.google.com/deletion-policy", "abandon")
 						create.DeleteResources(h, create.CreateDeleteTestOptions{Create: []*unstructured.Unstructured{obj}})
 						// continue to export the resource
 						shouldGetKubeObject = false
-
 					default:
 						t.Errorf("unknown TEST command %q", testCommand)
 						continue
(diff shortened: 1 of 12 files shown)


tests/e2e/testdata/scenarios/ccc_pause_change_reconcile/_http02.log
acpana, line 1: this being an empty file is consequential because it means that while KCC was paused to interaction happened with the cloud provider, hence no actuation happened. Checking that this file is empty might be sufficient as such.

tests/e2e/script_test.go
```diff
@@ -36,6 +38,14 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
+var runScenarios string
```
acpana, line 41: i think i want this to be a repeatable flag;

then we can run a bunch of pause related scenarios under the pause CI step

justinsb: My _personal_ view is that (1) flags are a pain to pass to tests, and we want (2) `go test ./...` to cover the "core" scenarios.  If you want to override a kubeconfig, by all means use an env var or a flag, but `go test ./...` should be what we run in the default CI.

In this case, I think the `-run`  flag on `go test` does this?

tests/e2e/script_test.go
```diff
@@ -119,6 +135,23 @@ func TestE2EScript(t *testing.T) {
 						// continue to export the resource
 						shouldGetKubeObject = false
 
+					case "CHECK-INTENT":
```
justinsb, line 138: Nit: I don't know what CHECK-INTENT does - is there a version of this name that is more self-documenting?

tests/e2e/script_test.go
```diff
@@ -192,7 +225,17 @@ func TestE2EScript(t *testing.T) {
 
 				}
 
-				create.DeleteResources(h, create.CreateDeleteTestOptions{Create: script.Objects})
+				objSet := []*unstructured.Unstructured{}
+				seen := map[string]struct{}{} // the key for the set will be the object name
```
justinsb, line 229: That's kind of dangerous, because if I have two objects of different kinds but the same name ...

I'm not entirely sure why the change here TBH (i.e. why aren't deleting all the objects).  Is it because we want to skip the CHECK-INTENT objects?  If so, maybe we could build up the objects to delete as we loop through the steps?

tests/e2e/testdata/scenarios/ccc_pause_change_reconcile/_http02.log
justinsb: Yes, I think this works!  It's not entirely self-documenting, but we could just put a README.md in the ccc_pause_change_reconcile directory (that's probably easier than e.g. ignoring comments and putting the comment in this file)

acpana: I started guarding against steps that we don't want logs for to reduce noise and added a README

tests/e2e/script_test.go
acpana: Reworked this and added an entry in the README

tests/e2e/script_test.go
acpana: I think I was being lazy 😅 ; I created a key by the (gvk, namespacedName) tuple instead.

tests/e2e/script_test.go
```diff
@@ -49,6 +52,12 @@ func TestE2EScript(t *testing.T) {
 		cancel()
 	})
 
+	type gvkByNN struct {
```
justinsb, line 55: Nit: technically it's GVKNN, (no "by"), which is actually an acronym I've used before.  You could use GroupVersionKindNamespaceName if you were worried that GVKNN was not very clear, but personally I think it's very clear once you're told and GVKNN is a little shorter to type & read!

tests/e2e/script_test.go
```diff
@@ -95,7 +107,25 @@ func TestE2EScript(t *testing.T) {
 					switch testCommand {
 					case "APPLY":
 						applyObject(h, obj)
-						create.WaitForReady(h, obj)
+						_, noWait := obj.Object["NO_WAIT"]
```
justinsb, line 110: Any reason not to have a different command APPLY-NO-WAIT (to mirror DELETE-NO-WAIT).

My gut was that we might end with commands like `APPLY(wait=false)` or something like that, but I wanted to see how it evolved.

In that spirit though, let's see where it goes... it's a test and doesn't escape our code!

tests/e2e/script_test.go
```diff
@@ -112,13 +142,41 @@ func TestE2EScript(t *testing.T) {
 						// The object probably still exists (that's probably
 						// why we're using DELETE-NO-WAIT), so export the
 						// resource and the kube object.
+					case "CHECK-LOG":
```
justinsb, line 145: I'm still not very clear what CHECK-LOG is doing...

tests/e2e/script_test.go
```diff
@@ -78,8 +87,11 @@ func TestE2EScript(t *testing.T) {
 				var eventsByStep [][]*test.LogEntry
 
 				eventsBefore := h.Events.HTTPEvents
-				for i, obj := range script.Objects {
+				// tracks the set of all applied objects as keyed by a gvk, namespaced name tuple
+				appliedObjects := map[gvkByNN]*unstructured.Unstructured{}
+				ignoreStepEvents := map[int]struct{}{} // ignore steps for which we don't records events for like CC/ CCC
```
justinsb, line 92: I don't think I understand why we need to skip this; if we write an empty file we are asserting that there are no events... We would skip if there was unavoidable non-determinism...

tests/e2e/script_test.go
```diff
@@ -201,7 +265,12 @@ func TestE2EScript(t *testing.T) {
 }
 
 func applyObject(h *create.Harness, obj *unstructured.Unstructured) {
-	if err := h.GetClient().Patch(h.Ctx, obj, client.Apply, client.FieldOwner("kcc-tests"), client.ForceOwnership); err != nil {
+	o := obj.DeepCopy()
+	delete(o.Object, "NO_WAIT")
+	delete(o.Object, "TEST")
```
justinsb, line 270: Good point ... any idea how was I getting away with no clearing this before?

tests/e2e/script_test.go
acpana: Ah yes I can change to `APPLY-NO-WAIT` 👍🏼

tests/e2e/script_test.go
acpana: I think it's because of the default being `TEST: APPLY` and not having to use write it in 😆

tests/e2e/script_test.go
acpana: parking some thoughts here

> what CHECK-LOG is doing...

Would there be a better name for it?

---

I think there's still some value in having a mechanism to check this scenario: "when CC/ CCC start reconciling again via an APPLY, let's check that the logs show KCC resources start getting reconciled onto the cloud provider again". IIRC, if it's sufficient to check the logs for when we apply CC/ CCC we could get away with just having the golden checks. But at present apply CC/ CCC is "instantaneous" and we move on without recording any http events. Wdyt ?

---

IIRC, we did chat about forcing the reconcile on the KCC object (ArtifactRegistryRepository here). That approach still requires some timing/ waiting that for the logs to get captured as part of apply a CC/ CCC resource that turns on `Reconciling`. 

On a separate note, I think we should add an option to turn off or cap the jiiter as that can add an extra level of non determinism here. (i.e. we wait on the object to get reconciled or worse case the jiiter pushes out past the time window that we are waiting for).

justinsb: Ah OK, so I think I understand now.

Let's see what happens with the jitter.  I imagine we don't want to waiting around for a re-reconcile in general (it'll make our tests slow), so we will want to trigger reconciles.  We may want to make the re-reconcile interval _longer_, in that if our test happened to go slowly, we might see a "random" re-reconcile from another object when we aren't expecting it.  But I propose we tackle that when we start to see it.

So I think CHECK-LOG could be "WAIT-FOR-HTTP-REQUEST" or something along those lines.  And then I think ideally it should not have an object at all, we should just use the "APPLY" task.  Now I think the problem with that is that we need to find a change for each object that is reflected in metadata.generation, and I'm not sure if labels are reflected in metadata.generation (which is its own problem; we know we need to rethink labels).

So I'm OK to have this now that I understand what's going on :-), but let's work in this direction and see what's possible.

justinsb:
Thanks for explaining what's going on @acpana  ... I have some notes on proposed direction, but I don't think they need to hold up this PR - it's test-code only.

/approve
/lgtm

PR #1452, opened 2024-03-28
Add a step in dynamic test to verify status fields after creation

maqiuyujoyce:
### Change description

In dynamic test workflow, add a step to verify the `status` fields are populated correctly for newly supported CRDs. This is in preparation for #1353 (currently reverted).

### Tests you have done

`go test -timeout 1200s -v -tags=integration ./pkg/controller/dynamic/ -test.run TestCreateNoChangeUpdateDelete -run-tests pubsubtopic` passed

### DIFF
--- a/pkg/controller/dynamic/dynamic_controller_integration_test.go
+++ b/pkg/controller/dynamic/dynamic_controller_integration_test.go
@@ -266,6 +266,8 @@ func validateCreate(ctx context.Context, t *testing.T, testContext testrunner.Te
 
 	// Check observedGeneration matches with the pre-reconcile generation
 	testcontroller.AssertObservedGenerationEquals(t, reconciledUnstruct, preReconcileGeneration)
+
+	verifyStatusFields(t, reconciledUnstruct)
 }
 
 // testNoChange verifies that reconciling a resource which has not changed does not result in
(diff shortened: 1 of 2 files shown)


maqiuyujoyce:
@justinsb I took a bit of your refactoring work in #1438 here.

justinsb:
I think we more coverage from #1447 ?

pkg/controller/dynamic/dynamic_controller_integration_test.go
```diff
@@ -774,6 +776,46 @@ func containsResourceIDTestVar(t *testing.T, u *unstructured.Unstructured) bool
...
+		t.Errorf("reconciled status didn't have type map[string]interface{} but was actually %T", statusUnstruct)
+	}
+
+	requiredFieldsMap := map[string]bool{"observedGeneration": true, "conditions": true}
+	if len(statusMap) < 2 {
+		t.Errorf("reconciled resource didn't have enough subfields under 'status' field: it should at least have fields 'conditions' and 'observedGeneration'")
+	}
+
+	if !k8s.OutputOnlyFieldsAreUnderObservedState(reconciledUnstruct.GroupVersionKind()) {
+		return
```
gemmahou, line 795: Do we want to return an error here?

gemmahou:
> I think we get more coverage from #1447 ?

I feel it's fine to check status in dynamic test because we run dynamic test periodically and we currently do not have a job to run unified test(i'm trying to configure to run unified test using prow job so we can run it as a part of our ci workflow). 

We will still need a golden test to verify that object status contains the necessary fields (conditions and observedGeneration for all resources, and for new resources, it might have observedState). If I did not miss anything, this test does not verify the existence of observedState, because it is unpredictable?

maqiuyujoyce:
Status validation can be achieved as part of the mock test object check. Though it only checks the status for the resource after update.

We should consider how to check status after create once we identified about this missing check.

PR #1460, opened 2024-03-29
mockgcp: support options on httpmux

justinsb:
We want to send empty values on e.g. the cloudbilling API


acpana:
🥇 

/lgtm
/approve

PR #1461, opened 2024-03-29
Composition POC code dump.

barney-s:
- Uses inline expansion via PlanCR
- Uses k8s Job to expand jinja templates

### Change description

Fixes:  None

Experimental folder dump from a POC code base.


justinsb:
Let's get this merged in as-is and then we can start the unification process in a structured way.

/approve
/lgtm

Though it looks like you will have to add headers etc

PR #1464, opened 2024-03-29
chore: create a (placeholder) CI task

justinsb:
This script doesn't do anything yet, but we can start to add the
wiring for running it.

### DIFF
--- a/dev/ci/postsubmits/build-all
+++ b/dev/ci/postsubmits/build-all
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+# Copyright 2024 Google LLC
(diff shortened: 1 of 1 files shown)


justinsb:
cc @xiaoweim - this is the (placeholder) CI script that we can wire up

xiaoweim:
Hi @justinsb, I want to clarify that did you tag me for reference of the script that we might have, or is this task assigned to me for me to work on it? And It seems like that I cannot be assigned because of permission issue?

justinsb:
I think we have other CI jobs now

/close

PR #1465, opened 2024-03-29
Move output-only fields to status.observedState for new CRDs

maqiuyujoyce:
### Change description

Fixes b/327226936

Rebased on top of #1438 and redid #1353.
* Reused the allowlist for the `state-into-spec: merge` feature to identify whether computed fields should be under `status.observedState` in a resource. Added util functions.
* Updated the logic when fetching resource-specific fields under `status` in TF-based IAM client, and the reconciliation flow for TF-based resources.
* Updated the logic when populating resource-specific fields under `status` for TF-based resources.
* Updated the CRD generation for newly added TF-based resources to move output-only fields to `status.observedState`.
* Added unit tests to cover the changes.
  * Moved allowlist_test.go to the k8s_test package to avoid circular dependency.

### Tests you have done

`go test -timeout 1200s -v -tags=integration ./pkg/controller/dynamic/ -test.run TestCreateNoChangeUpdateDelete -run-tests pubsubtopic` passed


pkg/k8s/allowlist.go
```diff
@@ -16,231 +16,247 @@ package k8s
...
+	// Any newly supported v1beta1 kinds should NOT have computed fields
+	// directly under 'status' in the schema.
+	v1beta1KindsWithComputedFieldsUnderStatus = append(v1beta1KindsWithStateIntoSpecMergeSupport,
+		"ComputeNetworkFirewallPolicyAssociation")
+)
 
-func SupportsStateIntoSpecMergeInKind(kind string) bool {
-	return isValueInAllowlist(kind, v1beta1KindsWithStateIntoSpecMergeSupport)
+func SupportsStateIntoSpecMerge(gvk schema.GroupVersionKind) bool {
+	if gvk.Version != KCCAPIVersionV1Beta1 {
```
maqiuyujoyce, line 249: With this logic change (from #1438 ), we are going to break all the v1alpha1 resources: All the new v1alpha1 resource (KRM object) will only support "state-into-spec: absent".

Looks to be an okay change to me. @justinsb as you've showed concern about breaking v1alpha1 resources, I just want to confirm you are aware of this change.

pkg/k8s/allowlist_test.go
```diff
@@ -12,38 +12,159 @@
...
+				// false, the CRD is not supposed to have 'status.observedState'.
+				if !(shouldHaveObservedState || mayHaveObservedState) && hasObservedState {
+					t.Errorf("'status.observedState' exists, but it shouldn't")
+				}
+
+				// If mayHaveObservedState is true, it means 'status' will only
+				// have subfields 'conditions' and 'observedGeneration', and
+				// optionally subfield 'observedState'.
+				if mayHaveObservedState {
+					statusProp := findOpenAPIProperty(openAPISchema, "status")
```
maqiuyujoyce, line 128: This check is similar with the check in #1452 .

maqiuyujoyce:
Mock test TestAllInSeries/samples/project-level-policy#01 failed. Retriggered the test to verify if it is a flake.

pkg/k8s/allowlist_test.go
```diff
@@ -12,38 +12,159 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package k8s
+package k8s_test
```
justinsb, line 15: Why the change?

pkg/k8s/allowlist_test.go
```diff
@@ -12,38 +12,159 @@
...
+				//     have observedFields configured.
+				mayHaveObservedState := k8s.OutputOnlyFieldsAreUnderObservedState(gvk)
+				rcs, err := smLoader.GetResourceConfigs(gvk)
+				// Ignore not found error because there are handwritten and
+				// DCL-based resources.
+				if err != nil && !strings.Contains(err.Error(), "no mapping with name") {
+					t.Errorf("error getting resource config(s) for gvk %+v: %v", gvk, err)
+				}
+				shouldHaveObservedState := false
+				for _, rc := range rcs {
```
justinsb, line 106: Not introduced in this PR, but I'm confused by the idea that there might be more than one service mapping.

pkg/k8s/allowlist_test.go
```diff
@@ -12,38 +12,159 @@
...
+						if _, ok := requiredFieldsMap[k]; ok {
+							foundInMaps = true
+							delete(requiredFieldsMap, k)
+						}
+						if _, ok := optionalFieldsMap[k]; ok {
+							foundInMaps = true
+							delete(optionalFieldsMap, k)
+						}
+						if !foundInMaps {
+							t.Errorf("CRD has non-boilerplate field %v under 'status'", k)
```
justinsb, line 145: Nit: I think it's an "unexpected field" rather than "non-boilerplate field"

justinsb:
A few nits/questions, but nothing blocking this PR

/approve
/lgtm

pkg/k8s/allowlist_test.go
maqiuyujoyce: Test function `TestOutputOnlyFieldsAreUnderObservedState` needs to import the `servicemappingloader` package, which imports `k8s` package. Not updating the package name here will result in circular dependency issue.

We probably want to move allowlist related code outside of k8s.

pkg/k8s/allowlist_test.go
maqiuyujoyce: We have several instances of multiple ResourceConfigs mapping to the same Kind. This is mostly because we merged the locational resources, e.g. [rc1](https://cnrm.git.corp.google.com/cnrm/+/refs/heads/archived/master/config/servicemappings/compute.yaml#1368), [rc2](https://cnrm.git.corp.google.com/cnrm/+/refs/heads/archived/master/config/servicemappings/compute.yaml#93). Also we had to do some merges for edge case resources like ComputeInstance: [rc1](https://cnrm.git.corp.google.com/cnrm/+/refs/heads/archived/master/config/servicemappings/compute.yaml#816), [rc2](https://cnrm.git.corp.google.com/cnrm/+/refs/heads/archived/master/config/servicemappings/compute.yaml#923).

PR #1466, opened 2024-03-30
[WIP] Identify output-only fields under 'spec' for TF-based resources

maqiuyujoyce:
Currently, we are identifying output-only `spec` fields that meet the following criteria:
1. CRD contains a v1beta1 version.
    -> We want to focus on more important CRDs (i.e. v1beta1 is more important).
3. CRD is generated based on a TF resource.
    -> DCL-based resources don't have output-only spec fields unless they are subfields under a spec list.
5. Field is not sensitive.
    -> Further discussion will be needed to determine the strategy to handle sensitive, output-only field.
7. Field is not ignored.
    -> Ignored fields are not `spec` fields.
9. Field is not already observed.
    -> Observed field shouldn't be duplicated.
11. Field doesn't contain any list parent.
    -> Duplicating a subfield under a list may result in the size increase of the CR, which may eventually hit the etcd size limit.
13. Field itself is not a list.
    -> Duplicating a list may result in the size increase of the CR, which may eventually hit the etcd size limit.

This PR is a playground-ish PR, and will be closed after #1675 is merged. It was uploaded for posterity.

### DIFF
--- a/pkg/crd/crdgeneration/tf2crdgeneration.go
+++ b/pkg/crd/crdgeneration/tf2crdgeneration.go
@@ -16,6 +16,7 @@ package crdgeneration
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 
 	corekccv1alpha1 "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/apis/core/v1alpha1"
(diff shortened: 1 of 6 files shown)


yuwenma:
I was trying to ask how https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1675 was generated. I really like this full context PR description and the consideration for maintainers. Everything looks good except the confirmation about if we should include already observed fields. BTW, do you want to actually merge this PR ? This seems to be useful in the long run if we need to fork the TF repo.

maqiuyujoyce:
> Everything looks good except the confirmation about if we should include already observed fields.

The information generated here are additive instead of a replacement of the existing observedFields list, so it should be fine.

> BTW, do you want to actually merge this PR ?

I was messing around the existing tf2crdgeneration code, which will create a lot of noise if we merge it. We can turn it into a script eventually if it turns out to be useful by other contributors. Right now, I'd prefer closing it with out merge and using it for informational purposes only.

maqiuyujoyce:
We decided to only support spec fields as part of the observed state for TF-based resources after being explicitly requested by customers.

PR #1470, opened 2024-03-30
Support dynamic names for valueFrom resources in composition

barney-s:
Add valuesFrom.resourceRef.nameSuffix to support the use case where he dependent resource name is dynamically generated by a previous stage.

For now it supports the pattern: <inputapi.name><suffix>
This is limiting but provides a way to achieve it.

### DIFF
--- a/experiments/compositions/composition/api/v1/composition_types.go
+++ b/experiments/compositions/composition/api/v1/composition_types.go
@@ -23,8 +23,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
-// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+//TODO(barney-s) : Remove proto annotations
 
 type ResourceRef struct {
 	// OPTION 1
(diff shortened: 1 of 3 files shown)


experiments/compositions/composition/api/v1/composition_types.go
```diff
@@ -32,10 +32,11 @@ type ResourceRef struct {
 	// resource: ServiceIdentity.serviceusage.cnrm.cloud.google.com/v1beta1//sqladmin.googleapis.com
 
 	// OPTION 2
-	Group   string `json:"group" protobuf:"bytes,1,name=group"`
-	Version string `json:"version,omitempty" protobuf:"bytes,2,opt,name=version"`
-	Kind    string `json:"kind" protobuf:"bytes,3,name=kind"`
-	Name    string `json:"name" protobuf:"bytes,4,name=name"`
+	Group      string `json:"group" protobuf:"bytes,1,name=group"`
```
justinsb, line 35: Do we need the proto annotations?

experiments/compositions/composition/api/v1/composition_types.go
```diff
@@ -153,9 +154,17 @@ func (s *Composition) Validate() bool {
 	s.Status.ClearCondition(ValidationFailed)
 	// Validate Expanders
 	message := ""
-	for i, expander := range s.Spec.Expanders {
+	for expanderIndex, expander := range s.Spec.Expanders {
 		if expander.Name == "" {
-			message += fmt.Sprintf("Expander: %d missing Name", i)
+			message += fmt.Sprintf("Expander: %d missing Name", expanderIndex)
```
justinsb, line 159: Nit: something like `spec.expanders[%d] missing name` might be clearer.  We could even use the validation errors from apimachinery, but it'll end up with something like  `spec.expanders[%d]`  IIRC

justinsb:
Some comments, but nothing blocking.  I'll approve & lgtm & hold, feel free to clear hold once you've seen comments.

/approve
/lgtm
/hold

experiments/compositions/composition/api/v1/composition_types.go
barney-s: may be not. I started off with @cheftako's initial code and continued having it. 
We can remove if not needed for k8s usecase.

barney-s: Added a TODO in the file.

barney-s (commented):
resolved comments

zicongmei:
Adding `nameSuffix` is not going to solve the problem. There can be multiple resources with a same suffix.

PR #1472, opened 2024-03-31
chore: refactor job execution in compositions with a view to making it pluggable

justinsb:
Starting by moving the code, then we can add a second implementation
to understand the underlying contract.


barney-s:
@justinsb - Would you rebase to the latest and run `make && make e2e-test` ? 

Or i can clone this CL and work. Let me know.

PR #1473, opened 2024-03-31
chore: update to controller-tools 0.14.0

justinsb:
Older versions crash with newer version of go.

Also we switch the makefile to use `go run`, avoiding the complexity
of trying to install binaries.


yuwenma:
/lgtm
/approve

Thanks for keeping the controller-tools update to date and fixing the crash

justinsb:
We are now on a later version of controller-tools

/close

PR #1490, opened 2024-04-02
mockgcp: fix makefile to only generate some of aiplatform

justinsb:
Some of the data-plane protos are incompatible with grpc-gateway, so
only generate the control-plane protos we need.


anhdle-sso:
/remove-lgtm I encounter the same issue with the unit test

Issue #1201, opened 2024-02-06
Need a schema-validation test on our samples/fixtures


justinsb:
### Checklist

- [X] I did not find a related open enhancement request.
- [X] I understand that enhancement requests filed in the GitHub repository are by default low priority.
- [X] If this request is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Describe the feature or resource

Some of our samples/fixtures would not pass schema validation.  I believe this is causing the TestCRUD tests to fail.

We should do schema validation on our samples/fixtures as a test, this test would not even need a cluster I believe.

ldanielmadariaga:
This item has been open for >12 months. As part of scheduled repository maintenance, this Issue/PR will be closed. If this is still a critical ask, please re-open and work with your Google Cloud Account team to prioritize it with GCP KCC Product Management.

PR #1494, opened 2024-04-02
Fix sample yaml

gemmahou:
### Change description

Fixes b/322827315

### Tests you have done


maqiuyujoyce:
lgtm/
approve/

justinsb:
It would be good to check the schemas of all our tests ... https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/1201, I think this would have caught the logging_config error.

Now that we have pause, maybe an easy way would be to apply them all without resetting the cluster in between cc @acpana

maqiuyujoyce:
> It would be good to check the schemas of all our tests ... https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/1201, I think this would have caught the logging_config error.

Agreed we should have test coverage for schema either via the webhook or the unit test. (Using the webhook may be more aligned with the customers' view though.)

PR #1497, opened 2024-04-03
Make goimports cover the entire repository

maqiuyujoyce:
### Change description

Added tests/ and experiments/ folders to goimports command.
I think it is simpler and easier to maintain if we enforce goimports across the entire repository but it seems to modify generated files that don't have a good import grouping rule. So I manually picked two more folders to cover.

### Tests you have done

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ manifests: generate
 fmt:
 	mockgcp/dev/fix-gofmt
 	make -C operator fmt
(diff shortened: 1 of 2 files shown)


justinsb:
Thanks @maqiuyujoyce  - now I understand why we didn't have consistent formatting in the tests directory!

/approve
/lgtm

PR #1499, opened 2024-04-03
feat: add direct logMetric

acpana:
Initial implementation of `LoggingLogMetric` resource as a direct api controller. Patch also adds KRM golden object e2e tests for this resource.

### DIFF
--- a/pkg/controller/direct/logging/logmetric_controller.go
+++ b/pkg/controller/direct/logging/logmetric_controller.go
@@ -1,0 +1,253 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package logging
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"strings"
+
+	api "google.golang.org/api/logging/v2"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
+	krm "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/clients/generated/apis/logging/v1beta1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
+)
+
+const ctrlName = "logmetric-controller"
+
+// AddLogMetricController creates a new controller and adds it to the Manager.
+// The Manager will set fields on the Controller and start it when the Manager is started.
+func AddLogMetricController(mgr manager.Manager, config *controller.Config, opts directbase.Deps) error {
+	gvk := krm.LoggingLogMetricGVK
+
+	// todo(acpana): plumb context throughout direct
+	ctx := context.TODO()
+	gcpClient, err := newGCPClient(ctx, config)
+	if err != nil {
+		return err
+	}
+	m := &logMetricModel{gcpClient: gcpClient}
+	return directbase.Add(mgr, gvk, m, opts)
+}
+
+type logMetricModel struct {
+	*gcpClient
+}
+
+// model implements the Model interface.
+var _ directbase.Model = &logMetricModel{}
+
+type logMetricAdapter struct {
+	resourceID string
+	parentID   string
+
+	desired         *krm.LoggingLogMetric
+	actual          *api.LogMetric
+	logMetricClient *api.ProjectsMetricsService
+}
+
+var _ directbase.Adapter = &logMetricAdapter{}
+
+// AdapterForObject implements the Model interface.
+func (m *logMetricModel) AdapterForObject(ctx context.Context, u *unstructured.Unstructured) (directbase.Adapter, error) {
+	projectMetricsService, err := m.newProjectMetricsService(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	obj := &krm.LoggingLogMetric{}
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &obj); err != nil {
+		return nil, fmt.Errorf("error converting to %T: %w", obj, err)
+	}
+
+	if obj.Spec.ProjectRef.External == "" {
+		// todo acpana: resolve the external ref
+		return nil, fmt.Errorf("project external reference is not set")
+	}
+
+	// validate that the project ref follows the "projects/PROJECT_ID" format
+	if parts := strings.Split(obj.Spec.ProjectRef.External, "/"); len(parts) != 2 || parts[0] != "projects" {
+		return nil, fmt.Errorf("project external reference is not in the format projects/PROJECT_ID")
+	}
+
+	return &logMetricAdapter{
+		resourceID:      ValueOf(obj.Spec.ResourceID),
+		parentID:        obj.Spec.ProjectRef.External,
+		desired:         obj,
+		logMetricClient: projectMetricsService,
+	}, nil
+}
+
+func (a *logMetricAdapter) Find(ctx context.Context) (bool, error) {
+	if a.resourceID == "" {
+		return false, nil
+	}
+
+	logMetric, err := a.logMetricClient.Get(a.fullyQualifiedName()).Context(ctx).Do()
+	if err != nil {
+		if IsNotFound(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("getting logMetric %q: %w", a.fullyQualifiedName(), err)
+	}
+
+	a.actual = logMetric
+
+	return true, nil
+}
+
+// Delete implements the Adapter interface.
+func (a *logMetricAdapter) Delete(ctx context.Context) (bool, error) {
+	// Already deleted
+	if a.resourceID == "" {
+		return false, nil
+	}
+
+	_, err := a.logMetricClient.Delete(a.fullyQualifiedName()).Context(ctx).Do()
+	if err != nil {
+		if IsNotFound(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("deleting log metric %s: %w", a.fullyQualifiedName(), err)
+	}
+
+	return true, nil
+}
+
+func (a *logMetricAdapter) Create(ctx context.Context, u *unstructured.Unstructured) error {
+	log := klog.FromContext(ctx).WithName(ctrlName)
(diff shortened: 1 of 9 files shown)


justinsb:
Looks good.  Does it pass tests if you wire it in through KCC_USE_DIRECT_RECONCILERS ?  (We don't expect the golden output to be identical - at least the user agent will change, and in practice terraform is very chatty.  But we do expect the test itself to pass e.g. the golden output should be the same the second time you run it!)

pkg/controller/direct/logging/logmetric_controller.go
```diff
@@ -0,0 +1,380 @@
...
+	update := &api.LogMetric{}
+	update.Name = a.fullyQualifiedName()
+
+	// description is the only field that can be updated
+	if ValueOf(a.desired.Spec.Description) != a.actual.Description {
+		update.Description = ValueOf(a.desired.Spec.Description)
+	}
+	// todo: acpana do the rest of the field updates
+
+	// DANGER: this is an upsert; it will create the LogMetric if it doesn't exists.
```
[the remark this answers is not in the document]
acpana: Yes I think the behavior is the same here:

- gen api: [ref](https://github.com/googleapis/google-api-go-client/blob/main/logging/v2beta1/logging-gen.go#L3167-L3193)
- terraform in our repo: [ref](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/logging/resource_logging_metric.go#L578-L586)

pkg/controller/direct/logging/logmetric_controller.go
```diff
@@ -0,0 +1,380 @@
...
+	// a shared validator? validate exists/ check set? validate is well formed?
+	if project == "" {
+		return fmt.Errorf("project is empty")
+	}
+	name := a.desired.GetName()
+	if name == "" {
+		return fmt.Errorf("name is empty")
+	}
+	lm := &api.LogMetric{
+		Name: fmt.Sprintf("projects/%s/metrics/%s", project, name),
```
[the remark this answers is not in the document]
acpana: I went back on some `fqn` calls as it seems the `resourceID` gets set as `projects/project_id/metrics/metric_id`.

acpana:
I think this is getting a little closer! Also:

```bash
$ KCC_USE_DIRECT_RECONCILERS=LoggingLogMetric BILLING_ACCOUNT_ID="REDACTED" ORGANIZATION_ID="REDACTED" go test -v -tags=integration ./config/tests/samples/create -timeout 1600s -test.count=1 -test.run TestAll -run-tests logginglogmetric 2>&1

...
--- PASS: TestAll (1.08s)
    --- PASS: TestAll/int-log-metric (10.13s)
    --- PASS: TestAll/exponential-log-metric (10.14s)
    --- PASS: TestAll/explicit-log-metric (10.31s)
    --- PASS: TestAll/linear-log-metric (11.07s)
PASS
...
ok  	github.com/GoogleCloudPlatform/k8s-config-connector/config/tests/samples/create	81.190s
```

(Although I think that test run has these commits cherry-picked for niceness -- https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1424/commits . I should use the builtin go run flag instead 🙃 )

pkg/controller/direct/logging/utils.go
```diff
@@ -0,0 +1,142 @@
...
+
+	return reflect.DeepEqual(kccObj.LinearBuckets, kccApiReducedLinearBuckets)
+}
+
+func expandKCCtoAPIForBucketOptions(kccObj *krm.LogmetricBucketOptions) *api.BucketOptions {
+	if kccObj == nil {
+		return nil
+	}
+
+	expanded := &api.BucketOptions{}
```
justinsb, line 123: I like this, I'm starting to imagine the auto-generated version of this.

```
func mapKCCToAPI_BucketOptions(in *krm.LogmetricBucketOptions) *api.BucketOptions {
  if (in == nil) { return nil }
  out := &api.BucketOptions{}
  out.ExplicitBuckets = mapKCCToApi_Explicit(in.ExplicitBuckets)
  out.ExponentialBuckets = mapKCCToApi_Exponential(in.ExponentialBuckets)
  out.LinearBuckets = mapKCCToApi_Linear(in.LinearBuckets)
   return out  
}

func mapKCCToAPI_Explciit(in *krm.Explicit) *api.Explicit {
  if (in == nil) { return nil }
...
```

The trick I've seen done in kubernetes (for version conversion) is to have a `<foo>_generated.go` (or something) and then only generate the functions that aren't explicitly defined.  So if you wanted to override `mapKCCToAPI_BucketOptions`, you would just have a `func mapKCCToAPI_BucketOptions(` in your non-generated code.  I think the original kube generator code used static analysis to parse the go code, but that is actually a bit of a pain because it doesn't work mid-edit when the code doesn't compile etc, so it might be more user-friendly just to regex match on e.g. `^func mapKCCToAPI_BucketOptions(` (with some escaping) and  just require people not to do things like this:

```
/* EDGE CASE */ func mapKCCToAPI_BucketOptions(
```

(although I can also see us getting cute here and removing anything in between `/*` and `*/` before doing the regexp)

cc @anhdle-sso

pkg/controller/direct/logging/logmetric_controller.go
```diff
@@ -0,0 +1,380 @@
...
+	name := a.desired.GetName()
+	if name == "" {
+		return fmt.Errorf("name is empty")
+	}
+	filter := a.desired.Spec.Filter
+	if filter == "" {
+		return fmt.Errorf("filter is empty")
+	}
+
+	lm := &api.LogMetric{
```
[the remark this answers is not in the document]
acpana: yes I am partial to shortened one letter derived var names hehe but please call me out on it for egregious ones!

pkg/controller/direct/logging/logmetric_controller.go
```diff
@@ -0,0 +1,380 @@
...
+}
+
+// todo acpana: make sure that the format of the metric is not changing here.
+func (a *logMetricAdapter) Find(ctx context.Context) (bool, error) {
+	if a.resourceID == "" {
+		return false, nil
+	}
+
+	// it seems the resource ID is the fully qualified name
+	// looks like: projects/acpana-kcc/metrics/blllinyzbesrgmbrvvtwbibkvpfjunkhsrugzzynp
```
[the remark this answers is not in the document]
acpana: Yes I did! Thanks for your patience but I was wrong. The `resourceID` remains an ID for the log metric and does not include the `projects/PROJECT_ID`.

I got confused by the ProjectMetricsCreateCall as it's the online on that expects the resourceID by itself and not the  fully qualified name used in all other request structs:

https://github.com/googleapis/google-api-go-client/blob/f8960e3fd83691d395ac826fdaf8f8de590704fd/logging/v2/logging-gen.go#L39090-L39105

acpana:
fyi folks I've added the KRM export golden checks as a wf step.

sample run: https://github.com/GoogleCloudPlatform/k8s-config-connector/actions/runs/8791915791/job/24127093191?pr=1499

pkg/test/utils.go
```diff
@@ -165,7 +169,42 @@ func CompareGoldenObject(t *testing.T, path string, got []byte) {
...
-	if diff := cmp.Diff(wantMap, gotMap); diff != "" {
+	if diff := cmp.Diff(postWriteNormalization(wantMap), gotMap); diff != "" {
 		t.Errorf("unexpected diff in %s: %s", path, diff)
 	}
+	if err := os.WriteFile(path+"_2d.yaml", got, 0644); err != nil {
+		t.Errorf("failed to write golden output %s: %v", path, err)
+	}
+}
+
+func postWriteNormalization(wantMap map[string]interface{}) map[string]interface{} {
```
acpana, line 180: Not a fan of this but short of re-capturing to exclude the annotations (which I don't think we should do just yet) I thought this was the lesser evil 🤷🏼

pkg/controller/direct/logging/logmetric_controller.go
```diff
@@ -0,0 +1,253 @@
...
+
+	status := &krm.LoggingLogMetricStatus{}
+	if err := logMetricStatusToKRM(actualUpdate, status); err != nil {
+		return err
+	}
+
+	// actualUpdate may not contain the description for the metric descriptor.
+	if update.Description != "" {
+		status.MetricDescriptor.Description = &update.Description
+	}
```
acpana, line 239: Highlighting this bc I did think it was a bit odd that the actualUpdate lacked the `Description` field.

acpana (commented):
some notes

pkg/test/utils.go
```diff
@@ -102,6 +102,10 @@ func CompareGoldenFile(t *testing.T, p string, got string, normalizers ...func(s
 		if diff := cmp.Diff(want, got); diff != "" {
 			t.Errorf("unexpected diff in %s: %s", p, diff)
 		}
+
+		if err := os.WriteFile(p+"_2.log", []byte(got), 0644); err != nil {
```
justinsb, line 106: Did you intend for this to be here?  (I'm not following what it is for TBH)

pkg/test/utils.go
justinsb: I think either we accept that the golden outputs are specific to a particular scenario (which highlights the gap) or we have two files (i.e. add a "scenario" into the filename / path).  I think I'd start with accepting the delta, and then if we can't close the gap, we can add it into the filename / path.

justinsb:
LGTM, I think everything is a nit other than the golden test workaround.  Personally I would rather live with the tests failing, but I can see that blocks us from running the tests in github actions.  I see you have a merge conflict here anyway, so let's figure out what makes sense to unblock this in chat...

pkg/test/utils.go
acpana: Nope! Good catch

pkg/controller/direct/logging/utils.go
```diff
@@ -0,0 +1,340 @@
...
+
+// todo acpana: end common things
+
+// todo acpana: house these somewhere else
+
+func compareMetricDescriptors(kccObj *krm.LogmetricMetricDescriptor, apiObj *api.MetricDescriptor) bool {
+	return reflect.DeepEqual(kccObj, reduceAPItoKRM_MetricDescriptor(apiObj))
+}
+
+func reduceAPItoKRM_MetricDescriptorStatus(apiObj *api.MetricDescriptor) *krm.LogmetricMetricDescriptorStatus {
```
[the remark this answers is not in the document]
acpana: No I went initially w reduce/ expand bc the API has some fields that are missing from the KRM representation but as we close the gap these words probably lose meaning so I'll rephrase.

pkg/controller/direct/logging/logmetric_controller.go
```diff
@@ -0,0 +1,253 @@
...
+	if project == "" {
+		return fmt.Errorf("project is empty")
+	}
+	name := a.desired.GetName()
+	if name == "" {
+		return fmt.Errorf("name is empty")
+	}
+	filter := a.desired.Spec.Filter
+	if filter == "" {
+		return fmt.Errorf("filter is empty")
```
justinsb, line 151: No objection to doing this here, though I suspect over time we want to refactor validation into its own method.  Letting the GCP API return an error is always an option also (although the error messages tend not to be as good as if we hand-crafted them, as long as we keep the field names reasonably close the error messages tend not to be _bad_)

justinsb:
Feature-flagged, so let's merge & iterate!

/approve
/lgtm

PR #1501, opened 2024-04-03
Changes to make compositions deployable to a kind cluster

barney-s:
### Change description
Compositions runnable in a kind cluster.

- Validate the composition manifests are working in a kind cluster
- Changes required to grant additional rbac permissions to the composition controller
- Changes to create jobs in the namespace of the composition object instead of default namespace
- Changes to grant appropriate permissions using Roles and bindings instead of ClusterRoles and bindings
- make target to generate manifests deployable to a kind cluster
- make target to build images, create a kind cluster, add images to the kind cluster and deploy the controller there. Allows for e2e tests to be run in kind clusters.

### DIFF
--- a/experiments/compositions/composition/Makefile
+++ b/experiments/compositions/composition/Makefile
@@ -1,10 +1,15 @@
 # GCP project to use for pushing manifests
 GCP_PROJECT_ID ?= $(shell gcloud config get-value project)
 # Image URL to use all building/pushing image targets
-IMG ?= gcr.io/$(GCP_PROJECT_ID)/composition:latest
-INLINE_IMG ?= gcr.io/$(GCP_PROJECT_ID)/manifests-inline:latest
+GIT_IMG_VERSION ?= $(shell git rev-parse --short HEAD)
+IMG_VERSION ?= v0.0.1.alpha
+IMG ?= gcr.io/$(GCP_PROJECT_ID)/composition:$(IMG_VERSION)
+INLINE_IMG ?= gcr.io/$(GCP_PROJECT_ID)/manifests-inline:$(IMG_VERSION)
+JINJA_IMG ?= gcr.io/$(GCP_PROJECT_ID)/expander-jinja2:$(IMG_VERSION)
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.28.0
+KIND_CLUSTER ?= kind
+
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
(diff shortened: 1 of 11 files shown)


experiments/compositions/composition/internal/controller/composition_controller.go
```diff
@@ -47,9 +47,15 @@ type CompositionReconciler struct {
 //+kubebuilder:rbac:groups=composition.google.com,resources=compositions,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=composition.google.com,resources=compositions/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=composition.google.com,resources=compositions/finalizers,verbs=update
-//+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list
+//+kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list;watch
 //+kubebuilder:rbac:groups="",resources=events,verbs=create;patch
+//+kubebuilder:rbac:groups=alice.alice,resources=*,verbs=get;list;watch;update;patch
```
justinsb, line 52: Not a blocker, but we probably want to include the alice permissions separately (as part of CRD creation?)

experiments/compositions/composition/Makefile
```diff
@@ -152,6 +157,30 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
 	- $(CONTAINER_TOOL) buildx rm project-v3-builder
 	rm Dockerfile.cross
 
+##@ Testing
+
+.PHONY: release-kind-manifests
+release-kind-manifests: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
```
justinsb, line 163: Idea for the future: can we have one manifest?  Then when you test on kind, you build confidence on not-kind

experiments/compositions/composition/internal/controller/composition_controller.go
barney-s: Agreed. Trying to figure this part out. Should we have a separate CRD that generates Roles or ask the user to edit a role.

barney-s: Ah i see.. part of alice crd onboarding is appropriate place.

experiments/compositions/composition/Makefile
barney-s: This is more of dev-testing since we auto fill the current project for the gcr images. 
Need a proper release target that generates manifests for images from release gcr + version

Misnamed. Release-devtest-manifests /?

experiments/compositions/composition/Makefile
```diff
@@ -1,10 +1,15 @@
 # GCP project to use for pushing manifests
 GCP_PROJECT_ID ?= $(shell gcloud config get-value project)
 # Image URL to use all building/pushing image targets
-IMG ?= gcr.io/$(GCP_PROJECT_ID)/composition:latest
-INLINE_IMG ?= gcr.io/$(GCP_PROJECT_ID)/manifests-inline:latest
+GIT_IMG_VERSION ?= $(shell git rev-parse --short HEAD)
+IMG_VERSION ?= v0.0.1.alpha
+IMG ?= gcr.io/$(GCP_PROJECT_ID)/composition:$(IMG_VERSION)
+INLINE_IMG ?= gcr.io/$(GCP_PROJECT_ID)/manifests-inline:$(IMG_VERSION)
```
cheftako, line 7: Do we want a common prefix for the images?

experiments/compositions/composition/Makefile
```diff
@@ -152,6 +157,30 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
 	- $(CONTAINER_TOOL) buildx rm project-v3-builder
 	rm Dockerfile.cross
 
+##@ Testing
+
+.PHONY: release-kind-manifests
+release-kind-manifests: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
+	cd ../alice && $(KUSTOMIZE) build config/crd -o ../composition/release/kind/alice_crds.yaml
```
cheftako, line 164: Chatting with Justin and was wondering if we wanted to change alice to facade.

experiments/compositions/composition/config/manager/kustomization.yaml
```diff
@@ -18,5 +18,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: gcr.io/wfender-gke-dev/composition
-  newTag: 0.0.18
+  newName: gcr.io/allotrope-barni/composition
```
cheftako, line 21: We need to get rid of both 'wfender-gke-dev' and 'barni'. Probably need to run a templating engine or sed script to set the appropriate project here.

experiments/compositions/composition/Makefile
barney-s: added.

experiments/compositions/composition/config/manager/kustomization.yaml
barney-s: I have a way of doing it temporarily in makefile.

barney-s: done

experiments/compositions/composition/Makefile
barney-s: can do in a subsequest change. It would be a set of small changes across a lot of files. Would prefer to do it after the e2e tests have been merged.

barney-s:
Fixes #1509 - Expander pods created in composition namespace.

PR #1506, opened 2024-04-04
compute_network.patch

katrielt:
### Change description

Change ForceNew to false for the field enable_ula_internal_ipv6. The Compute API allows this field to be enabled/disabled on-the-fly.

Fixes b/332834635

### Tests you have done

### DIFF
--- a/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_computenetworks.compute.cnrm.cloud.google.com.yaml
+++ b/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_computenetworks.compute.cnrm.cloud.google.com.yaml
@@ -79,7 +79,7 @@ spec:
                 type: string
               enableUlaInternalIpv6:
                 description: |-
-                  Immutable. Enable ULA internal ipv6 on this network. Enabling this feature will assign
+                  Enable ULA internal ipv6 on this network. Enabling this feature will assign
                   a /48 from google defined ULA prefix fd20::/20.
                 type: boolean
               internalIpv6Range:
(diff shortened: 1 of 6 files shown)


justinsb:
Looks like our presubmit validation wants you to squash the commits - sorry!  (I _think_ it's right though, we really should have our "terraform LTS provider" in a separate repo, and from that perspective separate commits makes sense)

It would be great if you could provide a testfixture that updates enable_ula_internal_ipv6, and then we can verify it and update the mocks etc.  Let me know if you need any pointers for that, it should "just" be a create.yaml with it set one way and then an update.yaml which flips the field!

katrielt:
> Looks like our presubmit validation wants you to squash the commits - sorry! (I _think_ it's right though, we really should have our "terraform LTS provider" in a separate repo, and from that perspective separate commits makes sense)
> 
> It would be great if you could provide a testfixture that updates enable_ula_internal_ipv6, and then we can verify it and update the mocks etc. Let me know if you need any pointers for that, it should "just" be a create.yaml with it set one way and then an update.yaml which flips the field!

Done. Squashed and pushed.

I'll add a to the test and update the docs after this PR is submitted?

justinsb:
Sorry @katrielt - I accidentally merged a PR before it was passing tests, because (apparently) sometimes our merge robot will merge a PR that is failing tests, as long as it has approve and lgtm.  I'm going to dig into what actually happened there, but the TLDR is that if you rebase HEAD should now be green.

(We're also thinking we need to implement better testing, like github's submit queue, but ... that's another story)

justinsb:
Thanks @katrielt 

/approve
/lgtm

PR #1512, opened 2024-04-05
vcr: Update vcr hook and matcher to avoid potential failures

gemmahou:
### Change description
1. Update vcr hook to record failed requests (404, 400, 403) but replace the internal error message. My initial idea was that removing them would speed up vcr replay, but in some cases, missing those requests would cause an "interaction not found" issue in replay mode.
2. Update vcr matcher to sort request json body before comparing. json can be same but reordered
3. Update to use example-project in replay mode

### Tests you have done

### DIFF
--- a/tests/e2e/unified_test.go
+++ b/tests/e2e/unified_test.go
@@ -15,14 +15,12 @@
 package e2e
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
 	testvariable "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test/resourcefixture/variable"
 	"hash/fnv"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


tests/e2e/unified_test.go
```diff
@@ -195,11 +193,14 @@ func testFixturesInSeries(ctx context.Context, t *testing.T, testPause bool, can
 					unique := make(map[string]bool)
 
 					hook := func(i *cassette.Interaction) error {
-						// Discard failed interactions
+						// Remove internal error message from failed interactions
 						resCode := i.Response.Code
 						if resCode == 404 || resCode == 400 || resCode == 403 {
-							i.DiscardOnSave = true
+							i.Response.Body = "fake error message"
```
justinsb, line 199: This may cause some problems, e.g. terraform does some string matching for project-not-found errors IIRC, but ... we can see.  The alternative is to parse it as json, strip the fields that we don't want, then serialize back to json.... that's what we do in the golden tests; a bit of a pain but not terrible.  We could consider extracting the response-body normalization out into its own file, maybe, and share it.

justinsb:
Thanks @gemmahou  ... one thing that set off alarm bells in my head, but if the tests are passing, it works :-)

/approve
/lgtm

tests/e2e/unified_test.go
gemmahou: > This may cause some problems, e.g. terraform does some string matching for project-not-found errors IIRC

Ah, that might explain why those 404/400/403 interactions sometimes cause issues (not always).

I haven't tested all resources, but for computenodegroup/template and alloydb, removing those requests didn't result in any issues. However, when I tested on pubsubtopic/schema, the replay mode failed, and I noticed that I had to keep those error requests.

I will keep an eye on this if I find similar issues with other resources, which might be because of this. We can then try to find a way to hide only the internal message in the response body.

PR #1513, opened 2024-04-05
chore: update to go 1.22.2

justinsb:
* Update our go.mod to use toolchain
* Make sure dockerfiles are using pinned versions
* Set the version to 1.22.2


justinsb:
I think this needs the bump in #1700  ... I'll merge that one in here to be sure.

justinsb:
The bump in #1700 means this one does now pass tests, so I think we need #1700 and then I can rebase this one (or we can just merge this one)

yuwenma:
/lgtm

Personally I'd prefer merging #1700 first so that people can tell how to bump a `controller-gen` and which files are affected in the future when we need to. Defer to you.

PR #1514, opened 2024-04-05
chore: update github actions to use go version in go.mod

justinsb:
This avoids having to repeat the go version, and in general we want to
keep them in sync.


acpana:
good call! thanks for making this change 

that's a handy new feature from GH actions! 💯 

/lgtm
/approve

PR #1515, opened 2024-04-05
Inject Plan as OwnerRef to all expanded resources

barney-s:
### Change description

Inject Plan as OwnerRef to all expanded resources

- Add a deisgn note on ownerref and implentation options
- Inject Plan as OwnerRef for all expanded resources
- Reread plan after create to get the uuid

### DIFF
--- a/experiments/compositions/composition/internal/controller/expander_reconciler.go
+++ b/experiments/compositions/composition/internal/controller/expander_reconciler.go
@@ -107,6 +107,12 @@ func (r *ExpanderReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 			logger.Error(err, "Unable to create Plan Object")
 			return ctrl.Result{}, err
 		}
+
+		// Read after create to get the UID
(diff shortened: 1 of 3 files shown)


experiments/compositions/composition/internal/controller/expander_reconciler.go
```diff
@@ -107,6 +107,12 @@ func (r *ExpanderReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 			logger.Error(err, "Unable to create Plan Object")
 			return ctrl.Result{}, err
 		}
+
+		// Read after create to get the UID
+		if err := r.Client.Get(ctx, planNN, &plancr); err != nil {
```
justinsb, line 112: Do you know for a fact this isn't set?  I ask because I thought client.Client actually populated back some values, like resourceVersion in particular.

barney-s: I dont know for a fact. I assumed it did not. Will check and remove this if that is indeed the case.

PR #1517, opened 2024-04-05
tests: skip secondary-instance sample test for now

justinsb:
Until we can enforce that secondary instance should only be created
once primary is created.


acpana:
thanks for following up on the test issue 💯 

/lgtm
/approve

do we know if the test ever passed reliably?

PR #1527, opened 2024-04-09
Soft delete feature implementation for Google Cloud Storage

cheftako:
### Change description

Backport https://github.com/hashicorp/terraform-provider-google-beta/pull/7119

Adds support for Soft Delete feature, which allows setting soft delete policy on 'google_storage_bucket' resource.

This fixes hashicorp/terraform-provider-google#17127

storage: added 'soft_delete_policy' to 'google_storage_bucket' resource Derived from GoogleCloudPlatform/magic-modules#10171

Fixes #

### Tests you have done

None


config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_storagebuckets.storage.cnrm.cloud.google.com.yaml
```diff
@@ -287,6 +287,23 @@ spec:
                 required:
                 - retentionPeriod
                 type: object
+              softDeletePolicy:
+                description: The bucket's soft delete policy, which defines the period
+                  of time that soft-deleted objects will be retained, and cannot be
+                  permanently deleted. If it is not provided, by default Google Cloud
+                  Storage sets this to default soft delete policy.
+                properties:
+                  effectiveTime:
```
justinsb, line 296: This really should be status (though it's not because it's nested)

Also, I think we want retentionDurationSeconds to be in both spec and status.observedState

And I think we want to be sure that we only write back to status.observedState (which I think is fair because it's a new field, nobody's going to be looking in the "old place").

justinsb:
Looks good from a backport perspective.  I think we have to split the terraform change from the non-terraform change until we can move the terraform provider v4 to its own repo.  But I'm glad you did, because it does suggest the things we wanted from a declarative interface are not what we get by default from terraform.

Assuming we won't be able to fix this in the 1.116 timeframe, we should get the features in for 1.116 (DNSRecordSet), cut the branch, then merge this and make whatever improvements we can for 1.117.  As we improve the build process we'll be less constrained here (because we'll be more able to cherry pick), but we're not there yet.

Do you agree?

jdub55:
Can this change be backported to version 4.x of the Terraform provider? IHAC that is using Terraform v4 provider in production and it is not a trivial task to upgrade to the latest version of the provider for a number of different reasons.

600lyy:
Just created an issue for the tracking purpose:
https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/2046

@justinsb If you want, I can continue to work on this and open a new PR if needed

justinsb:
We merged @600lyy 's version of this; ok to close @cheftako ?"

PR #1529, opened 2024-04-09
chore: Add CRD category to generated types

justinsb:
We want to be able to check that the generated-types round-trip.

### DIFF
--- a/scripts/generate-go-crd-clients/general_types.go
+++ b/scripts/generate-go-crd-clients/general_types.go
@@ -102,6 +102,16 @@ func (g *GeneralTypes) Generate() {
 
 	g.Print("// +genclient")
 	g.Print("// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object")
+	g.Print("// +kubebuilder:resource:categories=%s,shortName=%s",
+		strings.Join(g.CRD.Spec.Names.Categories, ";"),
+		strings.Join(g.CRD.Spec.Names.ShortNames, ";"))
+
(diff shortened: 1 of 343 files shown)


justinsb:
WIP because I would like to see our tests fail because I have not (yet) regenerated the clients.

justinsb:
No longer WIP, this does produce a CRD that is not perfect, but is much closer.

(e.g. I've been doing `go run sigs.k8s.io/controller-tools/cmd/controller-gen@v0.14.0 +crd paths=./pkg/clients/generated/apis/storage/v1beta1/`)

scripts/generate-go-crd-clients/general_types.go
acpana, line 1: Would it be worth having a smoke test for the client generation? I see some `client_test` cases but if I were to guess those were added to ensure regressions don't happen for a given resource. Not blocking of course.

acpana (commented):
I spot checked some generated clients too, including `logginglogmetric`. I mostly wanted to make sure that the tags get applied to both v1alpha1 and v1beta1

acpana:
/lgtm
/approve

thanks for adding this 🥇

acpana:
> a CRD that is not perfect, but is much closer.

@justinsb I forgot to mention but if you have a mental list of all the CRD gen improvements that we can make, let's throw them in an issue so we can work on them together!

scripts/generate-go-crd-clients/general_types.go
justinsb: We do check in the generated code, and I believe I verified here that if we don't check in the correct generated code our CI catches that.  So we at least can see any regressions to this visually in a PR.

PR #1533, opened 2024-04-09
replace instances of charlie with Platform Admin and variants

barney-s:
### Change description

Cleanup. Replace instances of charlie with Platform Admin and variants


justinsb:
Thanks for rebasing

/approve
/lgtm

PR #1534, opened 2024-04-09
ci:refactor: reorder steps

acpana:
We first must checkout the repo before we can use its `go.mod` file in the setup go step.


barney-s (approved):

justinsb:
Great point :-)  Thanks for fixing! I guess there's a fallback for when go.mod isn't there, which is why this wasn't obvious...

justinsb:
Hmm.. well it's definitely failing now on other PRs - thanks for fixing!

PR #1536, opened 2024-04-10
tests: Add custom command line flags in unified test

gemmahou:
### Change description

Similar to samples and dynamic test, add `-run-tests` flag so we can run specific test cases in unified test by providing the test name. If it is not provided, all tests will be run.

Before: ` go test -test.count=1 -timeout 3600s -v ./tests/e2e -test.run TestAllInSeries/fixtures/computenodegroup`
After: ` go test -test.count=1 -timeout 3600s -v ./tests/e2e -test.run TestAllInSeries/fixtures -run-tests computenodegroup`

This is helpful when we want to run a subset of all test cases during presubmit, and also allow us to reuse the presubmit constants([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/test/constants/presubmitconstants.go#L31))

### Tests you have done

### DIFF
--- a/tests/e2e/unified_test.go
+++ b/tests/e2e/unified_test.go
@@ -15,16 +15,19 @@
 package e2e
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
+	"flag"
 	"fmt"
 	"hash/fnv"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"os"
 	"path/filepath"
-
+	"regexp"
 	"strconv"
 	"strings"
 	"testing"
@@ -48,6 +51,14 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+func init() {
+	// run-tests allows you to limit the tests that are run by specifying
+	// regexes to be used to match test names.
+	flag.StringVar(&runTestsRegex, "run-tests", "", "run only the tests whose names match the given regex")
+}
+
+var runTestsRegex string
+
 func TestAllInSeries(t *testing.T) {
 	if os.Getenv("RUN_E2E") == "" {
 		t.Skip("RUN_E2E not set; skipping")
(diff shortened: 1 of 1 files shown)


tests/e2e/unified_test.go
```diff
@@ -276,6 +296,8 @@ func testFixturesInSeries(ctx context.Context, t *testing.T, testPause bool, can
 							if err != nil {
 								t.Fatal("[VCR] Failed to read request body")
 							}
+							r.Body.Close()
+							r.Body = ioutil.NopCloser(bytes.NewBuffer(reqBody))
```
gemmahou, line 300: I noticed that I removed it by accident in another pr.

justinsb:
I'm guessing the primary motivation here is to reuse the existing skips?

Directionally though, I would like to be more "go".  I would like for us to be able to run `go test ./...` and it runs the tests with mock-kubeapiserver, mockkgcp or vcr etc.  Those tests should be reasonably fast, side-effect free, and not require configuration.  (And then if we want to run against real GCP or a real kube-apiserver we either pass a test or an environment variable).

In particular though, `go test -run` recognizes regexes, so I don't want to add this here, it's moving away from "normal go"

One flag I did notice that might be helpful; go test supports a `-skip` flag.  So one thing we could do is have a file with each skipped test listed explicitly, and then build the skip expression from that; technically it's regex but I think we would get away with just joining them all with `|` (though my regexp is not strong!)

gemmahou:
> I'm guessing the primary motivation here is to reuse the existing skips?
> 
> Directionally though, I would like to be more "go". I would like for us to be able to run `go test ./...` and it runs the tests with mock-kubeapiserver, mockkgcp or vcr etc. Those tests should be reasonably fast, side-effect free, and not require configuration. (And then if we want to run against real GCP or a real kube-apiserver we either pass a test or an environment variable).
> 
> In particular though, `go test -run` recognizes regexes, so I don't want to add this here, it's moving away from "normal go"
> 
> One flag I did notice that might be helpful; go test supports a `-skip` flag. So one thing we could do is have a file with each skipped test listed explicitly, and then build the skip expression from that; technically it's regex but I think we would get away with just joining them all with `|` (though my regexp is not strong!)

We may want to specify certain test to run it in presubmit. Even they are faster than realGCP tests, each of them still needs about 1min to finish and we have hundreds of test cases. I'm thinking we just run a subset of all tests in presubmit and all of them in another job(prow)? `-run` accept regex but the name needs to be a full path i.e. TestAllInSeries/fixtures/computenodegroup 

I agree we should also have `-skip`, I can add the flag in this PR.

gemmahou:
Obsolate PR.

PR #1539, opened 2024-04-10
Don't upload artifacts for pause test

gemmahou:
### Change description

There's no artifacts file to upload for pause test: https://screenshot.googleplex.com/BRn7vRxPpCVQUpo.png

Also I noticed that the GH presubmit/pause-test failed in some PRs([job](https://github.com/GoogleCloudPlatform/k8s-config-connector/actions/runs/8635527978/job/23673443666?pr=1498)). Unsure about the root cause but seems the error comes from [http_recorder.go](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/204782804df09f0ec10187bcdc7b13891b0203b7/pkg/test/http_recorder.go#L126). Removing `ARTIFACTS` env variable would be a quick fix.

### Tests you have done

### DIFF
--- a/.github/workflows/presubmit.yaml
+++ b/.github/workflows/presubmit.yaml
@@ -86,18 +86,11 @@ jobs:
       - uses: actions/setup-go@v4
         with:
           go-version-file: 'go.mod'
-      - name: "Run mock tests"
+      - name: "Run pause tests"
         run: |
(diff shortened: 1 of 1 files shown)


justinsb:
(I think the actual problem was that we were shutting down the controllers, but I realized we should log the error anyway: #1541 )

PR #1541, opened 2024-04-10
tests: log error if we fail to read response

justinsb:
Should help us understand what's going on here.


gemmahou:
Thanks!
/lgtm
/approve

PR #1544, opened 2024-04-11
chore: run tests in ./tests as part of github actions

justinsb:
The linter tests are inexpensive, we should run them as unit tests.

We probably want to run the mockgcp tests separately by default,
as they are not speedy enough to be considered unit tests.  Today,
we skip them anyway because we don't set RUN_E2E; if we make them
easy to run by default we likely want to skip them here.


anhdle-sso:
@justinsb Do you think we need to set RUN_E2E to run more test?

ldanielmadariaga:

PR #1546, opened 2024-04-11
mockgcp: support for pubsub

justinsb:
- mockgcp: add generated pubsub grpc code
- tests: capture some golden output for pubsub
- mockgcp: Implement pubsub
- mockgcp: Allow services to rewrite errors for accurate golden testing


pkg/test/resourcefixture/testdata/basic/pubsub/v1beta1/pubsubschema/_http.log
gemmahou, line 1: I'm wondering if it's possible to reuse the vcr log for golden testing? Perhaps we need to make sure the log's format matches.

justinsb: No objection if you want to try!  I think you would parse _http.log and use that as the cassette.  We might need to tweak _http.log to be a little more parseable (e.g. include an explicit separator because we removed Content-Length), but it shouldn't be _too_ bad.

It does mean basically rewriting the go-vcr library, but it seemed pretty thing anyway...

justinsb:
/assign @gemmahou 

This should let us get closer with mockgcp coverage for the samples (I think we also need KMS for the malformed sample  you hit)

PR #1549, opened 2024-04-11
ci: exclude **.md from presubmits

acpana:
Per a suggestion from @gemmahou , we can skip presubmits for PRs that are mostly doc changes (like README updates, docs, etc.)

Note per https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths if the diff contains additional files the workflows trigger.

### DIFF
--- a/.github/workflows/presubmit.yaml
+++ b/.github/workflows/presubmit.yaml
@@ -17,8 +17,12 @@ name: Presubmit
 on:
   pull_request:
     types: [opened, synchronize, reopened]
+    paths-ignore:
+      - "**.md"
   push:
     branches: ["master"]
+    paths-ignore:
+      - "**.md"
 
 jobs:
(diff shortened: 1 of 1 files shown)


gemmahou:
Thanks Alex for making this change! I'd like to hear others' opinions on this to ensure that we don't unintentionally cause any issues or raise copyright concerns.

One of our presubmit job checks Google License in every newly added markdown files, even for auto-generated ones(like logs). Once I re-generate these files, the License is overwritten and I have to manually add it back. I find this process annoying and License might be unnecessary for generated files.

cc: @justinsb, @maqiuyujoyce

gemmahou:
Oh I noticed that we just ignore .md files, my concern in another comment was about the generated ymal files.

/lgtm
/approve

PR #1551, opened 2024-04-12
Correct the path for storage proto

ziyue-101:
### Change description

Fix a path to store the generated storage proto from openapi

### Tests you have done 

### DIFF
--- a/mockgcp/Makefile
+++ b/mockgcp/Makefile
@@ -56,5 +56,5 @@ generate-protos-from-openapi:
 	cd tools/gapic; go run . ../../servicenetworking-api.json > ../../apis/mockgcp/cloud/servicenetworking/v1/servicenetworking.proto
 
 	wget -O storage-v1.json https://raw.githubusercontent.com/googleapis/google-api-go-client/main/storage/v1/storage-api.json
-	mkdir -p apis/mockgcp/cloud/storage/v1/
+	mkdir -p apis/mockgcp/storage/v1/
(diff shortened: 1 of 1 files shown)


mockgcp/Makefile
```diff
@@ -57,4 +57,4 @@ generate-protos-from-openapi:
 
 	wget -O storage-v1.json https://raw.githubusercontent.com/googleapis/google-api-go-client/main/storage/v1/storage-api.json
 	mkdir -p apis/mockgcp/cloud/storage/v1/
-	cd tools/gapic; go run . --proto-version=2 ../../storage-v1.json > ../../apis/mockgcp/storage/v1/service.proto
+	cd tools/gapic; go run . --proto-version=2 ../../storage-v1.json > ../../apis/mockgcp/cloud/storage/v1/service.proto
```
justinsb, line 60: So because we don't have a proto, it's not obvious whehter it should be google.cloud.storage or google.storage

However, we do have v2 of the API, and it is under google.storage: https://github.com/googleapis/googleapis/blob/36a3012c2c19e76b94972b4a60b01a7592824ea2/google/storage/v2/storage.proto

As such, I think this line is right (and the protoc command is also right), and the mkdir (line 59) is wrong

ziyue-101:
That makes sense. Updated the mkdir command

justinsb:
Thanks @ziyue-101 

/approve
/lgtm

PR #1552, opened 2024-04-12
vcr: generate cassette for presubmit

gemmahou:
### Change description

Generate cassette for presubmit resources. I'm using presubmit constants([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/test/constants/presubmitconstants.go#L31)) to determine presubmit resources. However, not all the test cases passed the record or replay mode(investigating...). All the tests in this PR I have verified locally, and they passed both modes, I didn't include failed tests in this PR.

How to review this PR:
Most of the files in this PR are generated. We need to double-check no internal data in the files that we do not want to expose to public.
lmk if we should split huge PRs like this into smaller ones.

### Tests you have done


justinsb:
> lmk if we should split huge PRs like this into smaller ones.

I think it's fine, for now I think the changes to the http logs are interesting, so I'm willing to just scan them when it comes to introducing golden output.  If/when it becomes annoying, we can add it to .gitattributes so it is automatically collapsed-by-default in github PR reviews.

pkg/test/resourcefixture/testdata/basic/accesscontextmanager/v1beta1/accesscontextmanagerserviceperimeter/dependencies.yaml
```diff
@@ -15,16 +15,11 @@
 apiVersion: resourcemanager.cnrm.cloud.google.com/v1beta1
 kind: Project
 metadata:
-  annotations:
-    cnrm.cloud.google.com/deletion-policy: "abandon"
```
justinsb, line 19: Can we (easily?) split out the changes to the tests?  The issue is that this PR then comingles two different ideas: (1) changing/fixing the tests and (2) generating the test golden output / casettes.

The second one is easy to merge, the first one needs a lot more consideration.  So now your "fast merge" PR becomes a "slow merge" PR

tests/e2e/testdata/vcr-cassette/TestAllInSeries_fixtures_androidrecaptchaenterprisekey-oauth.yaml
```diff
@@ -0,0 +1,17 @@
+# Copyright 2024 Google LLC
+#
```
justinsb, line 2: I thought we were going to put each casette under a vcr directory alongside the test?  So in this case this would be `pkg/test/resourcefixture/testdata/basic/recaptchaenterprise/v1beta1/recaptchaenterprisekey/androidrecaptchaenterprisekey/_vcr/oauth.yaml`  I am probably confused / misremembering though!

pkg/test/resourcefixture/testdata/basic/accesscontextmanager/v1beta1/accesscontextmanagerserviceperimeter/dependencies.yaml
justinsb: In this case, as there's only two, I think it's less work to just do it in one PR.  Can you (briefly) explain the reason for this change (and the other accesscontextmanager one), just so we have a record?

tests/e2e/testdata/vcr-cassette/TestAllInSeries_fixtures_androidrecaptchaenterprisekey-oauth.yaml
gemmahou: Ah right we agreed to keep all test data(human-crafted & generated) for same resource at one place. 

However this PR is already huge, I addressed this in: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1564.

justinsb:
OK, moving in another PR sgtm!  And thanks for splitting out #1564

/approve
/lgtm

PR #1557, opened 2024-04-15
tests: add a project-id annotation defaulter to mock-kubeapiserver

justinsb:
Because our mock-kubeapiserver doesn't support webhooks, we need to
find another way.  Defaulters are a good mechanism, and generally we
can now start to think about reducing our reliance on webhooks.


yuwenma (commented):
/lgtm
/approve
/hold 

Code looks good and I like the goal of this PR. The only thing I'd like to double check is that if the changes to state-into-spec need to be updated. 
Also, I think this change shall be reflected in the golden object, like some resources have the project-id annotation using the default value.

ldanielmadariaga:

PR #1560, opened 2024-04-15
chore: generate logginglogmetric krm exports

acpana:
Sample commands used for generation:

```bash
$ WRITE_GOLDEN_OUTPUT=1 GOLDEN_OBJECT_CHECKS=1 BILLING_ACCOUNT_ID="REDACTED" ORGANIZATION_ID="REDACTED" E2E_KUBE_TARGET=envtest RUN_E2E=1 E2E_GCP_TARGET=real go test -test.count=1 -timeout 600s -v ./tests/e2e -run TestAllInSeries/fixtures/linearlogmetric
```

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/logging/v1beta1/logginglogmetric/explicitlogmetric/_generated_object_explicitlogmetric.golden.yaml
+++ b/pkg/test/resourcefixture/testdata/basic/logging/v1beta1/logginglogmetric/explicitlogmetric/_generated_object_explicitlogmetric.golden.yaml
@@ -1,0 +1,56 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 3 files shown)


gemmahou:
I made the change in #1447, but looking at the golden files, I think we may not need to have the test name in the file name because we place the golden file under its corresponding test folder, so we know it belongs to that test.

i.e.
Before: logginglogmetric/explicitlogmetric/_generated_object_explicitlogmetric.golden.yaml
After: logginglogmetric/explicitlogmetric/_generated_object.golden.yaml

Would like to hear your opinion, @justinsb.

If we agree on this, I can make the code change and regenerate files very quick. And let's hold this PR for now. Sorry for the inconvenience, I feel as we have more generated files, we'll need to create a documentation about the naming and the usage of them.

justinsb:
Agreed on the naming being something we should change, and I think a code comment where we write the files is all we need.  I think a reasonable name is probably:  `{fixture.SourceDir}/_objects/{kind}_{name}.yaml` (sticking to the convention proposed in another PR that generated files start with `_` or are in a directory that starts with `_`).   I think `{fixture.SourceDir}/_object_{kind}_{name}.yaml` would also work.  But happy to riff on ideas (e.g. do we need the namespace?)  I think that's a separate PR thought.

I don't know that we need to hold this PR, as long as @acpana is willing to rename the golden files (or just regenerate them).  I think the github diff tooling will detect them as a rename if they are just renames (and can detect diffs across renames with small changes).

acpana:
Happy to rename (and regenerate) the files to the new format -- should that be done as part of this PR ?

justinsb:
> Happy to rename (and regenerate) the files to the new format -- should that be done as part of this PR ?

Either way, as you prefer.  I'll approve and hold and you can decide.  I think @gemmahou 's PRs to move the vcr test output have now landed.

/approve
/lgtm
/hold

acpana:
let me pull this in and follow up as needed 🙏🏼 

/hold cancel

PR #1500, opened 2024-04-03
mockgcp: improve StorageBucket mock

justinsb:
- tests: add golden http output for StorageBucket
- mockgcp: Update generated proto for storage API
- mockgcp: improve StorageBucket fidelity

PR #1563, opened 2024-04-15
mockgcp: use hand-coded storage proto (partially revert 1250)

jingyih:
Temporary fix for https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1250#issuecomment-2057929418


justinsb:
Thanks for doing this.

I'm hoping we can get #1500 in, which will fix this, but agree that if that's not happening today then we should fix HEAD.

jingyih:
Closing in favor of #1500.

PR #1564, opened 2024-04-16
vcr: Update output path

gemmahou:
### Change description

Fixes https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1552#discussion_r1564096664

### Tests you have done


tests/e2e/unified_test.go
```diff
@@ -171,6 +171,10 @@ func testFixturesInSeries(ctx context.Context, t *testing.T, testPause bool, can
 					create.MaybeSkip(t, fixture.Name, opt.Create)
 				}
 
+				if testVCR {
+					os.Setenv("VCR_PATH", fixture.SourceDir)
```
justinsb, line 175: I think I would prefer to pass this in to the harness via a field or method parameter (because env vars will break in parallel and are unobvious i.e. are invisible to static analysis).  If we need it in the constructor (`NewHarness`), then I propose creating a `HarnessOptions` struct.

tests/e2e/unified_test.go
```diff
@@ -711,3 +604,118 @@ func addTestTimeout(ctx context.Context, t *testing.T, timeout time.Duration) co
 
 	return ctx
 }
+
+func vcrConfig(t *testing.T, h *create.Harness) {
```
justinsb, line 608: Nit: Maybe `configureVCR`?  (Because `vcrConfig` sounds more like a function that returns the vcr configuration, rather than a function that sets the VCR configuration)

justinsb:
One nit over using an env var to pass parameters (and a suggestion around function naming), but generally lgtm.  I'm guessing this is going to need a rebase anyway, but we'll see what happens after #1552 merges...

tests/e2e/unified_test.go
gemmahou: Done

tests/e2e/unified_test.go
gemmahou: Done

justinsb:
Thanks @gemmahou 

/approve
/lgtm

PR #1568, opened 2024-04-16
tests: wait for cache to sync

acpana:
As we've turned on caching for CC & CCC I think we should wait on the informer caches to sync. This may actually reduce the flakes (from the `pause-tests` which are some of the only e2e tests today that test the fetching of CC/ CCC) .

### DIFF
--- a/config/tests/samples/create/harness.go
+++ b/config/tests/samples/create/harness.go
@@ -483,6 +483,9 @@ func NewHarness(ctx context.Context, t *testing.T) *Harness {
 		}
 	}()
 
+	// Given the informers a chance to sync
+	mgr.GetCache().WaitForCacheSync(ctx)
+
 	return h
 }
 
--- a/config/tests/samples/create/samples_test.go
+++ b/config/tests/samples/create/samples_test.go
@@ -301,6 +301,9 @@ func setup() {
 			logging.Fatal(err, "error starting manager")
 		}
 	}()
+
+	// Given the informers a chance to sync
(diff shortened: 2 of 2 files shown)


config/tests/samples/create/samples_test.go
```diff
@@ -301,6 +302,10 @@ func setup() {
 			logging.Fatal(err, "error starting manager")
 		}
 	}()
+
+	if couldNotSyncACache := mgr.GetCache().WaitForCacheSync(ctx); couldNotSyncACache {
```
justinsb, line 306: This _could_ be it, but I'm not sure.  I think cached reads are blocked until the caches are synced (I can't remember if the controllers _also_ wait, I think they do not).  If it's the pause tests though, it should be in harness, not here

acpana: yeah I realized after I ran a test locally that I had the wrong mgr 🤦🏼

config/tests/samples/create/harness.go
```diff
@@ -483,6 +483,9 @@ func NewHarness(ctx context.Context, t *testing.T) *Harness {
 		}
 	}()
 
+	// Given the informers a chance to sync
+	mgr.GetCache().WaitForCacheSync(ctx)
```
cheftako, line 487: If we need to wait for the cache anyway, would it make sense to wait for the cache to be synced and then run the tests?

anhdle-sso:
closing, will keep an eye if needed

PR #1570, opened 2024-04-16
WIP: Attempt to bypass shutdown issue

justinsb:


justinsb:
We don't seem to be hitting this problem currently

/close

PR #1576, opened 2024-04-17
tests: add switch board

acpana:
If we can short circuit the reconcile loop, we *should* avoid any panics that we've seen in the CI when we cancel a context because the test is finished. This patch allows us to signal to controllers that we don't want them to do any more work.

Follow ups:
* [ ] add to other controllers
* [ ] log if short circuiting

### DIFF
--- a/pkg/controller/switch.go
+++ b/pkg/controller/switch.go
@@ -1,0 +1,52 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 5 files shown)


pkg/controller/switch.go
```diff
@@ -0,0 +1,52 @@
...
+	defer r.lock.RUnlock()
+
+	return r.systemRunning
+}
+
+func GetSwitchBoard() *SwitchBoard {
+	return mainSwitch
+}
+
+func SetSwitchBoard(s *SwitchBoard) {
```
barney-s, line 48: Whats the advantage of having a local scoped var and setting it via SetSwitchBoard() vs having an Exported variable initialized using init() ?

acpana:
not needed anymore; still a good idea

PR #1586, opened 2024-04-17
WIP: Direct controller for MonitoringDashboard

justinsb:
- mockgcp: add golden output for monitoringdashboard
- mockgcp: fix a few inaccuracies in MonitoringDashboard mock
- genapi: Direct controller for MonitoringDashboard


pkg/controller/direct/monitoring/mapping.go
acpana, line 1: Nice! so IIUC, these should be auto-generable in the future, right? And keeping them somewhat with concrete types allows the code to be less magic-y?

justinsb: Not sure where this should have been bound to, but yes, I'm trying to see to what extend we can make this look like auto-generated code :-)

The `dest **pb.Dashboard` thing I think is unnecessary - easier just to return *pb.Dashboard.  It might be that some of the enum mapper generic functions benefit from a dest pointer, for type inference, but ... still iterating there!

justinsb:
I think we've merged the important stuff here in other PRs.

/close

PR #1587, opened 2024-04-17
chore: update hack scripts to generate golden objects

justinsb:
When we're using these scripts locally, we should capture/compare as
much golden output as we can, to maximize signal.


gemmahou:
Could you help my understand why we need both "ARTIFACTS" and "WRITE_GOLDEN_OUTPUT" in the script?

justinsb:
WRITE_GOLDEN_OUTPUT is our golden output, which we normally expect to commit back to the repository, so lives in the repo, under each test folder.  Currently it's more used for local testing.

ARTIFACTS is a location that we will archive and store alongside the test log after running the tests, primarily for periodic testing, but it's useful here too.  It captures a non-normalized http log.

PR #1590, opened 2024-04-18
tests: plumb context in samples

acpana:
Small patch adds a context plumbed through for the samples tests setup.


justinsb:
Looks good.  Is this just hygiene, or was there an actual issue?

/approve
/lgtm

PR #1594, opened 2024-04-18
tests: handle a nil object when normalizing responses

justinsb:
Probably shouldn't happen, but we shouldn't panic if it does.


acpana (approved):
Thank you 🥇 !

/lgtm
/approve

PR #1596, opened 2024-04-18
chore: Add kubebuilder labels & additionalPrinterColumns

justinsb:
- autogen
- chore: Add kubebuilder labels & additionalPrinterColumns


pkg/clients/generated/apis/logging/v1beta1/logginglogmetric_types.go
acpana, line 1: Nice, I have spot checked a couple of these and they lgtm overall.

Is there a good way for us to test these automatically moving forward? Or are they implicitly tested already?

acpana:
/hold 

qq -- would it be possible to put the command that you used to generate these in the description or a comment here? or the commit body? In the future, it may be helpful if we have to regenerate these

acpana:
/lgtm
/approve

thanks for this 💯 one non blocking question

justinsb:
> qq -- would it be possible to put the command that you used to generate these in the description or a comment here? or the commit body? In the future, it may be helpful if we have to regenerate these

God call - so I normally try to do that.  In this case, it's `make ready-pr` (or the faster `GOWORK=off make generate-go-client`) .. but I will next time!

/hold cancel

PR #1597, opened 2024-04-18
chore: generate int64 instead of int in client types

justinsb:
- chore: generate int64 instead of int in client types
- autogen: make generate-go-clients


justinsb:
/assign @acpana 

I think this gets rid of one problem in the directly-actuated resource

acpana:
thanks for writing this!

I_think_ this change won't cause any issues with #1645 since I already made these changes for the LLM CRD but I guess we will find out :) 

/lgtm
/approve

PR #1599, opened 2024-04-18
tests: use cmp.Diff for golden object check

acpana:
While trying to diff 2 KRM exports, I figured it could be easier to use `cmp.Diff` to get all the diff at once!

before
```
...
    utils.go:170: unexpected diff in <MY_PATH>/k8s-config-connector/pkg/test/resourcefixture/testdata/basic/logging/v1beta1/logginglogmetric/linearlogmetric/_generated_object_linearlogmetric.golden.yaml: field metricDescriptor in the golden file is missing

...
```

after
```
  utils.go:170: unexpected diff in<MY_PATH>/k8s-config-connector/pkg/test/resourcefixture/testdata/basic/logging/v1beta1/logginglogmetric/linearlogmetric/_generated_object_linearlogmetric.golden.yaml:   map[string]any{
                ... // 2 identical entries
                "metadata": map[string]any{"annotations": map[string]any{"cnrm.cloud.google.com/management-conflict-prevention-policy": string("none")}, "finalizers": []any{string("cnrm.cloud.google.com/finalizer"), string("cnrm.cloud.google.com/deletion-defender")}, "generation": float64(3), "labels": map[string]any{"cnrm-test": string("true")}, ...},
                "spec":     map[string]any{"bucketOptions": map[string]any{"linearBuckets": map[string]any{"numFiniteBuckets": float64(4), "offset": float64(0.5), "width": float64(2.5)}}, "description": string("An updated sample log metric"), "disabled": bool(true), "filter": string("resource.type=gae_app AND severity>=ERROR"), ...},
                "status": map[string]any{
                        "conditions": []any{map[string]any{"lastTransitionTime": string("1970-01-01T00:00:00Z"), "message": string("The resource is up to date"), "reason": string("UpToDate"), "status": string("True"), ...}},
                        "createTime": string("1970-01-01T00:00:00Z"),
        -               "metricDescriptor": map[string]any{
        -                       "description": string("An updated sample log metric"),
        -                       "name":        string("projects/${projectId}/metricDescriptors/logging.googleapis.com/user/logginglogmetric-${uniqueId}"),
        -                       "type":        string("logging.googleapis.com/user/logginglogmetric-${uniqueId}"),
        -               },
                        "observedGeneration": float64(3),
                        "updateTime":         string("1970-01-01T00:00:00Z"),
                },
          }
```

### DIFF
--- a/pkg/test/utils.go
+++ b/pkg/test/utils.go
@@ -15,7 +15,6 @@
 package test
 
 import (
-	"fmt"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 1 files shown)


gemmahou:
I think that's my intention, for KRM object golden testing, I only care about the fields in the object, values may not that important, and we'll need to normalize the generated values. However, people may have different opinions, see this [comment](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1447#discussion_r1550753910).

acpana:
The main thing that I'd like is to see the whole diff at once for the things that we care about that are not matching.

gemmahou:
> The main thing that I'd like is to see the whole diff at once for the things that we care about that are not matching.

That's fair. looks good to me!

/lgtm
/approve

acpana:
@anhdle-sso @justinsb I'm fairly certain I observed a situation in which the PR was merged here without all checks passing. Luckily, the check that was not passing was a flake (pretty confident).

More to the point, I think this situation can be reproduced on any PR that has a "lgtm" and "approve" label but at least a failing check. If that check is then retriggered manually (I did so on GH here), then the prow bot might race w the GH API and not realize that not ALL checks have completed/ passed.

PR #1605, opened 2024-04-19
chore: tool to add advanced OpenAPI validation for our ResourceRefs

justinsb:
It's not possible to do this with controller-runtime /
controller-tools annotations, so instead we post-process the CRD.
    
Adding the tool as-is, with a plan to polish and integrate into the
CRD generation pipeline when needed.

### DIFF
--- a/scripts/add-validation-to-crds/parse-crds.go
+++ b/scripts/add-validation-to-crds/parse-crds.go
@@ -1,4 +1,4 @@
-// Copyright 2022 Google LLC
+// Copyright 2024 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


scripts/add-validation-to-crds/parse-crds.go
```diff
@@ -0,0 +1,245 @@
+// Copyright 2022 Google LLC
```
acpana, line 1:
```suggestion
// Copyright 2024 Google LLC
```

acpana (commented):
/lgtm
/approve

I can give this a go with #1645

acpana:
thanks for rebasing 🙏🏼 

/lgtm
/approve

PR #1611, opened 2024-04-19
Add mockgcp Cloud sql v1beta4.

yuwenma:
### Change description

Fixes #

### Tests you have done


mockgcp/mocksql/operations_store.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+func (s *operations) startLRO(ctx context.Context, op *pb.Operation, obj proto.Message, callback func() (proto.Message, error)) (*pb.Operation, error) {
+	log := klog.FromContext(ctx)
+
+	now := time.Now()
+
+	switch obj := obj.(type) {
+	case *pb.DatabaseInstance:
+		op.TargetId = obj.Name
```
justinsb, line 43: Oooh, I like this approach!

pkg/webhook/manifests.go
```diff
@@ -26,7 +26,7 @@ import (
 )
 
 var (
-	admissionReviewVersions = []string{"v1"}
+	admissionReviewVersions = []string{"v1", "v1beta1"}
```
justinsb, line 29: Is this needed?  It looks accidental...  maybe split into another (non-test) PR if it is needed, or add comments etc

justinsb:
PR lgtm, other than the `admissionReviewVersions` change which looks like a rebase issue?  (I didn't poke around to see where it is used though....)

pkg/webhook/manifests.go
yuwenma: Not by accident. The mock test hits the webhook that only allows v1 version. My gut feeling is we may need to further separate some of the mock process from kcc code.  I'll dig deeper next week.

yuwenma: Fixed. It really took me a while to figure out what happened. In short, [KCC development setup](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/scripts/environment-setup/repo-setup.sh) installs an older kube-apiserver (via kubebuilder v2.3.1) and failed to install a new one (in a later step, the e2e test uses) due to my cloudtop permissions.  The old kube-apiserver only supports v1beta1 but not v1 that breaks the webhook admission review.  What confused me and sent me to the wrong direction is that I'm in a new test env (no cluster, no docker, no gcloud default setup) so I was assuming the dependencies were all up to date.

justinsb:
Thanks @yuwenma  - great stuff!

/approve
/lgtm

PR #1624, opened 2024-04-23
refactor: stage gostructs as sot for logMetric

acpana:
Mostly a cp for relevant fields for logMetric resource.


apis/resources/logging/v1beta1/groupversion_info.go
```diff
@@ -0,0 +1,39 @@
...
+	// SchemeGroupVersion is the group version used to register these objects.
+	SchemeGroupVersion = schema.GroupVersion{Group: "logging.cnrm.cloud.google.com", Version: "v1beta1"}
+
+	// SchemeBuilder is used to add go types to the GroupVersionKind scheme.
+	SchemeBuilder = &scheme.Builder{GroupVersion: SchemeGroupVersion}
+
+	// AddToScheme is a global function that registers this API group & version to a scheme
+	AddToScheme = SchemeBuilder.AddToScheme
+
+	LoggingLogMetricGVK = schema.GroupVersionKind{
```
justinsb, line 34: FYI I've started putting these into the relevant types.go; it makes generation easier

apis/resources/logging/v1beta1/logmetric_types.go
```diff
@@ -0,0 +1,224 @@
...
+// limitations under the License.
+
+package v1beta1
+
+import (
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/clients/generated/apis/k8s/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type LogmetricBucketOptions struct {
```
justinsb, line 22: BTW, I think we will eventually want an annotation that links the go type to the underlying API (normally proto) type.  I think that will help the generators.  Not something we need to worry about yet!

apis/resources/logging/v1beta1/logmetric_types.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+type LogmetricExponentialBuckets struct {
+	/* Must be greater than 1. */
+	// +optional
+	GrowthFactor *float64 `json:"growthFactor,omitempty"`
+
+	/* Must be greater than 0. */
+	// +optional
+	NumFiniteBuckets *int `json:"numFiniteBuckets,omitempty"`
```
justinsb, line 49: If you want to switch this to int64, that is likely the correct type (and is not a breaking API change).

(float64 triggers warnings because apparently float's don't round-trip (?), but I don't think there's a non-breaking API change there ... at least I haven't convinced myself there is one)

apis/resources/logging/v1beta1/logmetric_types.go
```diff
@@ -0,0 +1,224 @@
...
+	// +optional
+	MetricDescriptor *LogmetricMetricDescriptor `json:"metricDescriptor,omitempty"`
+
+	/* Immutable. The Project that this resource belongs to. */
+	ProjectRef v1alpha1.ResourceRef `json:"projectRef"`
```
justinsb, line 149: My 2c: we'll likely want ProjectRef in the future, both so we can have the better comments (that talk about "projects" instead of "referrents") and because it will probably help our type-conversion logic.  But not a blocker!

justinsb:
Looks good and as it's purely additive I don't mind if we want to merge early and iterate!

apis/resources/logging/v1beta1/logmetric_types.go
acpana: I went ahead and changed the `*int`s to `*int64`s

acpana:
> Looks good and as it's purely additive I don't mind if we want to merge early and iterate!

Rebase onto main 👍🏼 

/hold cancel

acpana:
The direct tests should not be running on this branch 🤔

apis/resources/logging/v1beta1/logmetric_types.go
```diff
@@ -0,0 +1,242 @@
+// Copyright 2024 Google LLC
```
justinsb, line 1: Super-nit (not a blocker) ... not sure we need `resources` in the path .... I think "apis" covers it

justinsb:
Purely additive, so :+1: 

/approve
/lgtm

PR #1625, opened 2024-04-23
feat: direct controller for RedisCluster

justinsb:
Direct controller support for Redis Cluster (a new KCC resource)


pkg/controller/direct/redis/cluster/refs.go
```diff
@@ -0,0 +1,66 @@
...
+package cluster
+
+import (
+	"context"
+
+	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func normalizeProjectRef(ctx context.Context, reader client.Reader, src client.Object, ref *refs.ProjectRef) (*refs.ProjectRef, error) {
```
yuwenma, line 24: Any concerns about moving this to the apis/v1beta1/refs?

pkg/test/resourcefixture/testdata/basic/redis/v1alpha1/rediscluster/redisclustersimple/_generated_object_redisclustersimple.golden.yaml
```diff
@@ -0,0 +1,52 @@
...
+    createTime: "1970-01-01T00:00:00Z"
+    discoveryEndpoints:
+    - address: 10.128.0.2
+      port: 6379
+      pscConfig:
+        networkRef:
+          external: projects/${projectId}/global/networks/network-${uniqueId}
+    preciseSizeGb: 39
+    pscConnections:
+    - address: 10.128.0.2
```
yuwenma, line 40: nit: This has two `10.128.0.2`.

pkg/controller/direct/redis/cluster/rediscluster_controller.go
```diff
@@ -0,0 +1,284 @@
...
+	projectRef, err := refs.ResolveProject(ctx, kube, obj, &obj.Spec.ProjectRef)
+	if err != nil {
+		return nil, err
+	}
+	projectID := projectRef.ProjectID
+	if projectID == "" {
+		return nil, fmt.Errorf("cannot resolve project")
+	}
+
+	// TODO: Move from monitoring package into shared package (and make refs implement an interface)
```
yuwenma, line 106: I like this.

pkg/controller/direct/redis/cluster/rediscluster_controller.go
```diff
@@ -0,0 +1,284 @@
...
+	desiredProto := RedisClusterSpec_ToProto(mapCtx, &obj.Spec)
+	if mapCtx.Err() != nil {
+		return nil, mapCtx.Err()
+	}
+
+	return &redisClusterAdapter{
+		projectID:      projectID,
+		location:       location,
+		resourceID:     resourceID,
+		desired:        desiredProto,
```
yuwenma, line 121: Oh, and we can also group the only-KCC "desired" field to structs for different purposes to normalize the solution, like the externalRef (real ID) https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/cloudbuild/externalresource.go#L28

pkg/controller/direct/redis/cluster/rediscluster_controller.go
```diff
@@ -0,0 +1,284 @@
...
+	if err != nil {
+		return fmt.Errorf("waiting for redisCluster create %s: %w", a.fullyQualifiedName(), err)
+	}
+
+	log.V(0).Info("created redisCluster", "redisCluster", created)
+
+	resourceID := lastComponent(created.Name)
+	if err := unstructured.SetNestedField(u.Object, resourceID, "spec", "resourceID"); err != nil {
+		return fmt.Errorf("setting spec.resourceID: %w", err)
+	}
```
yuwenma, line 212: Is this write back to spec? I think RedisCluster is alpha resource and we don't need to write it back or maybe I mis-understand this.

pkg/controller/direct/redis/cluster/refs.go
justinsb: Good call, I think this predates that / is the origin of that function.  I'll move!

pkg/controller/direct/redis/cluster/rediscluster_controller.go
justinsb: This is writing back to spec, yes.  Remember this code was originally written 3 months ago ... basically prehistoric.

I don't think we've written down (on GitHub?) that we don't want to write back to spec.resourceID for new SciFi resources.  I'd be in favor of that cc @cheftako

pkg/test/resourcefixture/testdata/basic/redis/v1alpha1/rediscluster/redisclustersimple/_generated_object_redisclustersimple.golden.yaml
justinsb: I think we've normalized the address to always be 10.128.0.2, because it's assigned by GCP

pkg/controller/direct/redis/cluster/rediscluster_controller.go
justinsb: Right!

pkg/controller/direct/redis/cluster/refs.go
justinsb: Actually, why don't we do separate PR to create a ref normalizer in a shared package (i.e. move it out of monitoring).  We're already using refs in the body of this function...

I'm happy to send that unless you want to...

pkg/controller/direct/redis/cluster/rediscluster_controller.go
justinsb: I changed it not to write back to spec!

PR #1628, opened 2024-04-23
tests: ignore duplicated GET requests

justinsb:
- tests: don't record duplicate GETs in golden log
- tests: update golden output for iamserviceaccount test

### DIFF
--- a/tests/e2e/unified_test.go
+++ b/tests/e2e/unified_test.go
@@ -397,6 +397,25 @@ func testFixturesInSeries(ctx context.Context, t *testing.T, testPause bool, can
 						return false
 					})
 
+					// Remove repeated GET requests
+					{
+						var previous *test.LogEntry
+						events = events.KeepIf(func(e *test.LogEntry) bool {
+							keep := true
+							if e.Request.Method == "GET" && previous != nil {
+								if previous.Request.Method == "GET" && previous.Request.URL == e.Request.URL {
+									if previous.Response.Status == e.Response.Status {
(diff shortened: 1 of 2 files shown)


justinsb:
/assign @gemmahou 

Not sure what people think.  Because we're comparing the body also, I think this helps get rid of some of the strictly unnecessary GETs which terraform makes, which makes the diff _much_ smaller (basically "real") when switching to a direct controller.

tests/e2e/unified_test.go
```diff
@@ -375,6 +375,25 @@ func testFixturesInSeries(ctx context.Context, t *testing.T, testPause bool, can
...
 
+					// Remove repeated GET requests
+					{
+						var previous *test.LogEntry
+						events = events.KeepIf(func(e *test.LogEntry) bool {
+							keep := true
+							if e.Request.Method == "GET" && previous != nil {
+								if previous.Request.Method == "GET" && previous.Request.URL == e.Request.URL {
+									if previous.Response.Status == e.Response.Status {
+										if previous.Response.Body == e.Response.Body {
```
gemmahou, line 386: Is that possible the response body is the same but reordered? I encountered this before, so that I [sort JSON ](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/tests/e2e/unified_test.go#L645) before comparing.

justinsb: I guess it's possible, but I don't think I've encountered it yet (_most_ JSON libraries sort the keys I believe).  Let's tackle it when we do?

PR #1629, opened 2024-04-23
tests: CompareGoldenObject should follow the golden output pattern

justinsb:
WRITE_GOLDEN_OUTPUT should always write the updated output, for use
with e.g. `git diff`

### DIFF
--- a/pkg/test/utils.go
+++ b/pkg/test/utils.go
@@ -142,18 +142,18 @@ func IgnoreLeadingComments(s string) string {
 	return strings.TrimSpace(strings.Join(out, "\n")) + "\n"
 }
 
-func CompareGoldenObject(t *testing.T, path string, got []byte) {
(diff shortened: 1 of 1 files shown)


pkg/test/utils.go
```diff
@@ -165,7 +165,19 @@ func CompareGoldenObject(t *testing.T, path string, got []byte) {
...
 	}
-	if diff := cmp.Diff(wantMap, gotMap); diff != "" {
-		t.Errorf("unexpected diff in %s: %s", path, diff)
+
+	diff := cmp.Diff(wantMap, gotMap)
+	if diff == "" {
+		return
+	}
+
+	if writeGoldenOutput {
```
acpana, line 174: ah yes good catch repeated writes on the file were not caught!

acpana:
/lgtm
/approve

thanks for adding this in 🥇

PR #1632, opened 2024-04-23
Direct support for NetworkConnectivityServiceConnectionPolicy

justinsb:
- Add NetworkConnectivityServiceConnectionPolicy test
- Implement direct controller for NetworkConnectivityServiceConnectionPolicy


apis/networkconnectivity/v1alpha1/serviceconnectivity.go
```diff
@@ -0,0 +1,213 @@
...
+	// ConsumerForwardingRule: The resource reference of the PSC Forwarding
+	// Rule within the consumer VPC.
+	ConsumerForwardingRule *string `json:"consumerForwardingRule,omitempty"`
+
+	// ConsumerTargetProject: The project where the PSC connection is
+	// created.
+	ConsumerTargetProject *string `json:"consumerTargetProject,omitempty"`
+
+	// // Error: The most recent error during operating this connection.
+	// Error *GoogleRpcStatus `json:"error,omitempty"`
```
cheftako, line 145: Should we have something which either tracks or explains why these fields are commented out?  If because its deprecated maybe we should track that?

apis/networkconnectivity/v1alpha1/serviceconnectivity.go
```diff
@@ -0,0 +1,213 @@
...
+
+	// // ErrorInfo: Output only. The error info for the latest error during
+	// // operating this connection.
+	// ErrorInfo *GoogleRpcErrorInfo `json:"errorInfo,omitempty"`
+
+	// // ErrorType: The error type indicates whether the error is consumer
```
cheftako, line 151: Ditto, is this deprecated?

apis/networkconnectivity/v1alpha1/serviceconnectivity.go
```diff
@@ -0,0 +1,213 @@
...
+	ConsumerForwardingRule *string `json:"consumerForwardingRule,omitempty"`
+
+	// ConsumerTargetProject: The project where the PSC connection is
+	// created.
+	ConsumerTargetProject *string `json:"consumerTargetProject,omitempty"`
+
+	// // Error: The most recent error during operating this connection.
+	// Error *GoogleRpcStatus `json:"error,omitempty"`
+
+	// // ErrorInfo: Output only. The error info for the latest error during
```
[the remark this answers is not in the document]
cheftako: Are we intending to reflect this in status? Might be good to indicate where?

pkg/controller/direct/networkconnectivity/mapping.go
```diff
@@ -0,0 +1,115 @@
...
+	// Labels map[string]string `json:"labels,omitempty"`
+
+	out.Network = in.Network
+
+	out.PscConfig = PscConfig_FromProto(ctx, in.PscConfig)
+
+	out.ServiceClass = in.ServiceClass
+
+	return out
+}
```
cheftako, line 27: nice to have blank lines between functions.

pkg/controller/direct/networkconnectivity/serviceconnectionpolicy_controller.go
```diff
@@ -0,0 +1,304 @@
...
+	if err := a.waitForOp(ctx, op); err != nil {
+		return false, fmt.Errorf("waiting for serviceConnectionPolicy delete %s: %w", a.fullyQualifiedName(), err)
+	}
+
+	return true, nil
+}
+
+func (a *serviceConnectionPolicyAdapter) waitForOp(ctx context.Context, op *api.GoogleLongrunningOperation) error {
+	// TODO: Only wait a short time, then update status?
+	for {
```
cheftako, line 179: Can this wait forever?

cheftako:
Do we have a test for setting and unsetting all the relevant fields?

pkg/controller/direct/networkconnectivity/serviceconnectionpolicy_controller.go
```diff
@@ -0,0 +1,304 @@
...
+		if IsNotFound(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("deleting serviceConnectionPolicy %s: %w", a.fullyQualifiedName(), err)
+	}
+
+	if err := a.waitForOp(ctx, op); err != nil {
+		return false, fmt.Errorf("waiting for serviceConnectionPolicy delete %s: %w", a.fullyQualifiedName(), err)
+	}
+
```
anhdle-sso, line 173: Do we need to empty out resourceID to mark deletion complete?

justinsb: Good question.   I don't know if we're allowed to make spec (or status) changes once an object has been deleted.   Normally we would expect that the object would be deleted immediately after we remove the finalized, so it's a moot point.  But I think you're right, someone else could have added a finalized, so if we can remove the status field, we probably should.  (Although maybe the other finalizer might want the resource ID - IDK)

I think the problem today for spec.resourceID is that we don't really know whether we set the field or the user did (technically we can consult the field ownership metadata, but I don't think we want to keep doing that).  When we have status.resourceID / status.selfLink though, we will want a consistent answer here.

pkg/controller/direct/networkconnectivity/serviceconnectionpolicy_controller.go
justinsb: Yes.  We need to figure out the right behaviour here (e.g. ideally we would update status during the update, ideally we would not block one of our reconciler threads etc)

pkg/controller/direct/networkconnectivity/serviceconnectionpolicy_controller.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+	if !reflect.DeepEqual(a.desiredProto.PscConfig, a.actual.PscConfig) {
+		updateMask.Paths = append(updateMask.Paths, "psc_config")
+	}
+	if (a.desiredProto.ServiceClass) != (a.actual.ServiceClass) {
+		updateMask.Paths = append(updateMask.Paths, "service_class")
+	}
+
+	if len(updateMask.Paths) != 0 {
+
```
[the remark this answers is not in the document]
justinsb: Concur - I'll add it.  We can also try to build it into the contract, to nudge people to providing this information.

acpana:
cc @justinsb would you want to revive this?

ldanielmadariaga:

PR #1644, opened 2024-04-24
enh: Rate-limit requeues on IAMPartialPolicy

justinsb:
We already rate-limit requeues on IAMPolicy, extend this to cover
IAMPartialPolicy also.

Also add a big description as to why we're doing this as a
code-comment.


yuwenma:
/lgtm

Thanks for the comment!

PR #1648, opened 2024-04-25
Add Mock for GKEHubMembership

ziyue-101:
* Fixed `GkeHubFeatureMembership` test case by removing the spec.resourceId field (the others resources refer the projectID by its metadata.name).
* Added MockGcp for `GKEHubMembership`
* Enabled Mock test for `GKEHubFeatureMembership` as the resource is relying on GKEHubfeature and GKEHubmembership Mocks.

### DIFF
--- a/experiments/compositions/composition/config/manager/manager.yaml
+++ b/experiments/compositions/composition/config/manager/manager.yaml
@@ -17,9 +17,9 @@ kind: Namespace
 metadata:
   labels:
     control-plane: controller-manager
-    app.kubernetes.io/name: namespace
-    app.kubernetes.io/instance: system
-    app.kubernetes.io/component: manager
+    app.kubernetes.io/name: composition-controller-manager-namespace
+    app.kubernetes.io/instance: composition-controller-manager-system
+    app.kubernetes.io/component: composition-controller-manager
     app.kubernetes.io/created-by: composition
     app.kubernetes.io/part-of: composition
     app.kubernetes.io/managed-by: kustomize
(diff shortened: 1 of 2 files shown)


mockgcp/mockgkehub/gkehubfeature.go
```diff
@@ -87,8 +87,10 @@ func (s *GKEHubFeature) UpdateFeature(ctx context.Context, req *pb.UpdateFeature
 		switch path {
 		case "labels":
 			obj.Labels = req.Resource.GetLabels()
-		case "multiclusteringress":
+		case "spec":
```
justinsb, line 90: Wow - that's a very broad update!  I think we need to replace the whole spec here?

justinsb:
Thanks @ziyue-101 this is looking good.  I think if the client specifies a fieldMask of `spec`, we are supposed to update the _whole_ `spec` field, not just part of it, but I don't think that's a blocker.  (Note for anyone reading this: this is a field named "spec" in the GCP API, not a KRM spec field!)

/approve
/lgtm

justinsb:
I think the test failure is probably real, I've tried it a few times and it is failing in `TestAllInSeries/samples/service-mesh-feature-membership `, which looks related.

justinsb:
We can try generating a proto from the [gapic](https://github.com/googleapis/google-api-go-client/blob/main/gkehub/v1beta/gkehub-api.json) , the "reversed" proto schemas aren't identical but do tend to be up-to-date.

We should also ask the gkehub team to update their proto, usually it's just an oversight and updating the SDKs makes GCP better for everyone!

ziyue-101:
b/338112177 is fixed, trying to update protos

ziyue-101:
@justinsb  the `TestAllInSeries/samples/service-mesh-feature-membership ` is passing now after adding the missing protos.

ziyue-101:
$ git push -f origin 3ca649085:mockgkehubfeaturemembershio  
Still can't re-open the PR, Opening a new one instead

PR #1650, opened 2024-04-26
fix: Fix panic during export

justinsb:
This was reported by a customer; though it's not yet fully root-caused
we can avoid the panic and report some diagnostic information.


jingyih (commented):
/lgtm
/approve

PR #1655, opened 2024-04-26
tests: capture GRPC traffic also

justinsb:
- tfprovider: support intercepting GRPC requests
- tests: record grpc requests in our golden tests
- tests: capture golden output for bigtabletable test

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/bigtable/v1beta1/bigtabletable/_http.log
+++ b/pkg/test/resourcefixture/testdata/basic/bigtable/v1beta1/bigtabletable/_http.log
@@ -1,0 +1,521 @@
+GRPC /google.bigtable.admin.v2.BigtableInstanceAdmin/ListInstances
+
+{
+  "parent": "projects/${projectId}"
+}
+
+OK
+
+{}
+
+---
+
+GRPC /google.bigtable.admin.v2.BigtableInstanceAdmin/CreateInstance
+
+{
+  "clusters": {
+    "cluster1-${uniqueId}": {
+      "defaultStorageType": "SSD",
+      "encryptionConfig": {},
+      "location": "projects/${projectId}/locations/us-east1-b",
+      "serveNodes": 3
+    },
+    "cluster2-${uniqueId}": {
+      "defaultStorageType": "SSD",
+      "encryptionConfig": {},
+      "location": "projects/${projectId}/locations/us-west1-a",
+      "serveNodes": 3
+    }
+  },
+  "instance": {
+    "displayName": "BigtableSample",
+    "labels": {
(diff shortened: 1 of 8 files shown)


pkg/test/resourcefixture/testdata/basic/bigtable/v1beta1/bigtabletable/_http.log
acpana, line 1: This probably needs to be called something else then, right? Like `_grpc.golden.log` ?

justinsb: _maybe_ ... the issue is that it's still http traffic, we're just decoding the GRPC requests.  There are some http logs where we intermingle http and GRPC!

justinsb: We could maybe rename `_http.log` to `_requests.log` or `_gcp.log` - it's really the log of non-kube requests/responses.

justinsb:
/assign @yuwenma 

This is on the path for reproducing a bigtable issue (#1780)

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/transport/bigtable_client_factory.go
```diff
@@ -32,6 +33,10 @@ func (s BigtableClientFactory) NewInstanceAdminClient(project string) (*bigtable
 	opts = append(opts, option.WithTokenSource(s.TokenSource), option.WithUserAgent(s.UserAgent))
 	opts = append(opts, s.gRPCLoggingOptions...)
 
+	if GRPCUnaryClientInterceptor != nil {
```
yuwenma, line 36: Is this change from the upstream tf repo fetch  or we manually added the logic?

pkg/test/resourcefixture/testdata/basic/bigtable/v1beta1/bigtabletable/_http.log
yuwenma: I think keeping the name as it is should be good. From the testing level, it doesn't really matter where the resource comes from or talks to. We care more on the resource itself.

yuwenma:
/lgtm
/approve

The fail test looks suspicious. Could you rerun?

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/transport/bigtable_client_factory.go
justinsb: Added manually

PR #1593, opened 2024-04-18
Add support for ComputeNetworkFirewallPolicyRule

katrielt:
### Change description

Add support for ComputeNetworkFirewallPolicyRule

Fixes b/315045017

### Tests you have done

- [x] Run `make ready-pr` to ensure this PR is ready for review.
- [X] Perform necessary E2E testing for changed resources.

PR #1658, opened 2024-04-30
Move output-only fields to `status.observedState` for v1alpha1 only CRDs

maqiuyujoyce:
### Change description

UI and UX improve for #1593 .

All the v1alpha1 only CRDs now have output-only fields to `status.observedState`. And they won't be supporting `state-into-spec: merge` any more.
CRDs supporting both v1alpha1 and v1beta1 (`grep -l 'name: v1alpha1' ./config/crds/resources/* | xargs grep 'name: v1beta1'`) are not touched: these resources are controlled by the `v1alpha1KindsWithStateIntoSpecMergeSupport` and `v1alpha1KindsWithOutputOnlyFieldsUnderStatus` allowlists.

### Tests you have done


maqiuyujoyce:
We decided to only support spec fields as part of the observed state for TF-based resources after being explicitly requested by customers.

PR #1664, opened 2024-04-30
chore: client-generator should visit objects in maps

justinsb:
Some APIs may include objects in maps, we want to output those objects
also.


justinsb:
cc @acpana 

cc @cheftako

yuwenma:
/lgtm
/approve

Thanks for the core.

acpana:
I'm going to go ahead and 

/hold cancel

PR #1667, opened 2024-05-01
tests: mutable but unreadable scenario for llm

acpana:
Captured with:

```bash
E2E_KUBE_TARGET=envtest E2E_GCP_TARGET=mock RUN_E2E=1 \
  go test -test.count=1 -timeout 360s -v ./tests/e2e -run TestE2EScript/scenarios/direct/mutable_but_unreadable 2>&1
```

Follow on PR will add support for mutable but unreadable for llm direct.

### DIFF
--- a/tests/e2e/testdata/scenarios/README.md
+++ b/tests/e2e/testdata/scenarios/README.md
@@ -36,3 +36,5 @@ a top-level field `TEST` on the object:
 * Setting `TEST: WAIT-FOR-HTTP-REQUEST`along with `VALUE_PRESENT: your value` will apply the object
   and inspect the http log to check that the value in VALUE_PRESENT appears. The step will
   wait ~ seconds for that value to show up.
+
+* Setting `WRITE-KUBE-OBJECT: false` will not export the KRM objects of the KCC resource.
--- a/tests/e2e/testdata/scenarios/direct/mutable_but_unreadable/SCENARIO_README.md
+++ b/tests/e2e/testdata/scenarios/direct/mutable_but_unreadable/SCENARIO_README.md
@@ -1,0 +1,7 @@
+For the LoggingLogMetric resource the metadata fields are:
+
+```yaml
+    cnrm.cloud.google.com/mutable-but-unreadable-fields: '{"spec":{"metricDescriptor":{"launchStage":"EARLY_ACCESS","metadata":{"ingestDelay":"1s","samplePeriod":"1s"}}}}'
+```
+
+What we want to see is that if we update the spec (the user intent) of our LoggingLogMetric resource WITH the same values for our mutable but unreadable fields, that update DOES NOT cause an underlying cloud provider API call. That is, there is no http log for the apply.
--- a/tests/e2e/httplog.go
+++ b/tests/e2e/httplog.go
@@ -246,6 +246,7 @@ func IgnoreAnnotations(annotations map[string]struct{}) func(string) string {
 
 			if !ignore {
 				sb.WriteString(line)
+				sb.WriteString("\n")
 			}
 		}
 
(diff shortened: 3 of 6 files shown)


tests/e2e/testdata/scenarios/README.md
```diff
@@ -35,4 +35,6 @@ a top-level field `TEST` on the object:
 
 * Setting `TEST: WAIT-FOR-HTTP-REQUEST`along with `VALUE_PRESENT: your value` will apply the object
   and inspect the http log to check that the value in VALUE_PRESENT appears. The step will
-  wait ~ seconds for that value to show up.
\ No newline at end of file
+  wait ~ seconds for that value to show up.
+
+* Setting `WRITE-KUBE-OBJECT: false` will not export the KRM objects of the KCC resource.
```
gemmahou, line 40: nit: To be more consist with other flags, maybe change the name to `TEST: NOT-EXPORT-KRM-OBJECT` or something similar?

tests/e2e/testdata/scenarios/direct/mutable_but_unreadable/SCENARIO_README.md
```diff
@@ -0,0 +1,7 @@
+For the LoggingLogMetric resource the metadata fields are:
+
+```yaml
+    cnrm.cloud.google.com/mutable-but-unreadable-fields: '{"spec":{"metricDescriptor":{"launchStage":"EARLY_ACCESS","metadata":{"ingestDelay":"1s","samplePeriod":"1s"}}}}'
+```
+
+What we want to see is that if we update the spec (the user intent) of our LoggingLogMetric resource WITH the same values for our mutable but unreadable fields, that update DOES NOT cause an underlying cloud provider API call. That is, there is no http log for the apply.
```
gemmahou, line 7: I may not have enough context about this test, but from this README `there is no http log for the apply`, I expect to not see a POST request. But I still see there's one POST in _http00.log. Could you help me understand the expectation of this test?

tests/e2e/testdata/scenarios/README.md
justinsb: I think the idea is that this modifier can be added to any `TEST` ?

tests/e2e/testdata/scenarios/direct/mutable_but_unreadable/SCENARIO_README.md
justinsb: I think the second object corresponds to _http01.log, and there is no _http01.log.  Maybe we should write an empty golden file, it's probably more an accident that we don't than a deliberate design decision

justinsb:
This looks good I think, I'll add a hold in case we want to populate empty files (per @gemmahou 's comment), but we could also do that in another PR - if so feel free to cancel the hold

/approve
/lgtm
/hold

tests/e2e/testdata/scenarios/README.md
acpana: Yes I am trying to experiment with this to see if there is value in it. I can come back and remove this if we are/ I am not getting the right mileage from excluding objects.

tests/e2e/testdata/scenarios/direct/mutable_but_unreadable/SCENARIO_README.md
acpana: Yes maybe the README needs to spell out that we won't have an `_http01.log` file but I can come back here when I add / activate the tests for the direct controller.

> that we won't have an `_http01.log` file

NOTE that having GET requests to begin with would be a-ok because just as long as we don't see a POST/ PUT in that log that we wouldn't expect!

acpana:
/hold cancel

will carry over some of the feedback to the activation of these tests fro the direct llm controller

PR #1671, opened 2024-05-01
tests: custom diff checks for llms

acpana:
Captured with:

```bash
E2E_KUBE_TARGET=envtest E2E_GCP_TARGET=mock RUN_E2E=1 \
  go test -test.count=1 -timeout 360s -v ./tests/e2e -run TestE2EScript/scenarios/direct/custom_diff 2>&1 
```

### DIFF
--- a/tests/e2e/testdata/scenarios/direct/llm_custom_diff/value_type/label/script.yaml
+++ b/tests/e2e/testdata/scenarios/direct/llm_custom_diff/value_type/label/script.yaml
@@ -1,0 +1,67 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: logging.cnrm.cloud.google.com/v1beta1
+kind: LoggingLogMetric
+metadata:
+  name: logginglogmetric-${uniqueId}
+  annotations:
+    cnrm.cloud.google.com/state-into-spec: absent
+spec:
+  filter: "resource.type=gae_app AND severity<=ERROR"
+  metricDescriptor:
+    displayName: "a concise name"
+    labels:
+    - description: "a label description"
+      key: "testkey"
+      valueType: "STRING"
+    metricKind: "DELTA"
+    valueType: "DISTRIBUTION"
+  valueExtractor: "EXTRACT(jsonPayload.response)"
+  labelExtractors:
+    testkey: "EXTRACT(jsonPayload.request)"
+  bucketOptions:
+    explicitBuckets:
+      bounds:
+      - 1.5
+      - 4.5
+  projectRef:
+    external: "projects/${projectId}"
+---
+apiVersion: logging.cnrm.cloud.google.com/v1beta1
+kind: LoggingLogMetric
+metadata:
+  name: logginglogmetric-${uniqueId}
+  annotations:
+    cnrm.cloud.google.com/state-into-spec: absent
+spec:
+  filter: "resource.type=gae_app AND severity<=ERROR"
+  metricDescriptor:
+    displayName: "a concise name"
+    labels:
+    - description: "a label description"
+      key: "testkey"
+      valueType: ""
+    metricKind: "DELTA"
+    valueType: "DISTRIBUTION"
+  valueExtractor: "EXTRACT(jsonPayload.response)"
+  labelExtractors:
+    testkey: "EXTRACT(jsonPayload.request)"
+  bucketOptions:
+    explicitBuckets:
+      bounds:
+      - 1.5
+      - 4.5
+  projectRef:
+    external: "projects/${projectId}"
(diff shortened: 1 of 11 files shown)


maqiuyujoyce:
Thank you so much for adding the new scenarios, @acpana !

IIRC, setting a string field to `""` in general means `unsetting this field` in a KCC resource (though it might be different because the LoggingLogMetric is a DCL-based resource). Could you check that the fields set to `""` in the scenarios are specialized or setting a string field to `""` actually doesn't unset a string field? Also would be great if you can check the behavior of dropping a field and see whether it means unsetting or unmanaging (which I remember is the latter).

acpana:
> or setting a string field to "" actually doesn't unset a string field?

I think this is the case for DCL based resources, or at least inasmuch as LoggingLogMetric is considered. We can see that from https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1691/files#r1589854992

And actually, from the http logs too (the lack of a POST/PUT w the dropped fields).

I think for 1.118 we can aim for parity for the LLM as direct release but can explore the behavior of set, unmanage or unset a field past 1.118?

tests/e2e/testdata/scenarios/direct/llm_custom_diff/value_type/label/script.yaml
```diff
@@ -0,0 +1,67 @@
...
+  annotations:
+    cnrm.cloud.google.com/state-into-spec: absent
+spec:
+  filter: "resource.type=gae_app AND severity<=ERROR"
+  metricDescriptor:
+    displayName: "a concise name"
+    labels:
+    - description: "a label description"
+      key: "testkey"
+      valueType: ""
```
maqiuyujoyce, line 55: This may be a bit outside of the scope of custom diff: What's the POST request when this value is `""` in the direct controller? In the DCL controller, field seems to be dropped by the request:

`{"bucketOptions":{"explicitBuckets":{"bounds":[1.5,4.5]}},"filter":"resource.type=gae_app AND severity\u003c=ERROR","labelExtractors":{"testkey":"EXTRACT(jsonPayload.request)"},"metricDescriptor":{"displayName":"a concise name","labels":[{"description":"a label description","key":"testkey"}],"metricKind":"DELTA","valueType":"DISTRIBUTION"},"name":"logginglogmetric-e3leul22li5i2y55gzaa","valueExtractor":"EXTRACT(jsonPayload.response)"}`

For compatibility, I think most field (if not all) of empty values will be dropped during the POST. Just want to double check - is it how direct controllers handle resource creations?

maqiuyujoyce (commented):
/lgtm
/approve

/hold
I don't think it is a blocker for LLM release as long as you've tested the direct controller's behavior locally. But feel free to remove the /hold if needed.

Meanwhile, I really doubt if `STRING` should be considered a valid value at all. We should reach out to the API team to gather more context and ensure our resource is working properly without misleading information.

I'm also curious how "setting to an empty value" works in the direct controller in general.

tests/e2e/testdata/scenarios/direct/llm_custom_diff/value_type/label/script.yaml
acpana: > For compatibility, I think most field (if not all) of empty values will be dropped during the POST

I think we can drop it as a parity at first then decide if we want to emulate that.

tests/e2e/testdata/scenarios/direct/llm_custom_diff/value_type/metric_descriptor/script.yaml
```diff
@@ -0,0 +1,41 @@
...
+apiVersion: logging.cnrm.cloud.google.com/v1beta1
+kind: LoggingLogMetric
+metadata:
+  name: logginglogmetric-${uniqueId}
+  annotations:
+    cnrm.cloud.google.com/state-into-spec: absent
+spec:
+  filter: "resource.type=gae_app AND severity<=ERROR"
+  metricDescriptor:
+    displayName: "a concise name"
```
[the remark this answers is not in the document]
acpana: Sorry could you double check? I applied this yaml and a cloud object got created on my end:

```yaml
apiVersion: logging.cnrm.cloud.google.com/v1beta1
kind: LoggingLogMetric
metadata:
  name: logginglogmetric-manual-test-alex
  annotations:
    cnrm.cloud.google.com/state-into-spec: absent
spec:
  filter: "resource.type=gae_app AND severity<=ERROR"
  metricDescriptor:
    displayName: "a concise name"
    labels:
    - description: "a label description"
      key: "testkey"
      valueType: "STRING"
    metricKind: "DELTA"
    valueType: "DISTRIBUTION"
  valueExtractor: "EXTRACT(jsonPayload.response)"
  labelExtractors:
    testkey: "EXTRACT(jsonPayload.request)"
  bucketOptions:
    explicitBuckets:
      bounds:
      - 1.5
      - 4.5
  projectRef:
    external: REDACTED
```

> Looks like this test doesn't match the real-world scenario, and the HTTP log and K8s object information can't really be refreshed, right? So should we remove it?

To be clear, we do want to keep our `mock` and `real` in sync so there may be genuine concerns that they are not but from my testing (hitting `real` as the target and `mock`), I've not seen variance just yet.

acpana:
/hold cancel

I'm going to let this merge in and address (any) behavior disparities in the direct controller when we activate these tests. Right now they are just used to capture the "what happened before" w the DCL controller in this case and for us to have a baseline.

acpana:
> Meanwhile, I really doubt if `STRING` should be considered a valid value at all. 

Will probably want to continue this conversation with the team offline

tests/e2e/testdata/scenarios/direct/llm_custom_diff/value_type/label/script.yaml
maqiuyujoyce: Just to double check - is "dropping the field of empty values" an existing behavior in the direct controller, or is it something we plan to do?

PR #1675, opened 2024-05-02
Support output-only spec fields as observed fields for v1beta1 TF-based CRDs

maqiuyujoyce:
### Change description

Fixes b/333562492, b/330942090

With #1466, output-only fields that are not sensitive, not ignored, not already observed, not in the path of a list, not a list itself, and in v1beta1 TF-based CRDs are identified.

In this PR, those fields are added under `status.observedState` of the CRDs.

### Tests you have done


yuwenma:
I thought we would include the "already observed" fields in `status.observedState` as well with an awareness of the potential duplications (See [comment](https://docs.google.com/document/d/1eyRvo58pisZrVm8y0uU4ispTJWCahDO_WS7aVMD3D3k/edit?disco=AAABKkiFxRU)).

maqiuyujoyce:
> I thought we would include the "already observed" fields in status.observedState as well with an awareness of the potential duplications (See [comment](https://docs.google.com/document/d/1eyRvo58pisZrVm8y0uU4ispTJWCahDO_WS7aVMD3D3k/edit?disco=AAABKkiFxRU)).

They are included already. They are excluded in #1466 because that PR's purpose is to generate the list of the observed fields to be copied over to the service mappings; not excluding the already observed fields will result in duplication.

Fields that are already observed may be configurable fields and we should not override the existing `observedFields` list.

maqiuyujoyce:
We decided to only support spec fields as part of the observed state after being explicitly requested by customers.

PR #1676, opened 2024-05-02
feat: scifi: Direct controller for Alloy DB Cluster

anhdle-sso:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/pkg/controller/direct/alloydb/cluster_controller.go
+++ b/pkg/controller/direct/alloydb/cluster_controller.go
@@ -1,0 +1,301 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package alloydb
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"strings"
+	"time"
+
+	api "google.golang.org/api/alloydb/v1beta"
+	"google.golang.org/protobuf/types/known/fieldmaskpb"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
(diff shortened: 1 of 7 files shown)


pkg/controller/direct/alloydb/cluster_controller.go
```diff
@@ -0,0 +1,301 @@
...
+		log.V(0).Info("change detected: displayName")
+		updateMask.Paths = append(updateMask.Paths, "displayName")
+	}
+	if a.desired.ClusterType != a.actual.ClusterType {
+		log.V(0).Info("change detected: clusterType")
+		updateMask.Paths = append(updateMask.Paths, "clusterType")
+	}
+	if !reflect.DeepEqual(a.desired.EncryptionConfig, a.actual.EncryptionConfig) {
+		log.V(0).Info("change detected: encryption_config")
+		updateMask.Paths = append(updateMask.Paths, "encryption_config")
```
justinsb, line 241: BTW, I'm trying to figure out if we've documented anywhere what our behaviour should be here.  mockgcp uses grpc-gateway, which is very strict about the names, but I think real-gcp allows either proto or json format.  If you find any docs, please LMK!

pkg/controller/direct/alloydb/mapping.go
```diff
@@ -0,0 +1,231 @@
...
+	if in == nil {
+		return nil
+	}
+	out := &api.UserPassword{
+		User: ValueOf(in.User),
+	}
+	if in.Password.Value != nil {
+		out.Password = ValueOf(in.Password.Value)
+	} else {
+		out.Password = in.Password.ValueFrom.SecretKeyRef.Key
```
justinsb, line 159: Is this right?  You can add errors to `ctx *MapContext` if that's more appropriate.

One thing I'm not certain about is when we should resolve these i.e. whether we resolve them on demand during the mapping, or whether we try to resolve them all up-front and (here) replace ValueFrom with  Value.  For ProjectRefs, we would replace Name with External. etc

pkg/controller/direct/alloydb/mapping.go
```diff
@@ -0,0 +1,231 @@
...
+		DatabaseVersion:      LazyPtr(in.DatabaseVersion),
+		EncryptionInfo:       EncryptionInfo_KRMFromApi(ctx, in.EncryptionInfo),
+		MigrationSource:      MigrationSource_KRMFromApi(ctx, in.MigrationSource),
+		Name:                 LazyPtr(in.Name),
+		Uid:                  LazyPtr(in.Uid),
+	}
+	return out
+}
+
+func BackupSource_KRMFromApi(ctx *MapContext, in *api.BackupSource) []krm.ClusterBackupSourceStatus {
```
justinsb, line 180: Ouch ... do we have the cardinality wrong in our API?  Still, at least it looks like we have it wrong the less-problematic way.

justinsb:
A few comments, but nothing that can't be done in follow-on (as this is feature-flagged)  I'll add hold, but feel free to cancel the hold when you've seen the comments.

/approve
/lgtm
/hold

acpana:
hm, it seems this included some out of date interfaces for the direct controller

PR #1691, opened 2024-05-04
tests: add gcloud exporter for llm

acpana:
A bit of a hacky way to use gcloud for the field tests (for dcl resources). Set, unmanage and unset.

The exports are captured against `real` gcp:

```bash
E2E_KUBE_TARGET=envtest E2E_GCP_TARGET=real RUN_E2E=1  \
  go test -test.count=1 -timeout 360s -v ./tests/e2e -run TestE2EScript/scenarios/direct/llm_set_unset/basic 2>&1
```

#### PR todos
* [ ] document the `EXPORTER` control for the scenario based testing
* [ ] more hacking on the normalizer to replace timestamps

### DIFF
--- a/tests/e2e/export.go
+++ b/tests/e2e/export.go
@@ -15,9 +15,14 @@
 package e2e
 
 import (
+	"bytes"
+	"fmt"
+	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strings"
 
+	"github.com/GoogleCloudPlatform/k8s-config-connector/apis/resources/logging/v1beta1"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/config/tests/samples/create"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/cli/cmd/export"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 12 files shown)


tests/e2e/testdata/scenarios/direct/llm_set_unset/basic/_gcloud_export02.yaml
```diff
@@ -0,0 +1,27 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+createTime: "2024-05-04T02:04:36.418290817Z"
+description: an e2e test
+disabled: true
```
acpana, line 17: Going to double check assumptions on whether we want these to be "deleted" from the cloud provider's api representation or not.

acpana:
The `gcloud` cli seems to not be on GH runner. I think rather than adding it, we can use the current adapter or use the api directly.

tests/e2e/export.go
```diff
@@ -26,6 +31,59 @@ import (
...
+
+	cmd := exec.Command("gcloud", args...)
+	var out bytes.Buffer
+	cmd.Stdout = &out
+
+	if err := cmd.Run(); err != nil {
+		return nil, fmt.Errorf("error from cmd.Run: %w, stdout: %q", err, out.String())
+	}
+
+	cleanOutput := regexp.MustCompile(`(?ms)^(Updates are available.*$|To take a quick.*$)`).ReplaceAll(out.Bytes(), nil)
```
justinsb, line 59: Sigh, this should not be on stdout

tests/e2e/export.go
```diff
@@ -26,6 +31,59 @@ import (
...
+	resourceID, _, _ := unstructured.NestedString(obj.Object, "spec", "resourceID")
+	if resourceID == "" {
+		resourceID = obj.GetName()
+	}
+
+	gvk := obj.GroupVersionKind()
+	switch gvk.GroupKind() {
+	case v1beta1.LoggingLogMetricGVK.GroupKind():
+		e := &Exporter{
+			Cmds:         []string{"logging", "metrics", "describe"},
```
justinsb, line 73: I guess this doesn't work with mockgcp?

tests/e2e/export.go
```diff
@@ -26,6 +31,59 @@ import (
...
+
+func exportResourceWithGcloud(h *create.Harness, obj *unstructured.Unstructured) []byte {
+	resourceID, _, _ := unstructured.NestedString(obj.Object, "spec", "resourceID")
+	if resourceID == "" {
+		resourceID = obj.GetName()
+	}
+
+	gvk := obj.GroupVersionKind()
+	switch gvk.GroupKind() {
+	case v1beta1.LoggingLogMetricGVK.GroupKind():
```
justinsb, line 71: If we're going to do something temporary, I think I'd prefer we just add something to the harness, vs calling gcloud.

We do have getCloudResourceManagerClient, if it's temporary I wouldn't mind just fetching the logginglogmetric here - we can move that to the export later.

Not sure about how to sequence things though, the evidence _that_ it works is probably all we need, rather than having the PR itself merged.

acpana: I don't think this is a blocker for 1.118 and since writing this poc I've wisened up a bit and would argue against this patch (that relies on gcloud). I'd much rather use the export from #1768 

Can this be a 1.119 effort?

tests/e2e/export.go
acpana: One more reason why I should redo this w the export func from #1768

tests/e2e/export.go
justinsb: That SGTM!

acpana:
I think the work here is "moot" with supporting the `Export` interface 💯

PR #1696, opened 2024-05-06
Fix mock test for AlloyDB Cluster

anhdle-sso:
### Change description

- Fix String function for Mock AlloyDB Cluster
- Fix GetCluster function for the mock
- Fix the incorrect API group for alloydb
- Add normalizer for http log
- Add golden file output and its normalizer

### Tests you have done


justinsb:
Looks good.  I think this would be caught by hack/record-gcp / hack/compare-mock, we just aren't running them here?

I think this is in your other PR also, so I'll hold but you can remove hold as you see fit (I would guess the other one, otherwise you'll have to rebase that one) 

/approve
/lgtm
/hold

anhdle-sso:
> Looks good. I think this would be caught by hack/record-gcp / hack/compare-mock, we just aren't running them here?
> 
> I think this is in your other PR also, so I'll hold but you can remove hold as you see fit (I would guess the other one, otherwise you'll have to rebase that one)
> 
> /approve /lgtm /hold

There is a few blunders in the previous PR. The API group was incorrect so the mock test just skipped and the test showed as passed. For the record_gcp, I still have good amount of flakes due to duplicate GET requests. I will debug that issue around the duplicate GET requests in a separate thread. Also, I add the golden file and its normalizer; which was not the norm previously.

config/tests/samples/create/harness.go
```diff
@@ -542,7 +542,7 @@ func MaybeSkip(t *testing.T, name string, resources []*unstructured.Unstructured
 			}
 
 			switch gvk.GroupKind() {
-			case schema.GroupKind{Group: "alloydb.googleapis.com", Kind: "AlloyDBCluster"}:
+			case schema.GroupKind{Group: "alloydb.cnrm.cloud.google.com", Kind: "AlloyDBCluster"}:
```
justinsb, line 545: lol - sorry I missed that!

tests/e2e/unified_test.go
```diff
@@ -563,6 +582,9 @@ func testFixturesInSeries(ctx context.Context, t *testing.T, testPause bool, can
 					for k := range operationIDs {
 						normalizers = append(normalizers, ReplaceString(k, "${operationID}"))
 					}
+					for k := range networkIDs {
```
justinsb, line 585: Might be easier to put this in as a pathID, but not a big deal!

PR #1698, opened 2024-05-06
Fix the error when 'state-into-spec: merge' is configured for handwritten IAM resources

maqiuyujoyce:
### Change description

Fixes b/338572343.

Issue:
`merge` is considered invalid value for the `state-into-spec` annotation in the handwritten IAM resources.

Root cause:
TypeMeta is missing in the retrieved object in the handwritten IAM controllers. Without type metadata, the kind is `""`, and kind `""` indeed doesn't support `merge`.

### Tests you have done

### DIFF
--- a/pkg/controller/iam/auditconfig/iamauditconfig_controller.go
+++ b/pkg/controller/iam/auditconfig/iamauditconfig_controller.go
@@ -45,7 +45,6 @@ import (
 	"golang.org/x/sync/semaphore"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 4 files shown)


pkg/controller/iam/auditconfig/iamauditconfig_controller.go
```diff
@@ -156,6 +157,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		}
 		return reconcile.Result{}, err
 	}
+	// r.Get() overrides the TypeMeta to empty value, so need to configure it
```
justinsb, line 160: Did `Get()` actually clear it, or just not set it?

pkg/controller/iam/policymember/iampolicymember_controller.go
```diff
@@ -164,6 +165,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		}
 		return reconcile.Result{}, err
 	}
+	// r.Get() overrides the TypeMeta to empty value, so need to configure it
+	// after r.Get().
+	memberPolicy.TypeMeta = metav1.TypeMeta{
```
justinsb, line 170: Can we use `memberPolicy.SetGroupVersionKind(iamv1beta1.IAMPolicyMemberGVK)` -- I think it's equivalent?

pkg/controller/iam/auditconfig/iamauditconfig_controller.go
maqiuyujoyce: It cleared it. I configured TypeMeta before the Get call and it got removed after the Get call.

pkg/controller/iam/policymember/iampolicymember_controller.go
maqiuyujoyce: Good point. Done!

pkg/controller/iam/auditconfig/iamauditconfig_controller.go
justinsb: TIL!

justinsb (approved):

justinsb:
Thanks @maqiuyujoyce 

/approve
/lgtm

PR #1700, opened 2024-05-07
chore: update controller-gen to v0.14.0

justinsb:
- chore: Simplify operator makefile calling of controller-gen
- chore: update operator manifests for newer controller-gen


yuwenma:
/lgtm 

Great! I hit this the other day when running on cloudtop which that the controller-gen 0.10.0 is incompatible with the go v1.22. I was planning to fix it later on. You beat me!

PR #1713, opened 2024-05-07
Add basic gkehubfeaturemembership direct controller

ziyue-101:
### Change description

Add the direct controller for gkehubfeaturemembership

### Tests you have done

### DIFF
--- a/pkg/controller/direct/gkehub/featuremembership_controller.go
+++ b/pkg/controller/direct/gkehub/featuremembership_controller.go
@@ -1,0 +1,214 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package gkehub
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+
+	featureapi "google.golang.org/api/gkehub/v1beta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
+	krm "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/clients/generated/apis/gkehub/v1beta1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
+)
+
+const ctrlName = "gkehubfeaturemembership-controller"
+
+// AddGkeHubController creates a new controller and adds it to the Manager.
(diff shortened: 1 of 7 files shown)


pkg/controller/direct/gkehub/featuremembership_controller.go
```diff
@@ -0,0 +1,213 @@
...
+		return nil, err
+	}
+	apiObj, err := featureMembershipSpecKRMtoMembershipFeatureSpecAPI(&obj.Spec)
+	if err != nil {
+		return nil, err
+	}
+	return &gkeHubAdapter{
+		membershipID:  mID,
+		featureID:     fID,
+		projectID:     "projects/" + projectID,
```
justinsb, line 99: Super-nit (just a tip): personally I find it easier if a field named `projectID` is just the ID, so I would name this as `projectFQN` or `projectLink` or `projectPath` or something like that.  Or, more probably, I would just store the projectID and add `projects/` when I combine it.

We'll probably end up with something strong-typed anyway once we define shared functions around resolving projects, so not a big deal!

(FQN = Fully Qualified Name)

pkg/controller/direct/gkehub/featuremembership_controller.go
```diff
@@ -0,0 +1,213 @@
...
+	log.V(2).Info("successfully created gkehubfeaturemembership")
+	// no need to set the status from the api response for the  &krm.GKEHubFeatureMembershipStatus{} as the it only has generic status.
+	return nil
+}
+
+func (a *gkeHubAdapter) Update(ctx context.Context, u *unstructured.Unstructured) error {
+	log := klog.FromContext(ctx).WithName(ctrlName)
+	log.V(2).Info("updating object", "u", u)
+	actual := a.actual.MembershipSpecs[a.membershipID]
+	if !reflect.DeepEqual(a.desired.Configmanagement, &actual.Configmanagement) || !reflect.DeepEqual(a.desired.Policycontroller, &actual.Policycontroller) || !reflect.DeepEqual(a.desired.Mesh, &actual.Mesh) {
```
justinsb, line 159: I think you get away with this here because there are no output fields, I've been exploring some tricks to make this comparison easy even when there are output fields.  (E.g. if we map _back_ to Spec, then by definition those are only the input fields so we can compare spec with reflect.DeepEqual; but the downside is then we have to map from proto -> spec; but we might want that for export anyway; etc etc)

ziyue-101: Added a comment to indicate no output fields in the API objects.

pkg/controller/direct/gkehub/featuremembership_controller.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	// no need to set the status from the api response for &krm.GKEHubFeatureMembershipStatus{} as the it only has generic status.
+	return nil
+}
+
+// fullyQualifiedNameForMembership constructions a fully qualified name for a gkehub resource
+// to be used in API calls. The format expected is: "projects/*/locations/*/memberships/{membershipId}".
+// Func assumes values are well formed and validated.
+func fullyQualifiedNameForMembership(obj krm.GKEHubFeatureMembership) (string, error) {
+	membershipLocation := *obj.Spec.MembershipLocation
```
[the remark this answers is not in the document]
ziyue-101: Good call, Wrapped it around ValueOf()

pkg/controller/direct/gkehub/featuremembership_controller.go
ziyue-101: Remove the "projects/" prefix in the projectID

ziyue-101:
cc/@hosseingolestani
cc/@shumiao

ziyue-101:
@shumiao has reviewed. 
/unhold

PR #1714, opened 2024-05-07
Reenable skipped AlloyDBInstance tests

199201shubhamsahu:
Adding back the skipped AlloyDBInstance sample tests as it was suspected to raise alerts. However, it is safe to enable them now as these tests did not contribute to the alerts


justinsb:
Great news, thanks @199201shubhamsahu 

We also have @anhdle-sso working on an AlloyDB mock so we should be able to test these super quickly going forwards (including with fault injection, if you want to tweak the mock!)

/approve
/lgtm

Issue #1654, opened 2024-04-26
KCC support for BigLake tables


600lyy:
### Checklist

- [X] I did not find a related open enhancement request.
- [X] I understand that enhancement requests filed in the GitHub repository are by default low priority.
- [X] If this request is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Describe the feature or resource

As a DB Admin, I want be able to deploy and manage my BigLake resources using KCC

This resource is already available in KCC's [terraform subtree ](https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/master/third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/biglake), therefore there is no need to update terraform in KCC

### Additional information

Terraform examples:
https://cloud.google.com/bigquery/docs/create-cloud-storage-table-biglake#terraform

600lyy:
Confirmed that using the current BigQueryTable resource I can create a BigLake table following this example:

https://cloud.google.com/bigquery/docs/create-cloud-storage-table-biglake#terraform

However, I'm not able to set the option
> requirePartitionFilter

This property is only under *hivePartitioningOptions* and *timePartitioning*

I see that *require_partition_filter* was added in TF5.2 as a [top level property](*https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/bigquery_table#require_partition_filter)

This PR is to add *requirePartitionFilter* as a top level property in KCC bigquery resource

600lyy:
PR submitted: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1710

600lyy:
PR submitted: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1733

PR #1733, opened 2024-05-08
bigquery table require_partition_filter.patch

600lyy:
require_partition_filter.patch

Make require_partition_filter a top level field.
This will replace the same field in the time_partitioning.

Fixes [1654](https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/1654)

### DIFF
--- a/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_bigquerytables.bigquery.cnrm.cloud.google.com.yaml
+++ b/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_bigquerytables.bigquery.cnrm.cloud.google.com.yaml
@@ -375,6 +375,10 @@ spec:
                 - field
                 - range
                 type: object
+              requirePartitionFilter:
+                description: If set to true, queries over this table require a partition
+                  filter that can be used for partition elimination to be specified.
+                type: boolean
               resourceID:
                 description: Immutable. Optional. The tableId of the resource. Used
                   for creation and acquisition. When unset, the value of `metadata.name`
@@ -466,9 +470,10 @@ spec:
                       time.
                     type: string
                   requirePartitionFilter:
-                    description: If set to true, queries over this table require a
-                      partition filter that can be used for partition elimination
-                      to be specified.
+                    description: DEPRECATED. This field is deprecated; please use
+                      the top level field with the same name instead. If set to true,
+                      queries over this table require a partition filter that can
+                      be used for partition elimination to be specified.
                     type: boolean
                   type:
                     description: The supported types are DAY, HOUR, MONTH, and YEAR,
--- a/config/samples/resources/bigquerytable/bigquery_v1beta1_bigquerytable.yaml
+++ b/config/samples/resources/bigquerytable/bigquery_v1beta1_bigquerytable.yaml
@@ -23,6 +23,7 @@ spec:
   description: "BigQuery Sample Table"
   datasetRef:
     name: bigquerytabledep
+  requirePartitionFilter: true
   friendlyName: bigquerytable-sample
   externalDataConfiguration:
     autodetect: true
--- a/config/tests/samples/create/harness.go
+++ b/config/tests/samples/create/harness.go
@@ -743,8 +743,9 @@ func MaybeSkip(t *testing.T, name string, resources []*unstructured.Unstructured
 		// case "fullalloydbcluster":
 		case "apikeyskeybasic":
 		case "artifactregistryrepository":
-		case "bigqueryjob":
 		case "bigqueryconnectionconnection":
+		case "bigqueryjob":
+		case "bigquerytable":
 		case "custombudget":
 		case "certificatemanagercertificatemapentry":
 		case "httpsfunction":
--- a/pkg/clients/generated/apis/bigquery/v1beta1/bigquerytable_types.go
+++ b/pkg/clients/generated/apis/bigquery/v1beta1/bigquerytable_types.go
@@ -275,7 +275,7 @@ type TableTimePartitioning struct {
 	// +optional
 	Field *string `json:"field,omitempty"`
 
-	/* If set to true, queries over this table require a partition filter that can be used for partition elimination to be specified. */
+	/* DEPRECATED. This field is deprecated; please use the top level field with the same name instead. If set to true, queries over this table require a partition filter that can be used for partition elimination to be specified. */
 	// +optional
 	RequirePartitionFilter *bool `json:"requirePartitionFilter,omitempty"`
 
@@ -330,6 +330,10 @@ type BigQueryTableSpec struct {
 	/* If specified, configures range-based partitioning for this table. */
 	// +optional
 	RangePartitioning *TableRangePartitioning `json:"rangePartitioning,omitempty"`
+
+	/* If set to true, queries over this table require a partition filter that can be used for partition elimination to be specified. */
+	// +optional
+	RequirePartitionFilter *bool `json:"requirePartitionFilter,omitempty"`
 
 	/* Immutable. Optional. The tableId of the resource. Used for creation and acquisition. When unset, the value of `metadata.name` is used as the default. */
 	// +optional
--- a/pkg/clients/generated/apis/bigquery/v1beta1/zz_generated.deepcopy.go
+++ b/pkg/clients/generated/apis/bigquery/v1beta1/zz_generated.deepcopy.go
@@ -646,6 +646,11 @@ func (in *BigQueryTableSpec) DeepCopyInto(out *BigQueryTableSpec) {
 		*out = new(TableRangePartitioning)
 		**out = **in
 	}
+	if in.RequirePartitionFilter != nil {
+		in, out := &in.RequirePartitionFilter, &out.RequirePartitionFilter
+		*out = new(bool)
+		**out = **in
+	}
 	if in.ResourceID != nil {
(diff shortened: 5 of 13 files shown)


yuwenma:
/lgtm

Thanks for the contribution, @600lyy PR looks good to me. 
I'll defer to @maqiuyujoyce for kcc-tf requirements.

maqiuyujoyce (commented):
Hi @600lyy , thank you for supporting the new field! It looks like a slightly tricky deprecation use case. I verified that the two `requirePartitionFilter` field can both be configured at the same time with different values, and the POST API calls captures the values properly - is it the expected behavior? If not, additional logic might be needed to do the check.

600lyy:
> Hi @600lyy , thank you for supporting the new field! It looks like a slightly tricky deprecation use case. I verified that the two `requirePartitionFilter` field can both be configured at the same time with different values, and the POST API calls captures the values properly - is it the expected behavior? If not, additional logic might be needed to do the check.

Thanks @maqiuyujoyce for your comment. 
Where did you set the two _requirePartitionFilter_ fields?

maqiuyujoyce:
This is the test yaml I used (you'll want to replace `${uniqueId}` and `${projectId}`):

```
apiVersion: bigquery.cnrm.cloud.google.com/v1beta1
kind: BigQueryTable
metadata:
  name: bigquerytablesample${uniqueId}
  annotations:
spec:
  friendlyName: bigquerytable-sample
  datasetRef:
    name: bigquerydatasetsample${uniqueId}
  requirePartitionFilter: false
  timePartitioning:
    type: DAY
    requirePartitionFilter: true
```

600lyy:
> This is the test yaml I used (you'll want to replace `${uniqueId}` and `${projectId}`):
> 
> ```
> apiVersion: bigquery.cnrm.cloud.google.com/v1beta1
> kind: BigQueryTable
> metadata:
>   name: bigquerytablesample${uniqueId}
>   annotations:
>     cnrm.cloud.google.com/project-id: ${projectId}
> spec:
>   friendlyName: bigquerytable-sample
>   datasetRef:
>     name: bigquerydatasetsample${uniqueId}
>   requirePartitionFilter: false
>   timePartitioning:
>     type: DAY
>     requirePartitionFilter: true
> ```

Those two fields with the same name are treated as conflicting fields in terraform 
```
~/workspace/tmp/terraform-test ❯ terraform apply                                                                              10776 11:36:14
╷
│ Error: Conflicting configuration arguments
│
│   with google_bigquery_table.default,
│   on main.tf line 17, in resource "google_bigquery_table" "default":
│   17:   require_partition_filter = false
│
│ "require_partition_filter": conflicts with time_partitioning.0.require_partition_filter
```

But KCC currently seems to allow for conflicting fields to exist in the same object so I'm not sure where to add that additional check?
Should I move timePartition. requirePartitionFilter to ignoredFields as it is due to deprecated?

maqiuyujoyce:
> But KCC currently seems to allow for conflicting fields to exist in the same object so I'm not sure where to add that additional check?

The conflicting fields are at TF-level but not KCC-level. KCC will respect the CRD instead of TF schema.
It can probably be done as an additional feature in KCC (i.e. supporting the generation of the conflicting schema in CRD based on TF schema), or we may implement the check in the local TF lib. Let's discuss offline for further details.

> Should I move timePartition. requirePartitionFilter to ignoredFields as it is due to deprecate?

Unfortunately, no. BigQueryTable is a v1beta1 CRD that we need to ensure backwards compatibility of. Removing a field (ignoring the field will remove it from the CRD) is a breaking change that we don't allow within the same version.

600lyy:
Hi @maqiuyujoyce ,
I was trying to put check sin the [local TF provider] (*https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/bigquery/resource_bigquery_table.go)

During testing, I noted that **time_partitioning.require_partition_filter** always gets a default value (false) when creating a new table even if the field is omitted in the KCC YAML. Do you know is it a KCC behaviour of adding zero value to fields in a  list object? ( I see **time_partitioning.expiration_ms** is also assigned a value of 0 if not specified).

With this limitation, it is very hard to resolve the conflict since we don't know if the *false* value is set by users or by KCC.

maqiuyujoyce:
> Do you know whether it is a KCC behaviour of adding zero value to fields in a list object? 

I believe not.

I verified by printing out the [config](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/v1.118.2/pkg/controller/tf/controller.go#L333) (desired state) and [diff](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/v1.118.2/pkg/controller/tf/controller.go#L347) with the following YAML:
```
apiVersion: bigquery.cnrm.cloud.google.com/v1beta1
kind: BigQueryTable
metadata:
  name: bigquerytablesample${uniqueId}
  annotations:
spec:
  friendlyName: bigquerytable-sample
  datasetRef:
    name: bigquerydatasetsample${uniqueId}
 # requirePartitionFilter: false
  timePartitioning:
    type: DAY
    requirePartitionFilter: true
```

And neither of them has the commented-out, top-level `requirePartitionFilter` field.

Config:
```
&{ComputedKeys:[] Raw:map[dataset_id:bigquerydatasetsamplezpcrpgtojmaqyrykhxea friendly_name:bigquerytable-sample 
labels:map[cnrm-test:true managed-by-cnrm:true] project:project-id 
table_id:bigquerytablesamplezpcrpgtojmaqyrykhxea time_partitioning:[map[require_partition_filter:true type:DAY]]] 
Config:map[dataset_id:bigquerydatasetsamplezpcrpgtojmaqyrykhxea friendly_name:bigquerytable-sample 
labels:map[cnrm-test:true managed-by-cnrm:true] project:project-id 
table_id:bigquerytablesamplezpcrpgtojmaqyrykhxea time_partitioning:[map[require_partition_filter:true type:DAY]]]}
```

Diff:
```
&{mu:{state:0 sema:0} Attributes:map[creation_time:0xc00921cec0 dataset_id:0xc00921d400 
deletion_protection:0xc00921d140 etag:0xc00921d380 expiration_time:0xc00921d300 friendly_name:0xc00921d340 
labels.%:0xc00921cf00 labels.cnrm-test:0xc00921cf40 labels.managed-by-cnrm:0xc00921cf80 
last_modified_time:0xc00921cfc0 location:0xc00921d100 num_bytes:0xc00921d3c0 
num_long_term_bytes:0xc00921d180 num_rows:0xc00921d240 project:0xc00921d2c0 self_link:0xc00921d1c0 
table_id:0xc00921d280 time_partitioning.#:0xc00921d000 time_partitioning.0.expiration_ms:0xc00921d040 
time_partitioning.0.require_partition_filter:0xc00921d0c0 time_partitioning.0.type:0xc00921d080 type:0xc00921d200] 
Destroy:false DestroyDeposed:false DestroyTainted:false RawConfig:{ty:{typeImpl:<nil>} v:<nil>} RawState:{ty:{typeImpl:
<nil>} v:<nil>} RawPlan:{ty:{typeImpl:<nil>} v:<nil>} Meta:map[]}
```

Same thing happens when I commented out `timePartitioning.requirePartitionFilter`. So I believe the defaulted empty values are from TF itself.

> During testing, I noted that raw["require_partition_filter"] always gets a default value (false) when creating a new table even if the field is omitted in the KCC YAML.

Just to double check, have you verified that the `ok` variable is `true`?

600lyy:
@maqiuyujoyce  

I believe the default value comes from API. 
Do you know in terraform provider, is there a good way to distinguish a value specified by user and a value set by API?
`d.GetOkExists / d.GetOK` both get value from tf state file, making it hard to tell if user sets the value explicitly.

regadas (approved):

yuwenma:
Actually we don't have any presubmit test coverage for `BigQueryTable` yet. Since the BigQueryTable is already a beta resource, I think we want to be more careful about accepting PRs on this resource until either one of VCR or MockGCP tests is added.  
You can enable VCR by following this guide https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/d49935e79ccc9e42e4192b80cb9e87a4f120bfb1/tests/README.VCRTesting.md, or add a MockGCP by following this guide https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/d49935e79ccc9e42e4192b80cb9e87a4f120bfb1/mockgcp

maqiuyujoyce:
> Do you know in terraform provider, is there a good way to distinguish a value specified by user and a value set by API?
d.GetOkExists / d.GetOK both get value from tf state file, making it hard to tell if user sets the value explicitly.

I'm not familiar with TF provider enough to have a good answer. This can be a question to TF team (about their legacy code). Feel free to reach out offline about it.

600lyy:
> Actually we don't have any presubmit test coverage for `BigQueryTable` yet. Since the BigQueryTable is already a beta resource, I think we want to be more careful about accepting PRs on this resource until either one of VCR or MockGCP tests is added. You can enable VCR by following this guide https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/d49935e79ccc9e42e4192b80cb9e87a4f120bfb1/tests/README.VCRTesting.md, or add a MockGCP by following this guide https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/d49935e79ccc9e42e4192b80cb9e87a4f120bfb1/mockgcp
> 
> cc @maqiuyujoyce @gemmahou

Thanks!

PR #1737, opened 2024-05-08
chore: refactor unified_test to split out scenario execution

justinsb:
Working towards being able to run samples & fixtures the same way.

### DIFF
--- a/tests/e2e/unified_test.go
+++ b/tests/e2e/unified_test.go
@@ -146,9 +146,7 @@ func testFixturesInSeries(ctx context.Context, t *testing.T, testPause bool, can
 			t.Run(fixture.Name, func(t *testing.T) {
 				ctx := addTestTimeout(ctx, t, subtestTimeout)
 
-				uniqueID := testvariable.NewUniqueID()
-
-				loadFixture := func(project testgcp.GCPProject) (*unstructured.Unstructured, create.CreateDeleteTestOptions) {
+				loadFixture := func(project testgcp.GCPProject, uniqueID string) (*unstructured.Unstructured, create.CreateDeleteTestOptions) {
 					primaryResource := bytesToUnstructured(t, fixture.Create, uniqueID, project)
 
 					opt := create.CreateDeleteTestOptions{CleanupResources: true}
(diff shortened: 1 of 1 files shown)


tests/e2e/unified_test.go
```diff
@@ -170,9 +168,26 @@ func testFixturesInSeries(ctx context.Context, t *testing.T, testPause bool, can
...
+		}
+	})
+
+	// Do a cleanup while we can still handle the error.
+	t.Logf("shutting down manager")
+	cancel()
+}
+
+func runScenario(ctx context.Context, t *testing.T, testPause bool, fixture resourcefixture.ResourceFixture, loadFixture func(project testgcp.GCPProject, uniqueID string) (*unstructured.Unstructured, create.CreateDeleteTestOptions)) {
+	// Extra indentation to avoid merge conflicts
```
acpana, line 182: to avoid merge conflicts in this PR 😛 ? or in the future, in which case I'd be curious how this works to fool git diff hehe ? 👀

acpana (commented):
/hold question about that extra indent

I think I can see how we are going to reuse `runScenario` for the fixtures too.

/lgtm
/approve

tests/e2e/unified_test.go
```diff
@@ -170,9 +168,26 @@ func testFixturesInSeries(ctx context.Context, t *testing.T, testPause bool, can
...
+			})
+		}
+	})
+
+	// Do a cleanup while we can still handle the error.
+	t.Logf("shutting down manager")
+	cancel()
+}
+
+func runScenario(ctx context.Context, t *testing.T, testPause bool, fixture resourcefixture.ResourceFixture, loadFixture func(project testgcp.GCPProject, uniqueID string) (*unstructured.Unstructured, create.CreateDeleteTestOptions)) {
```
acpana, line 181: But also could we rename this to `runTest`? Scenario makes me think of the step/ scenario based testing but if that's on purpose then never mind 😛

tests/e2e/unified_test.go
yuwenma: Just learnt this trick today 😆

tests/e2e/unified_test.go
yuwenma: Let's have this PR goes in and then we can discuss/improve in follow up changes. 🥂

yuwenma:
/lgtm
/approve

Thank you!

PR #1739, opened 2024-05-08
Diversify AlloyDB test and sample regions

199201shubhamsahu:
Diversify AlloyDB test and sample regions

- [N.A] Perform necessary E2E testing for changed resources.

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/alloydb/v1beta1/alloydbbackup/fullalloydbbackup/create.yaml
+++ b/pkg/test/resourcefixture/testdata/basic/alloydb/v1beta1/alloydbbackup/fullalloydbbackup/create.yaml
@@ -18,11 +18,11 @@ metadata:
   name: alloydbbackup-${uniqueId}
 spec:
   clusterNameRef:
-    external: "projects/${projectId}/locations/us-east1/clusters/alloydbcluster${uniqueId}"
-  location: us-east1
+    name: alloydbcluster-${uniqueId}
+  location: europe-west10
   encryptionConfig:
     kmsKeyNameRef: 
-      external: "projects/${projectId}/locations/us-east1/keyRings/kmskeyring-${uniqueId}/cryptoKeys/kmscryptokey-${uniqueId}"
+      name: kmscryptokey-${uniqueId}
   projectRef:
     external: ${projectId}
   resourceID: alloydbbackup${uniqueId}
--- a/config/samples/resources/alloydbcluster/regular-cluster/alloydb_v1beta1_alloydbcluster.yaml
+++ b/config/samples/resources/alloydbcluster/regular-cluster/alloydb_v1beta1_alloydbcluster.yaml
@@ -17,7 +17,7 @@ kind: AlloyDBCluster
 metadata:
   name: alloydbcluster-sample-regular
 spec:
-  location: us-east1
+  location: asia-south2
   networkConfig:
     networkRef: 
       name: alloydbcluster-dep-regular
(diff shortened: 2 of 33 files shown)


199201shubhamsahu:
@maqiuyujoyce Are there any known `regions` or more precisely `gcp.resourceLocations` where the KCC projects used for tests not allowed to run?

pkg/test/resourcefixture/testdata/basic/alloydb/v1beta1/alloydbbackup/fullalloydbbackup/create.yaml
```diff
@@ -18,11 +18,11 @@ metadata:
   name: alloydbbackup-${uniqueId}
 spec:
   clusterNameRef:
-    external: "projects/${projectId}/locations/us-east1/clusters/alloydbcluster${uniqueId}"
-  location: us-east1
+    external: alloydbcluster-${uniqueId}
```
maqiuyujoyce, line 21: The value of `external` field should be the value accepted by `clusterName` field in the GCP API, which should be the relative resource name instead of the resource ID: https://cloud.google.com/alloydb/docs/reference/rest/v1beta/projects.locations.backups

I assume you want to use `name` instead of `external`?

pkg/test/resourcefixture/testdata/basic/alloydb/v1beta1/alloydbbackup/fullalloydbbackup/create.yaml
```diff
@@ -18,11 +18,11 @@ metadata:
...
+  location: europe-west10
   encryptionConfig:
     kmsKeyNameRef: 
-      external: "projects/${projectId}/locations/us-east1/keyRings/kmskeyring-${uniqueId}/cryptoKeys/kmscryptokey-${uniqueId}"
+      external: kmscryptokey-${uniqueId}
```
maqiuyujoyce, line 25: ditto

maqiuyujoyce (commented):
Thank you for the change, @199201shubhamsahu ! Have you verified those tests still pass? It's okay if you want to defer the runs to the integration test framework after the merge, though let me know if you need any context about the integration tests.

pkg/test/resourcefixture/testdata/basic/alloydb/v1beta1/alloydbbackup/fullalloydbbackup/create.yaml
199201shubhamsahu: Yes, thanks. Updated field to use name instead of external

pkg/test/resourcefixture/testdata/basic/alloydb/v1beta1/alloydbbackup/fullalloydbbackup/create.yaml
199201shubhamsahu: Done

199201shubhamsahu:
Yes, I plan to defer the runs to the integration test framework after the merge. Is there another better way to test all the changes at once, instead of running tests one by one for both sample tests and dynamic tests?

I assume the Presubmit will take care of all the dynamic tests, but not sure about the sample tests

maqiuyujoyce:
> Is there another better way to test all the changes at once, instead of running tests one by one for both sample tests and dynamic tests?

There is a internal workaround, but I think it's okay to defer it to the continuous integration tests. Will reach out if we run into new AlloyDB test failures after this merge.

> I assume the Presubmit will take care of all the dynamic tests, but not sure about the sample tests

Actually Presubmit on GitHub doesn't cover most of the dynamic tests or the samples test. It leverages some testdata to run vcr tests and some samples to run mock tests; none of the presubmit tests talks to the real GCP endpoint.

/lgtm
/approve

PR #1741, opened 2024-05-08
tests: add acquire tests

acpana:
This patch adds an acquire test for step based testing.

### DIFF
--- a/tests/e2e/script_test.go
+++ b/tests/e2e/script_test.go
@@ -24,6 +24,9 @@ import (
 	"testing"
 	"time"
 
+	api "google.golang.org/api/logging/v2"
+	"google.golang.org/api/option"
+
 	"github.com/GoogleCloudPlatform/k8s-config-connector/config/tests/samples/create"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test"
 	testcontroller "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test/controller"
@@ -115,6 +118,9 @@ func TestE2EScript(t *testing.T) {
 					}
 
 					switch testCommand {
+					case "CLOUD-APPLY":
+						applyToCloud(h, obj)
+						shouldGetKubeObject = false
 					case "APPLY":
 						applyObject(h, obj)
 						create.WaitForReady(h, obj)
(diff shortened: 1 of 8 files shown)


tests/e2e/testdata/scenarios/direct/logging/acquire/_http01.log
```diff
@@ -0,0 +1,199 @@
...
+
+{
+  "filter": "resource.type=gae_app AND severity\u003c=INFO",
+  "metricDescriptor": {
+    "labels": [],
+    "metricKind": "DELTA",
+    "unit": "1",
+    "valueType": "INT64"
+  }
+}
```
acpana, line 141: surprised to see this PUT from the dcl controller while acquiring. Curious what the diff computed was that required an upgrade, it looks like it's adding the `labels` slice at first sight.

acpana:
I ran this against `real` and then turned on the direct controller to prove that acquisition works in the happy path for using a direct controller for LLM.

tests/e2e/script_test.go
```diff
@@ -265,6 +271,40 @@ func TestE2EScript(t *testing.T) {
 	})
 }
 
+// applies objects directly to the cloud provider
```
justinsb, line 274: Could we instead apply and then abandon them?  Even if we use a custom verb to do that, this feels like it's duplicating our direct controller layer (?)

tests/e2e/testdata/scenarios/direct/logging/acquire/_http01.log
justinsb: Yeah, maybe we need the PR that logs the diff in the controller to be sure.

justinsb:
My inclination here is that we use the fact that this passes for confidence in our release, but we don't merge this in 1.118.  WDYT @acpana  ?

acpana:
Yes, this should not be a blocker for 1.118.

tests/e2e/script_test.go
acpana: We could do that but technically we will be testing the acquisition or adoption of a resource that we created and then un-managed. As opposed to to acquiring a previsouly unknown resource to us. Although I am not convinced that this distinction is meaningful That is, we can probably use the scenario you proposed to better effect.

If we do want to acquire a "clean" resource, I can expose the adapter or reuse some of the ResourceContext methods I added in #1768

acpana:
obsolete now after controller migration

PR #1745, opened 2024-05-09
feat: trigger ssa metadata for direct controller

acpana:
Following the DCL/ TF controllers to trigger ssa metadata annotation.

### DIFF
--- a/pkg/controller/direct/directbase/directbase_controller.go
+++ b/pkg/controller/direct/directbase/directbase_controller.go
@@ -188,6 +188,10 @@ func (r *DirectReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 
 func (r *reconcileContext) doReconcile(ctx context.Context, u *unstructured.Unstructured) (requeue bool, err error) {
 	logger := log.FromContext(ctx)
+	u, err = k8s.TriggerManagedFieldsMetadata(ctx, r.Reconciler.Client, u)
+	if err != nil {
+		return false, fmt.Errorf("error triggering Server-Side Apply (SSA) metadata: %w", err)
+	}
 
 	cc, ccc, err := kccstate.FetchLiveKCCState(ctx, r.Reconciler.Client, r.NamespacedName)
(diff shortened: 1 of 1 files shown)


maqiuyujoyce:
I traced back to the change when [TriggerManagedFieldsMetadata](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/k8s/meta.go) was added. It looks like this is used to enable the externally managed fields on clusters between K8s 1.16 and 1.18 because the metadata is not automatically triggered. 

I think it'll be safe if we keep it; though @yuwenma do you think it'll be more sustainable if we keep track of versions of clusters where Config Connector is installed so that we can clean up version-specific logic over time?

yuwenma:
I would say may be not a necessity. But I'm not very familiar with the KCC `cnrm.cloud.google.com/supports-ssa` annotation. By reading its doc https://cloud.google.com/config-connector/docs/concepts/managing-fields-externally#behavior_with_server-side_apply, it is not clear to me how it works with the k8s server-side apply.  @justinsb would you mind justifying?

justinsb:
As far as I can tell we never read this annotation nor the field manager.  I concur with @maqiuyujoyce that this is not needed in modern k8s version, where you basically always have managedFields.  Also, given that we aren't using managedFields in our direct controllers (at least not yet), I don't think we need this annotation.  If it turns out we do need it, we can always add it later.

justinsb:
/hold

Because I don't think we want this in 1.118.  If people agree, not sure if we move to 1.119 or just close?

maqiuyujoyce:
Yeah I'm okay with not having it as we can always add it if there are some unexpected issues.
Regarding `cnrm.cloud.google.com/supports-ssa`, I never had to interact with this annotation. But we should update the public doc to reflect our change: https://cloud.google.com/config-connector/docs/reference/annotations

yuwenma:
Close the PR to clear up the PR review list. We can re-open this PR if the question comes up to the table in the future.

PR #1746, opened 2024-05-09
feat: honor reconcile period anon in direct controller

acpana:
We already honor the `reconcile-period` annotation when generating a jittered value for re-enqueuing but we are not honoring it when it's 0.

### DIFF
--- a/pkg/controller/direct/directbase/directbase_controller.go
+++ b/pkg/controller/direct/directbase/directbase_controller.go
@@ -234,7 +234,7 @@ func (r *reconcileContext) doReconcile(ctx context.Context, u *unstructured.Unst
 		}
 		if k8s.HasFinalizer(u, k8s.DeletionDefenderFinalizerName) {
 			// deletion defender has not yet finalized; requeuing
-			logger.Info("deletion defender has not yet been finalized; requeuing", "resource", k8s.GetNamespacedName(u))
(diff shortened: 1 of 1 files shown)


pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -223,7 +233,7 @@ func (r *reconcileContext) doReconcile(ctx context.Context, u *unstructured.Unst
 			return false, nil
 		}
 		if k8s.HasFinalizer(u, k8s.DeletionDefenderFinalizerName) {
-			// deletion defender has not yet been finalized; requeuing
+			// deletion defender has not yet finalized; requeuing
 			logger.Info("deletion defender has not yet been finalized; requeuing", "resource", k8s.GetNamespacedName(u))
```
justinsb, line 237: Nit: I agree `not yet finalized` is clearer; want to change the logger.Info message also?

justinsb:
Thanks for fixing!

/approve
/lgtm
/hold in case you want to fix the nit in this PR

Feel free to self-remove hold

pkg/controller/direct/directbase/directbase_controller.go
acpana, line 237:
```suggestion
			// deletion defender has not yet finalized; requeuing
			logger.Info("deletion defender has not yet finalized; requeuing", "resource", k8s.GetNamespacedName(u))
```

PR #1750, opened 2024-05-09
feat: Support immutability checks for llm

gemmahou:
### Change description

Fixes b/339129197

Add immutability checks in webhook and llm controller.

Golden log captured using:
GOLDEN_REQUEST_CHECKS=1 E2E_KUBE_TARGET=envtest E2E_GCP_TARGET=real RUN_E2E=1 KCC_USE_DIRECT_RECONCILERS=LoggingLogMetric   go test -test.count=1 -timeout 360s -v ./tests/e2e -run TestE2EScript/scenarios/direct/immutable_field

### Tests you have done

### DIFF
--- a/pkg/controller/direct/logging/logmetric_controller.go
+++ b/pkg/controller/direct/logging/logmetric_controller.go
@@ -243,7 +243,11 @@ func (a *logMetricAdapter) Update(ctx context.Context, u *unstructured.Unstructu
 				update.Filter = a.actual.Filter
 			}
 		}
+
 		if !compareMetricDescriptors(a.desired.Spec.MetricDescriptor, a.actual.MetricDescriptor) {
+			if err := validateImmutableFieldsUpdated(a.desired.Spec.MetricDescriptor, a.actual.MetricDescriptor); err != nil {
+				return fmt.Errorf("logMetric update failed: %w", err)
+			}
 			update.MetricDescriptor = convertKCCtoAPIForMetricDescriptor(a.desired.Spec.MetricDescriptor)
 		}
 
--- a/pkg/controller/direct/logging/utils.go
+++ b/pkg/controller/direct/logging/utils.go
@@ -89,6 +89,28 @@ func compareMetricDescriptors(kccObj *krm.LogmetricMetricDescriptor, apiObj *api
 	return reflect.DeepEqual(kccObj, convertAPItoKRM_MetricDescriptor(apiObj))
 }
 
+func validateImmutableFieldsUpdated(kccObj *krm.LogmetricMetricDescriptor, apiObj *api.MetricDescriptor) error {
+	actualMetricDescriptor := convertAPItoKRM_MetricDescriptor(apiObj)
+
+	modified := []string{}
+	if kccObj == nil && apiObj == nil {
+		return nil
+	}
+	if kccObj == nil || apiObj == nil {
+		return fmt.Errorf("cannot make changes to immutable field: metricDescriptor")
+	}
+	if !reflect.DeepEqual(kccObj.MetricKind, actualMetricDescriptor.MetricKind) {
+		modified = append(modified, "metricDescriptor.MetricKind")
+	}
+	if !reflect.DeepEqual(kccObj.ValueType, actualMetricDescriptor.ValueType) {
+		modified = append(modified, "metricDescriptor.ValueType")
+	}
+	if len(modified) != 0 {
+		return fmt.Errorf("cannot make changes to immutable field(s): %v", modified)
+	}
+	return nil
+}
+
 func convertAPItoKRM_MetricDescriptorStatus(apiObj *api.MetricDescriptor) *krm.LogmetricMetricDescriptorStatus {
 	if apiObj == nil {
 		return nil
@@ -131,9 +153,9 @@ func convertAPItoKRM_LogMetricLabels(apiLabels []*api.LabelDescriptor) []krm.Log
 	kccLabels := make([]krm.LogmetricLabels, len(apiLabels))
 	for i, apiLabel := range apiLabels {
 		kccLabels[i] = krm.LogmetricLabels{
-			Description: &apiLabel.Description,
-			Key:         &apiLabel.Key,
-			ValueType:   &apiLabel.ValueType,
+			Description: &apiLabel.Description, // immutable
+			Key:         &apiLabel.Key,         // immutable
+			ValueType:   &apiLabel.ValueType,   // immutable
 		}
 	}
 	return kccLabels
@@ -290,6 +312,7 @@ func convertKCCtoAPIForMetricDescriptor(kccObj *krm.LogmetricMetricDescriptor) *
 	if kccObj.Metadata != nil {
 		metricDescriptor.Metadata = convertKCCtoAPIForLogMetricMetadata(kccObj.Metadata)
 	}
+	// immutable
 	if kccObj.MetricKind != nil {
 		metricDescriptor.MetricKind = ValueOf(kccObj.MetricKind)
 	}
--- a/pkg/webhook/immutable_fields_validator.go
+++ b/pkg/webhook/immutable_fields_validator.go
@@ -23,6 +23,8 @@ import (
 	"regexp"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
 	corekccv1alpha1 "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/apis/core/v1alpha1"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/dcl"
 	dclextension "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/dcl/extension"
@@ -36,7 +38,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/util/pathslice"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/util/typeutil"
 
-	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	tfschema "github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-provider-google-beta/google-beta/provider"
 	"github.com/nasa9084/go-openapi"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -55,7 +57,7 @@ var (
 
 type immutableFieldsValidatorHandler struct {
 	smLoader              *servicemappingloader.ServiceMappingLoader
-	tfResourceMap         map[string]*schema.Resource
+	tfResourceMap         map[string]*tfschema.Resource
 	dclSchemaLoader       dclschemaloader.DCLSchemaLoader
 	serviceMetadataLoader dclmetadata.ServiceMetadataLoader
 }
@@ -112,6 +114,12 @@ func (a *immutableFieldsValidatorHandler) Handle(_ context.Context, req admissio
 
 	if err := validateImmutableStateIntoSpecAnnotation(obj, oldObj); err != nil {
 		return admission.Errored(http.StatusForbidden, err)
+	}
+
+	gk := oldObj.GroupVersionKind().GroupKind()
+	switch gk {
+	case schema.GroupKind{Group: "logging.cnrm.cloud.google.com", Kind: "LoggingLogMetric"}:
+		return validateImmutableFieldsForLoggingLogMetricResource(oldSpec, spec)
 	}
 
 	if dclmetadata.IsDCLBasedResourceKind(obj.GroupVersionKind(), a.serviceMetadataLoader) {
@@ -278,7 +286,7 @@ func getQualifiedFieldName(prefix string, fieldName string) string {
 	return qualifiedName
 }
 
-func validateImmutableFieldsForTFBasedResource(obj, oldObj *unstructured.Unstructured, spec, oldSpec map[string]interface{}, smLoader *servicemappingloader.ServiceMappingLoader, tfResourceMap map[string]*schema.Resource) admission.Response {
+func validateImmutableFieldsForTFBasedResource(obj, oldObj *unstructured.Unstructured, spec, oldSpec map[string]interface{}, smLoader *servicemappingloader.ServiceMappingLoader, tfResourceMap map[string]*tfschema.Resource) admission.Response {
 	rc, err := smLoader.GetResourceConfig(obj)
 	if err != nil {
 		return admission.Errored(http.StatusBadRequest,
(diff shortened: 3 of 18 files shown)


pkg/webhook/immutable_fields_validator.go
```diff
@@ -114,6 +117,15 @@ func (a *immutableFieldsValidatorHandler) Handle(_ context.Context, req admissio
 		return admission.Errored(http.StatusForbidden, err)
 	}
 
+	gk := oldObj.GroupVersionKind().GroupKind()
+	//TODO(yuhou): remove `if` condition once we switch to llm direct controller
```
gemmahou, line 121: Can we remove this now?

tests/e2e/testdata/scenarios/direct/llm_immutable_field/description/_http01.log
```diff
@@ -0,0 +1,214 @@
...
+    "unit": "1",
+    "valueType": "INT64"
+  },
+  "name": "logginglogmetric-${uniqueId}",
+  "updateTime": "2024-04-01T12:34:56.123456Z"
+}
+
+---
+
+GET https://logging.googleapis.com/v2/projects/${projectId}/metrics/logginglogmetric-${uniqueId}?alt=json
```
acpana, line 35: I think Justin had a PR to remove some of the repeated GETs if you had time to rebase but no worries we can regenerate these.

acpana (commented):
/hold letting decide on merging

/lgtm
/approve

thanks for working on this 💯

tests/e2e/testdata/scenarios/direct/llm_immutable_field/metric_kind_and_value_type/_object01.yaml
```diff
@@ -0,0 +1,68 @@
...
+    labels:
+    - description: a label description
+      key: testkey
+      valueType: STRING
+    launchStage: EARLY_ACCESS
+    metadata:
+      ingestDelay: 1s
+      samplePeriod: 1s
+    metricKind: CUMULATIVE
+    valueType: INT64
```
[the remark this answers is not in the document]
gemmahou: Correct!

An additional thought as mentioned in our troubleshooting doc, [error_updating_the_immutable_fields_when_there_is_no_update](https://cloud.google.com/config-connector/docs/troubleshooting#error_updating_the_immutable_fields_when_there_is_no_update) `It's possible that the GCP API has made a change to an immutable field that was managed by you in the Config Connector resource. This caused the mismatch between the desired state and the live state of the immutable fields.` 

Cx will need to update their KCC resource to match the live state. I don't think this is a good UX, though. Maybe we can handle this scenario better with a direct control(in the long term).

pkg/controller/direct/logging/utils.go
```diff
@@ -76,6 +76,21 @@ func compareMetricDescriptors(kccObj *krm.LogmetricMetricDescriptor, apiObj *api
...
+	actualMetricDescriptor := convertAPItoKRM_MetricDescriptor(apiObj)
+	modified := []string{}
+	if !reflect.DeepEqual(kccObj.MetricKind, actualMetricDescriptor.MetricKind) {
+		modified = append(modified, "metricDescriptor.MetricKind")
+	}
+	if !reflect.DeepEqual(kccObj.ValueType, actualMetricDescriptor.ValueType) {
+		modified = append(modified, "metricDescriptor.ValueType")
+	}
+	if len(modified) != 0 {
+		return fmt.Errorf("cannot make changes to immutable field(s): %v", modified)
```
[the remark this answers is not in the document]
gemmahou: We have the "logMetric update failed:" prefix [here](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/cd9aaedfd1bdd21bbaa0b85d90c93599d832e1a4/pkg/controller/direct/logging/logmetric_controller.go#L273)!

tests/e2e/testdata/scenarios/direct/llm_immutable_field/labels/_object01.yaml
```diff
@@ -0,0 +1,75 @@
...
+  - lastTransitionTime: "1970-01-01T00:00:00Z"
+    message: |-
+      Update call failed: error applying desired state: googleapi: Error 400: Cannot change value type of label: testkey
+      Details:
+      [
+        {
+          "@type": "type.googleapis.com/google.rpc.DebugInfo",
+          "detail": "[ORIGINAL ERROR] generic::invalid_argument: Cannot change value type of label: testkey [google.rpc.error_details_ext] { message: \"Cannot change value type of label: testkey\" }"
+        }
+      ]
```
gemmahou, line 65: I captured the golden files using DCL controller, the error messages are slightly different between DCL and direct. 
i.e. DCL: `Update call failed: error applying desired state: googleapi: Error 400: Cannot change value type of label: testkey`, direct: `Update call failed: logMetric update failed: googleapi: Error 400: Cannot change value type of label: testkey`

If we run the golden check with direct controller, the test will fail because of the diff in the error message. That is expected and validates that the direct controller behaves the same as DCL (with a different error message). I want to leave a note here for future reference in case anyone encounters issues with golden tests.

tests/e2e/testdata/scenarios/direct/llm_immutable_field/metric_kind_and_value_type/_object01.yaml
```diff
@@ -0,0 +1,68 @@
...
+  projectRef:
+    external: projects/${projectId}
+  resourceID: logginglogmetric-${uniqueId}
+  valueExtractor: EXTRACT(jsonPayload.response)
+status:
+  conditions:
+  - lastTransitionTime: "1970-01-01T00:00:00Z"
+    message: 'Update call failed: error applying desired state: infeasible update:
+      ({true <nil> MetricDescriptor.MetricKind}) would require recreation'
```
gemmahou, line 58: Same here, error message is slightly different between DCL and direct.
i.e. DCL: `Update call failed: error applying desired state: infeasible update:
      ({true <nil> MetricDescriptor.MetricKind}) would require recreation'`, direct: `Update call failed: logMetric update failed: cannot make changes to immutable field(s): MetricDescriptor.MetricKind`

pkg/controller/direct/logging/logmetric_controller.go
```diff
@@ -208,6 +208,9 @@ func (a *logMetricAdapter) Update(ctx context.Context, u *unstructured.Unstructu
 		}
 	}
 	if !compareMetricDescriptors(a.desired.Spec.MetricDescriptor, a.actual.MetricDescriptor) {
+		if err := validateImmutableFieldsUpdated(a.desired.Spec.MetricDescriptor, a.actual.MetricDescriptor); err != nil {
```
justinsb, line 211: We might not want to enforce this here (in the controller), I'm not sure vs just letting the GCP API call fail....  I'm guessing this is what our controllers do today?

pkg/controller/direct/logging/utils.go
```diff
@@ -76,6 +76,21 @@ func compareMetricDescriptors(kccObj *krm.LogmetricMetricDescriptor, apiObj *api
 	return reflect.DeepEqual(kccObj, convertAPItoKRM_MetricDescriptor(apiObj))
 }
 
+func validateImmutableFieldsUpdated(kccObj *krm.LogmetricMetricDescriptor, apiObj *api.MetricDescriptor) error {
+	actualMetricDescriptor := convertAPItoKRM_MetricDescriptor(apiObj)
+	modified := []string{}
+	if !reflect.DeepEqual(kccObj.MetricKind, actualMetricDescriptor.MetricKind) {
```
justinsb, line 82: I have a helper that might be useful - see #1770

pkg/webhook/immutable_fields_validator.go
justinsb: I believe so, yes

pkg/webhook/immutable_fields_validator.go
```diff
@@ -320,6 +332,45 @@ func validateImmutableFieldsForTFBasedResource(obj, oldObj *unstructured.Unstruc
...
+		}
+		if !ok1 || err1 != nil {
+			return true
+		}
+		if !ok2 || err2 != nil {
+			return true
+		}
+		return !reflect.DeepEqual(oldVal, newVal)
+	}
+	ImmutableFields := []string{}
```
justinsb, line 356: Nit: `immutableFields` because it isn't a global, and also probably simplest to do:

```
immutableFields := []string{
  "foo",
  "bar",
}
```

pkg/webhook/immutable_fields_validator.go
```diff
@@ -320,6 +332,45 @@ func validateImmutableFieldsForTFBasedResource(obj, oldObj *unstructured.Unstruc
...
 
+func validateImmutableFieldsForLoggingLogMetricResource(oldSpec, spec map[string]interface{}) admission.Response {
+	if isResourceIDModified(oldSpec, spec) {
+		return admission.Errored(http.StatusForbidden,
+			k8s.NewImmutableFieldsMutationError([]string{k8s.ResourceIDFieldPath}))
+	}
+
+	isImmutableFieldsModified := func(oldSpec, newSpec map[string]interface{}, field string) bool {
+		tokens := strings.Split(field, ".")
+		oldVal, ok1, err1 := unstructured.NestedFieldCopy(oldSpec, tokens...)
```
justinsb, line 343: This is nice (though it does break down with slices)

pkg/webhook/immutable_fields_validator.go
```diff
@@ -320,6 +332,45 @@ func validateImmutableFieldsForTFBasedResource(obj, oldObj *unstructured.Unstruc
...
+		}
+		if !ok2 || err2 != nil {
+			return true
+		}
+		return !reflect.DeepEqual(oldVal, newVal)
+	}
+	ImmutableFields := []string{}
+	ImmutableFields = append(ImmutableFields, "metricDescriptor.metricKind")
+	ImmutableFields = append(ImmutableFields, "metricDescriptor.valueType")
+	ImmutableFields = append(ImmutableFields, "projectRef")
```
justinsb, line 359: Ugh, projectRef is a tricky case.  As long as it's the existing behaviour, this seems reasonable.

justinsb:
This is looking good.  Some thoughts but I don't think anything blocking... thanks for getting this done in time - it's a big relief!

pkg/controller/direct/logging/logmetric_controller.go
gemmahou: Currently the operation fails at DCL level(i.e. DCL might checks for immutable fields too), and won't make API calls. Do we want to check the immutability earlier rather than making calls to GCP?

pkg/webhook/immutable_fields_validator.go
gemmahou: immutibility webhook for DCL checks projectRef via "hierarchicalRefs"([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/cd9aaedfd1bdd21bbaa0b85d90c93599d832e1a4/pkg/webhook/immutable_fields_validator.go#L144)).

pkg/controller/direct/logging/utils.go
gemmahou: Is that PR still WIP or ready? it looks good to me and we can get it merged.

pkg/controller/direct/logging/logmetric_controller.go
justinsb: I don't think it's _required_ that we check it here also, but if our controllers do it today then sgtm!

pkg/controller/direct/logging/utils.go
justinsb: Let's go with what you've got for 1.118 and then we can look more in 1.119

pkg/controller/direct/logging/utils.go
```diff
@@ -84,6 +84,22 @@ func compareMetricDescriptors(kccObj *krm.LogmetricMetricDescriptor, apiObj *api
...
 }
 
+func validateImmutableFieldsUpdated(kccObj *krm.LogmetricMetricDescriptor, apiObj *api.MetricDescriptor) error {
+	actualMetricDescriptor := convertAPItoKRM_MetricDescriptor(apiObj)
+
+	modified := []string{}
+	if !reflect.DeepEqual(kccObj.MetricKind, actualMetricDescriptor.MetricKind) {
+		modified = append(modified, "metricDescriptor.MetricKind")
+	}
+	if !reflect.DeepEqual(kccObj.ValueType, actualMetricDescriptor.ValueType) {
```
justinsb, line 94: I think we should guard against kccObj being nil or actualMetricDescriptor being nil.  Maybe we do punt this "double-check" to 1.119, IDK

tests/e2e/testdata/scenarios/direct/llm_immutable_field/labels/_object01.yaml
```diff
@@ -0,0 +1,62 @@
...
+    valueType: DISTRIBUTION
+  projectRef:
+    external: projects/${projectId}
+  resourceID: logginglogmetric-${uniqueId}
+  valueExtractor: EXTRACT(jsonPayload.response)
+status:
+  conditions:
+  - lastTransitionTime: "1970-01-01T00:00:00Z"
+    message: |-
+      Update call failed: error updating: logMetric update failed: googleapi: Error 400: Cannot change value type of label: testkey
```
justinsb, line 44: Does this mean that the immutable webhook was not catching this change before?

tests/e2e/testdata/scenarios/direct/llm_immutable_field/description/SCENARIO_README.MD
```diff
@@ -0,0 +1,3 @@
+Step1: Create a logginglogmetric resource, operation succeeds
+Step2: Update spec.description field for the resource, operation succeeds. We'll see PUT request with updated
```
justinsb, line 2: I don't think we have a scenario test where an update is blocked by the immutability webhook (unless I'm missing something)?

yuwenma: IIUC, the `description` is mutable, so this path is valid. Are you looking for the label field (immutable, fails without PUT request)  
https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1750/files#diff-168ebeca2eba2ea5385e8de5887d460b5bd7726f884323e116dc82ca4a43b113

yuwenma: oh, I see the problem. @gemmahou I think we want to add the immutability check for `metricDescriptor.description` (output only, immutable) rather than the `description` field.

gemmahou: `spec.description` is mutable, I have this test here to make sure mutable fields can be updated.

`spec.metricDescriptor.labels.description` field is immutable, however, with the current DCL controller, we do not have webhook or controller level checks for labels field, instead, GCP API will failed if any existing labels fields(description, key, valueType) are changed. That's why I have a separate test for labels: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/7164701044d5bd747540aed65c3603564031ed6a/tests/e2e/testdata/scenarios/direct/llm_immutable_field/labels/SCENARIO_README.MD

I guess the reason that we do not have checks on `spec.metricDescriptor.labels` field is, it's not allowed to modify existing labels, but it's allowed to add a new label with description, key and valueType(deletion is not allowed as well if i remember correctly). So It's a bit tricky for us to handle it in our webhook or controller.

tests/e2e/testdata/scenarios/direct/llm_immutable_field/labels/_object01.yaml
gemmahou: Yes, current webhook for DCL controller does not handle immutability on `spec.metricDescriptor.labels` fields(including `labels.description`,`labels. key`, `labels.valueType`). With DCL controller, we'll just let the GCP API to fail.

I also replied in this [comment](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1750#discussion_r1600925434).

pkg/controller/direct/logging/utils.go
gemmahou: I guess they can be nil? If both are nil then we consider it as "no changes on field".

tests/e2e/testdata/scenarios/direct/llm_immutable_field/description/SCENARIO_README.MD
yuwenma: Can we add another test to make sure updating the `spec.metricDescriptor.labels.description` will fail? 
We don't need specific test for `spec.description` .

yuwenma:
/lgtm
/approve

Thank you, @gemmahou !

gemmahou:
Discussed offline, after rebasing with #1772, the generated http_log in immutability scenarios tests need to be updated. To unblock to release, we decided to merge this PR and I will have a followup PR to fix the tests and probably the readme.

Regarding the spec.metricDescriptor.labels field, I create an [issue](https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/1805) to track it.

PR #1751, opened 2024-05-09
chore: create tool for crd manipulation

justinsb:
A bunch of highly-specialized commands to sync-up the CRD generation.

### DIFF
--- a/scripts/crd-tools/pkg/objectvisitor/objectvisitor.go
+++ b/scripts/crd-tools/pkg/objectvisitor/objectvisitor.go
@@ -1,0 +1,115 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


justinsb:
I think we want to create more special-purpose tools, but they should be reusable-ish, and driven by scripts.  Script is to-follow.

A lot of this is very hard-coded (e.g. set fields is only implemented for the one field we want to set), though we could enhance it relatively easily I think as we get more requirements.
 
/assign @acpana


cc @yuwenma

scripts/crd-tools/pkg/objectvisitor/objectvisitor.go
```diff
@@ -0,0 +1,115 @@
...
+			return err
+		}
+
+		if d.IsDir() {
+			return nil
+		}
+
+		if filepath.Ext(p) != ".yaml" {
+			return nil
+		}
```
acpana, line 46: this is a bikeshedding thought but we probably want to avoid files like `kustomize.yaml` (and all of its forms) if we use a mix of our own tooling and controller gen.

acpana:
This is great 💯 It'll help automate the CRD generation for direct resources 🥇 

/hold let author decide on merging sequence

/lgtm
/approve

justinsb:
If nobody objects, let's go for it, and then I can put the last pieces together!

/hold cancel

yuwenma (commented):
Thanks!

PR #1752, opened 2024-05-09
LoggingLogMetric: Use metav1.Time to avoid schema change

justinsb:
This allows us to match the existing CRD schema here.


justinsb:
/assign @acpana 

This is going to fix the format so that it is date-time in the CRD

acpana:
/hold I'll leave it up to you if we want to change/stage the new CRD in this PR too.

thanks for looking into this 💯 

/lgtm
/approve

justinsb:
>  I'll leave it up to you if we want to change/stage the new CRD in this PR too.

There's a few more changes, I'm trying to do it thematically.  This is the big one though.  One thing I was surprised by is that we aren't using the CRD directly ... should we be?

/hold cancel

PR #1754, opened 2024-05-09
Simple tooling to generate CRDs for LoggingLogMetric

justinsb:
- LoggingLogMetric: sync up schema labels & printColumns
- LoggingLogMetric: make spec required
- fix: deepcopy-gen annotations on shared package
- chore: create simple script to generate crds from direct apis
- autogen: generate crds using dev/tasks/generate-crds


config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_logginglogmetrics.logging.cnrm.cloud.google.com.yaml
```diff
@@ -71,7 +76,6 @@ spec:
                       bounds:
                         description: The values must be monotonically increasing.
                         items:
-                          format: double
```
acpana, line 74: IIRC, we were told this info is not exactly sent over to the api server but 🙈 we still want these, right?

scripts/generate-go-crd-clients/k8s/v1alpha1/register.go
```diff
@@ -30,7 +30,7 @@
 
 // Package v1alpha1 contains API Schema definitions for the k8s v1alpha1 API group
 // +k8s:openapi-gen=false
-// +k8s:deepcopy-gen=
+// +k8s:deepcopy-gen=package,register
```
yuwenma, line 33: Is this a manual change? I think we cannot override this and the other `register.go` file since they are autogenerated.

dev/tasks/generate-crds
```diff
@@ -0,0 +1,52 @@
...
+# Add OpenAPI validation to refs
+cd ${REPO_ROOT}
+go run ./scripts/add-validation-to-crds --dir apis/config/crd/
+
+# Apply a few small tweaks to the CRDs
+cd ${REPO_ROOT}
+go run ./scripts/crd-tools set-field spec.preserveUnknownFields=false --dir apis/config/crd/
+go run ./scripts/crd-tools delete-field status --dir apis/config/crd/
+go run ./scripts/crd-tools set-annotation cnrm.cloud.google.com/version=0.0.0-dev --dir apis/config/crd/
+go run ./scripts/crd-tools delete-annotation controller-gen.kubebuilder.io/version --dir apis/config/crd/
```
yuwenma, line 39: just curious, any of this can be achieved by kustomize?

justinsb: Maybe, but I think we'd have to use KRM functions (e.g. delete-annotation).  I figure we can start this way and then hopefully create KRM functions.

scripts/generate-go-crd-clients/k8s/v1alpha1/register.go
justinsb: I think this is the template and the other (pkg/clients/generated/apis/k8s/v1alpha1/register.go) is generated.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_logginglogmetrics.logging.cnrm.cloud.google.com.yaml
justinsb: Yes, we believe this is not actually used.  That we specify double (float64) but the official tools drop it supports this.  So I don't think we need it.

Maybe we should try to figure out how to add it just so we don't need to worry about it, but it would be a pain I think...

justinsb: Actually, scratch that, just found the `// +kubebuilder:validation:Format` directive

justinsb: Although this particular one is a problem because the format has moved to the slice, which is likely wrong

yuwenma: I like using kubebuilder to validate the CRD. Also, we shall have` // +kubebuilder:validation:Required ` on `spec.filter`.

One thing I don't figure out how it works is that the [LLM CRD has `filter` set as required](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/crds/logging_v1beta1_logginglogmetric.yaml#L389). But I was assuming the CRD is generated from this file which should have `filter`  as `optional` (default). Thoughts?

pkg/clients/generated/apis/logging/v1beta1/logginglogmetric_types.go
```diff
@@ -162,7 +162,7 @@ type LoggingLogMetricSpec struct {
 	bucket has to be in the same project as the metric. For
 	example:projects/my-project/locations/global/buckets/my-bucket
 	If empty, then the Log Metric is considered a non-Bucket Log Metric.
-	Only `external` field is supported to configure the reference. */
+	Only `external` field is supported to configure the reference for now. */
```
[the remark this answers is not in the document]
yuwenma: Actually, can we delete the TF-based APIs? Otherwise it could be really easy to get confused.

justinsb:
I split out the last commit as https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1843 but we can just merge this if you want

justinsb:
OK, #1843 merged, so let's see if github can figure out this rebase

/hold cancel

PR #1757, opened 2024-05-10
Resolve project references in LoggingLogMetric

justinsb:
- **chore: refactor direct controller interface to pass client.Reader**
- **feat: resolve project in direct controller for LoggingLogMetric**


acpana:
💯 this LGTM overall thanks for taking it on.

AFAICT., we don't have a test for this (fixtures/ samples, dynamic, or scenarios) but we can add it as a follow on.

/lgtm
/approve

PR #1759, opened 2024-05-10
chore: don't overwrite status conditions/observedGeneration in direct controller

justinsb:
This should avoid problems in particular with the conditions timestamps.


acpana (approved):
thanks for catching this 💯 

/lgtm
/approve

PR #1768, opened 2024-05-10
tests:LoggingLogMetric: make dynamic reconciler work for direct resources

acpana:
Patch tracks "teaching" the dynamic controller how to use direct controller and resources.  It also grew to include any LoggingLogMetric changes needed as a result of turning on the dynamic tests for direct resources.


acpana:
This latest patches have a green "postsubmit" run locally:

```bash
$ go test -v -tags=integration ./pkg/controller/dynamic/ -timeout 1600s -test.count=1 -test.run 'TestCreateNoChangeUpdateDelete/logging/basic-explicitlogmetric|TestCreateNoChangeUpdateDelete/logging/basic-exponentiallogmetric|TestCreateNoChangeUpdateDelete/logging/basic-linearlogmetric'
...
--- PASS: TestCreateNoChangeUpdateDelete (0.13s)
    --- PASS: TestCreateNoChangeUpdateDelete/logging (0.00s)
        --- PASS: TestCreateNoChangeUpdateDelete/logging/basic-exponentiallogmetric (78.57s)
        --- PASS: TestCreateNoChangeUpdateDelete/logging/basic-explicitlogmetric (78.68s)
        --- PASS: TestCreateNoChangeUpdateDelete/logging/basic-linearlogmetric (78.80s)
PASS
...
ok  	github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/dynamic	88.442s
```

pkg/controller/direct/logging/utils.go
```diff
@@ -130,7 +192,16 @@ func convertAPItoKRM_LogMetricLabels(apiLabels []*api.LabelDescriptor) []krm.Log
 			Key:         &apiLabel.Key,
 			ValueType:   &apiLabel.ValueType,
 		}
+		// TODO(acpana): this looks like a bug on the api, as the value should be defaulted already.
+		if *kccLabels[i].ValueType == "" {
```
justinsb, line 196: Nit: It's more robust to use ValueOf, I know you set it two lines above, but the problem is that when/if you refactor it's easy to introduce bugs that way.

pkg/controller/direct/logging/utils.go
```diff
@@ -130,7 +192,16 @@ func convertAPItoKRM_LogMetricLabels(apiLabels []*api.LabelDescriptor) []krm.Log
 			Key:         &apiLabel.Key,
 			ValueType:   &apiLabel.ValueType,
 		}
+		// TODO(acpana): this looks like a bug on the api, as the value should be defaulted already.
```
justinsb, line 195: I think I've seen this from the API, where `STRING` gets returned from the API as `""`.  I would call it a "quirk" of the API :-)  I bet if you look at the proto, it's an enum, and STRING == 0.

pkg/controller/direct/logging/utils.go
```diff
@@ -130,7 +192,16 @@ func convertAPItoKRM_LogMetricLabels(apiLabels []*api.LabelDescriptor) []krm.Log
...
+		if *kccLabels[i].ValueType == "" {
+			*kccLabels[i].ValueType = "STRING" // "" defaults to "STRING"
+		}
 	}
+
+	sort.Slice(kccLabels, func(i, j int) bool {
```
justinsb, line 201: I like this, and I've noticed the API returning in an order I can't discern.  The one gotcha is that an input KRM object does not round-trip, but I guess that's a function of the API.

pkg/controller/dynamic/dynamic_controller_integration_test.go
```diff
@@ -249,7 +251,10 @@ func validateCreate(ctx context.Context, t *testing.T, testContext testrunner.Te
 
 	// Check that an "Updating" event was recorded, indicating that the
 	// controller tried to update the resource at all.
-	testcontroller.AssertEventRecordedforUnstruct(t, kubeClient, reconciledUnstruct, k8s.Updating)
+	// TODO(acpana): figure out if we want to expose Updating event for direct resources
```
justinsb, line 254: Good call.  Can you open an issue for 1.119?  We should be creating the same events, and we should have test coverage.

pkg/controller/dynamic/dynamic_controller_integration_test.go
```diff
@@ -728,8 +735,12 @@ func assertObjectContains(t *testing.T, obj, changedFields map[string]interface{
 			}
 			assertObjectContains(t, objVal.(map[string]interface{}), changedVal.(map[string]interface{}))
 		default:
-			if !reflect.DeepEqual(objVal, changedVal) {
-				t.Fatalf("unexpected value for %v: got %v, want %v", changedKey, objVal, changedVal)
+			if diff := cmp.Diff(objVal, changedVal, cmpopts.SortSlices(
```
justinsb, line 738: This is because of the label randomization I take it?  I don't love the fact that it potentially covers up issues in other resources, but I don't have a great answer.

pkg/dcl/clientconfig/config.go
```diff
@@ -40,15 +40,15 @@ type Options struct {
 	controller.Config
 }
 
-func New(ctx context.Context, opt Options) (*dcl.Config, error) {
+func NewWithHttpClient(ctx context.Context, opt Options) (*dcl.Config, *http.Client, error) {
```
justinsb, line 43: Nit: NewConfigAndClient (as you did below) is probably a clearer name. I would expect `...WithHttpClient` to take the http.Client as input.

pkg/dcl/clientconfig/config.go
```diff
@@ -108,11 +122,11 @@ func NewForIntegrationTest() *dcl.Config {
 		opt.HTTPClient = &http.Client{Transport: t}
 	}
 
-	config, err := New(ctx, opt)
+	config, httpClient, err := NewWithHttpClient(ctx, opt)
```
justinsb, line 125: It looks like you don't really need `NewWithHttpClient`, because the only place it's called where we look for the http client return value is in this method, and opt.HTTPClient is used instead.  Maybe just don't export `NewWithHTTPClient` (i.e. rename to `newWithHTTPClient`?)

pkg/controller/direct/logging/logmetric_controller.go
```diff
@@ -51,6 +51,14 @@ func AddLogMetricController(mgr manager.Manager, config *controller.Config, opts
 	return directbase.Add(mgr, gvk, m, opts)
 }
 
+func GetModel(ctx context.Context, config *controller.Config) (directbase.Model, error) {
```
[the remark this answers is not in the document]
justinsb: (For 1.119 or later, though!)

justinsb:
A few nits but generally this LGTM and I think we should consider it for 1.118, I don't think it's very risky.

cc @yuwenma

acpana:
Yes, this is mostly tests and it makes our LLM related dynamic controller postsubmits pass 💯

yuwenma:
Thank you @acpana ! Could you fixing the nits, and I think we are ready to go for 1.118.

yuwenma:
/lgtm
/approve

Needs to move this to the v1.118 release. @acpana could you resolve the comments in a follow up PR?

pkg/controller/dynamic/dynamic_controller_integration_test.go
acpana: Yeah I figure anyplace where there are slices it would run into some non deterministic ordering.

acpana:
rebased and re-ran the tests locally ✅

PR #1771, opened 2024-05-10
feat: resolve bucket references in LoggingLogMetric

justinsb:


pkg/test/resourcefixture/testdata/basic/logging/v1beta1/logginglogmetric/logbucketmetric/create.yaml
```diff
@@ -48,4 +48,5 @@ spec:
   projectRef:
     external: "projects/${projectId}"
   loggingLogBucketRef:
-    external: "projects/${projectId}/locations/global/buckets/logginglogbucket-dep-${uniqueId}"
\ No newline at end of file
+    name: logginglogbucket-dep-${uniqueId}
+    kind: LoggingLogBucket
```
acpana, line 52: Do we want to keep a test with the external name only as well ?

PR #1772, opened 2024-05-10
feat: LoggingLogMetric - use updateTime to avoid spurious updates

justinsb:
generation/observedGeneration tells us if the KRM side has changed;
updateTime tells us if the GCP side has changed.

Note we still have a spurious update after the first creation because
of setting spec.resourceID


pkg/controller/direct/logging/logmetric_controller.go
```diff
@@ -228,66 +228,110 @@ func logMetricStatusToKRM(in *api.LogMetric, out *krm.LoggingLogMetricStatus) er
...
+	if obj.Status.UpdateTime == nil {
+		log.Info("status.updateTime is not set")
+		return true
+	}
+
+	objUpdateTimestamp := obj.Status.UpdateTime.Time
+	if objUpdateTimestamp == gcpUpdateTimestamp {
+		log.V(2).Info("status.updateTime matches gcp updateTime", "status.updateTime", ValueOf(obj.Status.UpdateTime), "gcpUpdateTime", gcpUpdateTime)
+		return false
+	}
```
acpana, line 329: It would be nice to see if we can see reduced requests in our http golden logs with a step scenario where we catch spurious updates.

More to the point, we should see how this does against our test for mutable but unreadable fields

pkg/controller/direct/logging/logmetric_controller.go
```diff
@@ -224,66 +222,122 @@ func logMetricStatusToKRM(in *api.LogMetric, out *krm.LoggingLogMetricStatus) er
...
-	if a.desired.Spec.Filter != a.actual.Filter {
-		// todo acpana: revisit UX, err out if filter of desired is empty
-		if a.desired.Spec.Filter != "" {
-			update.Filter = a.desired.Spec.Filter
-		} else {
-			// filter is a REQUIRED field
-			update.Filter = a.actual.Filter
+	latest := a.actual
+
+	if a.hasChanges(ctx, u) {
```
yuwenma, line 227: 👍

tests/e2e/script_test.go
```diff
@@ -130,6 +130,11 @@ func TestE2EScript(t *testing.T) {
 						exportResource = nil
 						shouldGetKubeObject = false
 
+					case "SLEEP":
```
yuwenma, line 133: Adding timer in test always worries me. Is the purpose of this SLEEP to guarantee single http request log? My concern is mostly about  flakiness and test running latencies. 
For this case, I think 2 seconds are fine, but maybe we shall have some constraints on adding timer in test. Any suggestions?

pkg/controller/direct/logging/logmetric_controller.go
```diff
@@ -224,66 +222,122 @@ func logMetricStatusToKRM(in *api.LogMetric, out *krm.LoggingLogMetricStatus) er
...
+		if ValueOf(a.desired.Spec.Description) != a.actual.Description {
+			update.Description = ValueOf(a.desired.Spec.Description)
+		}
+		if ValueOf(a.desired.Spec.Disabled) != a.actual.Disabled {
+			update.Disabled = ValueOf(a.desired.Spec.Disabled)
+		}
+		if a.desired.Spec.Filter != a.actual.Filter {
+			// todo acpana: revisit UX, err out if filter of desired is empty
+			if a.desired.Spec.Filter != "" {
+				update.Filter = a.desired.Spec.Filter
```
yuwenma, line 240: +1 to err out if the spec.filter is unset. We shall not allow unspecified required fields.

pkg/controller/direct/logging/logmetric_controller.go
```diff
@@ -224,66 +222,122 @@ func logMetricStatusToKRM(in *api.LogMetric, out *krm.LoggingLogMetricStatus) er
...
+			} else {
+				// filter is a REQUIRED field
+				update.Filter = a.actual.Filter
```
yuwenma, line 243: Just FYI, passing empty Filter to logging server is allowed. It will do a no-op.

pkg/controller/direct/logging/logmetric_controller.go
```diff
@@ -224,66 +222,122 @@ func logMetricStatusToKRM(in *api.LogMetric, out *krm.LoggingLogMetricStatus) er
...
 		}
 	}
 
 	return setStatus(u, status)
 }
 
+func (a *logMetricAdapter) hasChanges(ctx context.Context, u *unstructured.Unstructured) bool {
+	log := klog.FromContext(ctx)
+
+	if u.GetGeneration() != getObservedGeneration(u) {
```
yuwenma, line 293: Great!

pkg/controller/direct/logging/logmetric_controller.go
```diff
@@ -224,66 +222,122 @@ func logMetricStatusToKRM(in *api.LogMetric, out *krm.LoggingLogMetricStatus) er
...
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &obj); err != nil {
+		log.Error(err, "error converting from unstructured")
+		return true
+	}
+	if obj.Status.UpdateTime == nil {
+		log.Info("status.updateTime is not set")
+		return true
+	}
+
+	if !gcpUpdateTimestamp.Equal(obj.Status.UpdateTime) {
```
yuwenma, line 319: I think you want to mean equal here (remove `!`).

tests/e2e/testdata/scenarios/direct/llm_lazy_reconcile/_http01.log
acpana, line 1: This is what we wanted to see, right 💯 !

The `GET` comes from our `Find` but we find no extra PUT/ POST.

tests/e2e/testdata/scenarios/direct/llm_lazy_reconcile/script.yaml
```diff
@@ -0,0 +1,40 @@
...
+  valueExtractor: "EXTRACT(jsonPayload.response)"
+  bucketOptions:
+    explicitBuckets:
+      bounds:
+      - 2.5
+      - 5
+  projectRef:
+    external: "projects/${projectId}"
+---
+TEST: SLEEP
```
acpana, line 36: We could also wrangle the `WAIT-FOR-HTTP-LOG` stanza.

justinsb: I was thinking I might be able to replace this with setting a magic annotation.... might not be a 1.118 blocker though...

tests/e2e/script_test.go
justinsb: I think I can replace this with a magic annotation.  What I'm not sure about is whether this should be 1.118 or whether we want to live with the potentially flaky test and fix it in 1.119.  There will be code changes in the non-test path to detect the magic annotation.

pkg/controller/direct/logging/logmetric_controller.go
justinsb: I don't think I touched this code here other than indentation.  Maybe an AI for @acpana  ?

pkg/controller/direct/logging/logmetric_controller.go
justinsb: We need a test here IMO

tests/e2e/testdata/scenarios/direct/llm_lazy_reconcile/script.yaml
justinsb: (I think the problem with WAIT-FOR-HTTP-LOG is that we want to wait for a request _after_ the GET, which is another SLEEP I think.  We could maybe log the "closing" requests - there are additional requests during cleanup, but that's sort of a hack)

tests/e2e/script_test.go
yuwenma: I'll vote to leave non-test change to 1.119.

pkg/controller/direct/logging/logmetric_controller.go
acpana: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1798

I don't think it's a blocker for 1.118 😛  but I'll leave that up to y'all

pkg/controller/direct/logging/logmetric_controller.go
acpana: > Just FYI, passing empty Filter to logging server is allowed. It will do a no-op.

From a previous channel, while the current dcl controller allows to send an ""/ empty filter to the platform, it errors out  ...

justinsb:
@yuwenma / @acpana  shall we unhold to unblock 1.118?  I did the annotation trick in #1800, it's perhaps a little contentious (?) though it is much more predictable

yuwenma:
> @yuwenma / @acpana shall we unhold to unblock 1.118? I did the annotation trick in #1800, it's perhaps a little contentious (?) though it is much more predictable

I'll vote for yes! Let's merge this otherwise we will miss the 1.118 release.

acpana:
I'm game but I don't think the CI ran on the most recent version of this 🙃

justinsb:
Just repushed in the hope that triggers CI (I assume it was the problem that @acpana fixed earlier where I had corrupted our githubactions yaml)

acpana:
/lgtm
/approve

thanks for rebasing 🥇 

/hold cancel

PR #1773, opened 2024-05-10
perf:remove expander type enum

xiaoweim:
### Change description

Remove the expander type enum to support third party expanders.

### Tests you have done
 `make e2e-test`


experiments/compositions/composition/go.mod
```diff
@@ -1,12 +1,11 @@
 module google.com/composition
 
-go 1.22
+go 1.22.0
```
barney-s, line 3: what is this change for ?

PR #1776, opened 2024-05-11
Cleanup: Fix release manifest target

barney-s:
### Change description

Cleanup: Fix release manifest target.

This is to avoid changes in release files everytime some-one tests locally.

- Add a separate make release-manifests target for release to CC
- Rename existing release-* to release-test-* to indicate they are not official release targets
- Generate release-test* manifests in release/test which is .gitignore'd
- Change test files to use manifests from release/test/* folder
- Add separate kustomization target under config/release


experiments/compositions/composition/config/release/manager_auth_proxy_patch.yaml
```diff
@@ -0,0 +1,53 @@
...
+          limits:
+            cpu: 500m
+            memory: 512Mi
+          requests:
+            cpu: 5m
+            memory: 128Mi
+      - name: manager
+        args:
+        - "--health-probe-bind-address=:8081"
+        - "--metrics-bind-address=127.0.0.1:8080"
```
cheftako, line 52: Shouldn't the metrics port also be declared as a container port?

barney-s: This is a copy of the kubebuilder generated patch. I dont normally touch these files. Can check.

barney-s: https://book.kubebuilder.io/reference/metrics.html#protecting-the-metrics

barney-s: https://github.com/brancz/kube-rbac-proxy/#motivation

PR #1721, opened 2024-05-07
feat: use LoggingLogBucketRef

acpana:
Start using the LoggingLogBucketRef!

PR #1779, opened 2024-05-13
LoggingLogMetric: honor resourceID on creation

justinsb:
By passing the spec to KRMToAPI, we can defend against this.

### DIFF
--- a/pkg/controller/direct/logging/logmetric_controller.go
+++ b/pkg/controller/direct/logging/logmetric_controller.go
@@ -173,7 +173,8 @@ func (a *logMetricAdapter) Create(ctx context.Context, u *unstructured.Unstructu
 		}
 	}
 
-	logMetric := convertKCCtoAPI(a.desired)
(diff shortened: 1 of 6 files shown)


pkg/controller/direct/logging/logmetric_controller.go
```diff
@@ -173,7 +173,8 @@ func (a *logMetricAdapter) Create(ctx context.Context, u *unstructured.Unstructu
 		}
 	}
 
-	logMetric := convertKCCtoAPI(a.desired)
+	logMetric := convertKCCtoAPI(&a.desired.Spec)
+	logMetric.Name = a.resourceID
```
yuwenma, line 177: I like this. Shall this be our general solution for resourceID <-> resource name mapping? 
Also, I'm wondering what we have and have not determined in the previous discussions https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1721#discussion_r1594810095

PR #1781, opened 2024-05-13
mockgcp: Support for bigtable

justinsb:
This is tricky because our current is actually not using REST
(uniquely so!), so our normal interception method doesn't work.


mockgcp/mockbigtable/cluster.go
```diff
@@ -0,0 +1,255 @@
...
+	longrunning "google.golang.org/genproto/googleapis/longrunning"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/emptypb"
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/pkg/storage"
+	pb "google.golang.org/genproto/googleapis/bigtable/admin/v2"
+	// pb "github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/generated/mockgcp/bigtable/admin/v2"
```
yuwenma, line 32: Is this intended?  I think we want to mock the proto anyways to control the expectations.
Do you mind adding a comment about this change?

mockgcp/mockbigtable/cluster.go
```diff
@@ -0,0 +1,255 @@
...
+
+	obj := &pb.Cluster{}
+	if err := s.storage.Get(ctx, fqn, obj); err != nil {
+		return nil, err
+	}
+
+	return obj, nil
+}
+
+func (s *instanceAdminServer) ListClusters(ctx context.Context, req *pb.ListClustersRequest) (*pb.ListClustersResponse, error) {
```
yuwenma, line 51: I'm curious about this LIST call exploration. Is it for the re-reconciliation improvement or this is how the current TF-based controller uses? 
Maybe we want to have a GET method anyways. So we can add the direct controller now .

mockgcp/mockbigtable/instance.go
```diff
@@ -0,0 +1,250 @@
...
+		obj.Name = clusterFQN
+		if err := s.populateDefaultsForCluster(obj); err != nil {
+			return nil, err
+		}
+		if err := s.storage.Create(ctx, clusterFQN, obj); err != nil {
+			return nil, err
+		}
+	}
+
+	zone := "us-west1-b" // TODO
```
yuwenma, line 135: nit: "us-central1-a"? So the instance, cluster and table will be in the same zone.

mockgcp/tools/gapic/pkg/sorted/map.go
```diff
@@ -51,16 +51,6 @@ func (m *Map[K, V]) Entries() []Entry[K, V] {
 	return m.entries
 }
 
-func (m *Map[K, V]) Get(k K) V {
```
yuwenma, line 54: This function seems to be good, shall we keep it?

justinsb: It was duplicated from a few lines up.  I assume we fixed it in another PR.  I'll rebase!

mockgcp/mockbigtable/cluster.go
justinsb: The challenge is that Terraform uses GRPC today, and so if we change the name we might have problems.  I agree consistency would probably be better, so let me take a look at what we can do.

(The other option is to always use the "real" protos when we can - i.e. take this approach everywhere - the problem is that we might lose some control over the evolution of the proto schema.  But OTOH we aren't changing them anyway)

justinsb: So the issue is that we get "unknown service google.bigtable.admin.v2.BigtableInstanceAdmin", because we have renamed that service to `mockgcp.bigtable.admin.v2...` to avoid collisions.

Given this, I propose we go with what we have here.  If we can use the Protos as is, and only generate the GRPC gateway bindings, that _is_ easier (as we don't have to mess with the Protos).  It won't work for everything, but I proposed we use this one as an experiment for this simpler approach.

I'll add a comment!

mockgcp/mockbigtable/cluster.go
justinsb: This is just the mockgcp, so these are the calls that the terraform uses.  A bigtable Instance has a list of clusters in it: https://cloud.google.com/config-connector/docs/reference/resource-docs/bigtable/bigtableinstance

During reconciliation, terraform lists all the clusters and tries to match them.

mockgcp/mockbigtable/instance.go
justinsb: Good call.  It is odd that this is zonal TBH :-)  But we don't need to worry about it _too_ much for our mocks!

mockgcp/mockbigtable/cluster.go
yuwenma: I see. guess TF has some special reason to do that.

yuwenma (commented):
/lgtm

justinsb:
I think I've tackled the questions, so cancelling the hold

/hold cancel

PR #1782, opened 2024-05-13
chore: Don't require headers on generated testdata

justinsb:
This doesn't seem to add anything, and it causes a lot of friction.


Makefile
```diff
@@ -100,6 +100,7 @@ fmt:
 	-ignore "operator/config/gke-addon/image_configmap.yaml" \
 	-ignore "operator/config/rbac/cnrm_viewer_role.yaml" \
 	-ignore "operator/vendor/**" \
+	-ignore "**/testdata/**/_*" \
```
yuwenma, line 103: Sure! But why not?

PR #1789, opened 2024-05-14
Add xiaoweim as approver

barney-s:
### Change description
Adding xiaoweim as owner to help with reviews.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -13,6 +13,7 @@ approvers:
   - roycaihw
   - yuwenma
   - ziyue-101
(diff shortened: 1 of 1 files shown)


justinsb:
LGTM once xiaoweim is in the org:

/verify-owners

Also maybe we should find an org where this restriction doesn't exist, I'm not sure how hard it would be to add non-googlers as members here.

justinsb:
Awesome, welcome @xiaoweim :tada: 

/approve
/lgtm

PR #1793, opened 2024-05-14
chore: A few style nits in LoggingLogMetric

justinsb:
Removing a few redundant checks, but also rely on ValueOf when
converting *T to an API value.

### DIFF
--- a/pkg/controller/direct/logging/utils.go
+++ b/pkg/controller/direct/logging/utils.go
@@ -227,15 +227,15 @@ func convertKCCtoAPIForBucketOptions(kccObj *krm.LogmetricBucketOptions) *api.Bu
 	}
 	if kccObj.ExponentialBuckets != nil {
 		apiObj.ExponentialBuckets = &api.Exponential{}
-		apiObj.ExponentialBuckets.NumFiniteBuckets = int64(ValueOf(kccObj.ExponentialBuckets.NumFiniteBuckets))
-		apiObj.ExponentialBuckets.GrowthFactor = *kccObj.ExponentialBuckets.GrowthFactor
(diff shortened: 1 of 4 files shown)


pkg/controller/direct/logging/utils.go
```diff
@@ -317,15 +306,9 @@ func convertKCCtoAPIForLogMetricLabels(kccLabels []krm.LogmetricLabels) []*api.L
...
 
-		if kccLabel.Description != nil {
-			apiLabels[i].Description = *kccLabel.Description
-		}
-		if kccLabel.Key != nil {
-			apiLabels[i].Key = *kccLabel.Key
-		}
-		if kccLabel.ValueType != nil {
-			apiLabels[i].ValueType = *kccLabel.ValueType
-		}
```
acpana, line 328: are we losing the "unset" intent if we are doing this? What was previously "nothing"/ unset would now be ""

yuwenma: (Sorry I missed your comment before approve/lgtm)
I think this is intended: the KCC object should be declarative meaning if it is "unset", users expect the server-side value to be nil/unset.  

One potential problem though is the `spec.Filter` field. It is a required field and when updating on the server side, the server will ignore the "unset" `filter` but keeping the orignal value. For that case, we shall make `spec.filter` required field. (I think I raised this question in some other chats).

yuwenma: For the above potential problem, I think we are good as well (checked the CRD) https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/4c4bccd8d7390a4d217e451fbbe6ac5936d80fcb/crds/logging_v1beta1_logginglogmetric.yaml#L346

PR #1795, opened 2024-05-14
tests:chore: update for logbucketmetric

acpana:
Path adds update struct  for `logbucketmetric` test.

```bash
$ go test -v -tags=integration ./pkg/controller/dynamic/ -timeout 1600s -test.count=1 -test.run 'TestCreateNoChangeUpdateDelete/logging/basic-logbucketmetric'  2>&1
...
--- PASS: TestCreateNoChangeUpdateDelete (0.12s)
    --- PASS: TestCreateNoChangeUpdateDelete/logging (0.00s)
        --- PASS: TestCreateNoChangeUpdateDelete/logging/basic-logbucketmetric (79.51s)
PASS
...
ok      github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/dynamic      89.455s
```

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/logging/v1beta1/logginglogmetric/logbucketmetric/update.yaml
+++ b/pkg/test/resourcefixture/testdata/basic/logging/v1beta1/logginglogmetric/logbucketmetric/update.yaml
@@ -1,0 +1,56 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: logging.cnrm.cloud.google.com/v1beta1
+kind: LoggingLogMetric
+metadata:
+  name: logginglogmetric-${uniqueId}
+spec:
+  description: "An updated sample log metric"
+  filter: "resource.type=gae_app AND severity>=ERROR"
+  disabled: true
+  metricDescriptor:
+    labels:
+    - key: "mass"
+      valueType: "STRING"
+      description: "amount of matter"
+    - key: "hasMass"
+      valueType: "BOOL"
+      description: "whether the item has a mass"
+    - key: "sku"
+      valueType: "INT64"
+      description: "identifying number for item"
+    metricKind: "DELTA"
+    valueType: "DISTRIBUTION"
+    unit: "s"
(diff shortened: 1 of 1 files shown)


justinsb:
What's the base ref?

acpana:
[acpana/test-reconcile-direct](https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/acpana/test-reconcile-direct)

pkg/test/resourcefixture/contexts/logging_contexts.go
```diff
@@ -0,0 +1,22 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package contexts
+
+func init() {
+	resourceContextMap["logbucketmetric"] = ResourceContext{
+		ResourceKind: "LoggingLogBucket",
+		SkipUpdate:   true,
```
yuwenma, line 20: Does this test run against DCL or Terraform only? And I assume the DCL does not support LLM update?  If so, could you add a comment explaining why skipping the update.

yuwenma (commented):
/lgtm

One minor comment just for maintenance nit.

pkg/test/resourcefixture/contexts/logging_contexts.go
acpana: LLM  is now a direct resource 😛 . So this is neither DCL nor Terraform.  Looking at the `logbucketmetric` [scenario](https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/master/pkg/test/resourcefixture/testdata/basic/logging/v1beta1/logginglogmetric/logbucketmetric), I didn't include an `update.yaml` as I was  only looking to test the reference to the LogBucket.

yuwenma: IIUC, "skip update" should only depend on whether the resource support the feature: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/README.NewResourceFromTerraform.md#optionally-add-the-resource-to-test-contexts
"Note that disabling either of these tests should ONLY be done if the resource does not support the feature. Otherwise, you will need to debug the failures, understand the root cause, and make the appropriate fix(es)."

I was assuming you were trying to disable the DCL/TF part from the test. Could you elaborate a little more how the update should be skipped?

yuwenma:
/hold 

Need further discussion about the skip update

pkg/test/resourcefixture/contexts/logging_contexts.go
acpana: I had added the SkipUpdate because this test was added to check external references for the LoggingLogBucket. It sounds like the ask here is to add an `update.yaml` so I did 👍🏼 .

acpana:
Added update.yaml and re-ran the test ✅

yuwenma:
Regarding b/331421837, do we have any test to have the `LoggingLogMetric` using the new field `spec.logginglogbucketRef`? Maybe I miss it, but I don't see any examples under `resourcefixture/testdata`.

PR #1798, opened 2024-05-14
enh: shift left empty filter on update

acpana:
Err out if filter is empty on Update.


acpana:
cc @yuwenma / @justinsb I think this is not strictly needed for 1.118 but I'll leave it up to you two to decide.

PR #1799, opened 2024-05-14
chore:ci:fix: concurrency stanza

acpana:
Fixes upset presubmit I think.

PR #1800, opened 2024-05-14
tests: Introduce internal annotation to force reconciliation

justinsb:
The test.cnrm.cloud.google.com/reconcile-cookie annotation can be
changed without affecting the "true" desired state, but we special-case the annotation in our filtering, so that it still triggers the
reconciler.


acpana:
/hold 

CI needs #1799 to run or an early sha as base ref. Feel free to cancel as needed.

pkg/k8s/constants.go
```diff
@@ -113,6 +114,9 @@ var (
 		OrgIDAnnotation,
 	}
 
+	// Internal Annotation to force reconciliation
+	InternalForceReconcileAnnotation = CNRMInternalGroup + "/reconcile-cookie"
```
acpana, line 118: this may be overkill but do we want to prevent modifying this anon for non test usage?

acpana (commented):
Looked at the last commit and I like it! 💯 

Would you want to rebase/ discard the commits that have already been merged into main?

pkg/k8s/constants.go
```diff
@@ -26,6 +26,7 @@ type ManagementConflictPreventionPolicy string
 // TODO: clean up old conditions used in handcrafted controllers
 const (
 	CNRMGroup                            = "cnrm.cloud.google.com"
+	CNRMInternalGroup                    = "internal.cnrm.cloud.google.com"
```
acpana, line 29:
```suggestion
	CNRMInternalGroup                    = "test.cnrm.cloud.google.com"
```

?

yuwenma:
Thank you for the quick fix! (I hit the LLM lazy reconcile flakiness as well). And it is really good to learn the predicate usage with such an example 🌞

yuwenma:
/hold cancel

#1799 has been merged.

pkg/k8s/constants.go
justinsb: I like your change to make it `test`, it clearly signals that this should not be relied on in prod.  If we discover that people are (ab)using it, that's interesting info and we should probably implement a similar production-suitable feature (which might not actually look very different, but .... shrug)

PR #1801, opened 2024-05-14
chore: update presubmit build-images script to override image names

justinsb:
This avoids depending on the current gcloud project (which is likely
not set in github actions, at least).

### DIFF
--- a/dev/tasks/build-images
+++ b/dev/tasks/build-images
@@ -19,6 +19,17 @@ set -o pipefail
 
 REPO_ROOT="$(git rev-parse --show-toplevel)"
 
+# Force the image names, to avoid depending on the current gcloud project
+export SHORT_SHA=$(git rev-parse --short=7 HEAD)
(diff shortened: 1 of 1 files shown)


justinsb:
Let's sequence this one behind #1799 - I want to rebase on that anyway

/hold

dev/tasks/build-images
```diff
@@ -19,6 +19,17 @@ set -o pipefail
 
 REPO_ROOT="$(git rev-parse --show-toplevel)"
 
+# Force the image names, to avoid depending on the current gcloud project
```
acpana, line 22: if we accept this we probably can remove my hack to inject the `mockProject` id to the GH action steps 😛

acpana:
You don't have to rebase this PR for the CI to run since we are only changing a script file.

/hold

in case you want to remove the hack for env otherwise I can do it as a follow on maybe.

/lgtm
/approve

PR #1810, opened 2024-05-15
Add presubmits for compositions

barney-s:
### Change description

Add presubmits for compositions.
- Add github action targetting experiments/**
- Run tests for jinja expander
- Run e2e tests
- fix test data in compositions to trigger tests
- increase timeouts for reducing flakes

### DIFF
--- a/.github/workflows/presubmit-compositions.yaml
+++ b/.github/workflows/presubmit-compositions.yaml
@@ -1,4 +1,4 @@
-# Copyright 2023 Google LLC
+# Copyright 2024 Google LLC
 #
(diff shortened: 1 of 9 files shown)


.github/workflows/presubmit-compositions.yaml
```diff
@@ -0,0 +1,45 @@
+# Copyright 2023 Google LLC
```
acpana, line 1: this is a super nit and it doesn't matter because  we don't have a linter that catches these 😝 
```suggestion
# Copyright 2024 Google LLC
```

PR #1814, opened 2024-05-16
LoggingLogMetric: use string instead of metav1.Time

justinsb:
metav1.Time is overly strict about the input format, and we can't
parse values that were written with nanosecond precision.

### DIFF
--- a/apis/resources/logging/v1beta1/logmetric_types.go
+++ b/apis/resources/logging/v1beta1/logmetric_types.go
@@ -206,7 +206,7 @@ type LoggingLogMetricStatus struct {
 	Conditions []v1alpha1.Condition `json:"conditions,omitempty"`
 	/* Output only. The creation timestamp of the metric. This field may not be present for older metrics. */
 	// +optional
-	CreateTime *metav1.MicroTime `json:"createTime,omitempty"`
+	CreateTime *string `json:"createTime,omitempty"`
 
 	// +optional
 	MetricDescriptor *LogmetricMetricDescriptorStatus `json:"metricDescriptor,omitempty"`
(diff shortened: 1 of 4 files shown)


apis/resources/logging/v1beta1/logmetric_types.go
```diff
@@ -217,7 +217,7 @@ type LoggingLogMetricStatus struct {
 
 	/* Output only. The last update timestamp of the metric. This field may not be present for older metrics. */
 	// +optional
-	UpdateTime *metav1.MicroTime `json:"updateTime,omitempty"`
+	UpdateTime *string `json:"updateTime,omitempty"`
```
yuwenma, line 220: Yeah, we need to make this a convention.

pkg/controller/direct/logging/utils.go
```diff
@@ -40,6 +40,17 @@ func ValueOf[T any](p *T) T {
 	return v
 }
 
+// LazyPtr returns a pointer to v, unless it is the empty value, in which case it returns nil.
+// It is essentially the inverse of ValueOf, though it is lossy
+// because we can't tell nil and empty apart without a pointer.
+func LazyPtr[T comparable](v T) *T {
```
yuwenma, line 46: cool!

yuwenma (commented):
/lgtm
/approve

apis/resources/logging/v1beta1/logmetric_types.go
justinsb: Right, we should probably have a test - that would be much more reliable than a doc.  However, I don't know what the rule would be... if the CRD previously existed, we should not have a metav1.Time or metav1.MicroTime, I guess?

PR #1803, opened 2024-05-15
terraform: fix perm diff on edition field in SQL instance

jingyih:
### Change description

Pick up the fix in https://github.com/GoogleCloudPlatform/magic-modules/pull/9198

Fixes #1802

### Tests you have done

TESTED

1. Create a SQL instance with KCC 1.117.0
```
$ kubectl apply -f create.yaml

$ cat create.yaml 
apiVersion: sql.cnrm.cloud.google.com/v1beta1
kind: SQLInstance
metadata:
  name: sqlinstance-sample-333079656-2
spec:
  region: us-central1
  databaseVersion: MYSQL_5_7
  instanceType: CLOUD_SQL_INSTANCE
  settings:
    tier: db-n1-standard-1
```

2. Update the KCC code to include the patch in this PR. This mimics the case where the users upgrade the KCC to the version with this fix.

3. Setting the "settings.edition" field in the created SQL instance
```
$ kubectl apply -f update.yaml
sqlinstance.sql.cnrm.cloud.google.com/sqlinstance-sample-333079656-2 configured

jingyih@jingyih:~/Project/oncall/b333079656/basic$ cat update.yaml 
apiVersion: sql.cnrm.cloud.google.com/v1beta1
kind: SQLInstance
metadata:
  name: sqlinstance-sample-333079656-2
spec:
  region: us-central1
  databaseVersion: MYSQL_5_7
  instanceType: CLOUD_SQL_INSTANCE
  settings:
    tier: db-n1-standard-1
    edition: ENTERPRISE
``` 

It looks like the originally created SQL instance was defaulted to "ENTERPRISE" edition (according to Pantheon). KCC controller sees no diff when applying the "update.yaml".

4. No update calls in the subsequent KCC reconciliations.
```
{"severity":"info","timestamp":"2024-05-15T01:01:17.380Z","logger":"sqlinstance-controller","msg":"starting reconcile","resource":{"name":"sqlinstance-sample-333079656-2","namespace":"default"}}
{"severity":"info","timestamp":"2024-05-15T01:01:18.084Z","logger":"sqlinstance-controller","msg":"underlying resource already up to date","resource":{"name":"sqlinstance-sample-333079656-2","namespace":"default"}}
{"severity":"info","timestamp":"2024-05-15T01:01:18.085Z","logger":"sqlinstance-controller","msg":"successfully finished reconcile","resource":{"name":"sqlinstance-sample-333079656-2","namespace":"default"},"time to next reconciliation":"13m1.159273067s"}
[...]

Issue #1802, opened 2024-05-15
SQLInstance: Setting "settings.edition" may cause permanent diff


jingyih:
### Checklist

- [X] I did not find a related open issue.
- [X] I did not find a solution in the troubleshooting guide: (https://cloud.google.com/config-connector/docs/troubleshooting)
- [X] If this issue is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Bug Description

N.A

### Additional Diagnostic Information

N.A

### Kubernetes Cluster Version

1.27

### Config Connector Version

HEAD

### Config Connector Mode

namespaced mode (default)

### Steps to reproduce the issue

1. Create a SQL instance
```
$ kubectl apply -f create.yaml

$ cat create.yaml 
apiVersion: sql.cnrm.cloud.google.com/v1beta1
kind: SQLInstance
metadata:
  name: sqlinstance-sample-333079656-2
spec:
  region: us-central1
  databaseVersion: MYSQL_5_7
  instanceType: CLOUD_SQL_INSTANCE
  settings:
    tier: db-n1-standard-1
```

2. Setting the "settings.edition" field in the created SQL instance
```
$ kubectl apply -f update.yaml

jingyih@jingyih:~/Project/oncall/b333079656/basic$ cat update.yaml 
apiVersion: sql.cnrm.cloud.google.com/v1beta1
kind: SQLInstance
metadata:
  name: sqlinstance-sample-333079656-2
spec:
  region: us-central1
  databaseVersion: MYSQL_5_7
  instanceType: CLOUD_SQL_INSTANCE
  settings:
    tier: db-n1-standard-1
    edition: ENTERPRISE
``` 

3. SQL instance becomes "UpToDate" after applying the updated yaml.
```
...
Status:
  Conditions:
    Last Transition Time:  2024-05-14T22:34:30Z
    Message:               The resource is up to date
    Reason:                UpToDate
    Status:                True
    Type:                  Ready
```

4. However, KCC controller still tries to update the SQL instance during the subsequent reconciliations.
```
{"severity":"info","timestamp":"2024-05-14T22:34:19.164Z","logger":"sqlinstance-controller","msg":"starting reconcile","resource":{"name":"sqlinstance-sample-333079656-2","namespace":"default"}}
[...]

PR #1823, opened 2024-05-16
test: test NoChange after Update in dynamic test

jingyih:
Make a no-change update after the update test. Ensure there is no unexpected update call.

This test could help catch bugs like #1802.

Without the fix in #1803, the test fails with
```
go test -timeout 1200s -v -tags=integration ./pkg/controller/dynamic/ -test.run TestCreateNoChangeUpdateDelete -run-tests mysqlinstance
...
k8s.go:84: expected event with reason 'Updating' to not be recorded for SQLInstance epqxuxkxxi5dnnbghdaq/sqlinstance-sample-epqxuxkxxi5dnnbghdaq, but it was
...
--- FAIL: TestCreateNoChangeUpdateDelete (0.16s)
    --- FAIL: TestCreateNoChangeUpdateDelete/sql (0.00s)
        --- FAIL: TestCreateNoChangeUpdateDelete/sql/basic-mysqlinstance (337.22s)
...
```

With the fix in #1803, the test passes
```
go test -timeout 1200s -v -tags=integration ./pkg/controller/dynamic/ -test.run TestCreateNoChangeUpdateDelete -run-tests mysqlinstance
...
--- PASS: TestCreateNoChangeUpdateDelete (0.16s)
    --- PASS: TestCreateNoChangeUpdateDelete/sql (0.00s)
        --- PASS: TestCreateNoChangeUpdateDelete/sql/basic-mysqlinstance (395.45s)
```

### DIFF
--- a/pkg/controller/dynamic/dynamic_controller_integration_test.go
+++ b/pkg/controller/dynamic/dynamic_controller_integration_test.go
@@ -53,6 +53,7 @@ import (
 	transport_tpg "github.com/hashicorp/terraform-provider-google-beta/google-beta/transport"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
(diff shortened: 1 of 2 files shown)


maqiuyujoyce:
@jingyih some thoughts regarding this change:

I agree that adding a nochange step after the update will help identify #1802 sooner, though I feel it might be an one-off case that simply adding one step won't help at scale.

The root cause of the issue seems to be that after updating the value from `""` to `"ENTERPRISE"`, the returned value is still `""` from the GCP API - basically it is a bug from the API.
If we want to identify such an issue (rare and hard to identify; related to field defaulting and update) at scale, it means that in tests:
1. We need to have field coverage for any new field.
2. We need to at least have value coverage of the variants of the default value for any field. We probably also need to cover all the possible values for any enum field.
3. We need to have use-case coverage for transitions among the values for the any field, i.e. updating any field multiple times to different values, and always having a nochange step after an update.
 
What this PR covers is part of step 3 above, which probably will help achieving our eventual goal of test coverage, but I'm concerned that additional time will be added to the integration test suite that causes timeout in the test job. Have you got the chance to verify that?

Also, how often have we run into similar issues so far? (This is the first one afaik.) If adding this steps makes the test take longer to run, I'm not convinced we should support it for all the test cases at the moment unless it is a common issue.

jingyih:
@maqiuyujoyce Thank you for sharing your thoughts!

> What this PR covers is part of step 3 above, which probably will help achieving our eventual goal of test coverage, but I'm concerned that additional time will be added to the integration test suite that causes timeout in the test job. Have you got the chance to verify that?

My thinking was that adding one more verification (after the update test) could improve the test coverage in general. But I agree this is a tradeoff between: 1. how much improvement in test coverage this PR brings, and 2. how much additional time it requires. For 1, we can argue that we already have similar verification after the create test. So the test coverage improvement by this PR is only for special cases such as the "edition" field of `SQLInstance`. For 2, I could copy this PR to Gerrit and find out.

The original ask is that we want to add test coverage for #1802. Maybe I was looking at the wrong direction. Do you see an simpler way to test #1802?

jingyih:
Based on two test runs, adding additional test step in this PR is increasing the test run time from "48m53s" to "1h3m11s".

Alternatively, we could ONLY "opt-in" SQL instance for the additional test step. We would likely need a new flag `resourceContext.TestNoChangeAfterUpdate`. WDYT? @maqiuyujoyce

maqiuyujoyce:
> Based on two test runs, adding additional test step in this PR is increasing the test run time from "48m53s" to "1h3m11s".

I think what I'm concerned about is a different job. Will discuss offline about it.

> Alternatively, we could ONLY "opt-in" SQL instance for the additional test step. We would likely need a new flag resourceContext.TestNoChangeAfterUpdate. WDYT? @maqiuyujoyce

This works for me; I'm also okay with a local allowlist for it.

jingyih:
Added a local allowlist. The "NoChangeAfterUpdate" test is only enabled for `SQLInstance` resource.

yuwenma (commented):
/lgtm
/approve

I was originally thinking of adding MockGCP test with custom steps.  This looks good.

yuwenma:
Sorry I just checked the code and didn't notice your previous discussions @maqiuyujoyce @jingyih.

The testing I originally thought of is using MockGCP that you can mimic how the SQL server UPDATE behaves, and make it as likely as possible to the real SQL server.
Here's [an example](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/mockgcp/mockspanner/admin/instance.go#L81), the SQL server has an unpleasant behavior about its CREATE and UPDATE: `nodeCount` and `processingUnit`  is correlated (1 nodecount = 1000 processing unit), but only one value can be given on CREATE and UPDATE only accept number increasing.

PR #1828, opened 2024-05-16
refactor: EnsureFinalizers

acpana:
WIP

Do we want to have the lifecycle handler update the api server with the finalizers?


maqiuyujoyce:
_Copying my thoughts over here for posterity:_

I think whether we want to have the lifecycle handler update the api server with the finalizers is more of a design question: Should we add the finalizer to the resource asap during the reconciliation, or are we okay to wait?

To prevent the resource from being deleted unexpectedly, we should add the finalizer at the beginning of the reconciliation (so something similar to #1179). And for compatibility purpose, we should probably do it anyways.

We can also choose to not updating the apiserver until the reconciliation is completed if we have solid reasons why that's better - This is the current behavior in the direct controller though I'm actually a bit concerned about it.

PR #1832, opened 2024-05-17
feat: support export of LoggingLogMetric

justinsb:
Wiring up support for our direct controllers here.

### DIFF
--- a/pkg/controller/direct/registry.go
+++ b/pkg/controller/direct/registry.go
@@ -1,0 +1,87 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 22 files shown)


pkg/controller/direct/registry.go
```diff
@@ -0,0 +1,87 @@
...
+			}
+			found, err := a.Find(ctx)
+			if err != nil {
+				return nil, err
+			}
+			if !found {
+				return nil, fmt.Errorf("resource %q is not found", url)
+			}
+
+			u, err := a.Export(ctx)
```
acpana, line 78: Should I rename this `Export` to avoid any confusion in the future?

justinsb: I think this `Export` is well-named.  It also makes sense to have a package-level `Export` function, which delegates to a particular Adapter's `Export`

acpana:
/lgtm
/approve

thanks for this 🥇 

i'll rebase/ rework https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1691 to leverage this 💯

PR #1834, opened 2024-05-17
Integrate GRPC based expander into the composition engine

barney-s:
### Change description

1. Call out to GRPC service evaluate() from composition expander reconciler
2. Use grpc service address from expander version status
3. Add `TestSimpleCompositionExpansionGrpc` test case to test this flow
4.  Continue to use Plan object as intermediate.
5. Expose Evaluate() error (on failure) in the condition.


experiments/compositions/composition/tests/data/TestSimpleCompositionExpansionJob/output.yaml
```diff
@@ -0,0 +1,64 @@
...
+  - apiVersion: facade.foocorp.com/v1alpha1
+    blockOwnerDeletion: true
+    controller: true
+    kind: PConfig
+    name: team-a-config
+spec:
+  stages:
+    project:
+      manifest: |2
+
```
xiaoweim, line 52: are these two blank lines intentional?

experiments/compositions/composition/internal/controller/expanderversion_controller.go
```diff
@@ -132,7 +132,7 @@ func (r *ExpanderVersionReconciler) processExpanderVersion(
 			value = fmt.Sprintf("%s/expander-%s:%s", ev.Spec.ImageRegistry, expander, key)
 		} else {
 			svcVersion := strings.Replace(key, ".", "-", -1)
-			value = fmt.Sprintf("composition-%s-%s", expander, svcVersion)
+			value = fmt.Sprintf("composition-%s-%s:8443", expander, svcVersion)
```
xiaoweim, line 135: qq: why do we want 8443 here? what does it mean?

experiments/compositions/composition/config/expanders/jinja2-v0.0.1.yaml
```diff
@@ -101,7 +101,7 @@ metadata:
     app.kubernetes.io/part-of: composition
     app.kubernetes.io/instance: jinja2-v0.0.1
     app.kubernetes.io/component: expanders
-  name: jinja2-v0-0-1
+  name: gjinja2-v0-0-1
```
xiaoweim, line 104: qq: does this mean that our current jinja2 expander is based only on grpc so it is called gjinja2?

experiments/compositions/composition/internal/controller/expanderversion_controller.go
barney-s: Its the port where the service listens for connections. 443 is considered ROOT level port.

experiments/compositions/composition/config/expanders/jinja2-v0.0.1.yaml
barney-s: we need to differentiate b/w job based and grpc based by name/type. 
Will send a PR in future where this will become `jinja` and job based one would be renamed to something else like `pjinja`

experiments/compositions/composition/internal/controller/expanderversion_controller.go
xiaoweim: Thank you for the explanation!

PR #1836, opened 2024-05-18
Bug fix: fix the old string in the replacement to be the right format

ziyue-101:
### Change description

The old string to be matched should be `projects/${xxx}`

### Tests you have done


justinsb:
Good call, though I think this needs a rebase

/approve
/lgtm

(If it merges, great, as it should be fine after rebasing!)

PR #1844, opened 2024-05-20
mockGCP: mock containeranalysisnote

gemmahou:
### Change description

Add mockGCP for ContainerAnalysisNote

### Tests you have done


mockgcp/Makefile
```diff
@@ -74,10 +75,15 @@ generate-protos-from-openapi:
 	mkdir -p apis/mockgcp/cloud/bigquery/v2/
 	cd tools/gapic; go run . --proto-version=2 ../../bigquery-v2.json > ../../apis/mockgcp/cloud/bigquery/v2/api.proto
 
+	wget -O containeranalysis-v1.json https://raw.githubusercontent.com/googleapis/google-api-go-client/main/containeranalysis/v1/containeranalysis-api.json
```
justinsb, line 78: It looks like we have a proto version of this API.  The OpenAPI -> proto conversion is always trickier, so you could try using the proto version (i.e. just reference it from `./third_party/google/devtools/containeranalysis/v1`.  You likely need to add a replacement to `./fixup-third-party.sh` to rewrite `google/devtools` => `mockgcp/devtools`

justinsb: Actually, it wasn't _that_ proto (so I understand why!), but the proto looks to be available under grafeas

I made some progress here: https://github.com/GoogleCloudPlatform/k8s-config-connector/compare/master...justinsb:k8s-config-connector:mock-containeranalysisnote?expand=1

gemmahou: Thanks a lot!! I was using a wrong proto so that's why none of my CRUD methods are invoked. 

Do you have any suggestions on identifying the proto used for a resources? I searched for the service name in [googleapis](https://github.com/googleapis/googleapis) and I tried code search the method name in file type `proto`. I hope the service team have this information documented somewhere...

mockgcp/Makefile
```diff
@@ -28,7 +28,6 @@ gen-proto:
 		./apis/mockgcp/cloud/bigquery/v2/*.proto \
 		./apis/mockgcp/cloud/servicenetworking/v1/*.proto \
 		./apis/mockgcp/cloud/resourcemanager/v1/*.proto \
-		./apis/mockgcp/storage/v1/*.proto \
```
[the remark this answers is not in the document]
gemmahou: I removed this by mistake! reverted

log.txt
```diff
@@ -0,0 +1,107 @@
+=== RUN   TestAllInSeries
```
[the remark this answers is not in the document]
gemmahou: removed!

mockgcp/mockcontaineranalysis/note.go
```diff
@@ -0,0 +1,122 @@
...
+	// TODO: Some sort of helper for fieldmask?
+	for _, path := range paths {
+		switch path {
+		case "short_description":
+			obj.ShortDescription = req.Note.GetShortDescription()
+		case "long_description":
+			obj.LongDescription = req.Note.GetLongDescription()
+		case "related_url":
+			obj.RelatedUrl = req.Note.GetRelatedUrl()
+		case "attestation.hint.human_readable_name": // skip, immutable field should not be updated
```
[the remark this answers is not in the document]
gemmahou: I noticed that this field is immutable： https://source.corp.google.com/piper///depot/google3/third_party/grafeas/proto/v1/grafeas.proto;l=312?q=%22Required.%20Immutable.%20The%20type%20of%20analysis%20this%20note%20represents.%22&sq=package:piper%20file:%2F%2Fdepot%2Fgoogle3%20-file:google3%2Fexperimental, however, the API allows updates on the field. I'm going to contact the service team to understand more about this.

mockgcp/mockcontaineranalysis/names.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"strings"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/common/projects"
+)
+
+type NoteName struct {
```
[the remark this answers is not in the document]
gemmahou: Updated!

mockgcp/mockcontaineranalysis/note.go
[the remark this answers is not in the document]
gemmahou: Got response from b/342209649, changing the value of a type(attestation's human_readable_name) is allowed but changing to another type(attestation to vulnerability) is not. I updated the mock field mask.

pkg/test/resourcefixture/testdata/basic/containeranalysis/v1beta1/containeranalysisnote/_vcr_cassettes/dcl.yaml
```diff
@@ -685,7 +685,7 @@ interactions:
       host: containeranalysis.googleapis.com
       remote_addr: ""
       request_uri: ""
-      body: '{"attestation":{"hint":{"humanReadableName":"Attestor Note2"}},"longDescription":"new long description","relatedUrl":[{"label":"test","url":"some.url"}],"shortDescription":"short description"}'
+      body: '{"attestation":{"hint":{"humanReadableName":"Attestor Note1"}},"longDescription":"new long description","relatedUrl":[{"label":"test","url":"some.url"}],"shortDescription":"short description"}'
```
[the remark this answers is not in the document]
gemmahou: Reverted, per the feedback from service team in b/342209649, it's allowed to update this value.

mockgcp/mockcontaineranalysis/note.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func (s *ContainerAnalysisV1) CreateNote(ctx context.Context, req *pb.CreateNoteRequest) (*pb.Note, error) {
+	reqName := req.Parent + "/notes/" + req.NoteId
+	name, err := s.parseNoteName(reqName)
+	if err != nil {
+		return nil, err
+	}
+
+	fqn := name.String()
```
yuwenma, line 65: I think we want to assign the note fields according to https://cloud.google.com/artifact-analysis/docs/reference/rest/v1beta1/projects.notes#Note Specifically I don't see field like `sbomReference`  `vulnerabilityAssessment`, etc

gemmahou: I'm unsure if they are required. 

Checking the [http_log](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/9f66475fcaa28abd8bfceec1035e2533668601bf/pkg/test/resourcefixture/testdata/basic/containeranalysis/v1beta1/containeranalysisnote/_http.log), fields are assigned according to the [test yaml](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/test/resourcefixture/testdata/basic/containeranalysis/v1beta1/containeranalysisnote/create.yaml). `sbomReference` and `vulnerabilityAssessment` are both options of union type, per the REST API doc, `// Union field type can be only one of the following`, once `attestation` is set, we may not need other types. 

However, I checked other mock resources, missing some field values might cause mock test to fail. So I'm unsure if we have an "official" way to determine the necessarily of the manual assigned values.

mockgcp/bigquery-v2.json
yuwenma, line 1: Can we group these json file and put them under `mockgcp/mockcontaineranalysis`?

mockgcp/mockcontaineranalysis/note.go
yuwenma: I didn't make it clear. 
1. I think we want to assign the output only fields here.  like "kind"
2. Regarding `vulnerabilityAssessment` and `sbomRefernece`, I saw they are one of "Union field type can be only one of the following". I want to know what error CREATE will give if specify two, and what message the UPDATE will give if we change the one of from sbomRefrence to valnerabilityAssessment, etc.  These can be done in follow up PRs.

pkg/test/resourcefixture/testdata/basic/containeranalysis/v1beta1/containeranalysisnote/_http.log
yuwenma, line 1: Could you also include the golden log for the `object` (`GOLDEN_OBJECT_CHECKS`)? (or maybe you did already)

mockgcp/generated/mockgrafeas/v1/attestation.pb.go
yuwenma, line 1: nit: Shall we put these under third_party/googlepais/mockcontaineranalysisnote/ ? What I"m thinking is for future maintenance, under mockgrafeas could make it harder to discover and reuse. But this is just a nit

mockgcp/mockcontaineranalysis/note.go
gemmahou: Make sense! Done.

pkg/test/resourcefixture/testdata/basic/containeranalysis/v1beta1/containeranalysisnote/_http.log
gemmahou: Added!

yuwenma (commented):
/lgtm
/approve

Thanks!

PR #1850, opened 2024-05-20
Resolve projectRef for the FM direct controller

ziyue-101:
### Change description

Resolve project Ref for hub controller
### Tests you have done

### DIFF
--- a/pkg/controller/direct/utils/projectref.go
+++ b/pkg/controller/direct/utils/projectref.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package logging
+package utils
 
 import (
 	"context"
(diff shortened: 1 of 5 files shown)


pkg/controller/direct/utils/projectref.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package logging
+package utils
```
acpana, line 15: this may be a bit if bikeshedding but could we rename this pkg to `references` ?

acpana (commented):
/hold

one bikeshed-y comment, can be overridden.

/lgtm
/approve

pkg/controller/direct/utils/projectref.go
ziyue-101: Done!

yuwenma: I share the same bikeshedding as @acpana 🤣

yuwenma:
/lgtm
/approve

Thanks.

PR #1852, opened 2024-05-20
Add structured config support for Expander

barney-s:
### Change description

Add ExpanderConfig to Expander
- Allows structured config for inbuilt expanders
- The ability to support refs for external configs
- remove proto annotations (cleanup)

### DIFF
--- a/experiments/compositions/composition/api/v1alpha1/composition_types.go
+++ b/experiments/compositions/composition/api/v1alpha1/composition_types.go
@@ -23,8 +23,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// TODO(barney-s) : Remove proto annotations
-
 // ConditionType defines the type of ManagedConfigSync condition
 type ConditionType string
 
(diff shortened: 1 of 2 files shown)


experiments/compositions/composition/api/v1alpha1/composition_types.go
```diff
@@ -45,57 +43,67 @@ type ResourceRef struct {
 	// resource: ServiceIdentity.serviceusage.cnrm.cloud.google.com/v1beta1//sqladmin.googleapis.com
 
 	// OPTION 2
-	Group      string `json:"group,omitempty" protobuf:"bytes,1,name=group"`
-	Version    string `json:"version,omitempty" protobuf:"bytes,2,opt,name=version"`
-	Kind       string `json:"kind" protobuf:"bytes,3,name=kind"`
-	Name       string `json:"name,omitempty" protobuf:"bytes,4,name=name"`
-	NameSuffix string `json:"nameSuffix,omitempty" protobuf:"bytes,2,name=nameSuffix"`
+	Group      string `json:"group,omitempty"`
```
cheftako, line 46: Do we still have/want an option 1?

experiments/compositions/composition/api/v1alpha1/composition_types.go
```diff
@@ -45,57 +43,67 @@ type ResourceRef struct {
...
 type ValuesFrom struct {
-	Name        string      `json:"name" protobuf:"bytes,1,name=name"`
-	ResourceRef ResourceRef `json:"resourceRef" protobuf:"resourceRef,2,name=resourceref"`
-	FieldRef    []FieldRef  `json:"fieldRef" protobuf:"fieldRef,3,name=fieldref"`
+	Name        string      `json:"name"`
+	ResourceRef ResourceRef `json:"resourceRef"`
+	FieldRef    []FieldRef  `json:"fieldRef"`
+}
+
+type Jinja2 struct {
```
cheftako, line 64: Shouldn't this be part of the Jinja expander rather than the composition?

experiments/compositions/composition/api/v1alpha1/composition_types.go
```diff
@@ -45,57 +43,67 @@ type ResourceRef struct {
...
+	Template string `json:"template"`
+}
+
+// ConfigReference - For BYO Expanders, we can extend it
+type ConfigReference struct {
+	APIGroup  string `json:"APIGroup"`
+	Name      string `json:"name"`
+	Namespace string `json:"namespace,omitempty"`
+}
+type ExpanderConfig struct {
```
cheftako, line 74: Missing empty line

experiments/compositions/composition/api/v1alpha1/composition_types.go
barney-s: Jinja Expander has no CRD for itself. ExpanderVersion is not the right place. 
Iam expecting few common built-in expanders (core expanders) that are part of the system. 

Along those lines jinja2 is part of the built-in for now.

experiments/compositions/composition/api/v1alpha1/composition_types.go
barney-s: fixed and added comments.

experiments/compositions/composition/api/v1alpha1/composition_types.go
xiaoweim: so does that mean that we will also have a struct for the second expander that we are adding?

barney-s: thats correct. 
Added a comment to that effect: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/c0b57d8e17aa41097a1a892019b0b3371e958b97/experiments/compositions/composition/api/v1alpha1/composition_types.go#L79

cheftako: Seems ok for now but we need to think through the adding a new expander CUJ. I think the best way to do that it actually be uniform in how we add expanders. Toward that end I think we should have the expanders configuration CRD as another part of the expander package along with relevant roles, permissions etc.

experiments/compositions/composition/api/v1alpha1/composition_types.go
```diff
@@ -45,57 +43,71 @@ type ResourceRef struct {
...
+	APIGroup  string `json:"APIGroup"`
+	Name      string `json:"name"`
+	Namespace string `json:"namespace,omitempty"`
+}
+
+type ExpanderConfig struct {
+	// Built in expanders
+	Jinja2 *Jinja2 `json:"jinja2,omitempty"`
+	// Soon..
+	// Getter *Getter `json:"getter,omitempty"`
```
cheftako, line 79: When we get to the "Getter" I would like us to take a look at how to make this more generic. It seems like this should be a reference to a config object appropriate to the expander and we should have 1 such field for all the expander types rather than for each expander type.

experiments/compositions/composition/api/v1alpha1/composition_types.go
```diff
@@ -45,57 +43,71 @@ type ResourceRef struct {
...
-	// NOTE: Tighten the Composition API to include fields that are used in the controller
-	//  As we add features we can uncomment these fields
+	Version    string       `json:"version,omitempty"`
+	ValuesFrom []ValuesFrom `json:"valuesFrom,omitempty"`
 
-	//ConfigAPIGroup  string `json:"configAPIGroup,omitempty" protobuf:"bytes,3,opt,name=configAPIGroup"`
-	//ConfigName      string `json:"configName,omitempty" protobuf:"bytes,4,opt,name=configName"`
-	//ConfigNamespace string `json:"configNamespace,omitempty" protobuf:"bytes,5,opt,name=configNamespace"`
-	//Image string `json:"image" protobuf:"bytes,6,opt,name=image"`
+	ExpanderConfig `json:""`
```
cheftako, line 97: We should not need ExpanderConfig and Jinja2/Getter configs.

cheftako: I think this is the correct direction but we should probably at least have a direction comment/todo.

experiments/compositions/composition/api/v1alpha1/composition_types.go
barney-s: The disadvantage with this approach is the composition is not inline. Meaning when defining a composition we would need to define a few expander objects and reference it in the composition. Is this the experience we want for built in expanders as well ?

experiments/compositions/composition/api/v1alpha1/composition_types.go
barney-s: I see your intent. 
At the same-time having inline-expander config has simpler UX.

Wondering whats the best way. May be i can add a third expander that is properly offline.

experiments/compositions/composition/api/v1alpha1/composition_types.go
barney-s: I was thinking a few inbuilt (inline) expanders and support for external expander configs (BYO)

experiments/compositions/composition/api/v1alpha1/composition_types.go
barney-s: Adding a comment and a TODO to not have inline expanders long run

experiments/compositions/composition/api/v1alpha1/composition_types.go
barney-s: Adding a comment and a TODO to not have inline expanders long run

experiments/compositions/composition/api/v1alpha1/composition_types.go
barney-s: Adding a comment and a TODO to not have inline expanders long run

PR #1861, opened 2024-05-21
tests: simplify monitoringdashboard test case, add to tests-e2e-direct

justinsb:
Changing to a random display name is not very intuitive (and doesn't
get us much).
    
Also add this to tests-e2e-direct, so we verify the golden output

### DIFF
--- a/scripts/github-actions/tests-e2e-direct.sh
+++ b/scripts/github-actions/tests-e2e-direct.sh
@@ -38,7 +38,7 @@ echo "Running e2e tests fixtures for LoggingLogMetric direct reconciliation..."
 GOLDEN_OBJECT_CHECKS=1 \
 GOLDEN_REQUEST_CHECKS=1 \
 E2E_KUBE_TARGET=envtest RUN_E2E=1 E2E_GCP_TARGET=mock \
-  go test -test.count=1 -timeout 600s -v ./tests/e2e -run 'TestAllInSeries/fixtures/explicitlogmetric|TestAllInSeries/fixtures/exponentiallogmetric|TestAllInSeries/fixtures/linearlogmetric|TestAllInSeries/fixtures/logbucketmetric'
+  go test -test.count=1 -timeout 600s -v ./tests/e2e -run 'TestAllInSeries/fixtures/explicitlogmetric|TestAllInSeries/fixtures/exponentiallogmetric|TestAllInSeries/fixtures/linearlogmetric|TestAllInSeries/fixtures/logbucketmetric|TestAllInSeries/fixtures/monitoringdashboard'
 
 echo "Running scenarios tests for LoggingLogMetric direct reconciliation..."
 
--- a/pkg/test/resourcefixture/testdata/basic/monitoring/v1beta1/monitoringdashboard/monitoringdashboardrefs/_generated_object_monitoringdashboardrefs.golden.yaml
+++ b/pkg/test/resourcefixture/testdata/basic/monitoring/v1beta1/monitoringdashboard/monitoringdashboardrefs/_generated_object_monitoringdashboardrefs.golden.yaml
@@ -52,7 +52,7 @@ spec:
           filter: metric.type="agent.googleapis.com/nginx/connections/accepted_count"
           resourceNames: []
         title: Widget 4
-  displayName: monitoringdashboard-${uniqueId}
+  displayName: monitoringdashboard updated
   projectRef:
     name: otherproject
   resourceID: monitoringdashboard-${uniqueId}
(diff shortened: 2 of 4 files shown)


scripts/github-actions/tests-e2e-direct.sh
```diff
@@ -38,7 +38,7 @@ KCC_USE_DIRECT_RECONCILERS=LoggingLogMetric \
 GOLDEN_OBJECT_CHECKS=1 \
 GOLDEN_REQUEST_CHECKS=1 \
 E2E_KUBE_TARGET=envtest RUN_E2E=1 E2E_GCP_TARGET=mock \
-  go test -test.count=1 -timeout 600s -v ./tests/e2e -run 'TestAllInSeries/fixtures/explicitlogmetric|TestAllInSeries/fixtures/exponentiallogmetric|TestAllInSeries/fixtures/linearlogmetric'
+  go test -test.count=1 -timeout 600s -v ./tests/e2e -run 'TestAllInSeries/fixtures/explicitlogmetric|TestAllInSeries/fixtures/exponentiallogmetric|TestAllInSeries/fixtures/linearlogmetric|TestAllInSeries/fixtures/monitoringdashboard'
```
acpana, line 41: two quick points:

- this is not using the direct controller for monitoring yet
- we probably want to ignore the `s-i-s: merge`

I don't think we can do the same trick as the `script.yaml` annotations for `absent` since these are fixtures but will leave that up to you if you want to modify the yaml or add a code check (specifically for `monitoring` I think) for now.

( these are probably all items I could've done better 😛 )

yuwenma: Can you share more context about the "script.yaml and the trick" (just in case others may hit a problem)?
 
I thought this go test  is a short path for LLM only: the `KCC_USE_DIRECT_RECONCILERS` specifies to LLM and the run only lists LLM related tests.
If we want to run tests for different scifi controllers, maybe we shall separate it from here?

acpana: > I don't think we can do the same trick as the script.yaml annotations for absent since these are fixtures

Oh I see, I think this comment was a bit confusing and I should've linked the code I had in mind 🤦🏼 Here's an example of using `state-into-spec: absent` in`script.yaml`:

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/86a55f99a46e5fb5a1072d6e9fa819edcfb03fb2/tests/e2e/testdata/scenarios/direct/llm_immutable_field/labels_value_type/script.yaml#L15-L20

But again, I don't think we can use that here because the `fixture` tests are shared in `test-e2e-fixtures` today, and more importantly, in post submits.

> I thought this go test is a short path for LLM only: the KCC_USE_DIRECT_RECONCILERS specifies to LLM and the run only lists LLM related tests.

Yes I meant to highlight that in with this point:

> * this is not using the direct controller for monitoring yet

So today, adding that would be a "no-op" and yes we would require a `KCC_USE_DIRECT_RECONCILERS` flag when we actually want to use the direct controller. Moreover, I'd recommend we split out the LLM and other resources 😛

justinsb:
Actually looks like I did most of this better in #1806, so let's get that one in first, then I'll rebase

/hold

scripts/github-actions/tests-e2e-direct.sh
```diff
@@ -38,7 +38,7 @@ KCC_USE_DIRECT_RECONCILERS=LoggingLogMetric \
 GOLDEN_OBJECT_CHECKS=1 \
```
yuwenma, line 38: Just to double check, do we want to run monitoringdashboard direct controller on the test "monitoringdashboard"?  If so, I think we want to include the Kind name in KCC_USE_DIRECT_RECONCILERS or using a different command (so you can  echo the steps more granularly)

justinsb: So I wanted to start by adding this to the "strict" tests where we check the golden output.  _Then_ I would add it to KCC_USE_DIRECT_RECONCILERS, and we would be able to see the changes in behaviour (if any) at the http and at the KRM level. 

I think what's going wrong is that we've turned off golden tests with KCC_USE_DIRECT_RECONCILERS, and I don't think we should have done that.

yuwenma:
PR looks good. could you capture the latest log and I think this PR is ready to go!

justinsb:
I think the flakes are due to races in the SLEEP approach, which was always intended as a temporary hack to be fixed in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1800. I updated that PR.

scripts/github-actions/tests-e2e-direct.sh
yuwenma: Sounds good. With #1910 , those LLM tests should no longer need the  KCC_USE_DIRECT_RECONCILERS flag (and thanks for the cleanup in #1919). 

When the monitoring dashboard is ready for the direct controller, you can simply remove the [dcl2crd label](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/b0451688317d853117d0f93e324f0d6cb000558c/apis/monitoring/v1beta1/monitoringdashboard_types.go#L870)  (KCC_USE_DIRECT_RECONCILERS should also work but I think it less preferred for beta resource)

PR #1613, opened 2024-04-19
chore: Tool to generate mapping functions from proto

justinsb:
A first-pass at a tool to generate types & map functions.

PR #1862, opened 2024-05-21
chore: create generated mappers and roundtrip test for Dashboard

justinsb:
This allows us to see unsupported fields.


pkg/controller/direct/monitoring/maputils.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	return outSlice
+}
+
+type ProtoEnum interface {
+	~int32
+	Descriptor() protoreflect.EnumDescriptor
+}
+
+func Enum_ToProto[U ProtoEnum](ctx *MapContext, in *string) U {
```
acpana, line 57: big nit: Going to leave it up to you but could we rename `ctx` to `mapCtx` or something that doesn't make me think of the `context.Context` one 😛

pkg/controller/direct/monitoring/maputils.go
```diff
@@ -0,0 +1,179 @@
...
+	var defaultU U
+	descriptor := defaultU.Descriptor()
+
+	inValue := ValueOf(in)
+	if inValue == "" {
+		unspecifiedValue := U(int32(0)) // defaultU.New(protoreflect.EnumNumber(0)).(U)
```
acpana, line 63: for my own learning, here and everywhere else, why not int64?

pkg/controller/direct/monitoring/maputils.go
```diff
@@ -0,0 +1,179 @@
...
+	ctx.Errorf("unknown enum value %q for %v (valid values are %v)", inValue, descriptor.FullName(), strings.Join(validValues, ", "))
+	return 0
+}
+
+func Enum_FromProto[U ProtoEnum](ctx *MapContext, v U) *string {
+	// if in == nil {
+	// 	return nil
+	// }
+
+	// v := *in
```
acpana, line 91:
```suggestion
```

remove?

pkg/controller/direct/monitoring/roundtrip_test.go
```diff
@@ -0,0 +1,388 @@
...
+
+func fillWithRandom0(t *testing.T, rand *rand.Rand, msg protoreflect.Message) {
+	descriptor := msg.Descriptor()
+	if string(descriptor.FullName()) == "google.protobuf.Duration" {
+		count := rand.Intn(10)
+		// Bias to zero
+		if count > 4 {
+			return
+		}
+		seconds := rand.Intn(365 * 24 * 60 * 60)
```
acpana, line 133: do we need to seed `rand` for determinism ?

pkg/controller/direct/monitoring/roundtrip_test.go
acpana, line 1: This is 💯 ; I'm wondering 1) if we want this to be in the `direct` package or test package long term and 2) if we can iterate over al the resources we care about with the same core.

pkg/controller/direct/monitoring/dashboard_generated.mappings.go
acpana, line 1: Wait how do we generate this?

pkg/controller/direct/monitoring/maputils.go
acpana, line 1: There's nothing "monitoring" about this right? Trying to be careful about creating yet another magic machine (YAMM) but do we want this to be used across `direct` package?

dev/tasks/generate-crds
```diff
@@ -24,7 +24,7 @@ cd ${REPO_ROOT}
 
 # Run controller-gen to generate CRDs
 cd ${REPO_ROOT}/apis
-go run sigs.k8s.io/controller-tools/cmd/controller-gen@v0.14.0 object crd:crdVersions=v1,allowDangerousTypes=true output:crd:artifacts:config=config/crd/ paths="./..."
+go run sigs.k8s.io/controller-tools/cmd/controller-gen@master object crd:crdVersions=v1,allowDangerousTypes=true output:crd:artifacts:config=config/crd/ paths="./..."
```
acpana, line 27: Do we want to keep this pinned ?

acpana (commented):
Some questions but overall looks really cool, especially the fuzzer

pkg/controller/direct/monitoring/maputils.go
justinsb: A much better name, thank you!

justinsb:
I've realized this is based on #1857

pkg/controller/direct/monitoring/maputils.go
justinsb: Hmmm, good question.  It shouldn't matter in go, because constant values are untyped (or something).  Removed the int32 entirely, and it seems to still work :-)

justinsb: (proto.EnumNumber is int32, so it is actually int32 at the proto level, though proto int64 vs int32 is largely academic anyway as they both use variable length encoding, I believe)

pkg/controller/direct/monitoring/maputils.go
justinsb: Thanks, done!

pkg/controller/direct/monitoring/roundtrip_test.go
justinsb: Will rename to randStream.  And we do, the first "real" line of FuzzFromProto

pkg/controller/direct/monitoring/roundtrip_test.go
justinsb: Agree.  I am just starting it here, but I think we want to move it out pretty fast into a shared library!

pkg/controller/direct/monitoring/dashboard_generated.mappings.go
justinsb: Using https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1613 ... the gotcha is that I don't think we are all agreed on whether we _should_ generate it this way.  I hope to achieve that consensus when we all meet and reconcile our approaches.

If you want to use the tool, feel free though!  I think we're coming up with a way of doing it in this stream of PRs:

1. Get mockgcp running to understand the API and make testing easier
1. Add a few more tests and capture some golden output
1. Add to the direct test script
1. Use the generated types and make sure that the CRD doesn't change "too much" (massive pain: the description strings are reformatted, making the diff painful)
1. Add the generated mapper functions and fuzz test to make sure we have coverage.
1. Write the "core controller" using those generated types & mapper functions
1. At some point switch the direct reconciler to be "always on"

It's a bit fuzzy right now, but this seems to be what's evolving.

dev/tasks/generate-crds
justinsb: Yes, but we need a PR.  I sent https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1868 to add a TODO

justinsb:
#1857 has now merged, so rebasing.  Thanks for the review @acpana!

pkg/controller/direct/monitoring/dashboard_generated.mappings.go
acpana: Thanks for this overview! Let's give this a go! 💯

acpana:
thanks for addressing comments 🥇 

/lgtm
/approve

pkg/controller/direct/monitoring/roundtrip_test.go
```diff
@@ -0,0 +1,419 @@
...
+	"github.com/google/go-cmp/cmp"
+	"google.golang.org/protobuf/encoding/prototext"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/testing/protocmp"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
+)
+
+// IDEA: Load all the samples, and check that we have all the KRM paths covered
```
yuwenma, line 34: +100. Love this direction.

yuwenma (commented):
/lgtm
/approve

PR #1863, opened 2024-05-21
feat: implement direct controller for MonitoringDashboard

justinsb:
- **feat: implement direct controller for MonitoringDashboard**
- **chore: update monitoringdashboard golden output**

### DIFF
--- a/pkg/controller/direct/monitoring/monitoringdashboard_controller.go
+++ b/pkg/controller/direct/monitoring/monitoringdashboard_controller.go
@@ -1,0 +1,281 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package monitoring
+
+import (
+	"context"
+	"fmt"
+
+	api "cloud.google.com/go/monitoring/dashboard/apiv1"
+	pb "cloud.google.com/go/monitoring/dashboard/apiv1/dashboardpb"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/monitoring/v1beta1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/config"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/references"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
+)
+
+func init() {
+	registry.RegisterModel(krm.MonitoringDashboardGVK, newDashboardModel)
+}
+
+func newDashboardModel(ctx context.Context, config *config.ControllerConfig) (directbase.Model, error) {
+	return &dashboardModel{config: config}, nil
+}
+
+type dashboardModel struct {
+	config *config.ControllerConfig
+}
+
+// model implements the Model interface.
+var _ directbase.Model = &dashboardModel{}
+
+type dashboardAdapter struct {
+	projectID  string
+	resourceID string
+
+	desired *pb.Dashboard
+	actual  *pb.Dashboard
+
+	dashboardsClient *api.DashboardsClient
+}
+
+// adapter implements the Adapter interface.
+var _ directbase.Adapter = &dashboardAdapter{}
+
+// AdapterForObject implements the Model interface.
+func (m *dashboardModel) AdapterForObject(ctx context.Context, kube client.Reader, u *unstructured.Unstructured) (directbase.Adapter, error) {
+	gcpClient, err := newGCPClient(ctx, m.config)
+	if err != nil {
+		return nil, fmt.Errorf("building gcp client: %w", err)
+	}
+
+	dashboardsClient, err := gcpClient.newDashboardsClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	obj := &krm.MonitoringDashboard{}
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &obj); err != nil {
+		return nil, fmt.Errorf("error converting to %T: %w", obj, err)
+	}
+
+	resourceID := ValueOf(obj.Spec.ResourceID)
+	if resourceID == "" {
+		resourceID = obj.GetName()
+	}
+	if resourceID == "" {
+		return nil, fmt.Errorf("cannot resolve resource ID")
+	}
+
+	projectRef, err := references.ResolveProject(ctx, kube, obj, &obj.Spec.ProjectRef)
+	if err != nil {
+		return nil, err
+	}
+	projectID := projectRef.ProjectID
+	if projectID == "" {
+		return nil, fmt.Errorf("cannot resolve project")
+	}
+
+	if err := VisitFields(obj, &refNormalizer{ctx: ctx, src: obj, kube: kube}); err != nil {
+		return nil, err
+	}
+
+	mapCtx := &MapContext{}
+	desiredProto := MonitoringDashboardSpec_ToProto(mapCtx, &obj.Spec)
+	if mapCtx.Err() != nil {
+		return nil, mapCtx.Err()
+	}
+
+	return &dashboardAdapter{
+		projectID:        projectID,
+		resourceID:       resourceID,
+		desired:          desiredProto,
+		dashboardsClient: dashboardsClient,
+	}, nil
+}
+
+// Find implements the Adapter interface.
+func (a *dashboardAdapter) Find(ctx context.Context) (bool, error) {
+	if a.resourceID == "" {
+		return false, nil
+	}
+
+	req := &pb.GetDashboardRequest{
+		Name: a.fullyQualifiedName(),
+	}
+	dashboard, err := a.dashboardsClient.GetDashboard(ctx, req)
+	if err != nil {
+		if IsNotFound(err) {
+			return false, nil
+		}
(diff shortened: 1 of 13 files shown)


justinsb:
I think we should merge this after we have more tests (at least #1806) - then I can also make the ref-test real

pkg/test/resourcefixture/testdata/basic/monitoring/v1beta1/monitoringdashboard/monitoringdashboardbasic/_http.log
```diff
@@ -23,26 +25,28 @@ X-Xss-Protection: 0
...
-        "weight": 2,
+        "weight": "2",
         "widgets": [
           {
             "title": "Widget 1",
             "xyChart": {
               "dataSets": [
                 {
-                  "plotType": "LINE",
+                  "plotType": 1,
```
justinsb, line 45: Kind of annoying, but the official client uses integer enum encoding, messing up our diffs.  I can try the proto form and see if it is any better.

pkg/test/resourcefixture/testdata/basic/monitoring/v1beta1/monitoringdashboard/monitoringdashboardbasic/_http.log
```diff
@@ -296,97 +302,11 @@ X-Xss-Protection: 0
 
 ---
 
-PATCH https://monitoring.googleapis.com/v1/projects/${projectId}/dashboards/monitoringdashboard-${uniqueId}?alt=json
+PATCH https://monitoring.googleapis.com/v1/projects/${projectId}/dashboards/monitoringdashboard-${uniqueId}?%24alt=json%3Benum-encoding%3Dint
 Content-Type: application/json
-User-Agent: kcc/controller-manager DeclarativeClientLib/0.0.1
```
justinsb, line 301: I think the diff is confused here because we don't do a GET before DELETE.  Maybe we should just do the GET!

justinsb: I added a semi-superfluous GET to delete, the diff now looks much better!

pkg/controller/direct/monitoring/monitoringdashboard_controller.go
```diff
@@ -0,0 +1,282 @@
...
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/references"
+)
+
+func init() {
+	directbase.ControllerBuilder.RegisterModel(krm.MonitoringDashboardGVK, newDashboardModel)
+}
+
+func newDashboardModel(config *controller.Config) directbase.Model {
+	ctx := context.TODO()
+	gcpClient, err := newGCPClient(ctx, config)
```
yuwenma, line 41: or we can move this to `AdapterForObject` so as we can use the passed in `context.Context` parameter.
The AdapterForObject will be initialized in each Reconcile loop.

scripts/github-actions/tests-e2e-direct.sh
```diff
@@ -24,23 +24,26 @@ cd ${REPO_ROOT}/
...
 
 echo "Running e2e tests fixtures for LoggingLogMetric direct reconciliation..."
-
 GOLDEN_OBJECT_CHECKS=1 \
 GOLDEN_REQUEST_CHECKS=1 \
 E2E_KUBE_TARGET=envtest RUN_E2E=1 E2E_GCP_TARGET=mock \
   go test -test.count=1 -timeout 600s -v ./tests/e2e -run 'TestAllInSeries/fixtures/explicitlogmetric|TestAllInSeries/fixtures/exponentiallogmetric|TestAllInSeries/fixtures/linearlogmetric|TestAllInSeries/fixtures/logbucketmetric'
 
-echo "Running scenarios tests for LoggingLogMetric direct reconciliation..."
+echo "Running e2e tests fixtures for MonitoringDashboard direct reconciliation..."
```
yuwenma, line 41: you may want to specify the `KCC_USE_DIRECT_RECONCILERS` flag because the monitoringDashboard is still using the `dcl2crd` label https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/monitoring/v1beta1/monitoringdashboard_types.go#L870 . Using `KCC_USE_DIRECT_RECONCILERS` flag can force using direct controller.

pkg/test/resourcefixture/testdata/basic/monitoring/v1beta1/monitoringdashboard/monitoringdashboardrefs/_http.log
```diff
@@ -175,26 +177,28 @@ X-Xss-Protection: 0
...
       {
-        "weight": 2,
+        "weight": "2",
         "widgets": [
           {
             "title": "Widget 1",
             "xyChart": {
               "dataSets": [
                 {
-                  "plotType": "LINE",
```
yuwenma, line 193: This is an interesting change. I guess we don't want to chnage it?

pkg/test/resourcefixture/testdata/basic/monitoring/v1beta1/monitoringdashboard/monitoringdashboardbasic/_generated_object_monitoringdashboardbasic.golden.yaml
```diff
@@ -3,7 +3,6 @@ kind: MonitoringDashboard
 metadata:
   annotations:
     cnrm.cloud.google.com/management-conflict-prevention-policy: none
-    cnrm.cloud.google.com/state-into-spec: merge
```
yuwenma, line 6: 👍

pkg/controller/direct/references/resourceid.go
```diff
@@ -0,0 +1,32 @@
...
+package references
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func GetResourceID(u *unstructured.Unstructured) (string, error) {
+	resourceID, _, err := unstructured.NestedString(u.Object, "spec", "resourceID")
```
yuwenma, line 24: I recalled to see a team discussion about the camel case  for `projectId` rather than `projectID`. Shall this be `resourceId` or `resourceID`?
(Not a blocker to this PR, just want to align the names here since this is a shareable function under direct/references)

pkg/controller/direct/monitoring/refs.go
```diff
@@ -0,0 +1,105 @@
...
+	"strings"
+
+	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/monitoring/v1beta1"
+	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/clients/generated/apis/k8s/v1alpha1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/references"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func normalizeResourceName(ctx context.Context, reader client.Reader, src client.Object, ref *v1alpha1.ResourceRef) (*v1alpha1.ResourceRef, error) {
```
yuwenma, line 29: Do you want to move this to `controller/direct/reference/` or something that can be sharable? I feel like this is a good piece to reuse and improve as we building up more resources.

pkg/controller/direct/monitoring/utils.go
```diff
@@ -25,3 +42,142 @@ func ValueOf[T any](p *T) T {
...
+	if old != nil {
+		status["conditions"] = old["conditions"]
+		status["observedGeneration"] = old["observedGeneration"]
+	}
+
+	u.Object["status"] = status
+
+	return nil
+}
+
```
yuwenma, line 113: qq: do you think we shall move this to the direct_controller`Reconcile`?

pkg/test/resourcefixture/testdata/basic/monitoring/v1beta1/monitoringdashboard/monitoringdashboardbasic/_http.log
yuwenma: I missed this message. Please ignore my other comment.

pkg/controller/direct/monitoring/monitoringdashboard_controller.go
justinsb: My one concern is that building the client may be expensive: if we have to open a new TCP connection, or if we have to get the token again.  But I don't want to block this PR on it, so I'll move it!

scripts/github-actions/tests-e2e-direct.sh
justinsb: I think I am .... I exported it on line 27.

pkg/controller/direct/monitoring/monitoringdashboard_controller.go
justinsb: Ah, I guess we're calling the expensive call every time anyway :-)

scripts/github-actions/tests-e2e-direct.sh
yuwenma: oh, I didn't realize the export there.

pkg/controller/direct/references/resourceid.go
justinsb: So ...  go says "ID".  KRM style guide says "ID".  Google APIs use "Id" (in JSON).  Most KCC fields use `Id`, but I guess here we use `resourceID`.

If the KCC field was called `resourceId`, I don't know whether I would follow go conventions or try to match the name of the field.

In this case though, we're reading a field, and I think that field's name is `resourceID`.  That's correct against KRM style guide, but possibly inconsistent with most other GCP-generated fields.

I think it's your decision @yuwenma , and it's going to be interesting!

(Our best bet might be to stop adding fields that end in Id at all, maybe it's another marker that something is actually a ref ... or we could just drop ID arguing that it is the "type" of the field, and we don't say "fooString" so why do we say "fooID" or "fooURL" ? )

pkg/controller/direct/monitoring/refs.go
justinsb: Yes, but my "rule of thumb" is to move it when we have two copies.  For this particular one, we only have one copy  (this is actually very specialized - there's a field called "resourceName" in MonitoringDashboard.  Note that this method does call into shared functions for most of its logic!

justinsb: Actually I guess this is resolving a resource of arbitrary kind, so is more reusable than I first thought.  Still, it's our first ... let's move it when we have two or three instances.

pkg/controller/direct/monitoring/utils.go
justinsb: We should move it, along with most/all of the functions in utils.  I just don't think we know where to move it to yet!

yuwenma:
/lgtm
/approve

 I think the presubmit should be fixed in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1979. Could you rebase this PR? It looks great!

PR #1866, opened 2024-05-21
Add MockGCP for SpannerInstance

yuwenma:
* Generate the Golden log against real GCP
* Add SpannerInstance mock API proto 
* Add SpannerInstance mock CRUD logic, specifically the special handling on `nodecound` and `processingunits`
* Adjust golden log against mock GCP (use TF-based spanner instance).


mockgcp/mockspanner/admin/instance.go
```diff
@@ -0,0 +1,150 @@
...
+	fqn := name.String()
+	obj := &pb.Instance{}
+	if err := s.storage.Get(ctx, fqn, obj); err != nil {
+		return nil, err
+	}
+	now := timestamppb.Now()
+	obj.UpdateTime = now
+	source := reflect.ValueOf(req.Instance)
+	target := reflect.ValueOf(obj).Elem()
+	for _, path := range req.FieldMask.Paths {
```
yuwenma, line 108: A simple approach to make UPDATE more like a real GCP UPDATE with update mask_field.

justinsb: Nice!  We should split this out into a helper method!

justinsb (approved):

justinsb:
Looks great - and that helper is cool!


/approve
/lgtm

PR #1869, opened 2024-05-21
Find the hubfeaturemembership resource when doing the Delete

ziyue-101:
### Change description

For the resource, the controller needs to do a get before `delete` because the `delete` only cleans up a field for the Gkehub feature resource. Otherwise , `adapter.actual` is not initialized in Delete in the base controller

### Tests you have done

### DIFF
--- a/pkg/controller/direct/gkehub/featuremembership_controller.go
+++ b/pkg/controller/direct/gkehub/featuremembership_controller.go
@@ -121,6 +121,14 @@ func (a *gkeHubAdapter) Find(ctx context.Context) (bool, error) {
 
 // Delete implements the Adapter interface.
 func (a *gkeHubAdapter) Delete(ctx context.Context) (bool, error) {
+	exist, err := a.Find(ctx)
+	if err != nil {
+		return false, fmt.Errorf("finding feature for %s:%w", a.featureID, err)
+	}
+	if !exist {
+		// return (false, nil) if the object was not found but should be presumed deleted.
+		return false, nil
+	}
 	// emptying the membershipspec is sufficient
 	a.desired = &featureapi.MembershipFeatureSpec{}
 	if _, err := a.patchMembershipSpec(ctx); err != nil {


pkg/controller/direct/gkehub/featuremembership_controller.go
```diff
@@ -121,6 +121,13 @@ func (a *gkeHubAdapter) Find(ctx context.Context) (bool, error) {
 
 // Delete implements the Adapter interface.
 func (a *gkeHubAdapter) Delete(ctx context.Context) (bool, error) {
+	exist, err := a.Find(ctx)
+	if err != nil {
+		return false, fmt.Errorf("finding feature for %s:%w", a.featureID, err)
+	}
+	if !exist {
+		return false, fmt.Errorf("expecting feature for %s, but not found", a.featureID)
+	}
```
acpana, line 130: sanity check: don't we want to support `exists == false` as it doesn't exist. I'm thinking what if end up reconciling a stale `delete` event  for instance, we will err out and fail the reconcile possibly?

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/aaa5bbdaea628288685cded6b54166d6fda795cf/pkg/controller/direct/directbase/directbase_controller.go#L241-L255

It may be sufficient to just error out not found.

ziyue-101: Do you have a scenario in your mind that KCC controller may reconcile a stale delete event?

I am leaning towards to error out on unexpected delete reconciliations.

ziyue-101: Thanks for explaining, returning (false, nil) then if notfound

acpana: Sweet, yes, not returning err on not found was what I wanted to make sure we do 😛 

Thanks Justin for clarifying and Ziyue for following up!

pkg/controller/direct/gkehub/featuremembership_controller.go
```diff
@@ -121,6 +121,14 @@ func (a *gkeHubAdapter) Find(ctx context.Context) (bool, error) {
 
 // Delete implements the Adapter interface.
 func (a *gkeHubAdapter) Delete(ctx context.Context) (bool, error) {
+	exist, err := a.Find(ctx)
+	if err != nil {
+		return false, fmt.Errorf("finding feature for %s:%w", a.featureID, err)
```
acpana, line 126:
```suggestion
		return false, fmt.Errorf("finding feature membership for %s:%w", a.featureID, err)
```

maybe it makes more sense but if so please feel free to follow up in a separate PR

ziyue-101: Actually, the Find gets an Gkehubfeature resource back, there is no GCP resource called featuremembership

PR #1871, opened 2024-05-21
tests: add logbucketmetric to direct tests

acpana:
Run the logbucketmetric in the direct tests. ~Note, I am not convinced we need it since the postsubmits hit the direct controller now but~ it's probably good practice to add it for now until we can better target all of the tests under a GVK.


pkg/test/resourcefixture/testdata/basic/logging/v1beta1/logginglogmetric/logbucketmetric/_generated_export_logbucketmetric.golden
acpana, line 1: oh I was really surprised by this. I'd've expected this to be a `.yaml`. I think we/ I raced with #1447

I can try to regenerate.

justinsb:
I think the big win from the "direct" test right now is that we're strict, e.g. we check the golden output (both object and the http log).  That's a much higher bar than we run with normally, and I think we only meet that bar on the resources that are implemented directly or where we intend to shortly.

So I like this direction!

yuwenma:
/lgtm
/approve

Thanks Alex!

PR #1872, opened 2024-05-22
Add APIKeysKey go types generated from proto

justinsb:
- **chore: reflow descriptions when generating CRDs**
- **chore: add APIKeysKey go types generated from proto**
- **autogen: generate direct-actuation CRDs**


dev/tasks/generate-crds
```diff
@@ -38,6 +38,7 @@ go run ./scripts/crd-tools set-field spec.preserveUnknownFields=false --dir apis
 go run ./scripts/crd-tools delete-field status --dir apis/config/crd/
 go run ./scripts/crd-tools set-annotation cnrm.cloud.google.com/version=0.0.0-dev --dir apis/config/crd/
 go run ./scripts/crd-tools delete-annotation controller-gen.kubebuilder.io/version --dir apis/config/crd/
+go run ./scripts/crd-tools reflow-descriptions --dir apis/config/crd/
```
yuwenma, line 41: We don't have apis/config/crd dir yet. Is this for future structure?

scripts/crd-tools/cmd/reflowdescriptions/reflowdescriptions.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func (v *visitor) visitSlice(s []any) []any {
+	for i, val := range s {
+		s[i] = v.visitAny(val)
+	}
+	return s
+}
+
+func (v *visitor) visitDescription(s string) string {
```
yuwenma, line 85: Very cool!

yuwenma (commented):
/lgtm

dev/tasks/generate-crds
justinsb: So actually this script generates that directory.  We don't check it in currently, and it does get "left behind" if you run it locally.  I'm not sure if we should delete it in the script (or generate it under `.build`).  WDYT?

apis/apikeys/v1alpha1/apikey_generated.go
```diff
@@ -0,0 +1,179 @@
...
+)
+
+// +kcc:proto=google.api.apikeys.v2.AndroidApplication
+type AndroidApplication struct {
+	// The SHA1 fingerprint of the application. For example, both sha1 formats are
+	//  acceptable : DA:39:A3:EE:5E:6B:4B:0D:32:55:BF:EF:95:60:18:90:AF:D8:07:09 or
+	//  DA39A3EE5E6B4B0D3255BFEF95601890AFD80709.
+	//  Output format is the latter.
+	// +required
+	Sha1Fingerprint *string `json:"sha1Fingerprint,omitempty"`
```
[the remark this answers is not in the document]
justinsb: So `+required` works now, after https://github.com/kubernetes-sigs/controller-tools/pull/944

We don't yet validate the input format (though I would expect GCP to reject it if invalid, so the user will still get a reasonable error, we shouldn't fail silently)

Because the input and output formats are potentially different, we'll need to be careful here, we can't just do a naive diff to determine if we need to re-reconcile.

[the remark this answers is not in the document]
yuwenma: > So `+required` works now, after [kubernetes-sigs/controller-tools#944](https://github.com/kubernetes-sigs/controller-tools/pull/944)
> 

Ah! This is so fresh and good news!

> We don't yet validate the input format (though I would expect GCP to reject it if invalid, so the user will still get a reasonable error, we shouldn't fail silently)
> 

Agree. I think I want to have basic and general validation on CRD, and have more server specific/usage related validation from the GCP server.   

> Because the input and output formats are potentially different, we'll need to be careful here, we can't just do a naive diff to determine if we need to re-reconcile.

Yes. This makes a lot of sense.

dev/tasks/generate-crds
yuwenma: Oh I see what this script does. (I was more thinking about keeping config/apis as the single CRD dir)

I"ll vote for `.build`.

yuwenma:
/lgtm
/approve

Thank you for this PR. very cool!

PR #1875, opened 2024-05-22
Getter as Second Expander

barney-s:
### Change description
 
Getter as Second Expander
- Implement grpc based getter expander
- Getter config is inline in Composition CRD
- Add release manifests for getter-expander
- Add e2e tests for getter-expander
- Add presubmit for getter-expander e2e

Next steps:

- Split out Getter config into its own CRD


experiments/compositions/scripts/github-actions/getter-test.sh
```diff
@@ -0,0 +1,21 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+set -o errexit
+set -o nounset
+set -o pipefail
+
+export PATH=$PATH:$(go env GOPATH)/bin
+cd experiments/compositions/composition
+GCP_PROJECT_ID=something make unit-test-expander-getter
```
xiaoweim, line 21: is this temporary :)

xiaoweim (commented):
/lgtm

cheftako:
Looks like a good start. I assume there will be a follow up to remove the getter functionality which is nestled in the jinja2 expander while keeping the existing CUJ's working?

experiments/compositions/scripts/github-actions/getter-test.sh
barney-s: No this is permanent for the kind based testing in github. Locally we have a GCP project context. In the github runner we dont have one. So setting it to a dummy value so that we dont have to change our scripts.

PR #1877, opened 2024-05-22
Support setting IAM on PrivateCACAPool

justinsb:
We do this in our direct controller, though we only need to implement the (new) IAM-related methods.

These methods are optional (at least for now!)


yuwenma (commented):
/lgtm
/approve

Looks great and thank you updating the controller template! (Sorry for the delayed review, this PR takes especially long time to understand).

PR #1813, opened 2024-05-15
enh: check failed updates in conditions 

acpana:
There can be a situation where the a failed update gets marked afterwards as `UpToDate` .

Big shout out to @gemmahou for catching this.

PR #1879, opened 2024-05-22
chore: tests: run all llm direct tests

acpana:
patch

- removes metric_descriptor custom_diff: we don't need this test now because we enforce field immutability and those fields are immutable
- write output for tests (mostly to capture the export since we support it now)
- refactor to subclass those steps field tests under `fields` not `direct`


tests/e2e/testdata/scenarios/direct/llm_custom_diff/value_type/label/_export0.yaml
gemmahou, line 1: Do you know why it's necessary to capture the exports?

tests/e2e/testdata/scenarios/direct/llm_immutable_field/labels_value_type/_object01.yaml
```diff
@@ -40,18 +40,9 @@ spec:
...
-          "@type": "type.googleapis.com/google.rpc.DebugInfo",
-          "detail": "[ORIGINAL ERROR] generic::invalid_argument: Cannot change value type of label: testkey [google.rpc.error_details_ext] { message: \"Cannot change value type of label: testkey\" }"
-        }
-      ]
-      , badRequest
-    reason: UpdateFailed
-    status: "False"
+    message: The resource is up to date
+    reason: UpToDate
+    status: "True"
```
gemmahou, line 45: This change is unexpected... we would expect an error message rather than "UpToDate". Have you rebased with #1813?

tests/e2e/testdata/scenarios/direct/llm_custom_diff/value_type/label/_export0.yaml
acpana: Yes I think with https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1832 in these need their exported objects now.

tests/e2e/testdata/scenarios/direct/llm_immutable_field/labels_value_type/_object01.yaml
acpana: Yes it is rebased and good catch! I had a similar moment of reflection here.

The caveat is that the test was captured against `real` AND as a DCL controller whereas this is being captured against `mock` and as a direct controller.

So this is where we are with the `direct` controller against `mock` ... that means we need to improve the mock! You can also see that the `export` is changed but that shouldn't happen for the `label`. 

If it's ok with you, could I improve the mock as a follow on ?

gemmahou:
/lgtm
/approve

Let's improve the mock in another PR.

PR #1883, opened 2024-05-23
chore: clean up after renaming

acpana:
Sigh, I forgot to git add the deleted files from https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1879


acpana:
Hm, I can see them here: https://github.com/GoogleCloudPlatform/k8s-config-connector/actions/runs/9200524734/job/25307226548

PR #1885, opened 2024-05-23
Switch jinja2 to grpc based expansion

barney-s:
### Change description

Switch jinja2 to grpc based expansion:
- rename gjinjg2 -> jinja2 (making jinja2 grpc based)
- rename existing pod based jinja2 -> pjinja2
- Add `image` field to ExpanderVersion
  - allow specifying an image that doesnt follow the
    removePrefix(expanderversion.name , "composition-") pattern.
  - provides more flexibility
  - In our case the type is pjinja and the image is expander-jinja
- Add a custom ratelimiter for the facade reconciler
  - Due to a bug, the reconciler is not restarted across tests.
  - Old failures add to rate limiting causing random failures
  - make ratelimiter max backoff 120s
- Add a small sleep b/w tests
- Move all ExistTimeout to CompositionReconcileTimeout
- Check if a plan was updated and if so check if the applier sees the
  latest plan object. If not retry again.
- Add InputGeneration and Generation to plan status

### DIFF
--- a/experiments/compositions/composition/api/v1alpha1/expanderversion_types.go
+++ b/experiments/compositions/composition/api/v1alpha1/expanderversion_types.go
@@ -38,11 +38,11 @@ const (
 
 // ExpanderVersionSpec defines the desired state of ExpanderVersion
 type ExpanderVersionSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
-
 	// ImageRegistry is the designated registry for where to pull the named expander image
 	ImageRegistry string `json:"imageRegistry,omitempty"`
+
+	// Image if different from removePrefix(expanderversion.name , "composition-")
+	Image string `json:"image,omitempty"`
 
 	// ValidVersions is a list of valid versions of the named expander
 	ValidVersions []string `json:"validVersions"`
--- a/experiments/compositions/composition/config/expanders/expander_versions.yaml
+++ b/experiments/compositions/composition/config/expanders/expander_versions.yaml
@@ -15,11 +15,12 @@
 apiVersion: composition.google.com/v1alpha1
 kind: ExpanderVersion
 metadata:
-  name: jinja2
+  name: jinja2-job
   namespace: system
 spec:
   type: job
   imageRegistry: gcr.io/krmapihosting-release
+  image: expander-jinja2
   validVersions:
   # current golang semver package only support version comparison
   # in the format of `v{num}.{num}.{num}-{alphabet}`, `v{num}.{num}.{num}` and `{num}.{num}.{num}`.
@@ -30,7 +31,7 @@ spec:
 apiVersion: composition.google.com/v1alpha1
 kind: ExpanderVersion
 metadata:
-  name: gjinja2
+  name: jinja2
   namespace: system
 spec:
   type: grpc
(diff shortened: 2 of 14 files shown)


experiments/compositions/composition/api/v1alpha1/expanderversion_types.go
```diff
@@ -38,12 +38,12 @@ const (
...
 // ExpanderVersionSpec defines the desired state of ExpanderVersion
 type ExpanderVersionSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
-
 	// ImageRegistry is the designated registry for where to pull the named expander image
 	ImageRegistry string `json:"imageRegistry,omitempty"`
 
+	// Image if different from removePrefix(expanderversion.name , "composition-")
+	Image string `json:"image,omitempty"`
```
xiaoweim, line 45: for my own clarification, do we have an example for of this image string, is it going to be version?

experiments/compositions/composition/config/expanders/expander_versions.yaml
```diff
@@ -30,7 +31,7 @@ spec:
 apiVersion: composition.google.com/v1alpha1
 kind: ExpanderVersion
 metadata:
-  name: gjinja2
+  name: jinja2
   namespace: system
 spec:
   type: grpc
```
xiaoweim, line 37: does this mean that for grpc we name it to the default jinja2, and for type job, we call it pjinja2? 
nit: Is there a reason why it is called pjinja?

experiments/compositions/composition/api/v1alpha1/expanderversion_types.go
cheftako: Guessing this will look like "gcr.io/MY-PROJECT/MY-EXPANDER:v1.0.12" or some such

experiments/compositions/composition/config/expanders/expander_versions.yaml
```diff
@@ -15,11 +15,12 @@
 apiVersion: composition.google.com/v1alpha1
 kind: ExpanderVersion
 metadata:
-  name: jinja2
+  name: pjinja2
```
cheftako, line 18: *p*jinja2 ?

cheftako: Ahh. I think I see. Should we make this more explicit? Jinja2-job ? The following then might be the jinja2-service, jinja2-grpc?

experiments/compositions/composition/api/v1alpha1/expanderversion_types.go
barney-s: Adding optional image that allows user to specify an image instead of us deducing image name from the expander version name. 

```
  imageRegistry: gcr.io/krmapihosting-release
  image: expander-jinja2
  validVersions:
     ...
  ```

experiments/compositions/composition/config/expanders/expander_versions.yaml
barney-s: It only matters if we are doing the same expander type in both modes. Jinja is an exception since we started as a job and are moving towards grpc. ideally we should drop the job,. but keeping it around for just-in-case.

 So will rename: pjinja2 -> jinja2-job 
But will not rename: jinja2 -> jinja2-grpc X

experiments/compositions/composition/config/expanders/expander_versions.yaml
barney-s: Default is grpc. Yes. 

pjinja2 -> pod jinja2 :) 

planning to rename it to jinja2-job to be explicit.

xiaoweim: Thanks for the explanation!

PR #1889, opened 2024-05-23
AlloyDB Public IP KCC Support

nancynh:
### Change description

Adds KCC support for AlloyDB instances to have an inbound public IP address, and set a list of authorized networks. Specifically adds the field: `networkConfig` which includes `authorizedExternalNetworks` and `enablePublicIp`. This is part of the [public IP feature](https://cloud.google.com/alloydb/docs/connect-public-ip).

Also updates the mock tests to allow instance updates with the `networkConfig` field.

Fixes b/313480670

### Tests you have done

Have ran the e2e test ([logs](https://paste.googleplex.com/5948731354775552)) to verify changes work as expected.


maqiuyujoyce:
/lgtm

Thank you for supporting the new feature, @nancynh ! Could you fix the unit test so that it can pass? I think you'll want to add the fields to tests/apichecks/testdata/exceptions/acronyms.txt.

mockgcp/mockalloydb/instance.go
yuwenma, line 1: Can you run `hack/compare-mock fixtures/alloydb` to update the MockGCP golden logs?

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_alloydbinstances.alloydb.cnrm.cloud.google.com.yaml
```diff
@@ -180,6 +180,27 @@ spec:
...
+                      is set to true.
+                    items:
+                      properties:
+                        cidrRange:
+                          description: CIDR range for one authorized network of the
+                            instance.
+                          type: string
+                      type: object
+                    type: array
+                  enablePublicIp:
```
yuwenma, line 198: The acronym should be `IP` I think.
Can we change this to `enablePublicIP`?

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_alloydbinstances.alloydb.cnrm.cloud.google.com.yaml
```diff
@@ -242,6 +263,11 @@ spec:
                   current reported status reflects the most recent desired state of
                   the resource.
                 type: integer
+              publicIpAddress:
```
yuwenma, line 266: ditto publicIPAddress

config/samples/resources/alloydbcluster/regular-cluster/alloydb_v1beta1_alloydbcluster.yaml
```diff
@@ -55,4 +55,4 @@ spec:
   initialUser:
     user: "postgres"
     password:
-      value: "postgres"
```
yuwenma, line 58: Thank you for updating this. 
This field seems to contain sensitive data. I think a better approach is to use k8s secret to [store this data](https://cloud.google.com/config-connector/docs/how-to/secrets). 
Not a blocker of this PR, this is something need to be considered to switch.

config/samples/resources/alloydbinstance/primary-instance/alloydb_v1beta1_alloydbinstance.yaml
```diff
@@ -23,5 +23,11 @@ spec:
     name: alloydbinstance-dep-primary
   databaseFlags:
     enable_google_adaptive_autovacuum: "off"
+    password.enforce_complexity: "on"
```
yuwenma, line 26: Thanks for updating this field example.
I think this field should be changed to `passwordPolicyFlags`, with struct value `enforce_complexity`, `expiration_in_days `, etc 

```
passwordPolicyFlags:
    enforce_complexity: "on"
    expiration_in_days: "90"
```

Also not a blocker.

pkg/test/resourcefixture/testdata/basic/alloydb/v1beta1/alloydbcluster/fullalloydbcluster/create.yaml
```diff
@@ -27,7 +27,7 @@ spec:
   initialUser:
     user: postgres
     password:
-      value: postgres
```
yuwenma, line 30: btw, I'm just curious about this change. Is there a specific reason to update the  `password.value` (like password requires to contain numbers)?

yuwenma:
cc @anhdle-sso for MockGCP review.

config/samples/resources/alloydbinstance/primary-instance/alloydb_v1beta1_alloydbinstance.yaml
maqiuyujoyce: `databaseFlags` is a freeform map field so I think the value will be whatever the API takes. And looks like `password.enforce_complexity` is exactly the value the API asks the user to pass in: https://cloud.google.com/alloydb/docs/manage-password-policy#enforce-password-complexity

I personally also don't completely agree with the structure/format, but I guess this is an API decision that's already GAed, and any improvement of it is out of the scope for this PR.

mockgcp/mockalloydb/instance.go
nancynh: Done

pkg/test/resourcefixture/testdata/basic/alloydb/v1beta1/alloydbcluster/fullalloydbcluster/create.yaml
nancynh: For a user to enable the public IP feature, they'll also need to enforce password complexity as a security requirement. Hence, the password change

config/samples/resources/alloydbinstance/primary-instance/alloydb_v1beta1_alloydbinstance.yaml
nancynh: Yes, like Joyce said we have to use `password.enforce_complexity` as the flag name - I think it'd be a bit difficult for us to change the AlloyDB API format with the database flags at this point unfortunately

yuwenma: Okay, as a FYI, this could block users to use some popular k8s tools and block configconnector to correctly populate the field value in certain cases. See
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#selecting-fields

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_alloydbinstances.alloydb.cnrm.cloud.google.com.yaml
maqiuyujoyce: The other suggestion is about an output-only field under `spec`. This output-only field is under `status` so it is a different situation.

But @nancynh, if this is a field you don't expect users to leverage and @yuwenma if you are feeling strongly about supporting the acronym then we can just ignore this field for now.

config/samples/resources/alloydbinstance/primary-instance/alloydb_v1beta1_alloydbinstance.yaml
maqiuyujoyce: Hi @yuwenma, do you mean that the field name `password.enforce_complexity` may be invalid for some K8s tools so that users cannot select this field in certain use cases?

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_alloydbinstances.alloydb.cnrm.cloud.google.com.yaml
nancynh: Ah okay, thanks for the explanation! I expect users will want to use this field to know which public IP address they can use to connect to the DB

config/samples/resources/alloydbinstance/primary-instance/alloydb_v1beta1_alloydbinstance.yaml
nancynh: > Okay, as a FYI, this could block users to use some popular k8s tools and block configconnector to correctly populate the field value in certain cases. See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#selecting-fields

Thanks for letting me know! This is surprising, we have other flags that follow this same format of having a period in the flag name e.g. [`google_columnar_engine.auto_columnarization_schedule`](https://cloud.google.com/alloydb/docs/reference/columnar-engine-flags#auto-columnarization-schedule) and [`google_db_advisor.max_statement_length`](https://cloud.google.com/alloydb/docs/reference/index-advisor-flags#max_statement_length)

CloudSQL which also has KCC support I think does a similar pattern for their db flags: https://cloud.google.com/sql/docs/postgres/flags#list-flags-postgres. I'd expect this to be a problem for them too

nancynh:
@maqiuyujoyce would you know how to fix the unit tests? I've added the fields in tests/apichecks/testdata/exceptions/acronyms.txt, but now it's complaining about a newline character:
```
=== FAIL: tests/apichecks TestCRDsAcronyms (1.52s)
    utils.go:99: unexpected diff in testdata/exceptions/acronyms.txt:   strings.Join({
          	... // 95608 identical bytes
          	"d=workflowsworkflows.workflows.cnrm.cloud.google.com version=v1a",
          	`lpha1: field ".status.revisionId" should be ".status.revisionID"`,
        - 	"\n",
          }, "")
```

nancynh:
FYI mock tests are failing in an non-AlloyDB test which I don't think are related to my changes `TestAllInSeries/fixtures/externalwithsubresource`:

```
2024-06-26T17:20:00.1400329Z     samples.go:122: error creating resource: Internal error occurred: failed calling webhook "container-annotation-handler.cnrm.cloud.google.com": failed to call webhook: Post "https://127.0.0.1:38699/container-annotation-handler?timeout=10s": tls: first record does not look like a TLS handshake
```

But they were passing before I rebased and fixed the unit tests in 32bfc458a4c96117b8854741d13c7b9eb46fbf9e

config/samples/resources/alloydbinstance/primary-instance/alloydb_v1beta1_alloydbinstance.yaml
yuwenma: Yes, I think so. https://kubernetes.io/docs/reference/kubectl/jsonpath (if users use json path)

maqiuyujoyce (commented):
/lgtm
/approve

/hold cancel

nancynh:
Thanks @yuwenma and @maqiuyujoyce for the review!!

PR #1892, opened 2024-05-23
feat: set ua per req

acpana:
While it is true that we attempt to set the ua in the `direct` codebase, it seems that it's overwritten when we use the `WithHttpClient` option.


acpana:
The tests fail and need regeneration with:

```
       - 	User-Agent: google-api-go-client/0.5
        + 	User-Agent: kcc/controller-manager
```

PR #1894, opened 2024-05-23
chore:tests: further cleanup

acpana:
I missed these files in #1883 somehow.

As more context we want to have `fields` tests for all resources eventually, even if they are not migrated necessarily to the `direct` controller.


gemmahou:
lol thanks!
/lgtm
/approve

PR #1901, opened 2024-05-28
mockgcp:feat: add ids mockgcp

acpana:
Staging mockids


acpana:
The proto published at https://github.com/googleapis/googleapis/blob/1e4137870560340a14700618a05e2d7162326af7/google/cloud/ids/v1/ids.proto is, unfortunately, not up to date. See how our (autogenerated) type pulls in `ThreatExceptions`.

I will be using the gapis to generate from https://github.com/googleapis/google-api-go-client/blob/b49e3b908a8ed562e068736f1c42e992538ba6e0/ids/v1/ids-api.json

acpana:
/hold cancel

Thanks for reviewing Yuwen! If I missed something I will follow up!

PR #1902, opened 2024-05-29
Change the accelerate type 'nvidia-tesla-k80' as it's no longer supported

maqiuyujoyce:
### Change description

We started receiving errors in integration tests like below:

```
Update call failed: error applying desired state: summary: error creating NodePool: googleapi: Error 400: Accelerator type "nvidia-tesla-k80" does not exist in zone us-east1-c.
```

This is likely due to https://cloud.google.com/compute/docs/eol/k80-eol. So updating it to `nvidia-tesla-t4` according to the recommendation in the doc.

### Tests you have done

`go test -v -tags=integration ./config/tests/samples/create -test.run TestAll -run-tests basic-node-pool` passed!


acpana:
/hold

qq -- is there a default value that would never go away here hehe? Wondering if we can future proof this some more?

Regardless, thanks for digging into this and making the changes 🏅 

/lgtm
/approve

maqiuyujoyce:
> qq -- is there a default value that would never go away here hehe? Wondering if we can future proof this some more?

Didn't find an default option: https://cloud.google.com/kubernetes-engine/docs/how-to/gpus, and according to the TF schema, it is required: https://github.com/hashicorp/terraform-provider-google/blob/v5.31.1/google/services/container/node_config.go#L147, so unfortunately we need to update the value whenever the type is EOL.

/hold cancel

PR #1910, opened 2024-05-30
Register direct controller from crd annotation

yuwenma:
### Change description
Currently, developers need to register each direct controller  in multiple places, and needs to list their dedicated GVK values. 

This PR proposes two changes:
1.  Register the direct controller's "Model" via go `init`.  This should simplify the code and increase the readability as: developing a direct controller starts from register a `Model` --> the `Model` implements a  `Adapter` --> the `Adapter` implements   `crud` + `export`. 
5. ~Introduce a new CRD label `cnrm.cloud.google.com/directcrd=true`, so~ developers do not need to list the new resource GVK in both CC registry and the e2e test base workflow. ~This label approach is the same as the TF and DCL based CRDs.~ Direct controller will be the default controller to use. If using TF/DCL controller, developers need to use the corresponding CRD labels. `KCC_USE_DIRECT_RECONCILERS` flag will override the TF/DCL label to force using the direct controller. 

 ### How to use

- Register your direct controller in `pkg/controller/direct/register/register.go`
```go
import  _ "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/apikeys"
```
- Register your direct controller `Model` to `directbase.ControllerBuilder`
```
func init() {
	directbase.ControllerBuilder.RegisterModel(krm.APIKeysKeyGVK, func (config *controller.Config) directbase.Model)
}
```


pkg/controller/direct/register/register.go
yuwenma, line 1: We can also restructure the `pkg/controller/direct` directly. But I don't plan to restructure the entire code base right now.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_apikeyskeys.apikeys.cnrm.cloud.google.com.yaml
```diff
@@ -5,10 +5,10 @@ metadata:
     cnrm.cloud.google.com/version: 0.0.0-dev
   creationTimestamp: null
   labels:
+    cnrm.cloud.google.com/directcrd: "true"
```
justinsb, line 8: Nit: `cnrm.cloud.google.com/direct` (it's the controller that's direct, not really the CRD).  Or it could just the default if we don't specify tf2crd or dcl2crd ?

pkg/controller/direct/apikeys/apikeyskey_controller.go
```diff
@@ -36,12 +35,12 @@ import (
 	. "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/mappings" //nolint:revive
 )
 
-// AddKeyReconciler creates a new controller and adds it to the Manager.
-// The Manager will set fields on the Controller and start it when the Manager is started.
-func AddKeyReconciler(mgr manager.Manager, config *controller.Config, opts directbase.Deps) error {
-	gvk := krm.APIKeysKeyGVK
+func init() {
+	directbase.ControllerBuilder.RegisterModel(krm.APIKeysKeyGVK, NewModel)
```
justinsb, line 39: I like this!

pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -53,22 +54,41 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
-// Add creates a new controller for reconciling objects of the specified GVK, delegating actual resource reconciliation to the provided Model.
-func Add(mgr manager.Manager, gvk schema.GroupVersionKind, model Model, opts Deps) error {
+var ControllerBuilder directControllerBuilder
+
+func init() {
+	ControllerBuilder = directControllerBuilder{modelMapper: map[schema.GroupVersionKind]func(*controller.Config) Model{}}
```
justinsb, line 60: Nit: We may be able to get away without an init function here.  But ... I can see it getting more complicated, so only a nit!

pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -53,22 +54,41 @@ import (
...
+}
+
+type directControllerBuilder struct {
+	modelMapper map[schema.GroupVersionKind]func(*controller.Config) Model
+}
+
+func (c *directControllerBuilder) RegisterModel(gvk schema.GroupVersionKind, modelFn func(*controller.Config) Model) {
+	c.modelMapper[gvk] = modelFn
```
justinsb, line 68: Nit: One trick, if it isn't perf sensitive (and this isn't, I believe), is to check if modelMapper is nil and initialize the map here.

pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -53,22 +54,41 @@ import (
...
+}
+
+func (c *directControllerBuilder) Add(mgr manager.Manager, config *controller.Config, gvk schema.GroupVersionKind, deps Deps) error {
```
justinsb, line 71: Nit: Maybe `AddController`?

pkg/controller/registration/registration_controller.go
```diff
@@ -247,6 +221,14 @@ func registerDefaultController(r *ReconcileRegistration, config *controller.Conf
 		}
 
 	default:
+		// register controllers for direct CRDs
+		if val, ok := crd.Labels[k8s.DirectCRDLabel]; ok && val == "true" {
```
justinsb, line 225: Might be easier to have a `directbase.ControllerBuilder.IsDirect` helper, than to require the additional label.

scripts/github-actions/tests-e2e-direct.sh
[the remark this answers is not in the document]
justinsb: Are we able to turn on the golden checks in tests-e2e-fixtures?  That's the big advantage of these tests, they are actually much stricter....

justinsb (commented):
Looks good, a few nits/suggestions

gemmahou:
I noticed the vcr test failed due to test timeout on "apikeyskey". Unsure if that's related to this PR, but we can fix it by 
1) Regenerate the vcr log and make sure the "replay" pass locally
2) Skip this test, and I'll regenerate the log later and re-enable it. An example: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/config/tests/samples/create/harness.go#L681-L682, we'll need to comment out this test from the allow list.

apis/apikeys/v1alpha1/apikey_type.go
```diff
@@ -41,7 +41,7 @@ var (
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:resource:categories=gcp,shortName=gcpapikeyskey;gcpapikeyskeys
 // +kubebuilder:subresource:status
-// +kubebuilder:metadata:labels="cnrm.cloud.google.com/tf2crd=true";"cnrm.cloud.google.com/managed-by-kcc=true";"cnrm.cloud.google.com/stability-level=alpha";"cnrm.cloud.google.com/system=true"
```
acpana, line 44: With this change, can we still have resources in transition? For instance, for ApiKey AFAICT, we didn't actually make the switch to a direct CRD but I could be wrong. 

What I like about the current state of the world is that we can turn on TF/ DCL based resources to use the DirectController via the feature flag (`KCC_USE_DIRECT_RECONCILERS`) and have developers iterate between the behavior of a TF/ DCL based controller and a direct one. Can we preserve that somehow?

.github/workflows/presubmit.yaml
```diff
@@ -112,17 +112,6 @@ jobs:
...
-    runs-on: ubuntu-22.04
-    timeout-minutes: 60
-    steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-go@v4
-        with:
-          go-version-file: 'go.mod'
-      - name: "run tests-e2e-direct"
-        run: |
-          ./scripts/github-actions/tests-e2e-direct.sh
```
acpana, line 125: I think in the end state we don't have `direct` tests specifically but I think these can be useful for folks, especially as they iterate on a resource conversion.

scripts/github-actions/tests-e2e-direct.sh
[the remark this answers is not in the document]
yuwenma: > Are we able to turn on the golden checks in tests-e2e-fixtures? 

I think we should. Do you happen to know why we don't add the golden check in  tests-e2e-fixtures previously?

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_apikeyskeys.apikeys.cnrm.cloud.google.com.yaml
yuwenma: I'll change this to "cnrm.cloud.google.com/direct".  And +1 to make the direct as default eventually.

pkg/controller/direct/directbase/directbase_controller.go
yuwenma: Sure. I was trying to keep the fn name the same as the one it restructured from. But I think `AddController` is better

.github/workflows/presubmit.yaml
yuwenma: The KCC_USE_DIRECT_RECONCILERS flag shall still work (which gives developers the testing approach during the resource conversion). For release, we can purely rely on the CRD annotations to tell whether a direct or TF controller shall manage a resource kind (and eventually direct will be the default so it does not need a annotation).

pkg/controller/direct/directbase/directbase_controller.go
yuwenma: Sg. Thanks!

pkg/controller/registration/registration_controller.go
yuwenma: I don't have a preference. This is just to keep the same code style as below (line 232 and line 241).

yuwenma: so as the function name "Add" vs "AddController"

pkg/controller/direct/directbase/directbase_controller.go
yuwenma: Done.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_apikeyskeys.apikeys.cnrm.cloud.google.com.yaml
yuwenma: I rethink about this. I think we can be more "direct aggressive" here (and we should be good if we do):
- no CRD labels. By default we expect a direct controller to reconcile the resource. To use TF/DCL controller, the CRD has to have the corresponding label.
- A direct controller needs to be registered under direct dir. Otherwise, without non-label CRD will return error on running time. WDYT?

scripts/github-actions/tests-e2e-direct.sh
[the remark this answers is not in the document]
yuwenma: I plan to turn on the stricter golden check on the generic e2e fixtures (that would be in a separate PR). Right now, I put back the direct e2e path.

pkg/controller/registration/registration_controller.go
```diff
@@ -196,44 +192,17 @@ func registerDefaultController(r *ReconcileRegistration, config *controller.Conf
 		JitterGen:    r.jitterGenerator,
 		Defaulters:   r.defaulters,
 	}
-
 	var schemaUpdater k8s.SchemaReferenceUpdater
 	if kccfeatureflags.UseDirectReconciler(gvk.GroupKind()) {
```
yuwenma, line 196: Just a note for reviewers: The previous code only turn on LLM in direct controller, all other resources that already have direct controller and API are only turning on via the `KCC_USE_DIRECT_RECONCILERS` flag (and those part is not covered by the e2e tests). 
When those resources are already, we can simply remove the `"cnrm.cloud.google.com/tf2crd=true"`  label from their CRD.

cc @justinsb @acpana

pkg/controller/direct/alloydb/cluster_controller.go
```diff
@@ -28,30 +28,23 @@ import (
...
 
-	// TODO: Share gcp client (any value in doing so)?
-	ctx := context.TODO()
-	gcpClient, err := newGCPClient(ctx, config)
-	if err != nil {
-		return err
-	}
-	m := &clusterModel{gcpClient: gcpClient}
-	return directbase.Add(mgr, gvk, m, opts)
+func NewModel(config *controller.Config) directbase.Model {
```
justinsb, line 41: Nit: We'll probably end up naming this NewClusterModel (or newClusterModel).  But not a blocker.

pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -53,22 +55,81 @@ import (
...
 		return err
 	}
 	return add(mgr, reconciler)
 }
 
-// NewReconciler returns a new reconcile.Reconciler.
-func NewReconciler(mgr manager.Manager, immediateReconcileRequests chan event.GenericEvent, resourceWatcherRoutines *semaphore.Weighted,
-	gvk schema.GroupVersionKind, model Model, jg jitter.Generator) (*DirectReconciler, error) {
+func (c *directControllerBuilder) IsDirectByCRD(crd *apiextensions.CustomResourceDefinition) bool {
+	for gvk, _ := range c.modelMapper {
```
justinsb, line 87: I think we can just call IsDirectByGK here - from inside IsDirectByCRD (the version shouldn't matter)

pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -53,22 +55,81 @@ import (
...
+				}
+			}
+
+		}
+	}
+	return false
+}
+
+func (c *directControllerBuilder) IsDirectByGK(gk schema.GroupKind) bool {
+	if c.modelMapper == nil {
```
justinsb, line 101: FYI, you can `range` over a nil map; it is treated like an empty map.  (You can also read from it, you just can't write to it)

pkg/controller/registration/registration_controller.go
justinsb: Cool!

pkg/test/controller/reconciler/testreconciler.go
```diff
@@ -240,15 +238,8 @@ func (r *TestReconciler) newReconcilerForCRD(crd *apiextensions.CustomResourceDe
...
-		switch crd.GetName() {
-		case "logginglogmetrics.logging.cnrm.cloud.google.com":
-			m, err := logging.GetModel(context.TODO(), &controller.Config{HTTPClient: r.httpClient})
-			if err != nil {
-				return nil, fmt.Errorf("error getting logging model: %w", err)
-			}
-
-			return directbase.NewReconciler(r.mgr, immediateReconcileRequests, resourceWatcherRoutines, loggingapis.LoggingLogMetricGVK, m, jg)
+		if directbase.ControllerBuilder.IsDirectByCRD(crd) {
+			return directbase.ControllerBuilder.NewReconciler(r.mgr, &kcccontroller.Config{HTTPClient: r.httpClient}, immediateReconcileRequests, resourceWatcherRoutines, crd.GroupVersionKind(), jg)
```
justinsb, line 242: Nit: IIRC crd.GroupVersionKind() is problematic, ideally ControllerBuilder would record the GVK

justinsb:
Thanks @yuwenma 

/approve
/lgtm

pkg/controller/direct/alloydb/cluster_controller.go
yuwenma: I like the private method name.

pkg/controller/direct/directbase/directbase_controller.go
yuwenma: Sure. Removed `IsDirectByCRD` and `IsDirectByGVK`, and only keep `IsDirectByGK`

pkg/controller/direct/directbase/directbase_controller.go
yuwenma: oh, thanks! I always check the map nil first and never realized it can be iterated. 🤣

PR #1913, opened 2024-05-30
Getter as a separate stage

barney-s:
### Change description
Move getter to a separate stage
- Remove inline ValuesFrom from Compositions
  - Remove `ValuesFrom` from the composition spec
  - Remove code that waits for and fetches values from the expander
    reconciler (Expander_fetcher.go)
  - Change e2e test to refelect this change
- Add a new CRD `Getter` that acts as a Getter configuration
  - Add validation for this CRD
- Proto changes
  - GRPC `Evaluate` returns Values or manifests.
  - WAIT as a return status
- New changes in expander reconciler
  - Accumulate values so that they are presented to all subsequent
    stages
  - Get Expander config by reference and pass it to Evaluate
  - Get expander config GVK by looking up the expander version
  - When expander returns WAIT retry sooner in 5s
- ExpanderVersion changes
  - Add GVK to bind an Expander with its Config CRD
- Expander GRPC Service changes
  - Receive the whole Getter config object as Config
  - Introduce WAIT as a return value when object or fieldpath doesnt
    exist

TODO:
- Check Getter CR validation error in expander reconciler or Getter grpc
  service
- Move Samples to use the new getter
- Move from a projection based value extraction to whole object getter
- Exp backoff for WAIT state ?


experiments/compositions/composition/api/v1alpha1/expanderversion_types.go
```diff
@@ -53,6 +59,9 @@ type ExpanderVersionSpec struct {
 	// +kubebuilder:validation:Enum=job;grpc
 	// +kubebuilder:default=job
 	Type ExpanderType `json:"type"`
+
+	// ExpanderConfig GVK
+	Config ExpanderConfigGVK `json:"config,omitempty"`
```
cheftako, line 64: Don't we need a name and namespace to point at an instance of the configuration?

experiments/compositions/composition/api/v1alpha1/getter_types.go
```diff
@@ -0,0 +1,114 @@
...
+
+type FieldRef struct {
+	Path string `json:"path"`
+	As   string `json:"as"`
+}
+
+type ValuesFrom struct {
+	Name        string      `json:"name"`
+	ResourceRef ResourceRef `json:"resourceRef"`
+	FieldRef    []FieldRef  `json:"fieldRef"`
```
cheftako, line 48: Do we need something like preconditions? (Eg status == ready?)

experiments/compositions/composition/api/v1alpha1/getter_types.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// GetterSpec defines the desired state of Getter
+type GetterSpec struct {
+	ValuesFrom []ValuesFrom `json:"valuesFrom,omitempty"`
```
cheftako, line 53: Not sure we need it yet but I do think that something like a timeout (not permanent we're not a web-hook, but for initial declaring of error) is something we will want.

experiments/compositions/composition/cmd/main.go
```diff
@@ -131,6 +131,13 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "ExpanderVersion")
 		os.Exit(1)
 	}
+	if err = (&controller.GetterReconciler{
```
cheftako, line 134: This feels a bit off. Deal with it for now but shouldn't the instanced "expander" reconcilers be triggered by the ExpanderVersionReconciler? If thats not true then how do we dynamically add a expander/plugin at runtime?

experiments/compositions/composition/api/v1alpha1/expanderversion_types.go
cheftako: NM missed this was metadata not individual expander config.

cheftako:
I think this gets the Getter expander working. I think we should soon start thinking about how we get a pluggable expander working. Ideally the template and getter based expanders would work the same way.

experiments/compositions/composition/api/v1alpha1/expanderversion_types.go
```diff
@@ -36,6 +36,12 @@ const (
 	ExpanderTypeGRPC ExpanderType = "grpc"
 )
 
+type ExpanderConfigGVK struct {
+	Group   string `json:"group"`
+	Version string `json:"version"`
+	Kind    string `json:"kind"`
```
xiaoweim, line 42: The Kind here is similar to expander type? like Getter/Jinja?

experiments/compositions/composition/cmd/main.go
barney-s: You are right. This is not required. Its only for CR validation. Validation can be done as part of CRD definition.

experiments/compositions/composition/api/v1alpha1/getter_types.go
barney-s: You mean, Instead of being stuck in WAITING mode we declare ERROR (due to timeout) ?

experiments/compositions/composition/api/v1alpha1/getter_types.go
barney-s: Yes we do. Need to add it (separate CL ? )

experiments/compositions/composition/api/v1alpha1/getter_types.go
```diff
@@ -0,0 +1,114 @@
...
+	// OPTION 1
+	// <Kind>.<group>/<version>/<namespace>/<name>
+	// resource: ServiceIdentity.serviceusage.cnrm.cloud.google.com/v1beta1//sqladmin.googleapis.com
+
+	// OPTION 2
+	Group      string `json:"group,omitempty"`
+	Version    string `json:"version,omitempty"`
+	Resource   string `json:"resource"`
+	Kind       string `json:"kind"`
+	Name       string `json:"name,omitempty"`
```
barney-s, line 36: one of comment.

experiments/compositions/composition/api/v1alpha1/expanderversion_types.go
barney-s: No this is GVK GVR kubernetes resource, kind

experiments/compositions/composition/config/crd/bases/composition.google.com_getters.yaml
```diff
@@ -0,0 +1,176 @@
...
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.14.0
+  name: getters.composition.google.com
+spec:
+  group: composition.google.com
+  names:
+    kind: Getter
```
cheftako, line 25: Would this be better named GetterConfiguration?

barney-s: Agreed. GetterConfig ?

PR #1919, opened 2024-05-31
tests: Don't ignore annotations with direct reconciler

justinsb:
We should update the golden data, rather than ignoring it.

### DIFF
--- a/pkg/test/utils.go
+++ b/pkg/test/utils.go
@@ -95,16 +95,16 @@ func CompareGoldenFile(t *testing.T, p string, got string, normalizers ...func(s
 		return
 	}
 
+	if diff := cmp.Diff(want, got); diff != "" {
+		t.Errorf("unexpected diff in %s: %s", p, diff)
+	}
+
 	if writeGoldenOutput {
 		// Write the output to the golden file
 		if err := os.WriteFile(p, []byte(got), 0644); err != nil {
 			t.Fatalf("failed to write golden output %s: %v", p, err)
 		}
-		t.Errorf("wrote updated golden output to %s", p)
-	} else {
-		if diff := cmp.Diff(want, got); diff != "" {
-			t.Errorf("unexpected diff in %s: %s", p, diff)
-		}
+		t.Logf("wrote updated golden output to %s", p)
 	}
 }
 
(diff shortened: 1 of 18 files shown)


pkg/test/utils.go
```diff
@@ -174,19 +174,19 @@ func CompareGoldenObject(t *testing.T, p string, got []byte) {
...
-		t.Errorf("wrote updated golden output to %s", p)
-	} else {
-		t.Errorf("unexpected diff in %s: %s", p, diff)
+		t.Logf("wrote updated golden output to %s", p)
 	}
 }
 
 func postWriteNormalization(wantMap map[string]interface{}) map[string]interface{} {
-	if os.Getenv("KCC_USE_DIRECT_RECONCILERS") != "" {
+	if os.Getenv("XXX_KCC_USE_DIRECT_RECONCILERS") != "" {
```
yuwenma, line 189: is this intended?

justinsb: Yes, it's a quick hack while this is WIP!

justinsb: That said it was confusing, so I just deleted it instead of hacking-it-out

acpana:
Yes I think this should work well now that the LoggingLogMetric resource is a direct controller based one. For resource conversions some normalization might still be needed.

pkg/test/utils.go
```diff
@@ -169,50 +169,18 @@ func CompareGoldenObject(t *testing.T, p string, got []byte) {
...
 		if err := os.WriteFile(p, []byte(got), 0644); err != nil {
 			t.Fatalf("failed to write golden output %s: %v", p, err)
 		}
-		t.Errorf("wrote updated golden output to %s", p)
-	} else {
-		t.Errorf("unexpected diff in %s: %s", p, diff)
-	}
-}
-
-func postWriteNormalization(wantMap map[string]interface{}) map[string]interface{} {
```
yuwenma, line 188: +1 to not override the test expectations.
What is the current LLM behavior if user specify those two annotations? Is it a no-op or we have some handling in the direct controller(I don't recall I see any, just to double check in case I miss anything)?

I recall in SIS, we decide to add some warnings if user specify the unexpected annotations.

scripts/github-actions/tests-e2e-direct.sh
```diff
@@ -24,24 +24,23 @@ cd ${REPO_ROOT}/
 echo "Downloading envtest assets..."
 export KUBEBUILDER_ASSETS=$(go run sigs.k8s.io/controller-runtime/tools/setup-envtest@latest use -p path)
 
+#export KCC_USE_DIRECT_RECONCILERS=LoggingLogMetric
```
yuwenma, line 27: cleanup?

scripts/github-actions/tests-e2e-direct.sh
```diff
@@ -24,24 +24,23 @@ cd ${REPO_ROOT}/
...
+
 echo "Running e2e tests samples for LoggingLogMetric direct reconciliation..."
 
-KCC_USE_DIRECT_RECONCILERS=LoggingLogMetric \
```
yuwenma, line 29: ah! you bit me. I fix this in my WIP PR.

tests/e2e/testdata/scenarios/fields/llm_custom_diff/value_type/label/_http00.log
```diff
@@ -255,74 +171,18 @@ X-Xss-Protection: 0
     "labels": [
       {
         "description": "a label description",
-        "key": "testkey"
```
yuwenma, line 258: Is the previous log from the real GCP service? From the deleted part, I think we want to modify the mock llm to use operation metadata to reflect the real http as much as possible. May be a good follow-up task for @acpana ?

yuwenma:
/lgtm
/approve

Thank you!

pkg/test/utils.go
justinsb: It's a no-op, I believe, but @acpana can confirm

scripts/github-actions/tests-e2e-direct.sh
justinsb: Will do!

scripts/github-actions/tests-e2e-direct.sh
justinsb: Happy to merge yours first if you prefer!

tests/e2e/testdata/scenarios/fields/llm_custom_diff/value_type/label/_http00.log
justinsb: I think the mock for LoggingLogMetric is pretty good, but agree we should ensure that we do have a recording with real GCP.  This is currently recorded with the mock

PR #1922, opened 2024-06-01
ci: run some of the validations on experiments too

acpana:
Some of the validations should be run on experiments/compositions too arguably. Currently there seems to be a formatting issue that's breaking HEAD.

### DIFF
--- a/.github/workflows/presubmit-compositions.yaml
+++ b/.github/workflows/presubmit-compositions.yaml
@@ -24,6 +24,24 @@ on:
     paths:
       - "experiments/**"
 jobs:
+  validations:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          # This is to get all the commits in order to validate them
+          fetch-depth: 0
+      - uses: actions/setup-go@v4
+        with:
+          go-version-file: 'go.mod'
+      - name: "Run validations"
+        run: |
+          ./scripts/github-actions/ga-validation.sh
+        env:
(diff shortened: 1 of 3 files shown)


.github/workflows/presubmit-compositions.yaml
```diff
@@ -24,6 +24,24 @@ on:
...
+      - uses: actions/checkout@v3
+        with:
+          # This is to get all the commits in order to validate them
+          fetch-depth: 0
+      - uses: actions/setup-go@v4
+        with:
+          go-version-file: 'go.mod'
+      - name: "Run validations"
+        run: |
+          ./scripts/github-actions/ga-validation.sh
```
acpana, line 40: I can pair this down.

barney-s:
https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1921 this should fix the breakage

justinsb:
I think we're thinking about experiments as a separate repo that just happens to temporarily be colocated with the "main" KCC repo.  Once they are no longer "experiments", we will promote them to their own repositories.

From that lens, I think we want experiments to have its own tests.  The issue here I believe is that we are checking formatting of experiments in the main CI.  I sent #1929 to split them out.

acpana:
> From that lens, I think we want experiments to have its own tests. The issue here I believe is that we are checking formatting of experiments in the main CI. I sent #1929 to split them out.

That is fair, thanks for splitting it up 🙏🏼 

Maybe I can add a simple validator/ lint as a follow-on PR that is scoped only to `/experiments` .

PR #1920, opened 2024-05-31
Add direct resource `cloudbuildworkerpool`

yuwenma:
### Change description

Fixes #

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #1924, opened 2024-06-02
tests: Don't ignore http log with direct reconcilers

justinsb:


tests/e2e/script_test.go
```diff
@@ -267,9 +267,10 @@ func TestE2EScript(t *testing.T) {
 				}
 
 				if os.Getenv("GOLDEN_REQUEST_CHECKS") != "" || os.Getenv("WRITE_GOLDEN_OUTPUT") != "" {
-					if os.Getenv("KCC_USE_DIRECT_RECONCILERS") != "" && os.Getenv("GOLDEN_REQUEST_CHECKS") != "" {
-						h.Logf("ignoring http log checking for now for direct resources")
-					} else {
+					// if os.Getenv("KCC_USE_DIRECT_RECONCILERS") != "" && os.Getenv("GOLDEN_REQUEST_CHECKS") != "" {
```
acpana, line 270: I think for LoggingLogMetric we can exclude this check and the logs _should_ pass now that the logs have been written with the `direct` controller.

For resource conversions, it may be hard still to compare a golden log of a direct controller vs a TF/ DCL one. But maybe we can flip this around to explictly not check resources during conversion.

justinsb: The idea is that we want to see that http.log change, and review it alongside the controller change to direct actuation.  _Ideally_ the changes would be small!  But the hope is that http.log will make it easy to see any changes, and validate that we are comfortable with them.

yuwenma: I hit this log-skip in #1920. The problem I have is that I'm adding a new resource which 1) has a dependency to network service. 2) does not have a TF based controller before.
This check skips the `fixture/cloudbuildworkerpool` test due to the service dependency. I tried remove the commend like this PR does and it resolves the problem.
     
+1 to @justinsb that we can review the http log change along the controller change.

justinsb:
I think the flakes are due to races in the SLEEP approach, which was always intended as a temporary hack to be fixed in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1800.  I updated that PR.

yuwenma:
/lgtm
/approve

With this fix in, I can rebase mine! Thanks for both fixes!

PR #1925, opened 2024-06-02
LoggingLogMetric - kind should not be require in projectRef

justinsb:
This regression was introduced because we reused the ResourceRef type.

### DIFF
--- a/apis/refs/v1beta1/projectref.go
+++ b/apis/refs/v1beta1/projectref.go
@@ -14,11 +14,15 @@
 
 package v1beta1
 
+// The Project that this resource belongs to.
 type ProjectRef struct {
 	/* The `projectID` field of a project, when not managed by KCC. */
 	External string `json:"external,omitempty"`
 	/* The `name` field of a `Project` resource. */
 	Name string `json:"name,omitempty"`
-	/* The `namespcae` field of a `Project` resource. */
+	/* The `namespace` field of a `Project` resource. */
 	Namespace string `json:"namespace,omitempty"`
+	// The kind of the Project resource; optional but must be `Project` if provided.
+	// +optional
+	Kind string `json:"kind,omitempty"`
 }
--- a/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_logginglogmetrics.logging.cnrm.cloud.google.com.yaml
+++ b/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_logginglogmetrics.logging.cnrm.cloud.google.com.yaml
@@ -324,29 +324,28 @@ spec:
                     - external
                   required:
                   - name
-                  - kind
                 - not:
                     anyOf:
                     - required:
                       - name
                     - required:
                       - namespace
-                    - required:
-                      - kind
                   required:
                   - external
                 properties:
                   external:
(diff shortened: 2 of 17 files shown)


scripts/add-validation-to-crds/parse-crds.go
```diff
@@ -223,6 +240,11 @@ func addValidationToRefs(props *apiextensions.JSONSchemaProps) error {
 		ruleYAML = refRuleWithKind
 	} else if signature == "external,kind,name,namespace" {
 		ruleYAML = refRuleWithKind
+		// kind is optional for projectRef (and maybe in future other well-known ref types)
+		// fieldPath is the best mechanism we have today (?)
```
jingyih, line 244: IIUC, when we migrate a resource to direct-based, we need to mark kind as "optional" for all its reference fields?

apis/refs/v1beta1/projectref.go
```diff
@@ -14,11 +14,15 @@
...
 	/* The `projectID` field of a project, when not managed by KCC. */
 	External string `json:"external,omitempty"`
 	/* The `name` field of a `Project` resource. */
 	Name string `json:"name,omitempty"`
-	/* The `namespcae` field of a `Project` resource. */
+	/* The `namespace` field of a `Project` resource. */
 	Namespace string `json:"namespace,omitempty"`
+	// The kind of the Project resource; optional but must be `Project` if provided.
+	// +optional
+	Kind string `json:"kind,omitempty"`
```
yuwenma, line 27: I guess this is for backward compatibility, right? I think we accept breaking changes for bug fix purposes.

scripts/add-validation-to-crds/parse-crds.go
justinsb: It's not really about direct or not, it's about for project references, where we (briefly) introduced kind, should we continue to support the kind field?

I think if we drop kind, we would break anyone that added it in 1.118.1.  That should not be many people, but it is _some_ people.

`kind` optional and defaulting to Project (and only Project is allowed) shouldn't break anyone, though it is inelegant.

justinsb:
I had to rebase (and regenerate the docs), so I've lost LGTM

acpana:
/lgtm
/approve

thanks for this change 🥇  and regenerating the docs 📚

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_logginglogmetrics.logging.cnrm.cloud.google.com.yaml
```diff
@@ -324,29 +324,28 @@ spec:
...
                     - required:
                       - namespace
-                    - required:
-                      - kind
                   required:
                   - external
                 properties:
                   external:
-                    description: The external name of the referenced resource
+                    description: The `projectID` field of a project, when not managed
```
maqiuyujoyce, line 337: We should not mention "KCC" in our CRD: this is not the approved branding name we should use.
Also, the field name is not `projectID` but `projectId` (in the API or gcloud) or `ID` (in the UI).

I think we should add checks in the CRD field description to avoid the mentioning of `KCC`.

PR #1928, opened 2024-06-03
chore: generate protos using a script

justinsb:
This should enable more reusability, as well as fixing the CI breakage at head.


experiments/compositions/composition/dev/tasks/generate-proto
```diff
@@ -0,0 +1,44 @@
...
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+cd ${REPO_ROOT}/experiments/compositions/composition
+
+BUILDDIR=$(pwd)/.build/
+BINDIR=${BUILDDIR}/bin/
+mkdir -p ${BINDIR}
+
+# Download protoc
+wget -N -O ${BUILDDIR}/protoc.zip https://github.com/protocolbuffers/protobuf/releases/download/v27.0/protoc-27.0-linux-x86_64.zip
```
barney-s: Using `@latest` for Go tools can lead to non-reproducible builds, as a new version of the tool could be released at any time with breaking changes. It's better to pin to a specific version to ensure that the build is consistent over time. For example: `go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.34.1`

experiments/compositions/composition/dev/tasks/generate-proto
```diff
@@ -0,0 +1,44 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+cd ${REPO_ROOT}/experiments/compositions/composition
+
+BUILDDIR=$(pwd)/.build/
+BINDIR=${BUILDDIR}/bin/
+mkdir -p ${BINDIR}
```
barney-s: The script downloads `protoc` using `wget` but does not remove the downloaded `.zip` file. While it's in a `.build` directory, it's good practice to clean up temporary files after they are no longer needed. Consider adding `rm ${BUILDDIR}/protoc.zip` after the `unzip` command.

experiments/compositions/composition/dev/tasks/generate-proto
```diff
@@ -0,0 +1,44 @@
...
+
```
barney-s: The `unzip` command could be made safer by adding the `-q` flag for quiet output, and it's good practice to check the exit code of the command to ensure it was successful. Although `set -o errexit` is used, explicit checks can sometimes be clearer. Also, `unzip` can be a source of security vulnerabilities if the archive is malicious. While the source is trusted here, it's a point to be aware of.

experiments/compositions/composition/dev/tasks/generate-proto
barney-s: `wget` is used here, but it's generally better to use `curl -fsSL` for scripts as it's more commonly available and has more robust error handling options.

experiments/compositions/composition/dev/tasks/generate-proto
```diff
@@ -0,0 +1,44 @@
...
+# Download protoc
+wget -N -O ${BUILDDIR}/protoc.zip https://github.com/protocolbuffers/protobuf/releases/download/v27.0/protoc-27.0-linux-x86_64.zip
+rm -f ${BINDIR}/protoc
+unzip -j -d ${BINDIR}/ ${BUILDDIR}/protoc.zip bin/protoc 
+
+${BINDIR}/protoc --version
```
barney-s: The script correctly sets the `PATH` for the `protoc` command. This is a good practice as it ensures that the locally downloaded `protoc` is used, rather than a system-installed one.

experiments/compositions/composition/expanders.mk
```diff
@@ -4,18 +4,9 @@
...
 ##@ Protos
 
-PROTOC ?= /usr/bin/protoc
-.PHONY: protoc
-protoc: $(PROTOC) ## Download protoc if necessary
-$(PROTOC):
-	sudo apt install -y protobuf-compiler
-	protoc --version
-	go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
-	go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
```
barney-s: The removal of the `sudo apt install` command is a significant improvement. It makes the build process more hermetic and removes the need for root privileges, which is a security best practice.

justinsb:
This was in the compositions experiment, I think we can now close

/close

PR #1932, opened 2024-06-04
fix:test: managedsslcertificate

acpana:
I updated this test to pass schema validations and reduced it to a basic form where we just create a `ComputeManagedSSLCertificate`.


pkg/test/resourcefixture/testdata/basic/compute/v1alpha1/computemanagedsslcertificate/managedsslcertificatebasic/dependencies.yaml
acpana, line 1: This test was not actually in need of these dependencies. It would still be useful to have a "less basic" test and I think I can add that as a follow on either as a fixture or scenario test.

pkg/test/resourcefixture/testdata/basic/compute/v1alpha1/computemanagedsslcertificate/managedsslcertificate/create.yaml
acpana, line 1: Given the test modifications, I think it's fair to no longer have this classified as an autogen test!

pkg/test/resourcefixture/testdata/basic/compute/v1alpha1/computemanagedsslcertificate/managedsslcertificatebasic/dependencies.yaml
yuwenma: Make sense

pkg/test/resourcefixture/testdata/basic/compute/v1alpha1/computemanagedsslcertificate/managedsslcertificate/create.yaml
```diff
@@ -11,15 +11,14 @@
...
 kind: ComputeManagedSSLCertificate
 metadata:
-  name: computemanagedsslcertificate-${uniqueId}
+  name: computemanagedsslcertificate${uniqueId}
 spec:
   managed:
-  - domains:
-    - sslcert.tf-test.club.
+    domains:
+    - sslcert.kcc-test.club.
```
yuwenma, line 21: can you change this to `direct` since we try to avoid exposing these codenames (I know... we refer them a lot in comments and PR descriptions. But we want to not expose them in the actual codebase.

pkg/test/resourcefixture/testdata/basic/compute/v1alpha1/computemanagedsslcertificate/managedsslcertificate/create.yaml
```diff
@@ -11,15 +11,14 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 apiVersion: compute.cnrm.cloud.google.com/v1alpha1
 kind: ComputeManagedSSLCertificate
 metadata:
-  name: computemanagedsslcertificate-${uniqueId}
+  name: computemanagedsslcertificate${uniqueId}
```
yuwenma, line 17: nit: I think the [naming convention](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/README.NewResourceFromTerraform.md#naming-conventions-special-variables-and-constants-in-testdata) is  computemanagedsslcertificate-${uniqueId}.

yuwenma:
/lgtm
/approve

Super minor comments. You can resolve in follow-up PRs.

pkg/test/resourcefixture/testdata/basic/compute/v1alpha1/computemanagedsslcertificate/managedsslcertificate/create.yaml
acpana: In this case, I think the dash violated a regex for the ssl cert name since we are using `metadata.name` for that BUT I am not able to reproduce that violation anymore so maybe I was getting confused 🤔

PR #1934, opened 2024-06-04
[WIP] Turn on stricter golden log check in e2e fixture and samples.

yuwenma:
### Change description

Fixes #

### Tests you have done


justinsb:
So I don't know if all our tests are consistent with record-gcp yet.

But ... I like this idea ... we can at least require that compare-mock produces stable output, and verify it!

justinsb:
I've been thinking about this ... you're right, we should have done this from the start.  It might be that our mockgcp results aren't 100% equal to GCP in all cases.  But we still get a lot more signal by checking these values against mockgcp than by not checking them!

scripts/github-actions/tests-e2e-direct.sh
```diff
@@ -24,7 +24,7 @@ cd ${REPO_ROOT}/
 echo "Downloading envtest assets..."
 export KUBEBUILDER_ASSETS=$(go run sigs.k8s.io/controller-runtime/tools/setup-envtest@latest use -p path)
 
-export KCC_USE_DIRECT_RECONCILERS=MonitoringDashboard
+KCC_USE_DIRECT_RECONCILERS=MonitoringDashboard
```
justinsb, line 27: I think we need export, otherwise it doesn't apply to subsequent requests, but as this is passing tests let's go with it!

yuwenma: I fix this in Part 2. Want to use this as an example to highlight some changes.

PR #1936, opened 2024-06-04
mockgcp: serve multiple hosts, including variables

justinsb:


yuwenma:
/lgtm
/approve

Is this related to the ComputeForwardingRule?  cc @gemmahou

justinsb:
Actually, it's more about regional endpoints like

```
func (s *MockService) ExpectedHosts() []string {
	return []string{"{region}-gkemulticloud.googleapis.com"}
}
```

But it will also let us support SecureSourceRepository, which listens on multiple host names (one for the control-plane API, a different one for the data-plane API) .

PR #1937, opened 2024-06-04
feat: create go types for SecureSourceManagerInstance

justinsb:
Generated from the proto by dev/tools/proto-to-mapper and then lightly
edited / moved around.


apis/securesourcemanager/v1alpha1/instance.go
```diff
@@ -0,0 +1,126 @@
...
+
+	/* Immutable. Optional. The name of the resource. Used for creation and acquisition. When unset, the value of `metadata.name` is used as the default. */
+	// +optional
+	ResourceID *string `json:"resourceID,omitempty"`
+
+	// Optional. Labels as key value pairs.
+	Labels map[string]string `json:"labels,omitempty"`
+
+	// Optional. Immutable. Customer-managed encryption key name, in the format
+	//  projects/*/locations/*/keyRings/*/cryptoKeys/*.
```
yuwenma, line 79: Shall this be a ref? What I'm actually trying to ask is if we shall validate the existence of this kmsKey

pkg/controller/direct/securesourcemanager/mapper.generated.go
yuwenma, line 1: What do you think about moving the auto-generated mapper under apis/<resource>/<version>/?  
It would be easier for generation, validation and version management.

pkg/controller/direct/securesourcemanager/roundtrip_test.go
```diff
@@ -0,0 +1,412 @@
...
+	"github.com/google/go-cmp/cmp"
+	"google.golang.org/protobuf/encoding/prototext"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/testing/protocmp"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
+)
+
+func FuzzSecureSourceManagerInstanceSpec(f *testing.F) {
```
yuwenma, line 33: @jingyih ^^ fuzz detecting resource field .

yuwenma (commented):
/lgtm
/approve
/hold  looks really great! One topic to discussion. Feel free to cancel the hold

apis/securesourcemanager/v1alpha1/instance.go
justinsb: Yes, it should be a ref!  It's alpha so we can still change it, but ... you're right.  I'll give it a go!

pkg/controller/direct/securesourcemanager/mapper.generated.go
justinsb: I like that idea!

justinsb:
I'm going to cancel the hold here because I have a cunning plan :-)

/hold cancel

pkg/controller/direct/securesourcemanager/roundtrip_test.go
jingyih: Cool. Thanks!

PR #1946, opened 2024-06-04
Use observedState to capture important output-only spec fields in ContainerCluster

maqiuyujoyce:
### Change description

Supports `status.observedState.masterAuth.clusterCaCertificate`, `status.observedState.privateClusterConfig.privateEndpoint`, and `status.observedState.privateClusterConfig.publicEndpoint`.

Also fixed the observed state CRD generation logic and population logic to allow more than one subfield in an object-typed field.

### Tests you have done


gemmahou:
Looks good to me. Do we want to capture golden object log to verify the output fields are populated correctly under obeservedState?

yuwenma:
/lgtm
/approve
/hold

Two quick questions:
1. Do we want to display the plain `clusterCaCertificate` in the status? 
2. Is there a dedicated requirement to expose these fields? 

If both answers are yes, please feel free to unhold and let the PR goes int.

+1 to @gemmahou comments on updating the golden log, I think you can do that in follow up PRs.

maqiuyujoyce:
> 1. Do we want to display the plain clusterCaCertificate in the status?

I think it should be acceptable because this field is also plain text when we do `gcloud container clusters describe`. Do you have any concern?

> 2. Is there a dedicated requirement to expose these fields?

Yes.

maqiuyujoyce:
> I think it should be acceptable because this field is also plain text when we do gcloud container clusters describe. Do you have any concern?

Discussed offline and agreed that as gcloud exposes the field in plain text so it should be safe for us to do it as well.

> +1 to @gemmahou comments on updating the golden log, I think you can do that in follow up PRs.

Yep, I'm working on that.

/unhold

PR #1947, opened 2024-06-05
tests: replace tagKeyId, update test data

justinsb:


yuwenma:
I think we no longer this PR since the golden object has been fixed in master.

PR #1953, opened 2024-06-05
tests: add CRD linter to detect unstylish acronym capitalization

justinsb:


tests/apichecks/crds_test.go
```diff
@@ -45,6 +49,107 @@ func TestCRDsDoNotHaveFooUrlRef(t *testing.T) {
...
+				}
+			})
+		}
+	}
+
+	sort.Strings(errs)
+
+	want := strings.Join(errs, "\n")
+
+	test.CompareGoldenFile(t, "errata/acronyms.txt", want)
```
yuwenma, line 130: Cool example! I would argue to rename it to testdata/...  (what does errata mean?)

justinsb: So it used to be that fixing mistakes in a printed book was expensive, so they would instead include a sheet with significant errors listed.   I found this fun picture: https://analoggamestudies.org/wordpress/wp-content/uploads/2014/11/lit_errata.jpg

Now it's mostly used to highlight changes in things like chip specifications, it's sort of like a changelog except the changes are to the description rather than to the chip itself - it's infeasible to change the chip.  e.g. https://www.infineon.com/dgdl/Infineon-XMC4100_XMC4200_BA_Errata_Sheet-ES-v01_03-EN.pdf?fileId=5546d46259b0420a0159d50916760b89

How about testdata/exceptions ?

justinsb: (The argument _against_ testdata BTW is that although we're implementing this in a test, it could also be a linter CLI.  But I agree that errata is pretty obscure at this point, given that "document tabs" are now also considered obscure!)

yuwenma: I LOVE this name. in case you haven't changed. Let me get them in!

yuwenma: oh no... do you want to change it back😄 ? 

I'll leave it to you to decide.

PR #1954, opened 2024-06-05
upload multi-cluster leader election to the experiment folder.

600lyy:
### Change description

This is a feature that extends the in-cluster, built-in leader election supported by controller runtime.
Multi-cluster leader election aims to support leader election for manages (and operators) across multiple clusters.


jingyih:
/assign
Thank you for uploading the PR!

PR #1955, opened 2024-06-05
chore: Refactor registry to be more reusable

justinsb:
We need to move things around to avoid import cycles as we add IAM
support.

### DIFF
--- a/pkg/controller/direct/registry/registry.go
+++ b/pkg/controller/direct/registry/registry.go
@@ -1,0 +1,107 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package registry
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/config"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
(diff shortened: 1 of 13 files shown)


third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/fwtransport/version.txt
yuwenma, line 1: It's weird that this file shows up here. (Since it looks good, no need to clean it up)

pkg/controller/direct/registry/registry.go
```diff
@@ -0,0 +1,107 @@
...
+
+func GetModel(gk schema.GroupKind) (directbase.Model, error) {
+	registration := singleton.registrations[gk]
+	if registration == nil {
+		return nil, fmt.Errorf("no model registered for %s", gk)
+	}
+	return registration.model, nil
+}
+
+func PreferredGVK(gk schema.GroupKind) (schema.GroupVersionKind, bool) {
```
yuwenma, line 48: I like this name

pkg/controller/direct/registry/registry.go
```diff
@@ -0,0 +1,107 @@
...
+	switch groupKind {
+	case schema.GroupKind{Group: "privateca.cnrm.cloud.google.com", Kind: "PrivateCACAPool"}:
+		return true
+	}
+	return false
+}
+
+// SupportsIAM returns true if this resource supports IAM (not all GCP resources do).
+// An error will be returned if IsDirect(groupKind) is not true.
+func SupportsIAM(groupKind schema.GroupKind) (bool, error) {
```
yuwenma, line 100: I don't see where the `registry.SupportsIAM` is called. Mind giving me some pointers?

justinsb: In pkg/cli/cmd/bulkexport/singleresourceiamclient/singleresourceiamclient.go  Generally this refactor is in working towards #1877

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/fwtransport/version.txt
justinsb: I'll rebase once we merge #1956 so this goes away

pkg/controller/direct/registry/registry.go
yuwenma: Got it.

PR #1957, opened 2024-06-05
ci: install golang if not currently installed from ci tasks

justinsb:


dev/tasks/install-tools
```diff
@@ -0,0 +1,32 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+cd "${REPO_ROOT}"
+
+if [[ ! -x go ]]; then
+  echo "Must install go; assuming we are running in a container"
+  curl -L -o /tmp/go.tar.gz https://go.dev/dl/go1.22.4.linux-amd64.tar.gz
```
yuwenma, line 25: I used to hit a controller-gen problem in go1.22 https://github.com/kubernetes-sigs/controller-tools/issues/880
If installing golang, can we install go1.23 instead?

justinsb: I think 1.22.4 is the latest: https://go.dev/doc/install

I think we updated controller-gen so we should be good with 1.22?

yuwenma: you are absolutely right!

PR #1958, opened 2024-06-05
WIP: feature: implement "watch" for GCP objects

justinsb:
We can't actually watch, but we can instead track dependencies and
poll them.  By doing this we are able to reduce the number of GCP
requests when multiple KCC resources share a GCP resource and we save
a few kube requests in the common (unchanged) case.  We may also be
able to implement a list optimization in future.

This is really a PoC, it only implements a narrow sliver of
functionality.

### DIFF
--- a/pkg/gcpwatch/fetcher.go
+++ b/pkg/gcpwatch/fetcher.go
@@ -1,0 +1,109 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package gcpwatch
+
+import (
+	"context"
+	"encoding/base64"
+	"fmt"
+	"strings"
+
+	"cloud.google.com/go/iam/apiv1/iampb"
+	resourcemanager "cloud.google.com/go/resourcemanager/apiv3"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/config"
+	"google.golang.org/api/option"
+	"k8s.io/klog/v2"
+)
+
+type IAMFetcher struct {
+	crmClient *resourcemanager.ProjectsClient
+}
+
+var _ Fetcher = &IAMFetcher{}
+
+func NewIAMFetcher(ctx context.Context, config *config.ControllerConfig) (*IAMFetcher, error) {
+	opts, err := options(config)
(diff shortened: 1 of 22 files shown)


justinsb:
/assign @yuwenma 

FYI - looking at whether we can "watch" GCP

justinsb:
(Broke out some of this into #1962 because that seems like a good idea independently of this, and adds a lot of noise to this PR)

pkg/gcpwatch/fetcher.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	policy, err := f.crmClient.GetIamPolicy(ctx, req)
+	if err != nil {
+		return nil, fmt.Errorf("fetching iam policy for %q: %w", req.Resource, err)
+	}
+	log.Info("got iam policy for project", "policy", policy, "projectID", projectID)
+	etag := base64.StdEncoding.EncodeToString(policy.Etag)
+	return &ResourceInfo{
+		Etag: etag,
```
yuwenma, line 107: Oh, I like this!

pkg/gcpwatch/tracker.go
```diff
@@ -0,0 +1,237 @@
...
+	tracker := &DependencyTracker{
+		fetcher:      fetcher,
+		controllers:  make(map[string]*ControllerRegistration),
+		gcpResources: make(map[gcpResouceKey]*dependenciesByResource),
+	}
+	return tracker
+}
+
+func (t *DependencyTracker) PollForever(ctx context.Context, initialDelay time.Duration, minInterval time.Duration) {
+	nextPoll := time.Now().Add(initialDelay)
```
yuwenma, line 98: Why do we need a configurable initialDelay?

pkg/gcpwatch/tracker.go
```diff
@@ -0,0 +1,237 @@
...
+
+func (t *DependencyTracker) pollOnce(ctx context.Context) error {
+	log := klog.FromContext(ctx)
+
+	var keys []gcpResouceKey
+	t.mutex.Lock()
+	for key := range t.gcpResources {
+		keys = append(keys, key)
+	}
+	t.mutex.Unlock()
```
yuwenma, line 126: Is this like taking a snapshot?

pkg/gcpwatch/tracker.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+func (t *DependencyTracker) getTarget(kind string, external string) *dependenciesByResource {
+	t.mutex.Lock()
+	defer t.mutex.Unlock()
+
+	key := gcpResouceKey{
+		Kind:     kind,
+		External: external,
+	}
```
yuwenma, line 229: In current KCC, what happens when a reference changes from external to KNN? i.e. a ProjectRef used to not have CC Project object, and a LLM use ProjectRef.external. then it has a CC object. Will the LLM controller update the projectRef.name and projectRef.namespace with corresponding value? If so, if the CC project is changed, will that trigger the LLM object to reconcile?

yuwenma (commented):
This looks great!

ldanielmadariaga:

PR #1959, opened 2024-06-06
Handle custom diff in gkehubfeaturemembership controller

ziyue-101:
### Change description

In old DCL controller, it has logic to convert {} back to all false fields within HNC and treats no diffs between KRM spec and GCP state.

### Tests you have done


acpana:
obsolete; will keep an eye out if needs re-opening

PR #1961, opened 2024-06-06
tests: create simple API linter to try to detect refs

justinsb:
- **tests: add CRD linter to detect unstylish acronym capitalization**
- **tests: create simple API linter to try to detect refs**


justinsb:
Trying out the simple heuristic here ... "can we identify refs by looking for something that looks like a selfLink template in the description".  My first attempt with a very dumb way of matching that seems to give reasonable results.

Note that I also deliberately (for now) deleted the exception for SecureSourceManager, so we can see it come up in CI (hopefully)

justinsb:
(Also this includes the acronym linter PR, so that one should probably be merged first)

justinsb:
Hooray!

```
=== Failed
=== FAIL: tests/apichecks TestMissingRefs (1.51s)
    utils.go:99: unexpected diff in testdata/exceptions/missingrefs.txt:   (
          	"""
exit status 1
make: *** [Makefile:49: test] Error 1
          	... // 48 identical lines
          	[refs] crd=gkebackupbackupplans.gkebackup.cnrm.cloud.google.com version=v1alpha1: field ".spec.backupConfig.encryptionKey.gcpKmsEncryptionKey" should be a reference
          	[refs] crd=notebooksenvironments.notebooks.cnrm.cloud.google.com version=v1alpha1: field ".spec.vmImage.project" should be a reference
        + 	[refs] crd=securesourcemanagerinstances.securesourcemanager.cnrm.cloud.google.com version=v1alpha1: field ".spec.kmsKey" should be a reference
          	[refs] crd=tagstagkeys.tags.cnrm.cloud.google.com version=v1beta1: field ".spec.parent" should be a reference
          	[refs] crd=vertexaifeaturestoreentitytypefeatures.vertexai.cnrm.cloud.google.com version=v1alpha1: field ".spec.entitytype" should be a reference
          	... // 6 identical lines
          	"""
```

scripts/unit-test.sh
```diff
@@ -18,7 +18,7 @@ set -o pipefail
 REPO_ROOT="$(git rev-parse --show-toplevel)"
 cd "${REPO_ROOT}"
 make -C operator test
-UNIT_TEST_PACKAGES=$(go list ./pkg/... ./cmd/... ./config/tests/... ./scripts/generate-go-crd-clients/... ./scripts/resource-autogen/... | grep -v mocktests)
+UNIT_TEST_PACKAGES=$(go list ./pkg/... ./cmd/... ./config/tests/... ./scripts/generate-go-crd-clients/... ./scripts/resource-autogen/... ./tests/... | grep -v mocktests | grep -v tests/e2e | grep -v tests/stress)
```
acpana, line 21:
```suggestion
UNIT_TEST_PACKAGES=$(go list ./pkg/... ./cmd/... ./config/tests/... ./scripts/generate-go-crd-clients/... ./scripts/resource-autogen/... ./tests/... | grep -v mocktests | grep -v tests/e2e)
```

I don't think we have stress just yet committed right ?

tests/apichecks/testdata/exceptions/acronyms.txt
```diff
@@ -0,0 +1,508 @@
...
+[acronyms] crd=computeinstancetemplates.compute.cnrm.cloud.google.com version=v1beta1: field ".spec.networkInterface[].ipv6AccessConfig[].externalIpv6PrefixLength" should be ".spec.networkInterface[].ipv6AccessConfig[].externalIPV6PrefixLength"
+[acronyms] crd=computeinstancetemplates.compute.cnrm.cloud.google.com version=v1beta1: field ".spec.networkInterface[].networkIp" should be ".spec.networkInterface[].networkIP"
+[acronyms] crd=computeinterconnectattachments.compute.cnrm.cloud.google.com version=v1beta1: field ".status.cloudRouterIpAddress" should be ".status.cloudRouterIPAddress"
+[acronyms] crd=computeinterconnectattachments.compute.cnrm.cloud.google.com version=v1beta1: field ".status.customerRouterIpAddress" should be ".status.customerRouterIPAddress"
+[acronyms] crd=computeinterconnectattachments.compute.cnrm.cloud.google.com version=v1beta1: field ".status.googleReferenceId" should be ".status.googleReferenceID"
+[acronyms] crd=computemanagedsslcertificates.compute.cnrm.cloud.google.com version=v1alpha1: field ".status.certificateId" should be ".status.certificateID"
```
acpana, line 158: So are we meant to chip away at these exceptions as a team? For instance, would we want to correct this field if we had the chance?

acpana (commented):
this LGTM overall but I am going to give others a chance to chime in too.

scripts/unit-test.sh
justinsb: Oh lol, right :-)

tests/apichecks/testdata/exceptions/acronyms.txt
justinsb: For beta: I don't think we can do anything about them.

For alpha: I would say yes, but probably not a priority unless we're already touching the resource.

The big thing is that we want to avoid introducing new exceptions if we can.  So the idea is that if you're adding an exception, it's not possible to PR that without it showing up as a diff here (otherwise it fails tests).  So it should draw the PR reviewer's attention to it, and thus it makes it easy to have the discussion.

Specifically for `Id` vs `ID` I don't know.  KRM conventions say `ID`, but we have done `Id` pretty consistently.  We might argue that `Id` is more "GCP API", and thus more appropriate for KCC.

I'd say we don't need to do anything yet for the existing field exception list.  But as we touch the alpha resources, or as we introduce new resources, that's when we should think twice about any new exceptions (or any acronyms that are missing from the "API linter rule"!)

cc @yuwenma

yuwenma: I really like this way to guide the developers! Let's merge this in and turn on the check!

PR #1962, opened 2024-06-06
chore: Refactor ControllerConfig into its own package

justinsb:
This makes it more reusable, and means we can likely DRY the
calculation of the client.Options


yuwenma:
/lgtm
/approve

Agree

PR #1964, opened 2024-06-06
fix: better handling for http proto annotations

justinsb:
- **fix: better handling for http proto annotations**
- **chore: generate protos from OpenAPI**
- **chore: generate update protos/grpc**


acpana:
Verified with mockcloudids in #1901. Prior, the routing was not hitting the handler endpoints.

/lgtm
/approve

PR #1965, opened 2024-06-06
Bump client-side rate limit to 80. 

yuwenma:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/pkg/controller/ratelimiter/ratelimiter.go
+++ b/pkg/controller/ratelimiter/ratelimiter.go
@@ -18,6 +18,8 @@ import (
 	"time"
 
 	"golang.org/x/time/rate"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/util/flowcontrol"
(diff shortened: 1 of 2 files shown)


pkg/controller/ratelimiter/ratelimiter.go
```diff
@@ -69,3 +71,16 @@ func RequeueRateLimiter() ratelimiter.RateLimiter {
 		&workqueue.BucketRateLimiter{Limiter: rate.NewLimiter(rate.Limit(5), 50)},
 	)
 }
+
+// SetMasterRateLimiter sets the the kubernetes client level rate limiter.
```
justinsb, line 75: Nit: Probably something like `SetOverallRateLimiter` to avoid the "master" term, but not a blocker

justinsb:
Suggested a new name, but it's not customer visible so I don't think we need to block

/approve
/lgtm

PR #1966, opened 2024-06-06
chore: gen-proto requires gen-api

acpana:
One liner to have `gen-proto` depend on generating from the gapi api. 

I got bit by this when working on `mockcloudids` and forgot to run `gen-proto`.


yuwenma:
/lgtm
/approve

Thanks!

PR #1971, opened 2024-06-07
fix: remove duplicate method

justinsb:
Looks like two PRs both added the same Get method to Map.


acpana:
Ah, good catch!

/lgtm
/approve

PR #1974, opened 2024-06-07
chore: fix detection of go on path

justinsb:
The -x option was not working when go _was_ installed - I don't think
it does a PATH lookup.


yuwenma (commented):
/lgtm
/approve

PR #1976, opened 2024-06-07
test: Use log instead of error when no TestAcquire tests 

gemmahou:
### Change description

We don't run TestAcquire on ALL dynamic tests, it's expected that there were no tests to run. That should not be an error and make the TestAcquire test to fail.

### Tests you have done


gemmahou:
Thanks @yuwenma 

/unhold

PR #1979, opened 2024-06-08
chore: make mock test RemoveHeader be case insensitive

yuwenma:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/pkg/test/http_recorder.go
+++ b/pkg/test/http_recorder.go
@@ -239,11 +239,21 @@ func prettifyJSON(s string, mutators ...JSONMutator) string {
 }
 
 func (r *Response) RemoveHeader(key string) {
+	// The http.header `Del` converts the `key` to `CanonicalHeaderKey`, which means
+	// it expects the passed-in parameter `key` to be case insensitive, but `Header` itself should
+	// use canonical keys.
 	r.Header.Del(key)
+	// Delete non canonical header keys like `x-goog-api-client`.
+	delete(r.Header, strings.ToLower(key))
 }
 
 func (r *Request) RemoveHeader(key string) {
+	// The http.header `Del` converts the `key` to `CanonicalHeaderKey`, which means
+	// it expects the passed-in parameter `key` to be case insensitive, but `Header` itself should
+	// use canonical keys.
 	r.Header.Del(key)
+	// Delete non canonical header keys like `x-goog-api-client`.
+	delete(r.Header, strings.ToLower(key))
 }
 
 func (r *Response) ParseBody() map[string]any {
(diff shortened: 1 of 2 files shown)


pkg/test/resourcefixture/testdata/basic/cloudbuild/v1alpha1/cloudbuildworkerpool/_http.log
```diff
@@ -499,7 +499,6 @@ X-Xss-Protection: 0
 
 GET https://cloudbuild.googleapis.com/v1/projects/${projectId}/locations/us-central1/workerPools/cloudbuildworkerpool-${uniqueId}?%24alt=json%3Benum-encoding%3Dint
 Content-Type: application/json
-x-goog-api-client: gl-go/1.22.4 gapic/1.16.1 gax/2.12.3 rest/UNKNOWN
```
yuwenma, line 502: I hit this because my test env is `gl-go/1.23.0-20240419-RC02 gapic/1.16.1 gax/2.12.3 rest/UNKNOWN`

pkg/test/http_recorder.go
```diff
@@ -239,11 +239,21 @@ func prettifyJSON(s string, mutators ...JSONMutator) string {
 }
 
 func (r *Response) RemoveHeader(key string) {
+	// The http.header `Del` converts the `key` to `CanonicalHeaderKey`, which means
+	// it expects the passed-in parameter `key` to be case insensitive, but `Header` itself should
+	// use canonical keys.
 	r.Header.Del(key)
+	// Delete non canonical header keys like `x-goog-api-client`.
+	delete(r.Header, strings.ToLower(key))
```
justinsb, line 247: We _could_ also populate Header with the keys it expects, I.e. use `r.Header.Add` instead of `r.Header[k] = values`

The downside is that our golden tests would then have all the headers canonicalized to lower case, which could be confusing.

Maybe we should stop exporting `Header`, and just have some methods like RemoveHeader, ReplaceHeader etc?  From that perspective, I like what you're doing here!

justinsb:
Some suggestions as to future direction, but I agree this is the first step!

/approve
/lgtm

pkg/test/http_recorder.go
yuwenma: Exporting `header` let us aware of some interesting information, like there's a header about `terraform:kcc-manage` (I plan to change it BTW). 

For mocktesting, +1 that I don't think we really need it.

pkg/test/resourcefixture/testdata/basic/cloudbuild/v1alpha1/cloudbuildworkerpool/_http.log
yuwenma: We did have dropped this header.  But the problem is the `RemoveHeader` method only can drop the header in the capital-letters. This PR is to drop lower case like `x-goog-api-client` 
https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/tests/e2e/normalize.go#L270

PR #1986, opened 2024-06-10
fix: Update contributing guide for running locally

jasonvigil:
This fixes errors returned by the immutable fields validating webhook encountered when creating resources using a local controller manager:

```
annotation cnrm.cloud.google.com/state-into-spec is immutable
```

In order to pass the validating webhook, the local controller manager must use the cnrm-controller-manager service account credentials (instead of local user credentials) when making requests that modify resource annotations.

### Tests you have done

- [n/a] Perform necessary E2E testing for changed resources.


maqiuyujoyce:
/lgtm
/approve

Thank you so much for adding the instructions, @jasonvigil !

PR #1988, opened 2024-06-11
feat: Add rate-limit specification in custom ControllerResource (Part 2)

yuwenma:
We want to allow users to customize the client side rate limit configuration via ControllerResource


operator/pkg/apis/core/customize/v1beta1/controllerresource_types.go
```diff
@@ -56,6 +56,12 @@ type ContainerResourceSpec struct {
 	// Resources specifies the resource customization of this container.
 	// Required
 	Resources ResourceRequirements `json:"resources"`
+
+	// RateLimit describes the token bucket rate limit to the kubernetes client.
+	// Please note this rate limit is shared among all the Config Connector resources' requests.
+	// If not specified, the default will be Token Bucket with qps 20, burst 30.
+	// +optional
+	RateLimit *RateLimit `json:"rateLimit,omitempty"`
```
yuwenma, line 64: The Config Connector rate limit is actually in the container `arg` (See below) rather than some `rateLimit` field.  But I want to be careful on user configurable fields, specifically I don't want to expose the access of other container args in the specification. So I defined a new field here that will append/override the `args`.

jingyih:
Which controllers do we plan to support? Only the cnrm-controller-manager?

operator/config/crd/bases/customize.core.cnrm.cloud.google.com_controllerresources.yaml
```diff
@@ -173,6 +173,21 @@ spec:
                       - recorder
                       - unmanageddetector
                       type: string
+                    ratelimit:
```
justinsb, line 176: Nit: rateLimit (?)

operator/pkg/controllers/utils.go
```diff
@@ -278,6 +279,32 @@ func customizeContainerResourcesFn(cMap map[string]customizev1beta1.ResourceRequ
...
+			if rateLimit.QPS > 0 {
+				expectedArgs = append(expectedArgs, fmt.Sprintf("--qps=%d", rateLimit.QPS))
+			}
+			if rateLimit.Burst > 0 {
+				expectedArgs = append(expectedArgs, fmt.Sprintf("--burst=%d", rateLimit.Burst))
+			}
+			if found {
+				for _, arg := range actualArgs {
+					if strings.Contains(arg, "--qps") || strings.Contains(arg, "--burst") {
+						continue
```
justinsb, line 298: Nit: Add a log message here?

operator/pkg/controllers/utils.go
```diff
@@ -278,6 +279,32 @@ func customizeContainerResourcesFn(cMap map[string]customizev1beta1.ResourceRequ
...
+					}
+					expectedArgs = append(expectedArgs, arg)
+				}
+			}
+			if err := unstructured.SetNestedStringSlice(container, expectedArgs, "args"); err != nil {
+				return fmt.Errorf("error setting token bucket rate-limit request: %w", err)
```
justinsb, line 304: Nit: "setting container args" or something like that?

jingyih:
Discussed offline, we could start with a new API (CRD) with version `v1alpha1` to customize the client rate limits.

PR #1989, opened 2024-06-11
feat: Make the client-side rate limit configurable via flags (Part 1)

yuwenma:
We want to allow users to configure the client side rate limit via flags.

follow up PR https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1988


justinsb:
Thanks @yuwenma 

/approve
/lgtm

PR #1992, opened 2024-06-11
chore: install zip in CLI build

justinsb:
Verified by running in docker that this now works when run with the
gcr.io/cloud-builders/docker image.


yuwenma:
/lgtm
/approve

Thanks!

PR #1994, opened 2024-06-12
Add validator to return a warning when `state-into-spec` is set to `merge`

maqiuyujoyce:
### Change description

Added a validating webhook that checks whether the `state-into-spec` annotation is explicitly set to `merge` and returns a warning if so.

Further improvement would be validating whether the `state-into-spec` annotation will be defaulted to `merge` for a resource, and/or validating whether `stateIntoSpec` field is set to `Merge` for CC or CCC.

### Tests you have done

Ran `make docker-build docker-push deploy` to install a locally built version, and verified that the warning is returned when `cnrm.cloud.google.com/state-into-spec: merge` is set:

```
$ kubectl apply -f pubsubtopic-merge.yaml 
Warning: 'cnrm.cloud.google.com/state-into-spec: merge' is unsupported for CRDs added in 1.114.0 and later. Use 'absent' instead. More details can be found at https://cloud.google.com/config-connector/docs/concepts/ignore-unspecified-fields.
pubsubtopic.pubsub.cnrm.cloud.google.com/pubsubtopic-sample-2 created
```

### DIFF
--- a/pkg/webhook/state_into_spec_annotation_validator.go
+++ b/pkg/webhook/state_into_spec_annotation_validator.go
@@ -1,0 +1,60 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package webhook
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+type stateIntoSpecAnnotationValidator struct {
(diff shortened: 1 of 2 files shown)


pkg/webhook/state_into_spec_annotation_validator.go
```diff
@@ -0,0 +1,60 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+type stateIntoSpecAnnotationValidator struct {
+	client client.Client
+}
+
+func NewStateIntoSpecAnnotationValidatorHandler() HandlerFunc {
```
yuwenma, line 35: nit: docstring for public methods.

pkg/webhook/state_into_spec_annotation_validator.go
```diff
@@ -0,0 +1,60 @@
...
+		klog.Error(err)
+		return admission.Errored(http.StatusBadRequest,
+			fmt.Errorf("error decoding object: %w", err))
+	}
+
+	value, ok := k8s.GetAnnotation(k8s.StateIntoSpecAnnotation, obj)
+	if ok && value == k8s.StateMergeIntoSpec {
+		return allowedResponse.WithWarnings(
+			fmt.Sprintf("'%v: %v' is unsupported for CRDs added in "+
+				"1.114.0 and later. Use '%v' instead. More details can be "+
```
yuwenma, line 54: Shall we suggest not using the `state-into-spec` annotation instead? (compared to using "absent" value)

maqiuyujoyce: At the moment, not using `state-into-spec` annotation means let KCC defaults the value, which is still `merge` for most resources. So I added a link to the doc which suggests users default it to `Absent` at cluster/namespace-level - this is the most recommended behavior right now.

pkg/webhook/state_into_spec_annotation_validator.go
maqiuyujoyce: Done.

gemmahou:
/lgtm
/approve

Looks good to me!

PR #1996, opened 2024-06-12
feat: templating tool for direct controller builder

yuwenma:
# Example

```bash
cd dev/tools/controllerbuilder
```

```bash
go run main.go --service cloudbuild --version v1alpha1 --resourceInKind WorkerPool
```


dev/tools/controllerbuilder/scaffold/controller.go
```diff
@@ -0,0 +1,102 @@
...
+		return "", fmt.Errorf("create controller directory %s: %w", controllerDir, err)
+	}
+	controllerFilePath := filepath.Join(controllerDir, strings.ToLower(kind)+"_controller.go")
+	if _, err := os.Stat(controllerFilePath); err == nil {
+		return "", fmt.Errorf("controller file %s may already exist: %w", controllerFilePath, err)
+	}
+	return controllerFilePath, nil
+}
+
+func FormatImports(path string, out []byte) error {
```
justinsb, line 79: Nice!

dev/tools/controllerbuilder/scaffold/controller.go
```diff
@@ -0,0 +1,102 @@
...
+		Fragment:  true}
+	formatedOut, err := imports.Process(path, out, importOps)
+	if err != nil {
+		return fmt.Errorf("format controller file %s: %w", path, err)
+	}
+	return WriteToFile(path, formatedOut)
+}
+
+func WriteToFile(path string, out []byte) error {
+	f, err := os.Create(path)
```
justinsb, line 92: Nit: I don't think you need (or should) call os.Create here.  os.WriteFile creates the file IIRC.

justinsb: (If you want to check that the file doesn't exist, you should basically inline WriteFile, passing some different options to os.OpenFile IIRC)

justinsb:
Looks great.  Once we have a template we can start adding commands / tweaking the template as we see fit.

Really happy to see this 🎉 

/approve
/lgtm

PR #1999, opened 2024-06-12
Support Export in gkehubfeaturemembership controller

ziyue-101:
### Change description

Implement KRM export

### Tests you have done

### DIFF
--- a/pkg/controller/direct/export.go
+++ b/pkg/controller/direct/export.go
@@ -44,9 +44,7 @@ func Export(ctx context.Context, url string, config *config.ControllerConfig) (*
 		if err != nil {
 			return nil, err
 		}
-
 		return u, nil
 	}
-
 	return nil, nil
 }
--- a/pkg/controller/direct/gkehub/acmconversion.go
+++ b/pkg/controller/direct/gkehub/acmconversion.go
@@ -49,6 +49,24 @@ func convertKRMtoAPI_ConfigManagement(r *krm.FeaturemembershipConfigmanagement) 
 
 }
 
+func convertAPItoKRM_ConfigManagement(r *featureapi.ConfigManagementMembershipSpec) *krm.FeaturemembershipConfigmanagement {
+	krmObj := krm.FeaturemembershipConfigmanagement{}
+	if r.Binauthz != nil {
+		krmObj.Binauthz = convertAPItoKRM_Binauthz(r.Binauthz)
+	}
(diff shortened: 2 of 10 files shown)


justinsb:
Could we add a test for this, I think you need to add the supported gkehub kinds to the URL-generation logic here: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/tests/e2e/export.go#L43

ziyue-101:
> Could we add a test for this, I think you need to add the supported gkehub kinds to the URL-generation logic here: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/tests/e2e/export.go#L43

Hi Justin, thanks for the pointer, looks like the `exportResource` method maps a Kind to a URL. However, , a `GKEhubfeaturemembership` resource maps to two URLs (feature and membership). What's the suggested way to handle this case ;-)?

justinsb:
> However, , a `GKEhubfeaturemembership` resource maps to two URLs (feature URL and membership URL).

I think the code in question is in our tests, so worst case we can map it to multiple URLs ....  but of course that begs the question of what happens next (do we export two objects if there are two URLs).

I actually thought that GKEHubFeature mapped to the feature URL and GKEHubFeatureMembership mapped to the membership URL - is that not right?

ziyue-101:
> Could we add a test for this, I think you need to add the supported gkehub kinds to the URL-generation logic here: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/tests/e2e/export.go#L43

Done with `exportURI = "//gkehub.googleapis.com/projects/" + projectID + "/locations/" + location + "/features/" + featureName + "/memberships/" + membershipName`

ziyue-101:
The test failure is 
`export.go:94: error from export.Execute: error getting unstructured: error getting next unstruct: unexpected url //bigquery.googleapis.com/projects/mock-project/datasets/bigquerydatasetup3etti4elu6ek5ri7xa`

though the change is not related to the resource.


Looks like to me the test is using the wrong(gkehub) adapter to export the resource. 


https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/8331376698541c2d1b929324c85cd566ad58f7b4/pkg/controller/direct/registry/registry.go#L68 looks a bug for me as it always returns the first adapater in the registry and use that for test

ldanielmadariaga:

PR #2000, opened 2024-06-12
Add explanation for state-into-spec in LoggingLogMetric reference doc

maqiuyujoyce:
### Change description

Added warning to the reference doc as the behavior for the `state-into-spec` annotation is changed after LoggingLogMetric uses the direct controller.

Also did a small clean up for the note in AlloyDBInstance.

### Tests you have done

### DIFF
--- a/scripts/generate-google3-docs/resource-reference/templates/logging_logginglogmetric.tmpl
+++ b/scripts/generate-google3-docs/resource-reference/templates/logging_logginglogmetric.tmpl
@@ -8,7 +8,7 @@
 
 Caution: After
 [v1.118.1](https://github.com/GoogleCloudPlatform/k8s-config-connector/releases/tag/v1.118.1),
-`cnrm.cloud.google.com/state-into-spec` annotation is ignored during
+`cnrm.cloud.google.com/state-into-spec` annotation is no longer used during
 LoggingLogMetric reconciliation, and [externally-managed
 fields](https://cloud.google.com/config-connector/docs/concepts/managing-fields-externally)
 in the existing LoggingLogMetric resources are considered managed, i.e.
--- a/scripts/generate-google3-docs/resource-reference/templates/logging_logginglogmetric.tmpl
+++ b/scripts/generate-google3-docs/resource-reference/templates/logging_logginglogmetric.tmpl
@@ -6,6 +6,14 @@
 {% block body %}
 {{template "alphadisclaimer.tmpl" .}}
 
+Caution: After
+[v1.118.1](https://github.com/GoogleCloudPlatform/k8s-config-connector/releases/tag/v1.118.1),
+`cnrm.cloud.google.com/state-into-spec` annotation is ignored during
+LoggingLogMetric reconciliation, and [externally-managed
+fields](https://cloud.google.com/config-connector/docs/concepts/managing-fields-externally)
+in the existing LoggingLogMetric resources are considered managed, i.e.
+Kubernetes _is_ the source of truth for these values.
+
 <table>
 <thead>
 <tr>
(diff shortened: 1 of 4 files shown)


scripts/generate-google3-docs/resource-reference/templates/logging_logginglogmetric.tmpl
```diff
@@ -6,6 +6,14 @@
 {% block body %}
 {{template "alphadisclaimer.tmpl" .}}
 
+Caution: After
+[v1.118.1](https://github.com/GoogleCloudPlatform/k8s-config-connector/releases/tag/v1.118.1),
+`cnrm.cloud.google.com/state-into-spec` annotation is ignored during
```
yuwenma, line 11: nit: maybe change "ignored" to "dropped" (and remove the corresponding doc from the doc)? ignore is ambiguous because it could mean no-op or keeping-the-existing-behavior (previous default is merge)

yuwenma:
/lgtm

Thank you! One minor comment. Shall we drop the State-into-Spec annotation from the [Cloud doc](https://cloud.google.com/config-connector/docs/reference/resource-docs/logging/logginglogmetric)?

scripts/generate-google3-docs/resource-reference/templates/logging_logginglogmetric.tmpl
maqiuyujoyce: What about directly mentioning it is no-op? I think "dropped" might also be misleading because it could indicate the annotation won't be persisted in the KRM object, but when we do `kubectl describe`, we actually can still see this annotation if it is configured explicitly.

I think we can say something like:

```
Caution: After
[v1.118.1](https://github.com/GoogleCloudPlatform/k8s-config-connector/releases/tag/v1.118.1),
`cnrm.cloud.google.com/state-into-spec` annotation is no longer used during
LoggingLogMetric reconciliation, and [externally-managed
fields](https://cloud.google.com/config-connector/docs/concepts/managing-fields-externally)
in the existing LoggingLogMetric resources are considered managed, i.e.
Kubernetes _is_ the source of truth for these values.
```

@tarynlucas do you have any suggestions?

maqiuyujoyce:
> Thank you! One minor comment. Shall we drop the State-into-Spec annotation from the [Cloud doc](https://cloud.google.com/config-connector/docs/reference/resource-docs/logging/logginglogmetric)?

We should. Though because this is configured in a shared template, removing it just for LLM may not be very straight-forward. I plan to tackle it separately.

scripts/generate-google3-docs/resource-reference/templates/logging_logginglogmetric.tmpl
tarynlucas: The suggested edit LGTM! I think "no longer used" makes sense since we're explaining this as a changed behavior from a previous version.

One nit: I think in the last sentence we're putting the emphasis on the wrong thing ("_is_"): could we change this to say something like "...considered managed, meaning Kubernetes is the source of truth for these values instead of [whatever the pre-1.118.1 behavior was]."

Also do we mean Kubernetes resources or objects? Just saying "Kubernetes is a the source of truth" seems a bit unclear to me, but maybe I'm just not familiar with the context.

yuwenma:
> > Thank you! One minor comment. Shall we drop the State-into-Spec annotation from the [Cloud doc](https://cloud.google.com/config-connector/docs/reference/resource-docs/logging/logginglogmetric)?
> 
> We should. Though because this is configured in a shared template, removing it just for LLM may not be very straight-forward. I plan to tackle it separately.

Okay, that's fair.

scripts/generate-google3-docs/resource-reference/templates/logging_logginglogmetric.tmpl
yuwenma: +1 to "no longer used" I like it as well.

yuwenma:
/lgtm
/approve

Thanks!

PR #2003, opened 2024-06-13
chore: simple license header yr lint

acpana:
wip


acpana:
Going to park this here and can pick up if it becomes more useful/ salient.

PR #2005, opened 2024-06-14
fix: support ref for network field

acpana:
Changes `network` field to be a ref field for `CloudIDSEndpoint`. This is a breaking change but the resource is in `v1alpha1`.


yuwenma:
/lgtm
/approve

Thanks. PR looks good. I think we also want to have the golden check on this resource (maybe a follow up PR?)

acpana:
> want to have the golden check on this resource (maybe a follow up PR?)

Yes I am working to have an http log check for the `CloudIDSEndpoint` resource 💯

PR #2012, opened 2024-06-14
feat: CBWP update with ETag

yuwenma:
Add Etag to CloudBuildWorkerPool resource `status.ObservedState`.
When updating, we should pass in the etag to the CloudBuild server and expect a different etag from the response body.

Right now, the mockgcp replaces the ETag, so we can only see the diff from the golden object.


apis/cloudbuild/v1alpha1/workerpool_types.go
```diff
@@ -88,6 +88,8 @@ type CloudBuildWorkerPoolObservedState struct {
 	// +required
 	WorkerConfig  *WorkerConfig       `json:"workerConfig,omitempty"`
 	NetworkConfig *NetworkConfigState `json:"networkConfig,omitempty"`
+
+	ETag *string `json:"eTag,omitempty"`
```
justinsb, line 92: It's still alpha, so not a blocker but ... I don't know how we should capitalize etag.  I think it might just be `etag`.  I don't think I've seen `eTag` anywhere else (?)

mockgcp/mockcloudbuild/workerpool.go
```diff
@@ -77,6 +77,7 @@ func (s *CloudBuildV1) CreateWorkerPool(ctx context.Context, req *pb.CreateWorke
 		result.CreateTime = now
 		result.UpdateTime = now
 		result.State = pb.WorkerPool_RUNNING
+		result.Etag = "W/11111111111111111111111111111111"
```
justinsb, line 80: FYI I have some helper functions `computeEtag` that I typically put in utils.go in each mock<service> directory.  (And note that there are two versions, some which return strings, some which return []byte, because some services need string and some []byte)

Would be great to figure out a better place for them, but it's pretty easy to have more realistic etags :-)

justinsb:
A few nits but lgtm.

I'll hold in case you want to address them here.

/hold
/approve
/lgtm

apis/cloudbuild/v1alpha1/workerpool_types.go
yuwenma: Ah! You are right. It's interesting that the camel case format is `etag` with lower case `e` where the capitalized format is `ETag`.

Updated.

mockgcp/mockcloudbuild/workerpool.go
yuwenma: I like the `computeEtag`!  Make some further changes (with weak indicator), please take a look at the latest commit. If the common `computeEtag` looks good, I can update the mockgcp

justinsb:
Thanks @yuwenma 

/approve
/lgtm

PR #2014, opened 2024-06-15
tests: golden log for cloudidsendpoint

acpana:


pkg/test/resourcefixture/testdata/basic/cloudids/v1alpha1/cloudidsendpoint/update.yaml
```diff
@@ -24,4 +24,4 @@ spec:
   projectRef:
     external: ${projectId}
   threatExceptions:
-    - exception1-${uniqueId}
+    - "123"
```
acpana, line 27: this needs to be a string representation of an numeric-only id

PR #2015, opened 2024-06-15
chore: Always create ARTIFACTS dir in CI, if set

justinsb:
This keeps fussy tools happy.


yuwenma (commented):
/lgtm
/approve


Thanks!

PR #2020, opened 2024-06-17
feat: add ExternalReference (selfLink) with GCP update/deletion uniqueness guardrail

yuwenma:
Follow up PR to resolve comment in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1920#discussion_r1630938449

### DIFF
--- a/apis/cloudbuild/v1alpha1/workerpool_types.go
+++ b/apis/cloudbuild/v1alpha1/workerpool_types.go
@@ -68,6 +68,10 @@ type CloudBuildWorkerPoolStatus struct {
 	// +optional
 	ObservedGeneration *int64 `json:"observedGeneration,omitempty"`
 
+	/* A unique specifier for the CloudBuild workerpool resource in GCP.*/
+	// +optional
+	ExternalRef *string `json:"externalRef,omitempty"`
+
 	/* ObservedState is the state of the resource as most recently observed in GCP. */
 	// +optional
 	ObservedState *CloudBuildWorkerPoolObservedState `json:"observedState,omitempty"`
--- a/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_cloudbuildworkerpools.cloudbuild.cnrm.cloud.google.com.yaml
+++ b/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_cloudbuildworkerpools.cloudbuild.cnrm.cloud.google.com.yaml
@@ -164,6 +164,10 @@ spec:
                       type: string
                   type: object
                 type: array
+              externalRef:
+                description: A unique specifier for the CloudBuild workerpool resource
+                  in GCP.
+                type: string
               observedGeneration:
                 description: ObservedGeneration is the generation of the resource
                   that was most recently observed by the Config Connector controller.
--- a/pkg/controller/direct/cloudbuild/externalresource.go
+++ b/pkg/controller/direct/cloudbuild/externalresource.go
@@ -1,0 +1,73 @@
+/*
+Copyright 2024.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cloudbuild
+
+import (
+	"fmt"
+	"strings"
+
+	cloudbuildpb "cloud.google.com/go/cloudbuild/apiv1/v2/cloudbuildpb"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/externalresource"
+)
+
+const (
+	serviceBaseURL = "https://cloudbuild.googleapis.com/v1/"
+)
+
+func NewResourceRef(gcpObj *cloudbuildpb.WorkerPool) (*ResourceRef, error) {
+	baseResourceRef := externalresource.New(serviceBaseURL, gcpObj)
+	extResRef := ValueOf(baseResourceRef.Get())
+	segments := strings.Split(extResRef, "/projects/")
+	if len(segments) != 2 {
+		return nil, fmt.Errorf("externalReference should be <baseUrl>/projects/<project>/locations/<location>/workerPools/<workerPool>, got %s",
+			extResRef)
+	}
+	segments = strings.Split(segments[1], "/")
+	if len(segments) == 5 && segments[1] == "locations" && segments[3] == "workerPools" {
+		return &ResourceRef{
+			project:     segments[0],
+			location:    segments[2],
+			resourceID:  segments[4],
+			externalRef: baseResourceRef.Get(),
+		}, nil
+	}
+	return nil, fmt.Errorf("externalReference should be in the form of <baseUrl>/projects/<project>/locations/<location>/workerPools/<workerPool>, got %s",
+		extResRef)
+}
+
+type ResourceRef struct {
+	resourceID  string
+	location    string
+	project     string
+	externalRef *string
+}
+
+func (e *ResourceRef) GetExternalReference() *string {
+	return e.externalRef
+}
+
+func (e *ResourceRef) GetResourceID() string {
+	return e.resourceID
+}
+
+func (e *ResourceRef) GetLocation() string {
+	return e.location
+}
+
+func (e *ResourceRef) GetProject() string {
+	return e.project
+}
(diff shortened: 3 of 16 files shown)


pkg/controller/direct/cloudbuild/externalresource.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"strings"
+
+	cloudbuildpb "cloud.google.com/go/cloudbuild/apiv1/v2/cloudbuildpb"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/externalresource"
+)
+
+func NewResourceRef(gcpObj *cloudbuildpb.WorkerPool) *ResourceRef {
+	return &ResourceRef{
```
yuwenma, line 27: This CBWP ResourceRef customize the common ResourceRef, mostly to provide dedicated user information.

pkg/controller/direct/cloudbuild/workerpool_controller.go
```diff
@@ -197,10 +197,14 @@ func (a *Adapter) Create(ctx context.Context, u *unstructured.Unstructured) erro
 	}
 	status.ObservedState.CreateTime = ToOpenAPIDateTime(created.GetCreateTime())
 	status.ObservedState.UpdateTime = ToOpenAPIDateTime(created.GetUpdateTime())
+	status.ExternalRef = NewResourceRef(created).Get()
 	return setStatus(u, status)
 }
 
 func (a *Adapter) Update(ctx context.Context, u *unstructured.Unstructured) error {
+	if err := a.ValidateExternalResource(); err != nil {
```
yuwenma, line 205: We do not allow user to update the `spec.resourceID` `spec.projectRef` or `spec.location` to a different value (which means a different GCP object). 

Ideally we can move this to the directbase_controller. (or we already have this check somewhere?)

pkg/controller/direct/externalresource/resource.go
```diff
@@ -0,0 +1,64 @@
...
+	GetName() string
+}
+
+// New builds the External Reference for CloudBuildWorkerPool when the object already exists on the GCP server.
+// Ideally, it expects to use the GCP object's selfLink value as the external reference. However, if
+// the resource does not have `selfLink` (like CloudBuildWorkerPool), it builds the URL from the GCP object.
+func New(host string, gcpObj GCPResource) *ExternalResourceReference {
+	baseExternalRef := &ExternalResourceReference{
+		hasSelfLink: false,
+		externalRef: host + gcpObj.GetName(),
```
yuwenma, line 38: Any better name than "host"? This externalRef is to make the value the same as the `seflLink`.

apis/cloudbuild/v1alpha1/workerpool_types.go
```diff
@@ -68,7 +68,10 @@ type CloudBuildWorkerPoolStatus struct {
 	// +optional
 	ObservedGeneration *int64 `json:"observedGeneration,omitempty"`
 
-	ResourceID *string `json:"resourceID,omitempty"`
+	/* The GCP URL to the cloudbuild workerpool resource.
+	The cloudbuild workerpool does not have selfLink. */
+	// +optional
+	ExternalRef *string `json:"externalRef,omitempty"`
```
justinsb, line 74: Oooh I like this metaphor, mirroring `external` in a ref.  If we go with this, Eventually we'll want to make sure that this format is always accepted as a ref.external format (we can always support additional formats as well, but it'll be much less confusing for users and for us if we have them match)

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_cloudbuildworkerpools.cloudbuild.cnrm.cloud.google.com.yaml
```diff
@@ -164,6 +164,10 @@ spec:
                       type: string
                   type: object
                 type: array
+              externalRef:
+                description: The GCP URL to the cloudbuild workerpool resource. The
```
justinsb, line 168: Maybe "A unique specifier for the CloudBuild workerpool resource in GCP."

I don't think you need to say "The cloudbuild workerpool does not have selfLink.".  (That might have been a code comment, it got pulled in here though)

pkg/controller/direct/cloudbuild/externalresource.go
```diff
@@ -0,0 +1,51 @@
...
+		externalresource.New("https://cloudbuild.googleapis.com/v1/", gcpObj),
```
justinsb, line 28: Nit: I don't know if we want v1.  Although I guess it depends on whether in OnePlatform a v1 and v2 with the same name would be the same resource.

pkg/controller/direct/cloudbuild/externalresource.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+}
+
+type ResourceRef struct {
+	*externalresource.ExternalResourceReference
+}
+
+func (e *ResourceRef) GetResourceID() string {
+	segments := strings.Split(*e.Get(), "/workerPools/")
```
justinsb, line 37: Two nits:

1) Dereferencing pointers is always scary (because it panics on nil), better to use a helper like ValueOf
2)  I think you should also check that segments actually has len != 0.

It might be the fluent way to do this is to move the parsing to NewResourceRef (which can then return an error)

pkg/controller/direct/cloudbuild/workerpool_controller.go
```diff
@@ -283,6 +287,8 @@ func (a *Adapter) Update(ctx context.Context, u *unstructured.Unstructured) erro
 	}
 	status.ObservedState.CreateTime = ToOpenAPIDateTime(updated.GetCreateTime())
 	status.ObservedState.UpdateTime = ToOpenAPIDateTime(updated.GetUpdateTime())
+	// This value should not be updated. Just in case.
+	status.ExternalRef = NewResourceRef(updated).Get()
```
justinsb, line 291: You may want to write a quick design-doc for "status.externalRef" including "why is it not status.selfLink", how it fixes resource deletion etc

pkg/controller/direct/cloudbuild/workerpool_controller.go
```diff
@@ -313,6 +320,38 @@ func (a *Adapter) Delete(ctx context.Context) (bool, error) {
 	return true, nil
 }
 
+// ValidateExternalResource compares the `status.externalRef` with the `spec` Project, Location and
+// (external) resourceID to make sure those fields are immutable and matches the previous deployed value.
```
justinsb, line 324: Oooh cool - another advantage!

pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -225,6 +225,18 @@ func (r *reconcileContext) doReconcile(ctx context.Context, u *unstructured.Unst
 		return false, r.handleUpdateFailed(ctx, u, err)
 	}
 
+	// To create, update or delete the GCP object, we need to get the GCP object first.
+	// Because the object contains the cloud service information like `selfLink` `ID` required to validate
+	// the resource uniqueness before updating/deleting.
+	existsAlready, err := adapter.Find(ctx)
+	if err != nil {
+		if unwrappedErr, ok := lifecyclehandler.CausedByUnresolvableDeps(err); ok {
+			logger.Info(unwrappedErr.Error(), "resource", k8s.GetNamespacedName(u))
```
justinsb, line 234: One gotcha is that if we're deleting we likely don't need the spec refs, we just need status.externalRef.

I guess we can tackle this later (as part of cleaning up resource cleanup)?

justinsb:
You have some conflicts, but...

/approve
/lgtm

pkg/controller/direct/externalresource/resource.go
yuwenma: Thanks! I like serviceBaseURL

apis/cloudbuild/v1alpha1/workerpool_types.go
yuwenma: Yes!

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_cloudbuildworkerpools.cloudbuild.cnrm.cloud.google.com.yaml
yuwenma: Sounds good!

pkg/controller/direct/cloudbuild/externalresource.go
yuwenma: I see what you mean. I think we need to specify a version in the URL anyways to make it valid. Maybe I'm wrong, but does the v1/v2 bundles in the godoc.org/cloud.google.com/go?

pkg/controller/direct/cloudbuild/externalresource.go
yuwenma: Thanks! The code looks much nicer now.

justinsb:
Thanks - looks great!

/approve
/lgtm

PR #2038, opened 2024-06-18
chore: bump hashicorp cc version to 1.119.0

yuwenma:
Related to #2035, fix the presubmit issue for 1.119.0 operator release


yuwenma:
this commit would be included in #2035

Issue #2046, opened 2024-06-18
Support soft delete retention for storage bucket


600lyy:
### Checklist

- [X] I did not find a related open enhancement request.
- [X] I understand that enhancement requests filed in the GitHub repository are by default low priority.
- [X] If this request is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Describe the feature or resource

The bucket's soft delete policy, which defines the period of time that soft-deleted objects will be retained, and cannot be permanently deleted. If it is not provided, by default Google Cloud Storage sets this to default soft delete policy`

We need this feature to be added in KCC before GCS team start the billing for retention as of Aug 25th

### Additional information

Found this PR addressed this feature, however it is not getting any update since April
https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1527

prakharjoshi:
@600lyy, it seems like there is no option to disable the soft delete policy via the config connector. As per the doc/PR it falls back to organisation-level soft delete policy. Is there a way to disable soft delete policy for specific buckets via config connectors?

ldanielmadariaga:
This item has been open for >12 months. As part of scheduled repository maintenance, this Issue/PR will be closed. If this is still a critical ask, please re-open and work with your Google Cloud Account team to prioritize it with GCP KCC Product Management.

PR #2051, opened 2024-06-18
Implementing soft delete for storage bucket

600lyy:
### Change description

Add support for soft delete policy for Google storage bucket.

The bucket's soft delete policy, which defines the period of time that soft-deleted objects will be retained, and cannot be permanently deleted. If it is not provided, by default Google Cloud Storage sets this to default soft delete policy`

This is the implementation for https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/2046


yuwenma:
/lgtm

PR looks good. @600lyy could you fix the presubmit check?

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_storagebuckets.storage.cnrm.cloud.google.com.yaml
```diff
@@ -287,6 +287,23 @@ spec:
                 required:
                 - retentionPeriod
                 type: object
+              softDeletePolicy:
+                description: The bucket's soft delete policy, which defines the period
+                  of time that soft-deleted objects will be retained, and cannot be
+                  permanently deleted. If it is not provided, by default Google Cloud
+                  Storage sets this to default soft delete policy.
+                properties:
+                  effectiveTime:
```
maqiuyujoyce, line 296: @yuwenma this field is an output-only field that gets inserted under `spec`. According to our previous discussion, we don't want to add any output-only spec fields any more - is my understanding correct?

pkg/test/resourcefixture/testdata/basic/storage/v1beta1/storagebucket/update.yaml
```diff
@@ -23,3 +23,5 @@ spec:
   versioning:
     enabled: true
   lifecycleRule: []
+  softDeletePolicy:
+    retentionDurationSeconds: 0
```
maqiuyujoyce, line 27: I ran the test locally and found that `0` was not used in the update, and the value kept to be `604800`. Could you verify that this is a valid update; or whether setting the value to `0` is actually effective? AFAIK, we don't support setting a primitive-typed field to its empty value.

maqiuyujoyce: I checked again and actually it worked.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_storagebuckets.storage.cnrm.cloud.google.com.yaml
maqiuyujoyce: This is a workaround to hide output-only spec field but still support it under observed state: #2075 .

justinsb:
Looks like I can do the rebase / cleanup; just pushed to the PR branch 🤞

justinsb:
Looks like a (simple) rebase was all that was needed - thank you @600lyy 

/approve
/lgtm

@maqiuyujoyce I think we should merge this before #2075

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_storagebuckets.storage.cnrm.cloud.google.com.yaml
yuwenma: Thank you @maqiuyujoyce for the quick workaround!

maqiuyujoyce:
@justinsb as long as we don't cut a release, the order to merge #2051 and #2075 shouldn't matter.

/lgtm

PR #2057, opened 2024-06-18
feat: Add support for export to MonitoringDashboard direct actuation

justinsb:
- **feat: Add support for export to MonitoringDashboard direct actuation**
- **chore: testdata for MonitoringDashboard including export**

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/monitoring/v1beta1/monitoringdashboard/monitoringdashboardbasic/_generated_export_monitoringdashboardbasic.golden
+++ b/pkg/test/resourcefixture/testdata/basic/monitoring/v1beta1/monitoringdashboard/monitoringdashboardbasic/_generated_export_monitoringdashboardbasic.golden
@@ -1,16 +1,7 @@
 apiVersion: monitoring.cnrm.cloud.google.com/v1beta1
 kind: MonitoringDashboard
 metadata:
-  annotations:
-    cnrm.cloud.google.com/management-conflict-prevention-policy: none
-  finalizers:
-  - cnrm.cloud.google.com/finalizer
-  - cnrm.cloud.google.com/deletion-defender
-  generation: 3
-  labels:
-    cnrm-test: "true"
   name: monitoringdashboard-${uniqueId}
-  namespace: ${uniqueId}
 spec:
   columnLayout:
     columns:
@@ -48,18 +39,7 @@ spec:
             scale: LINEAR
       - logsPanel:
           filter: metric.type="agent.googleapis.com/nginx/connections/accepted_count"
-          resourceNames: []
         title: Widget 4
   displayName: monitoringdashboard-updated
   projectRef:
     external: ${projectId}
-  resourceID: monitoringdashboard-${uniqueId}
-status:
-  conditions:
-  - lastTransitionTime: "1970-01-01T00:00:00Z"
-    message: The resource is up to date
-    reason: UpToDate
-    status: "True"
-    type: Ready
-  etag: abcdef123456
-  observedGeneration: 3
--- a/pkg/test/resourcefixture/testdata/basic/monitoring/v1beta1/monitoringdashboard/monitoringdashboardfull/_generated_export_monitoringdashboardfull.golden
+++ b/pkg/test/resourcefixture/testdata/basic/monitoring/v1beta1/monitoringdashboard/monitoringdashboardfull/_generated_export_monitoringdashboardfull.golden
@@ -1,16 +1,7 @@
 apiVersion: monitoring.cnrm.cloud.google.com/v1beta1
 kind: MonitoringDashboard
 metadata:
-  annotations:
-    cnrm.cloud.google.com/management-conflict-prevention-policy: none
-  finalizers:
-  - cnrm.cloud.google.com/finalizer
-  - cnrm.cloud.google.com/deletion-defender
-  generation: 3
-  labels:
-    cnrm-test: "true"
   name: monitoringdashboard-${uniqueId}
-  namespace: ${uniqueId}
 spec:
   columnLayout:
     columns:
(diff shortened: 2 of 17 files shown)


pkg/test/resourcefixture/testdata/basic/monitoring/v1beta1/monitoringdashboard/monitoringdashboardbasic/_http.log
```diff
@@ -160,7 +160,7 @@ X-Xss-Protection: 0
             "xyChart": {
               "dataSets": [
                 {
-                  "plotType": "STACKED_BAR",
+                  "plotType": "LINE",
```
justinsb, line 163: I had to change these values as part of cross-validating with GCP (which is also why we have some mockgcp changes)

pkg/test/resourcefixture/testdata/basic/monitoring/v1beta1/monitoringdashboard/monitoringdashboardbasic/_generated_export_monitoringdashboardbasic.golden
```diff
@@ -0,0 +1,45 @@
...
+              unitOverride: "1"
+          yAxis:
+            label: y1Axis
+            scale: LINEAR
+      - logsPanel:
+          filter: metric.type="agent.googleapis.com/nginx/connections/accepted_count"
+        title: Widget 4
+  displayName: monitoringdashboard-updated
+  projectRef:
+    external: ${projectId}
```
yuwenma, line 45: hmmm, do you know why this does not have the `.status`?

pkg/test/resourcefixture/testdata/basic/monitoring/v1beta1/monitoringdashboard/monitoringdashboardfull/_generated_export_monitoringdashboardfull.golden
```diff
@@ -0,0 +1,48 @@
...
+            scale: LINEAR
+      - logsPanel:
+          filter: metric.type="agent.googleapis.com/nginx/connections/accepted_count"
+          resourceNames:
+          - external: projects/${projectId}
+            kind: Project
+        title: Widget 4
+  displayName: monitoringdashboard-full
+  projectRef:
+    external: ${projectId}
```
yuwenma, line 48: Same question. Also, these files does not have the yaml extension. Is that intended?

tests/e2e/export.go
```diff
@@ -22,44 +22,35 @@ import (
...
 )
 
 func exportResource(h *create.Harness, obj *unstructured.Unstructured) string {
 	exportURI := ""
 
-	projectID := obj.GetAnnotations()["cnrm.cloud.google.com/project-id"]
-	if projectID == "" {
-		projectID = h.Project.ProjectID
-	}
+	projectID := resolveProjectID(h, obj)
```
yuwenma, line 32: 👍 Love this!

pkg/test/resourcefixture/testdata/basic/monitoring/v1beta1/monitoringdashboard/monitoringdashboardrefs/create.yaml
```diff
@@ -51,7 +51,6 @@ spec:
                   perSeriesAligner: ALIGN_RATE
               unitOverride: "1"
             plotType: "STACKED_BAR"
-          timeshiftDuration: 0s
```
yuwenma, line 54: Is `0s` similiar as a `reset` ?

pkg/test/resourcefixture/testdata/basic/monitoring/v1beta1/monitoringdashboard/monitoringdashboardbasic/_generated_export_monitoringdashboardbasic.golden
justinsb: So I _think_ export should not have status, the intent is that we're going to feed it back into `kubectl apply` (for example).  I could be wrong on that, but I don't think I am.

It wouldn't be hard to support status, particularly once we are generating the KRM <-> proto mapping functions.  But today I don't think we do.

pkg/test/resourcefixture/testdata/basic/monitoring/v1beta1/monitoringdashboard/monitoringdashboardfull/_generated_export_monitoringdashboardfull.golden
justinsb: I think the naming of these files could be improved, but this is the name that we're currently using.  No objection to changing it, but I think it should be another PR.

For example, we should probably:

* Make the extension .yaml
* Name it after the kind and name of what we are exporting
* Consider exporting more than the "primary" object

pkg/test/resourcefixture/testdata/basic/monitoring/v1beta1/monitoringdashboard/monitoringdashboardrefs/create.yaml
justinsb: It's more banal than that, timeshiftDuration must be a minimum of 60 seconds, and cannot be used with STACKED_BAR.  Because we're testing it above, I thought it easier just to drop this invalid value.

(I only realized this when I ran it against real GCP!)

pkg/test/resourcefixture/testdata/basic/monitoring/v1beta1/monitoringdashboard/monitoringdashboardbasic/_generated_export_monitoringdashboardbasic.golden
yuwenma: Ah! Didn't realize this is *_generated_export_* 
That makes sense.

PR #2059, opened 2024-06-18
fix: guard against nil model

acpana:
The `model`s for the dynamic controller presubmits are nil bc we never `Init` the registry. This also proved to me that we probably want to error out in `AddController` if `model` is nil for a `gvk`. Finally, if the `registration` for a `gk` contains a `nil` model, we probably want to err out to signal to consumers that `Init` was not called.

### DIFF
--- a/pkg/controller/direct/registry/registry.go
+++ b/pkg/controller/direct/registry/registry.go
@@ -39,7 +39,7 @@ type ModelFactoryFunc func(ctx context.Context, config *config.ControllerConfig)
 
 func GetModel(gk schema.GroupKind) (directbase.Model, error) {
 	registration := singleton.registrations[gk]
-	if registration == nil {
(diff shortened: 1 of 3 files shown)


pkg/controller/direct/registry/registry.go
```diff
@@ -39,7 +39,7 @@ type ModelFactoryFunc func(ctx context.Context, config *config.ControllerConfig)
 
 func GetModel(gk schema.GroupKind) (directbase.Model, error) {
 	registration := singleton.registrations[gk]
-	if registration == nil {
+	if registration == nil || registration.model == nil {
```
justinsb, line 42: We could also test this one separately and say "Init not called" (I guess?)

justinsb (approved):

PR #2052, opened 2024-06-18
chore: update protobuf libraries

justinsb:

PR #2064, opened 2024-06-20
chore: normalize status.observedState.etag in golden tests

justinsb:
Also simplify etag computation, as we don't rely on the value itself
any more.

### DIFF
--- a/mockgcp/common/fields/etag.go
+++ b/mockgcp/common/fields/etag.go
@@ -19,53 +19,17 @@ import (
 	"encoding/base64"
 	"fmt"
 
-	"google.golang.org/protobuf/encoding/prototext"
 	"google.golang.org/protobuf/proto"
 )
 
-var mustFields = []string{
-	"displayName",
-	"state",
-}
-
-// ComputeEtag computes the etag of the proto object with weak indicator.
-func ComputeEtag(obj proto.Message) string {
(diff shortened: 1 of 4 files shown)


justinsb:
(I did this because it blocked https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2052 )

mockgcp/common/fields/etag.go
```diff
@@ -19,53 +19,17 @@ import (
...
-		panic(fmt.Sprintf("converting to prototext: %v", err))
+		panic(fmt.Sprintf("converting to proto: %v", err))
 	}
-	h := sha256.Sum256([]byte(m))
-	str := base64.StdEncoding.EncodeToString(h[:])
-	strong := fmt.Sprintf(`"%s"`, str) // ETag must be quoted.
-	return "W/" + strong
-}
-
-func ComputeEtagBytes(obj proto.Message) []byte {
```
yuwenma, line 61: I checked the current in-use `ComputeEtag`. Some expects a []byte form and some needs a pointer. So I'm trying to make this a generic lib for other  contributors to use out of the box, so they don't need to handle the dereference pointer badly (like what I did previously).

yuwenma:
/lgtm
/approve



> (I did this because it blocked #2052 )

Ah! Fair

PR #2065, opened 2024-06-20
Make context optional for evaluation

barney-s:
### Change description

This is to allow simpler compositions and especially compositions that don't use any cloud resources.

### DIFF
--- a/experiments/compositions/composition/internal/controller/expander_reconciler.go
+++ b/experiments/compositions/composition/internal/controller/expander_reconciler.go
@@ -424,17 +424,23 @@ func (r *ExpanderReconciler) evaluateAndSavePlan(ctx context.Context, logger log
 	}
 
 	// read context in cr.namespace
+	var contextBytes []byte
 	contextcr := unstructured.Unstructured{}
 	contextcr.SetGroupVersionKind(contextGVK)
 	contextNN := types.NamespacedName{Namespace: cr.GetNamespace(), Name: "context"}
 	if err := r.Get(ctx, contextNN, &contextcr); err != nil {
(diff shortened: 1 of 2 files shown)


xiaoweim:
Looks good, does this mean that not all composition uses context? (since so far my understanding is that all compositions use context and will generate a plan)

experiments/compositions/composition/internal/controller/expander_reconciler.go
```diff
@@ -424,17 +424,20 @@ func (r *ExpanderReconciler) evaluateAndSavePlan(ctx context.Context, logger log
 	}
 
 	// read context in cr.namespace
+	var contextBytes []byte
 	contextcr := unstructured.Unstructured{}
 	contextcr.SetGroupVersionKind(contextGVK)
 	contextNN := types.NamespacedName{Namespace: cr.GetNamespace(), Name: "context"}
 	if err := r.Get(ctx, contextNN, &contextcr); err != nil {
 		logger.Error(err, "unable to fetch Context CR", "context", contextNN)
```
justinsb, line 432: Can we ignore _only_ `apierrors.IsNotFound(err)`, and still return an error if it's a different error.  The reason it matters is that you don't want e.g. timeout talking to kube-apiserver to cause a different, temporary reconciliation result.

barney-s: Ah yes. That was the intent.

barney-s:
> Looks good, does this mean that not all composition uses context? (since so far my understanding is that all compositions use context and will generate a plan)

Yes for non-cloud resources, `context` is optional.

PR #2073, opened 2024-06-20
fix: add proto-to-mapper support for fixed64

justinsb:
It is the same as uint64, just represented differently on the wire.


yuwenma:
/lgtm
/approve

Interesting.. thanks for the PR description!

PR #2074, opened 2024-06-20
fix: Trigger updates for mysqlinstance + sqlserverinstance mockgcp tests

jasonvigil:
### Change description

Seems that the existing `metadata.labels` changes do not incur updates to the resources. So that we exercise the mockgcp update path, made some changes to `spec.settings.diskSize` for these resources.

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/mysqlinstance/_http.log
+++ b/pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/mysqlinstance/_http.log
@@ -24,10 +24,11 @@ User-Agent: google-api-go-client/0.5 Terraform/ (+https://www.terraform.io) Terr
   "settings": {
     "activationPolicy": "ALWAYS",
     "availabilityType": "ZONAL",
+    "dataDiskSizeGb": "10",
     "dataDiskType": "PD_SSD",
     "edition": "ENTERPRISE",
     "pricingPlan": "PER_USE",
-    "storageAutoResize": true,
+    "storageAutoResize": false,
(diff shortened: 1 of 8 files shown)


jasonvigil:
WIP: the disk size field isn't updating

pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/mysqlinstance/_http.log
```diff
@@ -176,8 +177,8 @@ Grpc-Metadata-Content-Type: application/grpc
     "cert": "-----BEGIN CERTIFICATE-----\n-----END CERTIFICATE-----\n",
     "certSerialNumber": "0",
     "commonName": "common-name",
-    "createTime": "2024-06-19T22:54:22.309289161Z",
-    "expirationTime": "2025-06-19T22:54:22.309289161Z",
+    "createTime": "2024-06-20T23:54:19.971010960Z",
+    "expirationTime": "2025-06-20T23:54:19.971010960Z",
```
jasonvigil, line 181: @yuwenma should we normalize these timestamps?

jasonvigil: Looks like this was fixed in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2094

yuwenma (commented):
/lgtm
/approve

Looks good! Is those two fields the only mutable fields? 
FYI, we don't need to verify each label changes (that logic is independent from the Direct controller code). But no harm.

PR #2061, opened 2024-06-20
feat: Turn on stricter golden log check in e2e fixture (part 2)

yuwenma:
Turn on the stricter golden log check in fixture presubmit. This is the second part and needs  https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1934 to merge first.

PR #2093, opened 2024-06-21
fix: add TODO to the mockgcp TagKey resourceID

yuwenma:

### DIFF
--- a/tests/e2e/normalize.go
+++ b/tests/e2e/normalize.go
@@ -115,6 +115,16 @@ func normalizeKRMObject(u *unstructured.Unstructured, project testgcp.GCPProject
 			name, _, _ = unstructured.NestedString(u.Object, "status", "name")
 		}
 		tokens := strings.Split(name, "/")
+		if len(tokens) == 1 {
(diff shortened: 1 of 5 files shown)


tests/e2e/normalize.go
```diff
@@ -115,6 +115,16 @@ func normalizeKRMObject(u *unstructured.Unstructured, project testgcp.GCPProject
 			name, _, _ = unstructured.NestedString(u.Object, "status", "name")
 		}
 		tokens := strings.Split(name, "/")
+		if len(tokens) == 1 {
+			switch u.GetKind() {
+			case "TagsTagKey", "TagsTagValue":
+				// TODO: The mock TagKey server returns the correct format `tagKeys/{number}`, but the golden object `status.name`
```
justinsb, line 121: I'm not sure this is wrong, I think resourceID is "just the name".

justinsb (approved):

tests/e2e/normalize.go
justinsb: (So we may not need the TODO)

PR #2095, opened 2024-06-21
tests: add more StorageBucket tests

justinsb:
Covering softDeletePolicy.retentionDurationSeconds in particular.


justinsb:
I think we're also going to need some with state-into-spec: absent, but this PR is already getting pretty big...

yuwenma:
/lgtm

Looks like we need to re-generate the http log.

yuwenma:
/approve 

PR looks good. 
The failed VCR test is for `sourcereporepository` but I don't see how that's related.  I reran the test. hope it can pass.

PR #2096, opened 2024-06-21
fix: sort mockgcp string slice fields

yuwenma:


justinsb:
We should probably have a bug for this one, if the values are being ordered "randomly" (non-deterministically I guess would be a better way of saying that).  If it's the (real) service though, I guess there's not a lot we can do about it.

That said, no objection...

/approve
/lgtm

PR #2099, opened 2024-06-22
fix: replace mockresourcemanager generated Project number

yuwenma:

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/resourcemanager/v1beta1/project/projectmovedfoldertofolder/_generated_object_projectmovedfoldertofolder.golden.yaml
+++ b/pkg/test/resourcefixture/testdata/basic/resourcemanager/v1beta1/project/projectmovedfoldertofolder/_generated_object_projectmovedfoldertofolder.golden.yaml
@@ -24,5 +24,5 @@ status:
     reason: UpToDate
     status: "True"
     type: Ready
-  number: "518915279"
+  number: ${projectNumber}
   observedGeneration: 3
--- a/pkg/test/resourcefixture/testdata/basic/resourcemanager/v1beta1/project/projectmovedfoldertofolder/_generated_object_projectmovedfoldertofolder.golden.yaml
+++ b/pkg/test/resourcefixture/testdata/basic/resourcemanager/v1beta1/project/projectmovedfoldertofolder/_generated_object_projectmovedfoldertofolder.golden.yaml
@@ -24,5 +24,5 @@ status:
     reason: UpToDate
     status: "True"
     type: Ready
-  number: "2589592254"
+  number: "518915279"
   observedGeneration: 3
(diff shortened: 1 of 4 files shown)


pkg/test/resourcefixture/testdata/basic/resourcemanager/v1beta1/project/projectmovedfoldertofolder/_generated_object_projectmovedfoldertofolder.golden.yaml
```diff
@@ -24,5 +24,5 @@ status:
     reason: UpToDate
     status: "True"
     type: Ready
-  number: "2589592254"
+  number: "518915279"
```
justinsb, line 27: This isn't a magic number, is it?

tests/e2e/normalize.go
```diff
@@ -95,9 +95,18 @@ func normalizeKRMObject(u *unstructured.Unstructured, project testgcp.GCPProject
 	visitor.stringTransforms = append(visitor.stringTransforms, func(path string, s string) string {
 		return strings.ReplaceAll(s, project.ProjectID, "${projectId}")
 	})
+
+	// Update the project number with variable marker.
+	if u.GetKind() == "Project" {
+		// The ProjectNumber is ProjectID based. For those tests relying on the Mock Resource Manager server,
+		// the ProjectID is dynamic, different from the default mock test Project "mock-project",
+		// so the number is different as well.
+		visitor.replacePaths[".status.number"] = fmt.Sprintf("%d", project.ProjectNumber)
```
justinsb, line 104: Maybe we should replace with `${projectNumber}` ?  Though TBH I'm not entirely sure why it doesn't get replaced below.  Maybe the order in which these replacements run?

pkg/test/resourcefixture/testdata/basic/resourcemanager/v1beta1/project/projectmovedfoldertofolder/_generated_object_projectmovedfoldertofolder.golden.yaml
yuwenma: not ready for review. I have a better idea.

tests/e2e/normalize.go
yuwenma: Yes, yes, I also want to use `${projectNumber}`. Please see the last commit. 
The reason of not replacing is that those tests have two Projects, one is the "mock-project", the other is the real project which `uniqId` is dynamically generated. The below replacement `strings.ReplaceAll(s, fmt.Sprintf("%d", project.ProjectNumber), "${projectNumber}")` can only find the `mock-project` number, but not the dynamic ones. This only happens for tests that call the mock resource manager.

justinsb (approved):

PR #2109, opened 2024-06-22
fix: replace targetId with id marker

yuwenma:

More context: many golden http log uses targetId (see https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2061/files#diff-45e74dc3e22d6fbc3f979edfe4959d40c702bf51bea40854de4e4ab393bb5330). Override the targetId may be a little bit too aggressive. 

cc @acpana

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/compute/v1alpha1/computemanagedsslcertificate/managedsslcertificate/_http.log
+++ b/pkg/test/resourcefixture/testdata/basic/compute/v1alpha1/computemanagedsslcertificate/managedsslcertificate/_http.log
@@ -64,7 +64,7 @@ X-Xss-Protection: 0
   "selfLink": "https://www.googleapis.com/compute/beta/projects/${projectId}/global/operations/${operationID}",
   "startTime": "2024-04-01T12:34:56.123456Z",
   "status": "RUNNING",
-  "targetId": "7780111738275460050",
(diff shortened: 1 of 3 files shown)


pkg/test/resourcefixture/testdata/basic/compute/v1alpha1/computemanagedsslcertificate/managedsslcertificate/_http.log
```diff
@@ -127,7 +127,7 @@ X-Xss-Protection: 0
   "selfLink": "https://www.googleapis.com/compute/beta/projects/${projectId}/global/operations/${operationID}",
   "startTime": "2024-04-01T12:34:56.123456Z",
   "status": "RUNNING",
-  "targetId": "7780111738275460050",
+  "targetId": "1719033726205916358",
```
justinsb, line 130: I don't get why this one wasn't replaced.  Looking at the code it should have been, I think.  Maybe/hopefully you just need to recapture the golden output here?

yuwenma: 😆 ah, yes. I reminded myself to recapture the golden log while waiting for another full fixtures run. It took too long and I totally forgot this.

justinsb (approved):

PR #2124, opened 2024-06-24
refactor: move IAMServiceAccount reference resolution into API package

justinsb:
This makes it easier to consume.

### DIFF
--- a/apis/refs/v1beta1/gcpserviceaccountref.go
+++ b/apis/refs/v1beta1/gcpserviceaccountref.go
@@ -14,6 +14,27 @@
 
 package v1beta1
 
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type IAMServiceAccountRef struct {
+	/* The GCP Service Account Email used for auth when secretType is gcpServiceAccount. Allowed value: The `email` field of an `IAMServiceAccount` resource. */
+	External string `json:"external,omitempty"`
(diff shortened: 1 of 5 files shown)


justinsb:
/assign @ziyue-101

Moving the reference resolution to replace the ref seems to make it much simpler to use.  WDYT?

apis/refs/v1beta1/gcpserviceaccountref.go
```diff
@@ -14,20 +14,111 @@
...
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type IAMServiceAccountRef struct {
+	/* The GCP Service Account Email used for auth when secretType is gcpServiceAccount. Allowed value: The `email` field of an `IAMServiceAccount` resource. */
```
justinsb, line 30: I'm wondering if we want this documentation to be on the field in the parent type, rather than having different references for each.  I think it might also make our docs a bit simpler if we just linked to documentation on how to use refs, rather than repeating it every time.

yuwenma: +1 to have the doc in the parent type.  Also, I'm wondering if we do need two SA resolve methods here (Maybe one is better?)

yuwenma (commented):
/lgtm
/approve

/hold one question about if we shall keep a single GSA

apis/refs/v1beta1/gcpserviceaccountref.go
ziyue-101: This document can be on the parent type as it is generic. The one on  line 39 can't as it is field specific.

justinsb: I proposed we merge this refactor, then look at de-duping the two types, by seeing whether it makes sense when we move documentation to the parent.

We need two SA resolve types if we have two types.  The bulk of the logic is shared between them though, they both call into resolveServiceAccount.

PR #2125, opened 2024-06-24
chore: reformat VCR command in docs for easier copy-paste

justinsb:
When developers are looking at the markdown, we still want copy-paste to work nicely.


gemmahou:
/lgtm
/approve

Thanks!

PR #2127, opened 2024-06-24
Invoke the expander when a composition changes to update resources.

hankfreund:
### Change description

When the composition controller reconciles a change in a composition, the desired outcome is that affected resources be re-processed to reflect the changes.
Set up a side-channel to the expander and send events on it when compositions change. The expander will then enqueue the necessary events for all relevant resources.

### Tests you have done
Added `TestSimpleCompositionUpdate` to e2e tests and ran the scenario manually. The test exercises updating multiple CRs and ensures CRs for a different GVK aren't affected.

### DIFF
--- a/experiments/compositions/composition/internal/controller/composition_controller.go
+++ b/experiments/compositions/composition/internal/controller/composition_controller.go
@@ -29,6 +29,7 @@ import (
 	pb "google.com/composition/proto"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
+	corev1 "k8s.io/api/core/v1"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -40,6 +41,7 @@ import (
 	"k8s.io/client-go/util/retry"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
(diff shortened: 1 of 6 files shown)


experiments/compositions/composition/internal/controller/expander_reconciler.go
```diff
@@ -584,6 +589,26 @@ func (r *ExpanderReconciler) evaluateAndSavePlan(ctx context.Context, logger log
...
+	inputcrList.SetGroupVersionKind(r.InputGVK)
+	if err := r.List(ctx, inputcrList); err != nil {
+		logger.Error(err, "unable to fetch Input API Objects")
+		return nil
+	}
+	if len(inputcrList.Items) == 0 {
+		return nil
+	}
+	var reqs []reconcile.Request
+	for _, inputcr := range inputcrList.Items {
```
barney-s, line 605: This would cause thundering herd and thats fine. Wondering if the controller Q can handle this for large numbers. Please add a comment to that effect so that we know this is a place where we could optimize later when needed.

experiments/compositions/composition/internal/controller/expander_reconciler.go
```diff
@@ -44,20 +44,25 @@ import (
...
-	Composition types.NamespacedName
+	Scheme            *runtime.Scheme
+	Recorder          record.EventRecorder
+	RESTMapper        meta.RESTMapper
+	Config            *rest.Config
+	Dynamic           *dynamic.DynamicClient
+	InputGVK          schema.GroupVersionKind
+	InputGVR          schema.GroupVersionResource
+	Composition       types.NamespacedName
+	CRDChangedWatcher chan event.GenericEvent
```
barney-s, line 65: Is this meant to indicate Composition change ? 
If so does this name `CompositionChangeWatcher` seems more appropriate ?

experiments/compositions/composition/internal/controller/expander_reconciler.go
hankfreund: Will do.

experiments/compositions/composition/internal/controller/expander_reconciler.go
hankfreund: Ahh, you're right. Must have gotten confused. Fixed.

experiments/compositions/composition/internal/controller/composition_controller.go
```diff
@@ -353,15 +353,15 @@ func (r *CompositionReconciler) processComposition(
...
-		CRDChangedWatcher: r.handoffChannels[gvk],
+		Client:                    r.Client,
+		Recorder:                  r.mgr.GetEventRecorderFor(crd.Spec.Names.Plural + "-expander"),
+		Scheme:                    r.Scheme,
+		InputGVK:                  gvk,
+		Composition:               types.NamespacedName{Name: c.Name, Namespace: c.Namespace},
+		InputGVR:                  gvk.GroupVersion().WithResource(crd.Spec.Names.Plural),
+		RESTMapper:                r.mgr.GetRESTMapper(),
+		Config:                    r.mgr.GetConfig(),
+		ComopsitionChangedWatcher: r.handoffChannels[gvk],
```
barney-s, line 364: typo: Comopstiton/Composition.

PR #2128, opened 2024-06-25
Support observed state validation via GitHub action

maqiuyujoyce:
### Change description

Leveraged the VCR testing and the golden object comparison to verify that observed state fields are populated properly after a successful reconciliation.

Added the observed state test into the presubmit test.

Added instructions about how to generate observed state test data.

### Tests you have done

### DIFF
--- a/scripts/github-actions/observedstate-test.sh
+++ b/scripts/github-actions/observedstate-test.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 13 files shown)


scripts/github-actions/observedstate-test.sh
```diff
@@ -0,0 +1,32 @@
...
+cd ${REPO_ROOT}
+source ${REPO_ROOT}/scripts/fetch_ext_bins.sh && \
+	fetch_tools && \
+	setup_envs
+
+cd ${REPO_ROOT}/
+echo "Running e2e fixtures test with vcr replay mode..."
+
+E2E_KUBE_TARGET=envtest RUN_E2E=1 \
+  E2E_GCP_TARGET=vcr VCR_MODE=replay GOLDEN_OBJECT_CHECKS=1 \
```
acpana, line 30: I may be missing some more context here but why can't we rely on the existing golden objects exports `GOLDEN_OBJECT_CHECKS=1 ` in our fixtures and samples ?

maqiuyujoyce: Good point! I used VCR originally because there was no mock for the specific resources at the time when I started working on it or the mock doesn't populate all the fields I want. Will take a look at the mock.

maqiuyujoyce:
Observed state validation can be achieved as part of the mock test object check. This test is not necessary.

PR #2130, opened 2024-06-25
CLI: generate CRD types

justinsb:
Example usage:

generate-crd --proto-source-path ../proto-to-mapper/build/googleapis.pb \
  --service google.cloud.kms.v1 --version v1beta1 --output-api ~/kcc/k8s-config-connector/


dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+	return nil
+}
+
+func (g *TypeGenerator) visitFile(f protoreflect.FileDescriptor) {
+	for _, msg := range sorted(f.Messages()) {
+		g.visitMessage(msg)
+	}
+
+	krmVersion := "v1beta1"
```
yuwenma, line 59: Shall we set the default to `v1alpha1`? Assuming all direct resources shall start from Alpha.

dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
```diff
@@ -0,0 +1,284 @@
...
+	{
+		protoPackagePath := string(f.Package())
+		protoPackagePath = strings.TrimPrefix(protoPackagePath, "google.")
+		protoPackagePath = strings.TrimPrefix(protoPackagePath, "cloud.")
+		protoPackagePath = strings.TrimSuffix(protoPackagePath, ".v1")
+		protoPackagePath = strings.TrimPrefix(protoPackagePath, ".v1beta1")
```
yuwenma, line 65: Do you mean `TrimSuffix` for v1beta1?

yuwenma (commented):
/approve

Two comments otherwise LGTM. I love this approach! Thank you.
Guess we can also simplify the proto-to-mapper code to use the generatetypes lib.

dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
justinsb: Great catch!  Fixed!

justinsb:
> Guess we can also simplify the proto-to-mapper code to use the generatetypes lib.

Right - I'm actually pulling out "the good bits" and thinking we can probably just delete the proto-to-mapper code when we're done.

dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
justinsb: Good point!

So we have the actual version in a flag passed to the command.  I changed the logic around the filter function so that it now returns the go path in which to write messages, meaning we use the version the user asked for :-)

justinsb:
@yuwenma would be great to get this one in :-)

PR #2131, opened 2024-06-25
monitoringdashboard: add support for text.style

justinsb:
Only implemented in the new reconciler, but (helpfully) ignored by the old reconciler.


yuwenma:
/lgtm
/approve

Looks great!

PR #2132, opened 2024-06-25
docs: update 1.120 release notes to document new fields

justinsb:


justinsb:
Ideally we would do this in each PR, but easy enough to catch up I think (and we may not want to overburden contributors!)

/assign @yuwenma

yuwenma:
/lgtm
/approve

Thanks!

PR #2133, opened 2024-06-25
refactor: Refactor lifecyclehandler test to avoid circular dependency

justinsb:
This test was using the full controller test machinery, which caused
circular dependencies when we used the lifecycle handler in
controllers.

### DIFF
--- a/pkg/test/kubeharness.go
+++ b/pkg/test/kubeharness.go
@@ -1,0 +1,272 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package test
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"os"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/kubebuilder-declarative-pattern/mockkubeapiserver"
+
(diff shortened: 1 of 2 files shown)


pkg/test/kubeharness.go
```diff
@@ -0,0 +1,272 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/kubebuilder-declarative-pattern/mockkubeapiserver"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/dynamic"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/crd/crdloader"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/logging"
+)
+
+// KubeHarness is a test harness that brings up a kube-apiserver (only).
```
justinsb, line 45: This is a subset of our full harness.  I hope we can share the code in future, but wanted to keep this PR short.

pkg/test/kubeharness.go
```diff
@@ -0,0 +1,272 @@
...
+		}
+
+		t.Cleanup(func() {
+			if err := env.Stop(); err != nil {
+				h.Errorf("error stopping envtest environment: %v", err)
+			}
+		})
+
+		h.restConfig = restConfig
+	} else if targetKube := os.Getenv("E2E_KUBE_TARGET"); targetKube == "mock" {
```
yuwenma, line 89: Shall this be `E2E_GCP_TARGET`?

pkg/test/kubeharness.go
```diff
@@ -0,0 +1,272 @@
...
+
+func (h *KubeHarness) waitForCRDReady(obj client.Object) {
+	logger := log.FromContext(h.Ctx)
+
+	apiVersion, kind := obj.GetObjectKind().GroupVersionKind().ToAPIVersionAndKind()
+	name := obj.GetName()
+	namespace := obj.GetNamespace()
+
+	id := types.NamespacedName{Name: name, Namespace: namespace}
+	if err := wait.PollImmediate(2*time.Second, 1*time.Minute, func() (bool, error) {
```
yuwenma, line 196: The `PollImmediate` is deprecated. The lib suggests to use `PollWithContextTimeout` instead.

pkg/test/kubeharness.go
```diff
@@ -0,0 +1,272 @@
...
+	ns.SetName(name)
+	if err := h.GetClient().Create(ctx, ns); err != nil {
+		if !apierrors.IsAlreadyExists(err) {
+			h.Fatalf("error creating namespace %v: %v", name, err)
+		}
+	}
+}
+
+// CreateDummyCRD registers a CRD so we can create objects in tests
+func (h *KubeHarness) CreateDummyCRD(group, version, kind string) {
```
yuwenma, line 238: This is cool. (What do we do today?  I guess we should have such a CreateCRD already, but cannot find it via code search)

yuwenma:
/lgtm
/approve

I like the direction of this PR.

pkg/test/kubeharness.go
yuwenma: Yes! And +1 for smaller PRs

pkg/test/kubeharness.go
justinsb: I don't believe so, this is to enable (fast) mock-kubeapiserver instead of (slow) envtest

yuwenma: ohoh, I see. Both envs use value "mock". 🤣

pkg/test/kubeharness.go
justinsb: I found it ... there's a function called `test.FakeCRDs()` that is called via a convoluted call chain.

I'll harmonize this implement in a follow-on PR.

PR #2137, opened 2024-06-25
fix: change CBWP CRD field to camel case

yuwenma:
The CBWP v1alpha1 is  released in 1.119.0. We can pick up this fix in 1.120.0

## Test
Added a apichecks test to guarantee all CRDs field use camel case rather than pascal case. 

Before fix, the test output is 
```
[refs] crd=cloudbuildworkerpools.cloudbuild.cnrm.cloud.google.com version=v1alpha1: field ".spec.privatePoolV1Config.networkConfig.PeeredNetworkIPRange" should use camel case
```


tests/apichecks/crds_test.go
```diff
@@ -242,3 +242,33 @@ func visitProps(props *apiextensions.JSONSchemaProps, fieldPath string, callback
...
+		t.Fatalf("error loading crds: %v", err)
+	}
+	var errs []string
+	for _, crd := range crds {
+		for _, version := range crd.Spec.Versions {
+			visitCRDVersion(version, func(field *CRDField) {
+				fieldPath := field.FieldPath
+				first := func() int32 {
+					tokens := strings.Split(fieldPath, ".")
+					// Only check the last token to avoid duplication.
```
justinsb, line 258: Not sure duplication is the end of the world here, but works for me!

tests/apichecks/crds_test.go
```diff
@@ -242,3 +242,33 @@ func visitProps(props *apiextensions.JSONSchemaProps, fieldPath string, callback
...
+					for _, first := range tokens[len(tokens)-1] {
+						return first
+					}
+					return 0
+				}()
+				if unicode.IsLetter(first) && unicode.IsUpper(first) {
```
justinsb, line 264: Is there anything that is not a letter?  That sounds like it might cause problems too :-)

tests/apichecks/crds_test.go
```diff
@@ -242,3 +242,33 @@ func visitProps(props *apiextensions.JSONSchemaProps, fieldPath string, callback
...
+					errs = append(errs, fmt.Sprintf("[refs] crd=%s version=%v: field %q should use camel case", crd.Name, version.Name, field.FieldPath))
```
justinsb, line 265: Nit: I was starting to put the "rule name" in square brackets, so maybe `[jsonNaming]`

justinsb (approved):

justinsb:
I have some suggestions on the linter rule, but nothing that blocks the field rename so 👍

tests/apichecks/crds_test.go
yuwenma: oh, that will end up as giving us something below if  the pascal case happens in the second substring place. 
```
 field ".spec.PrivatePoolV1Config" should use camel case
 field ".spec.PrivatePoolV1Config.networkConfig" should use camel case
 field ".spec.PrivatePoolV1Config.networkConfig.peeredNetworkIPRange" should use camel case
```

tests/apichecks/crds_test.go
yuwenma: ah, removed.

tests/apichecks/crds_test.go
justinsb: What you have done is better :-)

justinsb (approved):

PR #2139, opened 2024-06-25
Deduplicate string replacement for startTime

ziyue-101:
### Change description

Normalize startTime in the `NormalizeHTTPLog`, which  is used in both script test and fixture test so we have more normalization than before

### Tests you have done


yuwenma (commented):
/lgtm
/approve

Thanks for the change!

PR #2155, opened 2024-06-26
monitoringdashboard: Add alertChart

justinsb:
- **monitoringdashboard: Add alertChart**


apis/monitoring/v1beta1/alertpolicyref.go
```diff
@@ -0,0 +1,24 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1beta1
+
+type MonitoringAlertPolicyRef struct {
```
yuwenma, line 17: nit: we can put this resource reference to https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/master/apis/refs/v1beta1 . Not a blocker

docs/releasenotes/release-1.120.md
```diff
@@ -25,6 +25,8 @@ output fields from GCP APIs are in `status.observedState.*`
   * Added `spec.severity` field.
 
 * `MonitoringDashboard`
+  * Added `alertChart` widgets.
+  * Added `collapsibleGroup` widgets.
```
yuwenma, line 29: I don't see `collapsibleGroup` in this PR (guess that's due to some diffbase). Hope this won't cause conflict.

pkg/controller/direct/monitoring/dashboard_mappings.go
```diff
@@ -40,6 +40,27 @@ func Empty_ToProto(mapCtx *MapContext, in *krm.Empty) *emptypb.Empty {
...
+		}
+	}
+	return out
+}
+func AlertChart_ToProto(mapCtx *MapContext, in *krm.AlertChart) *pb.AlertChart {
+	if in == nil {
+		return nil
+	}
+	out := &pb.AlertChart{}
+	out.Name = in.AlertPolicyRef.External
```
yuwenma, line 60: Do we need to check the null pointer `in.AlertPolicyRef` here? (I hit a similar issue but use a different auto proto method)

pkg/controller/direct/monitoring/monitoringdashboard_controller.go
```diff
@@ -90,12 +90,12 @@ func (m *dashboardModel) AdapterForObject(ctx context.Context, kube client.Reade
 	if err != nil {
 		return nil, err
 	}
-	projectID := projectRef.ProjectID
-	if projectID == "" {
+	if projectRef == nil || projectRef.ProjectID == "" {
```
yuwenma, line 93: I think we will never hit `projectRef == nil`(?) [code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/references/projectref.go#L35) , or maybe this infers you want to change the `ResolveProject`? Not a blocker, just curious about this change.

pkg/controller/direct/monitoring/refs.go
```diff
@@ -85,10 +89,76 @@ func normalizeResourceName(ctx context.Context, reader client.Reader, src client
...
 
+func normalizeMonitoringAlertPolicyRef(ctx context.Context, reader client.Reader, src client.Object, project references.Project, ref *krm.MonitoringAlertPolicyRef) (*krm.MonitoringAlertPolicyRef, error) {
+	if ref == nil {
+		return nil, nil
+	}
+
+	if ref.Name == "" && ref.External == "" {
+		return nil, fmt.Errorf("must specify either name or external on reference")
+	}
+	if ref.Name != "" && ref.External != "" {
```
yuwenma, line 100: I saw the CRD already use oneOf.
But anyways, I think this entire function is helpful for all ResRef.

pkg/controller/direct/monitoring/refs.go
```diff
@@ -85,10 +89,76 @@ func normalizeResourceName(ctx context.Context, reader client.Reader, src client
...
+	if err != nil {
+		return nil, err
+	}
+
+	alertPolicyProjectID, err := references.ResolveProjectIDForObject(ctx, reader, alertPolicy)
+	if err != nil {
+		return nil, err
+	}
+
+	ref = &krm.MonitoringAlertPolicyRef{
```
yuwenma, line 150: oh, this alertPolicyRef is really complicated (wasn't expect it allows both `projects/<project>/alertPolicies/<id>` and `alertPolicies/<id>`

I have a question about the `AlertPolicy` struct. Since the external has  `projects/<project>/alertPolicies/<id>` I think the resource itself should have a projectRef. But I don't see one https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/e9a7d548ec105d9b36c0c003976b971b8d14027d/pkg/clients/generated/apis/monitoring/v1beta1/monitoringalertpolicy_types.go#L619
Do you have pending PRs or this is a special case?

tests/e2e/normalize.go
```diff
@@ -106,6 +106,20 @@ func normalizeKRMObject(u *unstructured.Unstructured, project testgcp.GCPProject
 	// Specific to Compute SSL Certs
 	visitor.replacePaths[".status.observedState.certificateId"] = "1.719337333063698e+18"
 
+	// Specific to MonitoringDashboard
+	visitor.stringTransforms = append(visitor.stringTransforms, func(path string, s string) string {
+		if strings.HasSuffix(path, ".alertChart.alertPolicyRef.external") {
+			tokens := strings.Split(s, "/")
+			if len(tokens) > 2 {
+				switch tokens[len(tokens)-2] {
```
yuwenma, line 114: I guess the second to last token is always `alertPolicies`, right?

yuwenma (commented):
/lgtm
/approve

/hold Have some minor comments, no blockers. Feel free to unhold

apis/monitoring/v1beta1/alertpolicyref.go
justinsb: I'll move it in a separate PR if that's OK.  I was toying with the idea of grouping them by service though.  Let's have a think in that other PR :-)

docs/releasenotes/release-1.120.md
justinsb: Right - it's because we've already merged it, I just forgot to update the release notes!

pkg/controller/direct/monitoring/dashboard_mappings.go
justinsb: Technically no, because the field is required.  But I agree with you, we should be checking it from a defensive point of view.  Fixed!

pkg/controller/direct/monitoring/monitoringdashboard_controller.go
justinsb: True.  I'm not sure why I did it.  Reverting.

pkg/controller/direct/monitoring/refs.go
justinsb: Right.  It doesn't cost us much, and then we don't need to rely on CRD validation.

BTW, I'm thinking we can greatly reduce code duplication here in future, maybe using AdapterForURL, or maybe just by adding a method to the Adapter called ResolveReference or something.  But ... not a problem right now.

pkg/controller/direct/monitoring/refs.go
justinsb: I _think_ that's one of the "older" resources that uses the project-id annotation instead of projectRef.

I also _think_ that's why project-id is listed here: https://cloud.google.com/config-connector/docs/reference/resource-docs/monitoring/monitoringalertpolicy#annotations

justinsb: (And ResolveProjectIDForObject tries projectRef first, then looks at the project-id annotation)

tests/e2e/normalize.go
justinsb: If it's an alertPolicy, I believe so.    It's normalization for tests, so it's not the end of the world if we're wrong.

(And similar to what you suggested the other day, I think we may be able to factor out some common logic here in future.  The other option though is to build the replacements by looking at the http log, and share them with object normalization, rather than having to figure them out each time.  TBD, I guess!)

justinsb:
Implemented the suggested nil-pointer check changes

apis/monitoring/v1beta1/alertpolicyref.go
yuwenma: Sg!

pkg/controller/direct/monitoring/refs.go
yuwenma: Yes! I like this idea

pkg/controller/direct/monitoring/refs.go
yuwenma: That makes sense. Follow up discussion: since we validate and resolve the project from GCP server now, do you think we shall add the `spec.projectRef` for  `MonitoringAlertPolicy`, and only support the previous project-id annotation for backward compatibility reasons? Not a blocker nor a necessary. But I think we have made a lot  of improvements around project for the monitoring dashboard, why not make one more step further.

yuwenma:
/lgtm

Thanks! One more comment (no blocker) about if we want to add `spec.projectRef`

PR #2161, opened 2024-06-26
chore: don't try to install protoc if already available

justinsb:
This avoids requiring sudo every invocation of make.


yuwenma:
/lgtm
/approve

Thanks

PR #2145, opened 2024-06-26
fix: cbwp update the API to use the auto-generated struct

yuwenma:

PR #2164, opened 2024-06-26
chore: cbwp change status networkRef to external

yuwenma:
Follow up to the offline discussion on https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2145#discussion_r1654354648 

Summary: 
* `spec` should use <resource>Ref
* `status.observedState` should use plain external link.
* Both structs can use `kcc:proto` go tag, but need to be separate to split the `spec` and `status` CRD fields.


pkg/test/resourcefixture/testdata/basic/cloudbuild/v1alpha1/cloudbuildworkerpool/_generated_object_cloudbuildworkerpool.golden.yaml
```diff
@@ -37,9 +37,8 @@ status:
     etag: abcdef123456
     networkConfig:
       egressOption: NO_PUBLIC_EGRESS
+      peeredNetwork: projects/${projectId}/global/networks/computenetwork-${uniqueId}
```
justinsb, line 40: 👍

justinsb (approved):

PR #2166, opened 2024-06-26
chore: run all scenario tests in tests-e2e-scenarios

justinsb:
I am likely missing something, but it looks like we are only running some of our scenario tests here?


yuwenma:
> I am likely missing something, but it looks like we are only running some of our scenario tests here?

Very likely. The scenario tests comes from the previous tests-e2e-fixtures and only contains what's already in the fixtures file.

cheftako:
@justinsb Do we still want this? If so can you rebase?

ldanielmadariaga:

PR #2169, opened 2024-06-27
monitoringdashboard - use etag to detect changes

justinsb:


pkg/controller/direct/monitoring/changedetection.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	if gcpEtag == "" {
+		log.V(2).Info("etag not set in GCP")
+		return true
+	}
+
+	objectEtag := ValueOf(obj.Status.ObservedState.Etag)
+	if objectEtag == "" {
+		objectEtag = ValueOf(obj.Status.Etag)
```
yuwenma, line 79: Ah, so the `objectWithEtagtatus` is only used as a holder of the current GCP object and we do not maintain a `status.observedState` for MD. I like it.

pkg/controller/direct/monitoring/changedetection.go
```diff
@@ -0,0 +1,113 @@
...
+
+// ShouldReconcileBasedOnEtag checks if we should reconcile based on the GCP etag matching the KRM etag.
+// If the etag in KRM status is the same as the GCP etag, we consider the GCP object not to have changed.
+// We also consider the object to have changes if the KRM object generation != observedGeneration (spec changes),
+// and we also reconcile again if the object is not healthy (based on status.conditions).
+//
+// A few problems with the approach:
+// * We miss changes due to labels or annotations.
+// * If there's a change in the GCP object that isn't reflected in etag, we miss that (seems unlikely)
+// * Because we set spec.resourceID, we do an extra reconciliation after first creation (because we bump generation).
```
yuwenma, line 57: I guess this can be resolved by setting `status.externalRef` rather than `spec.resourceID`.

pkg/test/resourcefixture/testdata/basic/monitoring/v1beta1/monitoringdashboard/monitoringdashboardfull/_http.log
```diff
@@ -735,6 +735,555 @@ X-Xss-Protection: 0
 
 ---
```
yuwenma, line 737: Yes! etag is only overridden when we output the http log

yuwenma (commented):
/lgtm
/approve

I like this approach!

pkg/controller/direct/monitoring/changedetection.go
justinsb: Exactly! I think the biggest blocker is the name... I like `status.externalRef` :-)

PR #2170, opened 2024-06-27
Updates on VPC PSC forwarding rule

gemmahou:
### Change description

Follow up on PR#1942

Currently, only regional compute forwarding rule supports target service attachment. Global forwarding rule does not support this feature. Make our samples and fixtures yaml specific to regional forwarding rule to avoid confusion.

More details: https://b.corp.google.com/issues/342233672#comment6

### Tests you have done


pkg/test/constants/constants.go
```diff
@@ -45,7 +45,7 @@ var TestNameRegexesToSkip = []string{
 	".*(containerattachedcluster).*",
 	// Disable due to TF bug https://github.com/hashicorp/terraform-provider-google/issues/16255.
 	// We can't specify labels in the create operation, that causes AssertLabelsMatchAndHaveManagedLabel check to fail.
-	".*(privateserviceconnectforwardingrule).*",
```
maqiuyujoyce, line 48: Could you help me understand why do we want to make this change?

maqiuyujoyce (commented):
/lgtm
/approve

/hold
for a quick question.

Meanwhile, as I may have the question, maybe we should document the context somewhere for other contributors?

pkg/test/constants/constants.go
gemmahou: privateserviceconnectforwardingrule is renamed to regionalforwardingrulepsc

maqiuyujoyce: Gotcha! I got confused because I found the sample is renamed to `regional-forwardingrule-vpc-psc` but didn't realize this is the name for testdata.

maqiuyujoyce (commented):
/hold cancel

gemmahou:
> /lgtm /approve
> 
> /hold for a quick question.
> 
> Meanwhile, as I may have the question, maybe we should document the context somewhere for other contributors?

I added a comment [here](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/8028bb1eb832b180eaa3cd3aacf8b80b3400d107/config/samples/resources/computeforwardingrule/regional-forwarding-rule-vpc-psc/compute_v1beta1_computeforwardingrule.yaml#L40), also added a note in b/342233672. Since our resource docs are generated, I can log a ticket/open an issue to service team/TF to update their docs.

PR #2176, opened 2024-06-27
move refs for monitoringdashboard

justinsb:
- **monitoringdashboard: Add alertChart**
- **refactor: move MonitoringAlertPolicyRef to refs package**


justinsb:
This is moving the MonitoringAlertPolicyRef to `refs/v1beta1`, with the other refs.

TBH I'm not sure whether we should centralize them, or whether there is going to be enough per-service logic that we want to distribute them.  We might want to wait-and-see, but moving them is pretty easy anyway.

/assign @yuwenma

PR #2178, opened 2024-06-27
fix: certificateId -> certificateID

acpana:
Adds resource overrides for `ComputeMangedSSLCertificate` to rename the output only field `certificateId ` to `certificateID`.


yuwenma (commented):
/lgtm

Looks good! I'm curious why you move the three files under resourceoverrides.

pkg/resourceoverrides/compute_overrides.go
acpana, line 1: @yuwenma Answering the question here:

>  I'm curious why you move the three files under resourceoverrides.

I figured this made more sense since we group by service elsewhere (resourcecontexts, servicemappings, etc.)

pkg/resourceoverrides/compute_overrides.go
```diff
@@ -0,0 +1,85 @@
...
+		Kind: "ComputeManagedSSLCertificate",
+		Overrides: []ResourceOverride{
+			mapCertificateIdToCertificateID(),
+		},
+	}
+}
+
+func mapCertificateIdToCertificateID() ResourceOverride {
+	return ResourceOverride{
+		CRDDecorate: func(crd *apiextensions.CustomResourceDefinition) error {
```
jingyih, line 53: For my own eduction, since the CRD field is already renamed to "certificateID" (as the result of the patch in TF), why do we still need CRDDecorate override?

PR #2179, opened 2024-06-27
fix: endpointIp -> endpointIP

acpana:
Adds resource overrides for `CloudIDSEndpoint` to rename the output only field `endpointIp` to `endpointIP`.

### DIFF
--- a/pkg/resourceoverrides/cloudids_overrides.go
+++ b/pkg/resourceoverrides/cloudids_overrides.go
@@ -59,12 +59,25 @@ func mapEndpointIpToEndpointIP() ResourceOverride {
 			return nil
 		},
 		PostActuationTransform: func(original, reconciled *k8s.Resource, tfState *terraform.InstanceState, dclState *unstructured.Unstructured) error {
-			endpointIp, ok := reconciled.Status["endpointIp"]
-			if !ok {
-				return fmt.Errorf("endpointIp field was not populated")
+			observedState, found := reconciled.Status["observedState"]
+			if !found {
+				// if there is no observedState there is nothing to do!
+				return nil
 			}
-			reconciled.Status["endpointIP"] = endpointIp
-			delete(reconciled.Status, "endpointIp")
+			observedStateM, ok := observedState.(map[string]interface{})
+			if !ok {
+				return fmt.Errorf("cannot parse observedState map")
+			}
+			endpointIp, found := observedStateM["endpointIp"]
+			if !found {
+				// field endpointIp not populated
+				return nil
+			}
+
+			observedStateM["endpointIP"] = endpointIp
+			delete(observedStateM, "endpointIp")
+
+			reconciled.Status["observedState"] = observedStateM
 
 			return nil
 		},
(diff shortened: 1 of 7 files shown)


pkg/resourceoverrides/cloudids_overrides.go
```diff
@@ -0,0 +1,65 @@
...
+			}
+
+			// Rename endpointIp to endpointIP
+			observedState.Properties["endpointIP"] = endpointIp
+			delete(observedState.Properties, "endpointIp")
+
+			return nil
+		},
+		PostActuationTransform: func(original, reconciled *k8s.Resource, tfState *terraform.InstanceState, dclState *unstructured.Unstructured) error {
+			reconciled.Status["endpointIP"] = reconciled.Status["endpointIp"]
```
maqiuyujoyce, line 60: Should the logic be:

If `status.endpointIp` exists, add field `status.endpointIP` and remove field `status.endpointIp`?

maqiuyujoyce (commented):
Thank you for adding the resource override, @acpana !

Could you add a observed state test to cover it? Context: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2128

pkg/resourceoverrides/cloudids_overrides.go
```diff
@@ -0,0 +1,72 @@
...
+				observedState.Properties["endpointIP"] = endpointIp
+				delete(observedState.Properties, "endpointIp")
+			}
+
+			return nil
+		},
+		PostActuationTransform: func(original, reconciled *k8s.Resource, tfState *terraform.InstanceState, dclState *unstructured.Unstructured) error {
+			endpointIp, ok := reconciled.Status["endpointIp"]
+			if !ok {
+				return fmt.Errorf("endpointIp field was not populated")
```
maqiuyujoyce, line 64: Why is it an error? I don't think we require all the output-only field be populated? Or maybe I miss anything?

pkg/resourceoverrides/cloudids_overrides.go
acpana: Yes initially the PR was a wip and I got a chance to polish it up now

maqiuyujoyce (commented):
Have you tried to apply this resource locally and verify the output of `kubectl describe` looks as expected? I'm especially curious whether managedFields matches the new field name.

pkg/resourceoverrides/cloudids_overrides.go
acpana: that's a good point! I thought we required all output only fields to be populated but it sounds like that's not the case today.

maqiuyujoyce: AFAIK, output-only fields can be optional in general. We will need to thoroughly understand the API to make a decision.

maqiuyujoyce:
> Could you add a observed state test to cover it? Context: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2128

Thank you for your comment in #2128, @acpana ! As long as there are golden object (that contains the observed state) comparison for this resource in presubmit, I'm good!

pkg/resourceoverrides/cloudids_overrides.go
acpana: `endpointIP` is always present in this case I think. From the integ test `endpointIP` seems to be well populated in the managedFields too.

yuwenma:
/lgtm
/approve

Thank you for the quick fix!

PR #2181, opened 2024-06-27
feat: use resource mapper to generate g3doc for direct resource

yuwenma:
Verified this in CBWP via `make ready-pr`


yuwenma:
It's interesting that the `ServiceMapping` looks like a CRD but not a real CRD 😆

justinsb (approved):

PR #2185, opened 2024-06-28
monitoringdashboard: Add `prometheusQuery` and `outputFullDuration` to timeSeriesQuery

justinsb:


yuwenma (commented):
/lgtm
/approve 


If I understand correctly, adding a "standard" field like this only need contributors to write  the `TimeSeriesQuery_PrometheusQuery_ToProto`  method and uncomment the code (others are auto-generated), right? This is beautiful

PR #2190, opened 2024-06-28
monitoringresource: support incidentList

justinsb:


dev/tools/proto-to-mapper/main.go
```diff
@@ -258,51 +258,64 @@ func (v *visitor) writeTypes(out io.Writer, msg protoreflect.MessageDescriptor)
 			jsonName := field.JSONName()
 			goType := ""
 
-			switch field.Kind() {
-			case protoreflect.MessageKind:
-				goType = protoNameForType(field.Message())
+			if field.IsMap() {
```
yuwenma, line 261: this is a good example to have update the proto-to-mapper logic as we are adding more resource field types. 
 
Are you comparing the generated code with MD cloud documentatino? I'm wondering how you notice the resources is a slice of structs than a slice of strings.

yuwenma (commented):
/lgtm
/approve

PR #2191, opened 2024-06-28
monitoringdashboard: round trip tests

justinsb:
Create a script to run our round-trip tests, also stop excluding some fields we
have added support for, and normalize more values to be valid
(e.g. alignment_period only supports whole seconds)

### DIFF
--- a/docs/releasenotes/release-1.120.md
+++ b/docs/releasenotes/release-1.120.md
@@ -35,6 +35,7 @@ output fields from GCP APIs are in `status.observedState.*`
   * Added `pieChart` widgets.
   * Added `sectionHeader` widgets.
   * Added `singleViewGroup` widgets.
+  * Added `timeSeriesTable` widgets.
 
   * Added `dataSets.targetAxis` and `y2Axis` fields to `xyChart` widgets.
   * Added `id` field to all widgets.
--- a/pkg/controller/direct/monitoring/maputils.go
+++ b/pkg/controller/direct/monitoring/maputils.go
@@ -142,8 +142,16 @@ func Duration_FromProto(mapCtx *MapContext, in *durationpb.Duration) *string {
 		return nil
 	}
 
-	d := in.AsDuration()
-	s := fmt.Sprintf("%vs", d.Seconds())
+	// We want to report the duration without truncation (do don't want to map via float64)
+	s := strconv.FormatInt(in.Seconds, 10)
+	if in.Nanos != 0 {
+		nanos := strconv.FormatInt(int64(in.Nanos), 10)
+		pad := 9 - len(nanos)
+		nanos = strings.Repeat("0", pad) + nanos
+		nanos = strings.TrimRight(nanos, "0")
+		s += "." + nanos
+	}
+	s += "s"
 	return &s
 }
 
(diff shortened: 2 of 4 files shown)


pkg/controller/direct/monitoring/maputils.go
```diff
@@ -142,8 +142,16 @@ func Duration_FromProto(mapCtx *MapContext, in *durationpb.Duration) *string {
 		return nil
 	}
 
-	d := in.AsDuration()
-	s := fmt.Sprintf("%vs", d.Seconds())
+	// We want to report the duration without truncation (do don't want to map via float64)
```
justinsb, line 145: The issue we're fixing is that if translation to float64 products a number that is too big (uses exponents) or too small (nanosecond precision), it wasn't round-tripping.

In theory this is in the production code, in practice it doesn't make a difference for reasonable values.

pkg/controller/direct/monitoring/roundtrip_test.go
```diff
@@ -58,33 +57,17 @@ func FuzzMonitoringDashboardSpec(f *testing.F) {
...
 
-			unimplementedFields.Insert(widgetPath + ".time_series_table")
-
-			unimplementedFields.Insert(widgetPath + ".error_reporting_panel")
-
-			unimplementedFields.Insert(widgetPath + ".incident_list")
-
-			unimplementedFields.Insert(widgetPath + ".id")
+			unimplementedFields.Insert(widgetPath + ".time_series_table.data_sets[].time_series_query.time_series_filter.statistical_time_series_filter")
+			unimplementedFields.Insert(widgetPath + ".time_series_table.data_sets[].time_series_query.time_series_filter.pick_time_series_filter.interval")
```
justinsb, line 68: I haven't found any examples for statisticial_time_series_filter or interval, so I'm not sure if we should try to support it yet.

pkg/controller/direct/monitoring/roundtrip_test.go
```diff
@@ -58,33 +57,17 @@ func FuzzMonitoringDashboardSpec(f *testing.F) {
...
-			unimplementedFields.Insert(widgetPath + ".xy_chart.data_sets[].time_series_query.prometheus_query")
-			unimplementedFields.Insert(widgetPath + ".xy_chart.y2_axis")
-			unimplementedFields.Insert(widgetPath + ".xy_chart.timeshift_duration")
-			unimplementedFields.Insert(widgetPath + ".xy_chart.thresholds[].target_axis")
-
-			// This one might be a bug?
-			unimplementedFields.Insert(widgetPath + ".xy_chart.data_sets[].min_alignment_period.nanos")
-
-			unimplementedFields.Insert(widgetPath + ".scorecard.thresholds[].target_axis")
 			unimplementedFields.Insert(widgetPath + ".scorecard.blank_view")
```
justinsb, line 60: Blank view I think we probably should support, but I think it can go into 1.121 if need be.

docs/releasenotes/release-1.120.md
```diff
@@ -35,6 +35,7 @@ output fields from GCP APIs are in `status.observedState.*`
   * Added `pieChart` widgets.
   * Added `sectionHeader` widgets.
   * Added `singleViewGroup` widgets.
+  * Added `timeSeriesTable` widgets.
```
yuwenma, line 38: I want to merge the PR without this change. Do you mind reverting?

yuwenma (commented):
/lgtm
/approve

This is very cool! Thank you so much

/hold one comment about reverting a release note.

cc @jingyih

docs/releasenotes/release-1.120.md
yuwenma: Okay, just checked that the `timeSeriesTable` field has been checked in. Thank you for updating this note!

PR #2192, opened 2024-06-29
monitoring: add blankView

justinsb:
- **monitoring: add emptyView**

Builds on #2191.


justinsb:
I think this is relatively minor, doesn't _need_ to be in 1.120, but also low risk if we would like it in there.

PR #2194, opened 2024-06-29
kms: add direct types for KMSKeyRing

justinsb:


apis/kms/v1beta1/kmskeyring.go
```diff
@@ -0,0 +1,102 @@
...
+	// +optional
+	ObservedGeneration *int64 `json:"observedGeneration,omitempty"`
+
+	//   // Output only. The resource name for the
+	// //  [KeyRing][google.cloud.kms.v1.KeyRing] in the format
+	// //  `projects/*/locations/*/keyRings/*`.
+	// Name *string `json:"name,omitempty"`
+
+	// The self link of the created KeyRing in the format projects/{project}/locations/{location}/keyRings/{name}.
+	SelfLink *string `json:"selfLink,omitempty"`
```
yuwenma, line 62: or `externalRef`?

yuwenma (commented):
/approve
/lgtm

/hold  One naming comment. feel free to unhold

apis/kms/v1beta1/kmskeyring.go
justinsb: So this PR is just creating go types for the existing structure, I don't think we want to change the schema.  Renaming selfLink -> externalRef might be considered a breaking change, but regardless it likely shouldn't happen in this PR (IMO)

justinsb:
I think I've addressed the hold question, cancelling hold

/hold cancel

PR #2196, opened 2024-06-29
refactor: combine reference resolvers into refs package

justinsb:
This makes it a bit easier to consume, and can enable our Ref objects
to implement an interface for consistent resolution.


yuwenma (commented):
/lgtm
/approve

PR #2197, opened 2024-06-29
fix: alertPolicyRef with external references

justinsb:
I neglected to return the value in the external case, previously.


justinsb:
Assign @yuwenma  this should probably go into 1.120.0 (or 1.120.1 ?)

yuwenma (commented):
/lgtm
/approve

Thank you for the fix before the release!

PR #2203, opened 2024-07-01
mockgcp: support enum-encoding=int alt specifier

justinsb:
Clients can specify a query parameter of $alt=json;enum-encoding=int
(with some escaping) to get integer encoded values from protos.

We support this for more fidelity with mockgcp (though it does make
things a bit harder to read - ideally the clients would not do this,
but if we use the default clients we don't have full control here)


yuwenma (commented):
/lgtm
/approve

oh, right... Thank you!

PR #2204, opened 2024-07-01
tests: add periodic tests that run one service per test

justinsb:


dev/ci/periodics/e2e-service-logging
acpana, line 1: Should this look more like `e2e-service-monitoring` 😛 ?

dev/tasks/run-e2e
```diff
@@ -0,0 +1,63 @@
...
+
+if [[ "${E2E_GCP_TARGET}" == "real" ]]; then
+  if [[ -z "${GCP_PROJECT_ID:-}" ]]; then
+    GCP_PROJECT_ID=$(gcloud config get-value project)
+  fi
+  echo "Using project: ${GCP_PROJECT_ID}"
+
+  TEST_FOLDER_ID=$(gcloud projects describe ${GCP_PROJECT_ID} --format='value(parent.id)')
+  export TEST_FOLDER_ID
+fi
```
acpana, line 56: This will not work in GH actions so we probably want to use `secrets` in the yaml wf definition if we want to run these in GH actions eventually.

gemmahou:
I had something similar months ago([cl](https://cnrm-review.git.corp.google.com/c/cnrm/+/70180)) to be able to run fixtures tests by services(and then run multiple services in parallel). 

The code looks like(might need revise):
```
t.Run("fixtures", func(t *testing.T) {
  fixtures := resourcefixture.Load(t)
  
  groups := sets.NewString()
  for _, fixture := range fixtures {
	  groups.Insert(fixture.GVK.Group)
  }
  
  for _, group := range groups.List() {
	  group := group
	  groupTestName := strings.TrimSuffix(group, ".cnrm.cloud.google.com")
	  t.Run(groupTestName, func(t *testing.T) {
		  // t.Parallel()
		  for _, fixture := range fixtures {
			  if fixture.GVK.Group != group {
				  continue
			  }
			  runTests(ctx, t, fixture, false)
		  }
	  })
  }
})
```
This way, we can use the  parameter `TestAllInSeries/fixtures/logging` to run tests in one service, which would avoid adding an extra environment variable ONLY_TEST_APIGROUP. However, the downside is that we need to always include the service name even when running a single test case, such as `TestAllInSeries/fixtures/logging/logginglogmetric`.

dev/tasks/run-e2e
justinsb: True.  I don't think E2E_GCP_TARGET=real will work in GitHub actions.

justinsb:
> However, the downside is that we need to always include the service name even when running a single test case, such as TestAllInSeries/fixtures/logging/logginglogmetric.

Right - I prefer your approach, my concern is that it just might break all our test infrastructure.  Maybe it would be OK because it's our newer unified tests though ...

I do think we should change the path, but I think we should clean it all up e.g. `TestFixtures/<servicename>/<kind>/<scenario>`, and then all our test fixtures would be structured that way also.  But it feels like a big lift right now.

yuwenma (commented):
/lgtm
/approve

I really like this idea.

yuwenma:
> > However, the downside is that we need to always include the service name even when running a single test case, such as TestAllInSeries/fixtures/logging/logginglogmetric.
> 
> Right - I prefer your approach, my concern is that it just might break all our test infrastructure. Maybe it would be OK because it's our newer unified tests though ...
> 
> I do think we should change the path, but I think we should clean it all up e.g. `TestFixtures/<servicename>/<kind>/<scenario>`, and then all our test fixtures would be structured that way also. But it feels like a big lift right now.

+100 ! And agree it's a big lift for now. @gemmahou Feel free to continue your approach send out PRs.

PR #2221, opened 2024-07-02
Update composition controller RBAC

hankfreund:
### Change description

Update composition controller RBAC
    
Tests all manually grant superuser permissions to the composition controller for simplicity, but that hid the fact that the controller did not have the right permissions to carry out its tasks.
This change removes the broad permissions from the tests and samples and updates the controller RBAC to cover more of the requirements (while still expecting users to grant any additional permissions as necessary).

As a hack to make preview smoother for customers, globally grant superuser to the controller for the short term.

### DIFF
--- a/experiments/compositions/composition/tests/data/TestSimpleAddFacadeField/input.yaml
+++ b/experiments/compositions/composition/tests/data/TestSimpleAddFacadeField/input.yaml
@@ -82,9 +82,6 @@ rules:
   verbs:
   - get
   - update
-- apiGroups: ["*"]
-  resources: ["*"]
-  verbs: ["*"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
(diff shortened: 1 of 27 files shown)


experiments/compositions/composition/tests/data/TestSimpleAddFacadeField/input.yaml
```diff
@@ -99,6 +96,32 @@ subjects:
   name: composition-controller-manager
   namespace: composition-system
 ---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
```
barney-s, line 100: Do we need this to work ?

barney-s: wouldnt * * * in the actual rbac be sufficient. Why do we need this ?

hankfreund: I believe at some point the plan is to remove the  * * * from the cluster role, at which point this will be necessary. I made all the changes without the superuser on the cluster role first, so I know all the tests will work as-is once that workaround is gone.

PR #2224, opened 2024-07-03
fix: truncate monitoringdashboard CRD to fit into cc operator check

yuwenma:
I tried the new CRD size in the operator and it passes the  `TestConfigConnectorE2E` check. 

related to https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2217

Capture some solutions to the root cause:
1.  Switch the cc operator's default applier to use server-side-apply. This requires upstream change to k-d-p  https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/399
2. Switch the cc operator to use the `ApplysetApplier` (use ssa already). This requires not setting the `metadata.managedFields`. But the DCL/TF based controllers already have many handlings around `managedFields`. So this may not be a good option.
3. Turn on the operator check in presubmit to capture this issue earlier.

### DIFF
--- a/apis/monitoring/v1beta1/monitoringdashboard_types.go
+++ b/apis/monitoring/v1beta1/monitoringdashboard_types.go
@@ -57,11 +57,7 @@ type XyChart struct {
 	// +required
 	DataSets []XyChart_DataSet `json:"dataSets,omitempty"`
 
-	// The duration used to display a comparison chart. A comparison chart
-	//  simultaneously shows values from two similar-length time periods
-	//  (e.g., week-over-week metrics).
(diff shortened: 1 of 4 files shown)


justinsb:
I was imagining we just remove all the descriptions with a transformation here: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/8331376698541c2d1b929324c85cd566ad58f7b4/dev/tasks/generate-crds#L51 , but I do agree that a lot of the information here is superfluous.

/approve
/lgtm

justinsb:
(We can always revert this PR and try the other approach in 1.121)

PR #2232, opened 2024-07-03
tests: basic dataform fixture tests

acpana:
Adds basic test for `DataformRepository`, an autogen tf resource. Will update the `update.yaml` and add log, golden files once we have a mock for this.


yuwenma:
ah!  I just realize that you can actually get the golden object against the real GCP server here, so that you can get a fair amount of good values to feed the create.yaml and update.yaml.

yuwenma (commented):
/lgtm
/approve

looks good.

PR #2238, opened 2024-07-04
feat: enable pure direct resource on dynamic test

yuwenma:
### Change description

Fixes #

### Tests you have done


justinsb:
This looks great (but is marked as a draft, so won't approve yet :-) )

yuwenma:
> This looks great (but is marked as a draft, so won't approve yet :-) )

Thanks! Yeah, this PR is approved working for the CBWP Export. But I want to think more on what we can get from the dynamic test for pure direct resource. I also want to take a deeper look at the `Export` before submit the PR.

pkg/krmtotf/resource.go
```diff
@@ -79,6 +79,15 @@ func NewResource(u *unstructured.Unstructured, sm *corekccv1alpha1.ServiceMappin
 
 func NewResourceFromResourceConfig(rc *corekccv1alpha1.ResourceConfig, p *tfschema.Provider) (*Resource, error) {
 	tfResource, ok := p.ResourcesMap[rc.Name]
+	// Pure Direct Resource does not have ResourceMap.
+	if rc.Direct {
```
justinsb, line 83: I am likely missing a trick.  Are we going to have to set `Direct` somewhere?  Can it be set from the direct registry?

yuwenma: This config here is to distinguish a Direct resource that does not need to have any resource config. The direct registry is more for controller, to determine whether to use a Direct Controller or TF/DCL controller when it has multiple options.

PR #2239, opened 2024-07-04
chore: refactor so the version is one file

justinsb:
Avoids the need to split the terrform version-bump.


third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/fwtransport/framework_utils.go
```diff
@@ -19,11 +19,11 @@ import (
...
 const uaEnvVar = "TF_APPEND_USER_AGENT"
 
-//go:embed version.txt
-var kccVersion string
+// KCC Version is included in the UserAgent, and is exported so it can be set by clients.
+var KCCVersion = "dev"
 
 func CompileUserAgentString(ctx context.Context, name, tfVersion, provVersion string) string {
-	ua := fmt.Sprintf("kcc/%s (+https://github.com/GoogleCloudPlatform/k8s-config-connector) kcc/controller-manager/%s", kccVersion, provVersion)
+	ua := fmt.Sprintf("kcc/%s (+https://github.com/GoogleCloudPlatform/k8s-config-connector) kcc/controller-manager/%s", KCCVersion, provVersion)
```
yuwenma, line 26: didn't notice this before. Do we want to send the codename "KCC" in the UserAgent?

yuwenma (commented):
/lgtm
/approve

/hold One question, Feel free to unhold
Thank you so much!

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/fwtransport/framework_utils.go
justinsb: I don't think we want to change it, and if we did I don't think we should change it in this PR

PR #2242, opened 2024-07-04
chore: have codebots create their own commits

justinsb:
Makes it a bit clearer e.g. which branch, which commit mesage etc


yuwenma:
/lgtm
/approve

Thanks!

Issue #2135, opened 2024-06-25
fix: handle uri skeleton non null


acpana:
Today it seems that the `ExpectedSkeleton` is not checked if it is not null BUT the servicemappings lists `idTemplateCanBeUsedToMatchResourceName == false` . We should at least throw a warning in this scenario, otherwise the `ExpectedSkeleton` is not checked which may lead contributors to believe that their test is passing!

acpana:
Example:

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/2f4f2c6ee65034f7bbb4b4f62d2ba15842ff114d/config/servicemappings/vertexai.yaml#L20-L31

But the `ExpectedSkeleton` is not null

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/2f4f2c6ee65034f7bbb4b4f62d2ba15842ff114d/pkg/resourceskeleton/testdata/uri-skeleton.yaml#L1114-L1122

This stood out to me for the "malformed" schema and then I realized that we don't check the yamls if the `idTemplateCanBeUsedToMatchResourceName` is unset:

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/2f4f2c6ee65034f7bbb4b4f62d2ba15842ff114d/pkg/resourceskeleton/resourceskeleton_test.go#L74-L75

ldanielmadariaga:
This item has been open for >12 months. As part of scheduled repository maintenance, this Issue/PR will be closed. If this is still a critical ask, please re-open and work with your Google Cloud Account team to prioritize it with GCP KCC Product Management.

PR #2244, opened 2024-07-06
Fix skeleton test for Vertex AI resources

jingyih:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/config/servicemappings/vertexai.yaml
+++ b/config/servicemappings/vertexai.yaml
@@ -64,7 +64,7 @@ spec:
     - name: google_vertex_ai_endpoint
       kind: VertexAIEndpoint
       idTemplate: "projects/{{project}}/locations/{{region}}/endpoints/{{name}}"
-      idTemplateCanBeUsedToMatchResourceName: false
+      idTemplateCanBeUsedToMatchResourceName: true
       resourceAvailableInAssetInventory: true
       v1alpha1ToV1beta1: true
       storageVersion: v1alpha1
@@ -116,7 +116,7 @@ spec:
     - name: google_vertex_ai_index
       kind: VertexAIIndex
       idTemplate: "projects/{{project}}/locations/{{region}}/indexes/{{name}}"
-      idTemplateCanBeUsedToMatchResourceName: false
+      idTemplateCanBeUsedToMatchResourceName: true
       resourceAvailableInAssetInventory: false
       v1alpha1ToV1beta1: true
       storageVersion: v1alpha1
--- a/config/servicemappings/vertexai.yaml
+++ b/config/servicemappings/vertexai.yaml
@@ -116,7 +116,7 @@ spec:
     - name: google_vertex_ai_index
       kind: VertexAIIndex
       idTemplate: "projects/{{project}}/locations/{{region}}/indexes/{{name}}"
-      idTemplateCanBeUsedToMatchResourceName: true
+      idTemplateCanBeUsedToMatchResourceName: false
       resourceAvailableInAssetInventory: false
       v1alpha1ToV1beta1: true
       storageVersion: v1alpha1
--- a/pkg/resourceskeleton/testdata/uri-skeleton.yaml
+++ b/pkg/resourceskeleton/testdata/uri-skeleton.yaml
@@ -1044,17 +1044,7 @@
       region: us-central1
   ResourceConfigId: VertexAIEndpoint
   URI: "https://us-central1-aiplatform.googleapis.com/v1beta1/projects/kcc-test/locations/us-central1/endpoints/vertexaiendpoint-test"
-- ExpectedSkeleton:
-    apiVersion: vertexai.cnrm.cloud.google.com/v1beta1
-    kind: VertexAIIndex
-    metadata: {}
-    spec:
-      projectRef:
-        external: kcc-test
-      region: us-central1
-    status:
-      observedState:
-        name: vertexaiindex-test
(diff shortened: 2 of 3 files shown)


jingyih:
@maqiuyujoyce Could you help take a look if this PR makes sense?

config/servicemappings/vertexai.yaml
```diff
@@ -64,7 +64,7 @@ spec:
     - name: google_vertex_ai_endpoint
       kind: VertexAIEndpoint
       idTemplate: "projects/{{project}}/locations/{{region}}/endpoints/{{name}}"
-      idTemplateCanBeUsedToMatchResourceName: false
+      idTemplateCanBeUsedToMatchResourceName: true
```
maqiuyujoyce, line 67: Is this change on purpose? This property/field is usually set to true when we want to enable import/export tool for this resource. Or is it configured because we want to use export for vertex AI resources in e2e tests?

pkg/resourceskeleton/testdata/uri-skeleton.yaml
```diff
@@ -1111,34 +1111,30 @@
...
-    apiVersion: vertexai.cnrm.cloud.google.com/v1beta1
-    kind: VertexAIDataset
-    spec:
-      projectRef:
-    external: kcc-test
-    location: us-central1
+- ExpectedSkeleton: null # VertexAIDataset does not support import
   ResourceConfigId: VertexAIDataset
-  URI: "https://us-central1-aiplatform.googleapis.com/v1beta1/projects/kcc-test/locations/us-central1/datasets/1234567890?alt=json"
+  URI: ""
```
maqiuyujoyce, line 1116: Looking at other examples, URI may or may not be kept when ExpectedSkeleton is null. Do you mind adding a unit test to ensure the behavior is consistent? I think either not setting it (if possible) or setting it to an empty string should work.

pkg/resourceskeleton/testdata/uri-skeleton.yaml
```diff
@@ -1111,34 +1111,30 @@
...
+        external: kcc-test
+      region: us-central1
   ResourceConfigId: VertexAIEndpoint
   URI: "https://us-central1-aiplatform.googleapis.com/v1beta1/projects/kcc-test/locations/us-central1/endpoints/vertexaiendpoint-test"
 - ExpectedSkeleton:
     apiVersion: vertexai.cnrm.cloud.google.com/v1beta1
     kind: VertexAIIndex
-    metadata:
-      name: vertexaiindex-test
+    metadata: {}
```
maqiuyujoyce, line 1131: Why is the metadata section cleared up?

maqiuyujoyce (commented):
LGTM

config/servicemappings/vertexai.yaml
jingyih: I am not familiar with this field. But according to #2135, if this field is set to `false`, test `TestNewFromURI` will skip on the resource?

maqiuyujoyce: You are right. And the field was designed for the export/import tool: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/apis/core/v1alpha1/servicemapping_types.go#L88

pkg/resourceskeleton/testdata/uri-skeleton.yaml
jingyih: Related to the previous comment, the test `TestNewFromURI` is currently skipped for `VertexAIDataset`. Do we still need URI?

pkg/resourceskeleton/testdata/uri-skeleton.yaml
jingyih: As the result of enabling the test `TestNewFromURI` for resource (by setting `idTemplateCanBeUsedToMatchResourceName` to `true`), this is the expected output of the test. It is an output only field in [API](https://cloud.google.com/vertex-ai/docs/reference/rest/v1beta1/projects.locations.endpoints). So it ended up showing up under `status.observedState.name`. Not sure if I miss configured the service mapping?

pkg/resourceskeleton/testdata/uri-skeleton.yaml
maqiuyujoyce: I don't think we need it!

jingyih: Just double check on my understanding. I saw that resource `VertexAIDataset` does not support import. Therefore we should set `idTemplateCanBeUsedToMatchResourceName` to `false` for this resource?

https://registry.terraform.io/providers/hashicorp/google-beta/4.84.0/docs/resources/vertex_ai_dataset#import

maqiuyujoyce: In most cases, you are correct. More specifically, the decision flow I have for this field is:

1. Do I need to support export for this resource? Yes: Go to step 2. No: Set the value to `false`.
2. Can the `IDTemplate` be used to match the resource's URL? Yes: Set the value to `true`. No: Go to step 3.
3. Figure out the export use case / design for this resource.

Whether TF support import of this resource or not can be helpful to answer the question in step 2 above.

pkg/resourceskeleton/testdata/uri-skeleton.yaml
jingyih: Thanks! I set `idTemplateCanBeUsedToMatchResourceName: false` for VertexAIIndex.

maqiuyujoyce (commented):
/lgtm
/approve

PR #2251, opened 2024-07-09
feat: enable golden export for direct controller (no TF dependency)

yuwenma:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/pkg/controller/direct/cloudbuild/workerpool_controller.go
+++ b/pkg/controller/direct/cloudbuild/workerpool_controller.go
@@ -296,7 +296,25 @@ func (a *Adapter) Update(ctx context.Context, u *unstructured.Unstructured) erro
 }
 
 func (a *Adapter) Export(ctx context.Context) (*unstructured.Unstructured, error) {
-	return nil, nil
+	if a.actual == nil {
+		return nil, fmt.Errorf("`Find()` not called")
+	}
+	u := &unstructured.Unstructured{}
+
+	wp := &krm.CloudBuildWorkerPool{}
+	mapCtx := &direct.MapContext{}
+	wp.Spec = direct.ValueOf(CloudBuildWorkerPoolSpec_FromProto(mapCtx, a.actual))
+	if mapCtx.Err() != nil {
+		return nil, mapCtx.Err()
+	}
+	wp.Spec.ProjectRef = &refs.ProjectRef{Name: a.id.project}
+	wp.Spec.Location = a.id.location
+	obj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(wp)
+	if err != nil {
+		return nil, err
+	}
+	u.Object = obj
+	return u, nil
 }
 
 // Delete implements the Adapter interface.
--- a/tests/e2e/export.go
+++ b/tests/e2e/export.go
@@ -15,18 +15,85 @@
 package e2e
 
 import (
+	"context"
+	"fmt"
 	"path/filepath"
 	"strings"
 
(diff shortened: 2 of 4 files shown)


pkg/controller/direct/cloudbuild/workerpool_controller.go
```diff
@@ -302,7 +302,23 @@ func (a *Adapter) Update(ctx context.Context, u *unstructured.Unstructured) erro
 }
 
 func (a *Adapter) Export(ctx context.Context) (*unstructured.Unstructured, error) {
```
justinsb, line 304: Idea: (unrelated to this PR) we could have this return a runtime.Object, and then have the caller convert it to unstructured if needed.  Would save a few lines in each Export method.

tests/e2e/export.go
```diff
@@ -15,17 +15,57 @@
...
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/cli/cmd/export"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
 )
 
+func directExport(u *unstructured.Unstructured, c client.Client) ([]byte, error) {
```
justinsb, line 33: I _thought_ we had this function somewhere, but .. no objection, lgtm

tests/e2e/export.go
```diff
@@ -15,17 +15,57 @@
...
+	ctx := context.TODO()
```
justinsb, line 34: Nit: probably easier to accept a `ctx context.Context` parameter on new methods (the caller can always pass context.TODO if needed)

tests/e2e/export.go
```diff
@@ -15,17 +15,57 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	found, err := a.Find(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if !found {
+		return nil, fmt.Errorf("GVK %s '%v' is not found", u.GroupVersionKind(), u.GetName())
```
justinsb, line 50: Nit: a message like `"GCP resource for %v with name %q not found", gvk.Kind, u.GetName` might be clearer.  My thought is that it's not the GVK that wasn't found (in the registry), it's the underlying GCP object that wasn't found.

tests/e2e/export.go
```diff
@@ -15,17 +15,57 @@
...
+	}
+
+	unst, err := a.Export(ctx)
+	if err != nil {
+		return nil, err
+	}
+	unst.SetGroupVersionKind(gvk)
+	unst.SetName(u.GetName())
+	if u.GetNamespace() != "" {
+		unst.SetName(u.GetName())
```
justinsb, line 60: I don't follow this logic... should it be SetNamespace?

justinsb:
A few nits but lgtm.  (Tests aren't passing though)

pkg/controller/direct/cloudbuild/workerpool_controller.go
yuwenma: Yes! This also makes the golden export more "correct". Right now, I have to re-write the name and namespace for golden export but this could miss other metadata like labels and annotations.

ldanielmadariaga:

PR #2253, opened 2024-07-09
feat [compositions]: Simple CEL expander

barney-s:
### Change description

- Adding a simple CEL based expander.
- Uses CEL engine to evaluate substitutions in the template.
- Unit-tests and kind based tests added
- GRPC service, deployments .. added
- Expanderversion to register with composition engine added
- First CEL Sample added in samples


experiments/compositions/expanders/cel-expander/config/expanders/cel-v0.0.1.yaml
```diff
@@ -0,0 +1,115 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: cel-v0.0.1
+  namespace: system
+  labels:
+    control-plane: expander-cel-v0.0.1
```
xiaoweim, line 21: So does it mean we can only use expander-cel for versionv0.0.1 since it is hardcoded here?
if we want to use newer versions, do we need to change this field and the file name?

experiments/compositions/expanders/cel-expander/config/release/kustomization.yaml
```diff
@@ -0,0 +1,31 @@
...
+- ../crd
+- ../expanders
+- ../expander-versions
+
+patches:
+- patch: '[{"op": "replace", "path": "/spec/template/spec/containers/0/image",
+    "value": "gcr.io/krmapihosting-release/expander-cel:v0.0.1"}]'
+  target:
+    kind: Deployment
+    name: cel-v0.0.1
```
xiaoweim, line 28: Similar here, seems the expander version is hardcoded?

experiments/compositions/expanders/cel-expander/pkg/resource/resource.go
```diff
@@ -0,0 +1,98 @@
...
+			case types.Double:
+				vars[variable.Expression] = fmt.Sprintf("%f", v.Value().(float64))
+			case bool:
+				vars[variable.Expression] = fmt.Sprintf("%t", v)
+			case types.Bool:
+				vars[variable.Expression] = fmt.Sprintf("%t", v.Value().(bool))
+			case types.Null:
+				vars[variable.Expression] = ""
+			default:
+				return fmt.Errorf("unknown variable type: %v: type=%v", variable.ResolvedValue, v)
```
xiaoweim, line 74: I may be wrong but I thought there's some function like fmt.Sprintf("%#v", v) that just convert any type to string?

experiments/compositions/samples/FirstCELComposition/composition/teampage.yaml
```diff
@@ -0,0 +1,165 @@
+# Copyright 2024 Google LLC
```
xiaoweim, line 1: will this scenario be included in the private preview user guide?

experiments/compositions/expanders/cel-expander/config/expanders/cel-v0.0.1.yaml
barney-s: ah good point. This is just a label. Needs to be fixed.

experiments/compositions/expanders/cel-expander/config/release/kustomization.yaml
barney-s: This is ok. We can override this once we add a build pipeline.

experiments/compositions/samples/FirstCELComposition/composition/teampage.yaml
barney-s: Not for private preview. I would consider this OSS integration work. And would be part of Public preview.,

experiments/compositions/expanders/cel-expander/pkg/cel/engine.go
```diff
@@ -0,0 +1,59 @@
+package cel
+
+import (
+	"github.com/google/cel-go/cel"
+	"github.com/google/cel-go/common/types/ref"
+	"github.com/wzshiming/easycel"
```
justinsb, line 6: No objections to starting here, but we should probably replace this library with something more kubernetes

justinsb:
So I think this is opt-in so I think we should merge & iterate.

I would like to see us get some golden testing in place, so we can actually see inputs & outputs.  Also I think that will be useful for composition authors!

PR #2255, opened 2024-07-09
MockGCP: Support for ComputeTargetVPNGateway

gemmahou:
### Change description

Add mock support for ComputeTargetVPNGateway, that's required for ComputeForwardingRule

Cherry-picked code from https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/936

### Tests you have done

### DIFF
--- a/tests/e2e/normalize.go
+++ b/tests/e2e/normalize.go
@@ -106,6 +106,7 @@ func normalizeKRMObject(u *unstructured.Unstructured, project testgcp.GCPProject
 	// Specific to Compute
 	visitor.replacePaths[".status.observedState.certificateID"] = "1.719337333063698e+18"
 	visitor.replacePaths[".status.instanceId"] = "1111111111111111"
+	visitor.replacePaths[".status.gatewayId"] = "1.719337333063698e+18"
 
 	// Specific to MonitoringDashboard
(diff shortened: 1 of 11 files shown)


tests/e2e/normalize.go
```diff
@@ -106,6 +106,7 @@ func normalizeKRMObject(u *unstructured.Unstructured, project testgcp.GCPProject
 	// Specific to Compute
 	visitor.replacePaths[".status.observedState.certificateID"] = "1.719337333063698e+18"
 	visitor.replacePaths[".status.instanceId"] = "1111111111111111"
+	visitor.replacePaths[".status.gatewayId"] = "1.719337333063698e+18"
```
justinsb, line 109: Ouch ... this might be a problem if the GCP IDs overflow int64!

justinsb (approved):

tests/e2e/normalize.go
gemmahou: Fixed in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2262

Should not break anything since we replace the value with fixed string here just for golden comparison, to avoid confusion I had a fix and ID should be an int64

PR #2263, opened 2024-07-09
chore: gitignore generated

acpana:
This is to help with cleanup when running `make generate-pb` as the artifacts probably don't need to live in the tree at `dev/tools/proto-to-mapper` .

### DIFF
--- a/dev/tools/proto-to-mapper/.gitignore
+++ b/dev/tools/proto-to-mapper/.gitignore
@@ -1,2 +1,3 @@
 build/
 third_party/
+generated/


justinsb (approved):

yuwenma:
The problem about git-ignore the generated folder is that developers then cannot notice what has happened. (maybe we can print out the generated files, but it is kind of get rid of the git based control and less intuitive)

yuwenma:
@acpana may I ask why this bothers you? it gives a visualized instruction about where those generated files should be placed, and you don't need to check in the `./dev/tools/proto-to-mapper/generated`

yuwenma:
Actually, if we have the `./dev/tools/proto-to-mapper/generated` write back to the `./apis` directly, it could be a better experience. The only question is how proto-to-mapper can tell which API to write back to (another make rule for single resource?)

PR #2277, opened 2024-07-11
minor fix: Move direct logging api under apis folder

gemmahou:
### Change description

Make `logging` consistent with others, under apis/ not apis/resources

### Tests you have done


acpana:
Thanks loads for submitting this 💯 . 

I actually brought up the structure of `apis` too recently. cc @yuwenma @justinsb 

I'm on the camp that resource apis may better live under `apis/resources/<service>/<version>` as `apis` may hold more than just resource apis (like core for CC/ CCCs resources, resourcerefs, etc.). That said, I don't have too strong an opinion on it 🤷🏼‍♂️ so I'll defer to y'all.

yuwenma:
> Thanks loads for submitting this 💯 .
> 
> I actually brought up the structure of `apis` too recently. cc @yuwenma @justinsb
> 
> I'm on the camp that resource apis may better live under `apis/resources/<service>/<version>` as `apis` may hold more than just resource apis (like core for CC/ CCCs resources, resourcerefs, etc.). That said, I don't have too strong an opinion on it 🤷🏼‍♂️ so I'll defer to y'all.

Yes, I think Alex and I have a discussion in doc comments. I won't worry too much about code structure/refactoring right now. Because our goal by the end of Q3 is to fully explore the apis, and I believe we will encounter more complicated API package structure soon. I think @jasonvigil has hit one on cloudsql already. So even if we find a good code structure that fit for today, it may be out of date soon. Thus, I'd like to wait and gather more inputs and then we can revisit this by the end of Q3. 

One big advantage of this PR is that our latest CLI can write the api file with the `kcc:proto` tag directly to the `./apis/<service>/<version>/` directory. Thus I think this PR is more handy and fixes more than "minor" things.  I'd like to have it go in.  Thank you, @acpana and @gemmahou for this discussion!

PR #2278, opened 2024-07-11
refactor: factor out maputils

acpana:
Proof of concept of sharing the map utils and other utils across all the `direct` package.


acpana:
/hold

A few more places that can be factored out if this is something we want to do!

yuwenma:
Actually I think we shall move the cbwp mapping to the direct controller dir (but since we are still developing the mapper tool, this may not be done anyways 😆 ).

acpana:
I don't mind taking on the tooling changes for this too.

pkg/controller/direct/maputils.go
acpana, line 1: Also I think it makes sense to do this now as we are starting to add more resources as opposed to doing it later.

justinsb (approved):

PR #2281, opened 2024-07-12
tests: skip update portions of the test if we don't have an update configured.

justinsb:
This may enable us to stop requiring SkipUpdate in future.


pkg/controller/dynamic/dynamic_controller_integration_test.go
```diff
@@ -368,7 +372,7 @@ func testUpdate(ctx context.Context, t *testing.T, testContext testrunner.TestCo
 	// Update resource from test data
 	updateUnstruct := testContext.UpdateUnstruct.DeepCopy()
 	if updateUnstruct == nil {
```
maqiuyujoyce, line 374: As we've added the check above, is this check still needed?

maqiuyujoyce (commented):
/lgtm

I'm okay with not requiring update here and requiring test coverage via other approaches later.

pkg/controller/dynamic/dynamic_controller_integration_test.go
justinsb: Good point; removed it!

pkg/controller/dynamic/dynamic_controller_integration_test.go
```diff
@@ -367,9 +371,6 @@ func testUpdate(ctx context.Context, t *testing.T, testContext testrunner.TestCo
 
 	// Update resource from test data
 	updateUnstruct := testContext.UpdateUnstruct.DeepCopy()
-	if updateUnstruct == nil {
-		t.Fatalf("updateUnstruct is nil for '%v'. should SkipUpdate be set to true in resourcefixture/contexts?", testContext.ResourceFixture.Name)
-	}
```
acpana, line 372: It seems like there's a plan to move away from `SkipUpdate` so disregard if this point will become moot in the future but as I onboarded on the code base, what I liked about this failure is that it gave me a strong signal for how to set up the fixtures tests.

maqiuyujoyce: We probably need a better way to ensure coverage. This change isn't really part of it.

Regarding testdata, if we do want to enforce the existence of update.yaml, we probably should do it via a unit test but not the integration test.

maqiuyujoyce (commented):
/lgtm
/approve

As long as we have the future plan to ensure update coverage, I'm good with it.

PR #2282, opened 2024-07-12
Compositions [feat] Remove hardcoded values and parameterize AppTeam sample

barney-s:
### Change description

Remove hardcoded values from the AppTeam sample.
Parameterize the additional fields.

### DIFF
--- a/experiments/compositions/samples/AppTeam/composition/appteam.yaml
+++ b/experiments/compositions/samples/AppTeam/composition/appteam.yaml
@@ -18,14 +18,13 @@ metadata:
   name: appteams
   namespace: default
 spec:
-  inputAPIGroup: appteams.facade.facade
+  inputAPIGroup: appteams.idp.mycompany.com
   namespaceMode: explicit
   expanders:
   - type: jinja2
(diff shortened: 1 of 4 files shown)


experiments/compositions/samples/AppTeam/composition/appteam.yaml
```diff
@@ -237,8 +235,17 @@ spec:
...
+              adminUser:
+                type: string
+              billingUser:
+                type: string
+              folder:
+                type: string
             required:
             - project
+            - adminUser
+            - billingAccount
```
xiaoweim, line 247: is it possible to make the billingAccount optional? So that in the automated e2e test, I do not have to specify a billingAccount in the facadeCR

xiaoweim: actually it is okay, I can alter the CRD a little bit in the test

barney-s: For this composition (which includes the bucket) we need a billing.

PR #2283, opened 2024-07-12
feat: direct dataformrepository ctrl

acpana:
Add `direct` controller for `DataformRepository`.

### DIFF
--- a/pkg/controller/direct/dataform/repository_controller.go
+++ b/pkg/controller/direct/dataform/repository_controller.go
@@ -1,0 +1,256 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dataform
+
+import (
+	"context"
+	"fmt"
+
+	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/dataform/v1alpha1"
+	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/config"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
+
+	gcp "cloud.google.com/go/dataform/apiv1beta1"
+	dataformpb "cloud.google.com/go/dataform/apiv1beta1/dataformpb"
+	"google.golang.org/api/option"
+	"google.golang.org/protobuf/types/known/fieldmaskpb"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const ctrlName = "dataform-controller"
+
+func init() {
+	registry.RegisterModel(krm.DataformRepositoryGVK, NewModel)
+}
+
+func NewModel(ctx context.Context, config *config.ControllerConfig) (directbase.Model, error) {
+	return &model{config: *config}, nil
+}
+
+var _ directbase.Model = &model{}
+
+type model struct {
+	config config.ControllerConfig
+}
+
+func (m *model) client(ctx context.Context) (*gcp.Client, error) {
+	var opts []option.ClientOption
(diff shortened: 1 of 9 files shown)


pkg/controller/direct/dataform/repository_controller.go
```diff
@@ -0,0 +1,253 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const ctrlName = "dataform-controller"
+
+func init() {
+	registry.RegisterModel(krm.GroupVersionKind, NewModel)
```
justinsb, line 42: Nit: This will have to be DataFormGroupVersionKind (eventually!)

pkg/controller/direct/dataform/repository_controller.go
```diff
@@ -0,0 +1,253 @@
...
+	if err != nil {
+		return nil, err
+	}
+	projectID := projectRef.ProjectID
+	if projectID == "" {
+		return nil, fmt.Errorf("cannot resolve project")
+	}
+
+	// Get location
+	location := obj.Spec.Region
```
justinsb, line 99: Nit: Even if we have a schema rule, I like to validate that it is not "", just so we get a nice error message

pkg/controller/direct/dataform/repository_controller.go
```diff
@@ -0,0 +1,253 @@
...
+
+var _ directbase.Adapter = &Adapter{}
+
+func (a *Adapter) Find(ctx context.Context) (bool, error) {
+	if a.resourceID == "" {
+		return false, nil
+	}
+
+	req := &dataformpb.GetRepositoryRequest{Name: a.fullyQualifiedName()}
+	dataformrepositorypb, err := a.gcpClient.GetRepository(ctx, req)
```
justinsb, line 136: I also like the name "actual", but .. `dataformrepositorypb` works

pkg/controller/direct/dataform/repository_controller.go
```diff
@@ -0,0 +1,253 @@
...
+
+	desired := a.desired.DeepCopy()
+	mapCtx := &direct.MapContext{}
+	resource := DataformRepositorySpec_ToProto(mapCtx, &desired.Spec)
+	if mapCtx.Err() != nil {
+		return fmt.Errorf("converting DataformRepository spec to api: %w", mapCtx.Err())
+	}
+
+	req := &dataformpb.UpdateRepositoryRequest{UpdateMask: updateMask, Repository: resource}
+	_, err := a.gcpClient.UpdateRepository(ctx, req)
```
justinsb, line 198: Just to check - if this returns an operation, you should wait on it.

pkg/controller/direct/dataform/repository_controller.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	return true, nil
+}
+
+func (a *Adapter) fullyQualifiedName() string {
+	return fmt.Sprintf("projects/%s/locations/%s/repositories/%s", a.projectID, a.location, a.resourceID)
+}
+
+func (a *Adapter) getParent() string {
```
justinsb, line 234: Oooh I like this and will be stealing this pattern!

pkg/controller/direct/dataform/repository_controller.go
```diff
@@ -0,0 +1,253 @@
...
+	return fmt.Sprintf("projects/%s/locations/%s", a.projectID, a.location)
+}
+
+func setStatus(u *unstructured.Unstructured, typedStatus any) error {
```
justinsb, line 238: If it's alpha (and therefore uses observedState) we might actually use a different helper `setObservedState`; it's a much simpler function to write

pkg/controller/direct/export.go
```diff
@@ -42,6 +43,9 @@ func Export(ctx context.Context, url string, config *config.ControllerConfig) (*
 
 		u, err := adapter.Export(ctx)
 		if err != nil {
+			if strings.Contains(err.Error(), "not supported") {
```
justinsb, line 46: Do we have to do this?  String matching against errors is pretty much an anti-pattern, only to be done if there's no choice

pkg/controller/direct/registry/registry.go
```diff
@@ -62,8 +63,11 @@ func AdapterForURL(ctx context.Context, url string) (directbase.Adapter, error)
 		}
 		adapter, err := registration.model.AdapterForURL(ctx, url)
 		if err != nil {
-			return nil, err
+			if !strings.Contains(err.Error(), "not supported") {
```
justinsb, line 66: Ditto - any alternatives?

justinsb:
LGTM.  Only blocker is whether we can do something better than string matching against the error

pkg/controller/direct/dataform/repository_controller.go
acpana: It returns an object

pkg/controller/direct/export.go
acpana: I think I'll run back the `return nil, ErrNotImplemented` for now.

pkg/controller/direct/dataform/repository_controller.go
acpana: Is it ok if I follow up here? There's a few changes to the API that are needed for the beta promotion and there's no ObservedState yet in the Status. I can add it with the other API changes.

justinsb: Of course!

justinsb (approved):

PR #2286, opened 2024-07-12
Use direct controller for presubmits

ziyue-101:
### Change description

Fixes #

### Tests you have done


pkg/test/resourcefixture/testdata/basic/gkehub/v1beta1/gkehubfeaturemembership/fullacmgkehubfeaturemembership/_http.log
```diff
@@ -1267,52 +1207,22 @@ User-Agent: kcc/controller-manager DeclarativeClientLib/0.0.1
...
             "syncWaitSecs": "30"
           },
           "sourceFormat": "hierarchy"
         },
-        "hierarchyController": {
-          "enableHierarchicalResourceQuota": false,
-          "enablePodTreeLabels": false,
-          "enabled": false
-        },
+        "hierarchyController": {},
```
ziyue-101, line 1217: This is because of custom diff, the GCP API treats all false == {}.  Cherry picking https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/1959 to make sure they match

pkg/controller/direct/gkehub/featuremembership_controller.go
```diff
@@ -32,7 +33,11 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
 )
 
-const ctrlName = "gkehubfeaturemembership-controller"
+const (
+	ctrlName    = "gkehubfeaturemembership-controller"
+	timeout     = 5 * time.Minute
```
gemmahou, line 38: Just curious if 5 min is enough? I remember gkehubfeaturemembership is a long running resources(takes more than 10min). Shall we have special handling for long running resources?

ziyue-101: /cc @shumiao 
The timeout is for a GKEHubfeature Update call,
What is the recommended timeout period for such operation?

yuwenma: We don't want to do this in a direct controller. Currently you can just get it pass through like the cbwp [controller](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/cloudbuild/workerpool_controller.go#L191C2-L191C9). We can improve this later

ziyue-101: The cbwp controller uses `https://github.com/googleapis/google-cloud-go/tree/main/cloudbuild` to create GCP client. However, the corresponding `gkehub` one misses gkehub feature fields in the repo https://github.com/googleapis/google-cloud-go/tree/main/gkehub. 

By reading the repo description, https://screenshot.googleplex.com/57dR6itytjCs6Ls seems like the repo is still under development

The gkehub controller uses `https://github.com/googleapis/google-api-go-client/tree/main/gkehub/v1beta`, which doesn't have `Wait` built-in

shumiao: Gemma is right that the long running operation's TTL is 20 minutes, meaning in the worst case, it will take that long be marked done, with either a success or a failure code. The client doesn't have `wait` built-in. This is the client side `wait` logic internal google3 controllers share: https://source.corp.google.com/piper///depot/google3/cloud/kubernetes/hub/common/hubclient/tools.go;l=38-50;rcl=657751230. If I understand correctly, they don't set a timeout, but wait for it to complete or until the context times out.

ziyue-101: @yuwenma Friendly ping, how about I go ahead implement the Wait similar to what shumiao shared?

justinsb: Yes, I think you just poll the operation every 2 seconds (for example).  I wouldn't bother to do a backoff, because we're going to make this much better soon (return from the reconciler etc)

If the operation finished with an error, return an error.  If not, return with no error.  controller-runtime will run the retry for you.

ziyue-101: Thanks for providing additional context here. @justinsb  Now the controller polls the operation every 5s and times out after 20 min, which mimics what the Hub client does today. Could you review this PR?

pkg/controller/direct/gkehub/featuremembership_controller.go
```diff
@@ -176,18 +180,48 @@ func (a *gkeHubAdapter) Delete(ctx context.Context, deleteOp *directbase.DeleteO
...
+	mSpecs := feature.MembershipSpecs
+	if mSpecs == nil {
+		mSpecs = make(map[string]featureapi.MembershipFeatureSpec)
+	}
 	// only change the feature configuration for the associated membership
-	feature.MembershipSpecs[a.membershipID] = *a.desired
-	op, err := a.featureClient.Patch(a.featureID, feature).UpdateMask("membershipSpecs").Context(ctx).Do()
+	mSpecs[a.membershipID] = *a.desired
+	feature.MembershipSpecs = mSpecs
+	op, err := a.hubClient.featureClient.Patch(a.featureID, feature).UpdateMask("membershipSpecs").Context(ctx).Do()
```
justinsb, line 190: Nit: can we pass an etag or a similar optimistic concurrency mechanism?

pkg/test/resourcefixture/testdata/basic/gkehub/v1beta1/gkehubfeaturemembership/basicacmgkehubfeaturemembership/_http.log
```diff
@@ -1464,14 +1470,20 @@ X-Xss-Protection: 0
...
 ---
 
-PATCH https://gkehub.googleapis.com/v1beta/projects/example-project-01/locations/global/features/configmanagement?alt=json&updateMask=membershipSpecs
+PATCH https://gkehub.googleapis.com/v1beta/projects/example-project-01/locations/global/features/configmanagement?alt=json&prettyPrint=false&updateMask=membershipSpecs
 Content-Type: application/json
-User-Agent: kcc/controller-manager DeclarativeClientLib/0.0.1
+User-Agent: kcc/controller-manager
 
 {
+  "labels": {
```
justinsb, line 1478: Q: Does this mean that the DCL client did not support labels?

justinsb (approved):

pkg/controller/direct/gkehub/featuremembership_controller.go
ziyue-101: I created https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/2592 to track the work

pkg/test/resourcefixture/testdata/basic/gkehub/v1beta1/gkehubfeaturemembership/basicacmgkehubfeaturemembership/_http.log
ziyue-101: Yes, the logs was generated using the old DCL client and it doesn't have labels.

ziyue-101:
In light of the conversions above. /hold cancel

PR #2287, opened 2024-07-13
Validate format of GoogleServiceAccount in CCC

jingyih:
Create a CCC object with invalid GSA.
```
$ cat ~/tmp/ccc.yaml 
kind: ConfigConnectorContext
metadata:
    name: configconnectorcontext.core.cnrm.cloud.google.com
    namespace: config-control
spec:
    googleServiceAccount: foo@bar.com

$ k apply -f ~/tmp/ccc.yaml 
configconnectorcontext.core.cnrm.cloud.google.com/configconnectorcontext.core.cnrm.cloud.google.com created
```

CCC Status:
```
$ k describe -f ~/tmp/ccc.yaml 
(...)
Spec:
  Google Service Account:  foo@bar.com
Status:
  Errors:
    error during reconciliation: invalid GoogleServiceAccount format for "foo@bar.com"
  Healthy:  false
Events:
  Type     Reason        Age                From                               Message
  ----     ------        ----               ----                               -------
  Warning  UpdateFailed  4s (x13 over 24s)  configconnectorcontext-controller  error during reconciliation: invalid GoogleServiceAccount format for "foo@bar.com"
```

KCC operator logs:
```
{"severity":"error","timestamp":"2024-07-13T08:40:57.453Z","msg":"preflight check failed, not reconciling","controller":"configconnectorcontext-controller","controllerGroup":"core.cnrm.cloud.google.com","controllerKind":"ConfigConnectorContext","ConfigConnectorContext":{"name":"configconnectorcontext.core.cnrm.cloud.google.com","namespace":"config-control"},"namespace":"config-control","name":"configconnectorcontext.core.cnrm.cloud.google.com","reconcileID":"7be584f2-b021-4adb-9e3e-14123b343030","error":"invalid GoogleServiceAccount format for \"foo@bar.com\""}
```

### DIFF
--- a/operator/pkg/preflight/configconnectorcontextchecker.go
+++ b/operator/pkg/preflight/configconnectorcontextchecker.go
@@ -17,6 +17,7 @@ package preflight
 import (
 	"context"
 	"fmt"
+	"regexp"
 
 	corev1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/apis/core/v1beta1"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/k8s"
@@ -52,5 +53,21 @@ func (c *ConfigConnectorContextChecker) Preflight(_ context.Context, o declarati
 		return fmt.Errorf("spec.billingProject must be set if spec.requestProjectPolicy is set to %v", k8s.BillingProjectPolicy)
 	}
 
+	if err := validateGSAFormat(ccc.Spec.GoogleServiceAccount); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func validateGSAFormat(gsa string) error {
+	if gsa == "" { // GoogleServiceAccount is a required field. We do not need to fail here.
+		return nil
+	}
+	validGSAPattern := `^[A-Za-z0-9._%+\-]+@[a-z0-9.\-]+\.gserviceaccount.com$`
+	emailRegex := regexp.MustCompile(validGSAPattern)
+	if !emailRegex.MatchString(gsa) {
+		return fmt.Errorf("invalid GoogleServiceAccount format for %q", gsa)
+	}
 	return nil
 }
--- a/operator/pkg/preflight/configconnectorcontextchecker_test.go
+++ b/operator/pkg/preflight/configconnectorcontextchecker_test.go
@@ -169,3 +169,49 @@ func TestConfigConnectorContextChecker(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateGSAFormat(t *testing.T) {
+	tests := []struct {
+		name string
+		gsa  string
+		err  error
+	}{
+		{
+			name: "empty",
+			gsa:  "",
+			err:  nil,
+		},
+		{
+			name: "valid GSA format",
+			gsa:  "foo@abc.gserviceaccount.com",
+			err:  nil,
+		},
+		{
+			name: "valid GSA format",
+			gsa:  "foo@abc.def.gserviceaccount.com",
+			err:  nil,
+		},
+		{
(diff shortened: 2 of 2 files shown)


operator/pkg/preflight/configconnectorcontextchecker.go
```diff
@@ -52,5 +53,21 @@ func (c *ConfigConnectorContextChecker) Preflight(_ context.Context, o declarati
...
 
+	if err := validateGSAFormat(ccc.Spec.GoogleServiceAccount); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func validateGSAFormat(gsa string) error {
+	if gsa == "" { // GoogleServiceAccount is a required field. We do not need to fail here.
```
justinsb, line 64: Nit: do you mean _not_ a required field?  Or do you mean that it's required at the schema level, so this is unreachable?

Either way, I agree that you should just return nil for empty values!

operator/pkg/preflight/configconnectorcontextchecker.go
```diff
@@ -52,5 +53,21 @@ func (c *ConfigConnectorContextChecker) Preflight(_ context.Context, o declarati
...
+		return nil
+	}
+	validGSAPattern := `^[A-Za-z0-9._%+\-]+@[a-z0-9.\-]+\.gserviceaccount.com$`
```
justinsb, line 67: Where did you get this regex from?  I found https://cloud.google.com/iam/docs/reference/rest/v1/projects.serviceAccounts/create which suggested a more restrictive regex for the first bit.

projectID matches description here https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects

I'm actually more concerned about us rejecting _valid_ gcpServiceAccount values though, so as long as our regex is _too_ permissive, that's OK IMO!

justinsb (commented):
LGTM, one nit on the comment and some background on the regex

operator/pkg/preflight/configconnectorcontextchecker.go
jingyih: > Where did you get this regex from

I think I got the regex from a back-and-forth discussion between me and AI... 

> I found https://cloud.google.com/iam/docs/reference/rest/v1/projects.serviceAccounts/create which suggested a more restrictive regex for the first bit.

Good find! We could certainly make the first bit more restrictive to match Cloud IAM, but I feel leaving it less restrictive has more benefits.
 - If Cloud IAM update the regex in the future, our validation will likely still work.
 - IMO the primary goal is to reject really wrong formats, like if user forgot to replace "${GSA}" with the actual email address, it should get rejected here. (today we don't reject it, and annotate KSA with any string user write to this field)

operator/pkg/preflight/configconnectorcontextchecker.go
jingyih: Yeah this code should be unreachable. It is a required field, so empty string will error early during CR creation.

yuwenma: A readability rule of thumb is to always consider the function for its own and always handle the edge cases for itself, it would be easier to maintenance and less error prone if the other parts changed. i.e. somehow we decide to disable the required field check in the CRD level.

jingyih: If we fail here, that effectively means the GoogleServiceAccount is a required field, even if we mark it as optional in the CRD level. So I think we should not fail here.

yuwenma (commented):
/lgtm
/approve

It seems to be reasonable to let the GCP service validate acceptable value because KCC has no logic built around GSA. But I don't have strong opinion here, and it seems to be a nicer user experience.

PR #2288, opened 2024-07-13
Generate CRDs & mappers for BigtableInstance

justinsb:
- **chore: generate CRDs for BigtableInstance from protos**
- **chore: generate mappers for bigtable**


apis/bigtable/v1beta1/bigtableinstance_types.go
```diff
@@ -0,0 +1,191 @@
...
+	*/
+
+	/* NOTYET
+	// ObservedState is the state of the resource as most recently observed in GCP.
+	// +optional
+	ObservedState *BigtableInstanceObservedState `json:"observedState,omitempty"`
+	*/
+}
+
+/* NOTYET
```
cheftako, line 138: If not yet, when?

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_bigtableinstances.bigtable.cnrm.cloud.google.com.yaml
```diff
@@ -70,12 +73,15 @@ spec:
                         cpuTarget:
                           description: The target CPU utilization for autoscaling.
                             Value must be between 10 and 80.
+                          format: int64
```
cheftako, line 76: Is this being explicit or does this represent a change?

dev/tools/proto-to-mapper/main.go
```diff
@@ -268,6 +268,7 @@ func (v *visitor) writeTypes(out io.Writer, msg protoreflect.MessageDescriptor)
 					goType = "map[string]int64"
 				} else {
 					fmt.Fprintf(out, "// TODO: map type %v %v\n", keyKind, valueKind)
```
cheftako, line 270: Do we want to include field or goFieldName or jsonName here?

apis/bigtable/v1beta1/bigtableinstance_types.go
justinsb: So typically we / I have started doing one phase where we get the CRD yaml to match the existing CRD yaml.  It's not a perfect match, the descriptions aren't exact, and we now mark status.observedGeneration as an int64.  But there should be no field changes at this stage (we aren't using a new controller yet)

NOTYET is how I "stash" the fields that are going to be useful as we update the CRD, but we aren't yet adding support.

justinsb: For this instance, we aren't introducing observedState yet, because it's an existing v1beta1 CRD.  The only change we know we need right now for BigtableInstance is around the problems on autoscaling vs numNodes.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_bigtableinstances.bigtable.cnrm.cloud.google.com.yaml
justinsb: Ugh .. good call...

/hold

justinsb:
I think I need to add mappings for the Cluster also - thanks for spotting the int64 change.  It led me to check how this field is mapped in the proto.

/hold

dev/tools/proto-to-mapper/main.go
justinsb: Good call - will do!

justinsb:
Actually, I don't think it matters _yet_ that we aren't handling the clusters, that needs to be done in the controller, not the mappers.

/hold cancel

apis/bigtable/v1beta1/bigtableinstance_types.go
cheftako: Not against having a TODO/NOTYET but I do think we should communicate when we would want to include it.
If we want other people to contribute, having sign posts which indicate if this fields is ready to be exposed (or not) will help.
It can indicate its time or that not to waste your effort yet.

dev/tools/controllerbuilder/README.md
```diff
@@ -0,0 +1,17 @@
+Examples:
```
cheftako, line 1: This feels a bit off. It seems like its copied from the generate.sh file. Do we need to keep this up to date? What do I learn here that I wouldn't get from the script and if there is something maybe we should highlight it? If we want to explain to someone how to run a generate an individual resource, maybe we should package the script into individual resources to make that easy and then have a top level which runs them all?

dev/tools/controllerbuilder/scaffold/apis.go
```diff
@@ -139,7 +139,7 @@ func scaffoldDocFile(path string, cArgs *apis.APIArgs) error {
 	if err := WriteToFile(path, out.Bytes()); err != nil {
 		return err
 	}
-	color.HiGreen("New file added %s!\n", path)
+	color.HiGreen("New file added %s\n", path)
```
cheftako, line 142: I would suggest we use %q for paths. I don't expect spaces, new lines etc in them but better to be safe.

pkg/controller/direct/bigtable/mapper.go
```diff
@@ -0,0 +1,233 @@
...
+	// MISSING: State
+	// MISSING: Type
+	// MISSING: Labels
+	// MISSING: CreateTime
+	// MISSING: SatisfiesPzs
+	return out
+}
+
+func BackupInfo_StartTime_FromProto(mapCtx *direct.MapContext, in *timestamppb.Timestamp) *string {
+	mapCtx.NotImplemented()
```
cheftako, line 59: Do we have a test to ensure that passing a StartTime in a big table resource does not cause an error?

dev/tools/controllerbuilder/README.md
justinsb: Agreed - they should be in generate.sh, and we will DRY generate.sh as we add more types.

I feel like I keep deleting this file, it just won't go away!  Deleted!

dev/tools/controllerbuilder/scaffold/apis.go
justinsb: In this case I removed the `!` so that it is clickable.   Agreed though, fixed

pkg/controller/direct/bigtable/mapper.go
justinsb: We have a fuzz test that checks the reachable functions, yes.  This one is not reachable, because the BackupInfo message is not reachable.  @jingyih is adding reachability testing at the proto level in another PR, so we may be able to not generate these unreachable messages in future (and thus we wouldn't need these methods)

I also think when we have the code to not emit functions that are already defined, we can generate default versions of these functions.  We just don't have enough data yet on what they should look like ... need to do more resources!

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_bigtableinstances.bigtable.cnrm.cloud.google.com.yaml
justinsb: Actually, it's a change, but I think it's a good change.  We should not be using `integer` in CRDs, we should be using int32 or int64.  Here it's an int64 according to the proto.

justinsb: (And we'll validate that it doesn't need to be quoted in our tests)

apis/bigtable/v1beta1/bigtableinstance_types.go
justinsb: Any suggestions?  We at least want to wait till we have the direct controller.  Maybe `TODO-AFTER-DIRECT-CONTROLLER` ?

cheftako: Sure. I'd even be happy with something that said something like we want the direct controller to have a minimum 3 month burn in as a beta resource before we add fields the non direct controller did not support.

PR #2289, opened 2024-07-13
chore: generate types & mappers for RedisCluster

justinsb:
This is (mostly) generated, and then we pull out a few types we want to override.

There is some duplicate generation of functions and types, detecting that and skipping duplicates that is in a separate branch/PR.


pkg/controller/direct/redis/cluster/mapper.go
```diff
@@ -0,0 +1,78 @@
...
+import (
+	"time"
+
+	pb "cloud.google.com/go/redis/cluster/apiv1/clusterpb"
+	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/redis/v1alpha1"
+	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+func Cluster_CreateTime_FromProto(mapCtx *MapContext, in *timestamppb.Timestamp) *string {
```
justinsb, line 26: I think we may be able to automatically generate these (or automatically generate them, then allow overrides) ... but I want to collect a few more examples first

dev/tools/proto-to-mapper/main.go
```diff
@@ -755,8 +755,10 @@ func (v *visitor) writeMapFunctions() {
 			v.generatedFiles[k] = out
 
 			out.contents.WriteString(fmt.Sprintf("package %s\n\n", lastGoComponent(goPackage)))
-			out.contents.WriteString(fmt.Sprintf("import pb %q\n\n", "cloud.google.com/go/monitoring/dashboard/apiv1/dashboardpb"))
-			out.contents.WriteString(fmt.Sprintf("import krm %q\n\n", "github.com/GoogleCloudPlatform/k8s-config-connector/apis/monitoring/v1beta1"))
+			out.contents.WriteString("import (\n")
```
justinsb, line 758: This is what gofmt does, saves us a gofmt (as do some of the tab changes)

dev/tools/controllerbuilder/template/apis/doc.go
```diff
@@ -15,8 +14,6 @@ const DocTemplate = `
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-{{- if .PackageProtoTag }}
-// +kcc:proto={{ .PackageProtoTag }}
-{{- end }}
+{{ if .PackageProtoTag }}// +kcc:proto={{ .PackageProtoTag }}{{ end }}
```
justinsb, line 17: Avoiding an awkward (lack of) line break

dev/tools/controllerbuilder/pkg/commands/generatetypes/generatetypescommand.go
```diff
@@ -109,18 +109,9 @@ func RunGenerateCRD(ctx context.Context, o *GenerateCRDOptions) error {
 		if strings.HasSuffix(fullName, "Response") {
 			return "", false
 		}
-		if !strings.HasPrefix(fullName, o.ServiceName) {
+		if !strings.HasPrefix(fullName, o.ServiceName+".") {
 			return "", false
 		}
-
-		protoPackagePath = string(msg.ParentFile().Package())
```
justinsb, line 116: We're now driven much more by the flags

dev/tools/controllerbuilder/pkg/commands/generatemapper/generatemappercommand.go
```diff
@@ -93,7 +93,7 @@ func RunGenerateMapper(ctx context.Context, o *GenerateMapperOptions) error {
 		if strings.HasSuffix(fullName, "Response") {
 			return "", false
 		}
-		if !strings.HasPrefix(fullName, o.ServiceName) {
+		if !strings.HasPrefix(fullName, o.ServiceName+".") {
```
justinsb, line 96: Avoids foo.v1alpha1 matching foo.v1 !

dev/tools/controllerbuilder/pkg/codegen/mappergenerator.go
```diff
@@ -499,7 +509,12 @@ func (v *MapperGenerator) writeMapFunctionsForPair(out io.Writer, pair *typePair
 				}
 
 				oneof := protoField.ContainingOneof()
-				if oneof != nil {
+				if protoField.HasOptionalKeyword() {
```
justinsb, line 512: A lot of this complexity is because we are guessing the go field types.  It might easier to actually look at the target go type, and it might also save us some overrides!

dev/tools/controllerbuilder/pkg/codegen/mappergenerator.go
```diff
@@ -174,8 +174,11 @@ func (v *MapperGenerator) GenerateMappers() error {
 			krmPackage := pair.KRMType.GoPackage
 
 			out.contents.WriteString(fmt.Sprintf("package %s\n\n", lastGoComponent(goPackage)))
-			out.contents.WriteString(fmt.Sprintf("import pb %q\n\n", pbPackage))
-			out.contents.WriteString(fmt.Sprintf("import krm %q\n\n", krmPackage))
+			out.contents.WriteString("import (\n")
+			out.contents.WriteString(fmt.Sprintf("\tpb %q\n", pbPackage))
+			out.contents.WriteString(fmt.Sprintf("\tkrm %q\n", krmPackage))
+			out.contents.WriteString(fmt.Sprintf("\t%q\n", "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"))
```
justinsb, line 180: All this direct stuff is using the code that @acpana just refactored out 🎉

pkg/test/fuzz/generate.go
```diff
@@ -46,6 +46,20 @@ func fillWithRandom0(t *testing.T, randStream *rand.Rand, msg protoreflect.Messa
 		return
 	}
 
+	if string(descriptor.FullName()) == "google.protobuf.Timestamp" {
+		count := randStream.Intn(10)
+		// Bias to zero
+		if count > 4 {
+			return
+		}
+		// Generate a "reasonable" timestamp
```
justinsb, line 55: "Unreasonable" dates are (literally) billions of years in the future.    Not a problem in practice, but the fuzzer finds these problems pretty fast if we don't stop them!

pkg/controller/direct/redis/cluster/roundtrip_test.go
```diff
@@ -0,0 +1,148 @@
...
+		outputFields := sets.New(".etag")
+
+		// A few fields are not implemented yet in KRM, don't test them
+		unimplementedFields := sets.New(
+			".name",
+			".labels",
+		)
+
+		// Status fields
+		unimplementedFields.Insert(".discovery_endpoints")
```
justinsb, line 49: This file is not currently generated.  I think it would be nice to collect more examples and then automate.  e.g. I _think_ we can tell which fields are spec and which are status by seeing which ones map to spec and which to status (and if they map to neither, _that's_ an error)

pkg/controller/direct/redis/cluster/mapper.go
```diff
@@ -0,0 +1,79 @@
...
+	"time"
+
+	pb "cloud.google.com/go/redis/cluster/apiv1/clusterpb"
+	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/redis/v1alpha1"
+	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+func Cluster_CreateTime_FromProto(mapCtx *direct.MapContext, in *timestamppb.Timestamp) *string {
```
justinsb, line 27: We should be able to automatically generate these in future (particularly once we're able to skip generation if there is a manually-written function with the same name).  But I want to collect more examples first.

dev/tools/controllerbuilder/pkg/codegen/mappergenerator.go
cheftako, line 180: Why is this a string we need to manually validate? Wouldn't it be better to use direct.GoPackage ?

dev/tools/controllerbuilder/template/apis/doc.go
```diff
@@ -1,7 +1,6 @@
 package apis
 
-const DocTemplate = `
-// Copyright 2024 Google LLC
+const DocTemplate = `// Copyright 2024 Google LLC
```
cheftako, line 3: Would it make sense to construct DocTemplate and GroupVersionInfoTemplate from a new CopyRightTemplate? This both allows us to only fix the copyright in 1 place. It also emphasizes what is different and what is the same between them.

cheftako:
/lgtm
/approve

Flagged a couple of spots I think we may want to improve but nothing which needs to hold this up.

dev/tools/controllerbuilder/template/apis/doc.go
justinsb: Yeah, we can also automatically add the copyright to the file as part of writing it ... I think that's in a PR that's coming!

dev/tools/controllerbuilder/pkg/codegen/mappergenerator.go
justinsb: Not sure I follow - did this comment get moved between lines?

PR #2294, opened 2024-07-15
enh: some template changes

acpana:
Some direct template changes

PR #2295, opened 2024-07-15
enh: check if not supported

acpana:
Goes with https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2294 . If the Export is not supported yet in the controller package return nil.


yuwenma (commented):
/lgtm
/approve

acpana:
Seems like we are going to stick with `nil, nil`.

PR #2298, opened 2024-07-16
mockgcp: more fidelity/tests for ComputeNetwork with autoCreateSubnets

justinsb:
- **mockgcp: fidelity for ComputeNetworks with autoCreateSubnets**
- **tests: create test for ComputeNetwork with autoCreateSubnets=true**


justinsb:
(But note I put this into 1.126, not urgent)

mockgcp/common/workflows/computenetwork_subnetworks.go
```diff
@@ -0,0 +1,109 @@
...
+	return c, nil
+}
+
+func (e *Engine) CreateComputeNetworkSubnetworks(ctx context.Context, projectID string, networkID string) error {
+	subnetsClient, err := e.subnetworksClient(ctx)
+	if err != nil {
+		return err
+	}
+
+	regions := []string{
```
yuwenma, line 44: This list could be out-of-date. I think it is fine, but I think this will give new contributors additional overhead on configuring out where goes wrong if we count "how long it takes in the Step 1. add mockGCP" as a factor to evaluate the SciFi improvements.  

Maybe we can keep the "could be out of date" variables in a shared lib, or add some logging for easy triage? WDYT?

yuwenma (commented):
/lgtm
/approve
/hold 

Want to discuss a little bit more about the contributors' overhead on the region list.

@gemmahou for ack since you are working on this service.

mockgcp/common/workflows/computenetwork_subnetworks.go
justinsb: I think it is going to cause fewer problems than we have today.  I don't think a human will be too confused by the idea that the list of zones changes over time as new GCP zones are added.  If we close the loop and have AI generate it, we can add a note to the prompt.

justinsb:
I rebased this, but the computeserviceattachment changes are huge ... might have to look at what's going on there

mockgcp/mockcompute/networksv1.go
```diff
@@ -207,6 +213,15 @@ func (s *MockService) parseNetworkName(name string) (*networkName, error) {
 	return nil, status.Errorf(codes.InvalidArgument, "name %q is not valid", name)
 }
 
+// parseNetworkSelfLink parses a selfLink string into a networkName.
+// The expected form is `https://www.googleapis.com/compute/beta/projects/*/global/networks/*`.
+func (s *MockService) parseNetworkSelfLink(selfLink string) (*networkName, error) {
+	name := selfLink
+	name = strings.TrimPrefix(name, "https://www.googleapis.com/compute/beta/")
```
gemmahou, line 220: Maybe change `beta` to `v1`? Our mock uses `v1, I think that's the reason why we see so many 400 Bad Request in http logs, and  error message "network \"https://www.googleapis.com/compute/v1/projects/example-project-01/global/networks/${networkID}\" is not valid"

mockgcp/common/workflows/computenetwork_subnetworks.go
justinsb: I moved this into its own package under "mockgcp/common/regions", so we only have to maintain it once

mockgcp/mockcompute/networksv1.go
justinsb: Good call - thanks.  I made it tolerate both beta and v1.  We do now have buildComputeSelfLink, so we know whether it should be beta or v1, but I figure it's better just to accept both (real gcp might do the same thing, I don't know)

PR #2299, opened 2024-07-16
mockgcp: add comments to protos generated from OpenAPI

justinsb:


mockgcp/tools/gapic/pkg/protogen/convert.go
```diff
@@ -553,12 +587,15 @@ func (c *OpenAPIConverter) buildServiceFromOpenAPI(pluralName string, resource *
...
-		serviceID := c.doc.Name
-		prefix := []string{"google", "cloud"}
-
-		name := path.Join(path.Join(prefix...), serviceID, version, serviceID+"pb.proto")
-		packageName := strings.Join(prefix, ".") + "." + serviceID + "." + version
+		// version := c.doc.Version
+		// serviceID := c.doc.Name
+		// prefix := []string{"google", "cloud"}
+
+		tokens := strings.Split(c.protoPackageName, ".")
```
cheftako, line 594: Should probably check the tokens is the expected length.

cheftako: If its more than 2 doesn't that also suggest we may be splitting it incorrectly?

PR #2304, opened 2024-07-17
chore: Generate NetworkConnectivityServiceConnectionPolicy types

justinsb:
- **mockgcp: add comments to protos generated from OpenAPI**
- **mockgcp: generate networkconnectivity from OpenAPI**
- **autogen: add generated networkconnectivity api**
- **chore: types for NetworkConnectivityServiceConnectionPolicy**
- **chore: don't generate types where we already have one**


dev/tools/proto-to-mapper/Makefile
```diff
@@ -9,6 +9,8 @@ generate-pb: install-protoc-linux
 	git clone https://github.com/googleapis/googleapis.git third_party/googleapis || (cd third_party/googleapis && git reset --hard origin/master && git pull)
 	protoc --include_imports --include_source_info \
 		-I ./third_party/googleapis/ \
+		-I ../../../mockgcp/apis \
+		../../../mockgcp/apis/mockgcp/cloud/networkconnectivity/*/*.proto \
```
yuwenma, line 13: Can I ask for more context about this, like why we cannot (or prefer not to) put the proto in `./third_party/` (I'm a little bit concerned on the amount of potential duplicated codes we have in cc)

mockgcp/Makefile
```diff
@@ -123,6 +123,10 @@ generate-protos-from-openapi:
 	mkdir -p apis/mockgcp/cloud/bigquery/v2/
 	cd tools/gapic; go run . --proto-version=2 --proto-package mockgcp.cloud.bigquery.v2 ../../temp/bigquery-v2.json > ../../apis/mockgcp/cloud/bigquery/v2/api.proto
 
+	wget -O temp/networkconnectivity-v1.json https://raw.githubusercontent.com/googleapis/google-api-go-client/main/networkconnectivity/v1/networkconnectivity-api.json
+	mkdir -p apis/mockgcp/cloud/networkconnectivity/v1/
+	cd tools/gapic; go run . --proto-package mockgcp.cloud.networkconnectivity.v1 ../../temp/networkconnectivity-v1.json > ../../apis/mockgcp/cloud/networkconnectivity/v1/networkconnectivity.proto
+
```
yuwenma, line 129: (I see why...)

mockgcp/tools/gapic/pkg/protogen/convert.go
```diff
@@ -669,7 +675,7 @@ func (c *OpenAPIConverter) resolveMessageType(ref string) string {
 	}
 
 	switch ref {
-	case "Operation":
+	case "Operation", "GoogleLongrunningOperation":
```
yuwenma, line 678: it's a little bit hard to trace here. May I ask what the issues are (guess that may be specific to the api from raw json)  ?

yuwenma (commented):
/lgtm
/approve

Thank you!!! I see many cool stuff in this PRs.

dev/tools/controllerbuilder/generate.sh
```diff
@@ -49,3 +49,20 @@ go run . generate-mapper \
     --api-go-package-path github.com/GoogleCloudPlatform/k8s-config-connector/apis \
     --output-dir ~/kcc/k8s-config-connector/pkg/controller/direct/ \
     --api-dir ~/kcc/k8s-config-connector/apis/
+
+# NetworkConnectivity
+go run . generate-types \
+    --proto-source-path ../proto-to-mapper/build/googleapis.pb \
+    --service mockgcp.cloud.networkconnectivity.v1 \
+    --version networkconnectivity.cnrm.cloud.google.com/v1alpha1 \
+    --output-api ~/kcc/k8s-config-connector/apis \
```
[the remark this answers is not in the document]
justinsb: Will do ... we can at least start de-duplicating this function as it gets bigger :-)

dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
```diff
@@ -109,19 +110,31 @@ func (g *TypeGenerator) writeVisitedMessages() {
...
+			GoPackage: goPackage,
+			FileName:  "types.generated.go",
 		}
 		out := g.getOutputFile(k)
 
+		goTypeName := goNameForProtoMessage(msg, msg)
+		skipGenerated := true
+		goType, err := g.findTypeDeclaration(goTypeName, out.OutputDir(), skipGenerated)
+		if err != nil {
+			g.Errorf("lookup up go type: %w", err)
```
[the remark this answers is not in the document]
justinsb: Will fix!

dev/tools/proto-to-mapper/Makefile
justinsb: So we're adding the generated proto here, because Google doesn't ship one for network connectivity (that I could find).  We happen to be reusing the proto we generated from the OpenAPI here, because it is duplicated.  We could regenerate it here.

The challenge is that I think of mockgcp as separate from this, this is part of a tool that generates KCC code.  mockgcp is a testing tool for GCP.  mockgcp, proto-to-mapper/controllerbuilder and KCC itself could all be in separate repos.

mockgcp/tools/gapic/pkg/protogen/convert.go
justinsb: So this is converting an OpenAPI schema to a proto.  That isn't a perfect operation - we have lost some information - not least the proto tag numbers.  We want to generate the right type for an LRO, and this is how LROs show up (at least for this OpenAPI)

PR #2305, opened 2024-07-17
Generate mappers for NetworkConnectivityServiceConnectionPolicy

justinsb:
- **chore: generate mappers for NetworkConnectivityServiceConnectionPolicy**
- **autogen: generate networkconnectivity proto/grpc bindings**
- **chore: handle some edge cases in generator tooling**


justinsb:
/assign @yuwenma 

Looks like there are 4 phases for a brand new resource:

1) types & CRD (no real exit criteria?)
2) mappers (should pass round trip tests)
3) controller & test data (should work with real GCP)
4) mocks (should pass tests with controller, mockgcp and realgcp should be the same)

This is (2).  I'm thinking that we might be able to combine 1 & 2, particularly because there's no exit criteria for (1)

pkg/controller/direct/networkconnectivity/refs.go
yuwenma, line 1: qq: the commented code also look good, why do we not include them?

pkg/controller/direct/networkconnectivity/roundtrip_test.go
yuwenma, line 1: 👍

yuwenma (commented):
/lgtm
/approve

Looks great! Need to take a deeper look at the changes under tools/controllerbuilder. But I think we use that more as experimental code and it shall not block this PR.

PR #2311, opened 2024-07-18
chore: update google APIs in mockgcp

justinsb:
- **chore: create scripts to update APIs in mockgcp**
- **autogen: update mockgcp**


cheftako:
I think it would be best to treat the Google APIs like a library. We should control which exact version we are building against. We should be have a record of which version the Google API was built against for any given KCC release. We should also have tests run any time we upgrade the version is we're building against and be able to reject an upgrade if its causing problems. (At least until we have fixes for it)

justinsb:
> I think it would be best to treat the Google APIs like a library. We should control which exact version we are building against. We should be have a record of which version the Google API was built against for any given KCC release. We should also have tests run any time we upgrade the version is we're building against and be able to reject an upgrade if its causing problems. (At least until we have fixes for it)

Great call.  @jingyih added the ability to pin to a particular SHA, so I updated this PR to source that from a file.  That file is updated by running `dev/tasks/update-mockgcp`.  We have a codebot that will (eventually) run that script periodically and send PRs, via `dev/codebots/update-mockgcp`.  Because it sends PRs, it goes through testing, we have to approve, we can revert etc

dev/tasks/update-mockgcp
```diff
@@ -0,0 +1,44 @@
...
+REPO_ROOT=$(git rev-parse --show-toplevel)
+cd ${REPO_ROOT}
+
+cd ${REPO_ROOT}/mockgcp
+
+# Find latest SHAs
+GOOGLEAPIS_SHA=$(git ls-remote --exit-code https://github.com/googleapis/googleapis  refs/heads/master | awk '{print $1}' )
+
+# Update version markers
+cat > git.versions <<EOF
```
jingyih, line 30: I recommend that we write the GIT SHA in the Makefile rule, as opposed to in the task script. There could be manual runs of the Makefile, in which case we also need to record the SHA. WDYT?

jingyih:
It seems to me that we are pulling protos from googleapis/googleapis in 2 different places, mockgcp and SciFi codegen. Should we consolidate them into one place for better consistency?

justinsb:
> It seems to me that we are pulling protos from googleapis/googleapis in 2 different places, mockgcp and SciFi codegen. Should we consolidate them into one place for better consistency?

Yes, but let's do that separately.  It's not trivial because we rewrite mockgcp (e.g. cloud => mockgcp), but we don't rewrite the SciFi version

jingyih:
> > It seems to me that we are pulling protos from googleapis/googleapis in 2 different places, mockgcp and SciFi codegen. Should we consolidate them into one place for better consistency?
> 
> Yes, but let's do that separately. It's not trivial because we rewrite mockgcp (e.g. cloud => mockgcp), but we don't rewrite the SciFi version

I see. Maybe we can pull from the same git commit?

jingyih:
/lgtm

Let's see what the generated PR looks like, what it breaks, and which part can potentially be fixed by automated tools.

PR #2313, opened 2024-07-18
Update the First GCP sample to use a pre generated CRD

barney-s:
Use a prebuilt CRD instead of Facade builder
- Adjust the CRD group
- Update the readme
- Rename the CRD fields

### DIFF
--- a/experiments/compositions/samples/FirstGCPComposition/composition/cors-bucket.yaml
+++ b/experiments/compositions/samples/FirstGCPComposition/composition/cors-bucket.yaml
@@ -17,7 +17,7 @@ kind: Composition
 metadata:
   name: cors-bucket
 spec:
-  inputAPIGroup: crbuckets.facade.compositions.google.com    # Facade API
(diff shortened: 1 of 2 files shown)


experiments/compositions/samples/FirstGCPComposition/composition/cors-bucket.yaml
```diff
@@ -48,21 +48,50 @@ spec:
...
-        type: integer
\ No newline at end of file
+  conversion:
+    strategy: None
+  group: idp.mycompany.com
+  names:
+    categories:
+    - facade
+    - facades
+    kind: CRBucket
```
justinsb, line 63: Why is it a "CR" bucket?

justinsb (approved):

experiments/compositions/samples/FirstGCPComposition/composition/cors-bucket.yaml
barney-s: CORS + Retention Bucket :)

PR #2318, opened 2024-07-19
Fixes for the first composition sample

barney-s:
Fixes for the first composition sample

- Use generated CRD instead of Facade
- change the CRD group to make it generic
- change the facade to get the list of team members instead of apps
- Update the landing page configmap to use the list of team members instead of apps

### DIFF
--- a/experiments/compositions/samples/FirstComposition/composition/teampage.yaml
+++ b/experiments/compositions/samples/FirstComposition/composition/teampage.yaml
@@ -15,106 +15,135 @@
 apiVersion: composition.google.com/v1alpha1
 kind: Composition
 metadata:
-  name: team-page
+ name: team-page
 spec:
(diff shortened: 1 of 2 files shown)


experiments/compositions/samples/FirstComposition/composition/teampage.yaml
```diff
@@ -15,106 +15,135 @@
 apiVersion: composition.google.com/v1alpha1
 kind: Composition
 metadata:
-  name: team-page
+ name: team-page
```
justinsb, line 18: Nit: I think I liked two spaces better :-)

justinsb:
Would be really nice to have a unit test and check in the output of running a composition (a golden test).  But not a blocker!

/approve
/lgtm

PR #2319, opened 2024-07-19
WIP: Can we run self-hosted runners?

justinsb:
Are they faster?


justinsb:
Looks like the answer is no, that the big win would be replacing envtest with mock-kubeapiserver

/close

PR #2320, opened 2024-07-19
mockgcp: more fidelity for secretmanager

justinsb:
- **mockgcp: update secretmanager**
- **mockgcp: update fidelity of secretmanager API**
- **tests: add updated golden output for secretmanager**


justinsb:
This is because the automatic update PR  #2311 was failing because we were (incorrectly) responding with all fields for secret manager from mockgcp.

mockgcp/mockiampolicies.go
```diff
@@ -75,17 +99,23 @@ func (m *mockIAMPolicies) serveSetIAMPolicy(resourcePath string, httpRequest *ht
 		return &http.Response{StatusCode: http.StatusConflict, Body: responseBody}, nil
 	}
 
-	request.Policy.Version = 3
+	// conditional role bindings must specify version 3
+	hasConditions := false
+	for _, binding := range request.Policy.Bindings {
+		if binding.Condition != nil {
+			hasConditions = true
```
cheftako, line 106: might as well break at this point. No point in going through the rest of the bindings....

mockgcp/mocksecretmanager/secretversions.go
```diff
@@ -286,9 +292,12 @@ func (s *SecretsV1) DestroySecretVersion(ctx context.Context, req *pb.DestroySec
 		return nil, err
 	}
 
+	now := timestamppb.Now()
```
cheftako, line 295: Insert standard diatribe against using actual time in tests.

pkg/test/resourcefixture/testdata/basic/alloydb/v1beta1/alloydbcluster/fullalloydbcluster/_http.log
```diff
@@ -278,11 +278,19 @@ X-Xss-Protection: 0
 GET https://cloudkms.googleapis.com/v1/projects/${projectId}/locations/us-central1/keyRings/kmskeyring-${uniqueId}/cryptoKeys/kmscryptokey-${uniqueId}:getIamPolicy?alt=json&options.requestedPolicyVersion=3&prettyPrint=false
```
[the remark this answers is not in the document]
cheftako: Also do we want to check in all of these _http.log files?

cheftako:
Looks good. Would like answers for Policy.Version and projectId before making it official.

mockgcp/mocksecretmanager/secretversions.go
justinsb: True, we have to deal with it in normalization.  We do want updateTime to be not-a-constant, because we might want to use it for change detection though...

pkg/test/resourcefixture/testdata/basic/alloydb/v1beta1/alloydbcluster/fullalloydbcluster/_http.log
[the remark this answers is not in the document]
justinsb: I think yes to checking in the _http.log files, this is how we can verify what the controller is doing (e.g. is it actually making reasonable calls).  It also costs us basically nothing (a few files in a repo), to give us almost full coverage of the controller behaviour.

$projectId is the placeholder we insert in normalization, so that no matter what project we run this against (real or mock), the _http.log should be the same.  We use the same value as the input parameter because it feels natural, but we could also replace it with a different value if you think that's clearer!

mockgcp/mockiampolicies.go
justinsb: Done!

mockgcp/mockiampolicies.go
```diff
@@ -75,17 +99,23 @@ func (m *mockIAMPolicies) serveSetIAMPolicy(resourcePath string, httpRequest *ht
...
+		}
+	}
+	// GCP returns the version as 1 if there are no conditions
+	if !hasConditions {
+		request.Policy.Version = 1
+	}
```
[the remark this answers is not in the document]
justinsb: I don't know yet.  What I saw is that when the caller passes in a v3 policy, we get back a v1 policy if there are no conditions set.

We should be able to tell as we gather more logs, compare `hack/record-gcp` to `hack/compare-mock`.

PR #2321, opened 2024-07-19
Sample EKS/AKS compositions: Fix input typos, add comments, clean up docs

ostrain:


experiments/compositions/samples/AttachedAKS/aks-alice.md
```diff
@@ -1,21 +1,11 @@
-# Copyright 2024 Google LLC
```
barney-s, line 1: We need a better way to fix this. Meanwhile run "make fmt" at the root directory to autogenerate these.

PR #2322, opened 2024-07-19
Remove 'state-into-spec' annotation from reference docs

maqiuyujoyce:
### Change description

We should avoid mentioning the `state-into-spec` annotation so removing it from reference docs.

### Tests you have done

### DIFF
--- a/scripts/generate-google3-docs/resource-reference/generated/resource-docs/accesscontextmanager/accesscontextmanageraccesslevel.md
+++ b/scripts/generate-google3-docs/resource-reference/generated/resource-docs/accesscontextmanager/accesscontextmanageraccesslevel.md
@@ -57,20 +57,6 @@
 ## Custom Resource Definition Properties
 
 
-### Annotations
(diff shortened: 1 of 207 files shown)


gemmahou:
/lgtm
/approve
/hold

I remember we have the `state-into-spec` annotation in some golden logs([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/5b8285bf5c91d4e59bca8fe8524ef57ea40f98b9/pkg/test/resourcefixture/testdata/basic/gkehub/v1beta1/gkehubmembership/_generated_object_gkehubmembership.golden.yaml#L8)). Do we want to update them to avoid diffs in golden tests?

maqiuyujoyce:
`state-into-spec` annotation is still functioning as before, so the golden logs should have this annotation. This PR is to remove it from the docs and stop advertising this feature.

yuwenma:
/hold cancel

Lovely! Thank you.

PR #2323, opened 2024-07-19
[WIP] Return a warning when the state-into-spec annotation will be defaulted to 'merge'

maqiuyujoyce:
### Change description

This is an experiment PR that checks the default value of `state-into-spec` annotation in the webhook via retrieving the ConfigConnector object and potentially the ConfigConnectorContext object, and returns a warning if the default value is `merge`.

In general, we shouldn't make API calls in the webhook because it has a short timeout (10s), though it should be fairly cheap to do the CC/CCC object retrievals.

### Tests you have done

Ran `make deploy-controller` locally with basic CCC and CC objects (no `spec.stateIntoSpec` configuration), then applied a PubsubTopic resource (no `state-into-spec` annotation). Got the following warning as expected:

```
kubectl apply -f pubsubtopic.yaml 
Warning: 'cnrm.cloud.google.com/state-into-spec: merge' is unsupported for CRDs added in 1.114.0 and later. Use 'absent' instead. More details can be found at https://cloud.google.com/config-connector/docs/concepts/ignore-unspecified-fields.
pubsubtopic.pubsub.cnrm.cloud.google.com/pubsubtopic-sample-2 created
```

### DIFF
--- a/pkg/webhook/state_into_spec_annotation_validator.go
+++ b/pkg/webhook/state_into_spec_annotation_validator.go
@@ -51,13 +51,28 @@ func (a *stateIntoSpecAnnotationValidator) Handle(ctx context.Context, req admis
 	}
 
 	value, ok := k8s.GetAnnotation(k8s.StateIntoSpecAnnotation, obj)
+	klog.Infof("maqiuyu...annotation value %v, ok? %v", value, ok)
 	if ok && value == k8s.StateMergeIntoSpec {
 		return allowedResponse.WithWarnings(
 			fmt.Sprintf("'%v: %v' is unsupported for CRDs added in "+
 				"1.114.0 and later. Use '%v' instead. More details can be "+
 				"found at https://cloud.google.com/config-connector/docs/concepts/ignore-unspecified-fields.",
 				k8s.StateIntoSpecAnnotation, k8s.StateMergeIntoSpec, k8s.StateAbsentInSpec))
+	} else if !ok {
+		stateIntoSpecDefaultVal, err := k8s.GetStateIntoSpecDefaultValue(ctx, a.client, obj)
+		klog.Infof("maqiuyu...default value %v, err? %v", stateIntoSpecDefaultVal, err)
+		if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/webhook/state_into_spec_annotation_validator.go
```diff
@@ -51,13 +51,28 @@ func (a *stateIntoSpecAnnotationValidator) Handle(ctx context.Context, req admis
 	}
 
 	value, ok := k8s.GetAnnotation(k8s.StateIntoSpecAnnotation, obj)
+	klog.Infof("maqiuyu...annotation value %v, ok? %v", value, ok)
```
gemmahou, line 54: nit: remove the debug message?

maqiuyujoyce: Thanks, @gemmahou ! Will clean up if we decide we'll move forward with this change.

maqiuyujoyce:
I think it's nice to have the warning in the webhook, though making API calls in the webhook seems to be violating the best practices and may have consequence about the performance and etc. IMO the risk is higher than the benefit, and it should be okay if we don't merge this PR (e.g. users don't receive warnings if their CRs are defaulted with `state-into-spec: merge`). @yuwenma WDYT?

pkg/webhook/state_into_spec_annotation_validator.go
```diff
@@ -51,13 +51,29 @@ func (a *stateIntoSpecAnnotationValidator) Handle(ctx context.Context, req admis
 	}
 
 	value, ok := k8s.GetAnnotation(k8s.StateIntoSpecAnnotation, obj)
+	klog.Infof("maqiuyu...kind %q, obj %q, annotation value %v, ok? %v", obj.GroupVersionKind().Kind, obj.GetName(), value, ok)
```
yuwenma, line 54: nit: clean up?

maqiuyujoyce:
Discussed offline and decided not to add this feature.

PR #2324, opened 2024-07-19
Add basic mockgcp test cases for CloudSQL

jasonvigil:
These test cases specify the "required" fields for postgres, mysql, and sqlserver and include updates for the non-immutable fields.


pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqluser/_generated_object_sqluser.golden.yaml
```diff
@@ -26,7 +26,6 @@ spec:
         key: password
         name: sqluser-dep-${uniqueId}
   resourceID: sqluser-${uniqueId}
-  type: BUILT_IN
```
yuwenma, line 29: I guess(?) this is a default optional field. Can we keep the server's behavior here?

mockgcp/mocksql/service.go
```diff
@@ -62,16 +62,25 @@ func (s *MockService) Register(grpcServer *grpc.Server) {
...
+	if err != nil {
 		return nil, err
 	}
-	if err := pb.RegisterSqlOperationsServiceHandler(ctx, mux, conn); err != nil {
-		return nil, err
+
+	mux.RewriteError = func(ctx context.Context, error *httpmux.ErrorResponse) {
+		// todo: This will affect all services, not just instances. Need to figure out some way
+		//       to limit the scope of custom errors to be per-service (?)
+		if error.Code == 404 {
```
yuwenma, line 76: You can modify the implementation of `httpmux`.

mockgcp/mocksql/sqluser.go
```diff
@@ -93,13 +93,6 @@ func (s *sqlUsersService) Insert(ctx context.Context, req *pb.SqlUsersInsertRequ
 	obj.Instance = name.Instance
 	obj.Kind = "sql#user"
 
-	if obj.PasswordPolicy == nil {
-		obj.PasswordPolicy = &pb.UserPasswordValidationPolicy{}
-	}
-	if obj.PasswordPolicy.Status == nil {
-		obj.PasswordPolicy.Status = &pb.PasswordStatus{}
-	}
-
```
yuwenma, line 102: mind explaining why removing this ?

pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/mysqlinstancebasic/_http.log
```diff
@@ -0,0 +1,765 @@
...
+      "project": "${projectId}"
+    }
+  ],
+  "kind": "sql#usersList"
+}
+
+---
+
+DELETE https://sqladmin.googleapis.com/sql/v1beta4/projects/${projectId}/instances/mysqlbasic-${uniqueId}/users?alt=json&host=%25&name=root&prettyPrint=false
+User-Agent: google-api-go-client/0.5 Terraform/ (+https://www.terraform.io) Terraform-Plugin-SDK/2.10.1 terraform-provider-google-beta/kcc/controller-manager
```
yuwenma, line 114: no blocker. I'm curious about this TF delete call. Do you happen to know why?

pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/mysqlinstancebasic/create.yaml
```diff
@@ -0,0 +1,31 @@
...
+  annotations:
+    cnrm.cloud.google.com/state-into-spec: absent
+spec:
+  databaseVersion: MYSQL_5_7
+  region: us-central1
+  settings:
+    # Location preference is not actually a required field. However, setting it for tests
+    # helps with with normalizing the GCP responses, because otherwise GCP chooses a zone
+    # preference based on availability. Therefore it could potentially vary if not
+    # explicity specified.
```
yuwenma, line 28: 👍

tests/e2e/unified_test.go
```diff
@@ -622,8 +622,100 @@ func runScenario(ctx context.Context, t *testing.T, testPause bool, fixture reso
...
+						if _, found, _ := unstructured.NestedString(obj, "name"); found {
+							// Set a hardcoded UUID.
+							if err := unstructured.SetNestedField(obj, "11111111-aaaa-2222-bbbb-123456789abc", "name"); err != nil {
+								t.Fatal(err)
+							}
+						}
+						if _, found, _ := unstructured.NestedString(obj, "selfLink"); found {
+							// Set a hardcoded self link (based on name).
+							if err := unstructured.SetNestedField(obj, "https://sqladmin.googleapis.com/sql/v1beta4/projects/${projectId}/operations/11111111-aaaa-2222-bbbb-123456789abc", "selfLink"); err != nil {
+								t.Fatal(err)
```
yuwenma, line 715: I think we would want a more generic normalize function for self link: it can replace the ID to something like ${uniqueId} for all resources that has a selflink in the response.  Check `extractIDsFromLinks`

yuwenma:
Looks good! Have some minor comments.

pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqluser/_generated_object_sqluser.golden.yaml
jasonvigil: It looks like GCP does not return this field by default. This was probably originally added due to `state-into-spec: merge`, combined with the mock http mux marshaler previously being configured to return "0" / nil / empty values (this is the "0" value for the enum). It should be removed, since that's the GCP behavior.

mockgcp/mocksql/sqluser.go
jasonvigil: For postgres + sqlserver, these fields are not sent in the GCP responses, so it felt "more correct" to remove it.

tests/e2e/unified_test.go
jasonvigil: Ah, good point. I think we handle this already. It may have been pointless to add this normalizer. Let me double-check.

pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/mysqlinstancebasic/_http.log
jasonvigil: I was kinda curious about this also. It seems to be a security-related feature of terraform. From the [docs](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance):

> NOTE on google_sql_database_instance:
> - Second-generation instances include a default 'root'@'%' user with no password. This user will be deleted by Terraform on instance creation. You should use google_sql_user to define a custom user with a restricted host and strong password.

pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqluser/_generated_object_sqluser.golden.yaml
yuwenma: Does this git diff come from the real GCP and MockGCP (that the previous object and http run against real GCP and this PR switched to mockgcp )? If so, the `state-into-spec` will not make a diff, we can tell that from the  _http.log response [here](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2324/files#diff-187f6b6f9b12d8c78beee54344ae65366049302eb4f6b2bd984448cb04bc3f90L91). I'm okay with some fields not handled in one PR, especially since this PR already contains a lot of good stuff and is really big now. Just want to make sure I'm on the same page about where the changes come from.

mockgcp/mocksql/sqluser.go
yuwenma: Good to know! Thank you for the context!

pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/mysqlinstancebasic/_http.log
yuwenma: Obviously this cloud sql is just so important that every declarative config tool is trying to handle it with a lot of consideration!

yuwenma (commented):
/lgtm
/approve

/hold one comment about the default BUILTIN. No blockers, feel free to cancel the hold and improve that in follow up PRs

pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/mysqlinstancebasic/_http.log
yuwenma: Obviously this cloud sql is just so important that every declarative config tool is trying to handle it with a lot of consideration!

pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqluser/_generated_object_sqluser.golden.yaml
jasonvigil: The reason for this diff is because the sqluser _http.log and golden.yaml were based on mockgcp. Real GCP does not return `type: BUILT_IN`. The only reason mockgcp was returning BUILT_IN was because it's a 0 value, and before this change, the mockgcp sql service (unlike real gcp) was configured to return all 0 values. Here's where the code defines BUILT_IN as the 0 value for that field: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/e29dd609e93ff648b0f6e06a45317d6b2e1b0cf2/mockgcp/generated/mockgcp/cloud/sql/v1beta4/cloud_sql_users.pb.go#L45. So, in this case, I think it's correct to remove this, so we can match the real GCP behavior.

tests/e2e/unified_test.go
jasonvigil: Ok, fixed. Added another check for sql operation ids. Thanks :)

PR #2334, opened 2024-07-22
chore: don't add cnrm prefix to the operator image path

justinsb:
When building images with dev/tasks/build-images, currently all the
images get the same base path, so the extra cnrm "directory" is not
what we do today.


yuwenma (approved):
/lgtm
/approve

PR #2337, opened 2024-07-22
Resolve resource reference(initial PR)

gemmahou:
### Change description

Rebased on #2352 

1. Move `resolve<resource>` functions into `apis/refs/` package.
2. Make `GetResourceID`, `GetLocation` helper functions.

TODO:
Draft a tool to auto-generate the resolve<resourcerefs> logic.

Note:
I only include [ComputeNetworkRef](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/compute/v1beta1/computeforwardingrule_types.go#L248) and [ComputeTargetHTTPProxyRef](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/compute/v1beta1/computeforwardingrule_types.go#L96) in this PR. More references will be required for ComputeForwardingRule.

### Tests you have done


apis/refs/v1beta1/resourceref.go
yuwenma, line 1: I think we want to get rid of a generic `resourceRef` but using more dedicated kind.   Any reasons to switch it back?

apis/refs/v1beta1/resourceref.go
```diff
@@ -0,0 +1,126 @@
...
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/clients/generated/apis/k8s/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func ComputeNetworkRef_ConvertToExternal(ctx context.Context, reader client.Reader, src client.Object, ref *v1alpha1.ResourceRef) (*v1alpha1.ResourceRef, error) {
```
yuwenma, line 26: I like this name! And I also think we can actually having our tools to auto-generate this functions when adding new resources from proto!

apis/refs/v1beta1/resourceref.go
yuwenma: by reading the code, I think this is just a naming thing. Maybe move the functions to the computenetoworkref.go file

apis/refs/v1beta1/util.go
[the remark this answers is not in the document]
gemmahou: I moved `GetResourceID` `GetLocation` and `GetProjectID`(not resolve project) to `adapterutils.go`(or maybe a better name), those methods can be reused.

apis/refs/v1beta1/resourceref.go
gemmahou: Removed generic `resourceRef`

pkg/controller/direct/cloudbuild/workerpool_controller.go
```diff
@@ -90,7 +91,7 @@ func (m *model) AdapterForObject(ctx context.Context, reader client.Reader, u *u
 	}
 
 	// Get GCP Project
-	projectRef, err := refs.ResolveProject(ctx, reader, obj, obj.Spec.ProjectRef)
+	projectRef, err := resolverefs.ResolveProjectRef(ctx, reader, obj, obj.Spec.ProjectRef)
```
justinsb, line 94: So I moved the resolvers into the API package recently.  The intent was that we could then have the resolvers implement an interface, and make a lot of this generic.

WDYT - can we put the resolver functions into the refs package?   (We could even just add a `ResolveRef` function to each ref in this PR?)

pkg/controller/direct/resolverefs/helper.go
```diff
@@ -0,0 +1,88 @@
...
+	if err != nil {
+		return "", fmt.Errorf("cannot get location for referenced %s %v: %w", obj.GetKind(), obj.GetNamespace(), err)
+	}
+	if location == "" {
+		return "", fmt.Errorf("cannot get location for referenced %s %v (spec.location not set)", obj.GetKind(), obj.GetNamespace())
+	}
+	return location, nil
+}
+
+func GetProjectID(ctx context.Context, reader client.Reader, obj *unstructured.Unstructured) (string, error) {
```
justinsb, line 50: So this was called something like "ResolveProjectID" as a hint that we might have to go chase down some references etc.  I don't mind renaming it if it's confusing, but maybe add a comment to the function?

pkg/controller/direct/resolverefs/computetargethttpproxyref.go
```diff
@@ -0,0 +1,114 @@
...
+	if ref.External != "" {
+		if ref.Name != "" {
+			return nil, fmt.Errorf("cannot specify both name and external on ComputeNetwork reference")
+		}
+
+		tokens := strings.Split(ref.External, "/")
+		if len(tokens) == 5 && tokens[0] == "projects" && tokens[2] == "global" && tokens[3] == "targetHttpProxies" {
+			return &ComputeTargetHTTPProxy{
+				Project:                  tokens[1],
+				ComputeTargetHTTPProxyID: tokens[4]}, nil
```
justinsb, line 58: Looks like we should set Location == global here, so it round trips?

pkg/controller/direct/resolverefs/projectref.go
```diff
@@ -0,0 +1,109 @@
...
+		External:  in.External,
+		Kind:      in.Kind,
+	}
+}
+
+type Project struct {
+	ProjectID string
+}
+
+// ResolveProjectRef will resolve a ProjectRef to a Project, with the ProjectID.
```
justinsb, line 49: TBD if this should actually just return a ProjectRef, to follow the pattern we seem to have built since.  But no need to address in this PR  (unless you want to)

pkg/controller/direct/cloudbuild/workerpool_controller.go
gemmahou: Regarding the placement of resolver functions, I don't have a strong preference. It could be `api/refs/networkref.go` or `direct/refs/networkref.go`. I think we should maintain consistency in the future. I'm open to moving the resolver functions to `api/refs` if that's preferable.

My main concern here is avoiding placing these functions in individual controller packages, for example `direct/compute/networkref.go`. Some references are used in multiple resources, like `ComputeNetWork`, which is referenced by `CloudBuildWorkerPool` and `ComputeForwardingRule` (and potentially others).

pkg/controller/direct/resolverefs/computetargethttpproxyref.go
gemmahou: Thanks, fixed!

pkg/controller/direct/resolverefs/helper.go
gemmahou: I reverted it back to ResolveProjectID, will address in another PR if needed.

pkg/controller/direct/resolverefs/projectref.go
gemmahou: I reverted it back, will address in another PR if needed.

gemmahou:
Chatted with @yuwenma offline and reverted the function name change. 

I will have a followup PR to resolve other refs(i.e. compute address, subnetwork etc), and a tool to auto-generate the resolver functions.

yuwenma (commented):
/lgtm

apis/refs/v1beta1/helper.go
```diff
@@ -30,3 +29,16 @@ func GetResourceID(u *unstructured.Unstructured) (string, error) {
 	}
 	return resourceID, nil
 }
+
+// TODO(yuhou): Location can be optional. Use provider default location when it's unset.
+func GetLocation(obj *unstructured.Unstructured) (string, error) {
```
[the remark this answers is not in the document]
gemmahou: Done! I think we'll eventually make `GetLocation` a helper function because it will be used in resolving more references. I agree that we need to do so when it's fully functional.

justinsb (approved):

justinsb:
Ooops ... didn't realize @yuwenma was looking at this also.

/hold

(for yuwen's reviews)

yuwenma:
/hold cancel
Looks good! Thank you for the changes!

PR #2339, opened 2024-07-23
Ensure files in PR snapshot matches 'make ready-pr' result

maqiuyujoyce:
### Change description

Updated the validation script so that it not only checks the modified files but also the added files. And reran `make ready-pr`.

### Tests you have done


maqiuyujoyce:
Hi @justinsb , I think we need to make sure `make manifests` doesn't generate any diff at master HEAD, but RedisCulster CRD got generated when I ran the command in a clean branch. Do you want it to be exempted or added?

PR #2341, opened 2024-07-23
Remove aws and azure support for BigQueryConnectionConnection

maqiuyujoyce:
### Change description

A couple of reasons why we should remove the aws and azure support in BigQueryConnectionConnection:

1. There are a few output-only fields under the top-level `spec.aws` and `spec.azure` fields. We need to handle them specially.
2. There is no test coverage for the aws or the azure support, but testing it out means we need to set up accounts in both AWS and Azure. This may be expensive.

Considering that there is no urgent request about those features, and BigQueryConnectionConnection is still a v1alpha1 kind, let's remove the aws and azure fields for now and only add them after we can support thorough testing.

### Tests you have done

Note that an unrelated CRD actually got generated when running `make ready-pr`: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2339


yuwenma (commented):
/lgtm
/approve

Thanks for the change. Nice to see some VRD exceptions are cleared out.

PR #2347, opened 2024-07-23
Update Attached samples to use CRD instead of Facade.

hankfreund:
### Change description

- Grabbed the CRD from the cluster after applying the Facade and replaced the contents in the yaml.
- Renamed the group something more generic to better demonstrate the definition is something users own.
- Tested both AKS and EKS flows.

### DIFF
--- a/experiments/compositions/samples/AttachedAKS/03-attached-1.yaml
+++ b/experiments/compositions/samples/AttachedAKS/03-attached-1.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-apiVersion: facade.compositions.google.com/v1
+apiVersion: idp.mycompany.com/v1
 kind: AttachedAKS
 metadata:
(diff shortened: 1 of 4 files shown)


experiments/compositions/samples/AttachedAKS/03-attached-1.yaml
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-apiVersion: facade.compositions.google.com/v1
+apiVersion: compositions.mycompany.com/v1
```
barney-s, line 15: we are using idp instead of compositions for the group.
Not a blocker, but just wanted to let you know.

barney-s: `idp.mycompany.com`

hankfreund: Didn't realize. I've updated to match.

PR #2355, opened 2024-07-24
mockgcp: firestore

acpana:
Showcase import failures in mockgco. Should be fixed with the go.work addition.


acpana:
No longer up to date

PR #2357, opened 2024-07-25
chore: use canonical message name for "Empty"

jingyih:


jingyih:
Hi @justinsb,

Is there a specific reason that we annotate Empty struct with "emptypb.Empty"? My understanding is that the types and mapping of this field is currently manually handled?

justinsb:
Good call, just a brain fail on my part :-)

/approve
/lgtm

PR #2364, opened 2024-07-26
feat: Direct controller for ComputeForwardingRule

gemmahou:
### Change description

- [x] #2337 
- [x] #2390  
- [x] #2433 
- [x] #2545 

### Tests you have done


pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/regionalcomputeforwardingrule/_http.log
```diff
@@ -449,22 +449,20 @@ X-Xss-Protection: 0
...
   "description": "A regional forwarding rule",
   "labels": {
     "cnrm-test": "true",
-    "label-one": "value-one",
-    "managed-by-cnrm": "true"
+    "label-one": "value-one"
   },
   "loadBalancingScheme": "EXTERNAL",
   "name": "computeregionalforwardingrule-${uniqueId}",
-  "region": "projects/${projectId}/global/regions/us-central1",
```
gemmahou, line 467: This region looks problematic. I will investigate where the value comes from.

apis/compute/v1beta1/computeforwardingrule_types.go
```diff
@@ -31,7 +31,7 @@ var (
 	// AddToScheme is a global function that registers this API group & version to a scheme
 	AddToScheme = SchemeBuilder.AddToScheme
 
-	ComputeForwardingRuleGVK = schema.GroupVersionKind{
+	GroupVersionKind = schema.GroupVersionKind{
```
justinsb, line 34: This is going to cause conflicts - we have multiple GVKs in this package I believe.

apis/refs/v1beta1/computerefs.go
```diff
@@ -39,7 +39,7 @@ type ComputeNetwork struct {
 	ComputeNetworkID string
 }
 
-func (c *ComputeNetwork) String() string {
+func (c *ComputeNetwork) Id() string {
```
justinsb, line 42: Nit: In go, we normally do `ID`

apis/refs/v1beta1/computerefs.go
```diff
@@ -74,31 +74,31 @@ func ResolveComputeNetwork(ctx context.Context, reader client.Reader, src client
 		key.Namespace = src.GetNamespace()
 	}
 
-	computenetwork := &unstructured.Unstructured{}
-	computenetwork.SetGroupVersionKind(schema.GroupVersionKind{
+	computeNetwork := &unstructured.Unstructured{}
+	computeNetwork.SetGroupVersionKind(schema.GroupVersionKind{
```
justinsb, line 78: Nit: we might want to use the GVK from the API if we already have a dependency.  Not a big deal though.

apis/refs/v1beta1/computerefs.go
```diff
@@ -120,6 +120,102 @@ type ComputeAddressRef struct {
...
 }
 
+type ComputeAddress struct {
+	Project          string
+	Location         string
+	ComputeAddressID string
+	Address          string
+}
+
+func (c *ComputeAddress) Url() string {
```
justinsb, line 130: Nit: In go, acronyms (like ID) are capitalized, so this should be `URL()`

apis/refs/v1beta1/computerefs.go
```diff
@@ -120,6 +120,102 @@ type ComputeAddressRef struct {
...
+				Project:          tokens[1],
+				Location:         "global",
+				ComputeAddressID: tokens[4]}, nil
+		} else if len(tokens) == 6 && tokens[0] == "projects" && tokens[2] == "location" && tokens[4] == "addresses" {
+			return &ComputeAddress{
+				Project:          tokens[1],
+				Location:         tokens[3],
+				ComputeAddressID: tokens[5]}, nil
+		}
+		return nil, fmt.Errorf("format of ComputeAddress external=%q was not known (use projects/<projectId>/global/addresses/<proxyId> or projects/<projectId>/regions/<region>/addresses/<addressId>)", ref.External)
```
justinsb, line 163: Nit: s/proxyId/addressId

apis/refs/v1beta1/computerefs.go
```diff
@@ -120,6 +120,102 @@ type ComputeAddressRef struct {
...
+	key := types.NamespacedName{
+		Namespace: ref.Namespace,
+		Name:      ref.Name,
+	}
+	if key.Namespace == "" {
+		key.Namespace = src.GetNamespace()
+	}
+
+	computeAddress := &unstructured.Unstructured{}
+	computeAddress.SetGroupVersionKind(schema.GroupVersionKind{
```
justinsb, line 179: Nit: would be nice to use the API GVK

apis/refs/v1beta1/computerefs.go
```diff
@@ -269,23 +365,120 @@ type ComputeTargetTCPProxyRef struct {
...
-		return "", fmt.Errorf("cannot get location for referenced %s %v (spec.location not set)", obj.GetKind(), obj.GetNamespace())
+		return "", fmt.Errorf("cannot get location or region for referenced %s %v (spec.location or spec.region is empty)", obj.GetKind(), obj.GetNamespace())
 	}
 	return location, nil
 }
+
+func getAddress(obj *unstructured.Unstructured) (string, error) {
+	address, _, err := unstructured.NestedString(obj.Object, "spec", "address")
+	if err != nil {
+		return "", fmt.Errorf("reading spec.resourceID from %v %v/%v: %w", obj.GroupVersionKind().Kind, obj.GetNamespace(), obj.GetName(), err)
```
justinsb, line 478: Nit: spec.address?

apis/refs/v1beta1/computerefs.go
```diff
@@ -269,23 +365,120 @@ type ComputeTargetTCPProxyRef struct {
...
+	}
 	if location == "" {
-		return "", fmt.Errorf("cannot get location for referenced %s %v (spec.location not set)", obj.GetKind(), obj.GetNamespace())
+		return "", fmt.Errorf("cannot get location or region for referenced %s %v (spec.location or spec.region is empty)", obj.GetKind(), obj.GetNamespace())
 	}
 	return location, nil
 }
+
+func getAddress(obj *unstructured.Unstructured) (string, error) {
+	address, _, err := unstructured.NestedString(obj.Object, "spec", "address")
```
justinsb, line 476: Nit: can we add a TODO that we should be source this from status or status.observedState?

apis/refs/v1beta1/computerefs.go
```diff
@@ -269,23 +365,120 @@ type ComputeTargetTCPProxyRef struct {
...
+		}
+
+		tokens := strings.Split(ref.External, "/")
+		if len(tokens) == 6 && tokens[0] == "projects" && tokens[2] == "location" && tokens[4] == "targetVpnGateways" {
+			return &ComputeTargetVPNGateway{
+				Project:                   tokens[1],
+				Location:                  tokens[3],
+				ComputeTargetVPNGatewayID: tokens[5]}, nil
+		}
+		return nil, fmt.Errorf("format of ComputeTargetVPNGateway external=%q was not known (use projects/<projectId>/regions/<region>/targetVpnGateways/<gatewayId>)", ref.External)
```
justinsb, line 403: Nit: you use "location" in the if statement, but "regions" here.

apis/refs/v1beta1/computerefs.go
```diff
@@ -269,23 +365,120 @@ type ComputeTargetTCPProxyRef struct {
...
+
+	computeTargetVPNGateway := &unstructured.Unstructured{}
+	computeTargetVPNGateway.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "compute.cnrm.cloud.google.com",
+		Version: "v1beta1",
+		Kind:    "ComputeTargetVPNGateway",
+	})
+	if err := reader.Get(ctx, key, computeTargetVPNGateway); err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, fmt.Errorf("referenced ComputeAddress %v not found", key)
```
justinsb, line 426: Nit: ComputeTargetVPNGateway (and in the next message)

apis/refs/v1beta1/computerefs.go
```diff
@@ -269,23 +365,120 @@ type ComputeTargetTCPProxyRef struct {
...
 		return "", fmt.Errorf("cannot get location for referenced %s %v: %w", obj.GetKind(), obj.GetNamespace(), err)
 	}
+	if !found {
+		// if region is set, use its value as location
+		location, found, err = unstructured.NestedString(obj.Object, "spec", "region")
+		if err != nil {
+			return "", fmt.Errorf("cannot get region for referenced %s %v: %w", obj.GetKind(), obj.GetNamespace(), err)
+		}
+		if !found {
+			return "", fmt.Errorf("cannot get location or region for referenced %s %v (spec.location or spec.region not set)", obj.GetKind(), obj.GetNamespace())
```
justinsb, line 466: This is going to be a confusing error message if the type only supports one of location or region...

pkg/controller/direct/cloudbuild/workerpool_controller.go
```diff
@@ -109,7 +109,7 @@ func (m *model) AdapterForObject(ctx context.Context, reader client.Reader, u *u
 			return nil, err
 
 		}
-		obj.Spec.PrivatePoolConfig.NetworkConfig.PeeredNetworkRef.External = networkRef.String()
+		obj.Spec.PrivatePoolConfig.NetworkConfig.PeeredNetworkRef.External = networkRef.Id()
```
justinsb, line 112: Maybe we should name this form `ExternalRef()`?

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrule/_generated_object_globalcomputeforwardingrule.golden.yaml
```diff
@@ -48,6 +32,5 @@ status:
     reason: UpToDate
     status: "True"
     type: Ready
-  labelFingerprint: abcdef0123A=
```
yuwenma, line 51: This means the adapter's Create and Update methods shall update the ComputeForwardingRule `status` using the response value from the compute service.

apis/refs/v1beta1/computerefs.go
```diff
@@ -120,6 +120,102 @@ type ComputeAddressRef struct {
...
+		if len(tokens) == 5 && tokens[0] == "projects" && tokens[2] == "global" && tokens[3] == "addresses" {
+			return &ComputeAddress{
+				Project:          tokens[1],
+				Location:         "global",
+				ComputeAddressID: tokens[4]}, nil
+		} else if len(tokens) == 6 && tokens[0] == "projects" && tokens[2] == "location" && tokens[4] == "addresses" {
+			return &ComputeAddress{
+				Project:          tokens[1],
+				Location:         tokens[3],
+				ComputeAddressID: tokens[5]}, nil
```
justinsb, line 161: Address does not appear to be set in this code path

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/regionalcomputeforwardingrule/_generated_object_regionalcomputeforwardingrule.golden.yaml
```diff
@@ -1,28 +1,13 @@
...
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
 apiVersion: compute.cnrm.cloud.google.com/v1beta1
 kind: ComputeForwardingRule
 metadata:
   annotations:
     cnrm.cloud.google.com/management-conflict-prevention-policy: none
     cnrm.cloud.google.com/project-id: ${projectId}
-    cnrm.cloud.google.com/state-into-spec: merge
```
yuwenma, line 21: I think not showing this label is expected (and great). Is the `loadBalancingScheme` a output-only field?

pkg/controller/direct/compute/forwardingrule_controller.go
```diff
@@ -0,0 +1,460 @@
...
+
+	desired := a.desired.DeepCopy()
+
+	forwardingRule := ComputeForwardingRuleSpec_ToProto(mapCtx, &desired.Spec)
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+	forwardingRule.Name = direct.LazyPtr(a.resourceID)
+	forwardingRule.Labels = desired.Labels
+	// TODO(yuhou): handle default spec values
```
justinsb, line 247: Who defaults the LoadBalancingScheme value today?  Terraform?

pkg/controller/direct/compute/forwardingrule_controller.go
```diff
@@ -0,0 +1,460 @@
...
+			ForwardingRuleResource: forwardingRule,
+			Project:                a.projectID,
+		}
+		getReq := &computepb.GetGlobalForwardingRuleRequest{
+			ForwardingRule: a.resourceID,
+			Project:        a.projectID,
+		}
+		op, err = a.globalForwardingRulesClient.Insert(ctx, req)
+		if err != nil {
+			return fmt.Errorf("creating ComputeForwardingRule %s failed: %w", a.resourceID, err)
```
justinsb, line 266: In other controllers, we've used `fullyQualifiedName()` instead of just the name, to give a more complete error message

pkg/controller/direct/compute/forwardingrule_controller.go
```diff
@@ -0,0 +1,460 @@
...
+			}
+			return fmt.Errorf("getting ComputeForwardingRule %q failed: %w", a.resourceID, err)
+		}
+	} else {
+		req := &computepb.InsertForwardingRuleRequest{
+			ForwardingRuleResource: forwardingRule,
+			Region:                 a.Location,
+			Project:                a.projectID,
+		}
+		getReq := &computepb.GetForwardingRuleRequest{
```
justinsb, line 283: Nit: move this down to where you are using it, it's confusing otherwise.

pkg/controller/direct/compute/forwardingrule_controller.go
```diff
@@ -0,0 +1,460 @@
...
+			ForwardingRule: a.resourceID,
+			Region:         a.Location,
```
justinsb, line 285: Nit: Why is location exported?

pkg/controller/direct/compute/forwardingrule_controller.go
```diff
@@ -0,0 +1,460 @@
...
+	var err error
+	// TODO(yuhou): Checked the realGCP logs, setLabels request is being sent even when there are no updates to labels.
+	// That might because of the generated labelsFingerPrint?
+	if a.Location == "global" {
+		setLabelsReq := &computepb.SetLabelsGlobalForwardingRuleRequest{
+			Resource:                       a.resourceID,
+			GlobalSetLabelsRequestResource: &computepb.GlobalSetLabelsRequest{LabelFingerprint: a.actual.LabelFingerprint, Labels: forwardingRule.Labels},
+			Project:                        a.projectID,
+		}
+		_, err = a.globalForwardingRulesClient.SetLabels(ctx, setLabelsReq)
```
justinsb, line 338: What does SetLabels return?  The worry is always that it returns an operation and that we need to wait for that operation to complete

pkg/controller/direct/compute/forwardingrule_controller.go
```diff
@@ -0,0 +1,460 @@
...
+		return mapCtx.Err()
+	}
+	forwardingRule.Name = direct.LazyPtr(a.resourceID)
+	forwardingRule.Labels = desired.Labels
+
+	// Patch only support update on networkTier field, which KCC does not support yet.
+	// Use setTarget and setLabels to update target and labels fields.
+	var err error
+	// TODO(yuhou): Checked the realGCP logs, setLabels request is being sent even when there are no updates to labels.
+	// That might because of the generated labelsFingerPrint?
```
justinsb, line 331: Possible, but I doubt it.  It's probably just the existing controller is suboptimal.  I would personally do something like this:

`bool alwaysSendLabelChanges = true // For compatability with existing controller`

Then you can send a follow-on PR to flip that to false, the golden log change should be easy to understand.

pkg/controller/direct/compute/forwardingrule_controller.go
```diff
@@ -0,0 +1,460 @@
...
+
+	// setTarget request is sent when there are updates to target.
+	if !reflect.DeepEqual(forwardingRule.Target, a.actual.Target) {
+		if a.Location == "global" {
+			setTargetReq := &computepb.SetTargetGlobalForwardingRuleRequest{
+				ForwardingRule:          a.resourceID,
+				TargetReferenceResource: &computepb.TargetReference{Target: forwardingRule.Target},
+				Project:                 a.projectID,
+			}
+			_, err = a.globalForwardingRulesClient.SetTarget(ctx, setTargetReq)
```
justinsb, line 361: As before, we need some way to be able to look at the code and know that this is not returning an operation.

pkg/controller/direct/compute/forwardingrule_controller.go
```diff
@@ -0,0 +1,460 @@
...
+	} else {
+		getReq := &computepb.GetForwardingRuleRequest{
+			ForwardingRule: a.resourceID,
+			Region:         a.Location,
+			Project:        a.projectID,
+		}
+		updated, err = a.forwardingRulesClient.Get(ctx, getReq)
+	}
+	if err != nil {
+		if direct.IsNotFound(err) {
```
justinsb, line 393: Not sure you need the IsNotFound check here, this is pretty surprising.

pkg/controller/direct/compute/forwardingrule_controller.go
```diff
@@ -0,0 +1,460 @@
...
+			return err
+		}
+		return fmt.Errorf("getting ComputeForwardingRule %q failed: %w", a.resourceID, err)
+	}
+
+	status := &krm.ComputeForwardingRuleStatus{}
+	// TODO(yuhou): Do we need other status beside conditions and observedGeneration?
```
justinsb, line 400: To merge, we need whatever is in there today :-)

pkg/controller/direct/compute/forwardingrule_controller.go
```diff
@@ -0,0 +1,460 @@
...
+		// return (false, nil) if the object was not found but should be presumed deleted.
+		return false, nil
+	}
+
+	if a.Location == "global" {
+		req := &computepb.DeleteGlobalForwardingRuleRequest{
+			ForwardingRule: a.resourceID,
+			Project:        a.projectID,
+		}
+		_, err = a.globalForwardingRulesClient.Delete(ctx, req)
```
justinsb, line 433: Also check if operation here

pkg/controller/direct/compute/forwardingrule_controller.go
```diff
@@ -0,0 +1,460 @@
...
+	return fmt.Sprintf("projects/%s/global/forwardingRules/%s", a.projectID, a.resourceID)
+}
+
+func (a *forwardingRuleAdapter) fullyQualifiedName() string {
+	return fmt.Sprintf("projects/%s/regions/%s/forwardingRules/%s", a.projectID, a.Location, a.resourceID)
+}
+
+func (a *forwardingRuleAdapter) getParent() string {
+	// TODO(user): Write the GCP URI parent for your resource
+	return fmt.Sprintf("projects/%s", a.projectID)
```
justinsb, line 459: I think this should include global or the region.  It's also possible this isn't used though..

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrule/_generated_object_globalcomputeforwardingrule.golden.yaml
```diff
@@ -37,7 +22,6 @@ spec:
   networkRef:
     name: default
   portRange: "80"
-  resourceID: computeglobalforwardingrule-${uniqueId}
```
justinsb, line 40: So I think we do need to keep setting spec.resourceID, for now, as the resource is a beta resource.

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrule/_generated_object_globalcomputeforwardingrule.golden.yaml
justinsb, line 51: Looks like we need to surface this field in status, because it's a beta resource.

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/regionalcomputeforwardingrule/_generated_object_regionalcomputeforwardingrule.golden.yaml
```diff
@@ -34,9 +19,7 @@ spec:
     addressRef:
       name: computeraddress-${uniqueId}
   ipProtocol: ESP
-  loadBalancingScheme: EXTERNAL
   location: us-central1
-  resourceID: computeregionalforwardingrule-${uniqueId}
```
yuwenma, line 39: Regarding resoruceID, here's the change (internal) https://docs.google.com/document/d/1A6YSEygeVJuXefb9cH52RiqEqp15nI5-UHNqG1T08kg/edit?tab=t.0#heading=h.yro65jsl04g3. You can focus on adding the `status.externalRef` for now.

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrule/_http.log
```diff
@@ -645,9 +645,9 @@ X-Xss-Protection: 0
 
 ---
 
-GET https://compute.googleapis.com/compute/v1/projects/${projectId}/global/forwardingRules/computeglobalforwardingrule-${uniqueId}?alt=json
+GET https://compute.googleapis.com/compute/v1/projects/${projectId}/global/forwardingRules/computeglobalforwardingrule-${uniqueId}
 Content-Type: application/json
-User-Agent: Terraform/ (+https://www.terraform.io) Terraform-Plugin-SDK/2.10.1 terraform-provider-google-beta/kcc/controller-manager
```
justinsb, line 650: Nit: Do you happen to know if we are sending _any_ UserAgent?

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrule/_http.log
```diff
@@ -676,17 +676,16 @@ X-Xss-Protection: 0
...
+x-goog-request-params: project=${projectId}
 
 {
   "IPAddress": "0.0.0.0",
   "description": "A global forwarding rule",
   "labels": {
     "cnrm-test": "true",
-    "label-one": "value-one",
-    "managed-by-cnrm": "true"
```
justinsb, line 689: If we are going to support labels (which I guess we have to, because it's a beta resource, then I guess we have to do the managed-by-cnrm label)

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrule/_http.log
```diff
@@ -793,46 +790,9 @@ X-Xss-Protection: 0
 
 ---
 
-GET https://compute.googleapis.com/compute/v1/projects/${projectId}/global/forwardingRules/computeglobalforwardingrule-${uniqueId}?alt=json
```
justinsb, line 796: You _could_ do the same trick where you make an extra (pointless) call to get here, just to keep the log simple.  But we can also just look at it and understand what's happening ... up to you!

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/regionalcomputeforwardingrule/_generated_object_regionalcomputeforwardingrule.golden.yaml
```diff
@@ -47,7 +30,5 @@ status:
     reason: UpToDate
     status: "True"
     type: Ready
-  creationTimestamp: "1970-01-01T00:00:00Z"
```
justinsb, line 50: Looks like you need to surface creationTimestamp also.

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/regionalcomputeforwardingrule/_generated_object_regionalcomputeforwardingrule.golden.yaml
```diff
@@ -34,9 +19,7 @@ spec:
     addressRef:
       name: computeraddress-${uniqueId}
   ipProtocol: ESP
-  loadBalancingScheme: EXTERNAL
```
justinsb, line 37: I think this is OK, this is a result of state-into-spec: absent and is maybe a candidate for adding to observedState

pkg/controller/direct/cloudbuild/workerpool_controller.go
yuwenma, line 135: I think Id is a new concept that needs some more definitions. The advantage of using String() is that then the Network implements the go Stringer by default.

pkg/controller/direct/compute/forwardingrule_controller.go
gemmahou: `GetForwardingRuleRequest` is regional, Region is required. The usage of `region` and `location` can be confusing, some APIs use `region`, some use `location`. We need a consistent pattern for this. For now, KCC supports `location` only, if API asks for `region` then we use `location` as the value.

apis/refs/v1beta1/computerefs.go
gemmahou: For ComputeForwardingRule, the addressRef.External field accepts "address" of ComputeAddress([screenshot](https://screenshot.googleplex.com/C2sRUPo9BoTLTBY.png), KCC [doc](https://cloud.google.com/config-connector/docs/reference/resource-docs/compute/computeforwardingrule)). While other resources might accept the ID of ComputeAddress as we checked in the if-else statement. 

I need to add another if statement here, if ref.External is an ip address(i.e. 8.8.8.8):
```
// When External is an ip address
if len(tokens) == 1 {
  return &ComputeAddress{
  Address: tokens[0]}, nil
// When External is ComputeAddress ID
} else if len(tokens) == 5 && tokens[0] == "projects" && tokens[2] == "global" && tokens[3] == "addresses" {
...
```

pkg/controller/direct/cloudbuild/workerpool_controller.go
gemmahou: Do you mean the proposal Yuwen shared [here](https://docs.google.com/document/d/1A6YSEygeVJuXefb9cH52RiqEqp15nI5-UHNqG1T08kg/edit?tab=t.0#heading=h.yro65jsl04g3)? I'll keep that in mind, but I think that's a more generic change, we do not support status.ExternalRef for now. We can chat about that and address in a future PR.

gemmahou: I'd like to link my previous [comment](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2364/files#r1704759524) here.

My intention to have `networkRef.ID` is to distinguish it from if there's another resource reference ComputeNetworks's full URL(i.e. ComputeSubnetWork's `networkRef` field requires full network url instead of the id). 

In Sci-fi, we should also support reference different fields of one resource.

pkg/controller/direct/compute/forwardingrule_controller.go
gemmahou: It can be provided by user, if not, then TF default it to "EXTERNAL"([TF code](https://source.corp.google.com/piper///depot/google3/third_party/golang/terraform_providers/google_beta/google_beta/services/compute/resource_compute_forwarding_rule.go;l=232?q=resource_compute_forwarding_rule.go&sq=package:piper%20file:%2F%2Fdepot%2Fgoogle3%20-file:google3%2Fexperimental))

pkg/controller/direct/compute/forwardingrule_controller.go
gemmahou: Removed. If update succeeds, the resource must exist.

pkg/controller/direct/compute/forwardingrule_controller.go
gemmahou: it returns an operation. I add a check for op.Wait()

pkg/controller/direct/compute/forwardingrule_controller.go
gemmahou: It returns an operation. I added a check for op.Wait()

pkg/controller/direct/compute/forwardingrule_controller.go
gemmahou: Done

pkg/controller/direct/compute/forwardingrule_controller.go
gemmahou: I'm removing this method as it isn't used anywhere

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/regionalcomputeforwardingrule/_generated_object_regionalcomputeforwardingrule.golden.yaml
gemmahou: It can be provided by user, if not, then TF defaults it to "EXTERNAL"([TF code](https://source.corp.google.com/piper///depot/google3/third_party/golang/terraform_providers/google_beta/google_beta/services/compute/resource_compute_forwarding_rule.go;l=232?q=resource_compute_forwarding_rule.go&sq=package:piper%20file:%2F%2Fdepot%2Fgoogle3%20-file:google3%2Fexperimental))

pkg/controller/direct/compute/forwardingrule_controller.go
```diff
@@ -0,0 +1,482 @@
...
+			Project:                a.projectID,
+		}
+		op, err = a.forwardingRulesClient.Insert(ctx, req)
+	}
+	if err != nil {
+		return fmt.Errorf("creating ComputeForwardingRule %s failed: %w", a.fullyQualifiedName(), err)
+	}
+	log.V(2).Info("successfully created ComputeForwardingRule", "name", a.fullyQualifiedName())
+	// Get the created resource
+	created := &computepb.ForwardingRule{}
```
yuwenma, line 279: The `Insert` returns a `computepb.Operation`, which `op.Status` could be either OpDone or OpPending [Code](https://github.com/googleapis/google-cloud-go/blob/main/container/container.go#L45-L56). I think we shall probably only proceed the following GCP `Get` when we get a OpDone. 

For OpPending, we probably want to return  quietly and update the KCC `.status` in the next Reconcile, maybe `Find/Update`? @justinsb I'll vote for Update, do you have a preference?

pkg/controller/direct/compute/forwardingrule_controller.go
```diff
@@ -0,0 +1,482 @@
...
+
+	forwardingRuleAdapter := &forwardingRuleAdapter{
+		resourceID: resourceID,
+		projectID:  projectID,
+		Location:   location,
+		desired:    obj,
+	}
+
+	// Get compute GCP client
+	if location == "global" {
```
yuwenma, line 172: This works for me! I think your approach is good enough for now and easy to maintain (which is really important).

pkg/controller/direct/compute/forwardingrule_controller.go
yuwenma: I think we have two questions:
1. What does the GCP service do if not providing the default value (i.e. does it use a different default)? 
2. I assume TF does this for a "good" reason. But we want to know the reason first to decide whether or not use the same behavior. And if you do not know why TF does it, you can assume you know nothing about this special handling. (I doubt if this is anything related to `state-into-spec`)

pkg/controller/direct/compute/forwardingrule_controller.go
```diff
@@ -0,0 +1,482 @@
...
+		}
+		updated, err = a.forwardingRulesClient.Get(ctx, getReq)
+	}
+	if err != nil {
+		return fmt.Errorf("getting ComputeForwardingRule %q failed: %w", a.resourceID, err)
+	}
+
+	status := &krm.ComputeForwardingRuleStatus{
+		LabelFingerprint:  updated.LabelFingerprint,
+		CreationTimestamp: op.Proto().InsertTime,
```
yuwenma, line 422: do we want to update this ?

pkg/controller/direct/compute/forwardingrule_controller.go
```diff
@@ -0,0 +1,482 @@
...
+// Delete implements the Adapter interface.
+func (a *forwardingRuleAdapter) Delete(ctx context.Context) (bool, error) {
+	if a.resourceID == "" {
+		return false, fmt.Errorf("resourceID is empty")
+	}
+
+	log := klog.FromContext(ctx).WithName(ctrlName)
+	log.V(2).Info("deleting ComputeForwardingRule", "name", a.resourceID)
+
+	exist, err := a.Find(ctx)
```
yuwenma, line 442: Have you tried deleting a non-exist forwarding rule? If the GCP service gives a clear signal. i.e. 404, then maybe we do not need this additional GCP call?

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrule/_generated_object_globalcomputeforwardingrule.golden.yaml
yuwenma: With `state-into-spec: Absent` (1.122 and 1.123), only users will set this `spec.ResourceID` field. Do you mean we want to have the test coverage here that users use the `spec.resourceID` as the primary name rather than the `metadata.name`?  
I see in this example the `metadata.name` and the  `spec.resourceID` are set to the same value, which does not check if `spec.resourceID` is used.

apis/refs/v1beta1/computerefs.go
gemmahou: We don't have ComputeNetworkGVK yet, currently no Sci-fi migration plan for this resource. I'll leave it for now and update later.

apis/refs/v1beta1/computerefs.go
gemmahou: We don't have ComputeAddressGVK yet, currently no Sci-fi migration plan for this resource. I'll leave it for now and update later.

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/regionalcomputeforwardingrule/_generated_object_regionalcomputeforwardingrule.golden.yaml
gemmahou: Handled TF default values [here](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/924dd21636081ba611483be4361f74068cf5c80d/pkg/controller/direct/compute/forwardingrule_controller.go#L129)

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrule/_generated_object_globalcomputeforwardingrule.golden.yaml
justinsb: Hmmm... maybe we don't then :-)  I was worried about compatibility, but as it's not server-generated I don't think it will trip anyone up!

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/regionalcomputeforwardingrule/_http.log
justinsb: It may not be an issue because it's in the URL

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/regionalcomputeforwardingrule/_http.log
```diff
@@ -624,22 +594,22 @@ X-Xss-Protection: 0
...
-User-Agent: Terraform/ (+https://www.terraform.io) Terraform-Plugin-SDK/2.10.1 terraform-provider-google-beta/kcc/controller-manager
+User-Agent: kcc/controller-manager
+x-goog-request-params: project=${projectId}&region=us-central1&resource=computeregionalforwardingrule-${uniqueId}
 
 {
   "labelFingerprint": "abcdef0123A=",
   "labels": {
     "cnrm-test": "true",
-    "label-one": "value-one",
+    "label-one": "value-two",
```
justinsb, line 612: This is the "always calls setLabels" behaviour you were talking about with the old controller?

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/regionalcomputeforwardingrule/_http.log
```diff
@@ -776,10 +715,12 @@ X-Xss-Protection: 0
 
 ---
 
-GET https://compute.googleapis.com/compute/v1/projects/${projectId}/global/operations/${operationID}?alt=json&prettyPrint=false
```
justinsb, line 779: I had thought this looked like we weren't checking the LRO on delete, but ... we are.

justinsb:
Looks good, a few questions but nothing blocking I don't think!  Will LGTM but add hold, feel free to address questions here or in a separate PR as you see fit.

/approve
/lgtm
/hold

pkg/controller/direct/compute/forwardingrule_controller.go
gemmahou: Removed the `Find` call here and verified the http logs

pkg/controller/direct/compute/forwardingrule_controller.go
gemmahou: This field is appearing in the realGCP log, so I think we should add it.

My bad I forgot to update this: `InsertTime` the insert of the operation, `CreationTimestamp` should be `updated.CreationTimestamp`

pkg/controller/direct/compute/forwardingrule_controller.go
```diff
@@ -0,0 +1,496 @@
...
+	if err != nil {
+		return fmt.Errorf("updating ComputeForwardingRule labels %s failed: %w", a.fullyQualifiedName(), err)
+	}
+	err = op.Wait(ctx)
+	if err != nil {
+		return fmt.Errorf("waiting ComputeForwardingRule %s update labels failed: %w", a.fullyQualifiedName(), err)
+	}
+	log.V(2).Info("successfully updated ComputeForwardingRule labels", "name", a.fullyQualifiedName())
+	// Get the updated resource
+	if a.id.location == "global" {
```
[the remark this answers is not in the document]
gemmahou: Call `Get` twice, once after updating labels and once after updating targets, to match the realGCP http log.

pkg/controller/direct/compute/forwardingrule_controller.go
```diff
@@ -0,0 +1,496 @@
...
+	externalRef := direct.ValueOf(obj.Status.ExternalRef)
+	if externalRef == "" {
+		id = BuildID(projectID, location, resourceID)
+	} else {
+		id, err = asID(externalRef)
+		if err != nil {
+			return nil, err
+		}
+
+		if id.project != projectID {
```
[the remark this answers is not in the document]
gemmahou: Yes, but this way, we have a better idea (via separate error messages) of which specific field has been changed.

pkg/controller/direct/compute/forwardingrule_controller.go
```diff
@@ -0,0 +1,496 @@
...
+	}
+
+	// Get network
+	if obj.Spec.NetworkRef != nil {
+		networkRef, err := ResolveComputeNetwork(ctx, reader, obj, obj.Spec.NetworkRef)
+		if err != nil {
+			return nil, err
+
+		}
+		obj.Spec.NetworkRef.External = networkRef.External
```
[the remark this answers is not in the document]
gemmahou: Do you mean cleaning up other fields like `Name` in `obj.Spec.NetworkRef`? I won't worry about other fields because we get value from External only, see the mapping function: [code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/e82cf9bb1aa018e2797eb5faeec634fb2fdd549d/pkg/controller/direct/compute/mapper.go#L88)

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/regionalcomputeforwardingrule/_http.log
gemmahou: Updated the log.
I think that's because I made a mistake in mockgcp/mockcompute/regionalforwardingrulev1.go: Should use return s.startRegionalLRO not return s.startGlobalLRO. I updated the mock as well.

pkg/controller/direct/compute/client.go
```diff
@@ -0,0 +1,101 @@
...
+
+type optionsRoundTripper struct {
+	config config.ControllerConfig
+	inner  http.RoundTripper
+}
+
+func (m *optionsRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	if m.config.UserAgent != "" {
+		req.Header.Set("User-Agent", m.config.UserAgent)
+		req.Header.Del("x-goog-request-params")
```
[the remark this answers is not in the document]
gemmahou: hmmm I'm actually unsure why this added... and seems it's not deleted by `Del`. I also noticed that other direct resources(monitoringdashboard, cloudbuildworkerpool) have this header as well, I would guess it comes from the underlying GCP?

pkg/controller/direct/compute/forwardingrule_controller.go
gemmahou: I won't worry about this because we call `op.[Wait](https://source.corp.google.com/piper///depot/google3/third_party/golang/cloud_google_com/go/compute/v/v1/apiv1/operations.go;l=48?q=apiv1%2Foperations.go&sq=package:piper%20file:%2F%2Fdepot%2Fgoogle3%20-file:google3%2Fexperimental)()` a few lines above, and it makes sure that the operation is Done.

pkg/controller/direct/compute/forwardingrule_controller.go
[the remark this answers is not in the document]
justinsb: Right, it might be clearer/safer if you did clear all those other fields, it's a nit because I agree it shouldn't change functionality.

pkg/controller/direct/compute/client.go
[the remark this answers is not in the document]
gemmahou: Removed!

justinsb (approved):

PR #2366, opened 2024-07-26
tools: implement types updater

jingyih:
A tool to update the Go types when a new field in GCP API proto is detected.

Future work
 - Automatically insert the generated code snippet into the right place.
 - Update mappers.
 - Update tests and docs, etc


dev/tools/controllerbuilder/pkg/updatetypes/main.go
```diff
@@ -0,0 +1,235 @@
...
+)
+
+// List of proto message types not mapped to Go structs
+var protoMsgsNotMappedToGoStruct = []string{
+	"google.protobuf.Timestamp",
+	"google.protobuf.Duration",
+	"google.protobuf.Int64Value",
+}
+
+func main() {
```
justinsb, line 39: Nit: please consider making this a subcommand of the main tool (using cobra)

dev/tools/controllerbuilder/pkg/updatetypes/main.go
```diff
@@ -0,0 +1,235 @@
...
+	config := parseFlags()
+	updator := NewTypeUpdater(config, protoMsgsNotMappedToGoStruct)
```
justinsb, line 41: Supernit: s/updator/updater!

dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
```diff
@@ -127,24 +127,24 @@ func (g *TypeGenerator) writeVisitedMessages() {
 		if out.contents.Len() == 0 {
 			fmt.Fprintf(w, "package %s\n", krmVersion)
 		}
-		writeMessage(w, msg)
+		WriteMessage(w, msg)
 	}
 }
 
-func writeMessage(out io.Writer, msg protoreflect.MessageDescriptor) {
+func WriteMessage(out io.Writer, msg protoreflect.MessageDescriptor) {
```
justinsb, line 134: Love the idea of refactoring our logic here to make it better & more reusable 👍

dev/tools/controllerbuilder/pkg/updatetypes/main.go
```diff
@@ -0,0 +1,235 @@
...
+	analyzeDependencies(field, deps, ignoredProtoFields)
+
+	// fmt.Println("\nDependencies:")
+	// for _, dep := range deps {
+	// 	fmt.Printf("%s\n", dep.FullName())
+	// }
+	return deps, nil
+}
+
+// analyzeDependencies recursively explores the dependent proto messages of the given field.
```
justinsb, line 151: Oooh this could be really useful.  Technically we only want to generate "reachable" messages from the "main resource proto".  Today, we just generate them all, which is overkill, and sometimes we have to fix bugs in messages we don't use anyway.

One for the future though, rather than this PR :-)

dev/tools/controllerbuilder/pkg/updatetypes/main.go
```diff
@@ -0,0 +1,235 @@
...
+func analyzeDependencies(field protoreflect.FieldDescriptor, deps map[string]protoreflect.MessageDescriptor, ignoreFields []string) {
```
justinsb, line 152: Nit: I think this could be called `findDependencies`?

dev/tools/controllerbuilder/pkg/updatetypes/main.go
```diff
@@ -0,0 +1,235 @@
...
+					analyzeDependencies(field, deps, ignoreFields)
+				}
+			}
+		case protoreflect.EnumKind:
+			// deps[string(field.Enum().FullName())] = true  // Skip enum because enum is mapped to Go string in code generation
+		}
+	}
+}
+
+func shouldIgnore(fieldName string, ignoreFields []string) bool {
```
justinsb, line 187: Tip: might be easier to pass a `StringSet` or a `Set[string]`, we already use the one from apimachinery in a few places.

dev/tools/controllerbuilder/pkg/updatetypes/main.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+func writeToLocalFile(field protoreflect.FieldDescriptor, parentMsg protoreflect.MessageDescriptor, msgs map[string]protoreflect.MessageDescriptor) error {
+	file, err := os.Create("output.txt")
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	fmt.Fprintf(file, "[The generated new field]\n\n")
```
justinsb, line 227: Small idea: I suggest being a little more verbose here, and possibly even printing a "link" to the file & line number if you know it.  (Many editors recognize [path]:[line-number], I believe)

Big idea: You could also try feeding those more verbose instructions into AI :-)  You could feed it in manually a few times to mess around with it and the prompt, and if that works ... maybe offer a mode where we try to apply the change!

justinsb (approved):

dev/tools/controllerbuilder/pkg/updatetypes/main.go
jingyih: Sounds good! I left a TODO in the code. Manually inserting the generated code back to the files was never the design intention. I am finding a way do this automatically.

dev/tools/controllerbuilder/pkg/updatetypes/main.go
jingyih: SG! #2373

dev/tools/controllerbuilder/pkg/updatetypes/main.go
jingyih: Done. Thanks!

dev/tools/controllerbuilder/pkg/updatetypes/main.go
jingyih: Thanks! Address in #2373

dev/tools/controllerbuilder/pkg/updatetypes/main.go
jingyih: Updated!

PR #2370, opened 2024-07-26
enh: direct controller template change

gemmahou:
### Change description

Fixes #

### Tests you have done


dev/tools/controllerbuilder/template/controller.go
```diff
@@ -253,18 +267,23 @@ func (a *Adapter) Export(ctx context.Context) (*unstructured.Unstructured, error
...
+	log.V(2).Info("deleting {{.Kind}}", "name", a.fullyQualifiedName())
+
 	if a.resourceID == "" {
 		return false, nil
 	}
 	req := &{{.Service}}pb.Delete{{.Kind}}Request{Name: a.fullyQualifiedName()}
 	op, err := a.gcpClient.Delete{{.Kind}}(ctx, req)
 	if err != nil {
-		return false, fmt.Errorf("deleting {{.Kind}} %s: %w", a.fullyQualifiedName(), err)
+		return false, fmt.Errorf("deleting {{.Kind}} %s failed: %w", a.fullyQualifiedName(), err)
```
yuwenma, line 279: nit: we don't really need to re-emphasize the failure because this returns the error already. It is also more like a convention. See error message in the doc https://go.dev/blog/error-handling-and-go

dev/tools/controllerbuilder/template/controller.go
```diff
@@ -297,10 +316,3 @@ func setStatus(u *unstructured.Unstructured, typedStatus any) error {
 
 	return nil
 }
-
-func ToOpenAPIDateTime(ts *timestamppb.Timestamp) *string {
```
yuwenma, line 301: Mind explaining why removing this?

dev/tools/controllerbuilder/template/controller.go
```diff
@@ -175,72 +179,82 @@ func (a *Adapter) Create(ctx context.Context, u *unstructured.Unstructured) erro
...
 	if err != nil {
-		return fmt.Errorf("{{.Kind}} %s waiting update failed: %w", resource.Name, err)
+		return fmt.Errorf("{{.Kind}} %s waiting update failed: %w", a.fullyQualifiedName(), err)
 	}
+	log.V(2).Info("successfully updated {{.Kind}}", "name", a.fullyQualifiedName())
+
 	status := &krm.{{.Kind}}Status{}
-	if err := krm.Convert_{{.Kind}}_API_v1_To_KRM_status(updated, status); err != nil {
-		return fmt.Errorf("update {{.Kind}} status %w", err)
+	status := {{.Kind}}Status_FromProto(mapCtx, updated)
```
yuwenma, line 252: This is not really a status-from-proto, because the `updated` maps to the `status.observedState`, not `.status`. 
You can look at this example to see how it should be called https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/cloudbuild/workerpool_controller.go#L196

yuwenma (commented):
some nites. overall looks good.

dev/tools/controllerbuilder/template/controller.go
gemmahou: This function has been moved to [maputils.go](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/2caba9831c3ba01b648504ca28e8c5d3fdcbfb99/pkg/controller/direct/maputils.go#L140)

dev/tools/controllerbuilder/template/controller.go
gemmahou: To facilitate debugging, I wanted to have a uniform pattern for our error messages.
Adding "failed" is consistent with other error messages (create, update, etc.) and distinguishes it from the above log.info message.

dev/tools/controllerbuilder/template/controller.go
gemmahou: Updated. 

Noted that not all beta resources have status.observedState now, we need to manually add this field for them.

dev/tools/controllerbuilder/template/externalresource.go
gemmahou, line 1: We could include a detailed example in this template in the future. However, I still chose to include it now so that we can use it as an example for users of the additional struct and functions required for the status.ExternalRef feature. Without this example, the entities({{.Kind}}Identity, BuildID, etc) in the direct controller file would appear out of nowhere.

pkg/controller/direct/statusutils.go
```diff
@@ -0,0 +1,40 @@
...
+package direct
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+func SetStatus(u *unstructured.Unstructured, typedStatus any) error {
```
yuwenma, line 24: Can you add a doc string to explain this method?

pkg/controller/direct/statusutils.go
```diff
@@ -0,0 +1,40 @@
...
+	status, err := runtime.DefaultUnstructuredConverter.ToUnstructured(typedStatus)
+	if err != nil {
+		return fmt.Errorf("error converting status to unstructured: %w", err)
+	}
+
+	old, _, _ := unstructured.NestedMap(u.Object, "status")
+	if old != nil {
```
yuwenma, line 31: This would require the API to contain all three fields (How can we verify that?) We shall call it out to make sure first-time contributor can understand.

pkg/controller/direct/statusutils.go
yuwenma, line 1: oh, just realized you put this file in direct. I think it may be not ready yet. i.e. What if we add another field in status? This would then break the resources added in between.

pkg/controller/direct/cloudbuild/externalresource.go
```diff
@@ -65,7 +65,7 @@ func asID(externalRef string) (*CloudBuildWorkerPoolIdentity, error) {
 	}, nil
 }
 
-// fromRaw builds a CloudBuildWorkerPoolIdentity from resource components.
+// BuildID builds a CloudBuildWorkerPoolIdentity from resource components.
```
yuwenma, line 68: Ah, thanks!

pkg/controller/direct/cloudbuild/workerpool_controller.go
```diff
@@ -106,7 +106,7 @@ func (m *model) AdapterForObject(ctx context.Context, reader client.Reader, u *u
 	if externalRef == "" {
 		id = BuildID(projectID, location, resourceID)
 	} else {
-		id, err = asID(externalRef)
```
yuwenma, line 109: So the reason we call it `asID` and `asExternalRef` at the beginning is because the structured ID and `status.externalRef` are equivalent. They are two different forms of representing a ConfigConnector real ID. I feel like `asID` better reveals this relationship. 

I'm good with the other renaming, it's nit. But I think we want to align here. 

@justinsb What do you feel about this?

dev/tools/controllerbuilder/template/externalresource.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// BuildIDFromExternal builds a {{.Kind}}Identity from a external reference
+func BuildIDFromExternal(external string) (*{{.Kind}}Identity, error) {
+	// TODO(user): Build resource identity from external reference
+}
+
+// BuildID builds a {{.Kind}}Identity from resource components
+func BuildID(project, location string) *{{.Kind}}Identity {
+	// TODO(user): Build resource identity from resource components, i.e. project, location, resource id
```
yuwenma, line 60: nit: Suggest also giving the code and ask users to fix. Ideally, they don't need to fix anything.

dev/tools/controllerbuilder/template/externalresource.go
```diff
@@ -0,0 +1,61 @@
...
+	"fmt"
+	"strings"
+)
+
+const (
+	// TODO(user): Add service domain
+	serviceDomain = "//{{.KindToLower}}.googleapis.com"
+)
+
+// TODO(user): Define resource identity
```
yuwenma, line 30: Suggest improving the comment to explain what the ResourceIdentity is.

dev/tools/controllerbuilder/template/externalresource.go
```diff
@@ -0,0 +1,61 @@
...
+type {{.Kind}}Identity struct {
+	project    string
+	location   string
+	{{.KindToLower}} string
```
yuwenma, line 34: I think we want to change the KindToLower to `kind` and `proto-source`
more detail in b/359404149. Can be in follow-up PRs

dev/tools/controllerbuilder/template/externalresource.go
yuwenma: I think we should. this is not a nice to have but a necessary feature. https://docs.google.com/document/d/1A6YSEygeVJuXefb9cH52RiqEqp15nI5-UHNqG1T08kg/edit?tab=t.0#heading=h.k6gmvtemah8s

dev/tools/controllerbuilder/template/externalresource.go
gemmahou: I can take that task if needed and follow up in another PR

pkg/controller/direct/statusutils.go
gemmahou: I see I can put it back and make it common method once we are ready

dev/tools/controllerbuilder/template/externalresource.go
gemmahou: Done. See commit https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2370/commits/fd8800efafc114b2b4f78968885c718c60e62f2b

Local test result: https://screenshot.googleplex.com/6f9xwCARtZs93km.png

pkg/controller/direct/cloudbuild/workerpool_controller.go
gemmahou: I reverted the naming changes back. I'm fine with either ways, as long as we have good docs to make the functions clear.

dev/tools/controllerbuilder/scaffold/controller.go
```diff
@@ -34,29 +34,78 @@ const (
 	directControllerRelPath = "pkg/controller/direct"
 )
 
-func Scaffold(path string, cArgs *ccTemplate.ControllerArgs) error {
+func Scaffold(service, kind string, cArgs *ccTemplate.ControllerArgs) error {
```
justinsb, line 37: Nit: This is pre-existing, but we normally import alias with all lower case i.e. `cctemplate`

dev/tools/controllerbuilder/scaffold/controller.go
```diff
@@ -34,29 +34,78 @@ const (
...
+	var errs []error
+	if err := generateController(service, kind, cArgs); err != nil {
+		errs = append(errs, err)
+	}
+	if err := generateControllerHelpers(service, kind, cArgs); err != nil {
+		errs = append(errs, err)
+	}
+	if len(errs) != 0 {
+		var finalError []string
```
justinsb, line 46: Nit: use `return errors.Join(errs...)`.  I think it even checks if len(errs) == 0 for you :-)

dev/tools/controllerbuilder/cmd/root.go
```diff
@@ -43,11 +43,7 @@ func buildAddCommand(baseOptions *options.GenerateOptions) *cobra.Command {
 				Kind:        kind,
 				KindToLower: strings.ToLower(kind),
 			}
-			path, err := scaffold.BuildControllerPath(baseOptions.ServiceName, kind)
-			if err != nil {
-				return err
-			}
-			return scaffold.Scaffold(path, cArgs)
+			return scaffold.Scaffold(baseOptions.ServiceName, kind, cArgs)
```
justinsb, line 46: Nit: I think we should just pass cArgs, as the first two args are already in cArgs. (it would get confusing if someone passed a different value for cArgs.service vs the first argument)

dev/tools/controllerbuilder/scaffold/controller.go
```diff
@@ -99,7 +155,8 @@ func WriteToFile(path string, out []byte) error {
 	if err := os.MkdirAll(filepath.Dir(path), 0755); err != nil {
 		return fmt.Errorf("failed to create directory %q: %w", filepath.Dir(path), err)
 	}
-	f, err := os.OpenFile(path, os.O_CREATE|os.O_WRONLY, 0644)
+	// Use O_TRUNC to truncate the file
+	f, err := os.OpenFile(path, os.O_TRUNC|os.O_CREATE|os.O_WRONLY, 0644)
```
justinsb, line 159: Nit: I think we can just use `os.WriteFile` ?

dev/tools/controllerbuilder/template/controller.go
```diff
@@ -114,15 +118,38 @@ func (m *model) AdapterForObject(ctx context.Context, reader client.Reader, u *u
...
+	externalRef := direct.ValueOf(obj.Status.ExternalRef)
+	if externalRef == "" {
+		id = BuildID(projectID, location, resourceID)
+	} else {
+		id, err = asID(externalRef)
+		if err != nil {
+			return nil, err
+		}
+
+		if id.project != projectID {
```
justinsb, line 131: There's opportunity to simplify this logic I think

justinsb (approved):

PR #2376, opened 2024-07-29
doc: fix description for CloudBuildTrigger

jingyih:
Fixes b/355644901


justinsb:
I'm actually wondering if we should fix these site-wide, it would be nice to have links in our docs where it is appropriate.  cc @yuwenma

jingyih:
@justinsb Are you suggesting we have different field description for CRD vs. webpage? AFAIK, today the same description is used by both CRD and on webpage, so using plain text makes sense to me.

yuwenma (commented):
/lgtm
/approve

@justinsb @jingyih we have tracked the work about improving the site doc in a buganizer issue. I'd like to get this PR merge in to clean the PR backlog

PR #2378, opened 2024-07-29
tools: refactor mapper generator to precompute the mapping

jingyih:
The mapping only needs to be computed once.

### DIFF
--- a/dev/tools/controllerbuilder/pkg/codegen/mappergenerator.go
+++ b/dev/tools/controllerbuilder/pkg/codegen/mappergenerator.go
@@ -35,12 +35,14 @@ type MapperGenerator struct {
 
 	goPackages []*gocode.Package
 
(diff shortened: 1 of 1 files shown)


dev/tools/controllerbuilder/pkg/codegen/mappergenerator.go
```diff
@@ -114,7 +114,7 @@ func (v *MapperGenerator) visitMessage(msg protoreflect.MessageDescriptor) {
 	}
 	goTypes := v.findKRMStructsForProto(msg)
 
-	if len(goTypes) == 0 {
+	if goTypes == nil || len(goTypes) == 0 {
```
justinsb, line 117: Nit: You don't need this, `len(nil) == 0`

justinsb (approved):

PR #2386, opened 2024-07-30
enh:kompanion: support goroutines

acpana:
Adds worker support so namespaces can be processed in parallel. 

I've also marked where I can add debug level logs and will add those as a follow on PR.

### DIFF
--- a/experiments/kompanion/cmd/export/export.go
+++ b/experiments/kompanion/cmd/export/export.go
@@ -26,11 +26,12 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v2"
-	v1 "k8s.io/api/core/v1"
+	// v1 "k8s.io/api/core/v1".
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -58,7 +59,7 @@ const (
 var ExportCmd = &cobra.Command{
 	Use:     "export",
 	Example: examples,
-	Run:     run,
+	RunE:    runE,
 	Args:    cobra.ExactArgs(0),
 }
 
@@ -100,14 +101,37 @@ var (
 	resources  []*metav1.APIResource
 )
 
-func processNamespace(ns v1.Namespace, dynamicClient *dynamic.DynamicClient) {
-	log.Printf("Processing namespace: %s", ns.Name)
-	if shouldExclude(ns.Name, ignoreNamespaces, targetNamespaces) {
+// tracks the namespaces to be exported.
+// thread safe.
+type namespacesWorkLog struct {
+	mu         sync.Mutex
+	namespaces []string // will be treated as a FIFO queue
+}
+
+var workLog *namespacesWorkLog
+
+func (n *namespacesWorkLog) GetWork() string {
+	n.mu.Lock()
+	defer n.mu.Unlock()
+
+	if len(n.namespaces) == 0 {
+		return ""
+	}
+
+	workItem := n.namespaces[0]
+	n.namespaces = n.namespaces[1:]
+
+	return workItem
+}
+
+func processNamespace(namespaceName string, dynamicClient *dynamic.DynamicClient) {
+	log.Printf("Processing namespace: %s", namespaceName)
+	if shouldExclude(namespaceName, ignoreNamespaces, targetNamespaces) {
 		return
 	}
 
-	if err := os.Mkdir(filepath.Join(reportName, ns.Name), os.ModePerm); err != nil {
-		log.Printf("Error creating directory for namespace %s:, err: %w; skipping", ns.Name, err)
+	if err := os.Mkdir(filepath.Join(reportName, namespaceName), 0o700); err != nil {
+		log.Fatalf("Error creating directory for namespace %s:, err: %w; skipping", namespaceName, err)
 		return
 	}
 
@@ -120,13 +144,13 @@ func processNamespace(ns v1.Namespace, dynamicClient *dynamic.DynamicClient) {
 		// todo acpana debug logs
 		// log.Printf("Looking for gvr %s in namespace %s", gvr, ns.Name)
 
-		resources, err := dynamicClient.Resource(gvr).Namespace(ns.Name).List(context.TODO(), metav1.ListOptions{})
+		resources, err := dynamicClient.Resource(gvr).Namespace(namespaceName).List(context.TODO(), metav1.ListOptions{})
 		if err != nil {
 			if apierrors.IsNotFound(err) {
-				log.Printf("gvr %s not found in namespace %s", gvr, ns.Name)
+				log.Printf("gvr %s not found in namespace %s", gvr, namespaceName)
 				continue
 			} else {
-				log.Printf("Error fetching gvr %s resources in namespace %s: %s\n", gvr, ns.Name, err)
+				log.Printf("Error fetching gvr %s resources in namespace %s: %s\n", gvr, namespaceName, err)
 				continue
 			}
 		}
@@ -140,21 +164,21 @@ func processNamespace(ns v1.Namespace, dynamicClient *dynamic.DynamicClient) {
 			}
 
 			// todo acpana if the file exists, log error but move on
-			filename := filepath.Join(reportName, ns.Name, r.GetName()+".yaml")
+			filename := filepath.Join(reportName, namespaceName, r.GetName()+".yaml")
 			data, err := yaml.Marshal(r)
 			if err != nil {
-				log.Printf("Error marshalling resource %s in namespace %s: %s\n", r.GetName(), ns.Name, err)
+				log.Printf("Error marshalling resource %s in namespace %s: %s\n", r.GetName(), namespaceName, err)
 				continue
 			}
 			err = ioutil.WriteFile(filename, data, 0o644)
 			if err != nil {
-				log.Printf("Error writing file for resource %s in namespace %s: %s\n", r.GetName(), ns.Name, err)
+				log.Printf("Error writing file for resource %s in namespace %s: %s\n", r.GetName(), namespaceName, err)
 			}
 		}
 	}
 }
 
-func run(_ *cobra.Command, _ []string) {
+func runE(_ *cobra.Command, _ []string) error {
 	log.Printf("Running kompanion export with kubeconfig: %s", kubeconfig)
 	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
 	if err != nil {
@@ -179,7 +203,7 @@ func run(_ *cobra.Command, _ []string) {
 	}
 
 	for _, apiResourceList := range apiResourceLists {
-		if !strings.Contains(apiResourceList.GroupVersion, "cnrm.") {
+		if !strings.Contains(apiResourceList.GroupVersion, ".cnrm.cloud.google.com/") {
 			// todo acpana log debug level
 			// log.Printf("ApiResource %s group doesn't contain \"cnrm\"; skipping", apiResourceList.GroupVersion)
 			continue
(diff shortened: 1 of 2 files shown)


experiments/kompanion/cmd/export/export.go
```diff
@@ -141,66 +248,60 @@ func run(_ *cobra.Command, _ []string) {
...
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
 
-		for _, res := range resources {
-			gvr := schema.GroupVersionResource{
-				Group:    res.Group,
-				Version:  res.Version,
-				Resource: res.Name,
+			if workLog.Done() {
```
justinsb, line 285: Can we just rely on GetWork() returning "" ?  Do we need the separate Done method?

experiments/kompanion/cmd/export/export.go
```diff
@@ -90,10 +95,110 @@ func init() {
 
 	ExportCmd.Flags().StringArrayVarP(&targetObjects, targetObjectsFlag, "", []string{}, "object name prefix to target. Targets all if empty. Can be specified multiple times.")
 	ExportCmd.Flags().StringArrayVarP(&ignoreObjects, ignoreObjectsFlag, "", []string{}, "object name prefix to ignore. Excludes nothing if empty. Can be specified multiple times.")
+
+	ExportCmd.Flags().IntVarP(&workerRountines, workerRoutinesFlag, "", 10, "Configure the number of worker rountines to export namespaces with. Defaults to 10. ")
```
justinsb, line 99: Nit: I don't think there's any reason to define a constant here.

experiments/kompanion/cmd/export/export.go
```diff
@@ -90,10 +95,110 @@ func init() {
...
+}
+
+var (
+	reportName string
```
justinsb, line 103: Nit: Global variables are generally an anti-pattern, better to use a struct to hold them.

experiments/kompanion/cmd/export/export.go
```diff
@@ -90,10 +95,110 @@ func init() {
...
+	resources  []*metav1.APIResource
+)
+
+// tracks the namespaces to be exported.
+// thread safe.
+type namespacesWorkLog struct {
```
justinsb, line 109: Nit: You could use generics here, and rename this to `ConcurrentQueue[T]` or something

experiments/kompanion/cmd/export/export.go
```diff
@@ -90,10 +95,110 @@ func init() {
...
+
+func (n *namespacesWorkLog) GetWork() string {
+	n.mu.Lock()
+	defer n.mu.Unlock()
+
+	if len(n.namespaces) == 0 {
+		return ""
+	}
+
+	workItem, remainig := n.namespaces[0], n.namespaces[1:]
```
justinsb, line 131: Nit: it's probably easier to write this out

```
workItem := n.namespaces[0]
n.namespaces = n.namespaces[1:]
```

(Also avoids the typo in remaining)

experiments/kompanion/cmd/export/export.go
```diff
@@ -90,10 +95,110 @@ func init() {
...
+}
+
+func processNamespace(namespaceName string, dynamicClient *dynamic.DynamicClient) {
+	log.Printf("Processing namespace: %s", namespaceName)
+	if shouldExclude(namespaceName, ignoreNamespaces, targetNamespaces) {
+		return
+	}
+
+	if err := os.Mkdir(filepath.Join(reportName, namespaceName), os.ModePerm); err != nil {
+		log.Printf("Error creating directory for namespace %s:, err: %w; skipping", namespaceName, err)
```
justinsb, line 144: I think you do want to return an error here, we're not going to get a valid dump

experiments/kompanion/cmd/export/export.go
```diff
@@ -90,10 +95,110 @@ func init() {
...
+	return workItem
+}
+
+func processNamespace(namespaceName string, dynamicClient *dynamic.DynamicClient) {
+	log.Printf("Processing namespace: %s", namespaceName)
+	if shouldExclude(namespaceName, ignoreNamespaces, targetNamespaces) {
+		return
+	}
+
+	if err := os.Mkdir(filepath.Join(reportName, namespaceName), os.ModePerm); err != nil {
```
justinsb, line 143: Nit: I would use 0700 or 0755 (depending on whether it was sensitive).  I thought ModePerm was a mask.  Are you hoping that the umask makes this reasonable?

(You might also consider streaming directly into the tar.gz or zip file)

experiments/kompanion/cmd/export/export.go
```diff
@@ -90,10 +95,110 @@ func init() {
...
+		log.Printf("Error creating directory for namespace %s:, err: %w; skipping", namespaceName, err)
+		return
+	}
+
+	for _, res := range resources {
+		gvr := schema.GroupVersionResource{
```
justinsb, line 149: Nit: I would have guess this would be the slower loop, rather than namespaces, and thus more important to parallelize

experiments/kompanion/cmd/export/export.go
```diff
@@ -90,10 +95,110 @@ func init() {
...
+			Group:    res.Group,
+			Version:  res.Version,
```
justinsb, line 151: Nit: we should double check that we only get the preferred version here.  That said, not a blocker, because if we get multiple versions of the same object we'll notice that pretty quickly in the tar.gz!

experiments/kompanion/cmd/export/export.go
```diff
@@ -90,10 +95,110 @@ func init() {
...
+		// todo acpana debug logs
+		// log.Printf("Looking for gvr %s in namespace %s", gvr, ns.Name)
+
+		resources, err := dynamicClient.Resource(gvr).Namespace(namespaceName).List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				log.Printf("gvr %s not found in namespace %s", gvr, namespaceName)
+				continue
+			} else {
+				log.Printf("Error fetching gvr %s resources in namespace %s: %s\n", gvr, namespaceName, err)
```
justinsb, line 163: I think I would always return an error.  A List only returns NotFound if the CRD isn't found, IIRC, which is surprising.  And I think you want to report all the errors at the end, and probably still write the tar.gz file, and (ideally) write the errors both to stderr and to the tar.gz file.

experiments/kompanion/cmd/export/export.go
```diff
@@ -90,10 +95,110 @@ func init() {
...
+		// log.Printf("Found %d resources of type gvr %s in namespace %s", len(resources.Items), gvr, ns.Name)
+
+		for _, r := range resources.Items {
+			if shouldExclude(r.GetName(), ignoreObjects, targetObjects) {
+				continue
+			}
+
+			// todo acpana if the file exists, log error but move on
+			filename := filepath.Join(reportName, namespaceName, r.GetName()+".yaml")
+			data, err := yaml.Marshal(r)
```
justinsb, line 178: Please use `sigs.k8s.io/yaml`.  We should always use that library with k8s objects.

experiments/kompanion/cmd/export/export.go
```diff
@@ -90,10 +95,110 @@ func init() {
...
+			if err != nil {
+				log.Printf("Error marshalling resource %s in namespace %s: %s\n", r.GetName(), namespaceName, err)
```
justinsb, line 180: Nit: I would return (and accumulate) this error

experiments/kompanion/cmd/export/export.go
```diff
@@ -90,10 +95,110 @@ func init() {
...
+				continue
+			}
+			err = ioutil.WriteFile(filename, data, 0o644)
+			if err != nil {
+				log.Printf("Error writing file for resource %s in namespace %s: %s\n", r.GetName(), namespaceName, err)
```
justinsb, line 185: Ditto: return & accumulate this error

experiments/kompanion/cmd/export/export.go
```diff
@@ -116,20 +221,22 @@ func run(_ *cobra.Command, _ []string) {
 		log.Fatalf("Failed to get preferred resources: %s", err)
 	}
 
-	resources := []*metav1.APIResource{}
 	for _, apiResourceList := range apiResourceLists {
 		if !strings.Contains(apiResourceList.GroupVersion, "cnrm.") {
```
justinsb, line 225: Nit: `.cnrm.cloud.google.com/` might be less likely to accidentally collect the wrong thing

experiments/kompanion/cmd/export/export.go
```diff
@@ -116,20 +221,22 @@ func run(_ *cobra.Command, _ []string) {
...
-			log.Printf("ApiResource %s group doesn't contain \"cnrm\"; skipping", apiResourceList.GroupVersion)
+			// todo acpana log debug level
```
justinsb, line 226: Just use `log := klog.FromContext(ctx)` and the `log.V(1).Info` (or whatever)

experiments/kompanion/cmd/export/export.go
```diff
@@ -116,20 +221,22 @@ func run(_ *cobra.Command, _ []string) {
...
 		}
 
 		for _, apiResource := range apiResourceList.APIResources {
 			if !apiResource.Namespaced {
-				log.Printf("ApiResource %s is not namespaced; skipping", apiResource.SingularName)
+				// todo acpana log debug level
+				// log.Printf("ApiResource %s is not namespaced; skipping", apiResource.SingularName)
 				continue
 			}
 			if !contains(apiResource.Verbs, "list") {
```
justinsb, line 237: Nice!

experiments/kompanion/cmd/export/export.go
```diff
@@ -141,66 +248,60 @@ func run(_ *cobra.Command, _ []string) {
...
 
 	namespaces, err := clientset.CoreV1().Namespaces().List(context.TODO(), metav1.ListOptions{})
 	if err != nil {
 		log.Fatalf("Error fetching namespaces: %s\n", err)
 	}
 
-	reportName := timestampedName("report")
+	reportName = timestampedName("report")
 	if err := os.Mkdir(filepath.Join(reportName), os.ModePerm); err != nil {
 		log.Fatalf("Could not create \"report\" directory: %w", err)
```
justinsb, line 262: Nit: I always like to return the error, rather than panic-ing.  Cobra has a RunE method you can implement instead of Run

experiments/kompanion/cmd/export/export.go
```diff
@@ -141,66 +248,60 @@ func run(_ *cobra.Command, _ []string) {
...
+	// create the work log for go routine workers to use
+	workLog = &namespacesWorkLog{}
 	for _, ns := range namespaces.Items {
+		workLog.namespaces = append(workLog.namespaces, ns.Name)
+	}
 
-		log.Printf("Processing namespace: %s", ns.Name)
-		if shouldExclude(ns.Name, ignoreNamespaces, targetNamespaces) {
-			continue
+	// todo acpana parametrize as a flag
```
justinsb, line 271: I think you did this one?

experiments/kompanion/cmd/export/export.go
```diff
@@ -141,66 +248,60 @@ func run(_ *cobra.Command, _ []string) {
...
+	log.Printf("Starting worker threads to process namespaces")
+	var wg sync.WaitGroup
+	for i := 0; i < workerRountines; i++ {
+		if workLog.Done() {
```
justinsb, line 275: Nit: I'd remove this as well, launching a goroutine is fast.

experiments/kompanion/cmd/export/export.go
```diff
@@ -141,66 +248,60 @@ func run(_ *cobra.Command, _ []string) {
...
-				if err != nil {
-					log.Printf("Error writing file for resource %s in namespace %s: %s\n", r.GetName(), ns.Name, err)
-				}
+				processNamespace(ns, dynamicClient)
 			}
-		}
+		}()
 	}
 
+	log.Printf("Waiting on wait group")
```
justinsb, line 301: Maybe rename to something more user-friendly, like "Dumping Config Connector objects to report.tar.gz"

experiments/kompanion/cmd/export/export.go
acpana: Should we assume no parallelism in the case where the flag is not set?

experiments/kompanion/cmd/export/export.go
acpana: Hm I thought `ModePerm` was o0777  as per https://pkg.go.dev/os but maybe i'm reading that wrong. Let's go w 0700 for now. 

Let me address the streaming in a separate PR since I'm also working on the cleanup and that will make cleaning up moot 😛 .

experiments/kompanion/cmd/export/export.go
acpana: That's a good point. Let me follow up on parallelizing the group iterations

experiments/kompanion/cmd/export/export.go
acpana: We can use a golanci linter for this!

experiments/kompanion/cmd/export/export.go
acpana: Sure I was thinking that in the case where goroutines is like 100 but we only have 2-3 work items then we might still be spawning goroutines when the work has been already completed but I'll take it out.

experiments/kompanion/cmd/export/export.go
acpana: Let me address the log verbosity as a follow on if that's ok! Right now the "debug" are off.

experiments/kompanion/cmd/export/export.go
justinsb: I'd go with 1000 :-)

experiments/kompanion/cmd/export/export.go
justinsb: Yes, but we have existing places in our code that don't follow the rule.

justinsb (approved):

experiments/kompanion/cmd/export/export.go
cheftako: Even with 10 already seeing a lot of :
I0802 16:28:52.334712 2829986 request.go:697] Waited for 1.156100829s due to client-side throttling, not priority and fairness, request: GET:https://34.134.156.190/apis/accesscontextmanager.cnrm.cloud.google.com/v1beta1/namespaces/gke-managed-system/accesscontextmanageraccesspolicies

PR #2390, opened 2024-07-31
fix: Update compute mocks

gemmahou:
### Change description

TF uses compute/beta/ endpoint, to avoid log diffs after migrating to Scifi, I made below changes:

The urls should be v1 `https://compute.googleapis.com/compute/v1/`
The self links should have format `https://www.googleapis.com/compute/v1/...`

### Tests you have done

### DIFF
--- a/tests/e2e/unified_test.go
+++ b/tests/e2e/unified_test.go
@@ -475,11 +475,20 @@ func runScenario(ctx context.Context, t *testing.T, testPause bool, fixture reso
 
 					// Replace any dynamic IDs that appear in URLs
 					for _, event := range events {
-						url := event.Request.URL
+						u := event.Request.URL
 						for k, v := range pathIDs {
-							url = strings.ReplaceAll(url, "/"+k, "/"+v)
+							u = strings.ReplaceAll(u, "/"+k, "/"+v)
 						}
-						event.Request.URL = url
+
+						// Specific to Compute
+						// Terraform uses the /beta/ endpoints, but mocks and direct controller should use /v1/
+						// This special handling to avoid diffs in http logs.
+						// This can be removed once all Compute resources are migrated to direct controller.
+						basePath := "https://compute.googleapis.com/compute"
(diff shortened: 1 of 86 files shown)


justinsb:
Could you start the direct controller using the same API version as the existing controllers?

gemmahou:
> Could you start the direct controller using the same API version as the existing controllers?

Per this [comment](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/mockgcp/mockcompute/service.go#L198) I guess we want to switch to v1 eventually? 
Are you suggesting using beta version as the existing controller for easier testing and ensure backward compatibility? 

Other than the API version, the selflink format in compute mocks are inconsistent(should be `https://www.googleapis.com/compute/v1`, b/131629538), I can address that in this PR.

justinsb:
> Are you suggesting using beta version as the existing controller for easier testing and ensure backward compatibility?

Right - I think that should be easier.  Otherwise I think all our tests will start failing with the TF-based controllers?

tests/e2e/unified_test.go
```diff
@@ -475,11 +475,20 @@ func runScenario(ctx context.Context, t *testing.T, testPause bool, fixture reso
...
 						for k, v := range pathIDs {
-							url = strings.ReplaceAll(url, "/"+k, "/"+v)
+							u = strings.ReplaceAll(u, "/"+k, "/"+v)
+						}
+
+						// Specific to Compute
+						// Terraform uses the /beta/ endpoints, but mocks and direct controller should use /v1/
+						// This special handling to avoid diffs in http logs.
+						// This can be removed once all Compute resources are migrated to direct controller.
+						basePath := "https://compute.googleapis.com/compute"
```
justinsb, line 487: Oh I see!  I guess this is a nice trick also.  @yuwenma  what do you think?

yuwenma: works for me.

PR #2403, opened 2024-08-01
[DO NOT MERGE] Smooth promo for cbwp 

yuwenma:
### Issue
Current alpha to beta promo breaks if user already manually installs alpha CRDs, which is what happened for cbwp.
   
### Quick fix

Support both alpha and beta in the  CRD level (mark alpha as deprecated), This should unblock CC and CCC installation bundle to release.

we cannot merge this PR directly because it has been integrated in the PRESUBMIT process. Merging it will block all PRs to merge in and itself will be overridden.

Long term fix should be on the CRD level and follow the k8s version promo process.


yuwenma:
cc @cheftako @justinsb @jingyih I plan to cherry-pick this change to patch on 1.121.0 release. thoughts?

crds/cloudbuild_v1beta1_cloudbuildworkerpool.yaml
```diff
@@ -23,6 +23,8 @@ metadata:
     cnrm.cloud.google.com/stability-level: beta
   name: cloudbuildworkerpools.cloudbuild.cnrm.cloud.google.com
 spec:
+  conversion:
+    strategy: None
```
yuwenma, line 27: Ideally we shall use a webhook here because the cbwp alpha to beta changes a field
`spec.privatePoolV1Config.networkConfig.peeredNetworkIpRange` --> `spec.privatePoolV1Config.networkConfig.peeredNetworkIPRange`

apis/cloudbuild/v1alpha1/conversion.go
```diff
@@ -0,0 +1,150 @@
...
+
+package v1alpha1
+
+import (
+	"fmt"
+
+	cloudbuildpb "cloud.google.com/go/cloudbuild/apiv1/v2/cloudbuildpb"
+)
+
+func Convert_WorkerPool_API_v1_To_KRM_status(in *cloudbuildpb.WorkerPool, out *CloudBuildWorkerPoolStatus) error {
```
justinsb, line 23: I suspect these functions are now unused, because our controller is using v1beta1 (?)

pkg/k8s/crds.go
```diff
@@ -20,22 +20,46 @@ import (
 	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 )
 
+var scoreForVersion = map[string]int{
+	"v1alpha1": 111,
+	"v1beta1":  121,
+	"v1":       131,
+}
+
+func PreferredVersion(crd *apiextensions.CustomResourceDefinition) *apiextensions.CustomResourceDefinitionVersion {
```
justinsb, line 29: Oh nice, I prefer this approach to mine - mine escalated way too quickly :-)

tests/apichecks/testdata/exceptions/jsonnaming.txt
```diff
@@ -0,0 +1 @@
+[[jsonNaming] crd=cloudbuildworkerpools.cloudbuild.cnrm.cloud.google.com version=v1alpha1: field ".spec.privatePoolV1Config.networkConfig.PeeredNetworkIPRange" should use camel case]
```
justinsb, line 1: I do think it's fine to rename the v1alpha1 version

justinsb:
LGTM, I'll approve and hold in case you want to wait for test etc

/approve
/lgtm
/hold

tests/apichecks/testdata/exceptions/jsonnaming.txt
yuwenma: sure.

apis/cloudbuild/v1alpha1/conversion.go
yuwenma: Good, point! removed.

PR #2406, opened 2024-08-02
Support multiple versions in a CRD

justinsb:


justinsb:
/hold

I think this is too risky for this stage of the release.

We would likely do better hard-coding the GVK for the Kind(s) which have multiple versions, at least for now

cc @yuwenma

PR #2407, opened 2024-08-02
feat: Use externalRef as the true identity in cbwp

yuwenma:
### Change description

Fixes #

### Tests you have done


pkg/controller/direct/cloudbuild/externalresource.go
```diff
@@ -19,56 +19,57 @@ package cloudbuild
...
-		return &ResourceRef{
-			project:     segments[0],
-			location:    segments[2],
-			resourceID:  segments[4],
-			externalRef: baseResourceRef.Get(),
-		}, nil
-	}
-	return nil, fmt.Errorf("externalReference should be in the form of <baseUrl>/projects/<project>/locations/<location>/workerPools/<workerPool>, got %s",
-		extResRef)
+type CloudBuildWorkerPoolIDentity struct {
```
justinsb, line 28: Nit: Typo

pkg/controller/direct/cloudbuild/externalresource.go
```diff
@@ -19,56 +19,57 @@ package cloudbuild
...
+// FullyQualifiedName builds a CloudBuildWorkerPool resource of the format projects/<project>/locations/<location>/workerPools/<workerPool>
+func (c *CloudBuildWorkerPoolIDentity) FullyQualifiedName() string {
+	return fmt.Sprintf("projects/%s/locations/%s/workerPools/%s", c.project, c.location, c.workerpool)
 }
 
-func (e *ResourceRef) GetResourceID() string {
-	return e.resourceID
+// fromID builds a externalRef from a CloudBuildWorkerPoolIDentity
+func asExternalRef(id *CloudBuildWorkerPoolIDentity) *string {
+	e := fmt.Sprintf("%sprojects/%s/locations/%s/workerPools/%s", serviceDomain, id.project, id.location, id.workerpool)
```
justinsb, line 46: Nit: You might prefer to have the service domain without the leading/trailing slashes.  Also you might consider having this one be more like `return serviceDomain + id.FullyQualifiedName()`

pkg/controller/direct/cloudbuild/externalresource.go
```diff
@@ -19,56 +19,57 @@ package cloudbuild
...
-	return e.externalRef
+// FullyQualifiedName builds a CloudBuildWorkerPool resource of the format projects/<project>/locations/<location>/workerPools/<workerPool>
+func (c *CloudBuildWorkerPoolIDentity) FullyQualifiedName() string {
+	return fmt.Sprintf("projects/%s/locations/%s/workerPools/%s", c.project, c.location, c.workerpool)
 }
 
-func (e *ResourceRef) GetResourceID() string {
-	return e.resourceID
+// fromID builds a externalRef from a CloudBuildWorkerPoolIDentity
+func asExternalRef(id *CloudBuildWorkerPoolIDentity) *string {
```
justinsb, line 45: Any reason for this not to be a method on CloudBuildWorkerPoolIdentity?

pkg/controller/direct/cloudbuild/externalresource.go
```diff
@@ -19,56 +19,57 @@ package cloudbuild
...
+	e := fmt.Sprintf("%sprojects/%s/locations/%s/workerPools/%s", serviceDomain, id.project, id.location, id.workerpool)
+	return &e
 }
 
-func (e *ResourceRef) GetLocation() string {
-	return e.location
+// asID builds a CloudBuildWorkerPoolIDentity from a externalRef
+func asID(externalRef string) (*CloudBuildWorkerPoolIDentity, error) {
+	if !strings.HasPrefix(externalRef, serviceDomain) {
+		return nil, fmt.Errorf("externalRef shall has prefix %s, got %s", serviceDomain, externalRef)
```
justinsb, line 53: Nit: `externalRef should have prefix %q, got %q` or `externalRef %q should have prefix %q`

pkg/controller/direct/cloudbuild/externalresource.go
```diff
@@ -19,56 +19,57 @@ package cloudbuild
...
+		project:    tokens[1],
+		location:   tokens[3],
+		workerpool: tokens[5],
+	}, nil
 }
 
-func (e *ResourceRef) GetProject() string {
-	return e.project
+// fromRaw builds a CloudBuildWorkerPoolIDentity from resource components.
+func fromRaw(project, location, workerpool string) *CloudBuildWorkerPoolIDentity {
```
justinsb, line 69: Nit: fromRaw is not the most obvious name when you're calling it.  Maybe `buildCloudBuildWorkerPoolIdentity` ?  (and then the string-parsing version could be `parseCloudBuildWorkerPoolIdentity`)

Because this is in the CloudBuild package we could also do `buildWorkerPoolIdentity`

pkg/controller/direct/cloudbuild/workerpool_controller.go
```diff
@@ -98,10 +97,34 @@ func (m *model) AdapterForObject(ctx context.Context, reader client.Reader, u *u
...
+
+	externalRef := direct.ValueOf(obj.Status.ExternalRef)
+	if externalRef == "" {
+		id = fromRaw(projectID, location, resourceID)
+	} else {
+		id, err = asID(externalRef)
+		if err != nil {
+			return nil, err
+		}
+		if id.project != projectID {
```
justinsb, line 113: I agree this is a good idea.

I wonder if we can just struct comparison (maybe reflect.DeepEqual ... we just want to make sure that non-exported fields are compared)

pkg/controller/direct/cloudbuild/workerpool_controller.go
```diff
@@ -163,25 +178,17 @@ func (a *Adapter) Create(ctx context.Context, u *unstructured.Unstructured) erro
...
 	if mapCtx.Err() != nil {
 		return mapCtx.Err()
 	}
-	wp.Name = a.fullyQualifiedName()
+	wp.Name = a.id.FullyQualifiedName()
 	req := &cloudbuildpb.CreateWorkerPoolRequest{
-		Parent:       a.getParent(),
-		WorkerPoolId: a.resourceID,
+		Parent:       a.id.Parent(),
+		WorkerPoolId: a.id.workerpool,
```
justinsb, line 191: This is really nice!

pkg/controller/direct/cloudbuild/workerpool_controller.go
```diff
@@ -198,19 +205,11 @@ func (a *Adapter) Create(ctx context.Context, u *unstructured.Unstructured) erro
 	if mapCtx.Err() != nil {
 		return mapCtx.Err()
 	}
-	resRef, err := NewResourceRef(created)
-	if err != nil {
-		return err
-	}
-	status.ExternalRef = resRef.GetExternalReference()
+	status.ExternalRef = asExternalRef(a.id)
```
justinsb, line 208: Yeah, I think this would read more fluently if asExternalRef was a method on the struct

pkg/controller/direct/cloudbuild/workerpool_controller.go
```diff
@@ -292,11 +291,7 @@ func (a *Adapter) Update(ctx context.Context, u *unstructured.Unstructured) erro
 		return fmt.Errorf("update workerpool status %w", mapCtx.Err())
 	}
 	// This value should not be updated. Just in case.
```
justinsb, line 293: Looks like we are no longer parsing the value, so maybe drop the comment (and possibly there's no point setting ExternalRef if we're just setting it back?)

justinsb (approved):

pkg/controller/direct/cloudbuild/externalresource.go
yuwenma: I like it. May prefer keeping the leading slash in serviceDomain, just to make it easier to claim the form as the asset inventory.

pkg/controller/direct/cloudbuild/externalresource.go
yuwenma: Changed.

pkg/controller/direct/cloudbuild/externalresource.go
yuwenma: good catch! Thanks!

pkg/controller/direct/cloudbuild/externalresource.go
yuwenma: So I defined 3 methods here and want them to discovered with a clear indication of their usage:
- `asID` asks for string ID and returns structured ID
- `asExternalRef` asks for structured ID and returns the string ID. It is the opposite of `asID`
- `fromRaw` asks for the each segment value, and returns the structured ID. 
-

pkg/controller/direct/cloudbuild/workerpool_controller.go
yuwenma: Gotcha. Updated.

pkg/controller/direct/cloudbuild/workerpool_controller.go
yuwenma: Sure

PR #2382, opened 2024-07-30
mockgcp: improve artifactregistry fidelity

justinsb:
- **mockgcp: improve artifactregistry fidelity**
- **tests: update artifactregistryrepository golden output**

PR #2410, opened 2024-08-02
mockgcp: improve fidelity of serviceusage

justinsb:
- **mockgcp: improve fidelity of serviceusage**
- **tests: update golden output for serviceusage**

### DIFF
--- a/tests/e2e/normalize.go
+++ b/tests/e2e/normalize.go
@@ -602,6 +602,7 @@ func normalizeHTTPResponses(t *testing.T, events test.LogEntries) {
 	visitor.replacePaths[".etag"] = "abcdef0123A="
 	visitor.replacePaths[".response.etag"] = "abcdef0123A="
 	visitor.replacePaths[".serviceAccount.etag"] = "abcdef0123A="
+	visitor.replacePaths[".response.uniqueId"] = "12345678"
 
 	// Compute operations
 	visitor.replacePaths[".fingerprint"] = "abcdef0123A="
--- a/mockgcp/mockserviceusage/service.go
+++ b/mockgcp/mockserviceusage/service.go
@@ -18,10 +18,10 @@ import (
 	"context"
 	"net/http"
 
-	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	"google.golang.org/grpc"
 
 	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/common"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/common/httpmux"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/common/operations"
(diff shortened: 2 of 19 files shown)


justinsb:
/hold

This includes #2382 so should not be merged before that

justinsb:
#2382 has merged, so I rebased.

/hold cancel

tests/e2e/normalize.go
```diff
@@ -438,6 +438,9 @@ func normalizeHTTPResponses(t *testing.T, events test.LogEntries) {
 	visitor.replacePaths[".fingerprint"] = "abcdef0123A="
 	visitor.replacePaths[".startTime"] = "2024-04-01T12:34:56.123456Z"
 
+	// ServiceIdentity operations
+	visitor.replacePaths[".response.uniqueId"] = "12345678"
```
yuwenma, line 442: Looks like we have two ways to override the uniqueID. I'll vote for the replace-by-kind approach
For Mock test development, one problem (the most time consuming part) is to figure out the polluted overridden fields.  
https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/a9c4afddb4f51a9dac376335303ba34f17e5e5df/mockgcp/mockserviceusage/serviceusagev1beta1.go#L53

yuwenma (commented):
/approve

One comment. Happy to discuss more.

yuwenma:
I think we don't need the "response.uniqueId" overrides. Otherwise the PR looks good

justinsb:
>I think we don't need the "response.uniqueId" overrides. Otherwise the PR looks good

Removed so we can see what happens.  I think it's going to fail the real-gcp tests

justinsb:
So the tests fail:
```
2024-10-17T14:51:32.6328314Z     utils.go:110: unexpected diff in /home/runner/work/k8s-config-connector/k8s-config-connector/pkg/test/resourcefixture/testdata/iammemberreferences/serviceidentityref/_http.log:   strings.Join({
2024-10-17T14:51:32.6330498Z           	... // 645 identical bytes
2024-10-17T14:51:32.6331680Z           	"beta1.ServiceIdentity\",\n    \"email\": \"service-${projectNumber}@g",
2024-10-17T14:51:32.6333144Z           	"cp-sa-pubsub.iam.gserviceaccount.com\",\n    \"uniqueId\": \"12345678",
2024-10-17T14:51:32.6334074Z         + 	"9002",
2024-10-17T14:51:32.6335174Z           	"\"\n  }\n}\n\n---\n\nGET https://artifactregistry.googleapis.com/v1/pro",
2024-10-17T14:51:32.6336612Z           	"jects/${projectId}/locations/us-west1/repositories/artifactregis",
2024-10-17T14:51:32.6337572Z           	... // 9331 identical bytes
2024-10-17T14:51:32.6338186Z           }, "")
```

We're using a fixed ID in the mocks, but when we run against real GCP the ID won't be that value...  I think we need to normalize therefore @yuwenma ?

PR #2414, opened 2024-08-05
tests: more complete compute normalization

justinsb:
We were only normalizing a handful of the URLs from the compute requests.

### DIFF
--- a/tests/e2e/normalize.go
+++ b/tests/e2e/normalize.go
@@ -431,11 +431,37 @@ func normalizeHTTPResponses(t *testing.T, events test.LogEntries) {
 	visitor.replacePaths[".fingerprint"] = "abcdef0123A="
 	visitor.replacePaths[".startTime"] = "2024-04-01T12:34:56.123456Z"
 
+	// Compute URLs: Replace any compute beta URLs with v1 URLs
+	// Terraform uses the /beta/ endpoints, but mocks and direct controller should use /v1/
+	// This special handling to avoid diffs in http logs.
+	// This can be removed once all Compute resources are migrated to direct controller.
+	for _, event := range events {
+		event.Request.URL = rewriteComputeURL(event.Request.URL)
+	}
+
+	visitor.stringTransforms = append(visitor.stringTransforms, func(path string, s string) string {
+		switch path {
+		case ".selfLink", ".targetLink", ".selfLinkWithId", ".subnetworks[]":
+			return rewriteComputeURL(s)
+		}
+		return s
+	})
+
 	events.PrettifyJSON(func(obj map[string]any) {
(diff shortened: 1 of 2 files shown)


pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagvalue/tagvalueproject/_http.log
```diff
@@ -148,7 +148,7 @@ X-Xss-Protection: 0
     "createTime": "2024-04-01T12:34:56.123456Z",
     "description": "For valuename resources.",
     "etag": "abcdef0123A=",
-    "name": "tagValues/${tagValueID}",
+    "name": "tagValues/${tagValueId}",
```
acpana, line 151: Should these be reverted?

tests/e2e/normalize.go
```diff
@@ -431,13 +431,39 @@ func normalizeHTTPResponses(t *testing.T, events test.LogEntries) {
...
 	})
 }
 
+// Compute URLs: Replace any compute beta URLs with v1 URLs
+func rewriteComputeURL(u string) string {
+	for _, basePath := range []string{"https://compute.googleapis.com/compute", "https://www.googleapis.com/compute"} {
+		if strings.HasPrefix(u, basePath+"/beta/") {
+			u = basePath + "/v1/" + strings.TrimPrefix(u, basePath+"/beta/")
+		}
+	}
```
acpana, line 463: Maybe outside of the scope of this PR -- do we want to only rewrite URLs for TF based resources?

acpana (commented):
/lgtm
/approve

Happy to take another look after the tagstagvalue id -> ID . I think the sample failure is a flake sadly.

tests/e2e/normalize.go
justinsb: I don't think we should be rewriting URLs at all.  But right now our mock is different from real.

pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagvalue/tagvalueproject/_http.log
justinsb: I don't believe so.  We've been inconsistent in our normalization, but ID seems to be much more common than Id.  The reason for the change is that we actually change tagValues in two places, I think I will actually just remove the second to avoid confusion.

justinsb:
Thanks for looking @acpana - the diff is now _much_ shorter now that the LRO stuff has been split out!  Hopefully it passes tests now!

yuwenma:
/lgtm

Looks good.

PR #2423, opened 2024-08-05
chore: set kube version to v0.27.11

justinsb:
This work makes go work work.


justinsb:
We are now well past kube 1.27 - thank you @cheftako 

/close

PR #2426, opened 2024-08-06
tests: capture compute LROs also

justinsb:
- **tests: capture compute LROs also**
- **tests: Capture golden output including compute LROs**


yuwenma (commented):
/lgtm
/approve

Ah, so the `isGetOperation` previously missed the LRO GET because the URL does not match? Good catch! thank you!

PR #2429, opened 2024-08-06
Add CloudSQL support for "create from clone"

jasonvigil:
### Change description

Depends on #2393 


apis/sql/v1beta1/sqlinstance_types.go
```diff
@@ -437,7 +437,36 @@ type InstanceValueFrom struct {
...
+	/* DatabaseNames: (SQL Server only) Clone only the specified databases from the
+	source instance. Clone all databases if empty. */
+	DatabaseNames []string `json:"databaseNames,omitempty"`
+
+	/* PointInTime: Timestamp, if specified, identifies the time to which the
+	source instance is cloned. */
+	PointInTime string `json:"pointInTime,omitempty"`
+
+	/* SourceID: The ID of the source database instance to clone */
+	SourceID string `json:"sourceID,omitempty"`
```
maqiuyujoyce, line 463: This looks like a reference to me. Can we reference a SQLInstance here?

apis/sql/v1beta1/sqlinstance_types.go
```diff
@@ -437,7 +437,36 @@ type InstanceValueFrom struct {
...
+
+type CloneSource struct {
+	/* BinLogCoordinates: Binary log coordinates, if specified, identify the
+	position up to which the source instance is cloned. If not specified, the
+	source instance is cloned up to the most recent binary log coordinates. */
+	BinLogCoordinates *BinLogCoordinates `json:"binLogCoordinates,omitempty"`
+
+	/* DatabaseNames: (SQL Server only) Clone only the specified databases from the
+	source instance. Clone all databases if empty. */
+	DatabaseNames []string `json:"databaseNames,omitempty"`
```
maqiuyujoyce, line 456: This can potentially be references to SQLDatabase resources, right?

maqiuyujoyce (commented):
Code LGTM except for the potential reference fields. I assume you'll add test cases for this feature, right?

apis/sql/v1beta1/sqlinstance_types.go
jasonvigil: Good catch. Thanks. Fixed!

apis/sql/v1beta1/sqlinstance_types.go
jasonvigil: Fixed! Thank you

apis/sql/v1beta1/sqlinstance_types.go
```diff
@@ -441,7 +441,36 @@ type InstanceValueFrom struct {
 	SecretKeyRef *v1alpha1.SecretKeyRef `json:"secretKeyRef,omitempty"`
 }
 
+type BinLogCoordinates struct {
+	// BinLogFileName: Name of the binary log file for a Cloud SQL instance.
```
maqiuyujoyce, line 445: The field description generated, right? The format is ... weird...

pkg/controller/direct/sql/mapping.go
```diff
@@ -23,13 +23,17 @@ import (
...
-func SQLInstanceKRMToGCP(in *krm.SQLInstance, refs *SQLInstanceInternalRefs) (*api.DatabaseInstance, error) {
+func SQLInstanceKRMToGCPInstance(in *krm.SQLInstance, refs *SQLInstanceInternalRefs) (*api.DatabaseInstance, error) {
 	out := &api.DatabaseInstance{}
 
 	if in == nil {
 		return nil, fmt.Errorf("cannot convert nil SQLInstance")
 	}
 
+	if in.Spec.CloneSource != nil {
+		return nil, fmt.Errorf("cannot convert SQLInstance with CloneSource specified")
```
maqiuyujoyce, line 34: ~Does it mean that while the field is supported in the CRD, the feature can't be used?~

Okay, so basically this conversion is only triggered by the insert function, and the insert function should never be called if `in.Spec.CloneSource != nil`... got it.

Maybe the error message can be improved in the caller (insertInstance() function) saying something like: `error inserting a Cloud SQL instance with no cloud source: %w`, or you can make the error message more verbose here; otherwise this is an implicit knowledge which is not intuitive to future contributors.

pkg/controller/direct/sql/mapping.go
```diff
@@ -793,10 +797,43 @@ func Convert_SQLInstance_API_v1_To_KRM_status(in *api.DatabaseInstance, out *krm
...
-	var defaultValue T
-	if v == defaultValue {
-		return nil
+func SQLInstanceKRMToGCPCloneRequest(in *krm.SQLInstance, refs *SQLInstanceInternalRefs) (*api.InstancesCloneRequest, error) {
+	if in == nil {
+		return nil, fmt.Errorf("cannot convert nil SQLInstance")
+	}
+
+	if in.Spec.CloneSource == nil {
+		return nil, fmt.Errorf("cannot convert nil CloneSource")
```
maqiuyujoyce, line 806: Similarly, I wondered if the error message can be more verbose and actionable. E.g. `cannot convert nil CloneSource, CloudSource is required`.

pkg/controller/direct/sql/normalize.go
```diff
@@ -290,6 +303,95 @@ func normalizeAuditLogBucketRef(ctx context.Context, kube client.Reader, obj *kr
...
+
+func normalizeSourceSQLDatabaseRefs(ctx context.Context, kube client.Reader, obj *krm.SQLInstance) ([]string, error) {
+	if obj.Spec.CloneSource == nil || obj.Spec.CloneSource.SQLDatabaseRefs == nil {
+		return nil, nil
+	}
+
+	sqlDatabaseRefs := obj.Spec.CloneSource.SQLDatabaseRefs
+	sqlDatabases := make([]string, len(sqlDatabaseRefs))
+
+	for i, sqlDatabaseRef := range sqlDatabaseRefs {
```
maqiuyujoyce, line 356: Not a blocker,  but I believe we also want to check and ensure the list contains all external values or name values, but not a mix of both.

pkg/webhook/immutable_fields_validator.go
```diff
@@ -380,6 +382,21 @@ func validateImmutableFieldsForLoggingLogMetricResource(oldSpec, spec map[string
 	return allowedResponse
 }
 
+func validateImmutableFieldsForSQLInstanceResource(oldSpec, spec map[string]interface{}) admission.Response {
+	ImmutableFields := []string{"cloneSource"}
```
maqiuyujoyce, line 386: It's out of scope, though we probably need to think about how to handle use cases when a user realizes they made a mistake and want to fix it in-place after the creation/clone failed. Making the field strictly immutable means a user needs to delete the object, edit the manifest, and reapply, and could be a bit troublesome for some GitOps workflow.

pkg/webhook/immutable_fields_validator.go
```diff
@@ -380,6 +382,21 @@ func validateImmutableFieldsForLoggingLogMetricResource(oldSpec, spec map[string
...
+	var res []string
+	for _, field := range ImmutableFields {
+		if isImmutableFieldModified(oldSpec, spec, field) {
```
maqiuyujoyce, line 389: I think we want to achieve the immutability check via CEL rules? Not a blocker though.

maqiuyujoyce (commented):
Overall lgtm with some nits and open questions, i.e. none of the comment is a blocker.

Have you got the chance to test the clone locally? Do you plan to support test coverage and mockgcp coverage in this PR?

pkg/controller/direct/sql/normalize.go
jasonvigil: Why? I think it should be ok to mix

pkg/webhook/immutable_fields_validator.go
jasonvigil: I think that we should require delete + recreate, rather than build in some special other mechanism, but maybe we can wait for product feedback on this.

pkg/controller/direct/sql/mapping.go
jasonvigil: This error would only happen if developer misuses the conversion function. Probably just adding a comment would be sufficient.

pkg/controller/direct/sql/mapping.go
jasonvigil: Same here, this error would only happen if developer misuses the function. I'll add a comment.

apis/sql/v1beta1/sqlinstance_types.go
jasonvigil: Field description was taken directly from the client. Will update it to match the other comments.

pkg/controller/direct/sql/normalize.go
maqiuyujoyce: @yuwenma proposed it and suggested it will help us avoid complicated edge cases. Yuwen, could you elaborate on it a bit?

jasonvigil: I think that it is not really complicated to implement this, and customer may want the functionality. There could be some databases managed by kcc, others managed externally, but customer might want to clone all of them.

pkg/webhook/immutable_fields_validator.go
jasonvigil: I think it's more idiomatic to require create + recreate. But maybe this could be partially solved by educating customer about k8s best practices. Or, perhaps there is some k8s-native pattern here the customer had suggested that I am unaware of.

apis/sql/v1beta1/sqlinstance_types.go
jasonvigil: fixed

pkg/controller/direct/sql/mapping.go
jasonvigil: fixed

pkg/controller/direct/sql/mapping.go
jasonvigil: fixed

pkg/webhook/immutable_fields_validator.go
jasonvigil: fixed

pkg/webhook/immutable_fields_validator.go
jasonvigil: fixed

jasonvigil:
> Overall lgtm with some nits and open questions, i.e. none of the comment is a blocker.
> 
> Have you got the chance to test the clone locally? Do you plan to support test coverage and mockgcp coverage in this PR?

Yes, tested locally. Added a basic mockgcp test.

pkg/controller/direct/sql/normalize.go
maqiuyujoyce: SG!

Personally, I like the flexibility of supporting a mix of both. Implementation-wise, it's fine to start with a stricter rule (don't allow the mix) and then loosen it (allow the mix) if we think this provides a better UX.

apis/sql/v1beta1/sqlinstance_types.go
```diff
@@ -441,7 +441,38 @@ type InstanceValueFrom struct {
...
+	// +optional
+	PointInTime *string `json:"pointInTime,omitempty"`
+
+	/* The source SQLInstance to clone */
+	SQLInstanceRef refsv1beta1.SQLInstanceRef `json:"sqlInstanceRef,omitempty"`
+}
+
 type SQLInstanceSpec struct {
+	/* Create this database as a clone of a source instance. Immutable. */
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="spec.cloneSource is immutable"
```
maqiuyujoyce, line 472: A silly question: Does the type of the field matter for the validation? E.g. is it able to understand "equals" for object type here?

pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/sqlinstanceclonebasic/create.yaml
```diff
@@ -0,0 +1,35 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: sql.cnrm.cloud.google.com/v1beta1
+kind: SQLInstance
+metadata:
+  name: postgres-clone-${uniqueId}
```
maqiuyujoyce, line 18: It's out of scope, so bringing it up just for discussion.

I personally like consistent naming for easier validation, maintenance, education (contributor may ask about how to name the resource), and potential automation. That's why we had this guide: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/archive/README.NewResourceFromTerraform.md#naming-conventions-special-variables-and-constants-in-testdata

Now we are moving to direct controllers. Do you find any preferred way of naming the test resources? Or do you find any other purposes (other than consistency, maintenance easiness, etc) more important when coming up with the names?

pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/sqlinstanceclonebasic/create.yaml
```diff
@@ -0,0 +1,35 @@
...
+  cloneSource:
+    sqlInstanceRef:
+      name: postgres-source-${uniqueId}
+  databaseVersion: POSTGRES_16
+  region: us-central1
+  settings:
+    backupConfiguration:
+      enabled: true
+      pointInTimeRecoveryEnabled: true
+    # Location preference is not actually a required field. However, setting it for tests
```
maqiuyujoyce, line 29: Nice!

maqiuyujoyce (commented):
/lgtm
/approve

/hold
I wondered if @yuwenma wants to take another look.

Besides, can we add a sample about it? https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/README.Samples.md

apis/sql/v1beta1/sqlinstance_types.go
jasonvigil: Yes, equality works for maps: https://github.com/google/cel-spec/blob/master/doc/langdef.md#equality. I also tested this manually.

pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/sqlinstanceclonebasic/create.yaml
jasonvigil: I feel like it doesn't really matter that much, as long as the test resource names are unique and meaningful / understandable at first glance. Much like variable names.

pkg/controller/direct/sql/normalize.go
jasonvigil: I think it's more complicated to add a check to ensure that all references in the list are either external or kcc-managed. That requires more code. However, am not sure about the scenario you mentioned where it could cause "complicated edge cases". Do you remember the details?

apis/sql/v1beta1/sqlinstance_types.go
justinsb: I am a little wary of introducing CEL (1) on a high profile resource and (2) accidentally i.e. as part of a wider PR.  (2) is in case we need to revert and so that we can focus on CEL.

pkg/controller/direct/sql/merge.go
```diff
@@ -626,7 +626,7 @@ func MergeDesiredSQLInstanceWithActual(desired *krm.SQLInstance, refs *SQLInstan
 	}
 
 	if desired.Spec.Settings.DiskAutoresize != nil {
-		if desired.Spec.Settings.DiskAutoresize != actual.Settings.StorageAutoResize {
+		if direct.ValueOf(desired.Spec.Settings.DiskAutoresize) != direct.ValueOf(actual.Settings.StorageAutoResize) {
 			// Change disk autoresize
 			updateRequired = true
```
justinsb, line 631: We might also be able to look at the updateTime...

pkg/controller/direct/sql/normalize.go
```diff
@@ -290,6 +303,95 @@ func normalizeAuditLogBucketRef(ctx context.Context, kube client.Reader, obj *kr
...
+				return nil, fmt.Errorf("error reading referenced SQLDatabase %v: %w", key, err)
+			}
+
+			sqlDatabaseName, err := refs.GetResourceID(sqlDatabase)
+			if err != nil {
+				return nil, err
+			}
+			sqlDatabases[i] = sqlDatabaseName
+		} else {
+			return nil, fmt.Errorf("must specify either spec.settings.cloneSource.sqlDatabaseRefs[].external or spec.settings.cloneSource.sqlDatabaseRefs[].name")
```
justinsb, line 388: FYI, k8s has fieldpath to help with this problem (writing annoyingly long field paths!)

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -138,7 +138,60 @@ func (a *sqlInstanceAdapter) Create(ctx context.Context, u *unstructured.Unstruc
...
+		Multiplier: 2,
+	}
+	for {
+		log.V(2).Info("polling", "op", op)
+
+		if op.Status == "DONE" {
+			break
+		}
+		if err := gax.Sleep(ctx, pollingBackoff.Pause()); err != nil {
+			return fmt.Errorf("wait SQLInstance %s clone failed: %w", a.desired.Name, err)
```
justinsb, line 171: Nit: I personally think it reads better to use the "ing" form

i.e.

`failed to reconcile: waiting for SQLInstance %s clone: ...`

vs

`failed to reconcile: wait SQLInstance %s clone: ...`

pkg/controller/direct/sql/normalize.go
yuwenma: Thanks @maqiuyujoyce for raising this. 
I put the rationale and some thoughts [here](https://docs.google.com/document/d/1gdbFRW0nfny8j7C_bUDuGG2Bk3z_WaC_eYhoLbuppG0/edit?tab=t.0#heading=h.g21v7n6bvl3z). The no-mix is a recommendation ("shall" rather than "must", see [RFC2119](https://datatracker.ietf.org/doc/html/rfc2119#section-3)) . And that it is for GCP unique list.

apis/sql/v1beta1/sqlinstance_types.go
yuwenma: I proposed using CEL for immutable check in the [CR Validation doc](https://docs.google.com/document/d/1nKB6hkMV34K0SVtOdFyGZ48D4NhKhGAOTGXPEVLNAqQ/edit?resourcekey=0-hbpIoeozx9it1GVTZDIbBw&tab=t.0#bookmark=id.6quka07soumj) and put the reasons on how the decision is made:

- Kubernetes supports CEL since 1.26 which is the oldest GKE cluster version. Using CEL shall not cause GKE cluster version issues, but we shall reevaluate this when the condition changes.
- Some KCC resources use webhook to validate immutable fields. We can continue using that, but for external contributors CEL is much easier for self-learning. 
- We only apply CEL on immutable fields. More complicated CEL validation requires further discussion.

Regarding to Justin's concerns, +1 to split the immutable check in a separate PR. Because 1. we do want to use CEL for immutable field check and makes it as the trends. 2. users may hit the CEL-supportability problem when they install KCC in a non-GKE cluster and the k8s is very old. Though the possibility of this case itself is low, SQL is a high profile resource like Justin mentioned. So splitting the PR is a good idea, we can still use CEL and keep us with the clean rollback options.

yuwenma:
/lgtm

Looks great! You may want to resolve the comments before merging.

pkg/controller/direct/sql/sqlinstance_controller.go
jasonvigil: Fixed, and for all the other SQLInstance error messages also.

pkg/controller/direct/sql/normalize.go
jasonvigil: Can you share a link to an example of this?

pkg/controller/direct/sql/merge.go
jasonvigil: It felt like this was a more reliable way to do it, but am planning to switch to use the mapper / generator code soon anyway.

apis/sql/v1beta1/sqlinstance_types.go
jasonvigil: Ok, moving this CEL validation to a separate PR.

pkg/controller/direct/sql/merge.go
jasonvigil: Though maybe the mapper generator code wouldn't really help here. Can you elaborate a bit on what you mean by using updateTime? Not sure I understand completely.

pkg/controller/direct/sql/normalize.go
jasonvigil: Just spoke w/ @justinsb about this. We are thinking it is overkill to use refs for these SQLDatabases anyway, since they are associated with the parent SQLInstanceRef. Decided to make this field just a list of strings of database names.

apis/sql/v1beta1/sqlinstance_types.go
jasonvigil: Actually we have gone back on this decision, and are thinking it's overkill to use refs here. Switched back to just a list of names. https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2429#discussion_r1729307605

justinsb (approved):

PR #2432, opened 2024-08-07
Use 'Absent' as the default state-into-spec behavior

maqiuyujoyce:
### Change description

Use 'Absent' as the default state-into-spec behavior.

Diff base is #2504.

This PR should NOT be merged before 1.122 is released.

### Tests you have done

Built locally and verified that `absent` is the default value for state-into-spec annotation when it is not set. And with the cluster-level and namespace-level override, the default value can be rolled back to `merge`.


yuwenma:
/approve

PR looks good. Can you fix the presubmit tests?

gemmahou:
/lgtm
/approve

Feel free to remove the hold tag!

pkg/controller/direct/compute/refs.go
```diff
@@ -154,9 +154,12 @@ func ResolveComputeAddress(ctx context.Context, reader client.Reader, src client
 
 	// targetField: address
 	// See compute servicemappings for details
-	address, _, err := unstructured.NestedString(computeAddress.Object, "spec", "address")
+	// Because `spec.address` field is optional, we can't guarantee it always
+	// exists in a successfully reconciled ComputeAddress CR, so we should use
+	// the `status.observedState.address` instead.
+	address, _, err := unstructured.NestedString(computeAddress.Object, "status", "observedState", "address")
```
gemmahou, line 160: Clarification question: What if the user specifies an address? Will the `address` field be under `spec` then? Shall we check both `spec` and `status. observedState` for optional fields?

maqiuyujoyce: `status.observedState.address` should be the same as the specified `spec.address` if ComputeAddress reconciles successfully, so I won't worry about the discrepancy.

maqiuyujoyce (commented):
/hold cancel

PR #2438, opened 2024-08-07
feat: add kmskeyring ref

acpana:
Realized I am not going to use the `KmsKeyRef` for `DataformRepository` as it's not part of their proto definition just yet. So parking my work here for now.


apis/refs/v1beta1/helper.go
```diff
@@ -30,3 +30,14 @@ func GetResourceID(u *unstructured.Unstructured) (string, error) {
 	}
 	return resourceID, nil
 }
+
+func GetLocation(u *unstructured.Unstructured) (string, error) {
```
gemmahou, line 34: We don't want to make this a helper function, see: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2337#discussion_r1696037535

apis/refs/v1beta1/kmskeyref.go
gemmahou, line 1: I wanted to group refs and their resolver functions in one file by service... i.e. my compute refs: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/refs/v1beta1/computerefs.go

But I think your approach is clear, I'm open to suggestions.

apis/refs/v1beta1/kmskeyref.go
```diff
@@ -14,11 +14,97 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
 type KMSCryptoKeyRef struct {
-	/* The KMSCryptoKeyRef selfLink, when not managed by KCC. */
+	//  If provided must be in the format `projects/[kms_project_id]/locations/[region]/keyRings/[key_ring_id]/cryptoKeys/[key]`.
```
gemmahou, line 30: I've been thinking about this, I feel checks on External is unnecessary. KCC does not validate External format for now, and for example ComputeAddressRef, its External field accepts ID, full URL, or an address. I do not think KCC has a fixed pattern for External field, and should let GCP to fail the request if the value if invalid. I'll have a PR to discuss this, before that let's process with the current approach.

apis/refs/v1beta1/kmskeyref.go
```diff
@@ -14,11 +14,97 @@
...
 	External string `json:"external,omitempty"`
-	/* The `name` field of a `KMSCryptoKeyRef` resource. */
+	// The `name` field of a `KMSCryptoKey` resource.
```
gemmahou, line 32: Good catch!!

apis/refs/v1beta1/helper.go
acpana: But it's going to be used a fair amount I think for any ref that has `locations` in its fully qualified name

PR #2442, opened 2024-08-08
[For discussion]refactor: Updates on ref resolvers

gemmahou:
### Change description

Currently, the KCC controller (TF/DCL) does not validate external values of refs, and the format of externals can vary. Therefore, it is difficult to have a fixed pattern for external values. We rely on the GCP API to determine the valid format and to fail the request if it is invalid.

External is used for non-KCC managed resource, so we should let underlying GCP to manage/validate it. 
I'm more concerned about the Beta to direct migration here; strict checks on External might introduce breaking changes to our users.

- Removed the check on fixed External format.
- Added "targetField" so that we are able to control the field of refs we want to reference if its KCC managed resource. 

### Tests you have done

### DIFF
--- a/apis/refs/v1beta1/computerefs.go
+++ b/apis/refs/v1beta1/computerefs.go
@@ -17,7 +17,6 @@ package v1beta1
 import (
 	"context"
 	"fmt"
-	"strings"
 
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -28,7 +27,7 @@ import (
 )
 
 type ComputeNetworkRef struct {
-	/* The compute network selflink of form "projects/<project>/global/networks/<network>", when not managed by Config Connector. */
+	/* An external value of a `ComputeNetwork` resource, when not managed by KCC. */
 	External string `json:"external,omitempty"`
 	/* The `name` field of a `ComputeNetwork` resource. */
 	Name string `json:"name,omitempty"`
@@ -36,16 +35,7 @@ type ComputeNetworkRef struct {
 	Namespace string `json:"namespace,omitempty"`
 }
 
-type ComputeNetwork struct {
-	Project          string
-	ComputeNetworkID string
-}
-
-func (c *ComputeNetwork) String() string {
-	return fmt.Sprintf("projects/%s/global/networks/%s", c.Project, c.ComputeNetworkID)
-}
-
-func ResolveComputeNetwork(ctx context.Context, reader client.Reader, src client.Object, ref *ComputeNetworkRef) (*ComputeNetwork, error) {
+func ResolveComputeNetwork(ctx context.Context, reader client.Reader, src client.Object, ref *ComputeNetworkRef, targetField string) (*ComputeNetworkRef, error) {
 	if ref == nil {
 		return nil, nil
 	}
@@ -54,14 +44,7 @@ func ResolveComputeNetwork(ctx context.Context, reader client.Reader, src client
 		if ref.Name != "" {
 			return nil, fmt.Errorf("cannot specify both name and external on computenetwork reference")
 		}
-
-		tokens := strings.Split(ref.External, "/")
-		if len(tokens) == 5 && tokens[0] == "projects" && tokens[2] == "global" && tokens[3] == "networks" {
-			return &ComputeNetwork{
-				Project:          tokens[1],
-				ComputeNetworkID: tokens[4]}, nil
-		}
-		return nil, fmt.Errorf("format of computenetwork external=%q was not known (use projects/<projectId>/global/networks/<networkid>)", ref.External)
+		return ref, nil
 	}
 
 	if ref.Name == "" {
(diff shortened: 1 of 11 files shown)


apis/refs/v1beta1/computerefs.go
```diff
@@ -27,24 +26,15 @@ import (
 )
 
 type ComputeNetworkRef struct {
-	/* The compute network selflink of form "projects/<project>/global/networks/<network>", when not managed by KCC. */
+	/* An external value of a `ComputeNetwork` resource, when not managed by KCC. */
```
acpana, line 29: I may not be the most up to date with the pain points here but I somewhat like the self documenting aspect of `projects/<project>/global/networks/<network>`. As a user, I see it in the CRD and know what the expected format is.

Regarding validation, I think about how most validation that GCP makes we probably want to shift left so that we don't make extra API calls. But I'll let others chime in here if it doesn't apply.

gemmahou: Updating the comment because I feel External can be any format today, there should be no "expected" format for external. But we can have a "suggested" format for user, based on our understanding about the API.

justinsb:
I think what we're doing today in refs is more correct.  How did this arise - what problem are you trying to solve?

gemmahou:
> I think what we're doing today in refs is more correct. How did this arise - what problem are you trying to solve?

Having a strict format for refs.External would cause breaking changes to Beta resources when migrating to Scifi. If I understand correctly, External can be an ID, name, URL, address, email, etc. We do not currently have validations on the External field, and for some unsupported refs, we suggest that users to use External to specify their preferred value instead.

apis/refs/v1beta1/computerefs.go
```diff
@@ -147,47 +130,139 @@ func ResolveComputeSubnetwork(ctx context.Context, reader client.Reader, src cli
...
+
+	computeAddress := &unstructured.Unstructured{}
+	computeAddress.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "compute.cnrm.cloud.google.com",
+		Version: "v1beta1",
+		Kind:    "ComputeAddress",
+	})
+	if err := reader.Get(ctx, key, computeAddress); err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, fmt.Errorf("referenced ComputeAddress %v not found", key)
```
maqiuyujoyce, line 217: I think we should return NewReferenceNotFoundError error here.

More context: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2393#discussion_r1699462125

apis/refs/v1beta1/computerefs.go
```diff
@@ -147,47 +130,139 @@ func ResolveComputeSubnetwork(ctx context.Context, reader client.Reader, src cli
...
+	if err != nil {
+		return nil, err
+	}
+
+	projectID, err := ResolveProjectID(ctx, reader, computeAddress)
+	if err != nil {
+		return nil, err
+	}
+
+	region, _, _ := unstructured.NestedString(computeAddress.Object, "spec", "region")
```
maqiuyujoyce, line 232: I may not have the full context, but I thought ComputeAddress CR doesn't have `region` field: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_computeaddresses.compute.cnrm.cloud.google.com.yaml?

apis/refs/v1beta1/computerefs.go
```diff
@@ -76,44 +59,54 @@ func ResolveComputeNetwork(ctx context.Context, reader client.Reader, src client
...
+	projectID, err := ResolveProjectID(ctx, reader, computeNetwork)
 	if err != nil {
 		return nil, err
 	}
-	return &ComputeNetwork{
-		Project:          computeNetworkProjectID,
-		ComputeNetworkID: computenetworkID,
-	}, nil
+
+	// Need to determine good targetField values, some common values: selfLink, name, id, email, or specific field name etc
```
justinsb, line 85: Can we make this methods instead e.g. SelfLink() and ID().  Then the compiler is working with us.

apis/refs/v1beta1/computerefs.go
```diff
@@ -54,14 +44,7 @@ func ResolveComputeNetwork(ctx context.Context, reader client.Reader, src client
...
 			return nil, fmt.Errorf("cannot specify both name and external on computenetwork reference")
 		}
-
-		tokens := strings.Split(ref.External, "/")
-		if len(tokens) == 5 && tokens[0] == "projects" && tokens[2] == "global" && tokens[3] == "networks" {
-			return &ComputeNetwork{
-				Project:          tokens[1],
-				ComputeNetworkID: tokens[4]}, nil
-		}
-		return nil, fmt.Errorf("format of computenetwork external=%q was not known (use projects/<projectId>/global/networks/<networkid>)", ref.External)
```
justinsb, line 64: Let's just change this to a warning.  We don't currently have a suitable context (like MapContext), so maybe just a klog or similar.

apis/refs/v1beta1/computerefs.go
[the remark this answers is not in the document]
justinsb: The idea is that the user should always be able to use `projects/<project>/global/networks/<network>`.  Our controllers will convert to the correct format - full URL or just the name etc.

I think @gemmahou is proposing in this PR that we allow users to specify other values, and when they do we pass them through to the underlying APIs as-is, so that if users used a different format with an existing resource, we don't break them.  That makes me think, maybe we should have a "strict" mode of parsing, and use strict for new resources?  cc @yuwenma

apis/refs/v1beta1/computerefs.go
justinsb: Good call!

apis/refs/v1beta1/computerefs.go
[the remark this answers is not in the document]
maqiuyujoyce: There are rare cases when it is very difficult to achieve this consistent interface. In [this reference field for a ComputeNetwork](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/v1.121.0/config/servicemappings/servicedirectory.yaml#L53), the only accepted format is `projects/<projectNumber>/locations/global/networks/<networkID>`. Parsing project ID to project number requires specific permission that the KCC GSA may not have. So I'm inclined to keep accepting any form of the external value.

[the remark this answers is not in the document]
yuwenma: +1 to Justin. We shall have a strict formatting rule for new Direct resources. 

For existing Beta resources, we have to do special handlings when migrating to Direct controller, but it should be isolated within that resource code base. Regarding the example @maqiuyujoyce mentioned, the direct resource shall not use the shared `computeRef`, but defines its own one which still accepting projectNumber for backward compatibility reasons.

gemmahou:
I'm closing this PR as we have decided to use the current implementation for direct controllers. Regarding the Beta resource migration and other edge cases, I will add separate resolver functions for different use cases. Comments in this PR will be addressed there.

apis/refs/v1beta1/computerefs.go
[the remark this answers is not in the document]
maqiuyujoyce: I think keeping a consistent format for most cases but still allow edge cases works. Regarding the example I mentioned above, I agree that we should support it separately, though I believe the purpose is not only for backwards compatibility, but is to avoid unnecessary permission issues. 

Here are the edge cases I think worth the effort to specialize:
1. Forming the actual value passed to the API requires an additional API call.
2. The value API accepts is not resource ID/self link/resource name related (e.g. the actual address managed by a ComputeAddress resource).

Though we want to add enough instructions for contributors to be able to make a decision themselves during the implementation.

@gemmahou I saw you closed the PR. Let me know if you want me to follow up somewhere else.

PR #2443, opened 2024-08-08
enh: towards DataformRepository beta

acpana:
Includes

* api changes as per the published proto
* refs additions
* test updates


apis/dataform/v1alpha1/repository_types.go
```diff
@@ -68,9 +94,26 @@ type DataformRepositorySpec struct {
 	// +optional
 	ResourceID *string `json:"resourceID,omitempty"`
 
+	// Repository user labels. An object containing a list of "key": value pairs. Example: { "name": "wrench", "mass": "1.3kg", "count": "3" }.
```
justinsb, line 97: I don't think we should do these yet, today in KCC labels come from metadata.labels on the object.  I think what you're doing here is correct, it's just a bigger discussion!

apis/refs/v1beta1/secretref.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	ref = &SecretManagerSecretRef{
+		External: fmt.Sprintf("projects/%s/secrets/%s", secretProjectID, secretResourceID),
+	}
+
+	return &SecretManagerSecret{Ref: ref, ResourceID: secretResourceID}, nil
+}
+
+func ResolveSecretIDForObject(ctx context.Context, reader client.Reader, obj *unstructured.Unstructured) (string, error) {
```
justinsb, line 107: Is this the same logic as  ResolveSecretManagerSecretRef, just with unstructured?

If so, can you use runtime.DefaultUnstructuredConverter to map unstructured -> structured?

apis/refs/v1beta1/secretversiontref.go
```diff
@@ -0,0 +1,105 @@
+// Copyright 2024 Google LLC
```
justinsb, line 1: Nit: there's a typo in this filename

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_dataformrepositories.dataform.cnrm.cloud.google.com.yaml
```diff
@@ -59,29 +59,134 @@ spec:
...
+                    description: Authentication fields for remote uris using SSH protocol.
+                    properties:
+                      hostPublicKey:
+                        description: Content of a public SSH key to verify an identity
+                          of a remote Git host.
+                        type: string
+                      userPrivateKeySecretVersionRef:
+                        description: The name of the Secret Manager secret version
+                          to use as a ssh private key for Git operations. Must be
+                          in the format projects/*/secrets/*/versions/* .
```
justinsb, line 113: Nit: we should probably remove the "Must be in the format" here from the docstring, not a blocker for alpha though  (and we might want a API linter in future!)

mockgcp/mockdataform/repository.go
```diff
@@ -101,7 +104,6 @@ func (r *RepositoryV1Beta1) UpdateRepository(ctx context.Context, request *pb.Up
 		return nil, err
 	}
 
-	fmt.Println("todo acpana did it")
```
justinsb, line 104: lol :-)

pkg/controller/direct/dataform/repository_controller.go
```diff
@@ -191,16 +190,43 @@ func (a *Adapter) Update(ctx context.Context, u *unstructured.Unstructured) erro
 	updateMask := &fieldmaskpb.FieldMask{}
 
 	if a.desired.Spec.GitRemoteSettings != nil {
-		if !reflect.DeepEqual(a.desired.Spec.GitRemoteSettings, a.actual.GitRemoteSettings) {
+		mapCtx := &direct.MapContext{}
+		protoDesired := RepositoryGitRemoteSettings_ToProto(mapCtx, a.desired.Spec.GitRemoteSettings)
```
justinsb, line 194: I would expect that  DataformRepositorySpec_ToProto would have done this already (we might well need to override  DataformRepositorySpec_ToProto to make this call)

pkg/controller/direct/dataform/repository_controller.go
```diff
@@ -191,16 +190,43 @@ func (a *Adapter) Update(ctx context.Context, u *unstructured.Unstructured) erro
...
+		if mapCtx.Err() != nil {
+			return fmt.Errorf("converting GitRemoteSettings to api: %w", mapCtx.Err())
+		}
+
+		if !reflect.DeepEqual(protoDesired, a.actual.GitRemoteSettings) {
```
justinsb, line 199: This bit (the reflect.DeepEqual) does look good though.  We might also be able to use `proto.Equal`

pkg/controller/direct/dataform/repository_mappings.go
```diff
@@ -28,43 +29,99 @@ func DataformRepositorySpec_FromProto(mapCtx *direct.MapContext, in *pb.Reposito
 	out := &krm.DataformRepositorySpec{}
 	out.GitRemoteSettings = RepositoryGitRemoteSettings_FromProto(mapCtx, in.GetGitRemoteSettings())
 	out.WorkspaceCompilationOverrides = RepositoryWorkspaceCompilationOverrides_FromProto(mapCtx, in.GetWorkspaceCompilationOverrides())
+	out.DisplayName = direct.LazyPtr(in.GetDisplayName())
+	out.GitRemoteSettings = RepositoryGitRemoteSettings_FromProto(mapCtx, in.GetGitRemoteSettings())
```
justinsb, line 33: Yeah, it looks like you're already calling  RepositoryGitRemoteSettings_FromProto so I'm confused why you need to call it again!

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_dataformrepositories.dataform.cnrm.cloud.google.com.yaml
acpana: As in we shouldn't share the expected format in the doc?

pkg/controller/direct/dataform/repository_controller.go
acpana: Chatting offline -- I think we have to convert the individual pieces to proto since `a.desired` is KRM based here. I was following along the cbwp as it was the latest (at the time?). I see in monitoringdashboards both desired and actual are proto-based 🤔

pkg/controller/direct/dataform/repository_mappings.go
acpana: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2443#discussion_r1712386454

apis/refs/v1beta1/secretref.go
acpana: Reworked with your suggestion 🙏🏼

apis/refs/v1beta1/helper.go
```diff
@@ -30,3 +30,14 @@ func GetResourceID(u *unstructured.Unstructured) (string, error) {
 	}
 	return resourceID, nil
 }
+
+func GetLocation(u *unstructured.Unstructured) (string, error) {
```
justinsb, line 34: Nice!

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_dataformrepositories.dataform.cnrm.cloud.google.com.yaml
justinsb: We should, but it goes on `external`, not on the parent ref

justinsb (approved):

PR #2446, opened 2024-08-08
feat: Add basic/full test suite for storageNotification

yuwenma:
The `_generated_object_storagenotification.golden.yaml` and `http.log` is generated from the following commands:
- `hack/record-gcp fixtures/storagenotificationbase`
- `hack/record-gcp fixtures/storagenotificationfull`


pkg/test/constants/presubmitconstants_test.go
```diff
@@ -89,7 +89,7 @@ func TestValidTestName(t *testing.T) {
 					return nil
 				}
 				// Test cases are the names of the leaf-node directory, only match on the lowest subdirectory name
-				if strings.HasSuffix(filepath.Dir(path), tc) {
```
justinsb, line 92: I think this is saying that we need to change the representative test to be `storagenotificationbase`

I hope we can simplify all this stuff in future, and just use `tests/e2e`.  I just don't think we know if there is anything we're testing in the "other" tests that we are not testing in `tests/e2e`

justinsb:
One blocker about the representative test name, but other lgtm

pkg/test/constants/presubmitconstants_test.go
yuwenma: I think this requires the test suite to be called `basestoragenotification`. 
I changed the path check mostly for 2 reasons:
1. The previous error message is confusing: `Test case storagenotification does not exist or is misspelled` when the test suit is called `storagenotificationbase`. Using `contains` seems to better fit the error. 
2. I see why we want a check, but I don't quite get why the `Kind` has to be the suffix.  I think as long as the `Kind` shows up in the path, we are good. This also makes it easier for contributors to use shorter and more descriptive names.

justinsb (approved):

PR #2447, opened 2024-08-08
feat: add mockgcp for storageNotification

yuwenma:
based from #2446

- add MockGCP for StorageNotification (v1beta1)
- Fixed MockGCP for `StorageBucket` and `PubSubTopic` to fit real GCP outcome.


justinsb:
Code looks good, and I think we should merge these on green (assuming it's roughly the same as the hack/record-gcp)!

/approve
/lgtm

pkg/test/resourcefixture/testdata/basic/pubsub/v1beta1/pubsubtopic/_generated_object_pubsubtopic.golden.yaml
```diff
@@ -16,6 +16,30 @@ metadata:
   name: pubsubtopic-sample-${uniqueId}
   namespace: ${uniqueId}
 spec:
```
yuwenma, line 18: This golden object change is due to the state-into-spec:merge in pubsubtopic. I would expect to see they are removed once we change the annotation's default to `absent`

PR #2448, opened 2024-08-08
fix: wrong flag reference in generate-type CLI

yuwenma:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/dev/tools/controllerbuilder/pkg/options/generateoptions.go
+++ b/dev/tools/controllerbuilder/pkg/options/generateoptions.go
@@ -27,6 +27,6 @@ func (o *GenerateOptions) InitDefaults() {
 
 func (o *GenerateOptions) BindPersistentFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVar(&o.ProtoSourcePath, "proto-source-path", o.ProtoSourcePath, "path to (compiled) proto for APIs")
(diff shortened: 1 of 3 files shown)


dev/tools/controllerbuilder/pkg/options/generateoptions.go
```diff
@@ -27,6 +27,6 @@ func (o *GenerateOptions) InitDefaults() {
 
 func (o *GenerateOptions) BindPersistentFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVar(&o.ProtoSourcePath, "proto-source-path", o.ProtoSourcePath, "path to (compiled) proto for APIs")
-	cmd.PersistentFlags().StringVarP(&o.APIVersion, "version", "v", o.APIVersion, "the KRM API version. used to import the KRM API")
+	cmd.PersistentFlags().StringVarP(&o.APIVersion, "apiVersion", "v", o.APIVersion, "the KRM API version. used to import the KRM API")
```
justinsb, line 30: Nit: this probably should be `api-version`, to be consistent with `proto-source-path`.

Also ... do we need to change generate.sh?

yuwenma: Good point! Done

justinsb (approved):

PR #2449, opened 2024-08-08
chore: change the _type.go template to not skip target resource when --kind is specified

yuwenma:
Problem:
When we export the generated type to `./apis`, the _types.go generation no longer works (triggered by `--kind` flag). 
Because the _type.go file template contains the same `kcc:proto` tag as the _type.generated.go file suppose to use. This skips the actual proto-to-go generation of that resource.

### DIFF
--- a/dev/tools/controllerbuilder/generate.sh
+++ b/dev/tools/controllerbuilder/generate.sh
@@ -30,7 +30,8 @@ go run . generate-types \
     --service google.cloud.securesourcemanager.v1 \
     --api-version securesourcemanager.cnrm.cloud.google.com/v1alpha1 \
     --output-api ${APIS_DIR} \
-    --kinds SecureSourceManagerInstance
+    --kind SecureSourceManagerInstance \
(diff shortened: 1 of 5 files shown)


dev/tools/controllerbuilder/template/apis/types.go
```diff
@@ -36,13 +36,12 @@ var {{ .Kind }}GVK = GroupVersion.WithKind("{{ .Kind }}")
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
 // {{ .Kind }}Spec defines the desired state of {{ .Kind }}
-{{- if .KindProtoTag }}
-// +kcc:proto={{ .KindProtoTag }}
```
justinsb, line 40: We want this tag, don't we?  Otherwise we won't generate a mapper...

yuwenma: resolved per offline discussion

justinsb (approved):

PR #2452, opened 2024-08-08
feat:add a label for kcc resource type

xiaoweim:
This PR adds a default label to KCC resource object specifying the resource type.

I would like to have a label in concord usage tables that helps me attribute the resource usage to the exact KCC resource. In order to concord to be able to get this information, we first need a label in KCC so that I can later add a label in concord with this information.


yuwenma:
Do we want to add the Kind name in the KCC metadata labels as well? This is a duplicate of the ` Kind` field itself, my concern is that this could bug some users.

justinsb:
Yeah, can you describe more what we're trying to achieve here.  AIUI the GCP resource implies the KCC resource, so I'm not sure what we're adding?  We have the `managed-by-cnrm` GCP label already if we want to know if the resource is managed by KCC

xiaoweim:
> Yeah, can you describe more what we're trying to achieve here. AIUI the GCP resource implies the KCC resource, so I'm not sure what we're adding? We have the `managed-by-cnrm` GCP label already if we want to know if the resource is managed by KCC

Essentially I would like to have a label in concord usage tables that helps me attribute the resource usage to the exact KCC resource. In order to concord to be able to get this information, we first need a label in KCC so that I can later add a label in concord with this information.

xiaoweim:
> Do we want to add the Kind name in the KCC metadata labels as well? This is a duplicate of the ` Kind` field itself, my concern is that this could bug some users.

So I am adding this label to pass this information to concord usage tables, not sure if KCC metadata label is needed or can do the same?

xiaoweim:
retest/

cheftako:
Seems like some types (Eg CloudSQLInstance) are not being labelled today.

cheftako:
Otherwise looks good.

PR #2456, opened 2024-08-08
tests: create project for periodic e2e tests

justinsb:


dev/tasks/cleanup-test-resources
```diff
@@ -0,0 +1,28 @@
...
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+cd ${REPO_ROOT}
+
+GCP_PROJECT_ID=$(gcloud config get project)
```
yuwenma, line 23: The `kcctest-` convention could make this harder to maintain or left some unreacable projects.
I like that we use labels below when creating project, wondering if we can filter by label here and only delete the matching project.

yuwenma (commented):
/lgtm
/approve

/hold no blocker. Feel free to unhold or add filters in the cleanup.

dev/tasks/cleanup-test-resources
justinsb: We don't want to delete a project from another test...

I included the date so that we can relatively easily delete old projects.  If we want to we can add a label with the timestamp, but I think we should OSS the cleanup logic first.  (And we should think about whether we're rebuilding boskos and should just use that)

justinsb:
/hold cancel

Thanks for reviewing @yuwenma  ... I'll look into what changes we want to make to better support periodic cleanup when we create that job!

PR #2460, opened 2024-08-09
chore: update golangci-lint-action

justinsb:
We were seeing some failures from it without corresponding error
messages, try updating.


yuwenma (commented):
/lgtm
/approve

PR #2461, opened 2024-08-09
Update GCP dependencies

justinsb:
- **chore: update-gcp-dependencies is go.work aware**
- **codebot: update GCP dependencies**


mockgcp/dev/images/protoc/Dockerfile
yuwenma, line 1: May I ask what this file would be used for?

justinsb: Good call, I will revert ... I think the script added some things from my experiments!

PR #2463, opened 2024-08-09
mockgcp: more BigQueryTable fidelity

justinsb:


yuwenma (commented):
/lgtm
/approve

Just to double check I read the code (and assumptions) correctly, I assume the `spec.schema` was originally from the `state-into-spec:merge`, because it only shows in the HTTP response.

PR #2465, opened 2024-08-09
chore: make e2e project creation use GOOGLE_APPLICATION_CREDENTIALS

justinsb:


dev/tasks/create-test-project
```diff
@@ -30,10 +30,19 @@ if [[ -z "${BILLING_ACCOUNT_ID:-}" ]]; then
   exit 1
 fi
 
+if [[ -n "${GOOGLE_APPLICATION_CREDENTIALS:-} "]]; then
+  echo "Activating service-account in gcloud (GOOGLE_APPLICATION_CREDENTIALS=${GOOGLE_APPLICATION_CREDENTIALS})"
+  gcloud auth activate-service-account --key-file=${GOOGLE_APPLICATION_CREDENTIALS}
+fi
```
yuwenma, line 36: Assume the SA will run configconnector and enable some services, then
 shall we do something like this 
`gcloud beta billing accounts add-iam-policy-binding "${BILLING_ACCOUNT}" --member "serviceAccount:${SA_EMAIL}" --role "roles/billing.admin"

yuwenma (commented):
/lgtm
/approve

dev/tasks/create-test-project
justinsb: I think we only have `roles/billing.user` on this account, so we can't add permissions to the billing account.

I think we should try running the tests under this account, hopefully that will happen anyway but ... we'll see.  One problem, then the next one, etc  😂

yuwenma: I see. Thanks!

PR #2468, opened 2024-08-10
tests: use GCP_PROJECT_ID (if set) as default project

justinsb:
Otherwise we fall back to the credentials and gcloud default project.

This allows us to override the project in our e2e tests.


yuwenma (commented):
/lgtm
/approve

PR #2483, opened 2024-08-12
feat: Implement "Export" for SQLInstance

jasonvigil:
This fixes some of the failing SQL Instance dynamic tests caused by nil pointer dereferences of nil objects returned by the (previously) empty export implementation.

### DIFF
--- a/pkg/controller/direct/sql/sqlinstance_controller.go
+++ b/pkg/controller/direct/sql/sqlinstance_controller.go
@@ -122,14 +122,14 @@ func (a *sqlInstanceAdapter) Find(ctx context.Context) (bool, error) {
 	a.actual = instance
 
 	log := klog.FromContext(ctx).WithName(ctrlName)
-	log.V(2).Info("found cloudsql instance", "actual", a.actual)
+	log.V(2).Info("found SQLInstance", "actual", a.actual)
 
 	return true, nil
 }
 
 func (a *sqlInstanceAdapter) Create(ctx context.Context, u *unstructured.Unstructured) error {
 	log := klog.FromContext(ctx).WithName(ctrlName)
-	log.V(2).Info("creating instance", "desired", a.desired)
+	log.V(2).Info("creating SQLInstance", "desired", a.desired)
 
 	if a.projectID == "" {
 		return fmt.Errorf("project is empty")
@@ -216,7 +216,7 @@ func (a *sqlInstanceAdapter) Create(ctx context.Context, u *unstructured.Unstruc
 
 func (a *sqlInstanceAdapter) Update(ctx context.Context, u *unstructured.Unstructured) error {
 	log := klog.FromContext(ctx)
-	log.V(2).Info("updating instance", "desired", a.desired)
+	log.V(2).Info("updating SQLInstance", "desired", a.desired)
 
 	// First, handle database version updates
 	if a.desired.Spec.DatabaseVersion != nil && *a.desired.Spec.DatabaseVersion != a.actual.DatabaseVersion {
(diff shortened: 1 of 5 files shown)


pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -326,7 +326,27 @@ func (a *sqlInstanceAdapter) Delete(ctx context.Context) (bool, error) {
...
 
 func (a *sqlInstanceAdapter) Export(ctx context.Context) (*unstructured.Unstructured, error) {
-	return nil, nil
+	if a.actual == nil {
+		return nil, fmt.Errorf("SQLInstance %q not found", a.resourceID)
+	}
+
+	sqlInstance, err := SQLInstanceGCPToKRM(a.actual)
+	if err != nil {
+		return nil, fmt.Errorf("error converting SQL Instance from API %w", err)
```
maqiuyujoyce, line 335: Nit: Usually I'd prefer we use `SQLInstance` (the kind name) as a special term and don't separate it using white spaces as `SQL Instance` can be a much broader term. If it is to represent the GCP resource, I might use the GCP brand name (e.g. `Cloud SQL instance`) instead. But we don't have an agreed error message convention, so it's up to you to decide how to describe it.

maqiuyujoyce (commented):
Overall lgtm. Could you add the export golden file like this: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2251/files#diff-07863e8584e6714d3ddf662e9258afc7a5e20d531dea410aafcecfa478baee3f?

yuwenma (commented):
/lgtm
/approve
/hold  +1 to @maqiuyujoyce comment. But this defers to you.

jasonvigil:
> Overall lgtm. Could you add the export golden file like this: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2251/files#diff-07863e8584e6714d3ddf662e9258afc7a5e20d531dea410aafcecfa478baee3f?

@maqiuyujoyce @yuwenma I think this is blocked on https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2251.. Any plans to merge it soon?

maqiuyujoyce:
If it is blocked on a draft PR, I think it's fine to merge this one for now. I assume you run the dynamic test locally and it passed?

/lgtm
/approve

Feel free to unhold.

jasonvigil:
@maqiuyujoyce I re-ran the tests manually, and noticed that it was still failing sometimes, due to the `testDriftCorrection` wait time of 10s not being long enough for SQLInstance resources. Added another commit to fix this issue. Can you please re-review?

pkg/controller/dynamic/dynamic_controller_integration_test.go
```diff
@@ -467,7 +467,7 @@ func testDriftCorrection(ctx context.Context, t *testing.T, testContext testrunn
 	}
 	// Underlying APIs may not have strongly-consistent reads due to caching. Sleep before attempting a re-reconcile, to
 	// give the underlying system some time to propagate the deletion info.
-	time.Sleep(time.Second * 10)
+	time.Sleep(time.Second * 60)
```
maqiuyujoyce, line 470: As there are hundreds of test cases we need to run in a test suite, sleeping for one minute for each test case will be expensive. Could you customize it for SQLInstance only?

jasonvigil: The tests run in batches of 20 (in parallel). Given 379 tests, this would maybe add ~20min of extra runtime. So, not really that bad, but it would be nice to minimize test time if possible, so sure, will make this change

jasonvigil: Ok, updated. PTAL @maqiuyujoyce. I noticed some of the other SQLInstance tests were previously skipping the drift detection test altogether. I think this was unnecessary, we could have just increased the wait time. So, I updated those tests to not skip drift detection, and increased their wait time. Assuming other tests have the same problem, I made the wait time customizable for every test, rather than add some special behavior to the drift detection test runner just for SQLInstance.

maqiuyujoyce (commented):
/lgtm
/approve

PR #2485, opened 2024-08-13
Update ContainerCluster and ContainerNodePool CRDs

marko7460:
### Change description

This change makes necessary changes that allow modifying taints on containernodepool object inp lace instead of destroying and recreating the object

This change will:
* Update resource_container_node_pool code only with taints
* Update ContainerCluster and ContainerNodePool CRDs
* Update integration tests for ContainerNodePool

Caveat:
* Taints can be added and modified
* Taints can't be deleted
   * To delete the a taint you can delete it via `gcloud` command

Fixes #

### Tests you have done
* Ran `make ready-pr`
* Modified existing `ContainerNodePool`

## Tests
* Ran `go test -v -tags=integration ./pkg/controller/dynamic/ -test.run TestCreateNoChangeUpdateDelete -run-tests containernodepool -timeout 900s`. All tests passed
* Ran `make test`. All tests passed ```DONE 7185 tests in 596.678s```


config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_containerclusters.container.cnrm.cloud.google.com.yaml
```diff
@@ -532,10 +578,9 @@ spec:
               enableAutopilot:
                 description: Immutable. Enable Autopilot for this cluster.
                 type: boolean
-              enableBinaryAuthorization:
```
maqiuyujoyce, line 535: Removing a field is a breaking change for a Config Connector CRD and right now we don't accept such breaking changes. This field is removed in Terraform because they had a major version bump, and this doesn't apply to Config Connector.

To unblock the in-place update of the taint, I suggest instead of copying over the entire go files, you copy the code logic only relevant to the taint feature to the Terraform library under third_party/ folder.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_containerclusters.container.cnrm.cloud.google.com.yaml
```diff
@@ -193,6 +193,48 @@ spec:
                     required:
                     - disabled
                     type: object
+                  rayOperatorConfig:
```
maqiuyujoyce, line 196: It's nice that a lot more fields are supported in this CRD, though we do need field coverage for spec in testing especially for big changes like this.

Considering that there are already breaking changes, I think you'll probably want to only focus on the updatable taint fields and not to add other new fields.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_containerclusters.container.cnrm.cloud.google.com.yaml
```diff
@@ -787,51 +851,8 @@ spec:
                     description: Base64 encoded public certificate that is the root
                       of trust for the cluster.
                     type: string
-                  password:
```
maqiuyujoyce, line 790: This also seem to be a breaking change to me.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_containernodepools.container.cnrm.cloud.google.com.yaml
```diff
@@ -271,6 +273,16 @@ spec:
                   enablePrivateNodes:
                     description: Whether nodes have internal IP addresses only.
                     type: boolean
+                  networkPerformanceConfig:
```
maqiuyujoyce, line 276: Similarly, it's nice to add many new features to ContainerNodePool, but test coverage will be needed. Considering the size of the change, I think you might want to focus the taint feature.

maqiuyujoyce (commented):
Hi @marko7460 , thank you so much for contribution!

Consider that the new version of Terraform code contains breaking changes and it also adds many new fields that should be tested well before merge, I suggest you manually pick the logic relevant to the taint in-place update feature and update the Terraform library with those logic.

You'll want to ensure all the presubmit tests pass and it'll be great if you cover the update of the taints in the test cases following the guide here: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/archive/README.NewResourceFromTerraform.md#test-your-changes

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_containernodepools.container.cnrm.cloud.google.com.yaml
marko7460: @maqiuyujoyce Thank you for your feedback. You are absolutely right. I will focus only on the `taint` feature. I will revert changes to the most of the `go` files I took from the terraform provider code and focus on `taints`.

pkg/test/resourcefixture/testdata/basic/container/v1beta1/containernodepool/_generated_object_containernodepool.golden.yaml
```diff
@@ -29,6 +29,10 @@ spec:
   nodeConfig:
     diskSizeGb: 100
     diskType: pd-balanced
+    effectiveTaints:
```
justinsb, line 32: It looks  like maybe this (1) doesn't update and (2) might be a terraform thing, not a GKE thing.   Should we include it?

marko7460: I think you are right. I will test without this. This is the terraform specific field that is read only.

PR #2486, opened 2024-08-13
Support optional 'spec' field as targetField in resource references

### LINKED ISSUE #2467: Resource reference stops working if the targetField is optional and the referenced resource has `state-into-spec: absent`
### Checklist

As we are going to default the `state-into-spec` annotation to `absent` (see [announcement](https://github.com/GoogleCloudPlatform/k8s-config-connector/releases/tag/v1.121.0)), optional field defaulted/generated by the API won't be populated back to `spec`. So a resource becomes non-referenceable when the targetField is optional but defaulted/generated.

To ensure the resource references still work, we should support those optional fields under `status.observedState` and resolve the reference differently.

### Additional information

_No response_

### Importance

_No response_

maqiuyujoyce:
### Change description

Fixes #2467

There are edge cases where a resource is referenced via its optional `spec` field. After `state-into-spec: absent` becomes the default behavior, the optional field won't be populated if it is not configured explicitly by the user, and the reference can't be resolved.

To ensure the resource reference can still be resolved correctly, we should:
1. add the particular `spec` fields under `status.observedState` so that they can still be retrievable, and
2. check the observed state when resolving the resource references.

The two edge cases are:
* AlloyDBCluster is referenced via optional field `spec.clusterType` in AlloyDBInstance: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/v1.121.0/config/servicemappings/alloydb.yaml#L179
* ComputeAddress is referenced via optional field `spec.address` in many other kinds, e.g. https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/v1.121.0/config/servicemappings/compute.yaml#L508
  * `status.observedState.address` is already supported: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/v1.121.0/crds/compute_v1beta1_computeaddress.yaml#L256

### Tests you have done

Verified that after setting `state-into-spec: absent`, servicedirectoryendpoint test failed without this change, but succeeded with this change.


justinsb:
Looks good; do we want to also do ComputeAddress in this PR?

maqiuyujoyce:
> Looks good; do we want to also do ComputeAddress in this PR?

ComputeAddress already has `status.observedState.address`. Will update the PR description.

PR #2487, opened 2024-08-13
WIP: feat: CRD & Mapper for workstation workstation, cluster and config

justinsb:
Because our mapper generator (currently) is package scoped, it proved to be easier to generate these 3 together (because of shared resources with an unhandled rpc.Status).


apis/workstations/v1alpha1/workstationsworkstation_types.go
```diff
@@ -0,0 +1,138 @@
...
+	ExternalRef *string `json:"externalRef,omitempty"`
+
+	// ObservedState is the state of the resource as most recently observed in GCP.
+	// +optional
+	ObservedState *WorkstationsWorkstationObservedState `json:"observedState,omitempty"`
+}
+
+// WorkstationsWorkstationSpec defines the desired state of WorkstationsWorkstation
+// +kcc:proto=google.cloud.workstations.v1.Workstation
+type WorkstationsWorkstationObservedState struct {
```
yuwenma, line 70: I saw the recent Direct resources all have the GCP `observedState` fields added in `.status`. This may be something worth discussing. I see the value of adding those fields (and maybe more than not adding them), though 
my concerns are: 
1. We may then need to update the KRM object more frequently without a clear reason. i.e. some output-only fields are changed, like `UpdateTime`. 
2. It makes the "snapshotting" harder. Snapshot refers to the ConfigConnector's approach to hash previous state changes, and determine whether to update/reconcile the GCP object or not, like etag.
3. Some observedState field may be too large, like the MonitoringDashboard that can hit size issues more quickly.

apis/workstations/v1alpha1/workstationsworkstationcluster_types.go
```diff
@@ -0,0 +1,164 @@
...
+	// // Optional.
+	// //  [Labels](https://cloud.google.com/workstations/docs/label-resources) that
+	// //  are applied to the workstation cluster and that are also propagated to the
+	// //  underlying Compute Engine resources.
+	// Labels map[string]string `json:"labels,omitempty"`
+
+	// Immutable. Name of the Compute Engine network in which instances associated
+	//  with this workstation cluster will be created.
+	// +required
+	Network *string `json:"network,omitempty"`
```
yuwenma, line 58: Shall this be a resourceRef to `ComputeNetwork`?

apis/workstations/v1alpha1/workstationsworkstationcluster_types.go
```diff
@@ -0,0 +1,164 @@
...
+
+	// Immutable. Name of the Compute Engine subnetwork in which instances
+	//  associated with this workstation cluster will be created. Must be part of
+	//  the subnetwork specified for this workstation cluster.
+	// +required
+	Subnetwork *string `json:"subnetwork,omitempty"`
```
yuwenma, line 64: ditto: how about using  `ComputeSubnetworkRef`? Not a blocker

yuwenma (commented):
/approve

This PR looks really great! Maybe adding a mockGCP so we can verify the outcome?)

ldanielmadariaga:

PR #2488, opened 2024-08-13
tests: don't fail on changes in the http log during e2e

justinsb:
Because we don't want to treat a new upstream field as an error here.

### DIFF
--- a/pkg/test/utils.go
+++ b/pkg/test/utils.go
@@ -96,7 +96,19 @@ func CompareGoldenFile(t *testing.T, p string, got string, normalizers ...func(s
 	}
 
 	if diff := cmp.Diff(want, got); diff != "" {
-		t.Errorf("unexpected diff in %s: %s", p, diff)
+		onlyWarn := false
+		for _, f := range strings.Split(os.Getenv("ONLY_WARN_ON_GOLDEN_DIFFS"), ",") {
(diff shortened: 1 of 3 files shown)


pkg/test/utils.go
yuwenma, line 1: I can see that we may want to bypass the real HTTP headers, and maybe some dynamic field like timestamp(which mostly have been handled in normalize.go), I'm a little bit concerned about passing the entire http golden log. 

Mind sharing more on which http part that you want to skip?

justinsb: So this is only in the e2e tests against real GCP.  We want to capture the logs, but the problem is that if there's an undocumented/new field being returned by GCP, we don't necessarily want to fail our e2e.

I do see the point though, it would be nice if we could get a failure signal here.  I think the problem is that some of the fields are not yet in proto, but I can give it a go....

justinsb:
Rebased because of conflicting merge.

Also added a comment about why we need this - because otherwise adding a field in the upstream API breaks our tests.  Maybe we can get there one day, but it's a big reach right now.

/assign @cheftako

pkg/test/utils.go
yuwenma: I see. That makes sense. Looks like the ONLY_WARN_ON_GOLDEN_DIFFS on-and-off can be one of the new field detector.

PR #2489, opened 2024-08-13
tests: capture our golden output after e2e

justinsb:


justinsb:
The copying and pasting is not ideal, but I propose we get it working first then de-dup

yuwenma (commented):
/lgtm
/approve

Thank you

PR #2491, opened 2024-08-13
More github actions for Compositions

barney-s:
### Change description

Add more github actions for experiments/compositions:

- verify gomod
- verify go imports

### DIFF
--- a/.github/workflows/presubmit-compositions.yaml
+++ b/.github/workflows/presubmit-compositions.yaml
@@ -23,7 +23,42 @@ on:
     branches: ["master"]
     paths:
       - "experiments/compositions/**"
+
+env:
+  GOWORK: off
 jobs:
+  verify-goimports:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.22'
+          stable: true
+      - uses: actions/checkout@v4
+      - name: verify goimports
+        run: |
+          ./experiments/compositions/scripts/github-actions/verify-goimports.sh
+  verify-gomod:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.22'
+          stable: true
+      - uses: actions/checkout@v4
+      - run: |
+          ./experiments/compositions/scripts/github-actions/format-gomod.sh
+      - run: |
+          changes=$(git status --porcelain)
+          if [[ -n "${changes}" ]]; then
+            echo "::error Changes detected from dev/format-gomod:"
+            echo "::error (You may need to run go clean -cache -modcache)"
+            git diff | head -n60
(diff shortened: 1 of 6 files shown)


.github/workflows/presubmit-compositions.yaml
```diff
@@ -23,7 +23,48 @@ on:
     branches: ["master"]
     paths:
       - "experiments/compositions/**"
+
+env:
+  GOWORK: off
 jobs:
+  verify-goimports:
```
acpana, line 30: could we rely on the golanci-lint uberlinter to run `goimports`? Can be done as a follow on

.github/workflows/presubmit-compositions.yaml
```diff
@@ -23,7 +23,48 @@ on:
...
+    steps:
+      - name: Set up go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.22'
+          stable: true
+      - uses: actions/checkout@v4
+      - name: verify goimports
+        run: |
+          files=$(go run golang.org/x/tools/cmd/goimports -format-only -l experiments/compositions/.)
```
acpana, line 41: It may be easier to reproduce any issues w this cmd if we throw these into a script file

acpana:
/lgtm
/approve

2 little call outs; can probably add them as a follow ons if you agree

/hold if you want to address them here.

.github/workflows/presubmit-compositions.yaml
barney-s: Will do. We have one more CL in progress: 
https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2493

The top level linter is not catching anything in compositions.

barney-s:
> /lgtm /approve
> 
> 2 little call outs; can probably add them as a follow ons if you agree
> 
> /hold if you want to address them here.

Addressed. Please LGTM again.

.github/workflows/presubmit-compositions.yaml
```diff
@@ -23,7 +23,42 @@ on:
...
+      - name: Set up go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.22'
+          stable: true
+      - uses: actions/checkout@v4
+      - run: |
+          ./experiments/compositions/scripts/github-actions/format-gomod.sh
+      - run: |
+          changes=$(git status --porcelain)
```
justinsb, line 54: Tip: I personally prefer to have the GitHub actions be 1:1 with scripts - I think if I was doing it today I would use `dev/ci/presubmits` and auto-generate the GitHub actions.  The advantage of doing so is that then it's easy to run the same thing as GitHub actions locally ... just run the script!

justinsb (approved):

.github/workflows/presubmit-compositions.yaml
barney-s: happy to change it in a follow up.
Q: Is there an example i can look at. ?

PR #2494, opened 2024-08-14
tests: log resource name for dependency failure

acpana:
Some simple log lines to help me to more easily narrow down which resources are failing to create dependencies.

### DIFF
--- a/pkg/test/controller/reconciler/testreconciler.go
+++ b/pkg/test/controller/reconciler/testreconciler.go
@@ -142,12 +142,12 @@ func (r *TestReconciler) ReconcileObjectMeta(ctx context.Context, om metav1.Obje
 }
 
 // Creates and reconciles all unstructureds in the unstruct list. Returns a cleanup function that should be defered immediately after calling this function.
-func (r *TestReconciler) CreateAndReconcile(ctx context.Context, unstructs []*unstructured.Unstructured, cleanupPolicy ResourceCleanupPolicy) func() {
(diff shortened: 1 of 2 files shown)


pkg/test/controller/reconciler/testreconciler.go
```diff
@@ -142,12 +142,12 @@ func (r *TestReconciler) ReconcileObjectMeta(ctx context.Context, om metav1.Obje
...
+func (r *TestReconciler) CreateAndReconcile(ctx context.Context, dependencies []*unstructured.Unstructured, cleanupPolicy ResourceCleanupPolicy, mainResource *unstructured.Unstructured) func() {
 	r.t.Helper()
-	cleanupFuncs := make([]func(), 0, len(unstructs))
-	for _, u := range unstructs {
+	cleanupFuncs := make([]func(), 0, len(dependencies))
+	for _, u := range dependencies {
 		if err := r.mgr.GetClient().Create(ctx, u); err != nil {
-			r.t.Fatalf("error creating resource '%v': %v", u.GetKind(), err)
+			r.t.Fatalf("error creating dependecy '%v' for resource '%v/%v': %v", u.GetKind(), mainResource.GetName(), mainResource.GetKind(), err)
 		}
```
justinsb, line 151: No real objection, but don't we know from the test name?

justinsb (approved):

pkg/test/controller/reconciler/testreconciler.go
acpana: My experience with the post submit logs was that sometimes they are out of order. Another use case for this (at least for my debugging) is grep-ing against "error" or some other string and gleaning as much information as possible.

PR #2500, opened 2024-08-14
chore: don't use "direct" package from mockgcp

justinsb:
It would be nice, but it does introduce a circular dependency that
go.mod can't figure out.


yuwenma (commented):
/lgtm
/approve

Ouch! Thank you for the fix!

PR #2501, opened 2024-08-14
kubectl-plan: try to pretty-print diffs

justinsb:
- **feat: include more diff details in kubectl-plan output**
- **tests: golden-test "pretty" output from kubectl-plan also**
- **[kubectl-plan] tests: pre-create namespace, trim dynamic fields**


experiments/kubectl-plan/ssh-auth-sock.socket
yuwenma, line 1: remove?

experiments/kubectl-plan/pkg/plan/testdata/simpleupdate/_plan.yaml
```diff
@@ -4,6 +4,19 @@ spec:
   actions:
   - action: ApplyChanges
     apiVersion: v1
+    diff:
```
yuwenma, line 7: Would like to see how we do diff for slice and map

yuwenma (commented):
/lgtm
/approve
/hold for ssh socket
Really cool!

experiments/kubectl-plan/ssh-auth-sock.socket
justinsb: Good call, thank you!

PR #2508, opened 2024-08-14
add commands to build binaries for controllers

xiaoweim:
Add unified simple commands to build binaries for KCC controllers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -239,3 +239,47 @@ ready-pr: lint manifests resource-docs generate-go-client
 upgrade-dcl:
 	go get github.com/GoogleCloudPlatform/declarative-resource-client-library
 	make ensure
+
+# Build all binaries
+.PHONY: all-binary
+all-binary: config-connector-bin deletiondefender manager-bin recorder unmanageddetector webhook gke-addon-poststart operator-manager-bin
+
+# Build config-connector binary from cmd/config-connector
+.PHONY: config-connector-bin
+config-connector-bin:
+	go build -o bin/config-connector github.com/GoogleCloudPlatform/k8s-config-connector/cmd/config-connector
+
+# Build deletiondefender binary from cmd/deletiondefender
+.PHONY: deletiondefender
+deletiondefender:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -239,3 +239,40 @@ ready-pr: lint manifests resource-docs generate-go-client
 upgrade-dcl:
 	go get github.com/GoogleCloudPlatform/declarative-resource-client-library
 	make ensure
+
+# Build all binaries
+all-binary: cmd-config-connector deletiondefender cmd-manager recorder unmanageddetector webhook gke-addon-poststart operator-cmd-manager
+.PHONY: all-binary cmd-config-connector deletiondefender cmd-manager recorder unmanageddetector webhook gke-addon-poststart operator-cmd-manager
```
cheftako, line 245: Generally better to leave 1 .PHONY directive per target, rather than grouping them this way. Makes it easier to keep the targets clean and clean.

Makefile
```diff
@@ -239,3 +239,40 @@ ready-pr: lint manifests resource-docs generate-go-client
...
+
+# Build cmd/config-connector binary
```
cheftako, line 247: Seems like this is actually building the bin/config-connector binary

Makefile
```diff
@@ -239,3 +239,40 @@ ready-pr: lint manifests resource-docs generate-go-client
...
+cmd-config-connector:
```
cheftako, line 248: Also worth mention that .PHONY is used to indicate that the target is not a "real" target, i.e. it does not correspond to an actual file you are trying to make. However in this case you are trying to make bin/config-connector. So you could change the target to bin/config-connector and then it would not need to be a phony target. The problem then is that running 'make bin/config-connector' would do nothing if bin/config-connector already exists and is at least as old as its dependencies. Its probably fine for now but if put the effort into tracking its dependencies we could get an optimization.

justinsb: This is true, but go at least claims to be at least as good at tracking dependencies as make (or bazel).  So `go build` should be fast if dependencies are unchanged.

cheftako: Fair and I did end with its fine for now :P

justinsb:
/approve
/lgtm

Thanks @xiaoweim

PR #2510, opened 2024-08-14
fixit: Run presubmit fixtures tests by services

gemmahou:
### Change description

Split out compute & gkehub

### Tests you have done

### DIFF
--- a/tests/e2e/unified_test.go
+++ b/tests/e2e/unified_test.go
@@ -31,6 +31,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/util/slice"
+	"k8s.io/klog/v2"
+
 	"github.com/GoogleCloudPlatform/k8s-config-connector/config/tests/samples/create"
 	opcorev1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/apis/core/v1beta1"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test"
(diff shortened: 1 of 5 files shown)


justinsb:
We could also try creating N presubmit tests, one for each service ...

Do we know why the time has built up?  Is it just that each test takes 30 seconds, and we now have 120 tests (for example?).  Or are there one or two slow ones?

Even splitting out one or two key services might be interesting.  We could also have an env var to change the test names e.g. to include the group and kind in the path, so e.g.

`fixtures/<group>/<kind>/<name>` instead of `fixtures/<name>`

The advantage of doing that on an env var is that we wouldn't break our existing tests.  But maybe none of those existing tests care about TestAllInSeries anyway.

gemmahou:
> We could also try creating N presubmit tests, one for each service ...

I wanted to chat about this as well. Seems like the http recorder and log checker do not work well if multiple tests run in one job. I could create multiple jobs for each services like you did before. 

> Do we know why the time has built up? Is it just that each test takes 30 seconds, and we now have 120 tests (for example?). Or are there one or two slow ones?

I guess GH job has setup and cleanup as well(?) that takes time. Also as we are adding more tests, we'll face the issue eventually. 

> Even splitting out one or two key services might be interesting. We could also have an env var to change the test names e.g. to include the group and kind in the path, so e.g.
> 
> `fixtures/<group>/<kind>/<name>` instead of `fixtures/<name>`
> 
> The advantage of doing that on an env var is that we wouldn't break our existing tests. But maybe none of those existing tests care about TestAllInSeries anyway.

Splitting out one or two key services makes sense. If we run each services in a separate job, I'm afraid there will be a long list of jobs in GH. I'm also thinking about if there's a way we can run the tests somewhere else(like prow) and use a bot to post the result to GH... I remember TF team does something similar but that might need some extra GH and internal settings.

tests/e2e/unified_test.go
```diff
@@ -146,8 +149,21 @@ func testFixturesInSeries(ctx context.Context, t *testing.T, testPause bool, can
 		fixtures := resourcefixture.Load(t)
 		for _, fixture := range fixtures {
 			fixture := fixture
+			group := fixture.GVK.Group
```
justinsb, line 152: Nice!

justinsb (approved):

PR #2514, opened 2024-08-15
tests: Always capture golden output in e2e tests

justinsb:


yuwenma:
/lgtm
/approve

looks good! (unfortunately the presubmit fails)

PR #2518, opened 2024-08-15
fix: rediscluster multiple field updates

justinsb:
- **Multi field update test case for redis cluster**
- **redis: handle multi-field updates**
- **tests: capture golden output for redis multi-field update**


yuwenma (commented):
/lgtm
/approve

Thanks!

PR #2519, opened 2024-08-16
Implement Export capability for AlloyDB

jasonvigil:
### Change description

Also, added support for user labels in the direct controller.

Had to fix some of the test cases so they pass when running with direct controller + realgcp:
- Remove `displayName` field (it is currently broken / not working in the GCP API).
- Remove `maintenanceUpdatePolicy` field (it is not implemented in direct controller yet).


anhdle-sso (approved):
/LGTM

PR #2529, opened 2024-08-17
mockgcp: support for PubSubSubscription

justinsb:
- **tests: capture golden output for PubSub**
- **mockgcp: support for PubSubSubscription**

### DIFF
--- a/mockgcp/mockpubsub/topic.go
+++ b/mockgcp/mockpubsub/topic.go
@@ -48,35 +48,40 @@ func (s *publisherService) CreateTopic(ctx context.Context, req *pb.Topic) (*pb.
 	if err := s.storage.Create(ctx, fqn, obj); err != nil {
 		return nil, err
 	}
+
 	return obj, nil
 }
 
 func (s *publisherService) populateDefaultsForTopic(obj *pb.Topic) {
-	obj.MessageStoragePolicy = &pb.MessageStoragePolicy{
-		AllowedPersistenceRegions: []string{
-			"asia-east1",
-			"asia-northeast1",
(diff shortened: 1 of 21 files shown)


mockgcp/mockpubsub/topic.go
```diff
@@ -48,35 +48,40 @@ func (s *publisherService) CreateTopic(ctx context.Context, req *pb.Topic) (*pb.
...
-			"us-east5",
-			"us-east7",
-			"us-south1",
-			"us-west1",
-			"us-west2",
-			"us-west3",
-			"us-west4",
-			"us-west8",
-		},
+	// TODO: When _is_ this populated?
```
yuwenma, line 56: Good question. I don't know,  but at least in PubSubTopic creation. (I verified against real GCP in the mock storage notification #2447)

pkg/test/resourcefixture/testdata/basic/storage/v1beta1/storagenotification/storagenotificationbase/_http.log
yuwenma, line 1: The real GCP populates the `messageStoragePolicy` in these 2 test suites for storage notification. (The pubSubTopic is a dependency which calls the create method )

yuwenma (commented):
/lgtm

you may want to update the populate `MessageStorageePolicy` for PubSubTopic create (at least).

justinsb:
I still haven't found any GCP APIs that are returning the message storage policy... might be an API change.  We should detect it with our e2e tests anyway...

/hold cancel

PR #2530, opened 2024-08-19
kompanion: refactor to avoid global variables

justinsb:
Also only create directories if we're writing a file, and a few general refactorings for cobra-style.


yuwenma (commented):
/lgtm
/approve

Thanks!

PR #2537, opened 2024-08-20
chore: add GCP token source to controller options

justinsb:
- **chore: refactor out options building for REST and GRPC**
- **chore: add GCP token source to controller options**


yuwenma:
/lgtm
/approve

Thank you! Really like it! And I can update the controller template to use the new format.

PR #2538, opened 2024-08-20
fix: don't raise errors in SupportsIAM

justinsb:
If we haven't registered the direct controller here, log a warning
rather than treating this as an error.


yuwenma (commented):
/lgtm
/approve

Agree! Thanks

PR #2541, opened 2024-08-20
feat: use much higher QPS if not configured

justinsb:


experiments/kompanion/cmd/export/export.go
```diff
@@ -212,6 +212,12 @@ func RunExport(ctx context.Context, opts *ExportOptions) error {
 		return fmt.Errorf("error building kubeconfig: %w", err)
 	}
 
+	// We rely more on server-side rate limiting now, so give it a high client-side QPS
+	if config.QPS == 0 {
```
yuwenma, line 216: What is the default value of the client-side QPS (I thought it's 30 but I could be wrong, which means this condition is never hit)? 
Then how about `if config.QPS < 100`?

yuwenma (commented):
/lgtm
/approve

/hold 
One question for clarification, feel free to unhold.

experiments/kompanion/cmd/export/export.go
justinsb: So if it's not configured, then QPS==0, but that is then interpreted as 30 qps.

If  we want to turn off rate limiting entirely, we set QPS==-1 (any value < 0 works).  That is what is used e.g. internally in the apiserver when we want to avoid all client-side rate limiting and just rely on server-side rate limiting.

I'm fine with .QPS < 100, but (1) we should make it `qps >= 0 && qps < 100` to avoid the -1 case, and (2) the advantage of doing `== 0` is that then users can still configure it explicitly to a value < 100 if they want to.  I can't imagine anyone will though TBH, but ... they could.

For now, I'll unhold if that's OK and we can iterate on this

yuwenma: Thank you! This makes a lot of sense.

PR #2542, opened 2024-08-20
kompanion: iterate over kinds by default

justinsb:


experiments/kompanion/cmd/export/export.go
```diff
@@ -280,25 +304,46 @@ func RunExport(ctx context.Context, opts *ExportOptions) error {
 		return fmt.Errorf("could not create %q directory: %w", reportTempDir, err)
 	}
 
+	shouldExcludeObject := func(id types.NamespacedName) bool {
+		if shouldExclude(id.Namespace, opts.ignoreNamespaces, opts.targetNamespaces) {
+			return true
+		}
+		return shouldExclude(id.Name, opts.ignoreObjects, opts.targetObjects)
+	}
```
yuwenma, line 312: The logic is kind of twisted. I'm wondering if we can check GKNN (as a whole struct)  instead.

yuwenma (commented):
/lgtm
/approve 

One minor regarding the complexity of the `shouldExclude` logic.

experiments/kompanion/cmd/export/export.go
```diff
@@ -280,25 +304,46 @@ func RunExport(ctx context.Context, opts *ExportOptions) error {
...
+				Resources:           []schema.GroupVersionResource{resource},
+				DynamicClient:       dynamicClient,
+				ReportDir:           reportTempDir,
+				ShouldExcludeObject: shouldExcludeObject,
+			})
+		}
 	}
 
-	log.Printf("Starting worker threads to process namespaces")
+	log.Printf("Starting worker threads to process KCC objects")
```
yuwenma, line 346: super nit: KCC --> Config Connector

experiments/kompanion/cmd/export/export.go
justinsb: Yes, the challenge is that we also want to check the namespace individually.  But yes, I think we can move in this direction!

PR #2543, opened 2024-08-20
fix: use KUBECONFIG if specified

justinsb:
Switching to calling into the kube config loader directly here.


yuwenma (commented):
/lgtm
/approve

PR #2545, opened 2024-08-20
direct: ComputeForwardingRule refs

gemmahou:
### Change description

Blocks #2364 

This is a another PR about resolve ComputeForwardingRule refs...😅

### Tests you have done

### DIFF
--- a/apis/compute/v1beta1/computeforwardingrule_types.go
+++ b/apis/compute/v1beta1/computeforwardingrule_types.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v1beta1
 
 import (
+	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/clients/generated/apis/k8s/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -50,7 +51,7 @@ type ForwardingruleFilterLabels struct {
 
 type ForwardingruleIpAddress struct {
 	// +optional
-	AddressRef *v1alpha1.ResourceRef `json:"addressRef,omitempty"`
+	AddressRef *refs.ComputeAddressRef `json:"addressRef,omitempty"`
 
 	// +optional
 	Ip *string `json:"ip,omitempty"`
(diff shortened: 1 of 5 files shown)


pkg/controller/direct/compute/refs.go
```diff
@@ -0,0 +1,243 @@
...
+func resolveResource(ctx context.Context, reader client.Reader, src client.Object, ref *v1alpha1.ResourceRef, gvk schema.GroupVersionKind) (*unstructured.Unstructured, error) {
+	if ref == nil {
+		return nil, nil
+	}
+
+	if ref.External != "" {
+		if ref.Name != "" {
+			return nil, fmt.Errorf("cannot specify both name and external on reference")
+		}
+		return nil, nil
```
justinsb, line 39: How do we handle the External case?

pkg/controller/direct/compute/refs.go
```diff
@@ -0,0 +1,243 @@
...
+		return nil, err
+	}
+
+	// targetField: self_link
+	// See compute servicemappings for details
+	return &v1alpha1.ResourceRef{
+		External: computeTargetTCPProxy.GetSelfLink()}, nil
+}
+
+func ResolveComputeTargetVPNGateway(ctx context.Context, reader client.Reader, src client.Object, ref *v1alpha1.ResourceRef) (*v1alpha1.ResourceRef, error) {
```
justinsb, line 229: A problem here is that previously we could have used the type of the ref to call the correct Resolver function (by making it a member function).

apis/compute/v1beta1/computeforwardingrule_types.go
```diff
@@ -88,25 +87,25 @@ type ForwardingruleServiceDirectoryRegistrations struct {
 
 type ForwardingruleTarget struct {
 	// +optional
-	ServiceAttachmentRef *refs.ComputeServiceAttachmentRef `json:"serviceAttachmentRef,omitempty"`
+	ServiceAttachmentRef *v1alpha1.ResourceRef `json:"serviceAttachmentRef,omitempty"`
```
justinsb, line 90: We do lose the ability to have nice per-ref-type comments now, IIUC

justinsb:
I'm not sure what problem we're trying to solve here.  I think we want to update the comments on External to say "should be of the form projects/<foo>....", instead of "must be of the form projects/<foo>..." ?

apis/compute/v1beta1/computeforwardingrule_types.go
gemmahou: I do not think that's required for existing Beta resources(?), ideality we should not introduce any new changes to Beta CRDs?

pkg/controller/direct/compute/refs.go
gemmahou: Updated: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2545/commits/a5741936fc53397e04d5b5c4fb4fba02d7478931

For External case it should return `ref, nil`

gemmahou:
> I'm not sure what problem we're trying to solve here. I think we want to update the comments on External to say "should be of the form projects/....", instead of "must be of the form projects/..." ?

Ok I changed back to use the v1beta1 refs... but I haven't think about a good comments on External. 
Originally the description is: 'Allowed value: The `selfLink` field of a `ComputeNetwork` resource.'
New description is : 'The compute network selflink of form "projects/<project>/global/networks/<network>", when not managed by Config Connector.'
This is very confusing, "projects/<project>/global/networks/<network>" is not a selfLink(selfLink should be the full URL format), and the format restriction is optional for some resources.

Also it's hard to reuse the existing resolver(and skip strict format checks), because the output are different as well(id vs selfLink vs address, depends on the API), it's easier and clearer to write separate resolver functions for Beta resources like ComputeForwardingRule. We can think about how to merge them in the future...

pkg/controller/direct/compute/refs.go
gemmahou: I changed it back

pkg/controller/direct/compute/refs.go
```diff
@@ -0,0 +1,475 @@
...
+	}
+
+	// targetField: address
+	// See compute servicemappings for details
+	address, _, _ := unstructured.NestedString(computeAddress.Object, "spec", "address")
+	if address == "" {
+		return nil, fmt.Errorf("cannot get address for referenced %s %v (spec.address is empty)", computeAddress.GetKind(), computeAddress.GetNamespace())
+	}
+	return &refs.ComputeAddressRef{
+		External: address}, nil
```
justinsb, line 149: Will be interesting to see if this is the "primary" external ref format .... TBD!

pkg/controller/direct/compute/refs.go
```diff
@@ -0,0 +1,475 @@
...
+	}
+
+	// targetField: self_link
+	// See compute servicemappings for details
+	return &refs.ComputeTargetHTTPSProxyRef{
+		External: computeTargetHTTPSProxy.GetSelfLink()}, nil
+}
+
+func ResolveComputeTargetSSLProxy(ctx context.Context, reader client.Reader, src client.Object, ref *refs.ComputeTargetSSLProxyRef) (*refs.ComputeTargetSSLProxyRef, error) {
+	if ref == nil {
```
justinsb, line 348: As we build more of these, we should look at whether it's possible to have more helper functions for the common cases (similar to what you've done with resolveResourceName) e.g.

```
func ResolveComputeTargetSSLProxy(ctx context.Context, reader client.Reader, src client.Object, ref *refs.ComputeTargetSSLProxyRef) (*refs.ComputeTargetSSLProxyRef, error) {
  external, err := resolveNormalRef(ctx, reader, src, ref.Name, ref.Namespace, ref.External, schema.GroupVersionKind{...})
  if err != nil { ... }
  return &refs.ComputeTargetSSLProxyRef{ External: external }
}
````

File: pkg/controller/direct/compute/refs.go [hunk 6]
```diff
@@ -0,0 +1,475 @@
...
+		Kind:    "ComputeTargetVPNGateway",
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// targetField: self_link
+	// See compute servicemappings for details
+	return &refs.ComputeTargetVPNGatewayRef{
+		External: computeTargetVPNGateway.GetSelfLink()}, nil
```
justinsb, line 461: Does GetSelfLink actually work?

justinsb (approved):

pkg/controller/direct/compute/refs.go
gemmahou: Hmmm probably not... will need to use `unstructured.NestedString(obj.Object, "status", "selfLink")`, I'll verify and update in my direct controller PR!

PR #2547, opened 2024-08-21
Fix -skip-tests logic in samples test

maqiuyujoyce:
### Change description

Fix -skip-tests logic in samples test so that when it is unset, no sample will be skipped.

### Tests you have done

Verified that a single test can be triggered properly with `-run-tests=custom-budget` but without `-skip-tests`.


gemmahou:
/lgtm
/approve

Thanks!!

PR #2548, opened 2024-08-21
[DO NOT MERGE] Update BillingBudgetsBudget test cases to make the resource more identifiable

maqiuyujoyce:
### Change description

Specified `spec.displayName` for test BillingBudgetsBudget resources and the start year for the custom budget to make them more identifiable when looking into the underlying GCP resources.

Right now we keep leaking the GCP budgets likely due to some tests. I verified that the test resources can be cleaned up properly via running the dynamic test locally, so it's weird that it got leaked constantly. As I couldn't find any logs for this API, and there is no identifiable information on the leaked budgets, it's hard to know where they come from. Adding the displayName may help the investigation.

### Tests you have done


gemmahou:
Seems like the vcr test need to be updated. However seems like this requires quota project to be set, I'm unsure how difficult to update it locally. If it's just some naming changes, we can manually update the log, or exclude the vcr test for this resource for now like the alloydb one([example](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/0fe1146aed9ad203701eac057e9e961fe66bc8c0/config/tests/samples/create/harness.go#L768)). I will collect the log via prow and copy it over to GH repo.

ldanielmadariaga:

PR #2551, opened 2024-08-21
Disallow duplicate composition definitions for the same GVK

hankfreund:
### Change description
When the controller reconciles a new Composition, ensure we haven't already seen one for the same GVK. If a match is found with a different name or namespace, set the error status on the resource and don't process it further.

### Tests you have done
New e2e test for this scenario.


experiments/compositions/composition/tests/testcases/simple_test.go
```diff
@@ -458,7 +458,7 @@ func TestUpdateCompositionModifyStage(t *testing.T) {
 
 func TestCustomStatusCELRule(t *testing.T) {
 	s := scenario.NewBasic(t)
-	//defer s.Cleanup()
+	defer s.Cleanup()
```
barney-s, line 461: Thankyou for fixing this.

experiments/compositions/composition/tests/scenario/scenario.go
```diff
@@ -153,11 +153,11 @@ func NewFromSample(t *testing.T, sample Sample, dependentSamples []Sample, hasCl
 }
 
 func (s *Scenario) Cleanup() {
+	defer cluster.ReleaseCluster(s.T, s.cluster)
+	defer s.GatherLogs()
```
barney-s, line 157: Assuming GatherLogs would be called first.

hankfreund: Yes, first gather, then release.

barney-s:
Will send a separate CL for linter fix. 
https://github.com/golangci/golangci-lint-action/issues/297#issuecomment-896821555

Could you meanwhile run `make lint` in composition folder just to verify there are no linter errors.

barney-s:
Thank you for this PR @hankfreund

PR #2552, opened 2024-08-21
Add a timeout param to compositions/lint action

barney-s:
### Change description

We have been seeing compositions linter timeout more often than not. Adding timeout as per resolution here: 'https://github.com/golangci/golangci-lint-action/issues/308#issuecomment-896815985'


justinsb:
We should probably make you an owner @hankfreund 

/approve
/lgtm

PR #2555, opened 2024-08-21
kompanion: default group version to resource list GV

justinsb:
With some kube versions (I believe) the group version is empty and
should be defaulted from the containing list.

### DIFF
--- a/experiments/kompanion/cmd/export/export.go
+++ b/experiments/kompanion/cmd/export/export.go
@@ -284,6 +284,12 @@ func RunExport(ctx context.Context, opts *ExportOptions) error {
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


jasonvigil (approved):

experiments/kompanion/cmd/export/export.go
```diff
@@ -302,6 +308,16 @@ func RunExport(ctx context.Context, opts *ExportOptions) error {
 				Resource: apiResource.Name,
 			}
 
+			if gvr.Group == "" {
```
yuwenma, line 311: May I ask in which case this will happen? Or is this related to any special treatment on the client-go lib(`clientset.Discovery()`)? I was assuming the kompanion is for ConfigConnector resources, which shall not have non-empty group ( always `*.cnrm.google.com`) nor version (always  `v1alpha1` or `v1beta1`).

PR #2556, opened 2024-08-21
kompanion: use "Config Connector" instead of KCC

justinsb:
We aim to use the full name to avoid "jargon".


yuwenma (commented):
/lgtm
/approve

Thanks!

PR #2557, opened 2024-08-22
feat: generate CRD & mapper for DataFlowFlexTemplateJob

justinsb:


apis/dataflow/v1beta1/dataflowflextemplatejob_types.go
```diff
@@ -0,0 +1,459 @@
...
+	AdditionalUserLabels map[string]string `json:"additionalUserLabels,omitempty"`
+	*/
+
+	// The Cloud KMS key for the job.
+	//  Key format is:
+	//  projects/<project>/locations/<location>/keyRings/<keyring>/cryptoKeys/<key>
+	KmsKeyNameRef *refs.KMSCryptoKeyRef `json:"kmsKeyNameRef,omitempty"`
+
+	// Configuration for VM IPs.
+	IpConfiguration *string `json:"ipConfiguration,omitempty"`
```
yuwenma, line 139: nit: `IPConfiguration`

apis/dataflow/v1beta1/dataflowflextemplatejob_types.go
```diff
@@ -0,0 +1,459 @@
...
+	*/
+
+	// The Cloud Storage path for staging local files.
+	//  Must be a valid Cloud Storage URL, beginning with `gs://`.
+	StagingLocation *string `json:"stagingLocation,omitempty"`
+
+	// Docker registry location of container image to use for the 'worker harness.
+	//  Default is the container for the version of the SDK. Note this field is
+	//  only valid for portable pipelines.
+	SdkContainerImage *string `json:"sdkContainerImage,omitempty"`
```
yuwenma, line 176: nit: shall this be `SDKContainerImage`?

pkg/controller/direct/context.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+type Context struct {
+	Object *unstructured.Unstructured
+}
+
+func (c *Context) StatusLROStarted() error {
```
yuwenma, line 28: This seems to be not used. Can you add a comment for future work in case we miss it?

apis/refs/v1beta1/projectref.go
```diff
@@ -60,6 +60,9 @@ type Project struct {
 // ResolveProject will resolve a ProjectRef to a Project, with the ProjectID.
 func ResolveProject(ctx context.Context, reader client.Reader, src client.Object, ref *ProjectRef) (*Project, error) {
 	if ref == nil {
+		if projectID := src.GetAnnotations()["cnrm.cloud.google.com/project-id"]; projectID != "" {
+			return &Project{ProjectID: projectID}, nil
+		}
```
yuwenma, line 65: Personally I'd prefer having this `project-id` annotation being handled in each individual KCC resource because this is more of a backward compatible feature and is not what we recommend developers to use (we recommend `spec.projectRef`). WDYT?

apis/refs/v1beta1/gcpserviceaccountref.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 type IAMServiceAccountRef struct {
-	/* The GCP Service Account Email used for auth when secretType is gcpServiceAccount. Allowed value: The `email` field of an `IAMServiceAccount` resource. */
+	/* The `email` field of an `IAMServiceAccount` resource. */
```
yuwenma, line 30: (Is this from some other PRs?)
I think we want to build a convention for resource reference: that the `<Kind>Ref.external`  is the GCP asset inventory-like value. So I'd vote against accepting email as the external value.
 
If it is for the backward compatibility, we can use `IAMServiceAccountRef.email` instead (warning that the `external` is deprecating in the `status`)? Not a blocker of this PR.

yuwenma (commented):
/lgtm

Some formats questions to discuss. No blockers

yuwenma:
/approve

Code looks good (unfortunately the presubmit fails)

justinsb:
Issue is #2571

apis/dataflow/v1beta1/dataflowflextemplatejob_types.go
justinsb: Good call.  I think it wasn't caught by the apichecker because that checks the JSON form (via the CRD)

apis/refs/v1beta1/gcpserviceaccountref.go
justinsb: So I removed the "The GCP Service Account Email used for auth when secretType is gcpServiceAccount. " prefix.  I think that's specific to one API (though I'm not sure which one).  But "The GCP Service Account Email used for auth when secretType is gcpServiceAccount. " should be on the Ref, not on external...

The problem/feature is that fixing it once here means it is fixed in all the APIs that use IAMServiceAccountRef

apis/refs/v1beta1/projectref.go
justinsb: So the challenge is that it's pretty easy to forget, and we might not have test coverage to catch it!

Have we (aka you :-) ) written the "don't use project-id" guidance?

Personally I think we should continue to support the "fallback to namespace name" functionality (which is what is currently implemented by the project-id annotation, though we could do it another way).  Where we have included projectRef, we have made it mandatory.  I actually disagree with that, because it breaks users' ability to have the k8s namespace == the project-id and not have to specify projects everywhere.

But I will move this so that we can reach a full decision / direction here, rather than sneaking it in an unrelated PR :-)

cc @cheftako

pkg/controller/direct/context.go
justinsb: Oops, I added it by mistake.  I'll remove.

apis/dataflow/v1beta1/dataflowflextemplatejob_types.go
justinsb: Done!

justinsb:
I split out the controller logic, that should probably be a separate PR (even though it was helpful to write it at the same time!)

apis/refs/v1beta1/projectref.go
yuwenma: "k8s namespace == the project-id" is nit. My concerns is that KCC have already built several conventions but does not give a complete story and users cannot purely based on the KCC conventions to guess out the correct answers. i.e. If using project-id, users cannot use it as a no-brainer, they need to check if each resource supports that annotation or not. For `spec.projectRef`, at least users can rely on the  CRD validation to make some reasonable guess and self-correct.

apis/refs/v1beta1/gcpserviceaccountref.go
yuwenma: Totally agree.
But still my concern is about the convention that we are using "external" to represent something other than the asset inventory format (`email`) . But as long as the IAMSARef is aligned among all its references, this is not a really big concern.

yuwenma (commented):
/lgtm

Thank you! Looks great

PR #2558, opened 2024-08-22
Add CEL-based validation rules for SQLInstance

jasonvigil:
### Change description
    
- OneOf for password values
- Immutable for CloneSource

Depends on https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2429


jasonvigil:
A bit risky to add CEL for a v1beta1 resource, will do this for an alpha resource first.

PR #2559, opened 2024-08-22
Add predicate to partially enable SQLInstance direct reconciler

jasonvigil:
### Change description

If spec.cloneSource is specified, use direct reconciler. Otherwise, use
the terraform-based reconciler.
    
However, if `KCC_USE_DIRECT_RECONCILERS=SQLInstance` is set, always use
the direct reconciler.
    
This commit adds a reusable way for other direct reconcilers to
accomplish a similar goal of partially enabling the direct reconciler,
based on the resource to-be-reconciled. The ReconcileGate can be
registered for any direct reconciler, with custom logic based on the
resource.

Depends on https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2429

### DIFF
--- a/apis/sql/v1beta1/sqlinstance_types.go
+++ b/apis/sql/v1beta1/sqlinstance_types.go
@@ -595,7 +595,7 @@ type SQLInstanceStatus struct {
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:resource:categories=gcp,shortName=gcpsqlinstance;gcpsqlinstances
 // +kubebuilder:subresource:status
-// +kubebuilder:metadata:labels="cnrm.cloud.google.com/managed-by-kcc=true";"cnrm.cloud.google.com/stability-level=stable";"cnrm.cloud.google.com/system=true";"cnrm.cloud.google.com/tf2crd=true"
+// +kubebuilder:metadata:labels="cnrm.cloud.google.com/managed-by-kcc=true";"cnrm.cloud.google.com/stability-level=stable";"cnrm.cloud.google.com/system=true"
 // +kubebuilder:printcolumn:name="Age",JSONPath=".metadata.creationTimestamp",type="date"
 // +kubebuilder:printcolumn:name="Ready",JSONPath=".status.conditions[?(@.type=='Ready')].status",type="string",description="When 'True', the most recent reconcile of the resource succeeded"
 // +kubebuilder:printcolumn:name="Status",JSONPath=".status.conditions[?(@.type=='Ready')].reason",type="string",description="The reason for the value in 'Ready'"
--- a/pkg/controller/direct/directbase/directbase_controller.go
+++ b/pkg/controller/direct/directbase/directbase_controller.go
@@ -26,7 +26,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/jitter"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/lifecyclehandler"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/metrics"
-	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/predicate"
+	kccpredicate "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/predicate"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/ratelimiter"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/resourceactuation"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/resourcewatcher"
@@ -48,6 +48,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
@@ -64,7 +65,7 @@ func AddController(mgr manager.Manager, gvk schema.GroupVersionKind, model Model
 	if err != nil {
 		return err
 	}
-	return add(mgr, reconciler)
+	return add(mgr, reconciler, deps.ReconcilePredicate)
 }
 
 // NewReconciler returns a new reconcile.Reconciler.
@@ -97,16 +98,21 @@ func NewReconciler(mgr manager.Manager, immediateReconcileRequests chan event.Ge
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler.
-func add(mgr manager.Manager, r *DirectReconciler) error {
+func add(mgr manager.Manager, r *DirectReconciler, reconcilePredicate predicate.Predicate) error {
 	obj := &unstructured.Unstructured{}
 	obj.SetGroupVersionKind(r.gvk)
+
+	predicateList := []predicate.Predicate{kccpredicate.UnderlyingResourceOutOfSyncPredicate{}}
+	if reconcilePredicate != nil {
+		predicateList = append(predicateList, reconcilePredicate)
+	}
 
 	_, err := builder.
 		ControllerManagedBy(mgr).
 		Named(r.controllerName).
 		WithOptions(crcontroller.Options{MaxConcurrentReconciles: k8s.ControllerMaxConcurrentReconciles, RateLimiter: ratelimiter.NewRateLimiter()}).
 		WatchesRawSource(&source.Channel{Source: r.immediateReconcileRequests}, &handler.EnqueueRequestForObject{}).
-		For(obj, builder.OnlyMetadata, builder.WithPredicates(predicate.UnderlyingResourceOutOfSyncPredicate{})).
+		For(obj, builder.OnlyMetadata, builder.WithPredicates(predicateList...)).
 		Build(r)
 	if err != nil {
 		return fmt.Errorf("error creating new controller: %w", err)
@@ -117,7 +123,8 @@ func add(mgr manager.Manager, r *DirectReconciler) error {
 var _ reconcile.Reconciler = &DirectReconciler{}
 
 type Deps struct {
-	JitterGenerator jitter.Generator
+	JitterGenerator    jitter.Generator
+	ReconcilePredicate predicate.Predicate
 }
 
 // DirectReconciler is a reconciler for reconciling resources that support the Model/Adapter pattern.
--- a/pkg/controller/direct/registry/registry.go
+++ b/pkg/controller/direct/registry/registry.go
@@ -20,6 +20,8 @@ import (
 
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/config"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/predicate"
+
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/klog/v2"
 )
@@ -34,6 +36,7 @@ type registration struct {
 	gvk     schema.GroupVersionKind
 	factory ModelFactoryFunc
 	model   directbase.Model
+	rg      predicate.ReconcileGate
 }
 
 type ModelFactoryFunc func(ctx context.Context, config *config.ControllerConfig) (directbase.Model, error)
@@ -44,6 +47,11 @@ func GetModel(gk schema.GroupKind) (directbase.Model, error) {
 		return nil, fmt.Errorf("no model registered for %s", gk)
 	}
 	return registration.model, nil
+}
+
+func GetReconcileGate(gk schema.GroupKind) predicate.ReconcileGate {
+	registration := singleton.registrations[gk]
+	return registration.rg
 }
 
 func PreferredGVK(gk schema.GroupKind) (schema.GroupVersionKind, bool) {
@@ -90,6 +98,17 @@ func RegisterModel(gvk schema.GroupVersionKind, modelFn ModelFactoryFunc) {
 	singleton.registrations[gvk.GroupKind()] = &registration{
 		gvk:     gvk,
 		factory: modelFn,
+	}
+}
+
+func RegisterModelWithReconcileGate(gvk schema.GroupVersionKind, modelFn ModelFactoryFunc, rg predicate.ReconcileGate) {
+	if singleton.registrations == nil {
+		singleton.registrations = make(map[schema.GroupKind]*registration)
+	}
+	singleton.registrations[gvk.GroupKind()] = &registration{
+		gvk:     gvk,
+		factory: modelFn,
+		rg:      rg,
 	}
 }
 
--- a/pkg/controller/direct/sql/sqlinstance_controller.go
+++ b/pkg/controller/direct/sql/sqlinstance_controller.go
@@ -31,6 +31,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/config"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
+	kccpredicate "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/predicate"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
 	"github.com/googleapis/gax-go/v2"
 )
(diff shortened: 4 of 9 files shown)


apis/sql/v1beta1/sqlinstance_types.go
```diff
@@ -441,7 +437,37 @@ type InstanceValueFrom struct {
...
+
+type CloneSource struct {
+	/* Binary log coordinates, if specified, identify the position up to which the source instance is
+	cloned. If not specified, the source instance is cloned up to the most recent binary log coordinates. */
+	// +optional
+	BinLogCoordinates *BinLogCoordinates `json:"binLogCoordinates,omitempty"`
+
+	/* (SQL Server only) Clone only the specified databases from the source instance. Clone all databases if empty. */
+	// +optional
+	SQLDatabaseRefs []refsv1beta1.SQLDatabaseRef `json:"sqlDatabaseRefs,omitempty"`
```
justinsb, line 456: This is a tricky one.... should this be a ref or just a list of strings.  They have to be scoped to the parent source instance, right?  Refs does allow for dynamic naming, but not sure how likely that is.

apis/sql/v1beta1/sqlinstance_types.go
```diff
@@ -569,7 +595,7 @@ type SQLInstanceStatus struct {
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:resource:categories=gcp,shortName=gcpsqlinstance;gcpsqlinstances
 // +kubebuilder:subresource:status
-// +kubebuilder:metadata:labels="cnrm.cloud.google.com/managed-by-kcc=true";"cnrm.cloud.google.com/stability-level=stable";"cnrm.cloud.google.com/system=true";"cnrm.cloud.google.com/tf2crd=true"
+// +kubebuilder:metadata:labels="cnrm.cloud.google.com/managed-by-kcc=true";"cnrm.cloud.google.com/stability-level=stable";"cnrm.cloud.google.com/system=true"
```
justinsb, line 598: We'll need to be careful that this resource isn't skipped by the TF reconciler - IIUC we want both?

pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -97,16 +98,21 @@ func NewReconciler(mgr manager.Manager, immediateReconcileRequests chan event.Ge
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler.
-func add(mgr manager.Manager, r *DirectReconciler) error {
+func add(mgr manager.Manager, r *DirectReconciler, rolloutPredicate predicate.Predicate) error {
```
justinsb, line 101: Nit: `rolloutPredicate` is good in that it says what it is used for, but "rollout" could refer to the resource rolling out or something like that.  So it _might_ be better to call this `additionalPredicate` or `additionalPredicates` or something like that... I'm not sure though!

pkg/controller/direct/registry/registry.go
```diff
@@ -46,6 +49,14 @@ func GetModel(gk schema.GroupKind) (directbase.Model, error) {
 	return registration.model, nil
 }
 
+func GetRolloutGate(gk schema.GroupKind) (predicate.RolloutGate, error) {
+	registration := singleton.registrations[gk]
+	if registration == nil || registration.model == nil {
+		return nil, fmt.Errorf("no model registered for %s", gk)
```
justinsb, line 55: I'm not sure that this should be an error (?)

pkg/controller/direct/sql/mapping.go
```diff
@@ -793,10 +800,44 @@ func Convert_SQLInstance_API_v1_To_KRM_status(in *api.DatabaseInstance, out *krm
...
+	}
+
+	resourceID := ValueOf(in.Spec.ResourceID)
+	if resourceID == "" {
+		resourceID = in.Name
 	}
-	return &v
+	cloneReq.CloneContext.DestinationInstanceName = resourceID
+
+	if in.Spec.Settings.IpConfiguration != nil {
```
justinsb, line 827: Will be interesting to see if we can get this under fuzz-testing, so we can detect new fields...

pkg/controller/direct/sql/normalize.go
```diff
@@ -65,19 +68,29 @@ func NormalizeSQLInstance(ctx context.Context, kube client.Reader, obj *krm.SQLI
 	if err != nil {
 		return nil, err
 	}
+	sourceSQLInstance, err := normalizeSourceSQLInstanceRef(ctx, kube, obj)
```
justinsb, line 71: Personally I've been liking converting the refs "in place", I.e. updating them to replace name/namespace refs with external refs.  The nice thing is that then I think we could create a generic function that walks a struct using reflection, finds Refs (by interface) and calls a Resolve function on them (defined on the interface).

But YMMV, I'm watching how this shapes out!  It also doesn't _really_ matter - the bulk of the logic here is in the resolvers anyway.

pkg/controller/direct/sql/normalize.go
```diff
@@ -290,6 +303,95 @@ func normalizeAuditLogBucketRef(ctx context.Context, kube client.Reader, obj *kr
...
+	sqlDatabaseRefs := obj.Spec.CloneSource.SQLDatabaseRefs
+	sqlDatabases := make([]string, len(sqlDatabaseRefs))
+
+	for i, sqlDatabaseRef := range sqlDatabaseRefs {
+		if sqlDatabaseRef.External != "" && sqlDatabaseRef.Name != "" {
+			return nil, fmt.Errorf("cannot specify both spec.settings.cloneSource.sqlDatabaseRefs[].external and spec.settings.cloneSource.sqlDatabaseRefs[].name")
+		}
+
+		if sqlDatabaseRef.External != "" {
+			sqlDatabases[i] = sqlDatabaseRef.External
```
justinsb, line 362: One gotcha here is that I expect (some) users will use the full path for most external databases (e.g. `projects/foo/locations/bar/instances/baz/databases/bat` ?

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -31,14 +31,33 @@ import (
...
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
 	"github.com/googleapis/gax-go/v2"
 )
 
 const ctrlName = "sqlinstance-controller"
 
 func init() {
-	registry.RegisterModel(krm.SQLInstanceGVK, newSQLInstanceModel)
+	rolloutGate := &SQLInstanceRolloutGate{}
+	registry.RegisterModelWithRolloutGate(krm.SQLInstanceGVK, newSQLInstanceModel, rolloutGate)
```
justinsb, line 43: Tip: This is fine, another pattern is to create a struct to give something like Python named arguments e.g.

```
registry.RegisterModel(registry.Options{GVK: krm.SQLInstanceGVK, ModelFn: newSQLInstanceModel, RolloutGate: rolloutGate}
```

At 3 arguments, not a problem.  But when the number creeps up, particularly if it's a variable number of arguments, becomes more useful.

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -31,14 +31,33 @@ import (
...
+}
+
+type SQLInstanceRolloutGate struct{}
+
+var _ kccpredicate.RolloutGate = &SQLInstanceRolloutGate{}
+
+func (*SQLInstanceRolloutGate) ShouldReconcile(o *unstructured.Unstructured) bool {
+	u, err := runtime.DefaultUnstructuredConverter.ToUnstructured(o)
```
justinsb, line 51: I'm not clear why we need this first ToUnstructured.  It looks like we're converting unstructured -> unstructured?

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -69,7 +88,7 @@ var _ directbase.Adapter = &sqlInstanceAdapter{}
 func (m *sqlInstanceModel) AdapterForObject(ctx context.Context, kube client.Reader, u *unstructured.Unstructured) (directbase.Adapter, error) {
 	obj := &krm.SQLInstance{}
 	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &obj); err != nil {
-		return nil, fmt.Errorf("error converting to %T: %w", obj, err)
+		return nil, fmt.Errorf("converting to %T failed: %w", obj, err)
```
justinsb, line 91: 👍

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -138,14 +157,67 @@ func (a *sqlInstanceAdapter) Create(ctx context.Context, u *unstructured.Unstruc
...
+	for {
+		log.V(2).Info("polling", "op", op)
+
+		if op.Status == "DONE" {
+			break
+		}
+		if err := gax.Sleep(ctx, pollingBackoff.Pause()); err != nil {
+			return fmt.Errorf("waiting for SQLInstance %s clone failed: %w", a.desired.Name, err)
+		}
+		op, err = a.sqlOperationsClient.Get(a.projectID, op.Name).Do()
```
justinsb, line 192: If you have ideas or a PoC for surfacing information during this reconciliation ... that is so hot right now!

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -185,7 +257,7 @@ func (a *sqlInstanceAdapter) Create(ctx context.Context, u *unstructured.Unstruc
 				// Ref: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance
```
justinsb, line 257: One problem with doing more than one thing in a reconciliation is what happens if the first one succeeds but the second fails.  So we might want to extract this into a function, so that we can also call it when handling the adoption case.  But of course, then we need to be careful that we don't delete the root user if the user created it.

(Arguably, this is why it's not a good idea to do things like this in the IaC layer, particularly if the user has not asked us to)

pkg/controller/predicate/rolloutgate.go
```diff
@@ -0,0 +1,150 @@
...
+
+func NewRolloutPredicate(c client.Client, gvk schema.GroupVersionKind, rg RolloutGate) *RolloutPredicate {
+	return &RolloutPredicate{
+		c:   c,
+		gvk: gvk,
+		rg:  rg,
+	}
+}
+
+func (p *RolloutPredicate) Create(e event.CreateEvent) bool {
```
justinsb, line 53: I personally like to add a comment like `// Create implements the Predicate interface` just to make it clearer what we're doing.  I've seen it done elsewhere, and I assume at some stage there will be better support in IDEs for going to the interface method definition (so it's not worth doing more)...

pkg/controller/predicate/rolloutgate.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	return p.rg.ShouldReconcile(obj)
+}
+
+func (p *RolloutPredicate) Generic(e event.GenericEvent) bool {
+	obj, err := getUnstructuredObjWithGVK(p.c, e.Object, p.gvk)
+	if err != nil {
+		return false
+	}
+	return p.rg.ShouldReconcile(obj)
```
justinsb, line 82: Nit: Feels like you could pass the event to ShouldReconcile?

pkg/controller/predicate/rolloutgate.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// InverseRolloutPredicate generates a controller-runtime predicate based on the inverse of a RolloutGate.
+type InverseRolloutPredicate struct {
```
justinsb, line 86: Nit: might be easier to just have an InversePredicate that always returns the inverse of a predicate.  You don't have to export `inversePredicate`, rather `NewInverseRolloutPredicate` could just return `k8spredicate.Predicate`.

pkg/controller/predicate/rolloutgate.go
justinsb: Ah I see, actually this makes sense because I think we're assuming that at some stage we will watch more intelligently and not the extra GET.  We could eliminate some of the duplication by having a helper method in `RolloutPredicate`, but ... not a blocker.

justinsb (commented):
A few comments, but I don't think anything blocking

apis/sql/v1beta1/sqlinstance_types.go
justinsb: You did handle this, so I think you can disregard!

justinsb:
Overall LGTM - comments but mostly not blocking.  This is going to be great and unblock moving important resources to direct actuation 🎉 

One blocker (I think): sqlDatabaseRefs or databaseNames.  The databases have to be scoped to the source sqlInstance, I believe, so I think databaseNames might be fine (?)

apis/sql/v1beta1/sqlinstance_types.go
jasonvigil: Yeah, good point. Fixed in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2429

apis/sql/v1beta1/sqlinstance_types.go
jasonvigil: Yeah, it should run both, unless there is an error with getting or converting the resource in the predicate. In that case, neither controller will run (currently).

pkg/controller/direct/directbase/directbase_controller.go
jasonvigil: Hmm, I had some trouble deciding on a name for this one. Just changed it to `ReconcileGate` / `ReconcilePredicate`, wdyt?

pkg/controller/direct/registry/registry.go
jasonvigil: Yeah, fair enough. Changed it to not return any error, just nil if there is no rollout gate set.

pkg/controller/direct/sql/normalize.go
jasonvigil: Hmm, that sounds cool! I am hopeful that there will be time to squeeze in some time to use the tools for generating mapping / normalizing functions for SQLInstance in the future. In any case, it's internal (shouldn't affect users), so it will be possible to perform this refactor later.

pkg/controller/direct/sql/normalize.go
jasonvigil: Agreed it's probably simpler to just use a list of strings of database names. Fixed in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2429.

pkg/controller/direct/sql/sqlinstance_controller.go
jasonvigil: I kinda intentionally created another function for this (and didn't change the original function signature), in order to sort of "hide" this feature. Not really sure that we want to encourage direct reconciler developers to use this capability unless there's a strong need. Thought it might be desirable to have the default / most obvious way to be without the predicate / gate, instead (for most resources) we would want to just switch over to direct reconciler completely.

pkg/controller/direct/sql/sqlinstance_controller.go
jasonvigil: Good point. Looking closer, it seems we can just use the internal `.Object` field as the input to `FromUnstructured`. Got a bit confused on this because `FromUnstructured` needs the `map[string]interface{}` as an input, which isn't exactly the same as `unstructured.Unstructured`.

pkg/controller/predicate/rolloutgate.go
jasonvigil: Yeah, it felt like this was the least hassle way to do it, because the `UpdateEvent` passes in `ObjectNew` and `ObjectOld` (unlike all of the other events, which only pass in `Object`). So, there wasn't really any way to make a generic handler for every single event.

pkg/controller/predicate/rolloutgate.go
jasonvigil: This was sorta intentional, but I am not sure if it was wise. The idea was that if there is an error, then `inversePredicate` would return false also (same as non-inverse predicate). In that case, neither direct or terraform reconcilers would run.

pkg/controller/predicate/rolloutgate.go
jasonvigil: For me, I usually just add a "self-documenting" variable assignment (like here: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2559/files#diff-b3ba7e1f2676394700779aab805a0ac09590b4d96b002e9f84eb581915a6eb17R43) to show that it fits the interface: `var _ k8spredicate.Predicate = &RolloutPredicate{}`. Not sure if comments are really all that necessary

pkg/controller/direct/sql/mapping.go
jasonvigil: Where is the fuzz testing code? I am still just trying to achieve test coverage of the basic functionality lol.

pkg/controller/direct/sql/sqlinstance_controller.go
jasonvigil: Hmm.. probably out of scope for this PR, but am def planning to improve the code for this soon.

pkg/controller/direct/sql/sqlinstance_controller.go
jasonvigil: Agreed. Though, maybe we don't need to delete the user at all? This was only done to achieve implementation parity with Terraform. Perhaps it would be ok to just.. not do this? Wdyt?

justinsb: I don't think we want to change the behaviour here.  But yeah, I'm sure it will come up again...

justinsb (approved):

PR #2563, opened 2024-08-23
Add SQLInstance network mockgcp tests

jasonvigil:
### Change description

These are probably among the most used fields of SQLInstance resources, we should add some test cases for them.

Also, refactor SQLInstance direct controller mapper / merge code (just for `settings.IpConfiguration`, so far) to fix some issues that were discovered along the way, and to make it so that the code more closely resembles what the generated type mapping code would be (to make it easier to migrate later).
- `ipConfiguration.ipv4Enabled` should NOT be sent to API unless explicitly specified by user.
- Detect differences between `ipConfiguration.authorizedNetworks` when list sizes do not change, but values do change.


mockgcp/mocksql/sqlinstance.go
```diff
@@ -113,17 +113,26 @@ func (s *sqlInstancesService) Insert(ctx context.Context, req *pb.SqlInstancesIn
...
+		if isPostgres(obj) {
+			obj.IpAddresses = append(obj.IpAddresses, &pb.IpMapping{
+				IpAddress: "10.10.10.11",
+				Type:      pb.SqlIpAddressType_OUTGOING,
+			})
+		}
+	} else {
+		obj.IpAddresses = []*pb.IpMapping{
+			{
+				IpAddress: "192.168.0.3",
```
jasonvigil, line 132: These IPs are currently hardcoded in the mockgcp implementation for SQLInstance. Should we consider extending the mock implementation to resolve ComputeAddress resource references (is this even possible?) and allocate IPs from their specified network ranges? Or, is this overkill? Not sure.

pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/sqlinstanceauthorizednetworks/create.yaml
```diff
@@ -0,0 +1,37 @@
...
+spec:
+  databaseVersion: POSTGRES_16
+  region: us-central1
+  settings:
+    ipConfiguration:
+      authorizedNetworks:
+        - name: all
+          value: 0.0.0.0/0
+          expirationTime: 2099-01-01T00:00:00+00:00
+      ipv4Enabled: true
```
jasonvigil, line 30: This field is defaulted to true in the terraform-based controller. Not sure how to handle converting existing resources to the direct controller, because I feel that it should not be defaulted to true. Thoughts / ideas? Is it ok to change the behavior to make it not true by default?

jasonvigil: Hmm.. actually, it seems like the API defaults to setting `ipv4Enabled: true` if the `ipConfiguration` is not set. So.. maybe it is better behavior to default ipv4Enabled to true actually. Though, not quite sure how to handle this. Maybe in the CRD? Will come back to this on Monday..

pkg/controller/direct/sql/mapping.go
```diff
@@ -473,6 +412,73 @@ func SQLInstanceKRMToGCPInstance(in *krm.SQLInstance, refs *SQLInstanceInternalR
...
+	return out
+}
+
+func InstancePscConfigKRMToGCP(in []krm.InstancePscConfig) *api.PscConfig {
+	if len(in) < 1 {
+		return nil
+	}
+
+	// Note:  For some reason, the KRM API allows []InstancePscConfig. However, in the GCP proto there is only
+	// a single *api.PscConfig. I think there is a mistake in the KRM API; it should not allow a list. For
```
justinsb, line 466: This seems likely, there are 

You might want to return an error from Reconcile `if len(in) > 1`, we can always _also_ add a CEL rule later.

pkg/test/resourcefixture/contexts/sql_context.go
```diff
@@ -59,13 +59,34 @@ func init() {
 		ResourceKind:  "SQLInstance",
 	}
 
+	resourceContextMap["sqlinstanceauthorizednetworks"] = ResourceContext{
```
justinsb, line 62: Do you understand what this does?  (I don't!)

I wonder if we should have this functionality in the unified_test.

justinsb (approved):

PR #2568, opened 2024-08-26
Remove TestSimpleCompositionCreate

barney-s:
### Change description

Remove TestSimpleCompositionCreate

- This test was useful at the begining when composition create was failing.
- Now code is mature and all other tests cover this aspect


hankfreund:
Should also delete the `data/TestSimpleCompositionCreate/` folder.

barney-s:
Good catch. updated the PR

PR #2569, opened 2024-08-27
tests: add golden objects for monitoring tests

justinsb:


yuwenma (commented):
/lgtm
/approve

Looks good. Do you happen to know why the presubmit tests does not require the golden object before?

PR #2572, opened 2024-08-27
tests: add periodics for pubsub and sql

justinsb:


jasonvigil:
How do these tests get run (aside from manually)? Are they run automatically yet?

jasonvigil:
> How do these tests get run (aside from manually)? Are they run automatically yet?

Ah, looks like this is WIP.

jasonvigil (approved):

PR #2577, opened 2024-08-28
feat:add a label for kcc resource type

xiaoweim:
This PR adds a default label to KCC resource objects specifying the resource type in our config connector E2E tests.

This label will help us generate a mapping between KCC resources and GCP/GCE resources, which will help attribute revenue to individual KCC resource type.

This PR is a newer version of the outdated PR https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2452 that was closed.


pkg/test/resourcefixture/testdata/stateabsentinspec/bigquerydataset/_generated_export_bigquerydataset#01.golden
```diff
@@ -5,6 +5,7 @@ metadata:
     cnrm.cloud.google.com/delete-contents-on-destroy: "false"
   labels:
     cnrm-test: "true"
+    config-connector-resource-type: bigquerydataset
```
yuwenma, line 8: I think we do not want to add this label to the exported objects. Or do you mind sharing more context on why this is needed?

yuwenma (commented):
/hold

One question for clarification. Otherwise looks good!

pkg/test/resourcefixture/testdata/stateabsentinspec/bigquerydataset/_generated_export_bigquerydataset#01.golden
yuwenma: I see your confusion: The golden log captures what will happen to the users. The label is added based on the code change in the PR. So my original question is that I don't think we want the code change that will reflect as a new label added in the exported object.

xiaoweim: So you are saying that it is expected for the golden http logs to be changed, but the golden object should not be changed? 
I tried updating only the golden http logs but presubmit tests failed and suggests changes to be made to a few golden objects. I am my questions is more on what exactly is the exported object and what do we do with it during tests?
I want to get all presubmit tests passing, and I would like to understand what is this exported object and its functionality, so that I can understand why the presubmit tests failed when I did not update it or what should I do to keep the exported object and also the tests passing.

PR #2580, opened 2024-08-28
feat: better status updates for DataflowFlexTemplateJob

justinsb:
We update the state of the object before reconciliation has completed.


pkg/test/resourcefixture/testdata/basic/dataflow/v1beta1/dataflowflextemplatejob/streamingdataflowflextemplatejob/_generated_object_streamingdataflowflextemplatejob.golden.yaml
```diff
@@ -3,7 +3,7 @@ kind: DataflowFlexTemplateJob
 metadata:
   annotations:
     cnrm.cloud.google.com/management-conflict-prevention-policy: none
-    cnrm.cloud.google.com/mutable-but-unreadable-fields: '{"spec":{"containerSpecGcsPath":"gs://dataflow-templates/2020-08-31-00_RC00/flex/PubSub_Avro_to_BigQuery","parameters":{"createDisposition":"CREATE_NEVER","inputSubscription":"projects/${projectId}/subscriptions/pubsubsubscription-${uniqueId}","outputTableSpec":"${projectId}:bigquerydataset${uniqueId}.bigquerytable${uniqueId}","outputTopic":"projects/${projectId}/topics/pubsubtopic1-${uniqueId}","schemaPath":"gs://config-connector-samples/dataflowflextemplate/numbers.avsc"}}}'
+    cnrm.cloud.google.com/mutable-but-unreadable-fields: '{"spec":{"containerSpecGcsPath":"gs://dataflow-templates/2020-08-31-00_RC00/flex/PubSub_Avro_to_BigQuery","parameters":{"createDisposition":"CREATE_NEVER","inputSubscription":"projects/${projectId}/subscriptions/pubsubsubscription-${uniqueId}","outputTableSpec":"${projectId}:bigquerydataset${uniqueId}.bigquerytable${uniqueId}","outputTopic":"projects/${projectId}/topics/pubsubtopic0-${uniqueId}","schemaPath":"gs://config-connector-samples/dataflowflextemplate/numbers.avsc"}}}'
```
yuwenma, line 6: May I ask why this is a mutable but unreadable fields?
I'm thinking that since we are migrating to the direct resource, we should drop this annotation and switch to some other mechanism (I don't think the mutable-but-unreadable-fields in general will break users).

justinsb: Ah sure - this test is actually _not_ using the direct controller.  This is actually me noticing another problem in our tests and fixing it.  I will just split out this PR, it's not really relevant here!

We do use the direct controller in  `batchdataflowflextemplatejob-direct`, but that is actually using the annotation to opt-in to the direct controller.

yuwenma: Ah, gotcha.

pkg/controller/direct/dataflow/dataflowflextemplatejob_controller.go
```diff
@@ -330,30 +333,54 @@ func (a *dataflowFlexTemplateJobAdapter) Create(ctx context.Context, createOp *d
...
-		case pb.JobState_JOB_STATE_RUNNING:
-			ready = true
-		default:
-			log.Info("unexpected job state waiting for job running", "state", latest.CurrentState)
+
+	case pb.JobState_JOB_STATE_FAILED:
+		readyCondition = &v1alpha1.Condition{
+			Type:    v1alpha1.ReadyConditionType,
+			Status:  v1.ConditionFalse,
+			Reason:  k8s.UpToDate,
```
yuwenma, line 366: May I ask why this is UpToDate when the job is failed?

pkg/controller/direct/dataflow/dataflowflextemplatejob_controller.go
```diff
@@ -330,30 +333,54 @@ func (a *dataflowFlexTemplateJobAdapter) Create(ctx context.Context, createOp *d
...
-		JobID: jobID,
+	if err := op.UpdateStatus(ctx, status, readyCondition); err != nil {
+		return fmt.Errorf("updating status: %w", err)
 	}
-	return setStatus(u, status)
+
+	return nil
 }
 
 // Update implements the Adapter interface.
```
yuwenma, line 386: Shall we also check the Job status to decide whether to update or not?  
i.e. pb.JobState_JOB_STATE_CANCELLED --> no update and stop reconciling

pkg/controller/direct/dataflow/dataflowflextemplatejob_controller.go
```diff
@@ -330,30 +333,54 @@ func (a *dataflowFlexTemplateJobAdapter) Create(ctx context.Context, createOp *d
...
+	status := &krm.DataflowFlexTemplateJobStatus{
+		JobID:        job.GetId(),
+		CurrentState: direct.PtrTo(job.CurrentState.String()),
+		Type:         direct.PtrTo(job.Type.String()),
+	}
+
+	var readyCondition *v1alpha1.Condition
+
+	switch job.CurrentState {
+	case pb.JobState_JOB_STATE_RUNNING:
```
yuwenma, line 354: I think pb.JobState_JOB_STATE_UPDATED also applies to this case,  right?

pkg/controller/direct/dataflow/dataflowflextemplatejob_controller.go
```diff
@@ -363,7 +390,31 @@ func (a *dataflowFlexTemplateJobAdapter) Update(ctx context.Context, updateOp *d
...
+		j, _ := u.MarshalJSON()
+		log.V(0).Info("object is", "json", string(j))
+
+		// If we are waiting on the existing job, update the status
+		if a.actual != nil {
+			if err := a.updateStatus(ctx, updateOp, a.actual); err != nil {
+				return err
+			}
+
+			// TODO: If job fails to update, we probably need to "put back" the old job id
```
yuwenma, line 413: May I ask what does "put back the old job ID" means? Does the JobID changes in pending status until it is Running? If so, it seems that we also need to handle  the `spec.resourceID`  in the case that a job is pending, and users uses a unstable JobID trying to acquire/manage that Job.

pkg/controller/direct/directbase/operations.go
```diff
@@ -14,41 +14,69 @@
...
+	client client.Client
 
-type UpdateOperation struct {
 	object *unstructured.Unstructured
+
+	// HasSetReadyCondition tracks whether the controller explcitly set the ready condition
+	HasSetReadyCondition bool
+
+	// ShouldRequeue tracks whether we need a re-reconciliation
+	ShouldRequeue bool
```
yuwenma, line 40: oh, I like this. Though I don't see it is used in the dataflowflextemplateJob `Update`, did I miss anything?

pkg/controller/direct/directbase/operations.go
```diff
@@ -57,3 +85,106 @@ func NewDeleteOperation(object *unstructured.Unstructured) *DeleteOperation {
...
+
+	old, _, _ := unstructured.NestedMap(o.object.Object, "status")
+	if old != nil {
+		if status["conditions"] == nil {
+			status["conditions"] = old["conditions"]
+		}
+		// status["observedGeneration"] = old["observedGeneration"]
+		if status["externalRef"] == nil {
+			status["externalRef"] = old["externalRef"]
+		}
```
yuwenma, line 103: Mind adding the condition that if not equal, return errors as well? I'm interested about the externalRef especially for the `DataflowFlexTemplateJob` since it looks like the JobID has some specialties.

pkg/controller/direct/dataflow/dataflowflextemplatejob_controller.go
```diff
@@ -330,30 +333,54 @@ func (a *dataflowFlexTemplateJobAdapter) Create(ctx context.Context, createOp *d
...
-			ready = true
-		default:
-			log.Info("unexpected job state waiting for job running", "state", latest.CurrentState)
+
+	case pb.JobState_JOB_STATE_FAILED:
+		readyCondition = &v1alpha1.Condition{
+			Type:    v1alpha1.ReadyConditionType,
+			Status:  v1.ConditionFalse,
+			Reason:  k8s.UpToDate,
+			Message: "Dataflow job is in FAILED state",
```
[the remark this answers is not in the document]
justinsb: Yes, but let's do that when we have something variable in the message.

pkg/controller/direct/dataflow/dataflowflextemplatejob_controller.go
justinsb: Added a comment.  It's because (I think) we want to stop reconciliation at this point: we have reconciled the desired state and the outcome is a failed job.

yuwenma:
/lgtm
/approve

Thank you for the enhancement for DFTJ!

pkg/controller/direct/dataflow/dataflowflextemplatejob_controller.go
justinsb: Actually, changed my mind and made it explicit - we now call op.RequestRequeue when we want to requeue.

pkg/controller/direct/dataflow/dataflowflextemplatejob_controller.go
justinsb: Agreed, added a bunch more terminal but not ready states.

Also made the requeuing explicit - it was just too magic.  This means we can set the ready condition however we best see fit.  I think the current ready logic is right - on a terminal condition we stop reconciling but ready=false, but IDK!

pkg/controller/direct/dataflow/dataflowflextemplatejob_controller.go
justinsb: There's a separate discussion about what the primary key for this job really should be.

Today jobs are technically immutable.  We can replace a job, but the new job gets a new job ID.  If that new job fails, the old job continues running/resumes.  In this code we now keep the new job ID, so that we don't keep launching new (failing) jobs forever.  However, it does mean that the job is still running, just with the old job ID.

pkg/controller/direct/directbase/operations.go
justinsb: It was used, I just made it much more explicit - the controller must now call `op.RequestRequeue()`

pkg/controller/direct/directbase/operations.go
justinsb: This externalRef will change, because it includes the job id :-(

That's one reason why I think we might currently have the wrong primary key for this resource ... trying to track that down internally.

PR #2581, opened 2024-08-28
refactor: pass operation objects to adapters

justinsb:
This should allow for richer behaviours, such as intermediate updates of status.

### DIFF
--- a/pkg/controller/direct/directbase/operations.go
+++ b/pkg/controller/direct/directbase/operations.go
@@ -1,0 +1,59 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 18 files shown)


justinsb:
I'm experimenting with using this in the DataflowFlexTemplateJob controller in #2580 , if you want to peek at the last commit...

jasonvigil:
/approve

Looks like there are some failing tests though

pkg/controller/direct/directbase/operations.go
```diff
@@ -0,0 +1,45 @@
...
+	object *unstructured.Unstructured
+}
+
+func NewUpdateOperation(object *unstructured.Unstructured) *UpdateOperation {
+	return &UpdateOperation{
+		object: object,
+	}
+}
+
+func (o *UpdateOperation) UnstructuredObject() *unstructured.Unstructured {
```
yuwenma, line 15: My first read from the fn name  is that I thought you mean the `unstructured.Unstructured` `Object`. https://github.com/kubernetes/apimachinery/blob/v0.31.0/pkg/apis/meta/v1/unstructured/unstructured.go#L45C2-L45C8 
Can we rename this something like `Unstructured()` or `GetUnstructured()`?

yuwenma (commented):
/lgtm
/approve

pkg/controller/direct/directbase/operations.go
justinsb: Good call, done!

jasonvigil (approved):

PR #2582, opened 2024-08-28
Add support for deleting resources in reverse order by stage.

hankfreund:
### Change description

This solution makes use of the ApplySet annotations for GroupKinds and Namespaces. When a facade deletion request is received, the ApplySet data is read from the Plan to know what GKs to find and delete, and which namespaces to search. The resources are further filtered by the ApplySet ID they are automatically labeled with.
Stage order information is added to the Plan so the deletion processing can iterate over them in reverse, and as a label on each expanded resource.


experiments/compositions/composition/internal/controller/expander_reconciler.go
```diff
@@ -69,6 +73,15 @@ type EvaluateWaitError struct {
 	msg string
 }
 
+const (
+	finalizerName    = "compositions.google.com/expander-finalizer"
+	stagesAnnotation = "compositions.google.com/expander-stages"
+	nsAnnotation     = "applyset.kubernetes.io/additional-namespaces"
```
barney-s, line 79: Do you think we can use this instead of defining it here ? 

https://pkg.go.dev/sigs.k8s.io/kubebuilder-declarative-pattern/applylib@v0.0.0-20240813003123-2a9fe229ddb2/forked/github.com/kubernetes/kubectl/pkg/cmd/apply#pkg-constants

experiments/compositions/composition/internal/controller/expander_reconciler.go
```diff
@@ -127,6 +140,9 @@ func (r *ExpanderReconciler) getPlanForInputCR(ctx context.Context, inputcr *uns
 				"name", plancr.Name, "namespace", plancr.Namespace)
 			return planNN, &plancr, err
 		}
+		// Add a finalizer to the Plan to discourage out of band deletions. Stage information is
+		// stored in the Plan to handle proper cleanup of resources when a Facade is deleted.
+		controllerutil.AddFinalizer(&plancr, finalizerName)
```
barney-s, line 145: Would we need this in the else case as well ? 
Existing Plan CRs ? 
I dont think it is a required. Just a thought.

experiments/compositions/composition/internal/controller/expander_reconciler.go
```diff
@@ -158,6 +174,18 @@ func (r *ExpanderReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	if !inputcr.GetDeletionTimestamp().IsZero() {
+		return r.reconcileDelete(ctx, logger, inputcr)
+	}
+	// Add a finalizer to prevent removal of facade before all applied objects are cleaned up.
+	if !controllerutil.ContainsFinalizer(&inputcr, finalizerName) {
+		controllerutil.AddFinalizer(&inputcr, finalizerName)
```
barney-s, line 182: Do we need a finalizer for the plan object ? If we add a finalizer here ?

experiments/compositions/composition/tests/testcases/simple_test.go
```diff
@@ -472,3 +453,64 @@ func TestMultipleCompositionsDisallowedForSameGVK(t *testing.T) {
...
+	if err := s.C.Update(context.Background(), cms[1]); err != nil {
+		t.Fatalf("failed removing finalizer from configmap: %v", err)
+	}
+
+	// Verify the remaining objects have been deleted
+	s.C.MustHaveEvent("configs", "Delete", "Deleting objects for stage createnamespace", scenario.DeleteTimeout)
+	s.C.MustNotExist([]*unstructured.Unstructured{cr}, scenario.DeleteTimeout)
+	s.C.MustNotExist(cms, scenario.DeleteTimeout)
+
+	cr = utils.GetUnstructuredObj("facade.foocorp.com", "v1alpha1", "DConfig", "configs", "config-team-b")
```
barney-s, line 503: 503-511 is this testing a regular delete  ?

barney-s: Wondering if we should have a separate test for this ?

experiments/compositions/composition/internal/controller/expander_reconciler.go
```diff
@@ -665,6 +702,147 @@ func (r *ExpanderReconciler) enqueueAllFromGVK(ctx context.Context, _ client.Obj
...
+	for i := len(stages) - 1; i >= 0; i-- {
+		r.Recorder.Eventf(&inputcr, corev1.EventTypeNormal, "Delete", "Deleting objects for stage %s", stages[i])
+		nsList, ok := annotations[nsAnnotation]
+		if !ok {
+			err := fmt.Errorf("Plan is missing Namespace annotation")
+			logger.Error(err, "Unable to fetch Namespaces from Plan", "Plan", planNN)
+			return ctrl.Result{}, err
+		}
+		namespaces := strings.Split(nsList, ",")
+		namespaces = append(namespaces, inputcr.GetNamespace())
```
barney-s, line 738: why is this necessary ?

experiments/compositions/composition/internal/controller/expander_reconciler.go
hankfreund: Finalizers on both resources won't prevent one another from being deleted. The Plan finalizer is mostly to keep the plan from being accidentally removed out of band, which would cause big problems for facade deletion.

experiments/compositions/composition/internal/controller/expander_reconciler.go
hankfreund: The facade namespace might have resources in it (as is the case in most of our tests). ApplySet adds a list of "additional" namespaces -- namely, namespaces other than the inputcr's namespace.

experiments/compositions/composition/internal/controller/expander_reconciler.go
hankfreund: Yeah, it can't hurt to check pre-existing plans and ensure they have the finalizer as well.

experiments/compositions/composition/internal/controller/expander_reconciler.go
hankfreund: Yes, good call.

experiments/compositions/composition/internal/controller/expander_reconciler.go
```diff
@@ -665,6 +706,147 @@ func (r *ExpanderReconciler) enqueueAllFromGVK(ctx context.Context, _ client.Obj
...
+				n, err = r.deleteClusterResources(ctx, logger, stages[i], mapping.Resource, opts)
+			}
+			if err != nil {
+				logger.Error(err, "Error deleting resources", "GroupKind", gk)
+				r.Recorder.Eventf(&inputcr, corev1.EventTypeWarning, "Delete", "Failed deleting objects of GroupKind %q for stage %q: %v", gk, stages[i], err)
+				return ctrl.Result{}, err
+			}
+			numFound += n
+		}
+		if numFound > 0 {
```
barney-s, line 774: Logic here implies if a resource in a particular stage is not deleted (for whatever reason), the deletion of rest of the resources are blocked (fits the mental model). This is great if the delete pending resource eventually is removed (O(mins)). Iam thinking if we need to add a timeout (by looking at delete timestamp of the resource) to make progress if a resource is stuck in deletion. We can proceed to delete other resources after a timeout. We dont have to delete the plan/inputcr.

barney-s: we can do it in another PR ?

PR #2585, opened 2024-08-29
tests: funnel all tests through run-e2e

justinsb:
This means we only need to maintain the KCC_USE_DIRECT_RECONCILERS list in one spot.


yuwenma (commented):
/lgtm
/approve

PR #2587, opened 2024-08-29
fix: Rename sqlinstance test fixtures to not end in "basic"

jasonvigil:
Also, take the opportunity to set a new standard pattern for sqlinstance test names.

Test names should not end in "basic" otherwise they will be placed in the "autogen" test category.


pkg/test/resourcefixture/contexts/sql_context.go
```diff
@@ -17,72 +17,72 @@ package contexts
 import "time"
 
 func init() {
-	resourceContextMap["mysqlinstance"] = ResourceContext{
+	resourceContextMap["sqlinstance-authorizednetworks"] = ResourceContext{
```
gemmahou, line 20: Very nit comment: For some reason we do not have special characters like "-" in fixtures testnames. I'm unsure if that would cause any issue, but I suggest to keep consistent with other test names.

jasonvigil: Yeah, I noticed that. But do you know of why we shouldn't? It seems nice to have the visual separator. Could easily revert this change if adding them breaks stuff.

gemmahou: I do not find any documentation around this... we can change the format and see how it goes.

PR #2590, opened 2024-08-29
Fix dynamic test scifi issues

jasonvigil:
Properly detect if the resource uses Terraform, DCL, or Direct controller.

Also, modify the dynamic tests to not check the spec.resourceID value for direct resources, because direct reconcilers should not write to this field.


pkg/test/resourcefixture/contexts/register.go
```diff
@@ -66,43 +71,57 @@ type ResourceContext struct {
...
 	if rc.ResourceGVK == emptyGVK {
 		rc.ResourceGVK = fixture.GVK
 	}
+
+	// Determine the type of controller for the resource: TF, DCL, or Direct.
+	crd, err := crdloader.GetCRDForGVK(rc.ResourceGVK)
+	if err != nil {
+		return ResourceContext{}, err
+	}
+	if crd.GetLabels()[crdgeneration.TF2CRDLabel] == "true" {
```
justinsb, line 99: I think we've merged the SQLInstance clone support... should we be applying the predicates here?

(A fine answer is "yes, but first we want to fix a lot of the tests")

justinsb:
One question but LGTM.  Will apply hold, but feel free to remove

/approve
/lgtm
/hold

pkg/test/resourcefixture/contexts/register.go
jasonvigil: yes, but first we want to fix a lot of the tests

jasonvigil: 😄

PR #2591, opened 2024-08-30
tests: capture streaming DataflowFlexTemplateJob update example

justinsb:
The other example does not seem to be updateable.


yuwenma (commented):
/lgtm


One comment regarding the default setting.

mockgcp/mockpubsub/subscription.go
```diff
@@ -64,6 +64,9 @@ func (s *subscriberService) CreateSubscription(ctx context.Context, req *pb.Subs
 }
 
 func (s *subscriberService) populateDefaultsForSubscription(obj *pb.Subscription) {
+	if obj.AckDeadlineSeconds == 0 {
```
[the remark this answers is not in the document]
justinsb: I moved it to only be on the create method.  I _think_ it applies to the update method, but I think we only have evidence of it being needed in our tests on the create method, so far!

pkg/test/resourcefixture/testdata/basic/dataflow/v1beta1/dataflowflextemplatejob/streamingdataflowflextemplatejob/_broken_update.yaml
```diff
@@ -1,3 +1,6 @@
+# This job does not appear to actually be updateable, so masking out the update operation.
```
yuwenma, line 1: 👍

pkg/test/resourcefixture/testdata/basic/dataflow/v1beta1/dataflowflextemplatejob/streamingdataflowflextemplatejob/_http.log
yuwenma, line 1: I would need some help on reviewing the golden http changes, because normally it shall only reflects the changes in the golden object and create/update files. Are the other changes in this file actually come from real gcp ?

tests/e2e/normalize.go
```diff
@@ -562,7 +579,8 @@ func normalizeHTTPResponses(t *testing.T, events test.LogEntries) {
 		visitor.replacePaths[".job.startTime"] = "2024-04-01T12:34:56.123456Z"
 		visitor.replacePaths[".job.createTime"] = "2024-04-01T12:34:56.123456Z"
 		visitor.replacePaths[".currentStateTime"] = "2024-04-01T12:34:56.123456Z"
-		visitor.sortSlices.Insert(".environment.experiments")
+		visitor.sortAndDeduplicateSlices.Insert(".environment.experiments")
+		visitor.sortAndDeduplicateSlices.Insert(".environment.sdkPipelineOptions.options.experiments")
```
yuwenma, line 583: Can we do sorting in the MockGCP side?

justinsb: We also need to sort the GCP output though, it currently is randomly ordered I believe

pkg/test/resourcefixture/testdata/basic/dataflow/v1beta1/dataflowflextemplatejob/streamingdataflowflextemplatejob/_http.log
justinsb: This _should_ be the real GCP output (although the pipelineUrl probably needs normalization) , I can run again and confirm.

In the long run we should be getting the golden output from our prow runs.

yuwenma: Thanks! Yeah, I like the direction of using prow runs.

justinsb: Verified that the only differences when running `hack/record-gcp` are on the GCS bucket expiry time and on the pipelineUrl.  The pipelineUrl is quite tricky because it's effectively random, so I'll add a normalization.

justinsb:
I added the pipelineUrl normalization (so I lost lgtm/approved!)

PR #2594, opened 2024-08-30
fix: Delete mysql root user after creation

jasonvigil:
MySQL requires that the host is passed-in along with the username. Otherwise, the delete request will fail.


yuwenma (commented):
/lgtm
/approve

Looks great!

PR #2598, opened 2024-08-30
tool: mark resourceID as immutable field in template

jingyih:
In controller builder template, use CEL to validate `resourceID` as immutable field.


yuwenma (commented):
/lgtm
/approve

Thank you!

PR #2599, opened 2024-08-30
direct: Generate crd and upload status.externalRef

gemmahou:
### Change description

Generated CRD and populate status.externalRef in golden log

### Tests you have done

### DIFF
--- a/dev/tasks/generate-crds
+++ b/dev/tasks/generate-crds
@@ -55,7 +55,7 @@ go run ./scripts/crd-tools reflow-descriptions --dir apis/config/crd/
 
 # excluded_resources are resources that are under development for a direct conversion
 # we don't modify the CRD just yet for those but will in the future
-excluded_resources=("computeforwardingrule" "bigquerydatatransferconfig")
+excluded_resources=("bigquerydatatransferconfig")
 
 cd ${REPO_ROOT}
 cd apis/config/crd/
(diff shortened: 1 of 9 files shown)


dev/tasks/generate-crds
```diff
@@ -55,7 +55,7 @@ go run ./scripts/crd-tools reflow-descriptions --dir apis/config/crd/
 
 # excluded_resources are resources that are under development for a direct conversion
 # we don't modify the CRD just yet for those but will in the future
-excluded_resources=("computeforwardingrule")
+#excluded_resources=("computeforwardingrule")
```
yuwenma, line 58: If the resource is ready, maybe not place it inside `excluded_resources` just in case it's turned back off by accident?

pkg/controller/direct/compute/forwardingrule_externalresource.go
```diff
@@ -53,14 +53,14 @@ func asID(externalRef string) (*ForwardingRuleIdentity, error) {
 	}
 	path := strings.TrimPrefix(externalRef, serviceDomain+"/")
 	tokens := strings.Split(path, "/")
-	if len(tokens) == 5 || tokens[0] == "projects" || tokens[2] == "global" || tokens[3] == "forwardingrules" {
+	if len(tokens) == 5 && tokens[0] == "projects" && tokens[2] == "global" && tokens[3] == "forwardingrules" {
```
yuwenma, line 56: Good catch!

yuwenma (commented):
/lgtm
/approve

/hold
one minor but no blockers. Feel free to resolve in follow up PRs.

dev/tasks/generate-crds
gemmahou: Thanks! I made this a part of the comment, so people have an example on how to use this flag when working on direct controllers.

```
# excluded_resources are resources that are under development for a direct conversion
# we don't modify the CRD just yet for those but will in the future
# i.e. excluded_resources=("computeforwardingrule")
```

yuwenma (commented):
/lgtm

PR #2601, opened 2024-08-31
fix: Skip testNoChangeAfterUpdate if update resource is empty

jasonvigil:
This issue was never hit before because testNoChangeAfterUpdate is currently only enabled for SQLInstances resources. We recently added a SQLInstance test case that does not set an update.yaml resource (sqlinstance-clone-minimal), and it revealed the problem causing a nil pointer dereference when getting update object.

### DIFF
--- a/pkg/controller/dynamic/dynamic_controller_integration_test.go
+++ b/pkg/controller/dynamic/dynamic_controller_integration_test.go
@@ -296,6 +296,15 @@ func testNoChangeAfterCreate(ctx context.Context, t *testing.T, testContext test
 
 // testNoChangeAfterUpdate is enabled only on resources allowlisted inside the function.
 func testNoChangeAfterUpdate(ctx context.Context, t *testing.T, testContext testrunner.TestContext, systemContext testrunner.SystemContext, resourceContext contexts.ResourceContext) {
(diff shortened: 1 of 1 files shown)


gemmahou:
We can also add `SkipUpdate: true` in sql_context for this [test](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/4d839f272139da0228ee07f58546fcb78001451f/pkg/test/resourcefixture/contexts/sql_context.go#L27), when there's no update.yaml

jasonvigil:
> We can also add `SkipUpdate: true`([example](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/4d839f272139da0228ee07f58546fcb78001451f/pkg/test/resourcefixture/contexts/sql_context.go#L96)) in sql_context for this [test](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/4d839f272139da0228ee07f58546fcb78001451f/pkg/test/resourcefixture/contexts/sql_context.go#L27), when there's no update.yaml

Alright, updated! I like having the nil check also, to prevent from accidental wide breakages like this (where we block other tests from running due to the nil pointer dereference panic).

yuwenma (commented):
/lgtm
/approve

Makes sense. Thanks!

PR #2605, opened 2024-09-01
Compositions: Fix manual install version

barney-s:
### Change description

Manual install version had a `v` prefix which our release images does not have.


cheftako:
Looks like we have a lint issue we need to look into

PR #2607, opened 2024-09-02
tests: update golden output for dataflowflextemplate jobs

justinsb:


justinsb:
It turned out this was #2608  - skew in the direct reconciler lists

/close

PR #2608, opened 2024-09-03
refactor: have compare-mock call into run-e2e

justinsb:
This avoids another copy of the direct-reconciler list, with the accompanying possibilities for skew.


yuwenma (commented):
/lgtm
/approve

PR #2613, opened 2024-09-03
tool: remove unnecessary optional annotation in template

jingyih:
In controller builder template, remove unnecessary `// +optional` annotations in resource `Status`.


yuwenma:
/lgtm
/approve

Thanks!

PR #2614, opened 2024-09-03
Fix presubmit samples

gemmahou:
### Change description

Tracked [commit history](https://screenshot.googleplex.com/8xPagSCmdGNoxkz.png), the samples tests failed probably related to the recent refactors.

Two refactoring PRs caught my eye: #2604 and #2608

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #2583, opened 2024-08-29
MockGCP: Support for ComputeServiceAttachment

gemmahou:
### Change description

Add mock support for ComputeServiceAttachment

Cherry-picked code from https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/936

Collect logs as well, using KCC_USE_DIRECT_RECONCILERS="ComputeForwardingRule"

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #2620, opened 2024-09-04
Revert "MockGCP: Support for ComputeServiceAttachment"

gemmahou:
Reverts GoogleCloudPlatform/k8s-config-connector#2583


justinsb:
Why do we want to revert?  Can you link a GitHub issue or similar?

gemmahou:
> Why do we want to revert? Can you link a GitHub issue or similar?

The presubmit samples failed after this commit: https://screenshot.googleplex.com/8cgjE9ChvekNQVp.png, job: https://github.com/GoogleCloudPlatform/k8s-config-connector/actions/runs/10681623224/job/29605684721, and that blocks the open PRs.

The two failed tests are samples/regional-forwarding-rule-vpc-psc and samples/computeserviceattachment(flaky).

Checking the logs, the forwarding rule is a dependency of the service attachment, and the deletion of the forwarding rule is blocked because there is no network (the network was deleted before). It seems like we need to handle the deletion ordering in the direct controller. It's strange that the presubmit on the original PR passed. I had a temporary fix: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2614, but the two tests are still failing, so I think there's a deeper issue (deletion ordering, etc.).

justinsb:
Maybe we can just revert the ComputeForwardingRule support in `func MaybeSkip` instead of a full revert?

justinsb:
And on the deeper fix, it sounds like you're saying that the old controller was able to tolerate the network not existing, and the new controller is not.  I think this is because we resolve all dependencies in AdapterForObject.  Maybe in practice we don't want to do that for a Delete.  Either we have a dedicated method/path (e.g. `AdapterForObjectDeletion`) or we pass a parameter like `delete bool` to `AdapterForObject`, as we know in this case we don't need to do any resolution other than for the "primary key" (selfLink) of the object.

Issue #2402, opened 2024-08-01
resource-stats-recorder complains about missing v1alpha1 resources


tsawada:
### Checklist

- [X] I did not find a related open issue.
- [X] I did not find a solution in the troubleshooting guide: (https://cloud.google.com/config-connector/docs/troubleshooting)
- [X] If this issue is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Bug Description

After updating Config Connector from v1.102.0 to v1.120.1, we get following ERROR logs for all kinds of v1alpha1 cnrm resources, a few hundreds / minute:

```
setup "msg"="error recording metrics for CRD %v: %v" "error"="error listing objects for compute.cnrm.cloud.google.com/v1alpha1, Kind=ComputeRegionPerInstanceConfig: error listing objects: no matches for kind \"ComputeRegionPerInstanceConfig\" in version \"compute.cnrm.cloud.google.com/v1alpha1\"" "compute.cnrm.cloud.google.com/v1alpha1, Kind=ComputeRegionPerInstanceConfig"="(MISSING)"
```

We are not sure if this actually indicates any breakage. If not, we prefer to have less ERROR logs so that we can notice real errors easier.

### Additional Diagnostic Information

None

### Kubernetes Cluster Version

v1.28.0

### Config Connector Version

1.120.1

### Config Connector Mode

namespaced mode (default)

### Steps to reproduce the issue

Update to v1.120.1

cheftako:
In which log are you seeing this?

tsawada:
Thanks. stderr of a pod `cnrm-resource-stats-recorder-***-**`, container `recorder`.

justinsb:
Should be fixed by #2631, which should ship in 1.122

/close

PR #2631, opened 2024-09-05
recorder: watch CRDs instead of using a static list

justinsb:
Otherwise we look for CRDs that are not installed.

Issue #2402

### DIFF
--- a/cmd/recorder/main.go
+++ b/cmd/recorder/main.go
@@ -21,28 +21,29 @@ import (
 	"net/http"
 	_ "net/http/pprof" // Needed to allow pprof server to accept requests
 	"os"
+	"strings"
 	"time"
 
-	dclmetadata "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/dcl/metadata"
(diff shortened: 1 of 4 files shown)


cmd/recorder/main.go
```diff
@@ -106,34 +122,52 @@ func main() {
...
+	crdGVR := schema.GroupVersionResource{Group: "apiextensions.k8s.io", Version: "v1", Resource: "customresourcedefinitions"}
+	crdInfos := kube.WatchKube(ctx, kubeTarget, crdGVR, buildCRDInfo)
+
 	for {
-		err := doRecord(c, supportedGVKs)
-		if err != nil {
-			logger.Error(err, "error recording metrics.")
-		}
 		time.Sleep(time.Duration(metricInterval) * time.Second)
+		if !crdInfos.HasSyncedOnce() {
```
yuwenma, line 151: 👍 Ah!

yuwenma (commented):
/lgtm
/approve

PR #2632, opened 2024-09-05
Add full test and minimal test for Certificate Manager DNS Authorization

anhdle-sso:
This is the 1st PR; create.yaml and update.yaml have the same content

### Change description

Fixes #

### Tests you have done

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/certificatemanager/v1beta1/certificatemanagerdnsauthorization-full/update.yaml
+++ b/pkg/test/resourcefixture/testdata/basic/certificatemanager/v1beta1/certificatemanagerdnsauthorization-full/update.yaml
@@ -1,0 +1,24 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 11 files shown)


pkg/test/resourcefixture/testdata/basic/certificatemanager/v1beta1/certificatemanagerdnsauthorization-full/update.yaml
```diff
@@ -0,0 +1,24 @@
+# Copyright 2024 Google LLC
```
gemmahou, line 1: Is Update supported for this resource? If yes, we need to modify the mutable fields or labels in update.yaml. If not then we do not need update.yaml, and we can create test context for this resource and add `SkipUpdate: true` in the test context, similar to https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/8cc82c9173f45ea06d1ee718a5aceb2540b1cfed/pkg/test/resourcefixture/contexts/alloydb_context.go#L21

gemmahou: nvm, per the manual the 1st PR should have the same create and update yaml.

anhdle-sso: This is the first PR that would have create and update the same. The second PR would have update.yaml changes every mutable fields; along with golden log/object change to reflect that

PR #2633, opened 2024-09-05
feat: add mappings for BigQueryDataTransferConfig

jingyih:
This PR is also making some improvements to the mapper generator.


pkg/controller/direct/maputils.go
```diff
@@ -137,11 +137,26 @@ func LazyPtr[V comparable](v V) *V {
 	return &v
 }
 
-func ToOpenAPIDateTime(ts *timestamppb.Timestamp) *string {
+func StringTimestamp_FromProto(mapCtx *MapContext, ts *timestamppb.Timestamp) *string {
+	if ts == nil {
```
jingyih, line 141: It seems that this nil check causes some differences in the cloudbuildworkerpool golden object. However, I believe the check is necessary to protect the function from a nil input. A nil input timestamp would default to the Unix epoch, which is probably unintended?

Looking at the http log, it does seem that "createTime" is missing after a PATCH call.

```
diff --git a/pkg/test/resourcefixture/testdata/basic/cloudbuild/v1beta1/cloudbuildworkerpool/_generated_object_cloudbuildworkerpool.golden.yaml b/pkg/test/resourcefixture/testdata/basic/cloudbuild/v1beta1/cloudbuildworkerpool/_generated_object_cloudbuildworkerpool.golden.yaml
index 5e803a477..abb95f4df 100644
--- a/pkg/test/resourcefixture/testdata/basic/cloudbuild/v1beta1/cloudbuildworkerpool/_generated_object_cloudbuildworkerpool.golden.yaml
+++ b/pkg/test/resourcefixture/testdata/basic/cloudbuild/v1beta1/cloudbuildworkerpool/_generated_object_cloudbuildworkerpool.golden.yaml
@@ -35,7 +35,6 @@ status:
   externalRef: //cloudbuild.googleapis.com/projects/${projectId}/locations/us-central1/workerPools/cloudbuildworkerpool-${uniqueId}
   observedGeneration: 2
   observedState:
-    createTime: "1970-01-01T00:00:00Z"
     etag: abcdef123456
     networkConfig:
       egressOption: NO_PUBLIC_EGRESS
```

yuwenma: Make sense! Could you fix the cbwp log (or I can do that)?

jingyih: I did not dig into the mock implementation to figure out why createTime was missing.

jingyih: Fixed in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2656

jingyih:
Rebased to pick up https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2656

yuwenma (commented):
/lgtm
/approve

PR #2637, opened 2024-09-06
fix: improve MockPusub fidelity on AllowedPersistenceRegions

yuwenma:
The PubSub is a dependency to many resources. This `AllowedPersistenceRegions` is trivial but has bothered me every time using Pubsub as a dependency. It is just easier to fix it now than later as we grow the mockgcp coverage.
The outcome is verified by comparing the real GCP log with the Mock one in MockSecretManager.

### DIFF
--- a/mockgcp/mockpubsub/topic.go
+++ b/mockgcp/mockpubsub/topic.go
@@ -53,35 +53,32 @@ func (s *publisherService) CreateTopic(ctx context.Context, req *pb.Topic) (*pb.
 }
 
 func (s *publisherService) populateDefaultsForTopic(obj *pb.Topic) {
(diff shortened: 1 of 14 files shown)


justinsb:
Something odd is happening here, because I'm _not_ seeing this field any more.  It could be something to do with projects...  Also I don't _think_ our real/internal tests have this field populated

yuwenma:
This is an interesting case and we probably will hit it in the future on different resources. 
The `allowedPersistenceRegion` from the PubSub service response could bind with the GCP project policies, which could be inherited from its ancestor according to its resource-hierarchy. So the real GCP outcome could be different depending on which org/folder policy it runs.

PR #2647, opened 2024-09-06
fix: Skip checking for diffs in resource reference fields

jasonvigil:
There is no way to check for diffs in these fields, because export can only return the external reference, not the in-cluster object name/namespace.

### DIFF
--- a/pkg/controller/dynamic/dynamic_controller_integration_test.go
+++ b/pkg/controller/dynamic/dynamic_controller_integration_test.go
@@ -799,7 +799,11 @@ func getChangedFields(initialObject, updatedObject map[string]interface{}, field
 		for k, v := range updated {
 			if !reflect.DeepEqual(initial[k], v) {
 				if _, ok := v.(map[string]interface{}); ok {
-					changedFields[k] = getChangedFields(initial, updated, k)
+					// Skip checking for changes in resource reference fields, because there
+					// is no way to export the name and namespace fields (only external).
+					if !strings.HasSuffix(k, "Ref") {
+						changedFields[k] = getChangedFields(initial, updated, k)
+					}
 				} else {
 					changedFields[k] = updated[k]
 				}


pkg/controller/dynamic/dynamic_controller_integration_test.go
```diff
@@ -799,7 +799,11 @@ func getChangedFields(initialObject, updatedObject map[string]interface{}, field
 		for k, v := range updated {
 			if !reflect.DeepEqual(initial[k], v) {
 				if _, ok := v.(map[string]interface{}); ok {
-					changedFields[k] = getChangedFields(initial, updated, k)
+					// Skip checking for changes in resource reference fields, because there
```
justinsb, line 802: What if we normalize to "external form" ?

jasonvigil: In this case, that won't really help. The test checks what is applied (to k8s api) vs the exported resource in gcp. If we normalize internally to use the "external" form, the test will still fail, because it cannot find the "changed" field (`*Ref.name`) in the exported resources (that only have `*Ref.external`).

justinsb: Oh I see, this is our tests!

justinsb:
/approve
/lgtm


@jasonvigil is this test functionality covered by unified_test.go?  If not, mind explaining what it is doing?

jasonvigil:
@justinsb the test checks for differences between the "create.yaml" and "update.yaml" spec fields. Then, after update, it performs an `Export()` from GCP and verifies that the exported GCP resource contains the updated values.

I think unified_test.go "sort of" covers this functionality, but in a manual way. We can inspect the `_http.log` from unified_test.go to verify that the returned object is correct. However, there is no explicit test for the `Export()` functionality in unified_test.go (as far as I know).

jingyih:
> because export can only return the external reference, not the in-cluster object name/namespace.

Hi @jasonvigil, for my own education purpose, do you know if this behavior is SciFi-implementation specific?

jasonvigil:
@jingyih I think it's a problem with both direct and TF/DCL. The `Export()` function has no knowledge of the k8s cluster that KCC is running in, so it cannot "reconstruct" the K8s resources. This issue is probably only appearing now because we previously did not have any test cases for modifying resource reference fields.

PR #2648, opened 2024-09-08
tests: turn off VCR tests for SQLUser

justinsb:
The tests are currently failing and it probably needs a re-record.


justinsb:
Example of failures is on https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2645

yuwenma:
/lgtm
/approve

`sqluser` is still a TF-based controller so I think the VCR would be beneficial. @gemmahou Could you help re-record the test?

PR #2651, opened 2024-09-09
fix: Update sqlinstance-authorizednetworks timestamp format

jasonvigil:
Use the GCP-preferred format.


yuwenma (commented):
/lgtm
/approve

Thanks

PR #2656, opened 2024-09-09
mockgcp: improve cloud build worder pool

jingyih:
When LRO is finished, we should update the resource object in mockgcp storage.

I checked against real GCP, a `GET` call on Cloud Build Worker Pool does include "createTime, etag, state" in the response.


yuwenma (commented):
/lgtm
/approve
Thanks

PR #2658, opened 2024-09-09
Add test case for SQLInstance location preference

jasonvigil:
Update locationPreference mapping code to improve readability, make similar to generated mapping code.
    
Also, update ipConfiguration export code (this was missed before).


yuwenma (commented):
/lgtm
/approve

PR #2659, opened 2024-09-10
Support PrivilegedAccessManagerEntitlement via the direct controller

maqiuyujoyce:
### Change description

Added the CRD schema, mapper, direct controller and basic test case for PrivilegedAccessManagerEntitlement.

### Tests you have done

Verified locally via dynamic test that the new test case works.


maqiuyujoyce:
@yuwenma WIP PR for PrivilegedAccessManagerEntitlement schema.

apis/privilegedaccessmanager/v1beta1/doc.go
yuwenma, line 1: We need to start from Alpha resource unless this is already in Beta. See https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/develop-resources/scenarios/new-resource.md#pr-reviews-2

apis/privilegedaccessmanager/v1beta1/privilegedaccessmanagerentitlement_types.go
```diff
@@ -0,0 +1,315 @@
...
+	// +required
+	Location *string `json:"location"`
+
+	// The PrivilegedAccessManagerEntitlement name. If not given, the
+	// 'metadata.name' will be used.
+	// +optional
+	ResourceID *string `json:"resourceID,omitempty"`
+
+	// Optional. Who can create grants using this entitlement. This list should
+	// contain at most one entry.
```
yuwenma, line 59: Do you mean this accepts a list with no more than one value?

apis/privilegedaccessmanager/v1beta1/privilegedaccessmanagerentitlement_types.go
```diff
@@ -0,0 +1,315 @@
...
+//
+// * Require an approval from any user in a group G.
+// * Require an approval from any k number of users from a Group G.
+// * Require an approval from any user in a group G and then from a user U.
+//
+// A single user might be part of the 'approvers' ACL for multiple steps in this
+// workflow, but they can only approve once and that approval is only considered
+// to satisfy the approval step at which it was granted.
+type ManualApprovals struct {
+	// Optional. Do the approvers need to provide a justification for their
```
yuwenma, line 121: Suggest removing this question since this is user facing comment.

apis/refs/v1beta1/resourceRef.go
```diff
@@ -0,0 +1,38 @@
...
+	// by Config Connector.
+	// This field must be set when 'name' field is not set.
+	// +optional
+	External string `json:"external,omitempty"`
+	// The apiVersion (<group>/<version>) of the referenced resource.
+	// +required
+	ApiVersion string `json:"apiVersion,omitempty"`
+	// The kind of the referenced resource.
+	// +required
+	Kind string `json:"kind,omitempty"`
```
yuwenma, line 28: We do not want to support generic Kind. And maybe we don't need this field entirely. See my other comment about using KCC IAM resources

apis/privilegedaccessmanager/v1beta1/privilegedaccessmanagerentitlement_types.go
```diff
@@ -0,0 +1,315 @@
...
+	// pending approval.
+	// +optional
+	ApproverEmailRecipients *[]string `json:"approverEmailRecipients,omitempty"`
+}
+
+// Privileged access that this service can be used to gate.
+type PrivilegedAccess struct {
+	// Access to a Google Cloud resource through IAM.
+	// +required
+	GcpIAMAccess GcpIamAccess `json:"gcpIAMAccess,omitempty"`
```
yuwenma, line 155: Shall this be a ref to IAMPolicy?

apis/refs/v1beta1/organizationref.go
```diff
@@ -0,0 +1,108 @@
...
+
+// OrganizationRef represents the Organization that this resource belongs to.
+type OrganizationRef struct {
+	// The 'name' field of an organization, when not managed by Config Connector.
+	// +required
+	External string `json:"external,omitempty"`
+	// The kind of the Organization resource; optional but must be
+	// 'Organization' if provided.
+	// +optional
+	Kind string `json:"kind,omitempty"`
```
yuwenma, line 36: We do not need the kind because this value will only be Organization.

apis/refs/v1beta1/folderref.go
```diff
@@ -0,0 +1,176 @@
...
+	Name string `json:"name,omitempty"`
+	// The 'namespace' field of a 'Folder' resource.
+	// If unset, the namespace is defaulted to the namespace of the referencer
+	// resource.
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
+	// The kind of the Folder resource; optional but must be 'Folder' if
+	// provided.
+	// +optional
+	Kind string `json:"kind,omitempty"`
```
yuwenma, line 49: ditto

apis/privilegedaccessmanager/v1beta1/privilegedaccessmanagerentitlement_types.go
maqiuyujoyce: Looks like so: https://cloud.google.com/iam/docs/reference/pam/rest/v1beta/organizations.locations.entitlements#resource:-entitlement

apis/privilegedaccessmanager/v1beta1/privilegedaccessmanagerentitlement_types.go
maqiuyujoyce: Will rephrase it. This field description comes directly from the API: https://cloud.google.com/iam/docs/reference/pam/rest/v1beta/folders.locations.entitlements#Entitlement.ManualApprovals - do we have any guideline on what is a good field description and what is not?

apis/privilegedaccessmanager/v1beta1/privilegedaccessmanagerentitlement_types.go
maqiuyujoyce: I don't think so. This resource is technically a IAMPolicyMember resource with additional restrictions. Referencing an IAMPolicy here means that the user need the IAM policy and a PAM entitlement containing the same policy with some restrictions. But these two resources have mutually exclusive functionatlities: If the policy can be configured, then the entitlement is not necessary; and if the entitlement is needed, then they shouldn't configure the policy.

apis/privilegedaccessmanager/v1beta1/doc.go
maqiuyujoyce: Done.

apis/refs/v1beta1/resourceRef.go
maqiuyujoyce: Used three different refs to represent this field instead.

apis/refs/v1beta1/organizationref.go
maqiuyujoyce: Done.

apis/refs/v1beta1/folderref.go
maqiuyujoyce: Done.

pkg/controller/direct/privilegedaccessmanager/entitlement_controller.go
```diff
@@ -0,0 +1,502 @@
...
+}
+
+func (m *entitlementModel) AdapterForObject(ctx context.Context, reader client.Reader, u *unstructured.Unstructured) (directbase.Adapter, error) {
+
+	obj := &krm.PrivilegedAccessManagerEntitlement{}
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &obj); err != nil {
+		return nil, fmt.Errorf("error converting to %T: %w", obj, err)
+	}
+
+	// Get ResourceID
```
yuwenma, line 92: Some general suggestion from ln92-215 and external ref:
We can treat projectID, orgID and folderID the same thing in externalRef

pseudo code
```go
func oneOfParent(obj) (string, error) {
   p1 := obj.Spec.ProjectRef
   p2 := obj.Spec.FolderRef
   p3 := obj.Spec.OrgRef 
   if p1 == nil &&  p2 == nil && p3 ==nil {
      return "",  fmt.Errorf()
   }
   if p2 == nil && p3 ==nil {
        project := ResolveProject(p1)
        return "projects/" + project.ID,   nil
   }
   if p1 == nil && p3 ==nil {
        folder := ResolveFolder(p2)
        return "folders/" + folder.ID, nil
   }
   if p1 ==nil && p2 == nil {
        org := ResolveOrg(p3)
         return "organizations/" + org.ID, nil
   }
   return "", fmt.Errorf()
}

func AdapterForObject() {
    parent, err := oneOfParent(obj)
    if ... {
         buildID(parent, resourceID) 
    } else {
          asID(parent)
    }
}

func buildID(parent, resourceID string) {
      return ID{
           Parent: parent,
           ResourceID: resourceID
       }
}

func ValidateExternal(obj *krm.PAM, externalRef *ExternalRef) error {
      parent, err := oneOfParent(obj)
      if parent != externalRef.Parent {
          return err
      }
}
 

```

dev/tools/controllerbuilder/pkg/codegen/mappergenerator.go
yuwenma, line 1: Suggest putting this in a separate PR so it can benefit others sooner

apis/refs/v1beta1/folderref.go
yuwenma, line 1: Thanks for adding these two refs! If you like, we can also put that in separate PRs

apis/privilegedaccessmanager/v1alpha1/types.generated.go
```diff
@@ -0,0 +1,195 @@
...
+	// One and only one of 'projectRef', 'folderRef', or 'organizationRef' must
+	// be set.
+	// +optional
+	FolderRef *refs.FolderRef `json:"folderRef,omitempty"`
+
+	// The Organization that this privileged access is granted to.
+	// One and only one of 'projectRef', 'folderRef', or 'organizationRef' must
+	// be set.
+	// +optional
+	OrganizationRef *refs.OrganizationRef `json:"organizationRef,omitempty"`
```
yuwenma, line 169: Do you know what is the relationship between `spec.projectRef` and `spec. privilegedAccess. gcpIAMAccess.projectRef`?

pkg/gvks/supportedgvks/supportedgvks.go
yuwenma, line 1: Suggest moving non PAM code in a separate PR.

pkg/test/resourcefixture/testdata/basic/privilegedaccessmanager/v1alpha1/privilegedaccessmanagerentitlement/_generated_object_privilegedaccessmanagerentitlement.golden.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: privilegedaccessmanager.cnrm.cloud.google.com/v1alpha1
```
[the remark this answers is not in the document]
maqiuyujoyce: Yes. `spec.approvalWorkflow` and `spec.additionalNotificationTargets` are optional. Any reason you asked this question?

pkg/test/constants/presubmitconstants.go
```diff
@@ -29,66 +29,67 @@ var (
 	// When adding a new service/updating this list, prioritize test cases with more
 	// dependencies for more resource coverage in our presubmit-lite test.
 	RepresentativeCRUDTestsForAllServices = map[string][]string{
```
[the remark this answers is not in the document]
maqiuyujoyce: Unit tests will fail. We should remove the unit test as it is kinda useless now; but I prefer we do it in a separate PR.

apis/privilegedaccessmanager/v1alpha1/types.generated.go
maqiuyujoyce: Right now they must be the same, but they also must both be specified.

I was thinking of hiding this field and generate the value for the users, but I'm concerned that the service team extends this field's behavior unexpectedly so decided to expose the exact same interface as the underlying GCP API.

Overall this resource was designed weirdly.

pkg/gvks/supportedgvks/supportedgvks.go
maqiuyujoyce: This is actually done in #2664; added here because it is a prerequisite.

dev/tools/controllerbuilder/pkg/codegen/mappergenerator.go
maqiuyujoyce: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2805

pkg/controller/direct/privilegedaccessmanager/entitlement_externalresource.go
```diff
@@ -0,0 +1,93 @@
...
+package privilegedaccessmanager
+
+import (
+	"fmt"
+	"strings"
+)
+
+// The Identifier for Config Connector to track the PrivilegedAccessManagerEntitlement resource from the GCP service.
+type PrivilegedAccessManagerEntitlementIdentity struct {
+	Parent      parent
```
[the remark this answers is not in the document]
maqiuyujoyce: Done.

pkg/controller/direct/privilegedaccessmanager/entitlement_mappings.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+	case "cloudresourcemanager.googleapis.com/Folder":
+		out.FolderRef = &refs.FolderRef{
+			External: strings.TrimPrefix(resource, "//cloudresourcemanager.googleapis.com/"),
+		}
+	case "cloudresourcemanager.googleapis.com/Organization":
+		out.OrganizationRef = &refs.OrganizationRef{
+			External: strings.TrimPrefix(resource, "//cloudresourcemanager.googleapis.com/"),
+		}
+	}
```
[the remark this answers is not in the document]
maqiuyujoyce: Done.

pkg/controller/direct/privilegedaccessmanager/entitlement_controller.go
maqiuyujoyce: Good point. Done.

pkg/test/resourcefixture/testdata/basic/privilegedaccessmanager/v1alpha1/privilegedaccessmanagerentitlement/_generated_object_privilegedaccessmanagerentitlement.golden.yaml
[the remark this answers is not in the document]
yuwenma: Yes, we want to distinguish basic and full so that it is more straightforward to use as reference. You can see more detailed guide here https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/develop-resources/guides/1-add-mockgcp-tests.md#a-basic-test-suite

pkg/test/constants/presubmitconstants.go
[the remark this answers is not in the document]
yuwenma: +1 agree to do that in separate PR.

apis/privilegedaccessmanager/v1alpha1/types.generated.go
yuwenma: I see. Thanks for the explanation.

pkg/controller/direct/privilegedaccessmanager/entitlement_controller.go
```diff
@@ -0,0 +1,449 @@
...
+
+func (a *Adapter) Update(ctx context.Context, updateOp *directbase.UpdateOperation) error {
+	u := updateOp.GetUnstructured()
+
+	log := klog.FromContext(ctx).WithName(ctrlName)
+	log.V(2).Info("updating PrivilegedAccessManagerEntitlement", "name", a.id.FullyQualifiedName())
+	mapCtx := &direct.MapContext{}
+
+	updateMask := &fieldmaskpb.FieldMask{}
+
```
yuwenma, line 309: Suggest calling the `<Kind>Spec_ToProto` func instead. This makes the code easier to maintain and each file has a more dedicated usage, otherwise both Controller and Mapper needs to handle non-top level mappings.

pkg/controller/direct/privilegedaccessmanager/entitlement_controller.go
```diff
@@ -0,0 +1,449 @@
...
+	// Get ResourceID
+	resourceID := direct.ValueOf(obj.Spec.ResourceID)
+	if resourceID == "" {
+		resourceID = obj.GetName()
+	}
+	if resourceID == "" {
+		return nil, fmt.Errorf("cannot resolve resource ID")
+	}
+
+	container, err := oneOfContainer(ctx, reader, obj,
```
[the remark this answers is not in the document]
maqiuyujoyce: We need to pass in the exact fields because in this resource, there are two different oneOfContainer situations: one at the top level, and one as subfields under `spec.privilegedAccess.gcpIAMAccess`.

However, I plan to extract out the util functions to support the top level oneOfContainer situation later.

pkg/controller/direct/privilegedaccessmanager/entitlement_externalresource.go
```diff
@@ -0,0 +1,86 @@
...
+			"got %s", serviceDomain, serviceDomain, serviceDomain, externalRef)
+	}
+	p := parent{Location: tokens[3]}
+	if tokens[0] == "projects" {
+		p.Container = fmt.Sprintf("projects/%s", tokens[1])
+	} else if tokens[0] == "folders" {
+		p.Container = fmt.Sprintf("folders/%s", tokens[1])
+	} else if tokens[0] == "organizations" {
+		p.Container = fmt.Sprintf("organizations/%s", tokens[1])
+	}
```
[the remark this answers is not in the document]
maqiuyujoyce: Good catch! Simplified it.

pkg/controller/direct/privilegedaccessmanager/entitlement_controller.go
maqiuyujoyce: Done.

maqiuyujoyce:
This PR is blocked by https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2862.

maqiuyujoyce:
@justinsb I think I've addressed all your comments!

justinsb (approved):

PR #2660, opened 2024-09-10
fix: Update sqluser vcr

gemmahou:
### Change description

Fixes #2648 

### Tests you have done


pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqluser/_vcr_cassettes/tf.yaml
```diff
@@ -44,29 +44,29 @@ interactions:
...
-      content_length: 316
+      content_length: 339
       transfer_encoding: []
       trailer: {}
       host: sqladmin.googleapis.com
       remote_addr: ""
       request_uri: ""
       body: |
-        {"databaseVersion":"MYSQL_5_7","name":"sqluser-dep-3vxr1tn21m3yp","region":"us-central1","settings":{"activationPolicy":"ALWAYS","availabilityType":"ZONAL","dataDiskType":"PD_SSD","pricingPlan":"PER_USE","storageAutoResize":true,"tier":"db-n1-standard-1","userLabels":{"cnrm-test":"true","managed-by-cnrm":"true"}}}
+          {"databaseVersion":"MYSQL_5_7","name":"sqluser-dep-3vxr1tn21m3yp","region":"us-central1","settings":{"activationPolicy":"ALWAYS","availabilityType":"ZONAL","dataDiskType":"PD_SSD","edition":"ENTERPRISE","pricingPlan":"PER_USE","storageAutoResize":true,"tier":"db-n1-standard-1","userLabels":{"cnrm-test":"true","managed-by-cnrm":"true"}}}
```
gemmahou, line 63: This could be the only reason that caused the previous test to fail. The other updates are all in responses; we do not worry about response content with replay mode, but only check the request body, URL, and method.

yuwenma: that's fair

gemmahou: Noticed that after SQL Instance is switched from TF to direct controller, the sqlinstance related requests in the old vcr `tf.yaml` is invalid. We do need to re-collect logs and save the direct log(and DCL log) in a new folder `nontf.yaml`.

PR #2661, opened 2024-09-10
feat: add api for secretmanagersecret  @yuwenma

yuwenma:
Adds the API for SecretManagerSecret (Beta) resource.

This PR is based on #5903


config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_secretmanagersecrets.secretmanager.cnrm.cloud.google.com.yaml
```diff
@@ -308,32 +365,25 @@ spec:
                       type: string
                   type: object
                 type: array
-              createTime:
-                description: The time at which the Secret was created.
-                type: string
-              name:
```
justinsb, line 314: I think it's OK to drop create time, but `name` seems more likely to have some people reading it ... WDYT?

justinsb: My 2c: mark as deprecated, encourage people to read externalRef instead, say that we intend to remove `name` in the v1 API

pkg/test/resourcefixture/testdata/basic/secretmanager/v1beta1/basicsecretmanagersecret/create.yaml
```diff
@@ -24,7 +24,4 @@ spec:
   replication:
     userManaged:
       replicas:
-      - location: us-central1
-  topics:
```
justinsb, line 28: Can we keep topics?  Any reason to drop test coverage here?

justinsb: Ah - maybe because we moved it to the full test?

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_secretmanagersecrets.secretmanager.cnrm.cloud.google.com.yaml
```diff
@@ -243,48 +299,49 @@ spec:
...
               ttl:
-                description: |-
-                  Immutable. The TTL for the Secret.
-                  A duration in seconds with up to nine fractional digits, terminated by 's'. Example: "3.5s".
+                description: Input only. The TTL for the [Secret][google.cloud.secretmanager.v1.Secret].
                 type: string
               versionAliases:
                 additionalProperties:
-                  type: string
+                  format: int64
```
justinsb, line 322: So I think we've changed from `versionAliases map[string]string` to `versionAliases map[string]int` .  Can we add test coverage to check that we aren't breaking existing resources, in particular

````
versionAliases:
  foo: 1
```

and

````
versionAliases:
  foo: "1"
```

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_secretmanagersecrets.secretmanager.cnrm.cloud.google.com.yaml
yuwenma: Marking it deprecation works for me. I assume people uses KCC mostly via GitOps, so the live status matters mostly when something goes wrong (and `name` doesn't weigh in too much unless it's server generated). But I like the idea of marking it deprecated, since this is a Beta resource

pkg/test/resourcefixture/testdata/basic/secretmanager/v1beta1/basicsecretmanagersecret/create.yaml
yuwenma: Right. Topic is an optional field. I recommend only put required fields in the basic test suite, in which way it is easier to tell how to configure a field.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_secretmanagersecrets.secretmanager.cnrm.cloud.google.com.yaml
yuwenma: Sure!

yuwenma (commented):
Thanks for the review! @justinsb resolved all comments

justinsb (approved):

PR #2664, opened 2024-09-10
Add GVKs supported by direct controllers to supportedgvks.All()

### LINKED ISSUE #2646: KCC controller crashloop due to missing permission
In a new KCC cluster with alpha resources installed, I'm noticing the `cnrm-unmanaged-detector` pod is crash-looping due to the following error. 

```
pkg/mod/k8s.io/client-go@v0.27.11/tools/cache/reflector.go:231: Failed to watch *v1.PartialObjectMetadata: failed to list *v1.PartialObjectMetadata: securesourcemanagerinstances.securesourcemanager.cnrm.cloud.google.com is forbidden: User "system:serviceaccount:cnrm-system:cnrm-unmanaged-detector" cannot list resource "securesourcemanagerinstances" in API group "securesourcemanager.cnrm.cloud.google.com" at the cluster scope
```

It looks like permission on `securesourcemanagerinstances.securesourcemanager.cnrm.cloud.google.com` is missing in both [cnrm_admin.yaml ](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/f4cfe0adcb96052c76e54f87b538df3e746d0142/config/installbundle/components/clusterroles/cnrm_admin.yaml) and [cnrm_viewer.yaml ](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/f4cfe0adcb96052c76e54f87b538df3e746d0142/config/installbundle/components/clusterroles/cnrm_viewer.yaml). It is possible that some other permission is missing as well.

I am not entirely sure, but we probably need to update the [script](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/f4cfe0adcb96052c76e54f87b538df3e746d0142/scripts/generate-cnrm-cluster-roles/main.go#L62) to include SciFi resources?

maqiuyujoyce:
### Change description

Fixes #2646

* Added direct GVKs into supportedgvks.All() function so that reference doc generation and cnrm role generation logic can cover direct GVKs.
   * Added supportedgvks.AllWithoutDirect() to handle the situations when direct GVKs are not necessary. We should avoid calling supportedgvks.All() repetitively, and we should avoid calling it during resource reconciliation. supportedgvks.All() loops through all the CRDs (i.e. 400+ files) and it is time consuming.
* Updated the reference doc generation logic to make it work for direct GVKs.

I think we should enforce samples and reference doc template work for Scifi as well. 

### Tests you have done

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,7 +70,6 @@ manifests: generate
 	rm -rf config/crds/resources
 	rm -rf config/crds/tmp_resources
 	go build -o bin/generate-crds ./scripts/generate-crds && ./bin/generate-crds -output-dir=config/crds/tmp_resources
-	go run ./scripts/generate-cnrm-cluster-roles/main.go
 	# add kustomize patches on all CRDs
 	mkdir config/crds/resources
 	cp config/crds/kustomization.yaml kustomization.yaml
@@ -81,6 +80,10 @@ manifests: generate
 
 	# for direct controllers
 	dev/tasks/generate-crds
+
+	# Generating cnrm cluster roles is dependent on the existence of directory
+	# config/crds/resources with all the freshly generated CRDs.
+	go run ./scripts/generate-cnrm-cluster-roles/main.go
 
 # Format code
 .PHONY: fmt
--- a/pkg/controller/direct/registry/registry.go
+++ b/pkg/controller/direct/registry/registry.go
@@ -79,6 +79,7 @@ func AdapterForURL(ctx context.Context, url string) (directbase.Adapter, error) 
 	}
 	return nil, nil
 }
+
 func Init(ctx context.Context, config *config.ControllerConfig) error {
 	for _, registration := range singleton.registrations {
 		model, err := registration.factory(ctx, config)
--- a/pkg/gvks/supportedgvks/supportedgvks.go
+++ b/pkg/gvks/supportedgvks/supportedgvks.go
@@ -15,7 +15,10 @@
 package supportedgvks
 
 import (
+	"fmt"
+
 	iamapi "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/apis/iam/v1beta1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/crd/crdloader"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/dcl/metadata"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/krmtotf"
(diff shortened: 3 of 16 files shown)


pkg/controller/direct/registry/registry.go
```diff
@@ -107,6 +139,17 @@ func RegisterModelWithReconcileGate(gvk schema.GroupVersionKind, modelFn ModelFa
 	}
 }
 
+func IsDirectGVK(gvk schema.GroupVersionKind) bool {
```
yuwenma, line 142: We don't want a `IsDirectGVK` function because version does not determine if a resource is Direct or not. You can use `IsDirectByGK` (below) instead.

pkg/controller/direct/registry/registry.go
```diff
@@ -32,6 +33,36 @@ type registry struct {
...
+	schema.GroupKind{Group: "networkconnectivity.cnrm.cloud.google.com", Kind: "NetworkConnectivityServiceConnectionPolicy"}: k8s.KCCAPIVersionV1Alpha1,
+	schema.GroupKind{Group: "redis.cnrm.cloud.google.com", Kind: "RedisCluster"}:                                             k8s.KCCAPIVersionV1Alpha1,
+	schema.GroupKind{Group: "tags.cnrm.cloud.google.com", Kind: "TagsTagKey"}:                                                k8s.KCCAPIVersionV1Beta1,
+	schema.GroupKind{Group: "sql.cnrm.cloud.google.com", Kind: "SQLInstance"}:                                                k8s.KCCAPIVersionV1Beta1,
+	schema.GroupKind{Group: "compute.cnrm.cloud.google.com", Kind: "ComputeForwardingRule"}:                                  k8s.KCCAPIVersionV1Beta1,
+	schema.GroupKind{Group: "privateca.cnrm.cloud.google.com", Kind: "PrivateCACAPool"}:                                      k8s.KCCAPIVersionV1Beta1,
+	schema.GroupKind{Group: "securesourcemanager.cnrm.cloud.google.com", Kind: "SecureSourceManagerInstance"}:                k8s.KCCAPIVersionV1Alpha1,
+}
+
+func GetDirectGVKs() []schema.GroupVersionKind {
```
yuwenma, line 58: Allow-list may not be a good idea because it is hard to keep track and it buries into the code. To encourage external contributors, we shall be careful every additional step we ask them to do.

If you want to list all the direct resource and make that self-maintained, you can iterate the CRDs and call  
 `IsDirectByGK`.

pkg/gvks/supportedgvks/supportedgvks.go
```diff
@@ -39,6 +40,8 @@ func ManualResources(smLoader *servicemappingloader.ServiceMappingLoader, servic
 func resources(smLoader *servicemappingloader.ServiceMappingLoader, serviceMetaLoader metadata.ServiceMetadataLoader, includesAutoGen bool) []schema.GroupVersionKind {
```
yuwenma, line 40: Mind explaining when the supportedgvk is called?

scripts/generate-google3-docs/resource-reference/templates/apikeys_apikeyskey.tmpl
yuwenma, line 1: Does the current alpha resource have reference? If not, maybe adding all of them (and make it automatic) rather than only adding the direct resource.

maqiuyujoyce: No, you are right, alpha resources should not be added. Good catch! This is by mistake. I thought this is a direct GK as it has the direct controller and it is registered in init(). But the CRD said it is still based on TF?

pkg/gvks/supportedgvks/supportedgvks.go
maqiuyujoyce: It is called at multiple places, like crd generation, resource doc generation, cnrm role generation, and etc, when the caller needs to know what all the GVKs are, what manual GVKs are.

pkg/controller/direct/registry/registry.go
maqiuyujoyce: This is not doable for unit tests/local call of functions () that doesn't dynamically register the direct controllers. E.g. when I ran `make resource-docs`, the singleton registry is empty so there is no direct GK for the caller to identify. That's why I think it's important to also have a static registry.

We can also just go through all the CRDs whenever we want to figure out what direct CRDs are.

pkg/controller/direct/registry/registry.go
maqiuyujoyce: IsDirectGVK here is used to match for the scenarios that require version information. E.g. when we generate reference docs, we only generate it for v1beta1 resources. So version information will be necessary for the caller.

yuwenma: Based on the scenario you described, can it be
``` 
if gvk.Version == "v1beta1" {
      if IsDirectByGK(gvk.GroupKind) {
         ...
      }
}
```

pkg/controller/direct/registry/registry.go
yuwenma: I think there's some confusion about the direct controller registry: when a direct controller is registered, it does not mean it would be used. It is the CRD's absence of tf2crd or dcl2crd label that determines whether a resource is direct resource or not. 

So the label could be frequently changed and this allow list approach could be error-prone (and out of date easily).

pkg/controller/direct/registry/registry.go
```diff
@@ -32,6 +33,36 @@ type registry struct {
 	registrations map[schema.GroupKind]*registration
 }
 
+// staticRegistrations contain all the GKs and their highest versions supported
+// via direct controllers.
+// It's the static set of the direct controllers, and it should contain all the
+// GKs NOT based on TF, DCL, or handwritten IAM types.
+var staticRegistrations = map[schema.GroupKind]string{
```
jasonvigil, line 40: Not sure I understand the purpose of this mapping. Don't the direct controllers reigster themselves with `RegisterModel` in their `init()` functions already? Can we use their registrations instead of these static ones?

pkg/controller/direct/registry/registry.go
maqiuyujoyce: As discussed, I replaced the static allowlist with a CRD go-through.

pkg/controller/direct/registry/registry.go
maqiuyujoyce: Thanks for the suggestion! I updated the logic so that registry.go is kept untouched.

pkg/controller/direct/registry/registry.go
maqiuyujoyce: The registration via init() doesn't happen in unit tests or local scripts. I have discussed it with Yuwen and decided to go with a different route: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2664#discussion_r1755254382

Makefile
```diff
@@ -82,6 +81,10 @@ manifests: generate
 	# for direct controllers
 	dev/tasks/generate-crds
 
+	# Generating cnrm cluster roles is dependent on the existence of directory
+	# config/crds/resources with all the freshly generated CRDs.
+	go run ./scripts/generate-cnrm-cluster-roles/main.go
```
yuwenma, line 86: Nice!

scripts/generate-google3-docs/resource-reference/main.go
```diff
@@ -293,13 +305,25 @@ func constructResourceForGVK(gvk schema.GroupVersionKind, smLoader *servicemappi
...
+			}
 		}
 	}
 
 	r.SampleYamls, err = buildSamples(r.Kind, sampleDirPathForGVK(gvk), smLoader)
 	if err != nil {
-		return nil, fmt.Errorf("error building samples: %w", err)
+		// TODO: Samples should also be required for direct CRDs.
+		if isDirectGVK &&
+			strings.Contains(err.Error(), fmt.Sprintf("k8s-config-connector/config/samples/resources/%s: no such file or directory", strings.ToLower(gvk.Kind))) {
```
yuwenma, line 322: I like this check, but I think this could fail on renamed repo.

yuwenma:
One comment, other looks good. (You may also want to update the PR description).

scripts/generate-google3-docs/resource-reference/main.go
maqiuyujoyce: By renamed repo, did you mean that we might rename our repo from `k8s-config-connector` to something else?

I consider this check temporary because samples are actually pretty important and we should fill the sample gap asap. I'd assume this check will be removed in early Q4. Based on this timeline, does it still matter?

yuwenma: no, I mean people could git clone the repo to a different local directory, in which case this code won't work. (I recalled someone named their repo `kcc` rather than k8s-config-connector).

maqiuyujoyce: Gotcha! Removed `k8s-config-connector` in the error message.

jingyih: nite: maybe we can construct the matching string using "repo.GetResourcesSamplesPath()" and gvk?

jingyih:
/lgtm
/approve


/hold 
for others to review. feel free to cancel when we need to merge it.

pkg/gvks/supportedgvks/supportedgvks.go
```diff
@@ -42,6 +67,43 @@ func resources(smLoader *servicemappingloader.ServiceMappingLoader, serviceMetaL
...
+	if err != nil {
+		return nil, fmt.Errorf("error loading crds: %w", err)
+	}
+	handWrittenIAMTypes := make(map[schema.GroupVersionKind]bool)
+	directResources := make(map[schema.GroupVersionKind]bool)
+	for _, gvk := range BasedOnHandwrittenIAMTypes() {
+		handWrittenIAMTypes[gvk] = true
+	}
+	for _, crd := range crds {
+		if crd.ObjectMeta.Labels["cnrm.cloud.google.com/tf2crd"] == "true" {
```
jasonvigil, line 81: For SQLInstance, we have "partially enabled" the direct controller. This means:
- The tf2crd label is present in the CRD
- Only when spec.cloneSource is specified, the direct controller will manage the resource. Otherwise, the tf controller will manage the resource.

Therefore, this function will not report SQLInstance as a direct resource (it will think it's a terraform-based resource). Do you anticipate this will cause any issues?

maqiuyujoyce: We only want to identify pure direct resources using this functionm, so the mixed resource won't be causing any issues.

jasonvigil: Cool, thanks @maqiuyujoyce !

PR #2672, opened 2024-09-10
fix: write finalizers eagerly in direct controller

justinsb:
This avoids exiting early and not setting them.


yuwenma (commented):
/lgtm
/approve

PR #2674, opened 2024-09-10
Add test case for SQLInstance replica configuration

jasonvigil:
Update replicaConfiguration mapping code to improve readability, make similar to generated mapping code.


jasonvigil:
Going to rebase this

PR #2678, opened 2024-09-11
tests: roundtrip for DataformRepository

acpana:
Adds a fuzz roundtrip for DataformRepository.


yuwenma (commented):
oh, I think you need to add it to https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/dev/tasks/find-missing-fields to trigger the test

acpana:
I was going to add that as part of https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2685 -- I noticed we don't run the `fuzz`ing in the presubmit CI. 🤔 I can also add it here and keep the other PR focused on the CI job.

PR #2686, opened 2024-09-12
tools: require "--kind" and "--proto-resource" when generating types

jingyih:
This PR introduces the following changes:

 - Marks the `--proto-resource` flag as explicitly required, since the proto resource name is required for identifying the starting point of type generation.
 - Ensures that the first character (rune) of the proto resource name is automatically capitalized within the tool, eliminating the need for users to remember to capitalize the first letter in their input.
 - Also marks the `--kind` flag as required, given that kind and proto message often do not have the same name.

### DIFF
--- a/dev/tools/controllerbuilder/pkg/commands/generatetypes/generatetypescommand.go
+++ b/dev/tools/controllerbuilder/pkg/commands/generatetypes/generatetypescommand.go
@@ -80,6 +80,9 @@ func RunGenerateCRD(ctx context.Context, o *GenerateCRDOptions) error {
 	if o.ResourceProtoName == "" {
 		return fmt.Errorf("`--proto-resource` is required")
 	}
+	if o.ResourceKindName == "" {
+		return fmt.Errorf("`--kind` is required")
+	}
 	o.ResourceProtoName = capitalizeFirstRune(o.ResourceProtoName)
 
 	gv, err := schema.ParseGroupVersion(o.APIVersion)
@@ -127,9 +130,6 @@ func RunGenerateCRD(ctx context.Context, o *GenerateCRDOptions) error {
 	}
 
 	kind := o.ResourceKindName
-	if kind == "" {
-		kind = o.ResourceProtoName
-	}
 	if !scaffolder.TypeFileNotExist(kind) {
 		fmt.Printf("file %s already exists, skipping\n", scaffolder.GetTypeFile(kind))
 	} else {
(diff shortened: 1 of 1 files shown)


dev/tools/controllerbuilder/pkg/commands/generatetypes/generatetypescommand.go
```diff
@@ -122,23 +126,26 @@ func RunGenerateCRD(ctx context.Context, o *GenerateCRDOptions) error {
...
-			fmt.Printf("file %s already exists, skipping\n", scaffolder.GetTypeFile(kind))
-		} else {
-			err := scaffolder.AddTypeFile(kind, o.ResourceProtoName)
-			if err != nil {
-				return fmt.Errorf("add type file %s: %w", scaffolder.GetTypeFile(kind), err)
-			}
+	kind := o.ResourceKindName
+	if kind == "" {
+		kind = o.ResourceProtoName
+	}
```
yuwenma, line 132: Actually my experience is that the Kind and resourceProtoName are always different, because Kind always has the service name prefixed, while resourceProtoName is always ignore that (which makes sense for both case due to the different conditions). So I think may be we shall make both --kind and --proto-resource as required to stabilize the output.

jingyih: Yeah makes sense. Updated

dev/tools/controllerbuilder/pkg/commands/generatetypes/generatetypescommand.go
```diff
@@ -77,6 +77,13 @@ func RunGenerateCRD(ctx context.Context, o *GenerateCRDOptions) error {
 	if o.GenerateOptions.ProtoSourcePath == "" {
 		return fmt.Errorf("`proto-source-path` is required")
 	}
+	if o.ResourceProtoName == "" {
+		return fmt.Errorf("`--proto-resource` is required")
+	}
+	if o.ResourceKindName == "" {
+		return fmt.Errorf("`--kind` is required")
+	}
+	o.ResourceProtoName = capitalizeFirstRune(o.ResourceProtoName)
```
yuwenma, line 86: What happens if neither is provided previously? IIRC, if --kind and --proto-resource is not given, this tool generates everything except the "<kind>_types.go" file. Maybe we'd better still keep this functionality and only require both if either one is given.

yuwenma (commented):
One suggestion, otherwise looks good!

yuwenma (commented):
/lgtm
/approve

We can discuss the flag usage further. But definitely not a blocker! Thanks for sending the improvements PR!

PR #2688, opened 2024-09-12
feat: add SciFi controller for BigQueryDataTransferConfig

### LINKED ISSUE #1432: BQ Data Transfer Configs stuck on "Updating"
### Checklist

- [X] I did not find a related open issue.
- [X] I did not find a solution in the troubleshooting guide: (https://cloud.google.com/config-connector/docs/troubleshooting)
- [X] If this issue is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Bug Description

Scheduled Queries created via BQ Data Transfer Configs keep the KCC object stuck in "Updating", while a new Scheduled Query is created in GCP each update cycle.

### Additional Diagnostic Information

I'm deploying a BQ Data Transfer Config via KCC, by applying a k8s manifest via Terraform. The Data Transfer Config is used to set up a Scheduled Query. The object is created in Kubernetes, and the actual Scheduled Query is created in GCP, but the k8s object is stuck continuously in the "Updating" state, and with each update cycle, a new Scheduled Query is created in GCP. The k8s object does not seem to be aware that the Scheduled Query is actually created, and will keep trying to create new ones.

Server Version: v1.27.8-gke.1067004

### Config Connector Version

1.113.0

### Config Connector Mode

namespaced mode (default)

### Log Output

_No response_

### Steps to reproduce the issue

1) deploy a BigQuery Scheduled Query via KCC
2) check the status/count of the Scheduled Queries in GCP, and the status of the KCC object

### YAML snippets

```YAML
apiVersion: bigquerydatatransfer.cnrm.cloud.google.com/v1alpha1
kind: BigQueryDataTransferConfig
metadata:
  annotations:
  creationTimestamp: "2024-03-26T09:54:58Z"
  finalizers:
  generation: 1
  name: tstquery9
  namespace: placeholder
  resourceVersion: "223432102"
  uid: aa945d85-6014-465b-bcd6-a99a17f98ac9
spec:
  dataSourceId: scheduled_query
  destinationDatasetId: test
  displayName: tstquery9
  location: EU
  params:
    destination_table_name_template: tsttable3
    query: SELECT qtr,nonrep FROM `project-placeholder.test.tsttable`
    write_disposition: WRITE_APPEND
  projectRef:
    name: project-placeholder
  schedule: every monday 08:00
```

jingyih:
Fixes #1432


pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
```diff
@@ -0,0 +1,351 @@
...
+func (m *model) AdapterForObject(ctx context.Context, reader client.Reader, u *unstructured.Unstructured) (directbase.Adapter, error) {
+	obj := &krm.BigQueryDataTransferConfig{}
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &obj); err != nil {
+		return nil, fmt.Errorf("error converting to %T: %w", obj, err)
+	}
+
+	// Get ResourceID
+	resourceID := direct.ValueOf(obj.Spec.ResourceID)
+	if resourceID == "" {
+		serviceGeneratedID, err := parseServiceGeneratedIDFromExternalRef(obj)
```
jingyih, line 86: resource ID is service generated. It could be "" prior to resource creation.

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
```diff
@@ -0,0 +1,351 @@
...
+
+var _ directbase.Adapter = &Adapter{}
+
+func (a *Adapter) Find(ctx context.Context) (bool, error) {
+	log := klog.FromContext(ctx).WithName(ctrlName)
+
+	if a.id == nil {
+		return false, nil
+	}
+	if a.id.transferConfigID == "" { // resource ID is not yet generated by the GCP service
```
jingyih, line 192: special handling for finding resource prior to being created.

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
```diff
@@ -0,0 +1,351 @@
...
+	status.ObservedState = BigQueryDataTransferConfigObservedState_FromProto(mapCtx, created)
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+	// The UUID is service generated. e.g. "projects/{project_id}/locations/{region}/transferConfigs/{config_id}"
+	serviceGeneratedID, err := parseServiceGeneratedIDFromName(created.Name)
+	if err != nil {
+		return fmt.Errorf("error converting %s to BigQueryDataTransferConfigIdentity: %w", created.Name, err)
+	}
+	a.id.transferConfigID = serviceGeneratedID
```
jingyih, line 258: need to write the service generated ID back

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
```diff
@@ -0,0 +1,351 @@
...
+		updateMask.Paths = append(updateMask.Paths, "display_name")
+	}
+	// TODO(jingyih): handle other mutable fields
+
+	desired := a.desired.DeepCopy()
+	resource := BigQueryDataTransferConfigSpec_ToProto(mapCtx, &desired.Spec)
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+	resource.Name = a.id.FullyQualifiedName() // need to pass service generated ID to GCP API
```
jingyih, line 290: need to pass the service generated ID when updating the resource.

jingyih (commented):
@yuwenma I added some notes on special handling for service generated ID.

apis/bigquerydatatransfer/v1alpha1/groupversion_info.go
```diff
@@ -30,4 +30,10 @@ var (
 
 	// AddToScheme adds the types in this group-version to the given scheme.
 	AddToScheme = SchemeBuilder.AddToScheme
+
+	GroupVersionKind = schema.GroupVersionKind{
+		Group:   GroupVersion.Group,
+		Version: GroupVersion.Version,
+		Kind:    "BigQueryDataTransferConfig",
+	}
```
yuwenma, line 38: A better format could be adding 
`var BigQueryDataTransferConfigGVK = GroupVersion.WithKind("BigQueryDataTransferConfig")` in the <Kind>_types.go, which is what the latest template does.

dev/tasks/generate-crds
```diff
@@ -55,7 +55,7 @@ go run ./scripts/crd-tools reflow-descriptions --dir apis/config/crd/
 
 # excluded_resources are resources that are under development for a direct conversion
 # we don't modify the CRD just yet for those but will in the future
-excluded_resources=("bigquerydatatransferconfig" "securesourcemanagerinstance")
```
yuwenma, line 58: 👍

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+
+	// Resolve KMSCryptoKey Ref
+	if obj.Spec.EncryptionConfiguration != nil {
+		kmsCryptoKeyRef, err := refv1beta1.ResolveKMSCryptoKeyRef(ctx, reader, obj, obj.Spec.EncryptionConfiguration.KmsKeyRef)
+		if err != nil {
+			return nil, err
+		}
+		obj.Spec.EncryptionConfiguration.KmsKeyRef = kmsCryptoKeyRef
+	}
```
yuwenma, line 124: I think we shall move the non-parent depedencies in the create and update calls.

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
```diff
@@ -0,0 +1,351 @@
...
+}
+
+var _ directbase.Adapter = &Adapter{}
+
+func (a *Adapter) Find(ctx context.Context) (bool, error) {
+	log := klog.FromContext(ctx).WithName(ctrlName)
+
+	if a.id == nil {
+		return false, nil
+	}
```
yuwenma, line 191: Mind explaining what this means?

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
```diff
@@ -0,0 +1,351 @@
...
+}
+
+func (m *model) AdapterForObject(ctx context.Context, reader client.Reader, u *unstructured.Unstructured) (directbase.Adapter, error) {
+	obj := &krm.BigQueryDataTransferConfig{}
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &obj); err != nil {
+		return nil, fmt.Errorf("error converting to %T: %w", obj, err)
+	}
+
+	// Get ResourceID
+	resourceID := direct.ValueOf(obj.Spec.ResourceID)
```
yuwenma, line 84: Did you miss a case that metadata.name --> resourceID if `spec.resourceID` is not given? Or is that not a valid case here? I mean the case that KCC *acquire* a GCP resource that matches the `metadata.name`.

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+	projectID := projectRef.ProjectID
+	if projectID == "" {
+		return nil, fmt.Errorf("cannot resolve project")
+	}
+
+	var id *BigQueryDataTransferConfigIdentity
+	externalRef := direct.ValueOf(obj.Status.ExternalRef)
+	if externalRef == "" {
+		id = BuildID(projectID, location, resourceID)
```
yuwenma, line 139: So the problem here is that since we do not assign `metadata.name` to resourceID, if `spec.resourceID` is not given and the GCP object is not created (no `status.externalRef`), this will always pass in an empty `resourceID`, which fails the following logics.

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
```diff
@@ -0,0 +1,351 @@
...
+
+		if id.projectID != projectID {
+			return nil, fmt.Errorf("BigQueryDataTransferConfig %s/%s has spec.projectRef changed, expect %s, got %s",
+				u.GetNamespace(), u.GetName(), id.projectID, projectID)
+		}
+		if id.location != location {
+			return nil, fmt.Errorf("BigQueryDataTransferConfig %s/%s has spec.location changed, expect %s, got %s",
+				u.GetNamespace(), u.GetName(), id.location, location)
+		}
+		if id.transferConfigID != resourceID {
```
yuwenma, line 154: I think this only works for acquire but not KCC-created.  Because the transferConfigID is server generated and KCC does not write it back to spec.resourceID (and not metadata.name)

yuwenma:
I feel like maybe we can merge the tool-generated code first, and handle the service-generated ID in a follow up  PR, because the change needs to happen in multiple places and I'd like to use that as a coding example.

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
yuwenma: Otherwise the presence of the dependency would block the target resource

apis/bigquerydatatransfer/v1alpha1/groupversion_info.go
jingyih: Cool. I am updating the controller template in #2692, so that we don't have a undefined `GroupVersionKind` right after scaffolding the controller code.

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
jingyih: IIUC, it mainly affects `Find()` call? In other words, if the dependencies are not present, `Create()` and `Update()` will be blocked anyway, but `Find()` should not be blocked.

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
jingyih: I will remove this block. I think it was from my previous iteration on how to handle the service generated ID. Thanks for spotting!

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
jingyih: For resource with service generated ID, I am not sure how useful is `metadata.name`. It will be ignored during GCP resource creation, and later it cannot be used to match (acquire) the GCP resource. To acquire the GCP resource, we have to give GCP API the service generated ID.

jingyih:
Rebased to resolve a conflict.

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
yuwenma: Mind elaborating a little bit more on why "we want users to specifically set spec.resourceID to the service generated ID"?

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
jingyih: From my tests, the generated ID is something like "66e60359-0000-2d33-8bc7-c82add80addc". I think k8s name needs to start and end with an alphanumeric character?

Besides, when user try to acquire a GCP resource, I prefer them to explicitly set `spec.ResourceID`.

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
jingyih: I was referring to 
https://screenshot.googleplex.com/6Dcx5KjqfXvuexB

I think it makes sense to have users explicitly set `spec.ResourceID` to express the intend to "acquire a resource".

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
```diff
@@ -0,0 +1,408 @@
...
+	var opts []option.ClientOption
+	if m.config.UserAgent != "" {
+		opts = append(opts, option.WithUserAgent(m.config.UserAgent))
+	}
+	if m.config.HTTPClient != nil {
+		opts = append(opts, option.WithHTTPClient(m.config.HTTPClient))
+	}
+	if m.config.UserProjectOverride && m.config.BillingProject != "" {
+		opts = append(opts, option.WithQuotaProject(m.config.BillingProject))
+	}
```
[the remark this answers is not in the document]
jingyih: Cool! update the code.

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
yuwenma: It affects `Delete`. If a reference is missing, it cannot delete the object.

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
yuwenma: kk, looks like all BigQuery service-generated ID is all using the UUID format. Then I'm good with this.

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
yuwenma: I don't think this is a blocker issue.

PR #2691, opened 2024-09-13
chore: improve controller-builder template and fix externalRef issues

yuwenma:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/dev/tools/controllerbuilder/template/controller/controller.go
+++ b/dev/tools/controllerbuilder/template/controller/controller.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package template
+package controller
 
 type ControllerArgs struct {
 	// The ConfigConnector Group without cnrm.google.com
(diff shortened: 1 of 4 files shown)


dev/tools/controllerbuilder/template/controller/controller.go
```diff
@@ -269,54 +259,70 @@ func (a *Adapter) Update(ctx context.Context, updateOp *directbase.UpdateOperati
...
 
 // Delete implements the Adapter interface.
 func (a *Adapter) Delete(ctx context.Context, deleteOp *directbase.DeleteOperation) (bool, error) {
 	log := klog.FromContext(ctx).WithName(ctrlName)
-	log.V(2).Info("deleting {{.Kind}}", "name", a.id.fullyQualifiedName())
+	log.V(2).Info("deleting {{.ProtoResource}}", "name", a.id.FullyQualifiedName())
 
-	if a.resourceID == "" {
-		return false, nil
-	}
```
yuwenma, line 308: This is actually the reason about this PR. Current controllers which use the old template shall fix it.

jingyih (approved):
/lgtm

dev/tools/controllerbuilder/template/controller/controller.go
jingyih: I think this is OK? In the template, `a.resourceID` is set from either `spec.ResourceID` or `metadata.Name`.

yuwenma: You are right. I was fixing here and there and don't know where this comment comes from

PR #2699, opened 2024-09-16
Update mockgcp and tests for attached clusters.

hankfreund:
### Change description

- Updated the mockgkemulticloud service to conform with current LRO handling guidelines
- Added all currently supported update mask paths to the mock service
- Added a "full" test fixture with all currently supported (by KCC) fields and updates

### Tests you have done

- Updated the golden files based on the MockGCP output
- Tested against the real service to ensure everything works as expected
- Ran make ready-pr to ensure this PR is ready for review.


yuwenma (commented):
/lgtm
/approve 

Looks good! Thank you

yuwenma:
Hi, @hankfreund 
One follow up question: have you run against the real gcp (via `hack/record-gcp fixtures/<test name>`)? If so, could you upload the golden http log as well? (We expect the real GCP log as the very first git-commit so as we can set that as the expectation for the feature development ).

yuwenma:
> I have run against the real GCP and ensured that it works. From my understanding, the recorded logs for attached clusters are only from MockGCP instead of the real service because of the difficulties ensuring the external cluster and resources are available for generating the logs. @gemmahou might have more details about this?

Awesome!  Could you upload the real GCP log as well? The real GCP is to validate the functionality of the MockGCP. And after that, only the MockGCP log is needed.

PR #2703, opened 2024-09-16
fix: Fix FirestoreDatabase postsubmit failures

jasonvigil:
Due to issue with FirestoreDatabase service LRO / eventual consistency, wait after creating / updating so that the changes eventually become visible.

Also, do not attempt to update if the concurrency_mode or point_in_time_recovery_enablement fields are unspecified.

### DIFF
--- a/pkg/controller/direct/firestore/firestoredatabase_controller.go
+++ b/pkg/controller/direct/firestore/firestoredatabase_controller.go
@@ -237,12 +237,18 @@ func (a *Adapter) Update(ctx context.Context, updateOp *directbase.UpdateOperati
 
 	updateMask := &fieldmaskpb.FieldMask{}
 	if !reflect.DeepEqual(resource.ConcurrencyMode, a.actual.ConcurrencyMode) {
-		newDb.ConcurrencyMode = resource.ConcurrencyMode
-		updateMask.Paths = append(updateMask.Paths, "concurrency_mode")
+		// Skip update if concurrency_mode is unspecified
+		if resource.ConcurrencyMode != firestorepb.Database_CONCURRENCY_MODE_UNSPECIFIED {
+			newDb.ConcurrencyMode = resource.ConcurrencyMode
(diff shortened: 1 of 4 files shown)


pkg/controller/direct/firestore/firestoredatabase_controller.go
```diff
@@ -237,12 +237,18 @@ func (a *Adapter) Update(ctx context.Context, updateOp *directbase.UpdateOperati
 
 	updateMask := &fieldmaskpb.FieldMask{}
 	if !reflect.DeepEqual(resource.ConcurrencyMode, a.actual.ConcurrencyMode) {
-		newDb.ConcurrencyMode = resource.ConcurrencyMode
-		updateMask.Paths = append(updateMask.Paths, "concurrency_mode")
+		// Skip update if concurrency_mode is unspecified
+		if resource.ConcurrencyMode != firestorepb.Database_CONCURRENCY_MODE_UNSPECIFIED {
```
yuwenma, line 241: If i read the code correctly, this is the `spec.concurrencyMode` field. What does `spec.concurrencyMode == Database_CONCURRENCY_MODE_UNSPECIFIED` mean? Is this the default value if user not specifying the spec.concurrencyMode? Can we add a test suite (maybe scenario test) for this case?

jasonvigil: You are right, `Database_CONCURRENCY_MODE_UNSPECIFIED` is the value that gets assigned by `FirestoreDatabaseSpec_ToProto` when `spec.concurrencyMode` is not specified. However, GCP assigns some other default value (`Database_PESSIMISTIC`). And, potentially, the field could be managed separately from KCC, assigning some other value. This case is already covered by the `firestoredatabase-minimal` test.

yuwenma: I see. That makes sense. So basically this is the default handling

yuwenma (commented):
/lgtm
/approve

PR #2704, opened 2024-09-16
Shift-left the defaulting of optional and immutable spec fields

maqiuyujoyce:
### Change description

Shift-left the defaulting of the following spec fields to fix the unexpected diffs in re-reconciliation when there is no change in the CR:

* `spec.location` in CloudBuildTrigger.
* `spec.initialGroupConfig` in CloudIdentityGroup.
* `spec.database` in FirestoreIndex.

This looks like an old bug and was only exposed when `state-into-spec: absent` became the default configuration.

_Following is the context:_

Some optional and immutable spec fields are not specified by the users, and are defaulted by TF before it makes the API call to create the resource.
During reconciliation, for an immutable field, KCC will intentionally [match the live state with the desired state when constructing the live state](https://github.com/maqiuyujoyce/k8s-config-connector/blob/master/pkg/krmtotf/fetchlivestate.go#L233). But this is an incorrect behavior for the optional and immutable spec fields defaulted by TF, because it will unset such fields in the live state while the actual desired state passed to the API is not unset, but has the implicit default value.

If a resource has a field `spec.f` like that, after applying a manifest without this field `spec.f` specified, here is what will happen.
1. During the first reconciliation, the live state is nil which indicates a creation will happen.
    * With `cnrm.cloud.google.com/state-into-spec: merge`, the successfully reconciled CR contains field `spec.f` with the default value.
    * With `cnrm.cloud.google.com/state-into-spec: absent`, the successfully reconciled CR does NOT contain field `spec.f` with the default value.
1. During the second reconciliation, 
    1. KCC controller first retrieves the live state.
        * With `cnrm.cloud.google.com/state-into-spec: merge`, the desired value of field `spec.f` is non-nil, so it is kept as is.
        * With `cnrm.cloud.google.com/state-into-spec: absent`, the desired value of field `spec.f` is nil, which indicates this field is `unset`, so live state will have this field removed, even if it exists in the underlying GCP resource.
    1.  KCC controller then uses Diff() from TF to compare the live state and the desired state. Diff will "hydrate" the desired state with the default values in the TF schema.
        * With `cnrm.cloud.google.com/state-into-spec: merge`, both the live state and the desired state of the resource have the default values for field `spec.f`, and there is no diff for this field.
        * With `cnrm.cloud.google.com/state-into-spec: absent`, the live state has field `spec.f` unset, while the desired state has field `spec.f` set to its default value, so there is an unexpected diff.

With the defaulting happening at the KCC level, the desired state will have the default value explicitly set, and avoid the issue above.

Alternatively, we can update the logic to NOT match the live state with the desired state when constructing the live state, e.g. #2716. This is a more effective fix, but it'll impact more resources (so might give us surprises), and this will keep the defaulting implicit, which isn't well-aligned with the declarative design principle. 

### Tests you have done

### DIFF
--- a/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_cloudbuildtriggers.cloudbuild.cnrm.cloud.google.com.yaml
+++ b/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_cloudbuildtriggers.cloudbuild.cnrm.cloud.google.com.yaml
@@ -1043,6 +1043,7 @@ spec:
                   type: string
                 type: array
               location:
+                default: global
                 description: |-
                   Immutable. The [Cloud Build location](https://cloud.google.com/build/docs/locations) for the trigger.
                   If not specified, "global" is used.
--- a/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_cloudidentitygroups.cloudidentity.cnrm.cloud.google.com.yaml
+++ b/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_cloudidentitygroups.cloudidentity.cnrm.cloud.google.com.yaml
@@ -96,6 +96,7 @@ spec:
                 - id
                 type: object
               initialGroupConfig:
+                default: EMPTY
                 description: |-
                   Immutable. The initial configuration options for creating a Group.
 
--- a/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_firestoreindexes.firestore.cnrm.cloud.google.com.yaml
+++ b/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_firestoreindexes.firestore.cnrm.cloud.google.com.yaml
@@ -62,6 +62,7 @@ spec:
                 description: Immutable. The collection being indexed.
                 type: string
               database:
+                default: (default)
                 description: Immutable. The Firestore database id. Defaults to '"(default)"'.
                 type: string
               fields:
--- a/pkg/resourceoverrides/cloudbuild_trigger.go
+++ b/pkg/resourceoverrides/cloudbuild_trigger.go
@@ -1,0 +1,39 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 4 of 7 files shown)


maqiuyujoyce:
FYI @yuwenma I believe this will be a release blocker for v1.123.0. Without this change, at least the three resources mentioned in the PR will not reconcile properly for drift correction.

yuwenma:
Could you elaborate why shift-left is preferred, compared to set the default in a later time (i.e. when propagating the Object to TF)

maqiuyujoyce:
Shift-left the defaulting is preferred because it explicitly updates each impacted CRD so the scope of the change is clear. It also aligns with the declarative design best practices: The desired state (including the default values) will be persisted explicitly.

On the other hand, if we update the logic in the TF-based controller to NOT match the live state with the desired state when constructing the live state (#2716), it's hard to predict which Kinds will be impacted and whether there are any unexpected side effects. So it'll be more risky.

yuwenma:
> Shift-left the defaulting is preferred because it explicitly updates each impacted CRD so the scope of the change is clear. It also aligns with the declarative design best practices: The desired state (including the default values) will be persisted explicitly.
> 
> On the other hand, if we update the logic in the TF-based controller to NOT match the live state with the desired state when constructing the live state (#2716), it's hard to predict which Kinds will be impacted and whether there are any unexpected side effects. So it'll be more risky.

Can we assign the default value when propagating the CR object to TF?

maqiuyujoyce:
Do you mean that we should default the fields in the desired state? We probably can, but we don't know the default values of those fields - the default values configured in the TF schema are accessible, but they may not be used consistently in TF resource logic so I'd delegate the defaulting to TF instead of using KCC.

yuwenma:
> Do you mean that we should default the fields in the desired state? We probably can, but we don't know the default values of those fields - the default values configured in the TF schema are accessible, but they may not be used consistently in TF resource logic so I'd delegate the defaulting to TF instead of using KCC.

Maybe I misread the lines,  but does this PR already override the defaulting value (before KCC even received the object) ?  Or do you mean something different?

maqiuyujoyce:
Yes, in this PR, the defaulting happens before KCC reconcile the resources.

Discussed with @yuwenma offline about it. The concern is that with the CRD level defaulting, KCC can't tell whether the value is defaulted or specified by the users.
The concern is valid if API treats `unset` and an implicit value differently. If when the passed in field is `unset`, users can then explicitly set a value in the future after resource creation, then defaulting the value for the users restricts their use cases.

On the other hand, if the API always assigns a default value to the field no matter it is `unset` or not, and doesn't allow further changes, then CRD level defaulting should have no issue. Users should not make any changes to such a field after resource creation in KCC, and this is the same behavior if they create a corresponding GCP resource.

Though it'll be a bit expensive to verify and distinguish different "defaulting" and "immutability" behaviors among different APIs so right now we decide not to shift left the defaulting to the CRDs, but have it moved to the TF controller to workaround the original issue.

maqiuyujoyce:
We picked a different option (#2735) to solve the problem.

PR #2705, opened 2024-09-16
Update attached clusters mockGCP responses to more closely match GCP.

hankfreund:
- Set all the fields that are returned by the attached cluster service
- Update resource not found message
- Regenerate golden files
- Update env var substitution for platform version so the samples tests pass

### DIFF
--- a/mockgcp/mockgkemulticloud/attachedcluster.go
+++ b/mockgcp/mockgkemulticloud/attachedcluster.go
@@ -17,6 +17,7 @@ package mockgkemulticloud
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"google.golang.org/genproto/googleapis/longrunning"
@@ -26,6 +27,7 @@ import (
 	"google.golang.org/protobuf/types/known/emptypb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
+	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/common/fields"
 	pb "github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/generated/mockgcp/cloud/gkemulticloud/v1"
 )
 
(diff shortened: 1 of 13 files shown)


mockgcp/mockgkemulticloud/attachedcluster.go
```diff
@@ -66,6 +67,28 @@ func (s *GKEMulticloudV1) CreateAttachedCluster(ctx context.Context, req *pb.Cre
...
+	}
+	if obj.GetMonitoringConfig() == nil {
+		obj.MonitoringConfig = &pb.MonitoringConfig{
+			ManagedPrometheusConfig: &pb.ManagedPrometheusConfig{},
+		}
+	}
+	obj.Fleet.Membership = obj.Fleet.Project + "/locations/global/memberships/" + req.AttachedClusterId
+	obj.CreateTime = timestamppb.New(now)
+	obj.UpdateTime = timestamppb.New(now)
+	obj.Etag = "abcdef0123A="
```
yuwenma, line 83: nit: you can use `computeEtag()` to make it more "real". This field will be handled later on when record the HTTP log.

mockgcp/mockgkemulticloud/attachedcluster.go
```diff
@@ -66,6 +67,28 @@ func (s *GKEMulticloudV1) CreateAttachedCluster(ctx context.Context, req *pb.Cre
...
+	ver, err := trimPlatformVersion(req.GetAttachedCluster().GetPlatformVersion())
```
yuwenma, line 84: Is `spec.platformVersion` required? My gut feeling is that the platform version is associated with the fleet membership which means it could have a default value if unset.

hankfreund: It is a required value, no defaults. There's no correlation between the fleet membership and the platform version.

hankfreund: Note that platform version is different from k8s version. K8s version is determined by the service by querying the underlying cluster. Platform version is the version of attached clusters that's being provisioned.

acpana:
FYI, I think some of the test failure were flakes unfortunately 😢 ; However, for `validations` job you might have to run `make ready-pr` or `make resource-docs` to pass.

mockgcp/mockgkemulticloud/attachedcluster.go
yuwenma: Good to know. Thank you!

yuwenma (commented):
/lgtm
/approve

Good job! Thanks!

hankfreund:
> FYI, I think some of the test failure were flakes unfortunately 😢 ; However, for `validations` job you might have to run `make ready-pr` or `make resource-docs` to pass.

Thanks for calling that out. I missed that failure!

PR #2709, opened 2024-09-17
fix: Fix shouldSkipDriftDetection to handle direct resources

jasonvigil:


acpana:
Ah yes we can do this but might need to manually exclude direct resources with server generated IDs.

Did you test this w any failing tests like `go test -v -tags=integration ./pkg/controller/dynamic/ -timeout 1600s -test.count=1 -test.run 'TestCreateNoChangeUpdateDelete/logging/basic-logbucketmetric'  2>&1` ?

jasonvigil:
@acpana we can exclude resources via https://github.com/jasonvigil/k8s-config-connector/blob/128e196f272348e30ddb1f3143e769ae0a9aa29f/pkg/test/resourcefixture/contexts/register.go#L70, if they don't work

jasonvigil:
@acpana the test passes with this change! 🎉

PR #2695, opened 2024-09-13
[DO NOT MERGE]For testing: Record realGCP and update the mocks

gemmahou:
### Change description

record realGCP for forwarding rule tests [commit1](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2695/commits/c6986ff4f339ac782b75723be67595cc66c2e4ba)

I know there are a lot of diffs, but we only need to focus on the HTTP requests of forwarding rule(search for keyword "forwardingrules"). 

diffs fixed in the [commit2](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2695/commits/0e4dac90c6aaaf1d7f47cf20dccb77084a0f3297):  
1. operation type SetLabel to setLabel
2. format `region` in regional forwardingrule mock server

diffs fixed in the [commit3](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2695/commits/6968136a4524bf225afcf620fdcbae9294b59c2b):  
1. Mock compute LRO default to start from RUNNING(progress 0) and then finish DONE(progress 100). But `setLabels` is a super fast request, the start operation is DONE. 

diffs fixed in the [commit4](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2695/commits/367dfdabaf931a39d62a89425fab5163fc579156):  
1. Rewrite 404 error message for compute resources. We updated the general mock error message in #1221, I rewrote to match the realGCP message(But I think it's intention for that PR is to have a general message, no need to 100% match the real?)

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #2711, opened 2024-09-17
direct: Labels can only be added or modified by the setLabels method

gemmahou:
### Change description

Follow up #2695, fixes: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2695#discussion_r1759592910

During real `Create`, the forwarding rule is first created without labels, then `setLabels` has to be called to add labels to it.

This is a special API design. During mock `Create`, we simply copy the desired resource (including labels) as the created resource. That's why in the mock GET response, we can see the `labels` field but not in the real GET response. I don't think it's worth the effort to change the mock server to match the real GCP for this special case, as long as we know the real GCP behaves differently here and we accept this difference.

If in the future, more resources have the same behavior, we can consider revising the mock server.

### Tests you have done

### DIFF
--- a/mockgcp/mockcompute/operations.go
+++ b/mockgcp/mockcompute/operations.go
@@ -77,7 +77,7 @@ func (s *computeOperations) newLRO(ctx context.Context, projectID string) (*pb.O
 	return op, nil
 }
 
-func (s *computeOperations) startLRO0(ctx context.Context, op *pb.Operation, fqn string, callback func() (proto.Message, error)) (*pb.Operation, error) {
+func (s *computeOperations) startLRO(ctx context.Context, op *pb.Operation, fqn string, callback func() (proto.Message, error), startProgress int32, startStatus pb.Operation_Status) (*pb.Operation, error) {
 	log := klog.FromContext(ctx)
 
 	now := time.Now()
@@ -91,12 +91,12 @@ func (s *computeOperations) startLRO0(ctx context.Context, op *pb.Operation, fqn
 	op.InsertTime = PtrTo(formatTime(now))
 	op.Id = PtrTo(uint64(nanos))
 
-	op.Progress = PtrTo(int32(0))
+	op.Progress = PtrTo(startProgress)
 
 	op.Kind = PtrTo("compute#operation")
 	op.SelfLink = PtrTo("https://www.googleapis.com/compute/v1/" + fqn)
 
-	op.Status = PtrTo(pb.Operation_RUNNING)
+	op.Status = PtrTo(startStatus)
 
 	log.Info("storing operation", "fqn", fqn)
 	if err := s.storage.Create(ctx, fqn, op); err != nil {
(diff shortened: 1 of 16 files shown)


mockgcp/mockcompute/operations.go
```diff
@@ -91,12 +91,12 @@ func (s *computeOperations) startLRO0(ctx context.Context, op *pb.Operation, fqn
 	op.InsertTime = PtrTo(formatTime(now))
 	op.Id = PtrTo(uint64(nanos))
 
-	op.Progress = PtrTo(int32(0))
+	op.Progress = PtrTo(startProgress)
```
justinsb, line 94: As we pass in `op`, it might be easier to set it to zero only if Progress is nil.  Avoids the long-list of unnamed args problem!

mockgcp/mockcompute/operations.go
```diff
@@ -91,12 +91,12 @@ func (s *computeOperations) startLRO0(ctx context.Context, op *pb.Operation, fqn
...
 
 	op.Kind = PtrTo("compute#operation")
 	op.SelfLink = PtrTo("https://www.googleapis.com/compute/v1/" + fqn)
 
-	op.Status = PtrTo(pb.Operation_RUNNING)
+	op.Status = PtrTo(startStatus)
```
justinsb, line 99: Ditto here, maybe only if Status is nil?

mockgcp/mockcompute/service.go
```diff
@@ -222,8 +223,22 @@ func (s *MockService) NewHTTPMux(ctx context.Context, conn *grpc.ClientConn) (ht
 
 	// Returns slightly non-standard errors
 	mux.RewriteError = func(ctx context.Context, error *httpmux.ErrorResponse) {
-		// Does not return status (at least for 404)
-		error.Status = ""
+		if error.Code == 404 {
+			// Transform the mock error message:
+			// i.e. "address \"projects/${projectId}/global/addresses/computeaddress-${uniqueId}\" not found"
+			// into the realGCP error message:
+			// i.e. "The resource 'projects/${projectId}/global/addresses/computeaddress-${uniqueId}' was not found"
```
justinsb, line 230: Can you just return the correct message from the Get method?  That's what we do elsewhere.  It looks like this:

```
  return status.Errorf(codes.NotFound, "The resource '%s' was not found", fqn)
}
```

Or something along those lines, anyway!

pkg/controller/direct/compute/forwardingrule_controller.go
```diff
@@ -358,6 +360,51 @@ func (a *forwardingRuleAdapter) Create(ctx context.Context, createOp *directbase
 		return fmt.Errorf("getting ComputeForwardingRule %q: %w", a.fullyQualifiedName(), err)
 	}
 
+	// Set labels for the created forwarding rule
```
justinsb, line 363: Should we update the mock to reflect the reality of what GCP does (i.e. set labels to nil in the Create method?)

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrule/_http.log
```diff
@@ -236,11 +236,11 @@ X-Xss-Protection: 0
     "errors": [
       {
         "domain": "global",
-        "message": "healthCheck \"projects/${projectId}/global/healthChecks/computehealthcheck-${uniqueId}\" not found",
+        "message": "The resource 'projects/${projectId}/global/healthChecks/computehealthcheck-${uniqueId}' was not found",
```
justinsb, line 239: Do _all_ the GCP resources follow this pattern?  It's probably still easier to update the Get methods (I think so, anyway)

justinsb:
A few suggestions, but this looks like a good fix!

gemmahou:
Discussed offline and reviewed diffs in this PR, opened a new PR #2723 to address the labels.

PR #2712, opened 2024-09-17
example: use SQLInstance Reference

yuwenma:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/apis/refs/v1beta1/sqlinstanceref.go
+++ b/apis/refs/v1beta1/sqlinstanceref.go
@@ -14,6 +14,18 @@
 
 package v1beta1
 
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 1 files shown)


apis/refs/v1beta1/sqlinstanceref.go
```diff
@@ -22,3 +34,83 @@ type SQLInstanceRef struct {
...
+type SQLInstance struct {
+	ProjectID       string
+	SQLInstanceName string
+}
+
+func (s *SQLInstance) String() string {
+	return "projects/" + s.ProjectID + "/instances/" + s.SQLInstanceName
+}
+
+func (r *SQLInstanceRef) ResolveRef(ctx context.Context, reader client.Reader, src client.Object) (*SQLInstance, error) {
```
justinsb, line 47: If we resolve to a SQLInstanceRef (with only external populated), then I don't think we need the second type, and I think we can resolve "in place" in the bigger object.  Is there a reason to prefer SQLInstance over SQLInstanceRef?

yuwenma: I think I see what you mean. Let me try another attempt. Though this PR is a temp solution to unblock the BigQueryConnectionConnection object

PR #2713, opened 2024-09-17
chore: refactor resource ref

yuwenma:
### Change description

Fixes #

### Tests you have done


apis/bigqueryconnection/v1alpha1/connection_reference.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (r *BigQueryConnectionConnectionRef) GVK() schema.GroupVersionKind {
+	return BigQueryConnectionConnectionGVK
+}
+
+func (r *BigQueryConnectionConnectionRef) String() string {
+	return r.External
+}
+
+func (r *BigQueryConnectionConnectionRef) Parent() string {
```
justinsb, line 70: This method might be tricky, because I don't think you can always call it...

apis/bigqueryconnection/v1alpha1/connection_reference.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	r.External = strings.TrimPrefix(r.External, "/")
+	tokens := strings.Split(r.External, "/")
+	if len(tokens) != 6 || tokens[0] != "projects" || tokens[2] != "locations" || tokens[4] != "connections" {
+		return fmt.Errorf("format of BigQueryConnectionConnection external=%q was not known (use projects/<projectId>/locations/<location>/connections/<connectionID>)", r.External)
+	}
+	return nil
+}
+
+func (r *BigQueryConnectionConnectionRef) Resolve(ctx context.Context, reader client.Reader, u *unstructured.Unstructured) error {
```
justinsb, line 86: Nit: This needs a comment.  Is `u` the object that contains `r`, or have we already read the object that `r` points to, and `u` is that object?

justinsb, line 86: I think this method should be called `ComputeExternalRef` or something like that.  Notice that we don't  read `r` (that I can see).  Maybe it should be a method on BigQueryConnectionConnection instead?

pkg/controller/direct/bigqueryconnection/connection_controller.go
```diff
@@ -157,19 +118,19 @@ var _ directbase.Adapter = &Adapter{}
 func (a *Adapter) Find(ctx context.Context) (bool, error) {
 	log := klog.FromContext(ctx).WithName(ctrlName)
 
-	log.V(2).Info("getting BigQueryConnectionConnection", "name", a.id.AsExternalRef())
+	log.V(2).Info("getting BigQueryConnectionConnection", "name", a.desired.Name)
```
justinsb, line 121: Probably `a.id.External`

pkg/controller/direct/bigqueryconnection/connection_controller.go
```diff
@@ -157,19 +118,19 @@ var _ directbase.Adapter = &Adapter{}
...
 	}
-	req := &bigqueryconnectionpb.GetConnectionRequest{Name: a.id.FullyQualifiedName()}
+	req := &bigqueryconnectionpb.GetConnectionRequest{Name: a.id.External}
 	connectionpb, err := a.gcpClient.GetConnection(ctx, req)
 	if err != nil {
 		if direct.IsNotFound(err) {
 			return false, nil
 		}
-		return false, fmt.Errorf("getting BigQueryConnectionConnection %q: %w", a.id.FullyQualifiedName(), err)
+		return false, fmt.Errorf("getting BigQueryConnectionConnection %q: %w", a.desired.Name, err)
```
justinsb, line 133: I don't think we normally populate a.desired.Name, at least not in the default mappers..

pkg/controller/direct/bigqueryconnection/connection_controller.go
```diff
@@ -190,31 +151,29 @@ func (a *Adapter) Create(ctx context.Context, createOp *directbase.CreateOperati
 	}
 
 	req := &bigqueryconnectionpb.CreateConnectionRequest{
-		Parent:     a.id.Parent.String(),
+		Parent:     a.id.Parent(),
```
justinsb, line 154: I would really consider making `Parent` a method that can return an error, just to avoid misuse

pkg/controller/direct/bigqueryconnection/connection_controller.go
```diff
@@ -263,8 +222,11 @@ func (a *Adapter) Export(ctx context.Context) (*unstructured.Unstructured, error
 	if mapCtx.Err() != nil {
 		return nil, mapCtx.Err()
 	}
-	obj.Spec.ProjectRef = &refs.ProjectRef{Name: a.id.Parent.Project}
-	obj.Spec.Location = a.id.Parent.Location
+	tokens := strings.Split(a.id.External, "/")
+	if len(tokens) == 6 {
+		obj.Spec.ProjectRef = &refs.ProjectRef{Name: tokens[1]}
```
justinsb, line 227: We could define some helper methods on the ref also, e.g. `ProjectID` and `Location`.  For 90%+ of our Refs (I would guess), they can delegate to a method like `extractKey("projects")` or `extractKey("locations")`

apis/bigqueryconnection/v1alpha1/connection_reference.go
```diff
@@ -0,0 +1,66 @@
...
+	External string `json:"external,omitempty"`
+
+	// The `name` of a `BigQueryConnectionConnection` resource.
+	Name string `json:"name,omitempty"`
+	// The `namespace` of a `BigQueryConnectionConnection` resource.
+	Namespace string `json:"namespace,omitempty"`
+}
+
+func (r *BigQueryConnectionConnectionRef) ValidateExternal() error {
+	if r.External == "" {
```
justinsb, line 48: Nit: a method called Validate probably should not mutate the value

apis/refs/v1beta1/interface.go
```diff
@@ -0,0 +1,83 @@
...
+	u := &unstructured.Unstructured{}
+	u.SetGroupVersionKind(gvk)
+	if err := reader.Get(ctx, key, u); err != nil {
+		if apierrors.IsNotFound(err) {
+			return "", k8s.NewReferenceNotFoundError(u.GroupVersionKind(), key)
+		}
+		return "", fmt.Errorf("reading referenced %s %s: %w", gvk, key, err)
+	}
+
+	external, err := r.Resolve(ctx, reader, u)
```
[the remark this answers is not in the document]
justinsb: Oh wait, or option 3 ....

3. The method should still be `func (r *BigQueryConnectionConnectionRef) NormalizeToExternalForm(ctx context.Context, reader client.Reader, namespace string) (error) {`,  and move the logic from `NormalizeExternal` to be a helper method that `NormalizeToExternalForm` calls (so the caller has a nice simple interface)

apis/bigqueryconnection/v1alpha1/connection_reference.go
yuwenma: Changed this to return `(string, error)` .

`you can always call it...` Yeah, But I guess this is a general challenge if merging the resourceRef and realID , that the `External` (or real ID) can not guarantee to be present (necessary for realID, but optional for resourceRef).

apis/bigqueryconnection/v1alpha1/connection_reference.go
yuwenma: Ah, I see the concern. Yes, I changed this function that does not require Unstructured, and also defined a New() that require the typed resource.

pkg/controller/direct/bigqueryconnection/connection_controller.go
yuwenma: a.id.External could be empty because this is a service-generated resource we store the serviceID in external.

apis/bigqueryconnection/v1alpha1/connection_reference.go
yuwenma: fixed. Good catch. I got lost on changing the code name and input/output back and forth.

apis/bigqueryconnection/v1alpha1/connection_reference.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
+	"github.com/google/uuid"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// New builds a BigQueryConnectionConnectionRef from the ConfigConnector BigQueryConnectionConnection object.
+func New(ctx context.Context, reader client.Reader, obj *BigQueryConnectionConnection) (*BigQueryConnectionConnectionRef, error) {
```
justinsb, line 33: Nit (no need to fix) this will probably have to be NewBigQueryConnectionConnectionRef - or something like that -  as we add more refs to the package

apis/bigqueryconnection/v1alpha1/connection_reference.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		return fmt.Errorf("reading referenced %s %s: %w", BigQueryConnectionConnectionGVK, key, err)
+	}
+	// Get external from status.externalRef. This is the most trustworthy place.
+	actualExternalRef, _, err := unstructured.NestedString(u.Object, "status", "externalRef")
+	if err != nil {
+		return fmt.Errorf("reading status.externalRef: %w", err)
+	}
+	if actualExternalRef == "" {
+		return fmt.Errorf("BigQueryConnectionConnection is not ready yet.")
```
justinsb, line 152: Nit (no need to fix here) - I think we want to start returning our special "not ready" error to trigger our special backoff-with-auto-watch etc

apis/refs/v1beta1/interface.go
```diff
@@ -0,0 +1,29 @@
...
+
+package v1beta1
+
+import (
+	"context"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type ExternalNormalizer interface {
```
justinsb, line 23: Nit: Maybe this should just be called Reference :-)

pkg/controller/direct/bigqueryconnection/connection_controller.go
```diff
@@ -189,32 +147,34 @@ func (a *Adapter) Create(ctx context.Context, createOp *directbase.CreateOperati
...
+	}
 	req := &bigqueryconnectionpb.CreateConnectionRequest{
-		Parent:     a.id.Parent.String(),
+		Parent:     parent,
 		Connection: resource,
 	}
 	created, err := a.gcpClient.CreateConnection(ctx, req)
 	if err != nil {
-		return fmt.Errorf("creating Connection %s: %w", *a.id.AsExternalRef(), err)
+		return fmt.Errorf("creating Connection %s: %w", *&created.Name, err)
```
justinsb, line 160: Nit: `*&` should probably trigger a linter rule :-)

pkg/controller/direct/bigqueryconnection/connection_controller.go
```diff
@@ -263,8 +223,17 @@ func (a *Adapter) Export(ctx context.Context) (*unstructured.Unstructured, error
...
 	}
-	obj.Spec.ProjectRef = &refs.ProjectRef{Name: a.id.Parent.Project}
-	obj.Spec.Location = a.id.Parent.Location
+	parent, err := a.id.Parent()
+	if err != nil {
+		return nil, fmt.Errorf("BigQueryConnectionConnection %s parent unset: %w", a.id.External, err)
+	}
+	if parent != "" {
+		tokens := strings.Split(parent, "/")
+		if len(tokens) == 4 && tokens[0] == "projects" && tokens[2] == "locations" {
```
justinsb, line 232: Interesting ... the Parent is the last thing here that I think is awkward, and even now Parent still isn't exactly what we need for export.  Let's see how this evolves!

justinsb (approved):

apis/bigqueryconnection/v1alpha1/connection_reference.go
yuwenma: Ouch! How about `NewConnectionRef`. This function shall only be called by the BigQueryConnectionConnection controller (other reference controller calls the `NormalizeToExternalForm`)

apis/bigqueryconnection/v1alpha1/connection_reference.go
yuwenma: Wait! what did i miss?

apis/refs/v1beta1/interface.go
yuwenma: Sg.
(I maybe want to add another function to unbreak the rule https://go.dev/doc/effective_go#interface-names)

PR #2714, opened 2024-09-17
mockgcp: stage BigLake

acpana:
Parking this here. We don't have to add it right now but it cuts a step for whoever works on this next.


gemmahou:
/lgtm
/approve

Thanks! A starting point for BigLake mocks!

PR #2720, opened 2024-09-17
(Part 1)Fix diffs in compute forwarding rule: operations

gemmahou:
### Change description

Fixed diffs identified in #2695

1. Updated operation type `SetLabels` -> `setLabels`
2. `POST setLabels` returns `operation DONE`(100) instead of `RUNNING`(0)
3. Compared realGCP w/ TFcontroller and realGCP w/ direct controller in #2711, noticed a diff:
TF:       POST setLabels(return operation done) -> GET forwarding rule
direct: POST setLabels(return operation done) -> GET operation(return operation done) -> GET forwarding rule
So we should not make extra `GET operation` call if it's already done.

4. Updated the 404 error message for forwarding rule
5. Reformat `regions` field

### Tests you have done


yuwenma (commented):
/lgtm
/approve

PR #2723, opened 2024-09-18
(Part 2)Fix diffs in compute forwarding rule: Labels can only be added or modified by the setLabels method

gemmahou:
### Change description

Cherry-picked from #2711

Based on #2720, check the last commit

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #2724, opened 2024-09-18
(Part 3)Fix diffs in compute forwarding rule: remove `networkTier` field from global forwarding rule

gemmahou:
### Change description

Based on #2723. To review the change, check the last commit in this PR.

TF does not support networkTier field for global forwarding rule, It will always use GCP's default value "PREMIUM". Any value set by the user will be ignored and not sent to GCP.

Not sure if that's an ideal behavior, but let's try to match TF controller at this point.

### Tests you have done

### DIFF
--- a/pkg/controller/direct/compute/forwardingrule_controller.go
+++ b/pkg/controller/direct/compute/forwardingrule_controller.go
@@ -299,6 +299,11 @@ func (a *forwardingRuleAdapter) Create(ctx context.Context, createOp *directbase
 	var err error
 	op := &gcp.Operation{}
 	if a.id.location == "global" {
+		// todo(yuhou): TF does not support networkTier field for global forwarding rule
+		// It will always use GCP's default value, which is "PREMIUM." Any value set by the user will be ignored and not sent to GCP.
+		// To align with the TF controller, I remove this field.
+		// Ideally, direct controller should support this field and validate that the value.
+		forwardingRule.NetworkTier = nil
 		req := &computepb.InsertGlobalForwardingRuleRequest{
 			ForwardingRuleResource: forwardingRule,
 			Project:                a.id.project,
--- a/pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrulefull/_http.log
+++ b/pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrulefull/_http.log
@@ -770,7 +770,6 @@ x-goog-request-params: project=${projectId}
   ],
   "name": "computeglobalforwardingrule-${uniqueId}",
   "network": "projects/${projectId}/global/networks/${networkID}",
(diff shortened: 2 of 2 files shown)


pkg/controller/direct/compute/forwardingrule_controller.go
```diff
@@ -318,6 +318,11 @@ func (a *forwardingRuleAdapter) Create(ctx context.Context, createOp *directbase
 	var err error
 	op := &gcp.Operation{}
 	if a.id.location == "global" {
+		// todo(yuhou): TF does not support networkTier field for global forwarding rule
+		// It will always use GCP's default value, which is "PREMIUM." Any value set by the user will be ignored and not sent to GCP.
+		// To align with the TF controller, I remove this field.
+		// Ideally, direct controller should support this field and validate that the value.
+		forwardingRule.NetworkTier = nil
```
yuwenma, line 325: I don't quite get this (but I like this comment). Could you describe what the Compute GCP service does for this field? I guess I'm confused because from your description it looks like TF sets the default "PREMIUM", in which case this value would be the default "PREMIUM" as well? Or, is the "PREMIUM" set by the GCP service? If it is the latter, is it a no-op or returns errors?

Regarding the previous TF behavior, I guess it could relate to the spec auto-correct in `state-into-spec:merge` mode.

gemmahou: TF controller for global forwarding rule does not have "networkTier" field(see their [documentation](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_global_forwarding_rule#argument-reference)), that means whatever value provided by user, will be ignored(I do no see this field in the request body). If the field is unspecified, GCP will default it to "PREMIUM", so I can see the response body has this field with "PREMIUM". I hope below example can explain that clearly:

Configuration yaml:
```
metadata:
 name: globalforwardingrule
spec:
 networkTier: "anything"
 (other fields)
```
request body with TF controller:
```
POST ...globalforwardingrule...
{
(other fields)
}
```
response body with TF controller:
```
GET ...gloablforwardingrule...
{
networkTier: "PREMIUM"
(other fields)
}
```
Quote the field description: `For regional ForwardingRule, the valid values are 'PREMIUM' and 'STANDARD'. For GlobalForwardingRule, the valid value is 'PREMIUM'. If this field is not specified, it is assumed to be 'PREMIUM'. ` I think that explains why TF does not support this field for global forwarding rule and uses GCP default value.

yuwenma: Thanks for the explanation. I think 
1. we want to include the `networkTier` field in `status.observedState`. 
2. I think we don't need this line `forwardingRule.NetworkTier = nil`, with the previous TF-based `state-into-spec: merge` default, this guarantees backward compatibility for existing forwarding rules.

gemmahou:
As discussed, we do not want to drop "networkTier" field in direct controller.

PR #2725, opened 2024-09-18
(Part 4)Fix diffs in compute forwarding rule: normalize `IPAddress` in golden logs

gemmahou:
### Change description

based on #2724, check the last commit in this PR

When executing the tests against realGCP, the ip address will be generated by GCP if not provided. Normalize the ip address to be `8.8.8.8` as the fake ip we wrote in ComputeAddress mock server, to avoid diffs between mock and real.

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #2727, opened 2024-09-18
(Part 5)Fix diffs in compute forwarding rule: collect realGCP logs to verify

gemmahou:
### Change description

Based on Fixes PR part 1-4
- #2720 
- #2723 
- #2724 
- #2725 

This is the final PR for diff fixes, and this is used to verify changes in part 1-4, we do not merge this PR.

Note: we only focus on logs related to `ComputeForwardingRule` in this PR. For diffs in other dependencies, we should fix them later during their migration to direct controller.

[realGCP with direct controller](https://github.com/GoogleCloudPlatform/k8s-config-connector/commit/733eb5b9b1fa2bb89f4f1dcb12bf147db3c7a6fd): No diffs in `ComputeForwardingRule` logs

[realGCP with TF controller](https://github.com/GoogleCloudPlatform/k8s-config-connector/commit/e22c4ad457fe24c203e4453f2fcc2b0340ae54c7): No diffs in `ComputeForwardingRule` logs, except request URLs&headers, and compute endpoint versions(TF uses `beta`, but direct controller uses v1 instead)

### Tests you have done

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/regionalcomputeforwardingrule/_http.log
+++ b/pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/regionalcomputeforwardingrule/_http.log
@@ -19,11 +19,11 @@ X-Xss-Protection: 0
     "errors": [
       {
         "domain": "global",
-        "message": "address \"projects/${projectId}/regions/us-central1/addresses/computeraddress-${uniqueId}\" not found",
+        "message": "The resource 'projects/${projectId}/regions/us-central1/addresses/computeraddress-${uniqueId}' was not found",
         "reason": "notFound"
       }
     ],
-    "message": "address \"projects/${projectId}/regions/us-central1/addresses/computeraddress-${uniqueId}\" not found"
+    "message": "The resource 'projects/${projectId}/regions/us-central1/addresses/computeraddress-${uniqueId}' was not found"
   }
 }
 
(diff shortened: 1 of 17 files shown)


pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/regionalcomputeforwardingrule/_http.log
```diff
@@ -568,6 +477,7 @@ x-goog-request-params: project=${projectId}&region=us-central1
   },
   "loadBalancingScheme": "EXTERNAL",
   "name": "computeregionalforwardingrule-${uniqueId}",
+  "region": "projects/${projectId}/global/regions/us-central1",
```
gemmahou, line 480: This field appears in TF-controller log, but not in direct controller log.

We have noticed this diff before, and I think this is a "bug" in TF-controller. `region` is [output only](https://screenshot.googleplex.com/EXdhuzf6oAYFNRU.png), it does nothing here in the request body. Also the format `global/regions` looks strange, I didn't find the source code of this value.

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/regionalcomputeforwardingrule/_http.log
```diff
@@ -658,101 +652,16 @@ X-Xss-Protection: 0
...
   "name": "computeregionalforwardingrule-${uniqueId}",
   "networkTier": "PREMIUM",
-  "region": "https://www.googleapis.com/compute/v1/projects/${projectId}/regions/us-central1",
-  "selfLink": "https://www.googleapis.com/compute/v1/projects/${projectId}/regions/us-central1/forwardingRules/computeregionalforwardingrule-${uniqueId}",
-  "target": "https://www.googleapis.com/compute/v1/projects/${projectId}/regions/us-central1/targetVpnGateways/computetargetvpngateway-${uniqueId}"
-}
-
----
-
-POST https://compute.googleapis.com/compute/v1/projects/${projectId}/regions/us-central1/forwardingRules/${forwardingRuleID}/setLabels
```
gemmahou: It's strange that there's no "setLabels" request with TF-controller(update `label-one: value-one` to `label-one: value-two`). I tried to update labels with the TF-controller and checked `gcloud` output, the label value did get updated. But there's no update labels request in the log. 

While, I think the direct-controller log is correct, we expect this "setLabels" request to update the label value.

PR #2728, opened 2024-09-18
Updating makefile to get latest changes for cloud/kms api

nb-goog:
### To sync CloudKMS autokey proto from googleapi

Working on supporting cloudkms autokey feature via KCC. As a first step syncing cloud kms api proto.

### Tests you have done
- [X] Run `make ready-pr` to ensure this PR is ready for review. (done)
- [ ] Perform necessary E2E testing for changed resources. (NA)


yuwenma:
/lgtm
/approve

Thanks for the change, @nb-goog! Ideally we would like to see the MockGCP change goes together with the Make proto update so we can see the differences. But I'm okay if you want to do that in follow up PRs.

PR #2729, opened 2024-09-18
fix: Compute service "External" value allows selfLink"

### LINKED ISSUE #2701: SQLInstance fails to update with self-link for privateNetworkRef
### Checklist

- [X] I did not find a related open issue.
- [X] I did not find a solution in the troubleshooting guide: (https://cloud.google.com/config-connector/docs/troubleshooting)
- [X] If this issue is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Bug Description

When upgrading to version 1.122.0 we found that the `settings.ipConfiguration.privateNetworkRef.external` field for a SQLInstance seems to have changed.  According to the message, the format is required to be `projects/<project id>/global/networks/<networkid>`.  Previously, it was just a self-link to the network to be used.  We do not and have not used a global network for our VPC and I'm not sure why that's required now.  The issue seems to be [here](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/sql/normalize.go#L245) which leads [here](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/refs/v1beta1/computerefs.go#L53-L65).

There doesn't seem to be any mention of this in the release notes either.  Please advise on what to do if this is not a bug.

### Additional Diagnostic Information

This happened on all our clusters running different versions of k8s.  Rolling back to version 1.121.0 brought everything back.

v1.29.7-gke.1274000

### Config Connector Version

1.122.0

### Config Connector Mode

namespaced mode (default)

### Log Output

_No response_

### Steps to reproduce the issue

1. Install ConfigConnector v1.121.0
1. Create a SQLInstance with an external privateNetworkRef to a shared VPC
1. Upgrade ConfigConnector to v1.122.0
1. Wait until after next reconcialiation, or alternatively apply an annotation/label to force a reconciliation, then see error on SQLInstance object

### YAML snippets

_No response_

yuwenma:
### Change description

Fixes #2701

### Tests you have done

### DIFF
--- a/apis/refs/v1beta1/computerefs.go
+++ b/apis/refs/v1beta1/computerefs.go
@@ -37,7 +37,8 @@ func fixStaleExternalFormat(external string) string {
 }
 
 type ComputeNetworkRef struct {
-	/* The compute network selflink of form "projects/<project>/global/networks/<network>", when not managed by Config Connector. */
+	// A reference to an externally managed Compute Network resource.
+	// Should be in the format `projects/<projectID>/global/networks/<network>`.
 	External string `json:"external,omitempty"`
 	/* The `name` field of a `ComputeNetwork` resource. */
 	Name string `json:"name,omitempty"`
--- a/apis/refs/v1beta1/computerefs.go
+++ b/apis/refs/v1beta1/computerefs.go
@@ -27,6 +27,17 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// fixStaleExternalFormat converts the "External" reference field to the right format if a SelfLink value is used.
+// This guarantees the backward compatibility for Compute Beta resources.
+func fixStaleExternalFormat(external string) string {
+	tokens := strings.SplitN(external, "/projects/", 2)
+	if len(tokens) == 2 {
+		external = strings.TrimPrefix(external, tokens[0])
+	}
+	external = strings.TrimPrefix(external, "/")
+	return external
+}
+
 type ComputeNetworkRef struct {
 	/* The compute network selflink of form "projects/<project>/global/networks/<network>", when not managed by Config Connector. */
 	External string `json:"external,omitempty"`
@@ -55,6 +66,7 @@ func ResolveComputeNetwork(ctx context.Context, reader client.Reader, src client
 			return nil, fmt.Errorf("cannot specify both name and external on computenetwork reference")
 		}
 
+		ref.External = fixStaleExternalFormat(ref.External)
 		tokens := strings.Split(ref.External, "/")
 		if len(tokens) == 5 && tokens[0] == "projects" && tokens[2] == "global" && tokens[3] == "networks" {
 			return &ComputeNetwork{
@@ -122,6 +134,7 @@ func ResolveComputeSubnetwork(ctx context.Context, reader client.Reader, src cli
 		if ref.Name != "" {
 			return nil, fmt.Errorf("cannot specify both name and external on computenetwork reference")
 		}
+		ref.External = fixStaleExternalFormat(ref.External)
 
 		tokens := strings.Split(ref.External, "/")
 		if len(tokens) == 6 && tokens[0] == "projects" && tokens[2] == "regions" && tokens[4] == "subnetworks" {
--- a/apis/refs/v1beta1/computerefs.go
+++ b/apis/refs/v1beta1/computerefs.go
@@ -30,10 +30,8 @@ import (
 // fixStaleExternalFormat converts the "External" reference field to the right format if a SelfLink value is used.
 // This guarantees the backward compatibility for Compute Beta resources.
 func fixStaleExternalFormat(external string) string {
-	tokens := strings.SplitN(external, "/projects/", 2)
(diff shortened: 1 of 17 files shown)


yuwenma:
This is the immediate fix. We need https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/2730 to prevent this from happen for other Compute resource migration.

apis/refs/v1beta1/computerefs.go
```diff
@@ -55,6 +66,7 @@ func ResolveComputeNetwork(ctx context.Context, reader client.Reader, src client
 			return nil, fmt.Errorf("cannot specify both name and external on computenetwork reference")
 		}
 
+		ref.External = fixStaleExternalFormat(ref.External)
 		tokens := strings.Split(ref.External, "/")
 		if len(tokens) == 5 && tokens[0] == "projects" && tokens[2] == "global" && tokens[3] == "networks" {
```
justinsb, line 71: Should we handle it here instead?  e.g.

external := ref.External
external = strings.TrimPrefix(external, "https://www.googleapis.com/compute/v1/")
external = strings.TrimPrefix(external, "https://www.googleapis.com/compute/v1beta1/")
tokens := strings.Split(external, "/")
		if len(tokens) == 5 && tokens[0] == "projects" && tokens[2] == "global" && tokens[3] == "networks" {
		```

justinsb: OTOH I do like the idea that you have a nice function here that we can share with ComputeNetwork, ComputeSubnetwork etc etc and that we can write-back to ref.External when we are following that pattern...

yuwenma: So I searched and checked around, I *think* all Compute Beta resource allows selfLink as "External" value, some special case allows IP Address as External value as well. I think when migrating this to the service API dir, we could have a top level function to guarantee all resource allows the old External value.

apis/refs/v1beta1/computerefs.go
```diff
@@ -27,6 +27,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// fixStaleExternalFormat converts the "External" reference field to the right format if a SelfLink value is used.
+// This guarantees the backward compatibility for Compute Beta resources.
+func fixStaleExternalFormat(external string) string {
+	external = strings.TrimPrefix(external, "https://www.googleapis.com/compute/v1/")
+	external = strings.TrimPrefix(external, "https://www.googleapis.com/compute/v1beta1/")
+	external = strings.TrimPrefix(external, "/")
```
yuwenma, line 35: allow user to specify something like `/projects/<projectID>/global/networks/<networkID>'. This is trivial but with a nicer experience.

justinsb (approved):

apis/refs/v1beta1/computerefs.go
```diff
@@ -27,6 +27,15 @@ import (
...
+	return external
+}
+
 type ComputeNetworkRef struct {
 	/* The compute network selflink of form "projects/<project>/global/networks/<network>", when not managed by Config Connector. */
```
gemmahou, line 40: Shall we update the comment? 

"projects/<project>/global/networks/<network>" is not a selfLink, it's a preferred value of KCC. We can simply say "The external compute network of form "projects/<project>/global/networks/<network>"" or similar.

apis/refs/v1beta1/computerefs.go
gemmahou: I'm unsure if we want to fix in this way. As discussed before, we do want to have strict format restriction on `external` field, at least for new resources.  I'm afraid allowing stale formats would break this expectation? 

I feel we may want to have a special handling for backward capability(existing Beta resources), write a separate network resolver accepts other formats, like what I had for forwarding rule([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/5b1307cd3451ec9fe3c252707f2a2ff32deaa3d9/pkg/controller/direct/compute/refs.go)) or reuse the `ResolveComputeNetwork` in SQLInstance.

yuwenma: Are you talking about distinguishing between existing object and new object? Yes, I like that idea. Though this would require each resource that references ComputeNetwork to change the ResolveRef between "create" and other functions. I don't know how feasible that would be. 
 
Another question is how much do we know about the abnormal format of the Compute resource "External" field.

apis/refs/v1beta1/computerefs.go
yuwenma: Good catch.

apis/refs/v1beta1/computerefs.go
gemmahou: I do not have a 100% sure answer for the first part, but since we just started the Beta->direct migration for a few resources, if we have a standard to follow now, then there will not be too much work.

As far as I know, allowing "selfLink" as External is very common for Compute resources and also for other resources.

justinsb (approved):

Issue #2706, opened 2024-09-16
SQLInstance Error 400: The incoming request contained invalid data., invalidRequest


PurseChicken:
### Checklist

- [X] I did not find a related open issue.
- [X] I did not find a solution in the troubleshooting guide: (https://cloud.google.com/config-connector/docs/troubleshooting)
- [X] If this issue is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Bug Description

After updating the config connector operator to 1.122.0, I am now seeing the following error on all SQLInstance manifests:

`Update call failed: error updating: updating SQLInstance my-sql-instance failed: googleapi: Error 400: The incoming request contained invalid data., invalidRequest`

My SQLInstance manifests all look very similar to the following:

```
apiVersion: sql.cnrm.cloud.google.com/v1beta1
kind: SQLInstance
metadata:
  annotations:
    cnrm.cloud.google.com/deletion-policy: abandon
    cnrm.cloud.google.com/project-id: my-gcp-project
  name: my-gcp-project-my-sql-instance
  namespace: my-namespace
spec:
  databaseVersion: POSTGRES_16
  region: us-west3
  resourceID: my-sql-instance
  settings:
    availabilityType: REGIONAL
    backupConfiguration:
      enabled: true
      pointInTimeRecoveryEnabled: true
    databaseFlags:
      - name: cloudsql.iam_authentication
        value: 'on'
    diskSize: 100
    ipConfiguration:
      requireSsl: true
    maintenanceWindow:
      day: 1
      hour: 4
      updateTrack: stable
    tier: db-custom-2-8192
```

### Additional Diagnostic Information

I have not been able to find any other logs that point to what the issue could be. That said, I have this same issue for 4 different SQLInstance manifests.

### Kubernetes Cluster Version

v1.30.3-gke.1639000

### Config Connector Version

1.122.0

### Config Connector Mode

namespaced mode (default)

### Log Output

[...]

PR #2731, opened 2024-09-18
fix: Enable TF-based reconciler for SQLInstances without clone source

jasonvigil:
Fixes https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/2706

### DIFF
--- a/pkg/controller/registration/registration_controller.go
+++ b/pkg/controller/registration/registration_controller.go
@@ -248,21 +248,12 @@ func registerDefaultController(r *ReconcileRegistration, config *config.Controll
 
 		// If we have a choice of controllers, construct predicates to choose between them
 		if hasDirectController && (hasTerraformController || hasDCLController) {
-			reconcileGate := registry.GetReconcileGate(gvk.GroupKind())
-			if reconcileGate != nil {
+			if reconcileGate := registry.GetReconcileGate(gvk.GroupKind()); reconcileGate != nil {
 				// If reconcile gate is enabled for this gvk, generate a controller-runtime predicate that will
 				// run the direct reconciler only when the reconcile gate returns true.
 				useDirectReconcilerPredicate = kccpredicate.NewReconcilePredicate(r.mgr.GetClient(), gvk, reconcileGate)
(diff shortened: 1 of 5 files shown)


yuwenma:
For the cases we found in #2706, I think we also want to add the support in the direct SQLInstance reconciler to remove system labels, and recognize/propagate the 0 int value. But I'm good if you want to do that in separate PRs.

jasonvigil:
> For the cases we found in #2706, I think we also want to add the support in the direct SQLInstance reconciler to remove system labels, and recognize/propagate the 0 int value. But I'm good if you want to do that in separate PRs.

Yeah, this is just a quick fix to unblock the release.

pkg/controller/registration/registration_controller.go
```diff
@@ -248,21 +248,12 @@ func registerDefaultController(r *ReconcileRegistration, config *config.Controll
...
-			if !hasTerraformController && !hasDCLController {
-				// We're always going to use the direct reconciler
-				useDirectReconcilerPredicate = nil
-				useLegacyPredicate = nil
-			}
-
-			if (hasTerraformController || hasDCLController) && useDirectReconcilerPredicate == nil {
+			} else {
 				logger.Error(fmt.Errorf("no predicate where we have multiple controllers"), "skipping direct controller registration", "group", gvk.Group, "version", gvk.Version, "kind", gvk.Kind)
 				hasDirectController = false
```
yuwenma, line 258: Previously we decouple the CRDs and controllers and let the labels on CRD to choose the controller. It only moves from TF to direct but not vice versa (which is an important precondition).  

Now the controller decision is switched to the CR object. 

I like the reconcile gate idea overall (except that it is very challenge to figure out which reconciler is being used, that could make triaging more difficult). I think my concern is mainly that the reconcile gate has changed the concept about some basic concept about the direct registry and label based reconciliations, that is not covered by any of our guides, tools, and test.

yuwenma (commented):
/lgtm
/approve

I'd like to have this PR goes in to unblock users and continue the reconcile gate discussion.

PR #2733, opened 2024-09-18
chore: build images with git tag, if we match exactly

justinsb:
This means the image tag matches the git tag.


yuwenma (commented):
/lgtm
/approve

PR #2734, opened 2024-09-19
chore: create simple script to deploy the operator to a kube cluster

justinsb:
We also try to create a storage bucket during development of this
script - we aren't setting up IAM permissions so it won't work yet.
Once this is more real we probably want to stop auto-creating storage
buckets.

Example usage: VERSION=7b7142e dev/tasks/deploy-to-kube


yuwenma (commented):
/lgtm
/approve

Awesome! And thanks for the comments.

PR #2735, opened 2024-09-19
Set default values to immutable and optional spec fields for certain kinds

maqiuyujoyce:
### Change description

Shift-left the defaulting of the following spec fields to the TF controller fix the unexpected diffs in re-reconciliation when there is no change in the CR:

* `spec.location` in `CloudBuildTrigger`.
* `spec.initialGroupConfig` in `CloudIdentityGroup`.
* `spec.database` in `FirestoreIndex`.

This looks like an old bug and was only exposed when `state-into-spec: absent` became the default configuration.

More context: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2704#issue-2529378867

### Tests you have done


yuwenma:
/lgtm
/approve

Thank you!

PR #2739, opened 2024-09-19
support bigqueryconnection cloudsql connection

600lyy:
### Change description
Add support for bigqueryconnection cloudsql connection

Run E2E test again realGCP:

```
KCC_USE_DIRECT_RECONCILERS=BigQueryConnectionConnection hack/record-gcp fixtures/cloudsqlconnectionbasic
```

- [] Run `make ready-pr` to ensure this PR is ready for review.


apis/refs/v1beta1/sqlinstanceref.go
```diff
@@ -22,3 +34,92 @@ type SQLInstanceRef struct {
 	/* The `namespace` field of a `SQLInstance` resource. */
 	Namespace string `json:"namespace,omitempty"`
 }
+
+type SQLInstance struct {
+	ProjectID       string
+	Location        string
+	SQLInstanceName string
+	ConnectionName  string
```
yuwenma, line 42: Normally the Reference struct only holds the fundamental identifies (parent + instance ID). 

Does connection always in the form of `${ProjectID}:${Location}:${SQLInstanceName}`? if so, we can build the ConnectionName from the other three and don't need this field.

pkg/controller/direct/bigqueryconnection/connection_mapping.go
```diff
@@ -36,15 +36,49 @@ func CloudResourcePropertiesSpec_FromProto(mapCtx *direct.MapContext, in *pb.Clo
...
+	}
+	return out
+}
+
+func CloudSqlPropertiesSpec_FromProto(mapCtx *direct.MapContext, in *pb.CloudSqlProperties) *krm.CloudSqlPropertiesSpec {
+	if in == nil {
+		return nil
+	}
+	out := &krm.CloudSqlPropertiesSpec{}
+	// out.InstanceID = direct.LazyPtr(in.InstanceId)
```
yuwenma, line 62: We want this. You can use assign the value to External 
```
out.InstanceRef = &SQLInstanceRef{
   External: ...
}
```

yuwenma (commented):
/approve

Great job! I left some minor comments, and it will be ready to go.

apis/refs/v1beta1/sqlinstanceref.go
600lyy: Yeah, makes sense. Even though the form would change in future, I think it still can be built based on other fields.

yuwenma (commented):
/lgtm

Thanks! All looks good

PR #2740, opened 2024-09-19
Fix the default field name for CloudIdentityGroup

maqiuyujoyce:
### Change description

This is to fix the bug introduced in #2735. The field name should be in snake case instead of camel case.

### Tests you have done

- [Pending] Perform necessary E2E testing for changed resources.


maqiuyujoyce:
Verified locally that the test cases for CloudIdentityGroup all passed.

/hold cancel

yuwenma (commented):
/lgtm
/approve

PR #2742, opened 2024-09-19
fix: bigqueryconnectionconnection misuse projectNum for projectID

yuwenma:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/pkg/controller/direct/bigqueryconnection/connection_controller.go
+++ b/pkg/controller/direct/bigqueryconnection/connection_controller.go
@@ -166,7 +166,14 @@ func (a *Adapter) Create(ctx context.Context, createOp *directbase.CreateOperati
 	if mapCtx.Err() != nil {
 		return mapCtx.Err()
 	}
-	status.ExternalRef = &created.Name
+
+	tokens := strings.Split(created.Name, "/")
+	parent, err = a.id.Parent()
+	if err != nil {
+		return err
+	}
+	externalRef := parent + "/connections/" + tokens[5]
+	status.ExternalRef = &externalRef
 	return setStatus(u, status)
 }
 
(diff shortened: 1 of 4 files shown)


pkg/controller/direct/bigqueryconnection/connection_controller.go
```diff
@@ -166,7 +166,14 @@ func (a *Adapter) Create(ctx context.Context, createOp *directbase.CreateOperati
 	if mapCtx.Err() != nil {
 		return mapCtx.Err()
 	}
-	status.ExternalRef = &created.Name
+
```
yuwenma, line 169: I"ll refactor this code with resource ref in a follow up PR.

pkg/controller/direct/bigqueryconnection/connection_controller.go
```diff
@@ -166,7 +166,14 @@ func (a *Adapter) Create(ctx context.Context, createOp *directbase.CreateOperati
...
+	tokens := strings.Split(created.Name, "/")
+	parent, err = a.id.Parent()
```
justinsb, line 171: Idea for future: I'm wondering if we should introduce a type ProjectLocationRef.  It wouldn't be used by any API (at least I haven't seen one yet!), but it might make Parent() easier.

pkg/controller/direct/bigqueryconnection/connection_controller.go
```diff
@@ -166,7 +166,14 @@ func (a *Adapter) Create(ctx context.Context, createOp *directbase.CreateOperati
...
+	if err != nil {
+		return err
+	}
+	externalRef := parent + "/connections/" + tokens[5]
```
justinsb, line 175: Another idea for future: maybe we should parse `created.Name` into a BigQueryConnectionConnectionRef, and then use the canonical form from BigQueryConnectionConnectionRef (I assume that's what we're doing here)

justinsb (approved):

pkg/controller/direct/bigqueryconnection/connection_controller.go
yuwenma: Yes, agree! That's also how I want to refactor this code.

PR #2751, opened 2024-09-19
test: add full test case for regional forwarding rule and mock servicedirectory

gemmahou:
mock

### Change description

- Add full test case for regional forwarding rule(keep create.yaml and update.yaml identical)
- Add mock for dependency servicedirectory

### Tests you have done


yuwenma (commented):
/approve
/lgtm

Thanks! Would be even better if you can upload the create and update.yaml diff in separate git-commits (e.g. 1st commit making create.yaml and update.yaml the same, 2nd commit update.ymal shows the mutable fields)

PR #2752, opened 2024-09-19
test deflake: Abandon feature resource in dedicated test project

ziyue-101:
### Change description

The gkehub tests are running in special project. We should abandon the global resource to prevent interfering with other tests running in parallel.

The change is similar to what we did for the featuremembership tests

### Tests you have done


maqiuyujoyce:
Hi @ziyue-101 , why do we want to abandon this resource? I checked in our repo and it isn't reused in other test cases.

Usually we don't want to abandon the target resource in the test case: We'll validate there is indeed an update happening. to the KRM resource and if we do abandon and acquisition, there doesn't always be an update.

ziyue-101:
> Hi @ziyue-101 , why do we want to abandon this resource? I checked in our repo and it isn't reused in other test cases.
> 
> Usually we don't want to abandon the target resource in the test case: We'll validate there is indeed an update happening. to the KRM resource and if we do abandon and acquisition, there doesn't always be an update.

The test uses [a special project](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/a1ffee0774b3a018301a2b7f0525e8402c0f5bd4/pkg/test/resourcefixture/testdata/basic/gkehub/v1beta1/gkehubfeature/acmfeature/dependencies.yaml#L23) where other Gke hub tests are running in parallel. If we delete the feature(a global resource within a prject) here, it will interfere with other tests also rely on the feature in this project.

maqiuyujoyce:
> if we delete the feature(a global resource within a prject) here, it will interfere with other tests also rely on the feature in this project.

Right now, I don't see it (acm feature in this specific project) being a dependency of other test cases. When the acm feature is used as the dependency, it seems to be in a different test project (https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/v1.123.1/pkg/test/resourcefixture/testdata/basic/gkehub/v1beta1/gkehubfeaturemembership/fullacmgkehubfeaturemembership/dependencies.yaml#L22) So I feel it shouldn't be the cause of any flake. Or did I miss anything?

ziyue-101:
@maqiuyujoyce you are right. They belong to different project. I guess we no longer need the Pr.

PR #2754, opened 2024-09-19
test: add fuzz test for BigQueryDataTransferConfigSpec

jingyih:


pkg/test/fuzz/generate.go
```diff
@@ -105,6 +105,13 @@ func fillWithRandom0(t *testing.T, randStream *rand.Rand, msg protoreflect.Messa
 					v := randomString(randStream)
 					mapVal.Set(protoreflect.ValueOf(k).MapKey(), protoreflect.ValueOf(v))
 				}
+			case "string->message":
+				if field.FullName() == "google.protobuf.Struct.fields" && field.MapValue().Message().FullName() == "google.protobuf.Value" {
+					// currently this is converted to "map[string]string" in "BigQueryDataTransferConfig"
+					// TODO: fill in random strings
```
jingyih, line 111: Not sure how to fill in random strings for this field yet. So currently `.spec.params" is not actually tested in fuzz.

yuwenma: "fill in random strings" --> Do you mean the random string map? Can we do something similar to "string -> string" but replace the "v" with a new mapVal?

yuwenma:
/lgtm
/approve
/hold

Feel free to unhold and add the random string map in a follow up PR

pkg/test/fuzz/generate.go
jingyih: I was experimenting with it but having some trouble. The fuzz test simply exit with "fuzzing process hung or terminated unexpectedly: exit status 255". I don't feel like a single field should block adding the test.

PR #2759, opened 2024-09-20
Add test cases for SQLInstance settings

jasonvigil:
Also, refactor mapping functions for improved readability / maintainability, and to make them more similar to generated mappings.


yuwenma (commented):
/approve
/lgtm

Thanks! Would be even better if you can upload the create and update.yaml diff in separate git-commits (e.g. 1st commit making `create.yaml` and `update.yaml` the same, 2nd commit `update.ymal` shows the mutable fields)

PR #2760, opened 2024-09-20
dev:tool: small fix for interface type

acpana:
When running the mapper tool (example: 

```
go run main.go generate-mapper \
     --service google.cloud.bigquery.analyticshub.v1 \
     --proto-source-path ../proto-to-mapper/build/googleapis.pb \
     --output-dir $REPO_ROOT/pkg/controller/direct/ \
     --api-dir $REPO_ROOT/apis \
     --api-go-package-path $REPO_ROOT/apis/bigqueryanalyticshub \
     --api-version "bigqueryanalyticshub.cnrm.cloud.google.com/v1alpha1"
```
), I ran into the following err:

```
...
inspecting go code: loading package "<MY_PATH>/kcc/8/k8s-config-connector/apis/refs/v1beta1": inspecting package "v1beta1": unhandled type spec in "ExternalNormalizer": *ast.InterfaceType, &{Interface:24559 Methods:0xc00d5d7620 Incomplete:false}
exit status 1
```

### DIFF
--- a/dev/tools/controllerbuilder/pkg/gocode/ast.go
+++ b/dev/tools/controllerbuilder/pkg/gocode/ast.go
@@ -145,9 +145,7 @@ func (p *Package) inspect(packageName string, pkg *ast.Package) error {
 			case *ast.Ident:
 				// type alias
 			case *ast.InterfaceType:
-				if n.Name.String() == "ExternalNormalizer" {
-					// skip this for now
-				}
+				// always skip, nothing to generate for interface
 			default:
 				errs = append(errs, fmt.Errorf("unhandled type spec in %q: %T, %+v", n.Name, n.Type, n.Type))
 			}
--- a/dev/tools/controllerbuilder/pkg/gocode/ast.go
+++ b/dev/tools/controllerbuilder/pkg/gocode/ast.go
@@ -144,6 +144,10 @@ func (p *Package) inspect(packageName string, pkg *ast.Package) error {
 				}
 			case *ast.Ident:
 				// type alias
+			case *ast.InterfaceType:
+				if n.Name.String() == "ExternalNormalizer" {
+					// skip this for now
+				}
 			default:
 				errs = append(errs, fmt.Errorf("unhandled type spec in %q: %T, %+v", n.Name, n.Type, n.Type))
 			}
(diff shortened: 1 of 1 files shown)


dev/tools/controllerbuilder/pkg/gocode/ast.go
```diff
@@ -144,6 +144,10 @@ func (p *Package) inspect(packageName string, pkg *ast.Package) error {
 				}
 			case *ast.Ident:
 				// type alias
+			case *ast.InterfaceType:
```
jingyih, line 147: I see. So it is trying to parse 
https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/a1ffee0774b3a018301a2b7f0525e8402c0f5bd4/apis/refs/v1beta1/interface.go#L23

Mapper do not need to parse Go interface. I think we can always skip on *ast.InterfaceType.

yuwenma: I'm wondering if we shall only load the packages that match the flag Service/Kind. I guess the loading of `./refs` may come from that the resource has references from `./refs`. In which case, the struct always contains `Name, Namespace, External` and no further dependencies. @jingyih WDYT?

jingyih: @acpana What happens if we pass `--api-dir $REPO_ROOT/apis/bigqueryanalyticshub` to the mapper generator?
(I assume you have the generated types under $REPO_ROOT/apis/bigqueryanalyticshub)

My understanding is that all the types for `bigqueryanalyticshub` should be under "$REPO_ROOT/apis/bigqueryanalyticshub", there is no need to parse other subfolders under "$REPO_ROOT/apis". Mapping functions for referenced fields are handled manually anyway.

jingyih: Thanks for confirmation! Let's skip "interface" for now. If we need it in future, we can always add it back!

jingyih (approved):
/lgtm

dev/tools/controllerbuilder/pkg/gocode/ast.go
yuwenma: Thanks for the suggestion, @jingyih! I'd like to have the PR goes in so it can unblock others who run `dev/tasks/generate.sh` or the generate-mapper tool as well. Feel free to propose new approaches for mapper generation/update.

yuwenma:
/lgtm
/approve

Thanks, @acpana!

jingyih (approved):
/lgtm

PR #2762, opened 2024-09-21
Merge the returned additionalExperiments into DataflowJob after successful reconciliation

maqiuyujoyce:
### Change description

Fixed  the following issue in DataflowJob exposed by `state-into-spec: absent`.

Issue:
Re-reconciliation of a DataflowJob that has `spec.additionalExperiments` configured and doesn't have any changes in the manifest will result in the update of an existing job. The DataflowJob resource may stuck in failed state if the update fails. Because the job is an imperative resource that takes time to run/to change between different states, the chances the update doesn't go through properly might be high.

Root cause:
Value of `additionalExperiments` is updated by the API during creation even if it is originally configured by the user. This caused permdiff in between the desired state and the live state unless we match the desired state to the live state.

Fix:
Merge the returned live state of `spec.additionalExperiments` into the DataflowJob CR after successful reconciliation.

### Tests you have done


pkg/krmtotf/tftokrm.go
```diff
@@ -233,7 +234,16 @@ func ResolveSpecAndStatusWithResourceID(resource *Resource, state *terraform.Ins
 func resolveDesiredStateInSpecAndObservedStateInStatus(resource *Resource, state *terraform.InstanceState) (
 	spec map[string]interface{}, status map[string]interface{}) {
 	spec = deepcopy.MapStringInterface(resource.Spec)
-	_, status = GetSpecAndStatusFromState(resource, state)
+	returnedSpec, status := GetSpecAndStatusFromState(resource, state)
+	if resource.Kind == "DataflowJob" {
```
gemmahou, line 238: nit: we can add a comment of this special handling in the code for future reference.

maqiuyujoyce: Done.

PR #2763, opened 2024-09-23
docs: release note for 1.123

yuwenma:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/docs/releasenotes/release-1.123.md
+++ b/docs/releasenotes/release-1.123.md
@@ -1,12 +1,8 @@
-# v1.123.0
+# v1.123.1
 
-** This version is not yet released; this document is gathering release notes for the future release **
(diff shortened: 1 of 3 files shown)


docs/releasenotes/release-1.123.md
```diff
@@ -1,12 +1,6 @@
-# v1.123.0
+# v1.123.1
```
justinsb, line 1: If we're planning on not releasing 1.123.0 (or recommending 1.123.1 instead) a little note here saying "We do not recommend using 1.123.0, as it had an issue around ...".  Or if we never released it, saying "1.123.0 contained a regression that was caught during pre-release testing, and was not released.".  Otherwise people will wonder what happened to 1.123.0 !

justinsb:
/approve
/lgtm

/hold
(Hold in case you want to add a note about 1.123.0, but we can also do that in another PR)

docs/releasenotes/release-1.123.md
yuwenma: Thanks for the guide!

justinsb (approved):

PR #2771, opened 2024-09-23
test: second PR to update mutable fields for bigQueryDataset

xiaoweim:
Run the test against real GCP with annotation to set the state-into-spec to both absent and merge, without setting the storagebillingmodel.
| state-into-spec | storagebillingmodel in create.yaml | storagebillingmodel in golden object yaml |
| ----------------|----------------------------------- | ------------------------------------------| 
| absent                | Not set                                               | Not populated                                               |
| merge                 | Not set                                               | Not populated                                               |

If annotations are not set explicitly in the `create.yaml` file, it will default to absent and in the golden object yaml file, the storagebillingmodel will not be defaulted to any value by the controller as well.

In the request sent to the server recorded in the http log, storagebillingmodel is also not set.


yuwenma:
Thanks for verifying. This means we do not need any additional handling for the field.

PR #2776, opened 2024-09-24
chore: let Direct beta resource to use `status.observedState`

yuwenma:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/pkg/k8s/allowlist_test.go
+++ b/pkg/k8s/allowlist_test.go
@@ -16,7 +16,6 @@ package k8s_test
 
 import (
 	"fmt"
-	"slices"
 	"strings"
 	"testing"
 
@@ -27,12 +26,6 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/servicemapping/servicemappingloader"
 )
-
-// TODO(yuwenma): This is a temp fix. We should use a more generic approach.
-var SkipServiceLoaderList = []string{
-	"CloudBuildWorkerPool",
-	"SecretManagerSecret",
-}
 
 func TestSupportsStateIntoSpecMerge(t *testing.T) {
 	tests := []struct {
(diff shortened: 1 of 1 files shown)


pkg/k8s/allowlist_test.go
```diff
@@ -130,12 +121,6 @@ func TestOutputOnlyFieldsAreUnderObservedState(t *testing.T) {
 					t.Errorf("'status.observedState' doesn't exist, but it should")
 				}
 
-				// If both shouldHaveObservedState and mayHaveObservedState are
```
maqiuyujoyce, line 133: I think this check is still reasonable. Instead of removing it, can we add a filter in the beginning of the for loop for the crds like below?

```
if crd.GetLabels()[crdgeneration.TF2CRDLabel] != "true" && crd.GetLabels()[crdgeneration.Dcl2CRDLabel] != "true" {
	continue
}
```

maqiuyujoyce: Discussed offline. Right now TF-based CRDs may have SciFi features and will make this filtering not easy to implement. So it's fine to remove this check for now.

Though could you update the comment for posterity?

```
// For a TF-based CRD, if both shouldHaveObservedState and
// mayHaveObservedState are false, the CRD is not supposed to have
// 'status.observedState'.
// But because some TF-based CRD starts introducing SciFi features,
// and it's hard to tell whether a TF-based CRD also supports SciFi,
// we decided to skip this check to unblock the SciFi features.
```

yuwenma: I think the existing logic could be confusing for people who work on migrating the TF/DCL based resource to the Direct approach, especially that the tf2crd or dcl2crd labels won't be removed until the migration is fully done. This means the contributors cannot add the `observedState` field when they need to during the development. 

What problems will we hit if we drop this check? If there's no known issues, I would suggest we drop the check. If there is, I think we shall update the comment to tell why this check is needed.

maqiuyujoyce: > What problems will we hit if we drop this check?

The logic to manage observedState for pure TF-based resources is quite complicated and may cause confusion if we simply drop it. Considering we have more than a hundred v1beta1 TF-based resources, I'd prefer we can at least keep the information for this removed check.

> If there is, I think we shall update the comment to tell why this check is needed.

The check is needed because it helps elaborate the implementation details/decisions for observed state. Created a separate PR based on our discussion: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2781

yuwenma:
Fixed in #2781

PR #2779, opened 2024-09-24
mockgcp: we don't need the alloydb maintenance_update_policy patch any more

justinsb:
maintenance_update_policy has now been published in the public protos.

Keep it around as a comment though, so we know how to use this for next time - it isn't trivial.


yuwenma (commented):
/lgtm
/approve

+100

PR #2780, opened 2024-09-24
mockgcp: fix makefile for KMS OpenAPI -> proto

justinsb:
Fix some typo-class error.


yuwenma (commented):
/lgtm
/approve

Thanks! I think @nb-goog also has a fix for this. You win!

PR #2783, opened 2024-09-24
tools: add default values to type and mapper generator flags

jingyih:
Add default values to some of the flags used by type generator and mapper generator. So that users do not need to specify them.


yuwenma (commented):
/lgtm
/approve

PR #2784, opened 2024-09-24
feat: Outbound public IP support for AlloyDB instances

nancynh:
### Change description

Adds KCC support for AlloyDB instances to have an outbound public IP address. Specifically adds the field `enableOutboundPublicIp` to the already existing `networkConfig`. This is for an upcoming feature outbound public IP feature.

Fixes b/361373492 and b/352047661

### Tests you have done

Ran the e2e tests ([output](https://paste.googleplex.com/6307481652559872)) and saw that it correctly [created the instance with outbound public ip enabled](https://screenshot.googleplex.com/AkvbXbQFNLFhnfu) and then [disabled it](https://screenshot.googleplex.com/5WdGGpCuTnUyqQd)


pkg/test/resourcefixture/testdata/basic/alloydb/v1beta1/alloydbinstance/fullalloydbinstance/update.yaml
```diff
@@ -28,4 +28,5 @@ spec:
   resourceID: alloydbinstance${uniqueId}
   networkConfig:
     enablePublicIp: false
+    enableOutboundPublicIp: false
```
maqiuyujoyce, line 31: In the TF-based controller logic, setting a boolean field to `false` means unset this field. Have you built the code locally, applied this manifest, and verified that the behavior is expected (i.e. even unsetting this field will set this field to false)?

nancynh: It looks like when I unset this field, it does not actually set it to false - it instead maintains whatever value was set before. I think this is because when we remove this field it's not included in the update request and update mask. 

So, the AlloyDB update API will only change the fields that are in the update mask because it assumes the user only wants those fields to be changed.

maqiuyujoyce: Hi @nancynh , not sure if I misunderstand anything, or there might be some confusion in between the Config Connector resource behavior versus the GCP resource behavior.

In Config Connector, at least for AlloyDBInstance (runs on top of the TF-based controller):
1. Setting a boolean field to `false` in the AlloyDBInstance resource, means unsetting the corresponding field in the underlying GCP AlloyDB instance.
2. Unsetting a boolean field in the AlloyDBInstance, means stopping managing the value of the corresponding field in the underlying GCP AlloyDB instance, which is what exactly you experienced: nothing happens in the GCP resource. This is called managing the field externally: https://cloud.google.com/config-connector/docs/concepts/managing-fields-externally

What I wanted to check in the original comment is that: What's the behavior of GCP AlloyDB instance when you set this field to `false` in AlloyDBInstance?

nancynh: Thanks for the clarification @maqiuyujoyce! I misunderstood your question. Sorry about that! When we do set this field to `false` in the Config Connector AlloyDBInstance, the underlying GCP AlloyDB Instance does end up setting the field to `false` as well. It can be seen in the activity logs for the project when I had tested this earlier ([screenshot](https://screenshot.googleplex.com/5WdGGpCuTnUyqQd)).

maqiuyujoyce: Great, thank you so much for the verification!

maqiuyujoyce:
@nancynh could you fix the failures in Presubmit / tests-e2e-fixtures? Looks like there are 8 AlloyDB test cases failed there.

nancynh:
@maqiuyujoyce Finally got the rest of the tests passing. Just a note, because I re-generated the mockgcp AlloyDB protos to include the `enableOutboundPublicIP` field, it also added another field `MaintenanceSchedule` which is unrelated to my changes. So the HTTP golden logs had to be updated to include that too

maqiuyujoyce (commented):
/lgtm
/approve

nancynh:
Thank you for the quick review!!

PR #2787, opened 2024-09-25
feat: add StringDuration <-> Proto mapper

yuwenma:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/pkg/controller/direct/maputils.go
+++ b/pkg/controller/direct/maputils.go
@@ -157,6 +157,25 @@ func StringTimestamp_ToProto(mapCtx *MapContext, s *string) *timestamppb.Timesta
 	return ts
 }
 
+func StringDuration_FromProto(mapCtx *MapContext, d *durationpb.Duration) *string {
+	if d == nil {
+		return nil
+	}
+	s := d.AsDuration().String()
+	return &s
+}
+
+func StringDuration_ToProto(mapCtx *MapContext, s *string) *durationpb.Duration {
+	if s == nil {
+		return nil
+	}
+	td, err := time.ParseDuration(*s)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/controller/direct/maputils.go
```diff
@@ -157,6 +157,25 @@ func StringTimestamp_ToProto(mapCtx *MapContext, s *string) *timestamppb.Timesta
 	return ts
 }
 
+func StringDuration_FromProto(mapCtx *MapContext, d *durationpb.Duration) *string {
+	if d == nil {
+		return nil
+	}
+	s := d.String()
```
justinsb, line 164: I don't think this is correct, I think (for example) we want something like `seconds:15` instead of `15s`.

yuwenma: I'm a little bit confused. Do you mean we want or don't want? This will give `seconds:15`, you can see the test case below

justinsb: Sorry, yes, I mean we want `15s` (that's the format we normally use in KRM), rather than `seconds:15`

yuwenma: Okay, changed. Please see the new test (I used the time duration so it only goes to Hour but not Month and Year).

justinsb (approved):

justinsb:
Approved, but I have realized that this means that we now will update our existing generated mappers.  We should be careful that this does not introduce regressions!

yuwenma:
> Approved, but I have realized that this means that we now will update our existing generated mappers. We should be careful that this does not introduce regressions!

Good point! This comes from an error I saw when running the generate-mapper tool. I assume that means the previous `Duration` case are all manually handled. Will that be overridden as well?

PR #2789, opened 2024-09-25
refactor: use new externalRef convention

acpana:
Refactor DataformRepository's externalRef as per https://github.com/yuwenma/k8s-config-connector/blob/real-id/docs/develop-resources/api-conventions/external-reference.md


yuwenma (commented):
/lgtm
/approve

PR #2791, opened 2024-09-25
refactor: use resource name in ctrl name

acpana:
As I'm working with more resources in the BigQueryAnalyticsHub service group, I want to start adding nice-ness to the tooling when having more than one resource.


anhdle-sso:
/lgtm

The change is reasonable and useful. I am not sure if it effects anything else down the line other than logging.

PR #2793, opened 2024-09-25
Fix newly-added SQLInstance periodic tests

jasonvigil:
Many of these cases are failing because the terraform-based controller behaves differently (read: worse) than the direct controller. For now, we can skip some of the test cases until we enable the direct controller, or figure out how to replicate the "partially-enable direct controller via annotations / predicates" capability that we have in the actual registration controller for our periodic tests.


yuwenma (commented):
/lgtm
/approve

Thank you! +1 to the partially-enable direct controller approach

PR #2794, opened 2024-09-25
fix: Extend sql periodic test timeout

jasonvigil:
It takes longer than 4 hours :/


jasonvigil:
Actually, for a long term solution it's probably better to figure out how to run these tests in parallel.

PR #2797, opened 2024-09-26
feat: direct controller for SecretManagerSecret (part 1)

yuwenma:
### Change description

Fixes #

### Tests you have done


dev/tasks/run-e2e
```diff
@@ -24,7 +24,7 @@ echo "Downloading envtest assets..."
 export KUBEBUILDER_ASSETS=$(go run sigs.k8s.io/controller-runtime/tools/setup-envtest@latest use -p path)
 
 if [[ -z "${KCC_USE_DIRECT_RECONCILERS:-}" ]]; then
-  KCC_USE_DIRECT_RECONCILERS=SQLInstance,ComputeForwardingRule,GKEHubFeatureMembership,CertificateManagerDNSAuthorization
+  KCC_USE_DIRECT_RECONCILERS=SQLInstance,ComputeForwardingRule,GKEHubFeatureMembership,CertificateManagerDNSAuthorization,SecretManagerSecret
```
justinsb, line 27: Not a blocker, but we probably to think about whether we can use the annotation so that we can continue to test both the TF and the direct reconcilers.

mockgcp/mocksecretmanager/secrets.go
```diff
@@ -62,22 +63,30 @@ func (s *SecretsV1) CreateSecret(ctx context.Context, req *pb.CreateSecretReques
 	obj.Name = fqn
 	obj.CreateTime = timestamppb.Now()
 	if obj.Replication == nil {
-		obj.Replication = &pb.Replication{}
-	}
-	if obj.Replication.Replication == nil {
-		obj.Replication.Replication = &pb.Replication_Automatic_{
-			Automatic: &pb.Replication_Automatic{},
-		}
+		return nil, fmt.Errorf("Secret.replication must be specified.")
```
justinsb, line 66: Really (I.e. this came up as a difference on record-gcp vs compare-mock)?  That's surprising!

pkg/controller/direct/secretmanager/mapper.generated.go
```diff
@@ -0,0 +1,203 @@
+// Copyright 2024 Google LLC
```
justinsb, line 1: Nit: we should add these to `.gitattributes` to indicate they are generated code (and probably add a "this is generated code" comment)!

pkg/controller/direct/secretmanager/secret_controller.go
```diff
@@ -0,0 +1,381 @@
...
+			return false, nil
+		}
+		return false, fmt.Errorf("getting SecretManagerSecret %q: %w", a.id.FullyQualifiedName(), err)
+	}
+
+	a.actual = secretpb
+	return true, nil
+}
+
+func MergeMap(a, b map[string]string) map[string]string {
```
justinsb, line 189: Nit: this might be safer/clearer if it copied the map, rather than mutating `a` in-place

pkg/controller/direct/secretmanager/secret_controller.go
```diff
@@ -0,0 +1,381 @@
...
+	return a
+}
+
+func UserAnnotations(secret *krm.SecretManagerSecret) map[string]string {
+	annotations := MergeMap(secret.GetAnnotations(), secret.Spec.Annotations)
+	annotations = RemoveByPrefixes(annotations, "cnrm.cloud.google.com", "alpha.cnrm.cloud.google.com")
+	return annotations
+}
+
+func UserLabels(secret *krm.SecretManagerSecret) map[string]string {
```
justinsb, line 213: Nit: maybe "ComputeGCPLabels" / "ComputeGCPAnnotations" ?  Also I think we should immediately put these into a shared package, because they aren't something we want each service to write, we need to be compatible with existing behaviours.

pkg/controller/direct/secretmanager/secret_controller.go
```diff
@@ -0,0 +1,381 @@
...
+	log.V(2).Info("creating Secret", "name", a.id.FullyQualifiedName())
+	mapCtx := &direct.MapContext{}
+
+	desired := a.desired.DeepCopy()
+	resource := SecretManagerSecretSpec_ToProto(mapCtx, &desired.Spec)
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+
+	resource.Annotations = UserAnnotations(desired)
```
justinsb, line 232: I assume we support annotations in the TF controller, from `metadata.annotations`, so we aren't free to choose our behaviour here?

pkg/controller/direct/secretmanager/secret_controller.go
```diff
@@ -0,0 +1,381 @@
...
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+	status.ExternalRef = a.id.AsExternalRef()
+	status.Name = created.Name
+	return setStatus(u, status)
+}
+
+func topicsEqual(desired []*krm.TopicRef, actual []*secretmanagerpb.Topic) bool {
+	externalsDesired := []string{}
```
justinsb, line 258: Nit: you should probably use `set[string]` here, it'll probably be clearer

pkg/controller/direct/secretmanager/secret_controller.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+	// the GCP service use *name* to identify the resource.
+	resource.Name = a.id.FullyQualifiedName()
+
+	resource.Annotations = UserAnnotations(desired)
+	resource.Labels = UserLabels(desired)
+
+	updateMask := &fieldmaskpb.FieldMask{}
+
+	if !reflect.DeepEqual(resource.Annotations, a.actual.Annotations) {
```
justinsb, line 291: I wonder if we can infer from this which fields are immutable (or at least we don't support update), to add another "interlock" so that resources are more correct-by-design.

pkg/test/resourcefixture/testdata/basic/run/v1beta1/runjob/jobwithsecretmanagersecret/_generated_object_jobwithsecretmanagersecret.golden.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: run.cnrm.cloud.google.com/v1beta1
```
justinsb, line 1: Nit: it would be nice to capture the output before we turned on the direct controller!  Maybe we should encourage test changes to be in a separate PR?

pkg/test/resourcefixture/testdata/basic/secretmanager/v1beta1/fullsecretmanagersecret/_generated_object_fullsecretmanagersecret.golden.yaml
```diff
@@ -18,25 +17,23 @@ spec:
   annotations:
     bar: secretmanagersecret-bar
     foo: secretmanagersecret
-  expireTime: "2025-10-02T15:01:23Z"
+  expireTime: "2025-10-03T15:01:23Z"
```
justinsb, line 20: Hmmm . we should probably normalize this!

pkg/test/resourcefixture/testdata/basic/secretmanager/v1beta1/fullsecretmanagersecret/_generated_object_fullsecretmanagersecret.golden.yaml
```diff
@@ -18,25 +17,23 @@ spec:
...
   replication:
     automatic: true
-  resourceID: secretmanagersecret-${uniqueId}
   rotation:
-    nextRotationTime: "2025-10-02T15:01:23Z"
-    rotationPeriod: 3600s
+    nextRotationTime: "2025-10-03T15:01:23Z"
```
justinsb, line 24: And this!

pkg/test/resourcefixture/testdata/basic/secretmanager/v1beta1/fullsecretmanagersecret/create.yaml
```diff
@@ -38,6 +38,7 @@ spec:
   rotation:
     nextRotationTime: "2025-10-02T15:01:23Z"
     rotationPeriod: "3600s"
-  versionAliases:
-    "foo": "111"
-    "bar": "222"
+  # FR to SecretManagerSecretVersion adding aliases.
```
justinsb, line 41: Is there a follow-on PR for this (or a bug?)  I'm just wary that because we're removing test coverage, it'll make it harder to track

yuwenma: Yup, this is exactly why this PR is "part 1" 😄. I want to handle the secretVersion more carefully and dig deeper into these two resources, because I think something may go wrong...

pkg/controller/direct/secretmanager/secret_controller.go
yuwenma: Yes! And update needs more helpers in general!

pkg/test/resourcefixture/testdata/basic/secretmanager/v1beta1/fullsecretmanagersecret/_generated_object_fullsecretmanagersecret.golden.yaml
yuwenma: Which format would you suggest to use ?  The current format is ISO 8601.

mockgcp/mocksecretmanager/tests/service_test.go
yuwenma, line 1: This unit test is stale and uses invalid value to test (see [failure](https://github.com/GoogleCloudPlatform/k8s-config-connector/actions/runs/11247055706/job/31269889219?pr=2797#step:5:1972)). This test has been covered by the mockgcp full test suite.

pkg/test/resourcefixture/testdata/basic/secretmanager/v1beta1/secretmanagersecretversion/_http.log
```diff
@@ -63,36 +65,6 @@ X-Xss-Protection: 0
 
 ---
 
-GET https://secretmanager.googleapis.com/v1/projects/${projectId}/secrets/secretmanagersecretversion-dep-${uniqueId}?alt=json
```
yuwenma, line 66: seems that the previous TF-based approach has an additional REST call to get the secret.

apis/secretmanager/v1beta1/secretmanagersecret_types.go
```diff
@@ -39,24 +39,9 @@ type SecretManagerSecretSpec struct {
 	//  The replication policy cannot be changed after the Secret has been created.
 	Replication *Replication `json:"replication,omitempty"`
 
-	/*NOTYET
```
justinsb, line 42: Nit: we should write a doc about labels in SciFi

dev/tasks/run-e2e
```diff
@@ -24,7 +24,7 @@ echo "Downloading envtest assets..."
 export KUBEBUILDER_ASSETS=$(go run sigs.k8s.io/controller-runtime/tools/setup-envtest@latest use -p path)
 
 if [[ -z "${KCC_USE_DIRECT_RECONCILERS:-}" ]]; then
-  KCC_USE_DIRECT_RECONCILERS=ComputeForwardingRule,GKEHubFeatureMembership
+  KCC_USE_DIRECT_RECONCILERS=ComputeForwardingRule,GKEHubFeatureMembership,SecretManagerSecret
```
justinsb, line 27: Nit: we should also think about how we maintain test coverage when the old and new controllers co-exist

pkg/controller/direct/common/labels.go
```diff
@@ -0,0 +1,33 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common
+
+import "strings"
+
+func ComputeGCPLabels(labels map[string]string) map[string]string {
+	RemoveByPrefixes(labels, "cnrm.cloud.google.com")
```
justinsb, line 20: Is this the same as the TF code?  I thought we also skipped invalid labels there...

pkg/controller/direct/common/normalize.go
```diff
@@ -0,0 +1,52 @@
...
+
+package common
+
+import (
+	"time"
+
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+func NormalizeStringToTimestamp(ts string) *timestamppb.Timestamp {
```
justinsb, line 23: This seems _very_ permissive.... did we previously support all this formats?

pkg/controller/direct/secretmanager/secret_controller.go
```diff
@@ -0,0 +1,367 @@
...
+	for k, v := range a {
+		copy[k] = v
+	}
+	for k, v := range b {
+		copy[k] = v
+	}
+	return copy
+}
+
+func ComputeAnnotations(secret *krm.SecretManagerSecret) map[string]string {
```
justinsb, line 200: Do we do this today in the existing controller?

pkg/controller/direct/secretmanager/secret_controller.go
```diff
@@ -0,0 +1,367 @@
...
+
+	updateMask := &fieldmaskpb.FieldMask{}
+
+	if !reflect.DeepEqual(resource.Annotations, a.actual.Annotations) {
+		updateMask.Paths = append(updateMask.Paths, "annotations")
+	}
+	if !topicsEqual(desired.Spec.TopicRefs, a.actual.GetTopics()) {
+		updateMask.Paths = append(updateMask.Paths, "topics")
+	}
+	if !common.DeepEqual_StringAndTimestampPb(*desired.Spec.ExpireTime, a.actual.GetExpireTime()) {
```
justinsb, line 283: Nit: It's probably easier to compare resource vs a.actual

pkg/controller/direct/secretmanager/secret_controller.go
```diff
@@ -0,0 +1,367 @@
...
+	status, err := runtime.DefaultUnstructuredConverter.ToUnstructured(typedStatus)
+	if err != nil {
+		return fmt.Errorf("error converting status to unstructured: %w", err)
+	}
+
+	old, _, _ := unstructured.NestedMap(u.Object, "status")
+	if old != nil {
+		status["conditions"] = old["conditions"]
+		status["observedGeneration"] = old["observedGeneration"]
+		status["externalRef"] = old["externalRef"]
```
justinsb, line 360: Nit: I'm not sure why we are copying externalRef - I don't think we have to...

pkg/controller/direct/secretmanager/secret_controller.go
```diff
@@ -0,0 +1,367 @@
...
+		status["name"] = old["name"]
```
justinsb, line 361: Nit: Or why we're copying name!

justinsb:
Some thoughts, nothing blocking I don't think.  Will hold and you can remove hold.

/approve
/lgtm
/hold

pkg/controller/direct/secretmanager/secret_controller.go
yuwenma: This is a TF beta resource. And we discussed this `name` [in the previous PR](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2661#discussion_r1767644685) (I know it's annoying..)

pkg/controller/direct/secretmanager/secret_controller.go
yuwenma: Ah! At the time I added this template, we don't have this wonderful code! https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/d4cc447191eddae1a6def67146e66c0658f283a1/pkg/controller/direct/directbase/operations.go#L95
You are right, we don't need to now

pkg/controller/direct/secretmanager/secret_controller.go
yuwenma: Yes! exactly what I plan to do next!

pkg/controller/direct/secretmanager/secret_controller.go
yuwenma: No, we don't. Some beta resources have this coverage. I think we need put the update in a shared function to avoid similiar issues happening again.

pkg/controller/direct/common/normalize.go
yuwenma: The GCP timestamp proto does. This is for comparison. I feel like some previous Update compare could be wrong.. (like the same timestamp meaning with a different format is considered different)

dev/tasks/run-e2e
yuwenma: we can switch from KCC_USE_DIRECT_RECONCILERS to the alpha direct annotation.

yuwenma:
/hold cancel

Thank you for the suggestions! 
I'll send out PRs to improve the structure.

pkg/controller/direct/common/normalize.go
yuwenma: I changed this in #2904 to make it less permissive.

pkg/controller/direct/secretmanager/secret_controller.go
yuwenma: Used the protos field_behavior to find out the immutable and add a function to do proto level comparison in #2904

pkg/controller/direct/secretmanager/mapper.generated.go
yuwenma: Thank you for updating the .gitattributes!

pkg/controller/direct/secretmanager/secret_controller.go
yuwenma: Updated the template so future controllers can use opt to update #2919 2919

PR #2808, opened 2024-09-27
Fix sqlinstance user labels to filter out KRM-style labels

jasonvigil:
Fixes issue reported in https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/2706#issuecomment-2357595555


jasonvigil:
CC @maqiuyujoyce --

Note, the [Cleanup duplicate label helper functions](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2808/commits/6591d25244fc7b86f202e788607253e9254ff028) refactor will cause a slight behavior change when exporting resources into terraform HCL format. Now, the exported HCL will include the "managed-by-cnrm" label. I think this is ok, as long as the export is only used for debugging purposes, and users do not terraform apply the exported HCL (and therefore potentially cause confusion when resource labels incorrectly report they are managed by cnrm when, in
fact, they are not managed by cnrm). It could be argued this new behavior is "more correct", because the exported HCL now exactly matches what KCC is applying.

But, want to get your opinion on this refactor. Can you PTAL?

pkg/label/label_test.go
```diff
@@ -48,56 +31,44 @@ func TestNilMap(t *testing.T) {
...
 		"managed-by-cnrm": "true",
 	}
 	if !reflect.DeepEqual(result, expectedResult) {
 		t.Errorf("results mismatch: got '%v', want '%v'", result, expectedResult)
 	}
-	result = label.NewGCPLabelsFromK8SLabels(labels)
-	expectedResult = map[string]interface{}{
-		"key1": "val1",
-		"key2": "val2",
+	result = label.NewGCPLabelsFromK8sLabels(labels)
```
maqiuyujoyce, line 49: Nit: Looks like  all the rest are duplicated test cases. Could you remove them?

maqiuyujoyce (commented):
Thank you for the fix and cleaning up the code, @jasonvigil ! Just one comment about unit test but feel free to unhold and address it in a separate PR if you prefer.

/lgtm
/approve
/hold

pkg/label/label_test.go
jasonvigil: Ah, yeah, good catch! Thanks. Fixed

maqiuyujoyce (approved):
/lgtm
/approve

PR #2809, opened 2024-09-27
refactor: make cmds kind aware

acpana:
This makes the `generate-type`, `add` and `generate-mapper` cmds work for multiple resource in a an API package. For instance, now for `bigqueryanalyticshub` (the API package), we can have multiple resources (`dataexchange, dataexchangelisting).

The folder structure for `apis` is:

```
bigqueryanalyticshub
├─dataexchange
└─dataexchangelisting
   ├─v1alpha1
   │  └─doc.go ; etc other files
   └─v1beta1
```

and for the `controller` pkg:
```
bigqueryanalyticshub
├─dataexchange
└─dataexchangelisting
   ├─dataexchangelisting_controller.go
   ├─dataexchangelisting_externalresource.go
   └─v1beta1.mapper.generated.go
```

Closes on one of the items in https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/2802

PR #2810, opened 2024-09-27
wip:feat: direct BigQueryAnalyticsHubDataExchangeListing

acpana:
Super rough draft to show how I used the changes in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2809


apis/bigqueryanalyticshub/dataexchangelisting/v1alpha1/bigqueryanalyticshubdataexchangelisting_types.go
```diff
@@ -0,0 +1,88 @@
+// Copyright 2024 Google LLC
```
jingyih, line 1: Just to clarify, the alternative I suggested would generate the types under
`apis/bigqueryanalyticshub/v1alpha1/bigqueryanalyticshubdataexchangelisting_types.go`

acpana: I thought the convention was usually `/apis/$GROUP_NAME/$VERSION`. In our case I figured the group would be the `service_name/actual_group` wdyt ? The benefits here would be for us to have each resource in its own versioned package.

What's the benefits of lumping all resources under the version instead ?

jingyih: I see pros and cons for both approaches. Here’s what I’m thinking:
 - Resources from the same API might share common dependencies, so if each resource is in its own package, we’d end up generating the same dependencies over and over.
 - I also think it’s better to stick with the same version of an API across resources, instead of importing different versions from the same API.

acpana: From an offline chat w Justin:

I think we are going to move towards the second option above. To elaborate on the structure for multiple resources, it will be more like:

```
bigqueryanalyticshub
├─v1alpha1
└─v1beta1
   ├─ dataexchangelisting _types.go
   ├─ dataexchange _types.go
   ├─ shared _types.go
   │--doc.go ; etc other files
   └─group_version.go
   ```

I will make the changes on the #2809 PR.

PR #2811, opened 2024-09-27
Add mappers for ComputeFirewallPolicyRule

gemmahou:
### Change description

- Generate KCC types, mappers and crd for ComputeFirewallPolicyRule
- Add Fuzz test

### Tests you have done


apis/compute/v1beta1/computefirewallpolicyrule_types.go
```diff
@@ -0,0 +1,160 @@
...
+type ComputeFirewallPolicyRuleSpec struct {
+	/* The Action to perform when the client connection triggers the rule. Valid actions are "allow", "deny" and "goto_next". */
+	Action string `json:"action"`
+
+	/* An optional description for this resource. */
+	// +optional
+	Description *string `json:"description,omitempty"`
+
+	/* The direction in which this rule applies. Possible values: INGRESS, EGRESS */
+	Direction string `json:"direction"`
```
acpana, line 91: does this default to anything?

apis/compute/v1beta1/computefirewallpolicyrule_types.go
```diff
@@ -0,0 +1,160 @@
...
+package v1beta1
+
+import (
+	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
+	commonv1alpha1 "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/apis/common/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +kcc:proto=google.cloud.compute.v1.FirewallPolicyRuleMatcherLayer4Config
+type FirewallpolicyruleLayer4Configs struct {
```
acpana, line 26: I think this type and the other one should be camel cased, right ?

apis/compute/v1beta1/computefirewallpolicyrule_types.go
```diff
@@ -0,0 +1,160 @@
...
+	/* The IP protocol to which this rule applies. The protocol type is required when creating a firewall rule. This value can either be one of the following well known protocol strings (`tcp`, `udp`, `icmp`, `esp`, `ah`, `ipip`, `sctp`), or the IP protocol number. */
+	IpProtocol string `json:"ipProtocol"`
```
acpana, line 28:
```suggestion
	IPProtocol string `json:"ipProtocol"`
```

?

apis/compute/v1beta1/computefirewallpolicyrule_types.go
```diff
@@ -0,0 +1,160 @@
...
+// +kcc:proto=google.cloud.compute.v1.FirewallPolicyRule
+type ComputeFirewallPolicyRuleStatus struct {
+	commonv1alpha1.CommonStatus `json:",inline"`
+	/* Type of the resource. Always `compute#firewallPolicyRule` for firewall policy rules */
+	// +optional
+	Kind *string `json:"kind,omitempty"`
+
+	/* Calculation of the complexity of a single firewall policy rule. */
+	// +optional
+	RuleTupleCount *int64 `json:"ruleTupleCount,omitempty"`
```
acpana, line 126: Shouldn't these be in ab ObservedState struct ?

apis/compute/v1beta1/computefirewallpolicyrule_types.go
```diff
@@ -0,0 +1,160 @@
...
+// +kubebuilder:printcolumn:name="Status",JSONPath=".status.conditions[?(@.type=='Ready')].reason",type="string",description="The reason for the value in 'Ready'"
+// +kubebuilder:printcolumn:name="Status Age",JSONPath=".status.conditions[?(@.type=='Ready')].lastTransitionTime",type="date",description="The last transition time for the value in 'Status'"
+
+// ComputeFirewallPolicyRule is the Schema for the compute API
+// +k8s:openapi-gen=true
+type ComputeFirewallPolicyRule struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   ComputeFirewallPolicyRuleSpec   `json:"spec,omitempty"`
```
acpana, line 145: (this probably needs to come from the template code but we should `//+require` the spec here

pkg/controller/direct/compute/firewallpolicyrule/mapper.generated.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+	out := &krm.ComputeFirewallPolicyRuleSpec{}
+	out.Action = in.GetAction()
+	out.Description = in.Description
+	out.Direction = in.GetDirection()
+	out.Disabled = in.Disabled
+	out.EnableLogging = in.EnableLogging
+	out.Match = FirewallpolicyruleMatch_FromProto(mapCtx, in.Match)
+	out.Priority = int64(in.GetPriority())
+	// MISSING: RuleName
```
acpana, line 35: For the missing fields, can we add a comment as to why? Something like "not yet or not supported in proto for now, etc.)

acpana (commented):
overall looks great! Small/ few comments on the types

apis/compute/v1beta1/computefirewallpolicyrule_types.go
gemmahou: This is a Beta resource, and we haven't have ObservedState supported for this resource. We'll keep output only fields under status.

apis/compute/v1beta1/computefirewallpolicyrule_types.go
gemmahou: Since this is a Beta resource, we should keep the types the same(?) I copied from the existing one: pkg/clients/generated/apis/compute/v1beta1/computefirewallpolicyrule_types.go

apis/compute/v1beta1/computefirewallpolicyrule_types.go
gemmahou: Same here, we should keep the types the same as the previous generated one?

apis/compute/v1beta1/computefirewallpolicyrule_types.go
gemmahou: This field is required, either "INGRESS" or "EGRESS" is needed. I do not think there's a "default" value.

apis/compute/v1beta1/computefirewallpolicyrule_types.go
gemmahou: I actually copied the existing types because this is Beta resource. While I do not see `//+required` in the [template](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/67b69fdebb742db5390e6b328d00c396765c9da9/dev/tools/controllerbuilder/template/apis/types.go#L90), do we have pending changes to the template recently?

pkg/controller/direct/compute/firewallpolicyrule/mapper.generated.go
gemmahou: Right, those fields we are not supported yet. Instead of adding comment here in the generated code, I feel we may want to mention that `"MISSING" unsupported fields for Beta resources is accepted` in the [guide](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/b5cc2e924037a053af625b1781aed2a71b7caf67/docs/develop-resources/guides/3-add-mapper.md#exit-criteria), WDYT?

acpana:
/lgtm
/approve

thanks for following up

/hold

I'll let you merge this

PR #2814, opened 2024-09-28
feat: add direct controller for BigQueryDataset

xiaoweim:


apis/refs/v1beta1/projectref.go
```diff
@@ -166,5 +166,13 @@ func ResolveProjectID(ctx context.Context, reader client.Reader, obj *unstructur
 		return projectID, nil
 	}
 
+	if selfLink, _, _ := unstructured.NestedString(obj.Object, "status", "selfLink"); selfLink != "" {
```
yuwenma, line 169: Did you hit an actual case that doesnot have the project ID from the spec.projectID or the annotation?

xiaoweim: In the mock test, there is no annotation for project ID and this resource does not have a spec.projectID. I thought about adding the annotation in the yamls in mock tests, but I am not sure if I am supposed to change the create/update yamls in the mock test when adding the direct controller since it was working with TF.

pkg/controller/direct/bigquerydataset/dataset_controller.go
```diff
@@ -0,0 +1,383 @@
...
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	ctrlName        = "bigquery-controller"
+	defaultLocation = "US"
```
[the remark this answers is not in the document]
xiaoweim: Yes.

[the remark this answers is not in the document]
yuwenma: I see this defaultlocation is called in ln 111 when `spec.location` is not specified. This is TF's approach to handle defaulting which has many problems in KCC. What we do here is that making `spec.location` as required (remember `location` is the parent so it is required) in the CRD level, so the line 110-112 will never hit. Rigth? 

Could you also try using a real location like `us-central1` or `us-central1-a` to see what happens?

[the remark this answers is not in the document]
xiaoweim: I see, I can definitely change the location to be required. Yes if the location is required, line 110-112 will not be hit. 

I ran the test against real GCP with the default location changed to `us-central1`, I don't see any issue with it except for the location in the http log being changed from `US` to `us-central1`. 

Let me update the location to be required.

pkg/controller/direct/bigquerydataset/dataset_controller.go
```diff
@@ -0,0 +1,383 @@
...
+	}
+
+	projectRef, err := refs.ResolveProject(ctx, reader, obj, obj.Spec.ProjectRef)
+	if err != nil {
+		return nil, err
+	}
+	projectID := projectRef.ProjectID
+	if projectID == "" {
+		return nil, fmt.Errorf("cannot resolve project")
+	}
```
yuwenma, line 97: Can we use the new controller template and the type-generator?  Mostly we want to get rid of serviceDomain

pkg/controller/direct/bigquerydataset/dataset_controller.go
```diff
@@ -0,0 +1,383 @@
...
+			return nil, err
+		}
+		obj.Spec.DefaultEncryptionConfiguration.KmsKeyRef = kmsRef
+	}
+
+	// Get location
+	// Today TF controller defaults location to US when not specify
+	if obj != nil && obj.Spec.Location == nil {
+		obj.Spec.Location = direct.LazyPtr(defaultLocation)
+	}
```
yuwenma, line 112: nit: we don't need this check because location is required now.

pkg/controller/direct/bigquerydataset/dataset_controller.go
```diff
@@ -0,0 +1,383 @@
...
+	err := deleteDatasetCall.Do()
+	if err != nil {
+		return false, fmt.Errorf("deleting Dataset %s: %w", a.id.FullyQualifiedName(), err)
+	}
+	log.V(2).Info("successfully deleted Dataset", "name", a.id.FullyQualifiedName())
+
+	return true, nil
+}
+
+func setStatus(u *unstructured.Unstructured, typedStatus any) error {
```
yuwenma, line 367: nit: we don't this (if using the new controller template)

yuwenma:
Good job! Overall looks good and I think the main question is about why you need to resolve project from the seflLink. Besides, there are a few improvements and potential bug fixes that you can just apply the latest controller-template to pick up.

xiaoweim:
> Good job! Overall looks good and I think the main question is about why you need to resolve project from the seflLink. Besides, there are a few improvements and potential bug fixes that you can just apply the latest controller-template to pick up.

Regarding our discussion on why the `resolveProjectRef` function did not give error, the reason is that when the adapterForObject function is called, the `spec.projectRef` is already set to the right project. It is not nil so it did not return error.

The reason why the projectRef is already set is because KCC does project defaulting. Here when we create the [harness](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/tests/e2e/unified_test.go#L247), we set the default [project](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/config/tests/samples/create/harness.go#L372) to the test harness. And then we apply the defaults and [project annotation](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/tests/e2e/unified_test.go#L260)


This is where it gets the default project https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/test/gcp/gcp.go#L95. I tried changing the default project in my local config and then it would retunr error.

yuwenma:
> This is where it gets the default project [`master`/pkg/test/gcp/gcp.go#L95](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/test/gcp/gcp.go#L95). I tried changing the default project in my local config and then it would retunr error.

ok, this could relate to the project defaulting https://cloud.google.com/config-connector/docs/how-to/organizing-resources/project-scoped-resources. In such case, the project is not required (but location is)

apis/bigquery/v1beta1/bigquerydataset_types.go
```diff
@@ -94,15 +94,16 @@ type BigQueryDatasetSpec struct {
...
+	Location *string `json:"location"`
 
 	// Optional. Defines the time travel window in hours. The value can be from 48
 	//  to 168 hours (2 to 7 days). The default value is 168 hours if this is not
 	//  set.
 	MaxTimeTravelHours *string `json:"maxTimeTravelHours,omitempty"`
 
 	//	The project that this resource belongs to.
-	//  optional.
+	//  required
```
yuwenma, line 106: Let's keep this optional if that's what BQD supports today.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_bigquerydatasets.bigquery.cnrm.cloud.google.com.yaml
```diff
@@ -97,9 +94,6 @@ spec:
                           items:
                             type: string
                           type: array
-                      required:
-                      - dataset
-                      - targetTypes
```
yuwenma, line 102: let's keep these as required to make it backward compatible.

pkg/controller/direct/bigquerydataset/dataset_controller.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+
+	projectRef, err := refs.ResolveProject(ctx, reader, obj, obj.Spec.ProjectRef)
+	if err != nil {
+		return nil, err
+	}
+	projectID := projectRef.ProjectID
+	if projectID == "" {
+		return nil, fmt.Errorf("cannot resolve project")
+	}
```
yuwenma, line 96: Nit: You don't need this code if using the `krm.NewBigQueryDatasetRef`

pkg/controller/direct/bigquerydataset/dataset_controller.go
```diff
@@ -0,0 +1,351 @@
...
+
+	// Resolve KMS key reference
+	if obj.Spec.DefaultEncryptionConfiguration != nil {
+		kmsRef, err := refs.ResolveKMSCryptoKeyRef(ctx, reader, obj, obj.Spec.DefaultEncryptionConfiguration.KmsKeyRef)
+		if err != nil {
+			return nil, err
+		}
+		obj.Spec.DefaultEncryptionConfiguration.KmsKeyRef = kmsRef
+	}
```
yuwenma, line 105: Let's move this to the create and update calls

pkg/controller/direct/bigquerydataset/dataset_controller.go
```diff
@@ -0,0 +1,351 @@
...
+	desiredKRM := a.desired.DeepCopy()
+	desired := BigQueryDatasetSpec_ToProto(mapCtx, &desiredKRM.Spec, desiredKRM.Name)
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+	actual := a.actual
+	resource := proto.Clone(a.actual).(*bigquerypb.Dataset) // this is the proto resource we are passing to GCP API update call.
+
+	// Check for immutable fields
+	if !reflect.DeepEqual(desired.Location, *actual.Location) {
```
yuwenma, line 212: Could you try the `common.CompareProto` to see if it fits the comparison here?

pkg/controller/direct/bigquerydataset/dataset_controller.go
xiaoweim: May I ask why do we move this into create and update calls? IIUC this is to set the full name to the` kmsref.External`

pkg/controller/direct/bigquerydataset/dataset_controller.go
xiaoweim: I tried using `common.CompareProtoMessage` as how secret_controller is using it here https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/secretmanager/secret_controller.go#L245, but the dataset update call is structured a bit differently. It does not have a updateMask field in the request to cover the changes, it only asks for an entire dataset object without the updateMask.

https://pkg.go.dev/google.golang.org/api/bigquery/v2#DatasetsUpdateCall

pkg/test/resourcefixture/testdata/stateabsentinspec/bigquerydataset/_generated_export_bigquerydataset#01.golden
```diff
@@ -18,7 +18,8 @@ spec:
   - role: WRITER
     specialGroup: projectWriters
   friendlyName: bigquerydataset-sample-updated
-  location: US
+  location: us-central1
+  maxTimeTravelHours: "168"
```
yuwenma, line 22: nice

yuwenma (commented):
few nits. Mostly good to go!

pkg/controller/direct/bigquerydataset/dataset_controller.go
xiaoweim: I see, thank you for the explanation!

yuwenma (commented):
/lgtm
/approve
/hold 

PR looks good. Defer to @jingyih to take another pass and unhold.

jingyih:
I’ll take a look today. For future PRs, I recommend splitting the changes into multiple commits for easier review. For instance, separate the API changes from the controller code.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_bigquerydatasets.bigquery.cnrm.cloud.google.com.yaml
```diff
@@ -77,18 +77,15 @@ spec:
                           description: The dataset this entry applies to.
                           properties:
                             datasetId:
-                              description: Required. A unique ID for this dataset,
-                                without the project name. The ID must contain only
+                              description: Required. A unique Id for this dataset,
```
jingyih, line 80: Maybe not for this PR. This feels like a regression. I believe we want "ID" as opposed to "Id"?
@yuwenma

apis/bigquery/v1beta1/types.generated.go
```diff
@@ -622,151 +274,25 @@ type Dataset struct {
...
-	//  Represents the maximum number of groups to which each protected entity can
-	//  contribute. Changing this value does not improve or worsen privacy. The
-	//  best value for accuracy and utility depends on the query and data.
-	MaxGroupsContributed *int64 `json:"maxGroupsContributed,omitempty"`
-
-	// Optional. The privacy unit column associated with this policy. Differential
-	//  privacy policies can only have one privacy unit column per data source
-	//  object (table, view).
-	PrivacyUnitColumn *string `json:"privacyUnitColumn,omitempty"`
+	DatasetId *string `json:"datasetId,omitempty"`
```
jingyih, line 292: For required fields, I believe we need to add `// +required`. I noticed that in the CRD, some fields that were previously marked as required are no longer required, which is likely not what we want.

jingyih (commented):
I reviewed API-related changes and left a few comments. Thanks!

apis/bigquery/v1beta1/types.generated.go
```diff
@@ -85,354 +85,6 @@ type Access struct {
...
-
-// +kcc:proto=google.cloud.bigquery.v2.DataPolicyOption
-type DataPolicyOption struct {
-	// Data policy resource name in the form of
-	//  projects/project_id/locations/location_id/dataPolicies/data_policy_id.
-	Name *string `json:"name,omitempty"`
-}
-
 // +kcc:proto=google.cloud.bigquery.v2.Dataset
 type Dataset struct {
```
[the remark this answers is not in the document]
xiaoweim: The type generator does not auto remove it. I can manually remove it but it is compatible with https://github.com/googleapis/googleapis/blob/master/google/cloud/bigquery/v2/dataset.proto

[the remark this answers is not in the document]
xiaoweim: Discussed offline, regenerated the file and the `Dataset` struct is removed.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/basicbigquerydataset-direct/_generated_object_basicbigquerydataset-direct.golden.yaml
```diff
@@ -0,0 +1,31 @@
...
+  - lastTransitionTime: "1970-01-01T00:00:00Z"
+    message: The resource is up to date
+    reason: UpToDate
+    status: "True"
+    type: Ready
+  creationTime: "1970-01-01T00:00:00Z"
+  etag: abcdef123456
+  lastModifiedTime: "1970-01-01T00:00:00Z"
+  observedGeneration: 2
+  selfLink: https://bigquery.googleapis.com/bigquery/v2/projects/${projectId}/datasets/bigquerydatasetsample${uniqueId}
```
jingyih, line 31: Hi @maqiuyujoyce, when migrating a beta Terraform-based resource to direct controller, do we need to replicate some of the existing output fields to `observedState`?

pkg/controller/direct/bigquerydataset/dataset_controller.go
```diff
@@ -0,0 +1,342 @@
...
+	mapCtx := &direct.MapContext{}
+
+	// Convert KRM object to proto message
+	desiredKRM := a.desired.DeepCopy()
+	desired := BigQueryDatasetSpec_ToProto(mapCtx, &desiredKRM.Spec, desiredKRM.Name)
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+	actual := a.actual
+	resource := proto.Clone(a.actual).(*bigquerypb.Dataset) // this is the proto resource we are passing to GCP API update call.
```
jingyih, line 192: Do we need a `bigquerypb` resource? Conceptually the controller only need to convert between KRM and API, like the implementation of the Create() function.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/basicbigquerydataset-direct/_generated_object_basicbigquerydataset-direct.golden.yaml
jingyih: Ah I missed that guideline. Thanks!

pkg/controller/direct/bigquerydataset/bigquerydataset_mappings.go
```diff
@@ -21,314 +21,246 @@
...
-	out.MaxTimeTravelHours = &maxTimeTravelHours
-	out.Tags = direct.Slice_FromProto(mapCtx, in.Tags, DatasetTag_FromProto)
+	out.IsCaseInsensitive = direct.LazyPtr(in.IsCaseInsensitive)
+	maxTime := strconv.FormatInt(in.MaxTimeTravelHours, 10)
+	out.MaxTimeTravelHours = direct.LazyPtr(maxTime)
 	out.StorageBillingModel = direct.LazyPtr(in.StorageBillingModel)
 	return out
 }
-func Dataset_ToProto(mapCtx *direct.MapContext, in *krm.Dataset) *pb.DatasetMetadata {
+func BigQueryDatasetStatusObservedState_FromProto(mapCtx *direct.MapContext, in *api.Dataset) *krm.BigQueryDatasetStatus {
```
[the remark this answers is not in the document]
xiaoweim: Make sense.

pkg/controller/direct/bigquerydataset/dataset_controller.go
xiaoweim: Hmm technically yes, I can remove the `bigquerypb` resource conversion.

pkg/controller/direct/bigquerydataset/dataset_controller.go
```diff
@@ -0,0 +1,342 @@
...
+	"google.golang.org/protobuf/types/known/fieldmaskpb"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	ctrlName      = "bigquery-controller"
```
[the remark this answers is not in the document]
xiaoweim: Let me remove it in a followup PR.

pkg/test/resourcefixture/testdata/stateabsentinspec/bigquerydataset/_generated_object_bigquerydataset#01.golden.yaml
```diff
@@ -28,6 +14,7 @@ metadata:
   namespace: ${uniqueId}
 spec:
   friendlyName: bigquerydataset-sample-updated
+  location: us-central1
```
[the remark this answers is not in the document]
xiaoweim: I added location here because the BigQueryDataset API is now marking location as a required field.

pkg/controller/direct/bigquerydataset/bigquerydataset_mappings.go
```diff
@@ -21,314 +21,246 @@
...
 
 import (
-	"fmt"
-	"time"
+	"strconv"
 
-	pb "cloud.google.com/go/bigquery"
 	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/bigquery/v1beta1"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
+	bigquerypb "github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/generated/mockgcp/cloud/bigquery/v2"
```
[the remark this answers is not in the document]
xiaoweim: I think you are right, I modify the controller to only convert between api and krm, so I do not need `bigquerypb` in the mapper file.

apis/bigquery/v1beta1/dataset_reference.go
```diff
@@ -0,0 +1,187 @@
...
+	resourceID := valueOf(obj.Spec.ResourceID)
+	if resourceID == "" {
+		resourceID = obj.GetName()
+	}
+	if resourceID == "" {
+		return nil, fmt.Errorf("cannot resolve resource ID")
+	}
+
+	// Use approved External
+	externalRef := ""
```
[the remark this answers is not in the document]
yuwenma: good catch. This bypasses the immutable validation, @xiaoweim could you elaborate more on this?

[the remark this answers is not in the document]
xiaoweim: There is no externalRef field in the "status" of this resource. I see that by having the externalRef as an empty string, it can bypass the immutable field `project`, `location` and `resourceID` validation. 
Is it necessary to add this field to status or it is okay if we can do this validation through other field such as `selfLink`?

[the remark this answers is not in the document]
yuwenma: We prefer using `status.externalRef` because the base direct reconciler builds additional logic around it.  For `selfLink`, we want to keep it (I know it's a dup), could you add a comment saying that we recommend users to use the `status.externalRef` instead?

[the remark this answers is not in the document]
xiaoweim: I see, sure I can add the recommendation. Let me add `externalRef` here in the status and update the PR. Thank you for the explanation!

pkg/controller/direct/bigquerydataset/dataset_controller.go
```diff
@@ -0,0 +1,334 @@
...
+	log.V(2).Info("updating Dataset", "name", a.id.External)
+	mapCtx := &direct.MapContext{}
+
+	// Convert KRM object to proto message
+	desiredKRM := a.desired.DeepCopy()
+	desired := BigQueryDatasetSpec_ToProto(mapCtx, &desiredKRM.Spec, desiredKRM.Name)
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+	resource := a.actual
```
jingyih, line 188: nit: consider using a clone of `a.actual`. Although it doesn't make a difference with the current logic in the base direct controller (which always retrieves a fresh live state and records it in `actual`). But it will make the code less error prone to future changes.

pkg/controller/direct/bigquerydataset/bigquerydataset_mappings.go
```diff
@@ -21,314 +21,219 @@
...
 
-	pb "cloud.google.com/go/bigquery"
 	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/bigquery/v1beta1"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
+	api "google.golang.org/api/bigquery/v2"
 )
 
-func Access_FromProto(mapCtx *direct.MapContext, in *pb.AccessEntry) *krm.Access {
+func BigQueryDatasetSpec_ToProto(mapCtx *direct.MapContext, in *krm.BigQueryDatasetSpec, name string) *api.Dataset {
```
jingyih, line 32: Can we cleanup the mapping function names? Currently we have `xxx_ToProto` and `xxx_ToAPI` which both referencing "google.golang.org/api/bigquery/v2". Let’s choose a consistent naming convention.

pkg/controller/direct/bigquerydataset/bigquerydataset_mappings.go
```diff
@@ -21,314 +21,219 @@
...
-	out.ProjectId = &in.ProjectID
-	out.TableId = &in.TableID
+	out := &api.TableReference{}
+	out.DatasetId = direct.ValueOf(in.DatasetId)
+	out.ProjectId = direct.ValueOf(in.ProjectId)
+	out.TableId = direct.ValueOf(in.TableId)
 	return out
 }
-func TableReference_ToProto(mapCtx *direct.MapContext, in *krm.TableReference) *pb.Table {
+func BigQueryDataset_CovertKRMToAPI(mapCtx *direct.MapContext, in *krm.BigQueryDataset) *api.Dataset {
```
jingyih, line 229: This function is probably not needed. We only need to send spec fields (i.e. fields that can be configured by users) to GCP when creating the resource.

pkg/controller/direct/bigquerydataset/dataset_controller.go
```diff
@@ -0,0 +1,334 @@
...
+
+func (a *Adapter) Find(ctx context.Context) (bool, error) {
+	log := klog.FromContext(ctx).WithName(ctrlName)
+	log.V(2).Info("getting BigQueryDataset", "name", a.id.External)
+
+	parent, datasetId, err := krm.ParseBigQueryDatasetExternal(a.id.External)
+	if err != nil {
+		return false, fmt.Errorf("failed to parse bigquery dataset full name, %w", err)
+	}
+	req := &bigquerypb.GetDatasetRequest{
```
jingyih, line 119: nit: it does not look like we need to import "bigquerypb" to construct a request. We are talking to "a.gcpService.Datasets.Get" which takes a project ID and dataset ID. Same for Update() and Delete().

pkg/controller/direct/bigquerydataset/dataset_controller.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+	if !reflect.DeepEqual(desired.IsCaseInsensitive, resource.IsCaseInsensitive) {
+		resource.IsCaseInsensitive = desired.IsCaseInsensitive
+		updateMask.Paths = append(updateMask.Paths, "is_case_sensitive")
+	}
+	if !reflect.DeepEqual(desired.MaxTimeTravelHours, resource.MaxTimeTravelHours) {
+		resource.MaxTimeTravelHours = desired.MaxTimeTravelHours
+		updateMask.Paths = append(updateMask.Paths, "max_time_interval_hours")
+	}
+	if desired.Access != nil && resource.Access != nil && len(desired.Access) > 0 && !reflect.DeepEqual(desired.Access, resource.Access) {
```
jingyih, line 234: This means users cannot unset (clear) `Access` list? Not sure if unsetting the entire `Access` list is a valid use case.
Another question is regarding the ordering of this list. Does the underlying API consistently return the list in the same order? If not, we might encounter unexpected diffs even when the user hasn't made any changes to this field.

pkg/controller/direct/bigquerydataset/dataset_controller.go
```diff
@@ -0,0 +1,323 @@
...
+	if len(updateMask.Paths) == 0 {
+		return nil
+	}
+	parent, datasetId, err := krm.ParseBigQueryDatasetExternal(a.id.External)
+	if err != nil {
+		return fmt.Errorf("failed to parse bigquery dataset full name, %w", err)
+	}
+
+	if desired.Access == nil || len(desired.Access) == 0 {
+		datasetGetCall := a.gcpService.Datasets.Get(parent.ProjectID, datasetId)
```
jingyih, line 254: Why do we need to call `Dataset.Get` again to fetch the Access list? Isn't it already in `a.actual / resource`?

jingyih (commented):
We had a offline discussion with @yuwenma. The controller is currently using the deprecated API for BigQuery datasets. We will need to update the controller and mapping functions to use the new API in follow-up PRs.

pkg/controller/direct/bigquerydataset/dataset_controller.go
```diff
@@ -26,7 +26,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
 
-	bigquerypb "github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/generated/mockgcp/cloud/bigquery/v2"
+	clone "github.com/huandu/go-clone"
```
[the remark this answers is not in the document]
xiaoweim: the `api.Dataset` is not a proto, and the `proto.Clone` does not work with the API.

pkg/controller/direct/bigquerydataset/dataset_controller.go
xiaoweim: So the behavior is that even if the user leave the access field empty, it will still return with the default access.

Regarding the ordering of the list, there is a sort access method in our mock to give the access list an order, but it is not entirely the same as the real GCP access order. I believe the real GCP returns the object with an access list of a different order consistently. But our test log is consistent with the mock.

mockgcp/mockbigquery/datasets.go
```diff
@@ -53,6 +53,9 @@ func (s *datasetsServer) GetDataset(ctx context.Context, req *pb.GetDatasetReque
 		}
 		return nil, err
 	}
+	if obj.MaxTimeTravelHours == nil {
```
jingyih, line 56: nit: remove this

jingyih (commented):
/lgtm

jingyih:
Defer to @yuwenma for another pass and final approval.

yuwenma (commented):
/approve

PR #2816, opened 2024-09-29
Fix broken Golang install, and make installed version managed by upda…

benjamin-maynard:
### Change description

The `scripts/environment-setup/golang-setup.sh` setup script as part of the first time contributor setup is broken. This is caused by Go version 1.21+ no longer trimming the patch version for the `0` patch release (https://go.dev/doc/go1.21#introduction).

Updated the install script to use an explicit `GO_VERSION` variable, and added logic to `dev/codebots/update-golang-version` to update this based on the go toolchain version.

### Tests you have done

Tested the installation, and the `update-golang-version` script.


justinsb:
Nice fix - thank you @benjamin-maynard 

/approve
/lgtm

PR #2818, opened 2024-09-30
WIP: Example output from dev/tasks/propose-tag

justinsb:
Some things to note:

* We do create a new version of the manifests that the operator installs.
* We do update the `version/VERSION` file, so KCC reports the correct version.  I think we should use this to create the git tag.
* We do update the `stable` channel for the operator, but I don't think we should do this until we have published the images.
* We don't update the `install-bundle/` directory.  I propose we remove this and point people to the operator directories instead (or potentially a GitHub release)
* We don't update the `crds/` directory.  I propose we remove this and point people to the operator directories instead (or potentially a GitHub release)


ldanielmadariaga:

PR #2819, opened 2024-09-30
feat: Add types for WorkstationConfig

jasonvigil:


jasonvigil:
/hold

Need to rebase and re-run `make manifests` after https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2820 lands.

apis/workstations/v1alpha1/workstationcluster_types.go
```diff
@@ -67,22 +67,6 @@ type WorkstationClusterSpec struct {
 	PrivateClusterConfig *WorkstationCluster_PrivateClusterConfig `json:"privateClusterConfig,omitempty"`
 }
 
-type WorkstationClusterAnnotation struct {
```
justinsb, line 70: Not a blocker, but it looks like you're (only) trying to reuse the Annotation type - I don't think that is worth it generally.  Of course if you're also doing something else, not a problem.

apis/workstations/v1alpha1/workstationconfig_types.go
```diff
@@ -0,0 +1,467 @@
...
+// WorkstationConfigSpec defines the desired state of WorkstationConfig
+// +kcc:proto=google.cloud.workstations.v1.WorkstationConfig
+type WorkstationConfigObservedState struct {
+	// Output only. A system-assigned unique identifier for this workstation
+	//  configuration.
+	UID *string `json:"uid,omitempty"`
+
+	// Output only. Indicates whether this workstation configuration is currently
+	//  being updated to match its intended state.
+	Reconciling *bool `json:"reconciling,omitempty"`
```
justinsb, line 406: Nit: we probably have this in status.conditions

justinsb (approved):

apis/workstations/v1alpha1/workstationcluster_types.go
jasonvigil: The "Annotation" type was not generated by our tooling. I had to create it manually. The original (generated) type for the field was `map[string]string`. Why not re-use the type?

apis/workstations/v1alpha1/workstationconfig_types.go
```diff
@@ -0,0 +1,467 @@
...
+
+	// Optional. Whether the instance has the vTPM enabled.
+	EnableVTPM *bool `json:"enableVTPM,omitempty"`
+
+	// Optional. Whether the instance has integrity monitoring enabled.
+	EnableIntegrityMonitoring *bool `json:"enableIntegrityMonitoring,omitempty"`
+}
+
+// +kcc:proto=google.cloud.workstations.v1.WorkstationConfig.Host.GceInstance.GceConfidentialInstanceConfig
+type WorkstationConfig_Host_GceInstance_GceConfidentialInstanceConfig struct {
```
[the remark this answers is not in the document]
jasonvigil: Hmm, what do you mean? I am using the auto-generated type here. I hit a couple small issues (similar to https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2819#discussion_r1783291960), and had to slightly adjust the auto-generated types, but nothing major.

apis/workstations/v1alpha1/workstationconfig_types.go
jasonvigil: Fixed!

jasonvigil: Accidentally just removed this from WorkstationCluster also, whoops. Will include that in this change as well.

apis/workstations/v1alpha1/workstationcluster_types.go
justinsb: There's no harm in reusing the type, it might just be premature optimization (that can then require custom mappings).  In this case, sgtm!

apis/workstations/v1alpha1/workstationconfig_types.go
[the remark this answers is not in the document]
justinsb: Ah, it looks like you moved the type from `types.generated.go` though?  If we keep it in types.generated.go we can pick up new fields automatically (which is why I'm also interested in the extent to which we can have the generator do more automatically)

jasonvigil:
/hold cancel

I think this is now ready for (re)review

apis/workstations/v1beta1/cluster_types.go
```diff
@@ -45,13 +45,13 @@ type WorkstationClusterSpec struct {
 	DisplayName *string `json:"displayName,omitempty"`
 
 	// Optional. Client-specified annotations.
-	Annotations []WorkstationClusterAnnotation `json:"annotations,omitempty"`
+	Annotations []WorkstationAnnotation `json:"annotations,omitempty"`
```
yuwenma, line 48: Suggest using pointer to get rid of some potential bugs.

apis/workstations/v1alpha1/config_types.go
```diff
@@ -0,0 +1,465 @@
...
+
+import (
+	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/apis/k8s/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var WorkstationConfigGVK = GroupVersion.WithKind("WorkstationConfig")
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
```
yuwenma, line 25: nit: cleanup

yuwenma:
/lgtm
/approve

Some nites. No blockers.

PR #2822, opened 2024-09-30
refactor:ci: remove dcl upgrade probe

acpana:
Removes a github action that tells us if we can upgrade the dcl library (doesn't actually auto-upgrade the dcl lib(. This is safe to do as we've not gotten too much mileage out of it AND we're going towards adding `direct` resources only/ first.


maqiuyujoyce (commented):
/lgtm
/approve

PR #2824, opened 2024-10-01
refactor:ci: selectively run pause tests

acpana:
When I first added the e2e pause tests, the idea was for us to make sure that when KCC is paused, we don't send requests to GCP. We easily do this with mocks (check the log). As we've added more fixtures tests, the run time has ballooned and I think it's overly ambitious to run these for every fixture. 

A given KCC resource will use one of the 4 controllers:

- the Direct controller
- the Terraform (TF) controller
- the DCL controller
- a hand-rolled/ special controller (usually IAM related)

So if any resource can be reduced to one of the 4 categories, it is sufficient (and necessary) to only test an example of each flavor of controller. 

This change will significantly decrease the runtime for the ci job for pause while still fully testing the feature functionality.

### DIFF
--- a/scripts/github-actions/ga-pause-test.sh
+++ b/scripts/github-actions/ga-pause-test.sh
@@ -23,8 +23,21 @@ source ${REPO_ROOT}/scripts/fetch_ext_bins.sh && \
 	fetch_tools && \
 	setup_envs
 
+if [[ -z "${RUN_TESTS:-}" ]]; then
+  RUN_TESTS=""
+  RUN_TESTS+="TestPauseInSeries/fixtures/iamserviceaccount" # IAM
+  RUN_TESTS+="|TestPauseInSeries/fixtures/logbucketmetric" # Direct
+  RUN_TESTS+="|TestPauseInSeries/fixtures/cloudidsendpoint" # DCL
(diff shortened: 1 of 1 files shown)


scripts/github-actions/ga-pause-test.sh
```diff
@@ -23,8 +23,21 @@ source ${REPO_ROOT}/scripts/fetch_ext_bins.sh && \
 	fetch_tools && \
 	setup_envs
 
+if [[ -z "${RUN_TESTS:-}" ]]; then
+  RUN_TESTS=""
+  RUN_TESTS+="TestPauseInSeries/fixtures/iamserviceaccount" # IAM
+  RUN_TESTS+="|TestPauseInSeries/fixtures/logbucketmetric" # Direct
+  RUN_TESTS+="|TestPauseInSeries/fixtures/cloudidsendpoint" # DCL
+  RUN_TESTS+="|TestPauseInSeries/fixtures/computemanagedsslcertificate" # TF
```
acpana, line 31: The challenge is for me/ us to know when a resource is not DCL/ TF anymore. If this is migrated to direct, there's no automated way to know that at the moment. I can gnaw on this issue outside of this PR.

PR #2831, opened 2024-10-01
Update test scripts and instructions to use them.

hankfreund:
### Change description

- Updated the `run-e2e` script to allow running tests without recording goldens
- Updated `record-gcp` to call `run-e2e` instead of reproducing most of its functionality
- Renamed `compare-mock` to `record-mock` to better reflect what it does
- Added `compare-mock` and `compare-gcp` to provide convenient ways of running each in non-record mode
- Updated mockgcp docs to reference the scripts instead of using the full command line
- Updated direct controller step 1 doc with the new scripts

### DIFF
--- a/dev/tasks/run-e2e
+++ b/dev/tasks/run-e2e
@@ -77,9 +77,15 @@ if [[ -z "${TEST_TIMEOUT:-}" ]]; then
 fi
 echo "Using TEST_TIMEOUT=${TEST_TIMEOUT}"
 
+# Check if WRITE_GOLDEN_OUTPUT is unset (does NOT match empty string, so an empty WRITE_GOLDEN_OUTPUT can be used to disable writing output)
+if [[ -z "${WRITE_GOLDEN_OUTPUT+set}" ]]; then
+  WRITE_GOLDEN_OUTPUT=1
+fi
+echo "WRITE_GOLDEN_OUTPUT=${WRITE_GOLDEN_OUTPUT}"
+
 # Run e2e tests against real gcp, capturing output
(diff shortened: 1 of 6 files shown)


hack/record-gcp
```diff
@@ -1,44 +1,17 @@
...
-fi
-echo "TEST_BILLING_ACCOUNT_ID=${TEST_BILLING_ACCOUNT_ID}"
-export TEST_BILLING_ACCOUNT_ID
-
-if [[ -z "${KCC_USE_DIRECT_RECONCILERS:-}" ]]; then
-  KCC_USE_DIRECT_RECONCILERS="ComputeForwardingRule"
-fi
-export KCC_USE_DIRECT_RECONCILERS
-echo "KCC_USE_DIRECT_RECONCILERS=$KCC_USE_DIRECT_RECONCILERS"
+# Clear existing artifacts
```
acpana, line 13: I think we are losing the `KCC_USE_DIRECT_RECONCILERS` here

hankfreund: run-e2e sets this variable (line 27), and has more items in that list (so I think the two got out of sync, which is probably not right).

justinsb:
I like the direction, but I'm wondering whether we should just encourage writing all golden output, all the time.  I expect most people will be using git anyway... what was your use-case for _not_ writing the golden output?

hankfreund:
I think it's specifically useful when first developing a mock service. You run a script to generate the golden files, but then you need to compare your mock against it. Since the goldens are brand new, allowing them to be overwritten could hide a bug where the mock is changing them.

ldanielmadariaga:

PR #2833, opened 2024-10-01
fix: Fix ResourceID field optional tag format

jasonvigil:


maqiuyujoyce (commented):
Thank you for fixing the syntax, @jasonvigil ! It's a bit concerning that the issue is hidden by the fact that `omitempty` is set. In the long-term we might want a different way to verify the syntax. E.g. any field should either have `+required` or `+optional`, or completely dropping `+optional`.

/lgtm
/approve

jingyih:
`// +optional` is optional... so maybe we just add `// +required` when necessary.

jasonvigil:
> `// +optional` is optional... so maybe we just add `// +required` when necessary.

@jingyih makes sense -- I just created https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2852 to fix.

PR #2786, opened 2024-09-25
feat: direct BigQueryAnalyticsHubDataExchange

acpana:
Converts *v1alpha* `BigQueryAnalyticsHubDataExchange` as a direct resource. Note we are choosing to treat this as a special case and first do the conversion and then add the mockgcp as the TF controller is using a proto version that is not published for this service.

PR #2835, opened 2024-10-01
refactor: name field is output only

acpana:
A small follow up on https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2786 and fleshing out the test suite for this resource more.


apis/bigqueryanalyticshub/v1alpha1/bigqueryanalyticshubdataexchange_types.go
```diff
@@ -100,6 +96,10 @@ type BigQueryAnalyticsHubDataExchangeStatus struct {
 // BigQueryAnalyticsHubDataExchangeSpec defines the desired state of BigQueryAnalyticsHubDataExchange
 // +kcc:proto=google.cloud.bigquery.analyticshub.v1.DataExchange
 type BigQueryAnalyticsHubDataExchangeObservedState struct {
+	// Output only. The resource name of the data exchange.
+	//  e.g. `projects/myproject/locations/US/dataExchanges/123`.
+	Name *string `json:"name,omitempty"`
```
justinsb, line 101: You _could_ also just drop it, as long as we are populating status.externalRef

acpana: 😱 actually I like that idea; today this is outputted in the format of an externalRef so it's actually redundant in this case.

justinsb (approved):

PR #2837, opened 2024-10-01
Use reconcile gate to determine reconciler in periodic tests

jasonvigil:
This PR updates our periodic tests so they use the same reconcile
gate logic as controller-manager. With this change, we can test
resources with both the legacy tf/dcl controllers and the new direct
controllers at the same time, to ensure feature parity during resource
migration.


pkg/test/resourcefixture/contexts/register.go
```diff
@@ -152,8 +147,13 @@ func (rc ResourceContext) getResourceConfig(smLoader *servicemappingloader.Servi
 	panic(fmt.Errorf("no resource config found for kind: %v", rc.ResourceKind))
 }
 
-func (rc ResourceContext) IsAutoGenerated(smLoader *servicemappingloader.ServiceMappingLoader) bool {
-	if rc.IsTFResource {
+func (rc ResourceContext) IsAutoGenerated(smLoader *servicemappingloader.ServiceMappingLoader, u *unstructured.Unstructured) bool {
+	rt, err := testreconciler.ReconcilerTypeForObject(u)
```
maqiuyujoyce, line 151: Similarly, I wondered if we should pass rt in as a parameter.

Or can we make some further refactoring and make reconciler type part of the ResourceContext? As we only call it against the initialUnstruct or the updateUnstruct (i.e. the same kind within the same test case), so I feel it is possible to only get the CRD once per test case. Considering the potential cost of make API calls, I think we should minimize the number of times we call ReconcilerTypeForObject().

pkg/controller/iam/auditconfig/iamauditconfig_controller_integration_test.go
```diff
@@ -606,6 +608,15 @@ func newIAMAuditConfigFixture(t *testing.T, refResource *unstructured.Unstructur
 	}
 }
 
+func reconcileIAMAuditConfig(ctx context.Context, t *testing.T, reconciler *testreconciler.TestReconciler, policy *iamv1beta1.IAMAuditConfig, expectedResult reconcile.Result, expectedErrorRegex *regexp.Regexp) {
```
maqiuyujoyce, line 611: The change looks good to me, though have you verified it locally?

pkg/controller/iam/partialpolicy/iampartialpolicy_controller_integration_test.go
```diff
@@ -688,6 +686,15 @@ func assertObservedGenerationEquals(t *testing.T, gcpPolicy *iamv1beta1.IAMParti
 	}
 }
 
+func reconcileIAMPartialPolicy(ctx context.Context, t *testing.T, reconciler *testreconciler.TestReconciler, policy *iamv1beta1.IAMPartialPolicy, expectedResult reconcile.Result, expectedErrorRegex *regexp.Regexp) {
```
maqiuyujoyce, line 689: Similarly, just want to double check if you've tested locally.

pkg/controller/iam/policy/iampolicy_controller_integration_test.go
```diff
@@ -496,6 +498,15 @@ func assertObservedGenerationEquals(t *testing.T, gcpPolicy *iamv1beta1.IAMPolic
 	}
 }
 
+func reconcileIAMPolicy(ctx context.Context, t *testing.T, reconciler *testreconciler.TestReconciler, policy *iamv1beta1.IAMPolicy, expectedResult reconcile.Result, expectedErrorRegex *regexp.Regexp) {
```
maqiuyujoyce, line 501: ditto

pkg/controller/iam/policymember/iampolicymember_controller_integration_test.go
```diff
@@ -317,6 +319,15 @@ func newIAMPolicyMemberFixture(t *testing.T, refResource *unstructured.Unstructu
 	}
 }
 
+func reconcileIAMPolicyMember(ctx context.Context, t *testing.T, reconciler *testreconciler.TestReconciler, policy *v1beta1.IAMPolicyMember, expectedResult reconcile.Result, expectedErrorRegex *regexp.Regexp) {
```
maqiuyujoyce, line 322: ditto

pkg/test/controller/reconciler/testreconciler.go
```diff
@@ -211,67 +264,72 @@ func (r *TestReconciler) NewReconcilerForKind(kind string) reconcile.Reconciler
...
-	default:
-		crd := testcontroller.GetCRDForKind(r.t, kind)
-		reconciler, err = r.newReconcilerForCRD(crd, defaulters, jg)
+	gvk := u.GroupVersionKind()
+	crd, err := crdloader.GetCRDForGVK(gvk)
+	if err != nil {
+		r.t.Fatal(err)
 	}
+
+	rt, err := ReconcilerTypeForObject(u)
```
maqiuyujoyce, line 273: crdloader.GetCRDForGVK() is called not only in this function, but also when ReconcilerTypeForObject() is called. Can the duplicate API call be avoided?

pkg/test/resourcefixture/contexts/register.go
```diff
@@ -95,41 +92,39 @@ func GetResourceContext(fixture resourcefixture.ResourceFixture, serviceMetadata
 		rc.ResourceGVK = fixture.GVK
 	}
 
-	// Determine the type of controller for the resource: TF, DCL, or Direct.
```
maqiuyujoyce, line 98: I feel ReconcilerTypeForObject() should probably be called here, and ReconcilerType should be a field under ResourceContext - it's intended to be a holder for any context related to the target test resource (the one specified in create.yaml/update.yaml).

If we want to extend the functionality to the dependent resources, it is still beneficial that we reduce the number of duplicate calls for the target resource.

pkg/test/runner/runner.go
```diff
@@ -79,7 +79,19 @@ func RunAllWithObjectCreated(ctx context.Context, t *testing.T, mgr manager.Mana
 
 func RunAllWithDependenciesCreatedButNotObject(ctx context.Context, t *testing.T, mgr manager.Manager, shouldRunFunc ShouldRunFunc, testCaseFunc TestCaseFunc) {
 	testFunc := func(ctx context.Context, t *testing.T, testContext TestContext, sysContext SystemContext) {
-		dependencyCleanup := sysContext.Reconciler.CreateAndReconcile(ctx, testContext.DependencyUnstructs, testreconciler.CleanupPolicyAlways, testContext.CreateUnstruct)
+		cleanupFuncs := make([]func(), 0, len(testContext.DependencyUnstructs))
```
maqiuyujoyce, line 82: I assume you'll also use ReconcileGate here?

maqiuyujoyce (commented):
Thank you for the refactoring, @jasonvigil ! The overall change is reasonable, though I'm a bit concerned about performance, so proposed some refactoring ideas.

pkg/controller/dynamic/dynamic_controller_integration_test.go
```diff
@@ -370,9 +374,13 @@ func testNoChange(ctx context.Context, t *testing.T, testContext testrunner.Test
...
 
 func testUpdate(ctx context.Context, t *testing.T, testContext testrunner.TestContext, systemContext testrunner.SystemContext, resourceContext contexts.ResourceContext) {
+	if testContext.UpdateUnstruct == nil {
+		t.Logf("UpdateUnstruct not set; skipping update")
+		return
+	}
 	// Tests with `SkipUpdate` explicitly set to 'true' or tests for
 	// auto-generated resources don't support update test.
-	if resourceContext.SkipUpdate || resourceContext.IsAutoGenerated(systemContext.SMLoader) &&
+	if resourceContext.SkipUpdate || resourceContext.IsAutoGenerated(systemContext.SMLoader, testContext.UpdateUnstruct) &&
```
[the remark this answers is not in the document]
jasonvigil: Hmm, I agree, this whole skip logic / if statement does feel a bit messy.

pkg/test/controller/reconciler/testreconciler.go
jasonvigil: Where is the API call? I felt this was ok, since it seemed like `crdloader.GetCRDForGVK(gvk)` was only reading from disk.

pkg/test/resourcefixture/contexts/register.go
jasonvigil: I had considered this also, but felt like it would be easier / more ergonomic for the caller of `Reconcile` (and other functions) to only have to specify the object they want reconciled. The details of figuring out which reconciler to use would be abstracted away. We could optimize a little bit by creating a cache, but I don't know that it would really have much of a performance impact, given most of the time these tests are spent waiting on GCP. It sort of feels like "premature optimization" at the expense of complicating our test code.

pkg/test/resourcefixture/contexts/register.go
```diff
@@ -95,41 +92,39 @@ func GetResourceContext(fixture resourcefixture.ResourceFixture, serviceMetadata
...
 		rc.DCLSchema = s
 	}
+
 	return rc, nil
 }
 
-func (rc ResourceContext) SupportsLabels(smLoader *servicemappingloader.ServiceMappingLoader) bool {
-	if rc.IsTFResource {
+func (rc ResourceContext) SupportsLabels(smLoader *servicemappingloader.ServiceMappingLoader, u *unstructured.Unstructured) bool {
+	rt, err := testreconciler.ReconcilerTypeForObject(u)
```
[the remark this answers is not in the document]
jasonvigil: I am not quite sure there is actually an API call caused by this? We are reading the CRDs from disk, so I felt it was ok to call repeatedly.

pkg/test/resourcefixture/contexts/register.go
jasonvigil: We could potentially cache the reconciler type in ResourceContext, but that will also reduce flexibility in the future. For example, we may want to add test cases for changing from TF/DCL in create.yaml -> Direct in update.yaml.

pkg/test/runner/runner.go
jasonvigil: I think so? `sysContext.Reconciler.Reconcile` will determine the correct reconciler to use.

pkg/controller/iam/auditconfig/iamauditconfig_controller_integration_test.go
jasonvigil: I have not! What is the correct way to run these tests locally? Ideally it would be possible to just run a single test case locally (instead of the whole suite).

jasonvigil: Ok I figured out how to run the IAM tests locally using vscode launch configurations like the following:

```
{
	"name": "dynamic - basic-spannerinstance",
	"type": "go",
	"request": "launch",
	"mode": "test",
	"program": "/usr/local/google/home/jasonvigil/code/k8s-config-connector/pkg/controller/dynamic/",
	"buildFlags": [
		"-tags=integration",
	],
	"args": [
		"-test.v",
		"-test.timeout", "1600s",
		"-test.count=1",
		"-test.run", "TestCreateNoChangeUpdateDelete/spanner/^basic-spannerinstance$"
	],
},
```

And, ended up catching a bug!

The testreconciler could not recognize the IAM resources because they had empty GVKs. The GVKs were erased by previously calling `kubeClient.Create` with the typed resources, which uses the typed client to decode the result back into the object and erases the GVK [[ref](https://github.com/kubernetes/apimachinery/blob/c463db196543ee48d2353c6ca3325de6454ad075/pkg/runtime/helper.go#L254-L263)].

The dynamic controller tests are immune to this issue because they use unstructured resources instead of typed resources. Therefore, they use the unstructured kube client, which decodes the result directly back into the provided unstructured object, and does not erase the GVK [[ref](https://github.com/kubernetes/apimachinery/blob/c463db196543ee48d2353c6ca3325de6454ad075/pkg/runtime/serializer/json/json.go#L171-L180)].

I was able to verify this fix for policy, partialpolicy, and policymember. However, I am struggling to run the auditconfig tests locally. Is there some special configuration necessary? Here is the error I am getting (on both master and this feature branch):

```
=== NAME  TestReconcileIAMAuditConfigResourceLevelCreate/resourcemanager/basic-folderinfolder
    runner.go:74: reconcile for Folder:jw5yhppa3xsx4yi/folder-jw5yhppa3xsx4yi took 110.626384ms, result was ({false 0s}, Update call failed: error expanding resource configuration for kind Folder: error getting referenced resource from API server: resource reference is missing required 'name' field)
```

Can you please help me to run this test locally?

pkg/test/resourcefixture/contexts/register.go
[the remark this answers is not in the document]
maqiuyujoyce: You're right! Sorry I misread the function (there are two GetCRDForGVK() in crdloader.go).

Though I'm still not a fan loading all CRDs (more than 300 files) and unmarshaling each into a CRD object repetitively.

pkg/test/resourcefixture/contexts/register.go
maqiuyujoyce: What I want to avoid is the repetitive CRD file reading, and I assume test cases for changing from TF/DCL in create.yaml -> Direct in update.yaml doesn't require updates in the CRD, but instead an annotation in the object. So can we store the CRD metadata (i.e. information about `TF2CRDLabel` and `DCL2CRDLabel`) into the ResourceContext and retrieve the info from ResourceContext in ReconcilerTypeForObject()?

pkg/controller/iam/auditconfig/iamauditconfig_controller_integration_test.go
maqiuyujoyce: AFAIK, we don't have any filter to just run a single test case in these integration tests. Feel free to hack them locally for testing purpose if you like, or support them as a new feature. Or you can let the whole suite run in the background. Note that because not all resources support specific IAM features, the test suite _may_ not be that large.

Here is an example command:
```
go test -v -tags=integration ./pkg/controller/iam/auditconfig -test.run ^TestReconcileIAMAuditConfigResourceLevelCreate$ -timeout 900s
```

pkg/test/controller/reconciler/testreconciler.go
maqiuyujoyce: You are right that there is no API call. Please find my response at https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2837#discussion_r1791029847 and https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2837#discussion_r1791031633

pkg/test/runner/runner.go
maqiuyujoyce: Ah, the reason I asked was that I was not sure why it is refactored here. Then I realized that you replaced `ReconcileIfManagedByKCC()` with `Reconcile()`. This change looks good to me.

yuwenma:
/approve

Thank you for improving the periodic tests! I'll defer to @maqiuyujoyce for the LGTM

pkg/controller/iam/auditconfig/iamauditconfig_controller_integration_test.go
maqiuyujoyce: Ah, glad to see you figured out how to test it.

Regarding `resource reference is missing required 'name' field` error message, can you verify that you've configured the `TEST_FOLDER_ID` env var? I think it looks for `folderRef.name` because `folderRef.external` is empty.

jasonvigil: Ah, thank you @maqiuyujoyce ! Was missing th env var. I have now verified this change with one test case from each of the IAM categories.

pkg/test/resourcefixture/contexts/register.go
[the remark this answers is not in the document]
jasonvigil: Ok I have updated this change to use the in-memory GVK map to avoid repetitively loading the CRDs. Though, we still need to load the CRD each time we reconcile in the tests (for the schema updater), but this is the same as before. In the future, maybe we can explore some ways to improve (or remove?) the schema updater so that we would be able to avoid this.

pkg/test/resourcefixture/contexts/register.go
jasonvigil: Fair enough. I have proposed a solution to use the GVK map here: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2837#discussion_r1792643849

maqiuyujoyce:
/lgtm

Thank you for the refactoring, @jasonvigil !

PR #2838, opened 2024-10-01
docs: modifications on SciFi migraiton documents

xiaoweim:

### DIFF
--- a/docs/develop-resources/api-conventions/validations.md
+++ b/docs/develop-resources/api-conventions/validations.md
@@ -80,6 +80,8 @@ One future improvement can be having generate-types detect the “Immutable.” 
 
 Parent field should be required and immutable.
 
+Note that parent is not a field name. It refers to the pre-requisite of the resource in question. You can identify the parent field through to the cloud resource's parent segment in the URL. For example, for GKE clusters, the [GET request URL](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.zones.clusters/get) `https://container.googleapis.com/v1/projects/{projectId}/zones/{zone}/clusters/{clusterId}` has `projectId` and `zone` in the parent segment. Therefore, for GKE clusters, projectRef and zone will be the parent fields.
+
 
 ### Required
 
--- a/docs/develop-resources/guides/3-add-mapper.md
+++ b/docs/develop-resources/guides/3-add-mapper.md
@@ -17,6 +17,8 @@ go run . generate-mapper \
    --api-dir $REPO_ROOT/apis 
 ```
 
+**Note**: We suggest using the same proto for your mock GCP and for you type-generation tool to generate the Config Connector API and mapper to avoid mismatch in schema definitions. There are some exceptions where you need to [replace the proto go package](https://github.com/xiaoweim/k8s-config-connector/blob/master/dev/tools/controllerbuilder/pkg/codegen/mappergenerator.go#L132).
+
 ### Simple path
 
 If no comments marked `MISSING` and all the mapper functions look good, you are done. You can move to “Add the controller” (step 4).
--- a/docs/develop-resources/guides/4-add-controller.md
+++ b/docs/develop-resources/guides/4-add-controller.md
@@ -45,6 +45,8 @@ hack/compare-mock fixtures/<your_resource_test>
 KCC_USE_DIRECT_RECONCILERS=<YOUR KIND> hack/compare-mock fixtures/<your_resource_test>
 ```
 
+ * Note: Differences in http.logs such as `user-agent` and method url are expected. Please regenerate test logs before `compare-mock`.
+
 ### Exit Criteria
 
 * The PRs shall pass the MockGCP tests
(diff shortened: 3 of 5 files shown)


docs/develop-resources/guides/3-add-mapper.md
```diff
@@ -17,6 +17,8 @@ go run . generate-mapper \
    --api-dir $REPO_ROOT/apis 
 ```
 
+Please use the same proto as your mock GCP tests to avoid mismatch in schema definitions. There are some exceptions where you need to [replace the proto go package](https://github.com/xiaoweim/k8s-config-connector/blob/master/dev/tools/controllerbuilder/pkg/codegen/mappergenerator.go#L132).
```
yuwenma, line 20: We can recommend users to use the same proto, but maybe not saying "please use ... to avoid mismatch". They can use different protos, as long as the serialized data is the same.

docs/develop-resources/api-conventions/validations.md
```diff
@@ -80,6 +80,8 @@ One future improvement can be having generate-types detect the “Immutable.”
 
 Parent field should be required and immutable.
 
+Note that parent is not a field name, it refers to the cloud resources parent segment in the URL. For example, for GKE clusters, the [GET request URL](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.zones.clusters/get) `https://container.googleapis.com/v1/projects/{projectId}/zones/{zone}/clusters/{clusterId}` has `projectId` and `zone` in the parent segment. Therefore, for GKE clusters, projectRef and zone will be the parent fields.
+
```
yuwenma, line 84: Let's explain "by code" (in the below session, we give a parent example.

docs/develop-resources/guides/4-add-controller.md
```diff
@@ -45,6 +45,15 @@ hack/compare-mock fixtures/<your_resource_test>
 KCC_USE_DIRECT_RECONCILERS=<YOUR KIND> hack/compare-mock fixtures/<your_resource_test>
 ```
 
+### Turn on SciFi Controller
+
+To fully turn on the SciFi controller, add your resource to [KCC_USE_DIRECT_RECONCILERS](https://github.com/xiaoweim/k8s-config-connector/blob/master/dev/tasks/run-e2e#L27) and make sure it passes all tests and there is a decent test coverage over customer use cases. Usually we do not turn it on for beta resources right away.
+ 
+
+ * Note: some resources are used as dependency resources in other test scenarios, to only test the direct controller with your resource test, you can use this annotation `alpha.cnrm.cloud.google.com/reconciler: "direct"` in the test yamls instead of using the `KCC_USE_DIRECT_RECONCILERS` flag at CLI.
```
yuwenma, line 53: Let's not introduce these two options right now. It could cause some confusions

mockgcp/README.md
```diff
@@ -36,6 +36,9 @@ Broadly the steps are:
 
    * Note: Ensure you pick the same version the controller uses to call the GCP API during GCP resource instantiation.
 
+   * Note: If the resource does not map to an actual GCP resource, you might not be able to find a proto file
```
yuwenma, line 39: I like that you share the experience and give a concrete example. I think this may not belong to the mockGCP session, because the MockGCP does not care ( or do not know ) the KCC objects.

docs/develop-resources/guides/3-add-mapper.md
xiaoweim: I see, thank you!

docs/develop-resources/api-conventions/validations.md
xiaoweim: I do think the code below does a very good job explaining how the parent field is set. I only added this explanation here, IMO is more about how to identify the parent fields of a resource. I had issues correctly identifying and understanding the parent fields.

docs/develop-resources/guides/4-add-controller.md
xiaoweim: Sure, my point is actually focused on that maybe we should not switch the resource to using direct controller right away after migration, but at the same time also test the direct controller in our e2e tests. That is why I mentioned `alpha.cnrm.cloud.google.com/reconciler: "direct"` annotation. I am not sure if there's a better way to emphasize on this point. But I see there could be some confusion rising from this.

docs/develop-resources/api-conventions/validations.md
yuwenma: ok, fair enough.

mockgcp/README.md
xiaoweim: I see your point. Do you have any idea where it would be better suited? 

I wanted to add this because I had some issues initially, trying to figure out the appropriate proto to use for the virtual resource bigquerydatasetaccess. I was following the README guide here under MockGCP when I encountered this, so I added here.

yuwenma (commented):
/lgtm
/approve

Thanks! Looks good!

PR #2845, opened 2024-10-02
test: Support TargetGrpcProxy for direct ComputeForwardingRule

gemmahou:
### Change description

- Add mock service for ComputeTargetGrpcProxy
- Add test case globalcomputeforwardingrulegrpcproxy

### Tests you have done


mockgcp/mockcompute/globalforwardingrulesv1.go
```diff
@@ -125,6 +125,12 @@ func (s *GlobalForwardingRulesV1) Insert(ctx context.Context, req *pb.InsertGlob
 		}
 	}
 
+	// network field is only used for global internal load balancing.
+	// If neither subnetwork nor network field is specified, the default network will be used.
+	if obj.Network == nil && obj.Subnetwork == nil && obj.LoadBalancingScheme != nil && *obj.LoadBalancingScheme != "EXTERNAL" {
```
acpana, line 130: for my own learning, could I get a pointer to the api here?

mockgcp/mockcompute/regionalbackendservicev1.go
```diff
@@ -41,7 +41,7 @@ func (s *RegionalBackendServicesV1) Get(ctx context.Context, req *pb.GetRegionBa
 
 	obj := &pb.BackendService{}
 	if err := s.storage.Get(ctx, fqn, obj); err != nil {
-		return nil, err
+		return nil, status.Errorf(codes.NotFound, "The resource '%s' was not found", fqn)
```
acpana, line 44: I don't think we want to capitalize these -> `the resource ...`

mockgcp/mockcompute/targetgrpcproxyv1.go
```diff
@@ -0,0 +1,141 @@
...
+func (n *targetGrpcProxyName) String() string {
+	return "projects/" + n.Project.ID + "/global/targetGrpcProxies/" + n.Name
+}
+
+// parseTargetGrpcProxyName parses a string into a targetGrpcProxyName.
+// The expected form is `projects/*/global/parseTargetGrpcProxies/*`.
+func (s *MockService) parseTargetGrpcProxyName(name string) (*targetGrpcProxyName, error) {
+	tokens := strings.Split(name, "/")
+
+	if len(tokens) == 5 && tokens[0] == "projects" && tokens[3] == "targetGrpcProxies" {
```
acpana, line 126: hm, I think we should also check that `tokens[2]` === "global"`

acpana (commented):
I think this lgtm. Just one quick note on the parsing in mockgcp

mockgcp/mockcompute/regionalbackendservicev1.go
gemmahou: I made this change to make sure the mock message matches the realGCP.

mockgcp/mockcompute/targetgrpcproxyv1.go
gemmahou: TargetGrpcProxy is a global only resource([API doc](https://cloud.google.com/compute/docs/reference/rest/v1/targetGrpcProxies)), so it has to be global

mockgcp/mockcompute/globalforwardingrulesv1.go
gemmahou: Do you mean comparing pointers? that will cause issues since the pointers are different, I want to check the value here... Not sure if there's a better way to achieve that.

mockgcp/mockcompute/targetgrpcproxyv1.go
gemmahou: Justin actually made a similar comment in another PR😅 I think you are right we'd better to check the "global" as well.

acpana:
/lgtm
/approve

thanks for making these changes

/hold 

I'll let you merge this in

PR #2851, opened 2024-10-03
refactor: make add cmd kind aware

acpana:
Breaking off from https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2839, cherry-picked the controller part of the effort which is more straightforward than the types gen.


jingyih (approved):
/lgtm

PR #2852, opened 2024-10-03
chore: Remove '+optional' marker from types template

jasonvigil:
It is optional (because omitempty is specified in the struct tag), and
our docs recommend only specifying '+required' markers.

Follow-on to https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2833#issuecomment-2389692484

### DIFF
--- a/apis/bigquery/v1beta1/bigquerydataset_types.go
+++ b/apis/bigquery/v1beta1/bigquerydataset_types.go
@@ -28,7 +28,6 @@ var BigQueryDatasetGVK = GroupVersion.WithKind("BigQueryDataset")
 // +kcc:proto=google.cloud.bigquery.v2.dataset
 type BigQueryDatasetSpec struct {
 	// The BigQueryDataset name. If not given, the metadata.name will be used.
-	// +optional
 	ResourceID *string `json:"resourceID,omitempty"`
 
 	// An array of objects that define dataset access for one or more entities.
--- a/apis/bigqueryconnection/v1alpha1/connection_types.go
+++ b/apis/bigqueryconnection/v1alpha1/connection_types.go
@@ -39,7 +39,6 @@ type BigQueryConnectionConnectionSpec struct {
 
 	// The BigQuery ConnectionID. This is a server-generated ID in the UUID format.
 	// If not provided, ConfigConnector will create a new Connection and store the UUID in `status.serviceGeneratedID` field.
-	// +optional
 	ResourceID *string `json:"resourceID,omitempty"`
 
 	// User provided display name for the connection.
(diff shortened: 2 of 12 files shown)


maqiuyujoyce:
This change LGTM. Though can it be enforced? Naturally people will add optional or required marker and I assume we don't plan to rely on reviewers to do manual checks of these not so important details.

jingyih:
Did we try to run "dev/tasks/generate-crds" and double check that the CRDs are not affected?

jasonvigil:
> This change LGTM. Though can it be enforced? Naturally people will add optional or required marker and I assume we don't plan to rely on reviewers to do manual checks of these not so important details.

Do you think it is necessary / desirable to enforce that `+optional` isn't specified in CRD fields? We could probably add a linter to check this during CRD generation, but I feel like because it's a (relatively) harmless annotation that has no real impact, I think that's not really worth the effort. This change was just intended to be a quick, low effort fix to update our template, to stop the proliferation for newly-generated types.

jasonvigil:
> Did we try to run "dev/tasks/generate-crds" and double check that the CRDs are not affected?

There is no diff after running `dev/tasks/generate-crds`

jingyih:
> > Did we try to run "dev/tasks/generate-crds" and double check that the CRDs are not affected?
> 
> There is no diff after running `dev/tasks/generate-crds`

Cool. Thanks for confirming!

jingyih (approved):
/lgtm

jingyih:
> This change LGTM. Though can it be enforced? Naturally people will add optional or required marker and I assume we don't plan to rely on reviewers to do manual checks of these not so important details.

I don't have a strong opinion on whether we want to enforce this rule. If we decided to enforce it, I think we can have a separate PR to add a validation / test.

maqiuyujoyce:
> > This change LGTM. Though can it be enforced? Naturally people will add optional or required marker and I assume we don't plan to rely on reviewers to do manual checks of these not so important details.
> 
> Do you think it is necessary / desirable to enforce that `+optional` isn't specified in CRD fields? We could probably add a linter to check this during CRD generation, but I feel like because it's a (relatively) harmless annotation that has no real impact, I think that's not really worth the effort. This change was just intended to be a quick, low effort fix to update our template so that it matches our documented recommendation
> 
> https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/bac1bde670beed1a0061dcbf34a2a884a326bd4b/docs/develop-resources/api-conventions/validations.md?plain=1#L38-L42

I think it's desirable to enforce one specific behavior when a CRD field is optional. Flexibility causes confusion, and thus cost additional time in making a decision.
The enforced behavior can either be a `+optional` tag is always required or no `+optional` tag should be configured. I personally don't have a preference.

PR #2853, opened 2024-10-03
feat: Watch resources to record metrics instead of List

jasonvigil:
This will help with system load. With this change, K8s API server no longer needs to collect all resources when we fetch metrics.

### DIFF
--- a/cmd/recorder/main.go
+++ b/cmd/recorder/main.go
@@ -35,21 +35,14 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	flag "github.com/spf13/pflag"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


jasonvigil:
@justinsb how can we test this change?

jingyih:
To test local changes in controllers, I used to run this. Hopefully it still works.

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/748718d642683fa5d9cbfc8c856dea49191365b8/Makefile#L212

yuwenma:
Suggest adding some tests or/and benchmark to show the result.

jasonvigil:
@yuwenma, we don't seem to have any existing tests for this code, but I did a manual benchmark here: https://diff.googleplex.com/#key=Opd89gWnrjGg. The reported metrics are the same, but (even in this small example), the new code uses less memory.

yuwenma:
/lgtm
/approve

awesome!

PR #2858, opened 2024-10-04
fix: Update ref template to return ReferenceNotReadyError

jasonvigil:
This way, we will properly detect the source of the error being an unresolveable dependency and can requeue immediately once it's ready.


yuwenma (commented):
/lgtm
/approve

Nice

PR #2862, opened 2024-10-05
Allow CC-BY-3.0 license

maqiuyujoyce:
### Change description

license-lint became flaky after `github.com/apache/arrow/go/v15` is added to go.mod and sometimes fails with the following error at master HEAD:

```
ERROR: Some modules have unrecognized licenses:
  github.com/apache/arrow/go/v15: similar to CC-BY-3.0, 1.000000 confidence, path '/home/runner/go/pkg/mod/github.com/apache/arrow/go/v15@v15.0.2/LICENSE.txt'
```

Considering that `CC-BY-3.0` is a notice license that should be fine to allow, we should just add it to `unrestricted_licenses` list in the config.

### Tests you have done

### DIFF
--- a/.github/license-lint-config.yml
+++ b/.github/license-lint-config.yml
@@ -18,6 +18,7 @@ unrestricted_licenses:
   - BSD-3-Clause
   - BSD-2-Clause
   - 0BSD
+  - CC-BY-3.0
 
 reciprocal_licenses:
   - MPL-2.0


jasonvigil:
@acpana why did you update the licensedb dependency here? https://github.com/istio/tools/commit/4054415f72c76fb65f422860bf5995ca22eeda9d

I wonder if the flake issue would be fixed if we upgraded our license-lint version?

maqiuyujoyce:
> @acpana why did you update the licensedb dependency here? [istio/tools@4054415](https://github.com/istio/tools/commit/4054415f72c76fb65f422860bf5995ca22eeda9d)
> 
> I wonder if the flake issue would be fixed if we upgraded our license-lint version?

I kinda doubt - I experimented against master HEAD of istio/tools and the result is still inconsistent. One way to deflake it could be turning `res` [here](https://github.com/istio/tools/blob/master/cmd/license-lint/analysis.go#L62) into a slice and sort the items before using it.

justinsb (approved):

PR #2863, opened 2024-10-07
feat: Direct controller for ComputeFirewallPolicyRule

gemmahou:
### Tests you have done


pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
```diff
@@ -0,0 +1,310 @@
...
+	priority               int64
+	firewallPoliciesClient *gcp.FirewallPoliciesClient
+	desired                *krm.ComputeFirewallPolicyRule
+	actual                 *computepb.FirewallPolicyRule
+	reader                 client.Reader
+}
+
+var _ directbase.Adapter = &firewallPolicyRuleAdapter{}
+
+func (m *firewallPolicyRuleModel) AdapterForObject(ctx context.Context, reader client.Reader, u *unstructured.Unstructured) (directbase.Adapter, error) {
```
gemmahou, line 62: I do not have externalRef for this resource yet, this resource does not have unique identifier, and since it's a field of FirewallPolicy, I do not think there'll be other resource reference to it. No sure if we want to add create an externalRef for virtual resource like this.

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computefirewallpolicyrule/computefirewallpolicyrule-minimal/_http.log
```diff
@@ -208,14 +209,14 @@ X-Xss-Protection: 0
...
+POST https://compute.googleapis.com/compute/v1/locations/global/firewallPolicies/${firewallPolicyID}/addRule
 Content-Type: application/json
-User-Agent: kcc/controller-manager DeclarativeClientLib/0.0.1
+User-Agent: kcc/controller-manager
+x-goog-request-params: firewall_policy=${firewallPolicyId}
 
 {
   "action": "deny",
   "direction": "INGRESS",
-  "firewallPolicy": "locations/global/firewallPolicies/${firewallPolicyId}",
```
gemmahou, line 211: Direct controller does not have `firewallPolicy` in request body, while DCL controller does. 
In the API [doc](https://cloud.google.com/compute/docs/reference/rest/v1/firewallPolicies/addRule), `firewallPolicy` is path parameter, not a field in the request body. I think this diff is acceptable.

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computefirewallpolicyrule/computefirewallpolicyrule-egress-full/_http.log
```diff
@@ -905,7 +803,7 @@ X-Xss-Protection: 0
     ]
   },
   "priority": 9000,
-  "ruleTupleCount": 109,
+  "ruleTupleCount": 4,
```
gemmahou: the `ruleTupleCount` is generated, it's used to determine the complexity of a rule. To make it easier, in the mock server, if a rule is more complicated, I assign it to 4, otherwise to 2. So this diff is expected.

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computefirewallpolicyrule/computefirewallpolicyrule-egress-full/_http.log
```diff
@@ -735,13 +640,6 @@ X-Xss-Protection: 0
 {
   "error": {
     "code": 400,
-    "errors": [
-      {
-        "domain": "global",
-        "message": "Invalid value for field 'priority': '9000'. The firewall policy does not contain a rule at priority 9000.",
-        "reason": "invalid"
-      }
-    ],
```
gemmahou: The errors field comes from mock memory([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/4e8eb92702823e8caa3abff45b6d88b37f36a117/mockgcp/pkg/storage/memory.go#L132)).

In the mock server, because rule is a virtual resource without unique identifier, I did not save it in to the mock storage. If a certain rule is not found in a policy, I log an error message directly, so it's expected to not have the extra "errors" field in the error response.

pkg/controller/direct/compute/firewallpolicyrule/client.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	return gcpClient, nil
+}
+
+func (m *gcpClient) options() ([]option.ClientOption, error) {
+	var opts []option.ClientOption
+	if m.config.UserAgent != "" {
+		opts = append(opts, option.WithUserAgent(m.config.UserAgent))
+	}
+	if m.config.HTTPClient != nil {
```
yuwenma, line 43: This gcpClient config seems to be handled in the template already. Can you see if this works for you? https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/dev/tools/controllerbuilder/template/controller/controller.go#L92

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
```diff
@@ -0,0 +1,290 @@
...
+
+	// Get the deleted rules
+	_, err = a.get(ctx)
+	if err != nil {
+		return true, nil
+	}
+	return true, nil
+}
+
+func (a *firewallPolicyRuleAdapter) get(ctx context.Context) (*computepb.FirewallPolicyRule, error) {
```
yuwenma, line 284: maybe add back the code inline is more readable.

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
```diff
@@ -0,0 +1,290 @@
...
+			return false, fmt.Errorf("waiting ComputeFirewallPolicyRule %d delete failed: %w", a.priority, err)
+		}
+	}
+	log.V(2).Info("successfully deleted ComputeFirewallPolicyRule", "priority", a.priority)
+
+	// Get the deleted rules
+	_, err = a.get(ctx)
+	if err != nil {
+		return true, nil
+	}
```
yuwenma, line 280: We don't need this. It makes an additional GCP call and is covered by above code ln268-274

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
```diff
@@ -0,0 +1,290 @@
...
+func (a *firewallPolicyRuleAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOperation) error {
+	var err error
+
+	err = resolveDependencies(ctx, a.reader, a.desired)
+	if err != nil {
+		return err
+	}
+
+	log := klog.FromContext(ctx).WithName(ctrlName)
+	log.V(2).Info("updating ComputeFirewallPolicyRule", "priority", a.priority)
```
yuwenma, line 187: nit: either remove this log or move it before the GCP call.

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
```diff
@@ -0,0 +1,290 @@
...
+	mapCtx := &direct.MapContext{}
+
+	desired := a.desired.DeepCopy()
+	firewallPolicyRule := ComputeFirewallPolicyRuleSpec_ToProto(mapCtx, &desired.Spec)
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+	firewallPolicyRule.Priority = nil
```
yuwenma, line 195: Shall we put this inside the mapper function? If there's a reason to not to, could you add a comment?

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
```diff
@@ -0,0 +1,290 @@
...
+
+	op := &gcp.Operation{}
```
yuwenma, line 197: nit: remove this line

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
```diff
@@ -0,0 +1,290 @@
...
+	// Get the updated resource
+	updated, err = a.get(ctx)
+	if err != nil {
+		return fmt.Errorf("getting ComputeFirewallPolicyRule %d: %w", a.priority, err)
+	}
+
+	status := &krm.ComputeFirewallPolicyRuleStatus{
+		RuleTupleCount: direct.PtrTo(int64(*updated.RuleTupleCount)),
+		Kind:           direct.PtrTo("compute#firewallPolicyRule"),
+	}
```
yuwenma, line 226: I think(?) this is not needed because https://github.com/gemmahou/k8s-config-connector/blob/b66713a2ee2c8a30842ec3cb8a5beb3a978f9055/pkg/controller/direct/compute/firewallpolicyrule/mapper.generated.go#L68

pkg/controller/direct/compute/firewallpolicyrule/refs.go
```diff
@@ -0,0 +1,212 @@
...
+	resourceID, err := refs.GetResourceID(computeFirwallPolicy)
+	if err != nil {
+		return nil, err
+	}
+
+	return &refs.ComputeFirewallPolicyRef{
+		External: fmt.Sprintf("%s", resourceID)}, nil
+}
+
+func ResolveComputeNetwork(ctx context.Context, reader client.Reader, src client.Object, ref *refs.ComputeNetworkRef) (*refs.ComputeNetworkRef, error) {
```
yuwenma, line 75: We should use the new reference format for the compute service resources,
https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/refs/v1beta1/interface.go

Not a blocker to this PR, you can do that in follow up PRs.

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
yuwenma: Maybe we can use the firewall policy external ref + "/rules/" + priority

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
```diff
@@ -0,0 +1,290 @@
...
+	firewallPolicyRef, err := ResolveComputeFirewallPolicy(ctx, reader, obj, obj.Spec.FirewallPolicyRef)
+	if err != nil {
+		return nil, err
+
+	}
+	obj.Spec.FirewallPolicyRef.External = firewallPolicyRef.External
+	firewallPolicy := obj.Spec.FirewallPolicyRef.External
+
+	// Get priority
+	priority := obj.Spec.Priority
```
yuwenma, line 81: What is priority? This seems to be the firewall policy rule's ID.

pkg/controller/direct/maputils.go
```diff
@@ -193,6 +193,11 @@ func IsNotFound(err error) bool {
 	return HasHTTPCode(err, 404)
 }
 
+// IsInvalidValue returns true if the given error is an HTTP 400.
+func IsInvalidValue(err error) bool {
+	return HasHTTPCode(err, 400)
```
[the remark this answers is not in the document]
gemmahou: Updated the method name to Bad request.

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
gemmahou: It's called several times, and I got a suggest previously to make it to a separate function. I can add a comment to make it clearer.

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
gemmahou: I added this because there was no additional GET call to verify the rule has been deleted. Removing the GET call here will cause a missing GET request compared to the TF controller.

pkg/controller/direct/compute/firewallpolicyrule/client.go
gemmahou: Updated.

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
gemmahou: Good point! updated to use `ComputeFirewallPolicyRuleStatus_FromProto`

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
gemmahou: I do not think we can use priority itself as **unique** ID. Different rule with same priority can be associated with different policy, i.e.

policy1: {rule1(priority 100), rule2(priority 300)}
policy2: {rule3(priority 100), rule4(priority 300)}

If we include the policy i.e. `policies/policyName/priorities/priorityNumber` that could be used as unique identifier. It's still a bit "unusual" because this is not an expression exist in GCP. The GCP request to get a rule looks like:` GET https://compute.googleapis.com/compute/v1/locations/global/firewallPolicies/${firewallPolicyID}/getRule?priority=9000`, not `GET https://compute.googleapis.com/compute/v1/locations/global/firewallPolicies/${firewallPolicyID}/priorities/9000`

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
gemmahou: Added a comment. For patch request, the priority field is not a part of the request body, but a query parameter.

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
gemmahou: Added externalRef in #3035

apis/compute/v1beta1/firewallpolicyrule_types.go
```diff
@@ -20,6 +20,15 @@ import (
 	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
 	commonv1alpha1 "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/apis/common/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+var (
+	ComputeFirewallPolicyRuleGVK = schema.GroupVersionKind{
```
justinsb, line 27: Nit: you can also do `GroupVersion.WithKind("ComputeFirewallPolicyRule")` IIRC

dev/tasks/run-e2e
```diff
@@ -26,7 +26,7 @@ if [[ -z "${KUBEBUILDER_ASSETS:-}" ]]; then
 fi
 
 if [[ -z "${KCC_USE_DIRECT_RECONCILERS:-}" ]]; then
-  KCC_USE_DIRECT_RECONCILERS=ComputeForwardingRule,GKEHubFeatureMembership,SecretManagerSecret
+  KCC_USE_DIRECT_RECONCILERS=ComputeFirewallPolicyRule,ComputeForwardingRule,GKEHubFeatureMembership,SecretManagerSecret
```
justinsb, line 29: Nit: the downside of this is that we currently lose test coverage on the non-direct controller.  I've been leaning towards copying the tests and using the annotation to force direct mode.

mockgcp/mock_http_roundtrip.go
```diff
@@ -326,6 +326,15 @@ func (m *mockRoundTripper) prefilterRequest(req *http.Request) error {
...
 		}
+	} else {
+		// When sending a delete request for a ComputeFirewallPolicyRule resource,
+		// The request URL looks like POST https://compute.googleapis.com/compute/v1/locations/global/firewallPolicies/${firewallPolicyID}/removeRule.
+		// It's uncommon to use POST requests for delete operations, and a nil request body for POST method is unexpected,
+		// I got the "missing form body" error. Ref: https://go.dev/src/net/http/request.go?s=41070:41129 line 1340
+		// So instead of sending a nil request body, send an empty request body to ensure successful processing of the remove rule request.
+		body := &bytes.Buffer{}
+		b := body.Bytes()
+		req.Body = io.NopCloser(bytes.NewBuffer(b))
```
justinsb, line 337: Nit: I _think_ this is the same as:

```
		body := &bytes.Buffer{}
		req.Body = io.NopCloser(body)
```

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+func (m *firewallPolicyRuleModel) AdapterForObject(ctx context.Context, reader client.Reader, u *unstructured.Unstructured) (directbase.Adapter, error) {
+	obj := &krm.ComputeFirewallPolicyRule{}
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &obj); err != nil {
+		return nil, fmt.Errorf("error converting to %T: %w", obj, err)
+	}
+
+	// Set label managed-by-cnrm: true
+	obj.ObjectMeta.Labels["managed-by-cnrm"] = "true"
```
justinsb, line 84: Nit: we should build a helper function for building the GCP labels.  There may be one in SQLInstance

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
```diff
@@ -0,0 +1,297 @@
...
+	return firewallPolicyRuleAdapter, nil
+}
+
+func (m *firewallPolicyRuleModel) AdapterForURL(ctx context.Context, url string) (directbase.Adapter, error) {
+	// TODO: Support URLs
+	return nil, nil
+}
+
+func (a *firewallPolicyRuleAdapter) Find(ctx context.Context) (bool, error) {
+	log := klog.FromContext(ctx).WithName(ctrlName)
```
justinsb, line 121: Nit: We've stopped using WithName, in the adapter I believe.  The log in the context should have the name (or something pretty similar)

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
```diff
@@ -0,0 +1,297 @@
...
+		return false, fmt.Errorf("getting ComputeFirewallPolicyRule %d: %w", a.priority, err)
+	}
+	a.actual = firewallPolicyRule
+	return true, nil
+}
+
+func (a *firewallPolicyRuleAdapter) Create(ctx context.Context, createOp *directbase.CreateOperation) error {
+	var err error
+
+	err = resolveDependencies(ctx, a.reader, a.desired)
```
justinsb, line 141: Nit: If you do `err := ...` then I think you won't need to declare `err` on line 139

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
```diff
@@ -0,0 +1,297 @@
...
+	if err != nil {
+		return err
+	}
+
+	log := klog.FromContext(ctx).WithName(ctrlName)
+	log.V(2).Info("creating ComputeFirewallPolicyRule", "priority", a.priority)
+	mapCtx := &direct.MapContext{}
+
+	desired := a.desired.DeepCopy()
```
justinsb, line 150: Do we need to DeepCopy?

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	req := &computepb.AddRuleFirewallPolicyRequest{
+		FirewallPolicyRuleResource: firewallPolicyRule,
+		FirewallPolicy:             a.firewallPolicy,
+	}
+	op, err := a.firewallPoliciesClient.AddRule(ctx, req)
+
+	if err != nil {
+		return fmt.Errorf("creating ComputeFirewallPolicyRule %d: %w", a.priority, err)
```
justinsb, line 164: If we start using `a.id` I think this will get clearer (I _think_ the full id is actually going to be project + firewallRule + priority)

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
```diff
@@ -0,0 +1,297 @@
...
+	if !op.Done() {
+		err = op.Wait(ctx)
+		if err != nil {
+			return fmt.Errorf("waiting ComputeFirewallPolicyRule %d create failed: %w", a.priority, err)
+		}
+	}
+	log.V(2).Info("successfully created ComputeFirewallPolicyRule", "priority", a.priority)
+
+	// Get the created resource
+	created := &computepb.FirewallPolicyRule{}
```
justinsb, line 175: Nit: no reason to allocate it and then reassign it on the next line

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
```diff
@@ -0,0 +1,297 @@
...
+
+	mc := &direct.MapContext{}
+	spec := ComputeFirewallPolicyRuleSpec_FromProto(mc, a.actual)
+	specObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(spec)
+	if err != nil {
+		return nil, fmt.Errorf("error converting firewallPolicyRule spec to unstructured: %w", err)
+	}
+
+	u := &unstructured.Unstructured{
+		Object: make(map[string]interface{}),
```
justinsb, line 250: Nit: you don't need to initialize Object, SetGroupVersionKind does it for you (IIRC)

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
```diff
@@ -0,0 +1,297 @@
...
+func (a *firewallPolicyRuleAdapter) Delete(ctx context.Context, deleteOp *directbase.DeleteOperation) (bool, error) {
+	log := klog.FromContext(ctx).WithName(ctrlName)
+	log.V(2).Info("deleting ComputeFirewallPolicyRule", "priority", a.priority)
+
+	delReq := &computepb.RemoveRuleFirewallPolicyRequest{
+		FirewallPolicy: a.firewallPolicy,
+		Priority:       direct.PtrTo(int32(a.priority)),
+	}
+	op, err := a.firewallPoliciesClient.RemoveRule(ctx, delReq)
+
```
justinsb, line 271: Nit: probably clearer to not have a blank line between the call and its error handling

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
justinsb: I think it's OK to have one fewer get, if TF is doing it wrong :-)

If you want to keep it to make the golden output look the same (until we fully get rid of the TF) then consider something like this:

```
// Temporarily add an extra GET to match Terraform behaviour
a.get(ctx)
```

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
```diff
@@ -0,0 +1,297 @@
...
+	if err != nil {
+		return true, nil
+	}
+	return true, nil
+}
+
+func (a *firewallPolicyRuleAdapter) get(ctx context.Context) (*computepb.FirewallPolicyRule, error) {
+	getReq := &computepb.GetRuleFirewallPolicyRequest{
+		FirewallPolicy: a.firewallPolicy,
+		Priority:       direct.PtrTo(int32(a.priority)),
```
justinsb, line 294: Nit: probably `a.priority` should be int32

pkg/controller/direct/compute/firewallpolicyrule/refs.go
```diff
@@ -0,0 +1,212 @@
...
+		return nil, err
+	}
+
+	projectID, err := refs.ResolveProjectID(ctx, reader, computeNetwork)
+	if err != nil {
+		return nil, err
+	}
+
+	return &refs.ComputeNetworkRef{
+		External: fmt.Sprintf("https://www.googleapis.com/compute/v1/projects/%s/global/networks/%s", projectID, resourceID)}, nil
```
justinsb, line 120: Nit: I would have expected `projects/%s/...` (no "https://...").  We should try to match the format that we want users to use.

pkg/controller/direct/compute/firewallpolicyrule/refs.go
```diff
@@ -0,0 +1,212 @@
...
+			targetResource.External = networkRef.External
+			targetResources = append(targetResources, targetResource)
+		}
+		obj.Spec.TargetResources = targetResources
+	}
+	// Get target service accounts
+	var targetServiceAccounts []*refs.IAMServiceAccountRef
+	if obj.Spec.TargetServiceAccounts != nil {
+		for _, targetServiceAccount := range obj.Spec.TargetServiceAccounts {
+			iamServiceAccount, err := ResolveIAMSetviceAccount(ctx, reader, obj, targetServiceAccount)
```
justinsb, line 202: Nit: typo in ServiceAccount

justinsb:
A lot of nits, but nothing blocking.

/approve
/lgtm

/hold so you can see the nits, feel free to remove hold!

pkg/controller/direct/compute/firewallpolicyrule/refs.go
gemmahou: Reformat the value because we want to match the realGCP log. In realGCP log, it looks like:
```
  "targetResources": [
    "https://www.googleapis.com/compute/v1/projects/${projectId}/global/networks/network-${uniqueId}"
  ],
```
Also the target resource url requires the prefix, if I remove it, the creation will fail with error: `Invalid target resource url: projects/cnrm-yuhou/global/networks/network-bflfo2mcesdexjq`

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
gemmahou: We use DeepCopy() to avoid unintended modifications to the original a.desired. However, I do not see we currently modify this object anywhere later in the code... The DeepCopy() might be unnecessary, but it's no harm to keep it(?) just in case... since we don't want to change any desired values...

yuwenma (commented):
/lgtm
/approve

PR #2868, opened 2024-10-07
Avoid loading CRDs when registering the webhook

maqiuyujoyce:
### Change description

Webhook failed with `unable to locate repo root in working directory "/"` because supportedgvks.All() calls LoadCRDs() which checks the local git path. This won't work if the webhook is registered/created at the path where a local [k8s-config-connector](https://github.com/GoogleCloudPlatform/k8s-config-connector) Git repo doesn't exist. LoadCRDs() should only be called under the repo path.

Using supportedgvks.AllWithoutDirect() in the webhook instead to avoid loading CRDs.

### Tests you have done

### DIFF
--- a/pkg/webhook/register.go
+++ b/pkg/webhook/register.go
@@ -78,10 +78,7 @@ func GetCommonWebhookConfigs() ([]Config, error) {
 		return nil, fmt.Errorf("error getting new dcl schema loader: %w", err)
 	}
 	serviceMetadataLoader := metadata.New()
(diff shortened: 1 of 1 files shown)


maqiuyujoyce:
@jasonvigil looks like all presubmit tests passed. Can I get an approval?

jasonvigil:
/approve

@maqiuyujoyce ah, did not realize that /approve was also necessary

Issue #2869, opened 2024-10-07
All the CRD names and categories should be retrievable without loading the CRD files locally


maqiuyujoyce:
### Checklist

- [X] I did not find a related open enhancement request.
- [X] I understand that enhancement requests filed in the GitHub repository are by default low priority.
- [X] If this request is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Describe the feature or resource

Right now when we want to retrieve all kind/CRD names and their categories (TF-base/DCL-based/direct), we call [supportedgvks.All()](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/ace7489767a246333bbc0ca0b4c210d36d084b98/pkg/gvks/supportedgvks/supportedgvks.go#L32). However, this calls [supportedgvks.resourcesWithDirect()](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/ace7489767a246333bbc0ca0b4c210d36d084b98/pkg/gvks/supportedgvks/supportedgvks.go#L46) and eventually calls [repo.GetRootOrTestFatal()](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/ace7489767a246333bbc0ca0b4c210d36d084b98/pkg/util/repo/repo.go#L57), which assumes that the caller is within a local copy of our Git repo. However, if supportedgvks.All() is called in the controllers (webhooks, direct controllers, etc) that is not under a local copy of our git repo, this will fail with error like:

```
unable to locate repo root in working directory "/"
```

We should figure out a way to retrieve all kind/CRD names and their categories.

jasonvigil:
Ref to the original fix: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2868

PR #2871, opened 2024-10-08
Generate list of supported GVKs

jasonvigil:
Fixes https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/2869


pkg/gvks/supportedgvks/supportedgvks.go
```diff
@@ -68,34 +67,18 @@ func resources(smLoader *servicemappingloader.ServiceMappingLoader, serviceMetaL
...
+	for gvk, metadata := range SupportedGVKs {
+		if metadata.Labels["cnrm.cloud.google.com/tf2crd"] == "true" {
 			continue
 		}
-		if crd.ObjectMeta.Labels["cnrm.cloud.google.com/dcl2crd"] == "true" {
+		if metadata.Labels["cnrm.cloud.google.com/dcl2crd"] == "true" {
 			continue
 		}
-		versions := crd.Spec.Versions
-		highestVersion := k8s.KCCAPIVersionV1Alpha1
```
jasonvigil, line 88: @maqiuyujoyce I am curious why this logic was necessary to filter out only the highest supported direct gvk version? Wouldn't we want to return all supported versions?

pkg/gvks/supportedgvks/mapping.go
```diff
@@ -0,0 +1,4326 @@
...
+
+package supportedgvks
+
+import "k8s.io/apimachinery/pkg/runtime/schema"
+
+type GVKMetadata struct {
+	Labels map[string]string
+}
+
+var SupportedGVKs = map[schema.GroupVersionKind]GVKMetadata{
```
maqiuyujoyce, line 25: Could we add `generated` either in the file name or in the file path?

pkg/gvks/supportedgvks/supportedgvks.go
maqiuyujoyce: The highest version was added for reference doc generation: Only GVK containing v1beta1 version will have the reference doc, and the doc is based on version v1beta1.

Adding same GK with different versions should be fine as long as the reference doc generation is not impacted.

pkg/gvks/supportedgvks/mapping.go
jasonvigil: Yeah, that makes sense. Done

pkg/gvks/supportedgvks/supportedgvks.go
jasonvigil: Ah, cool. Looks like reference doc generation is unaffected because it filters out alpha versions. So I think we are good then.

maqiuyujoyce (commented):
/lgtm
/approve

PR #2885, opened 2024-10-09
dev:tools:fix: {

acpana:
Small quick fix so others don't run into this.


acpana:
We should probably have a light weight smoke check for these tools.

yuwenma (commented):
/lgtm
/approve

Thanks! I also have a PR to fix this. You beat me 😄

PR #2887, opened 2024-10-09
dev:tools: various small changes

acpana:
Best viewed by commit

### DIFF
--- a/dev/tools/controllerbuilder/template/apis/types.go
+++ b/dev/tools/controllerbuilder/template/apis/types.go
@@ -69,7 +69,8 @@ type {{ .Kind }}ObservedState struct {
 
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
-// +kubebuilder:resource:categories=gcp
+// TODO(user): make sure the pluralizaiton below is correct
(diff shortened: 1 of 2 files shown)


dev/tools/controllerbuilder/template/apis/types.go
```diff
@@ -83,6 +84,7 @@ type {{ .Kind }} struct {
 	metav1.TypeMeta   ` + "`" + `json:",inline"` + "`" + `
 	metav1.ObjectMeta ` + "`" + `json:"metadata,omitempty"` + "`" + `
 
+	// +required
```
acpana, line 87: I think this is fine to do because we expected to have the common case to always want to require spec

dev/tools/controllerbuilder/template/apis/types.go
```diff
@@ -69,7 +69,8 @@ type {{ .Kind }}ObservedState struct {
 
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
-// +kubebuilder:resource:categories=gcp
+// TODO(user): make sure the pluralizaiton below is correct
+// +kubebuilder:resource:categories=gcp,shortName=gcp{{ .Kind | ToLower }};gcp{{ .Kind | ToLower }}s
```
yuwenma, line 73: I like it.

yuwenma (commented):
/lgtm
/approve

PR #2890, opened 2024-10-10
feat: Record events when updating SQLInstances

jasonvigil:
This will help detect issues with the direct reconciler in the periodic tests.

Now, "testNoChange" will be able to detect when SQLInstance is incorrectly / unnecessarily performing updates.

Also, it provides a similar UX as the TF/DCL controllers, which record events and conditions on update.


jasonvigil:
@yuwenma / @justinsb -- for direct controllers, do we want to record conditions and events for updates (and other operations)? Similar to how TF/DCL do with https://github.com/jasonvigil/k8s-config-connector/blob/bf3b1ccd582833c53c64381d907ab6a647ace0eb/pkg/controller/lifecyclehandler/handler.go#L296.

This change only records events, but does not update status or set observed generation.

PR #2892, opened 2024-10-10
Remove non-configurable fields in PrivilegedAccessManagerEntitlement

maqiuyujoyce:
### Change description

In PrivilegedAccessManagerEntitlement, `projectRef`/`folderRef`/`organizationRef` and `resourceType` under `spec.privilegedAccess.gcpIamAccess` must match the parent container, so they are basically non-configurable fields. We decided to remove those fields from CRD and default them in the controller.

### Tests you have done


pkg/controller/direct/privilegedaccessmanager/entitlement_controller.go
```diff
@@ -283,15 +264,20 @@ func (a *Adapter) Create(ctx context.Context, createOp *directbase.CreateOperati
 	mapCtx := &direct.MapContext{}
 
 	desired := a.desired.DeepCopy()
-	resource := PrivilegedAccessManagerEntitlementSpec_ToProto(mapCtx, &desired.Spec)
+	resourceType, resource, err := getResourceTypeAndResourceFromContainer(a.id.Parent.Container)
+	if err != nil {
+		return fmt.Errorf("error getting resourceType and resource from container: %w", err)
+	}
+	hiddenFields := gcpIAMAccessResource{resourceType: resourceType, resource: resource}
+	entitlement := PrivilegedAccessManagerEntitlementSpec_ToProto(mapCtx, &desired.Spec, hiddenFields)
```
maqiuyujoyce, line 272: It looks weird if we pass in the hiddenFields to PrivilegedAccessManagerEntitlementSpec_ToProto() so I'm also exploring the option to improve the readability in #2893.

apis/privilegedaccessmanager/v1alpha1/types.generated.go
```diff
@@ -137,6 +135,9 @@ type PrivilegedAccess struct {
 // about IAM.
 // +kcc:proto=google.cloud.privilegedaccessmanager.v1.PrivilegedAccess.GcpIamAccess
 type GcpIamAccess struct {
+	/* The following fields should be hidden because their values must match
```
yuwenma, line 138: nit: Suggest listing the reasons for this decision so we can revisit when the condition changes (and I think it will): 
1. The current API has several redundant and duplicate fields between the parent URL and the IAMAccess. The usage is unclear.
2. It is cheaper to add a new API field than changing an existing one (and make it backward compatible)
3. We have pending discussions with PAM to better understand the API. But we don't want to be blocked right now.

yuwenma (commented):
/lgtm
/approve

apis/privilegedaccessmanager/v1alpha1/types.generated.go
maqiuyujoyce: Done.

PR #2893, opened 2024-10-10
[WIP] Use customization func to remove non-configurable fields in PrivilegedAccessManagerEntitlement

maqiuyujoyce:
5e1dab7b8fbf26c5457a37716df2f32e3ec4eea6 is an alternative option to implement #2892. Passing in the customization function (and in the future generalizing the customization function) seems to look more reasonable comparing with just passing in the additional values.


maqiuyujoyce:
PR existed for prototyping only.

PR #2895, opened 2024-10-10
chore: secretManagerSecret use new external ref

yuwenma:
diff base from #2797 ;fix some minor bugs


apis/secretmanager/v1beta1/secret_reference.go
```diff
@@ -0,0 +1,181 @@
...
+	// Should be in the format "projects/<projectID>/locations/<location>/secrets/<secretID>".
+	External string `json:"external,omitempty"`
+
+	// The name of a SecretManagerSecret resource.
+	Name string `json:"name,omitempty"`
+
+	// The namespace of a SecretManagerSecret resource.
+	Namespace string `json:"namespace,omitempty"`
+
+	parent *SecretManagerSecretParent
```
gemmahou, line 45: nit: suggest adding a comment saying that this field is optional. Some resources does not necessary to have `parent` field.

apis/secretmanager/v1beta1/secret_reference.go
gemmahou, line 1: No sure if it's better to put this filed under apis/refs/v1beta1/?

yuwenma: Can you share more on how you think the ./apis/refs is better?

apis/secretmanager/v1beta1/secret_reference.go
yuwenma: do you mean adding the comment in the code template? We can do that, but I would assume the contributors can figure it out themselves.

apis/secretmanager/v1beta1/secret_reference.go
gemmahou: This file defines resource references and associated logic. All resource references are currently located in the apis/refs directory for consistency and easier code search. 

Checking the refs we already have `SecretManagerSecretRef` type(https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/refs/v1beta1/secretmanagerrefs.go#L30). Shall we merge these two types? This could potentially lead to confusion in the future.

PR #2898, opened 2024-10-10
tests: small changes for dataformrespository

acpana:
Some small changes to the DataformRepository tests and implementation for Export.


anhdle-sso:
you might need to add the resource to https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/test/resourcefixture/sets.go#L89

acpana:
Also renamed to `basic` now that https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2896 is in

PR #2899, opened 2024-10-10
feat: implement missing exports [cloudbuild]

acpana:
Trying to implement any missing Exports in the direct package.


justinsb:
Looks good.  I know you say "implement any missing exports" but it's much easier to review these one at a time I think.  I'm not sure why the tests aren't failing here though, I would expect to see that we are now generating an exported version?  Maybe we need to add support to the export function in unified_test.go ?

acpana:
Going to take up Justin on having separate PRs for these. This one is for `cloudbuild`. The dynamic (postsubmit) tests should pass now 💯

yuwenma (commented):
Thanks for adding the export function. Could you generate the golden export file so we can validate?

pkg/controller/direct/cloudbuild/workerpool_controller.go
```diff
@@ -142,6 +142,29 @@ func (m *model) AdapterForObject(ctx context.Context, reader client.Reader, u *u
 }
 
 func (m *model) AdapterForURL(ctx context.Context, url string) (directbase.Adapter, error) {
+	// Format: //cloudbuild.googleapis.com/projects/<project>/lcoations/<location>/workerPools/<id>
```
yuwenma, line 145: can we build this from the existing `CloudBuildWorkerPoolIdentity`?

acpana: Chatted offline, we're going to punt on this for a bit as return nil, nil is really important to the registry.Export func

pkg/test/resourcefixture/testdata/basic/cloudbuild/v1beta1/cloudbuildworkerpool/_generated_export_cloudbuildworkerpool.golden
```diff
@@ -0,0 +1,18 @@
...
+    networkConfig:
+      egressOption: NO_PUBLIC_EGRESS
+      peeredNetworkIPRange: /29
+      peeredNetworkRef:
+        external: projects/${projectId}/global/networks/computenetwork-${uniqueId}
+    workerConfig:
+      diskSizeGb: 100
+      machineType: e2-medium
+  projectRef:
+    name: ${projectId}
```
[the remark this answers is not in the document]
acpana: That makes sense! Building this into the generated code then -- https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2941

yuwenma (commented):
/lgtm
/approve

PR #2904, opened 2024-10-11
feat: add a proto comparison to simplify Update field mask operations

yuwenma:
This PR contains several features:
1. Tell users a field is immutable (from proto). This saves a lot of GCP calls.
2. Contributors does not need to manually build the update field-mask by comparing every  mutable field
3. read OUTPUT-ONLY from proto field_behavior extension, so that the comparison function doesn't  compare output-only field.


pkg/controller/direct/secretmanager/secret_controller.go
```diff
@@ -242,34 +242,21 @@ func (a *Adapter) Update(ctx context.Context, updateOp *directbase.UpdateOperati
...
-	}
-	if !common.DeepEqual_StringAndTimestampPb(*desired.Spec.ExpireTime, a.actual.GetExpireTime()) {
-		updateMask.Paths = append(updateMask.Paths, "expire_time")
-	}
-	if !reflect.DeepEqual(desired.Spec.Rotation.NextRotationTime, a.actual.GetRotation().GetNextRotationTime()) {
-		updateMask.Paths = append(updateMask.Paths, "rotation.next_rotation_time")
-	}
-	if !reflect.DeepEqual(desired.Spec.Rotation.RotationPeriod, a.actual.GetRotation().GetRotationPeriod()) {
-		updateMask.Paths = append(updateMask.Paths, "rotation.rotation_period")
+	paths, err := common.CompareProtoMessage(resource, a.actual)
```
jingyih, line 248: We probably need the ability to provide "custom diff" function when comparing proto messages.

yuwenma: Done! Thank you for the suggestion

justinsb (approved):

justinsb:
Looks good.  Specific to Secret I think, but as we extend to more types we will discover the patterns we need (and don't need!)

PR #2908, opened 2024-10-11
Fix sqlinstance periodics

jasonvigil:


pkg/controller/direct/sql/sqlinstance_mappings.go
```diff
@@ -30,12 +30,6 @@ func SQLInstanceKRMToGCP(in *krm.SQLInstance) (*api.DatabaseInstance, error) {
 		return nil, fmt.Errorf("cannot convert nil KRM SQLInstance to GCP DatabaseInstance")
 	}
 
-	if in.Spec.CloneSource != nil {
```
maqiuyujoyce, line 33: Could you share some context why this check is not needed?

pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/sqlinstance-denymaintenanceperiod-direct/_generated_object_sqlinstance-denymaintenanceperiod-direct.golden.yaml
```diff
@@ -18,8 +18,8 @@ spec:
   region: us-central1
   settings:
     denyMaintenancePeriod:
-      endDate: 01-15
-      startDate: 11-01
+      endDate: 0-1-15
```
maqiuyujoyce, line 21: Does it mean there is a breaking change to SQLInstance users?

pkg/controller/direct/sql/sqlinstance_mappings.go
jasonvigil: From the commit message:
> Direct controller needs to be able to update SQLInstance with cloneSource specified.

pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/sqlinstance-denymaintenanceperiod-direct/_generated_object_sqlinstance-denymaintenanceperiod-direct.golden.yaml
jasonvigil: Not breaking, just unnecessary re-update. I can improve the equality check though, so both formats do not cause the controller to re-issue update requests.

yuwenma: Just to share some more thought on this:
I think a reliable check (to determine if update or not) could be to convert the `spec.settings.startDate` from `*string` to proto `timestamp` and compare on the proto level (strong typed).

You may need this to capture different types of valid timestamp value like`2024-01-01` `2024 Jan 1` `Jan, 1 24` (https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2934)

yuwenma:
/approve

Defer to @maqiuyujoyce for the lgtm vote.

pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/sqlinstance-denymaintenanceperiod-direct/_generated_object_sqlinstance-denymaintenanceperiod-direct.golden.yaml
maqiuyujoyce: Agreed with @yuwenma that in general same timestamp in different formats shouldn't cause any diff. Though if the API treats them as different values then we should make a decision based on the specific use case.

For SQLInstance, we should support the existing timestamp format in the direct controller, and the diffing logic should be the same.

Feel free to address this issue in a separate PR.

PR #2909, opened 2024-10-11
chore: Add SQLInstance presubmit test suite

jasonvigil:
There are enough sql tests at this point that we should split them into
a separate suite to reduce overall presubmit time.


jasonvigil:
Seems to save about 20 min on fixtures test suite

yuwenma (commented):
/lgtm
/approve

Great! Thanks!

PR #2910, opened 2024-10-14
feat: dataflowjob export

acpana:
Adding export for the dataflow job. This should make the postsubvmits happy (or -ier).

### DIFF
--- a/pkg/controller/direct/dataflow/dataflowflextemplatejob_controller.go
+++ b/pkg/controller/direct/dataflow/dataflowflextemplatejob_controller.go
@@ -60,8 +60,9 @@ type dataflowFlexTemplateJobAdapter struct {
 	resourceID string
 	jobID      string
 
-	desired *pb.LaunchFlexTemplateParameter
-	actual  *pb.Job
+	desiredObj *krm.DataflowFlexTemplateJob
+	desired    *pb.LaunchFlexTemplateParameter
+	actual     *pb.Job
 
 	flexTemplatesClient *api.FlexTemplatesClient
 	jobsClient          *api.JobsV1Beta3Client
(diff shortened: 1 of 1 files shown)


pkg/controller/direct/dataflow/dataflowflextemplatejob_controller.go
```diff
@@ -306,7 +308,31 @@ func (a *dataflowFlexTemplateJobAdapter) getJob(ctx context.Context, jobID strin
...
 func (a *dataflowFlexTemplateJobAdapter) Export(ctx context.Context) (*unstructured.Unstructured, error) {
-	return nil, nil
+	if a.actual == nil {
+		return nil, fmt.Errorf("Find() not called")
+	}
+	u := &unstructured.Unstructured{}
+
+	obj := &krm.DataflowFlexTemplateJob{}
+	mapCtx := &direct.MapContext{}
+	actual, err := toLaunchParameter(ctx, a.resourceID, a.desiredObj)
```
gemmahou, line 318: Can we use `a.actual` here, instead of calling another method?

gemmahou: Ah I just noticed that they are different type... to avoid confusion, maybe use actualFlexTemplateParameter is clearer?

pkg/controller/direct/dataflow/dataflowflextemplatejob_controller.go
```diff
@@ -306,7 +308,31 @@ func (a *dataflowFlexTemplateJobAdapter) getJob(ctx context.Context, jobID strin
...
+		return nil, mapCtx.Err()
+	}
+
+	obj.Spec.Region = direct.PtrTo(a.location)
+	uObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
+	if err != nil {
+		return nil, err
+	}
+
+	u.Object = uObj
```
gemmahou, line 334: Looking at other Exports, we want to set Name, GVK or Labels as well: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/4c22705c3fb2d93937a659eba94123d44614b7ec/pkg/controller/direct/compute/forwardingrule/forwardingrule_controller.go#L402-L404

Do we want to do the same for this resource?

acpana: Sure! Let's make these part of the tooling too!

PR #2918, opened 2024-10-15
Update operator to 1.124.0

yuwenma:


operator/channels/packages/configconnector/1.124.0/crds.yaml
```diff
@@ -113291,93 +113386,6 @@ spec:
                   gcpIAMAccess:
                     description: Access to a Google Cloud resource through IAM.
                     properties:
-                      folderRef:
```
cheftako, line 113294: Concerned that folderRef is disappearing.

cheftako: Yuwen pointed out its an alpha resource so dropping field is fine.

PR #2927, opened 2024-10-15
operator: support remapping of images

justinsb:
If the IMAGE_PREFIX env var or the --image-prefix flag is provided
to the operator, remap images to pull from that registry/mirror.

For example, with IMAGE_PREFIX=gcr.io/foo, an image like
gcr.io/example/cnrm/manager:latest will be rewritten to
gcr.io/foo/manager:latest.


justinsb:
Verified locally, though we do have to copy prom-to-sd

/hold cancel

PR #2900, opened 2024-10-10
Support Private Service Connect in AlloyDBCluster and AlloyDBInstance

vmiglani:
### [AlloyDB] Private Service Connect Support

This PR adds support for Private Service Connect to KCC.
Internal ticket: https://b.corp.google.com/issues/314366636

Corresponding Terraform PR: https://github.com/GoogleCloudPlatform/magic-modules/pull/10783
TF Beta Changes: https://github.com/hashicorp/terraform-provider-google-beta/pull/7429

### Tests you have done

- [x] Run `make ready-pr` to ensure this PR is ready for review.
- [x] Perform necessary E2E testing for changed resources. Updated pscalloydbcluster/create.yaml, pscalloydbinstance/create.yaml and pscalloydbinstance/update.yaml and ran local tests

PR #2929, opened 2024-10-16
Support pscConfig in mockalloydb cluster

maqiuyujoyce:
### Change description

Added `psc_config` field and corresponding message definition for mockalloydb Cluster in mockgcp/apply-proto-patches.sh, and ran `make gen-proto` to update alloydb proto.

This PR only covers changes related to AlloyDB.

### Tests you have done


yuwenma:
Could you share more information about this change for maintenance purpose? Basically we want to know
1. if the `psc_config`  is a GA field in AlloyDB cluster
2. When the client proto will be updated (if you know), so we can remove this patch.
3. what is the expected behavior of this field?

maqiuyujoyce:
@vmiglani , as you know more about the service, could you answer the questions asked by @yuwenma ?

vmiglani:
Sure Joyce! Thanks @yuwenma 
1. It is in GA
2. I am not sure I have all the context to answer that as of yet. 
3. When enabled i.e. the cluster is created with Private Service Connect which is the next generation Service Networking infrastructure.

nancynh:
On question 2, we need to work with the client libraries team since there are some breaking changes with respect to the PSC fields in our v1alpha/v1beta API. I can work with @vmiglani on that, though I don't know how long this will take. Maybe a week?

maqiuyujoyce:
Thank you, @nancynh ! Will follow up offline to determine when to remove this patch.

pkg/test/resourcefixture/testdata/basic/alloydb/v1beta1/alloydbcluster/fullalloydbcluster/_http.log
yuwenma, line 1: Suggest cover the new `pscConfig` field in the full test suite (assign a value in create.yaml, change in update.yaml)?

yuwenma (commented):
/approve

One comment about improving the test coverage, otherwise looks good!

pkg/test/resourcefixture/testdata/basic/alloydb/v1beta1/alloydbcluster/fullalloydbcluster/_http.log
maqiuyujoyce: `pscConfig` field will be supported in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2900.

@vmiglani could you also cover the newly supported fields in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2900 (including `pscConfig` field) in alloydb test cases that have `full` in their names?

vmiglani: We cannot do that. Because those tests use Service Networking as their architecture. That is the default. 
PSC is a different networking architecture, and both cannot be used together. Also, once selected, this cannot be changed either. So we do not want to touch those tests at all.

vmiglani: Thanks Joyce!

maqiuyujoyce: Thank you for the context, @vmiglani !

@yuwenma looks like the newly added fields will be mutually exclusive with existing fields, so it doesn't seem to be possible to cover them in the full test cases. However in #2900, the new fields are covered by a couple of newly added test cases already.

yuwenma: makes sense.

PR #2930, opened 2024-10-16
doc: 1.124.0 release notes

yuwenma:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/docs/releasenotes/release-1.124.md
+++ b/docs/releasenotes/release-1.124.md
@@ -1,24 +1,23 @@
 # v1.124.0
 
-** This version is not yet released; this document is gathering release notes for the future release **
+* Special shout-outs to @600lyy, @acpana, @anhdle-sso, @benjamin-maynard, @cheftako, @gemmahou, @hankfreund, @jasonvigil, @jingyih, @justinsb, @maqiuyujoyce, @nancynh, @svetakvsundhar, @xiaoweim, @yuwenma for their contributions to this release.
 
-* ...
+## Announcement
 
-* Special shout-outs to ... for their
-  contributions to this release.
-TODO: list contributors with `git log v1.123.0... | grep Merge | grep from | awk '{print $6}' | cut -d '/' -f 1 | sort | uniq`
+* We launched a major improvement to the Config Connector resource development!  Our new approach significantly enhances reliability and provides a more native Kubernetes experience. Learn more in our [guide](https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/master/docs/develop-resources)   
+
+* You can use the `alpha.cnrm.cloud.google.com/reconciler: direct` annotation on `ComputeForwardingRule` resource to opt-in
+the Direct Cloud Reconciler, which provides sustainable reconciliation support for different types of ForwardingRule uses, including setting labels for ServiceAttachments and Private Service Connect.
 
 ## Resources promoted from alpha to beta:
 
-*When resources are promoted from alpha to beta, we (generally) ensure they follow our best practices: use of refs on fields where appropriate,
-output fields from GCP APIs are in `status.observedState.*`
-
-* `PlaceholderKind`
+* `RedisCluster` (v1beta1)
 
 ## Direct Cloud Reconciler:
-* CertificateManagerDNSAuthorization (v1beta1).
-* ComputeForwardingRule (v1beta1).
-* BigQueryAnalyticsHubDataExchange (v1alpha1).
+
+* `BigQueryAnalyticsHubDataExchange` (v1alpha1).
+* `CertificateManagerDNSAuthorization` (v1beta1).
+* `ComputeForwardingRule` (v1beta1).
 
 ## New Resources:
 
@@ -27,6 +26,9 @@ output fields from GCP APIs are in `status.observedState.*`
 ## New Fields:
 
 * CertificateManagerDNSAuthorization
+
   * Added `spec.Location` field.
+
 * ComputeForwardingRule
+
   * Added `spec.target.googleApisBundle` field
--- a/docs/releasenotes/release-1.124.md
+++ b/docs/releasenotes/release-1.124.md
@@ -8,16 +8,9 @@
 
 * We launched a major improvement to the Config Connector resource development!  Our new approach significantly enhances reliability and provides a more native Kubernetes experience. Learn more in our [guide](https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/master/docs/develop-resources)   
 
-
-### Enhanced `ComputeForwardingRule` Reliability (Direct reconciliation)
-
-* We improves the reliability of `ComputeForwardingRule` reconciliation and introduces advanced configuration options.  These enhancements include fixes for defaulting and label issues, and adding the Private Service Connect setup using the new `spec.target.googleApisBundle` field (accept value `all-apis` or `vpc-sc`). To enable these improvements, add the `alpha.cnrm.cloud.google.com/reconciler: direct` annotation to your `ComputeForwardingRule` resources.
-
 ## New Beta Resources (Direct Reconciler):
 
 * `RedisCluster`
-* `CertificateManagerDNSAuthorization` (migrated)
-* `ComputeForwardingRule` (migrated)
 
 ## New Fields:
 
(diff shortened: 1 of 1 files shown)


docs/releasenotes/release-1.124.md
```diff
@@ -1,24 +1,23 @@
...
-* ...
+## Announcement
 
-* Special shout-outs to ... for their
-  contributions to this release.
-TODO: list contributors with `git log v1.123.0... | grep Merge | grep from | awk '{print $6}' | cut -d '/' -f 1 | sort | uniq`
+* We launched a major improvement to the Config Connector resource development!  Our new approach significantly enhances reliability and provides a more native Kubernetes experience. Learn more in our [guide](https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/master/docs/develop-resources)   
 
-## Resources promoted from alpha to beta:
+* You can use the `alpha.cnrm.cloud.google.com/reconciler: direct` annotation on `ComputeForwardingRule` resource to opt-in
```
justinsb, line 9: `provides sustainable reconciliation support` is probably too focused on us, rather than the user.

It sounds like we are introducing functionality only in the direct reconciler?  I don't know if that's a great idea (because we can't indicate that in the CRD schema), but regardless we should focus on the user here.  Something like:

> This functionality requires setting the `alpha.cnrm.cloud.google.com/reconciler: direct` annotation: spec.labels on kind=ServiceAttachment and kind=PrivateServiceConnect

docs/releasenotes/release-1.124.md
```diff
@@ -1,24 +1,23 @@
...
+* `RedisCluster` (v1beta1)
 
 ## Direct Cloud Reconciler:
-* CertificateManagerDNSAuthorization (v1beta1).
-* ComputeForwardingRule (v1beta1).
-* BigQueryAnalyticsHubDataExchange (v1alpha1).
+
+* `BigQueryAnalyticsHubDataExchange` (v1alpha1).
+* `CertificateManagerDNSAuthorization` (v1beta1).
+* `ComputeForwardingRule` (v1beta1).
```
justinsb, line 20: I'm confused - have we moved ComputeForwardingRule to the direct reconciler?

docs/releasenotes/release-1.124.md
```diff
@@ -1,24 +1,23 @@
...
+the Direct Cloud Reconciler, which provides sustainable reconciliation support for different types of ForwardingRule uses, including setting labels for ServiceAttachments and Private Service Connect.
 
-*When resources are promoted from alpha to beta, we (generally) ensure they follow our best practices: use of refs on fields where appropriate,
-output fields from GCP APIs are in `status.observedState.*`
+## Resources promoted from alpha to beta:
 
-* `PlaceholderKind`
+* `RedisCluster` (v1beta1)
 
 ## Direct Cloud Reconciler:
```
justinsb, line 16: Can we rephrase this?  Maybe we should say something more like:

```
# New Resources

* BigQueryAnalyticsHubDataExchange (v1alpha1, uses direct reconciler model)
```

But I don't know where that leaves the other two resources ...

docs/releasenotes/release-1.124.md
yuwenma: It is similiar to CloudSQL: 
the `spec.target.googleApisBundle` is SciFi only controller by the reconcile gate. The CRD still has the tf2crd label so by default it uses tf2crd if not adding the alpha direct annotation.

docs/releasenotes/release-1.124.md
cheftako: I would suggest leading with the result for the customer and then the steps they would need to take.

To use the Direct Cloud Reconciler for `ComputeForwardingRule`, set the `alpha.cnrm.cloud.google.com/reconciler: direct` annotation on `ComputeForwardingRule`. . .

docs/releasenotes/release-1.124.md
cheftako: Can we have separate lists for alpha and beta. 

New Enabled By Default Resources:
* `CertificateManagerDNSAuthorization` (uses direct reconciler model)

New Alpha Resources:
* BigQueryAnalyticsHubDataExchange (uses direct reconciler model)

docs/releasenotes/release-1.124.md
```diff
@@ -1,24 +1,23 @@
...
-TODO: list contributors with `git log v1.123.0... | grep Merge | grep from | awk '{print $6}' | cut -d '/' -f 1 | sort | uniq`
+* We launched a major improvement to the Config Connector resource development!  Our new approach significantly enhances reliability and provides a more native Kubernetes experience. Learn more in our [guide](https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/master/docs/develop-resources)   
 
-## Resources promoted from alpha to beta:
+* You can use the `alpha.cnrm.cloud.google.com/reconciler: direct` annotation on `ComputeForwardingRule` resource to opt-in
+the Direct Cloud Reconciler, which provides sustainable reconciliation support for different types of ForwardingRule uses, including setting labels for ServiceAttachments and Private Service Connect.
 
-*When resources are promoted from alpha to beta, we (generally) ensure they follow our best practices: use of refs on fields where appropriate,
-output fields from GCP APIs are in `status.observedState.*`
+## Resources promoted from alpha to beta:
```
cheftako, line 12: Not sure customers care about the difference between alpha to beta and new in beta. The distinction they are about is Enabled (Beta however it got there) and Experimental (Alpha/test/not really usable).

docs/releasenotes/release-1.124.md
yuwenma: If users do not care Alpha resource, how about drop the `Direct Cloud Reconciler` section?

docs/releasenotes/release-1.124.md
yuwenma: This is really helpful! Thank you. I restructured all the release note format to make it more user centric. Let me know  how it looks now

docs/releasenotes/release-1.124.md
yuwenma: I restructured the release sections. Let me know how it looks now.

docs/releasenotes/release-1.124.md
```diff
@@ -1,32 +1,34 @@
...
+  * Added `spec.Location` field.
 
-## New Fields:
+* `ComputeForwardingRule`
 
-* CertificateManagerDNSAuthorization
-  * Added `spec.Location` field.
-* ComputeForwardingRule
-  * Added `spec.target.googleApisBundle` field
+  * Added `spec.target.googleApisBundle` field (accept value `all-apis` or `vpc-sc`). Note, when configured this field, you are using the new Direct reconciliation.
```
justinsb, line 23: Nit: should be "(allowed values are `all-apis` or `vpc-sc`)", but I think we just drop the list of values.

docs/releasenotes/release-1.124.md
```diff
@@ -1,32 +1,34 @@
...
-## New Fields:
+* `ComputeForwardingRule`
 
-* CertificateManagerDNSAuthorization
-  * Added `spec.Location` field.
-* ComputeForwardingRule
-  * Added `spec.target.googleApisBundle` field
+  * Added `spec.target.googleApisBundle` field (accept value `all-apis` or `vpc-sc`). Note, when configured this field, you are using the new Direct reconciliation.  
+
+## Modified Beta Reconciliation
```
justinsb, line 25: Nit: "Resources moved to direct reconciliation" might be clearer

justinsb:
Some nits but LGTM

Will approve & hold in case you want to tweak

/approve
/lgtm
/hold

docs/releasenotes/release-1.124.md
yuwenma: Yeah... because it should be in the reference doc (API comment), but I think we missed it, and this field does not directly map to a GCP object field (several fields share the `target` in the GCP object see [doc](https://cloud.google.com/compute/docs/reference/rest/v1/forwardingRules), so I "temporarily" add it in this release note before we can fix the doc in the next release.

docs/releasenotes/release-1.124.md
yuwenma: Fixed. thanks

yuwenma:
@justinsb I resolved the comments in the actual release note. Thank you and Walter for all the suggestions!

PR #2934, opened 2024-10-17
chore: extend supported timestamp parsing layout

yuwenma:
### Change description

Fixes #

### Tests you have done


justinsb:
I think the same comment as before.  Do we actually support all these formats today?  Do we have a bug that we are trying to fix?  A good test could be "if we merged this PR, and someone proposed rolling it back, why would we _not_ roll it back".

yuwenma:
> I think the same comment as before.

This is a code refactoring. The same code is used [here](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/common/normalize.go#L23C6-L23C32). But since we have the `common.CompareProtoMessage` function, we don't need the individual timestamp function any more. 

> Do we actually support all these formats today? 

We do: the proto does, so the GCP server supports these formats. When KCC accepts the timestamp value as `*string`, and pass it to GCP server, everything just works well and the GCP server tells whether the format is valid or not. But what we are trying to do in proto comparison is that "Jan 1, 2024" should be the same as "2024-01-01", and we do not need to send the UPDATE request to GCP server.  
 
> Do we have a bug that we are trying to fix? A good test could be "if we merged this PR, and someone proposed rolling it back, why would we _not_ roll it back".

This list is the standard google proto timestamp allowed layout. If not supporting this, the mapper function could fail in runtime when converting a string, say in the RFC850 format, to timestmap.

pkg/test/resourcefixture/testdata/basic/secretmanager/v1beta1/basicsecretmanagersecret/update.yaml
```diff
@@ -0,0 +1,28 @@
...
+    cnrm.cloud.google.com/project-id: ${projectId}
+  name: secretmanagersecret-${uniqueId}
+  labels:
+    label-one: value-one
+spec:
+  replication:
+    userManaged:
+      replicas:
+      - location: us-central1
+  expireTime: "Monday, 02-Jan-06 15:04:05 MST"
```
yuwenma, line 28: same timestamp in different layout form.

ldanielmadariaga:

PR #2939, opened 2024-10-17
feat: Add types and mappers for compute target TCP proxy

gemmahou:
### Change description

Prerequisite for b/246759060
Migrate compute target TCP proxy to direct controller, and add support for regional resource. So we can reference to it in compute forwarding rule.

### Tests you have done


apis/compute/v1alpha1/computeregiontargettcpproxy_types.go
```diff
@@ -0,0 +1,122 @@
...
+
+package v1alpha1
+
+import (
+	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/apis/k8s/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var ComputeRegionTargetTCPProxyGVK = GroupVersion.WithKind("ComputeRegionTargetTCPProxy")
```
yuwenma, line 23: Just to confirm, do you plan to add the global  targetTCPProxy as a different kind?

apis/compute/v1alpha1/computeregiontargettcpproxy_types.go
```diff
@@ -0,0 +1,122 @@
...
+
+// ComputeRegionTargetTCPProxySpec defines the desired state of ComputeRegionTargetTCPProxy
+// +kcc:proto=google.cloud.compute.v1.TargetTcpProxy
+type ComputeRegionTargetTCPProxySpec struct {
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="BackendServiceRef is immutable"
```
yuwenma, line 28: Maybe not check the immutable on the CRD level (or even not check the immutable here for simplicity), because users can switch between `.external` and `.name` if they refer to the same `backendService`.  https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/develop-resources/api-conventions/resource-reference.md#rule-2-form-switch-allowed

apis/compute/v1alpha1/computeregiontargettcpproxy_types.go
```diff
@@ -0,0 +1,122 @@
...
+	BackendServiceRef *refs.ComputeBackendServiceRef `json:"backendServiceRef"`
+
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Description is immutable"
+	// Immutable. An optional description of this resource.
+	Description *string `json:"description,omitempty"`
+
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Region is immutable"
+	// Immutable. The geographical location of the ComputeTargetTCPProxy.
+	// Reference: GCP definition of regions/zones (https://cloud.google.com/compute/docs/regions-zones/)
+	Region *string `json:"region"`
```
yuwenma, line 40: nit: required field ?

apis/compute/v1alpha1/computeregiontargettcpproxy_types.go
```diff
@@ -0,0 +1,122 @@
...
+// +kcc:proto=google.cloud.compute.v1.TargetTcpProxy
+type ComputeRegionTargetTCPProxySpec struct {
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="BackendServiceRef is immutable"
+	// Immutable. A reference to the ComputeBackendService resource.
+	// +required
+	BackendServiceRef *refs.ComputeBackendServiceRef `json:"backendServiceRef"`
+
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Description is immutable"
+	// Immutable. An optional description of this resource.
+	Description *string `json:"description,omitempty"`
```
yuwenma, line 35: oh, it's interesting that the description is immutable.

apis/compute/v1alpha1/computeregiontargettcpproxy_types.go
```diff
@@ -0,0 +1,122 @@
...
+	// this target proxy has a loadBalancingScheme set to INTERNAL_SELF_MANAGED.
+	ProxyBind *bool `json:"proxyBind,omitempty"`
+
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="ProxyHeader is immutable"
+	// Immutable. Specifies the type of proxy header to append before sending data to
+	// the backend. Default value: "NONE" Possible values: ["NONE", "PROXY_V1"].
+	ProxyHeader *string `json:"proxyHeader,omitempty"`
+
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="ResourceID field is immutable"
+	// Immutable.
```
yuwenma, line 53: suggest not checking the immutability. User can specify `spec.resourceID` if it refers to the same obj. But I realized it is too trivial to capture all these conditions, and it further confuses the users.

apis/compute/v1alpha1/computeregiontargettcpproxy_types.go
```diff
@@ -0,0 +1,122 @@
...
+	CreationTimestamp *string `json:"creationTimestamp,omitempty"`
+
+	// Type of the resource. Always compute#targetTcpProxy for target TCP proxies.
+	Kind *string `json:"kind,omitempty"`
+
+	// The unique identifier for the resource.
+	ProxyID *int64 `json:"proxyID,omitempty"`
+
+	// The SelfLink for the resource.
+	SelfLink *string `json:"selfLink,omitempty"`
```
yuwenma, line 83: since this is an Alpha resource, I assume we don't want (at least not need) these fields now?

apis/compute/v1alpha1/computeregiontargettcpproxy_types.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// ComputeRegionTargetTCPProxyObservedState is the state of the ComputeRegionTargetTCPProxy resource as most recently observed in GCP.
+type ComputeRegionTargetTCPProxyObservedState struct {
+}
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// TODO(user): make sure the pluralizaiton below is correct
```
yuwenma, line 92: nit: clean up

apis/compute/v1alpha1/computeregiontargettcpproxy_types.go
```diff
@@ -0,0 +1,122 @@
...
+// +kubebuilder:resource:categories=gcp,shortName=gcpcomputeregiontargettcpproxy;gcpcomputeregiontargettcpproxys
```
yuwenma, line 93: nit: plural `proxies`

dev/tools/controllerbuilder/template/apis/refs.go
```diff
@@ -88,15 +88,11 @@ func (r *{{.Kind}}Ref) NormalizedExternal(ctx context.Context, reader client.Rea
...
+func New{{.Kind}}Ref(ctx context.Context, reader client.Reader, obj *{{.Kind}}, u *unstructured.Unstructured) (*{{.Kind}}Ref, error) {
 	id := &{{.Kind}}Ref{}
 
 	// Get Parent
-	projectRef, err := refsv1beta1.ResolveProject(ctx, reader, obj, obj.Spec.ProjectRef)
-	if err != nil {
-		return nil, err
-	}
-	projectID := projectRef.ProjectID
+	projectID, err := refsv1beta1.ResolveProjectID(ctx, reader, u)
```
yuwenma, line 95: maybe not changing the template. I want to modify the ResolveProject the other way around.

pkg/controller/direct/compute/regiontargettcpproxy/mapper.go
yuwenma, line 1: qq: any reasons to change the import package struct? 
I would expect this to be `package compute`, so that we can the service as a whole.

apis/compute/v1alpha1/computeregiontargettcpproxy_types.go
gemmahou: global is already a [TF-based resource](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/crds/compute_v1beta1_computetargettcpproxy.yaml#L25), I want to take this chance to support regional and split it from global.
We also have some compute regional resources(alpha): https://github.com/search?q=repo%3AGoogleCloudPlatform%2Fk8s-config-connector%20compute_v1alpha1_computeregion&type=code

apis/compute/v1alpha1/computeregiontargettcpproxy_types.go
gemmahou: Actually regional TCP proxy is immutable, there's no "update" function. https://cloud.google.com/compute/docs/reference/rest/v1/regionTargetTcpProxies

apis/compute/v1alpha1/computeregiontargettcpproxy_types.go
gemmahou: They are all output only fields, I'll move them to status.observedState.

yuwenma:
Per offline discussion, we will mgirate the global compute TCP Proxy to SCiFi and extend it to support regional resource.

apis/compute/v1beta1/computetargettcpproxy_types.go
```diff
@@ -0,0 +1,118 @@
...
+	BackendServiceRef *refs.ComputeBackendServiceRef `json:"backendServiceRef"`
+
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Description is immutable"
+	// Immutable. An optional description of this resource.
+	Description *string `json:"description,omitempty"`
+
+	// The geographical location of the ComputeTargetTCPProxy.
+	// Reference: GCP definition of regions/zones (https://cloud.google.com/compute/docs/regions-zones/)
+	// +optional
+	Location *string `json:"location"`
```
gemmahou, line 46: Have to explicitly add optional tag otherwise location will be marked as required field in the CRD, while it's not required for existing global tcp proxy, this change will break existing global tcp proxy resources.

I'm not familiar with how the CRD is generated and why location is marked as required by default. I noticed that in the previous generated types, we specify +optional but not +required([example](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/8dabfc844d7bb35d243216047f140f5d7e770441/pkg/clients/generated/apis/compute/v1beta1/computetargettcpproxy_types.go#L43)), is there any reason we switch to only specify +required not +optional now?

apis/compute/v1alpha1/computeregiontargettcpproxy_types.go
gemmahou: ResourceID is an immutable field in the existing types: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/6e9579347aadd08c07cfb0f1bd1747c4c9f4b197/pkg/clients/generated/apis/compute/v1beta1/computetargettcpproxy_types.go#L56. During direct controller migration I prefer to not change this behavior.

yuwenma (commented):
/lgtm
/approve

PR #2941, opened 2024-10-17
dev:tools: add export changes

acpana:
Small changes to add learnings from the `Export` PRs I have been working on.

### DIFF
--- a/dev/tools/controllerbuilder/template/controller/controller.go
+++ b/dev/tools/controllerbuilder/template/controller/controller.go
@@ -254,17 +254,21 @@ func (a *Adapter) Export(ctx context.Context) (*unstructured.Unstructured, error
 	if mapCtx.Err() != nil {
 		return nil, mapCtx.Err()
 	}
-	// TODO(user): Update other resource reference 
+	// TODO(user): Update other resource references 
 	parent, err := a.id.Parent()
 	if err != nil {
 		return nil, err
 	}
-	obj.Spec.ProjectRef = &refs.ProjectRef{Name: parent.ProjectID}
+	obj.Spec.ProjectRef = &refs.ProjectRef{External: parent.String()}
 	obj.Spec.Location = parent.Location
 	uObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
 	if err != nil {
 		return nil, err
 	}
+
+	u.SetName(a.actual.Id)
+	u.SetGroupVersionKind(krm.{{.Kind}}GVK)
+
 	u.Object = uObj
 	return u, nil
 }


dev/tools/controllerbuilder/template/controller/controller.go
```diff
@@ -254,17 +254,22 @@ func (a *Adapter) Export(ctx context.Context) (*unstructured.Unstructured, error
...
 	}
-	obj.Spec.ProjectRef = &refs.ProjectRef{Name: parent.ProjectID}
+	obj.Spec.ProjectRef = &refs.ProjectRef{External: parent.String()}
 	obj.Spec.Location = parent.Location
 	uObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
 	if err != nil {
 		return nil, err
 	}
+
+	u.SetName(a.actual.Id)
```
maqiuyujoyce, line 269: Maybe add a TODO comment about "Remove underscore and other chars that not allowed in the K8s naming convention."

yuwenma (commented):
/hold

This PR touch several topics that we need to provide a more concrete solution for. I suggest we do not update the template now to avoid raising signals to the contributors that we will change later on.

@acpana what is the problem you hit ?

dev/tools/controllerbuilder/template/controller/controller.go
acpana: let's add an issue to track this instead 🙏🏼

dev/tools/controllerbuilder/template/controller/controller.go
```diff
@@ -254,17 +254,22 @@ func (a *Adapter) Export(ctx context.Context) (*unstructured.Unstructured, error
...
 		return nil, mapCtx.Err()
 	}
-	// TODO(user): Update other resource reference 
+	// TODO(user): Update other resource references 
 	parent, err := a.id.Parent()
 	if err != nil {
 		return nil, err
 	}
-	obj.Spec.ProjectRef = &refs.ProjectRef{Name: parent.ProjectID}
+	obj.Spec.ProjectRef = &refs.ProjectRef{External: parent.String()}
```
acpana, line 262: @yuwenma regarding your question:

> what is the problem you hit ?

Some of these changes like setting external were review comments. Let's have the tooling do the right thing here.

Setting the name and gvk is the right thing to do for exporting.

Should I take out the labels for now?

jasonvigil: I think it was necessary to set name and GVK for marshaling SQLInstance when exporting: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/cae8025dee7e762f431bdb9a44946551b3edbc31/pkg/controller/direct/sql/sqlinstance_controller.go#L473-L474. It is probably also the right thing to set labels, though we don't have any tests for it yet: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/cae8025dee7e762f431bdb9a44946551b3edbc31/pkg/test/resourcefixture/contexts/register.go#L133-L136

yuwenma: oh, didn't realize this is for Export. That's fair

dev/tools/controllerbuilder/template/controller/controller.go
```diff
@@ -254,17 +254,22 @@ func (a *Adapter) Export(ctx context.Context) (*unstructured.Unstructured, error
...
 	obj.Spec.Location = parent.Location
 	uObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
 	if err != nil {
 		return nil, err
 	}
+
+	u.SetName(a.actual.Id)
+	u.SetGroupVersionKind(krm.{{.Kind}}GVK)
+	u.SetLabels(a.actual.Labels) // TODO(user): Only keep if the proto resource has a Labels field
```
yuwenma, line 271: I'm not sure about this. For labels, I think we may want to
1. Remove kcc its own labels (as sanity check)
2. if the `.spec` has a label field, assign to that field.
3. Otherwise, assign to the metadata label.

And we have another question: Shall we check if the CRD is `tf2crd` or `dcl2crd` and add the `alpha.cnrm.cloud.google.com/reconciler: direct` annotation? That depends on how the Export is used (for debugging or re-building the KCC object purpose seems like this is the right approach). What if the resource has reconcile gate? I think we may want to consider more about the Export behavior before putting that in the template, because giving a divergent solution is harder to fix than waiting until the customers really needs the `Export`.

jasonvigil: Maybe we can just skip labels for now, wdyt @yuwenma ? In any case, I think we should not add `alpha.cnrm.cloud.google.com/reconciler: direct` annotation, and instead make the default behavior to enable the direct reconciler, with an optional label to disable it, for specific resources that require care to migrate.

yuwenma (commented):
/lgtm
/approve
/hold cancel

PR #2942, opened 2024-10-17
Refactor SQLInstance defaulting

jasonvigil:
Separate defaulting from reference resolving. Create a function for filling in GCP defaults, to consolidate API defaulting logic.

### DIFF
--- a/pkg/controller/direct/sql/sqlinstance_defaults.go
+++ b/pkg/controller/direct/sql/sqlinstance_defaults.go
@@ -21,11 +21,17 @@ import (
 )
 
 func ApplySQLInstanceGCPDefaults(in *krm.SQLInstance, out *api.DatabaseInstance) {
+	if in.Spec.InstanceType == nil {
+		out.InstanceType = "CLOUD_SQL_INSTANCE"
+	}
 	if in.Spec.Settings.ActivationPolicy == nil {
 		out.Settings.ActivationPolicy = "ALWAYS"
 	}
 	if in.Spec.Settings.AvailabilityType == nil {
 		out.Settings.AvailabilityType = "ZONAL"
+	}
+	if in.Spec.Settings.ConnectorEnforcement == nil {
+		out.Settings.ConnectorEnforcement = "NOT_REQUIRED"
 	}
 	if in.Spec.Settings.DiskType == nil {
 		out.Settings.DataDiskType = "PD_SSD"
--- a/pkg/controller/direct/sql/sqlinstance_defaults.go
+++ b/pkg/controller/direct/sql/sqlinstance_defaults.go
@@ -1,0 +1,42 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sql
+
+import (
+	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/sql/v1beta1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
+	api "google.golang.org/api/sqladmin/v1beta4"
+)
+
+func ApplySQLInstanceGCPDefaults(in *krm.SQLInstance, out *api.DatabaseInstance) {
+	if in.Spec.Settings.ActivationPolicy == nil {
+		out.Settings.ActivationPolicy = "ALWAYS"
+	}
+	if in.Spec.Settings.AvailabilityType == nil {
+		out.Settings.AvailabilityType = "ZONAL"
+	}
+	if in.Spec.Settings.DiskType == nil {
+		out.Settings.DataDiskType = "PD_SSD"
+	}
+	if in.Spec.Settings.Edition == nil {
+		out.Settings.Edition = "ENTERPRISE"
+	}
+	if in.Spec.Settings.PricingPlan == nil {
+		out.Settings.PricingPlan = "PER_USE"
+	}
+	if in.Spec.Settings.DiskAutoresize == nil {
+		out.Settings.StorageAutoResize = direct.PtrTo(true)
+	}
+}
--- a/pkg/controller/direct/sql/sqlinstance_controller.go
+++ b/pkg/controller/direct/sql/sqlinstance_controller.go
@@ -108,7 +108,7 @@ func (m *sqlInstanceModel) AdapterForObject(ctx context.Context, kube client.Rea
 		return nil, fmt.Errorf("building gcp client: %w", err)
 	}
 
-	if err := NormalizeSQLInstance(ctx, kube, obj); err != nil {
+	if err := ResolveSQLInstanceRefs(ctx, kube, obj); err != nil {
 		return nil, err
 	}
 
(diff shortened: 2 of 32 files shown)


pkg/controller/direct/sql/sqlinstance_defaults.go
```diff
@@ -0,0 +1,48 @@
...
+
+package sql
+
+import (
+	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/sql/v1beta1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
+	api "google.golang.org/api/sqladmin/v1beta4"
+)
+
+func ApplySQLInstanceGCPDefaults(in *krm.SQLInstance, out *api.DatabaseInstance) {
```
yuwenma, line 23: I have some general questions about the backward compatibility of Cloud SQL and the sustainability of moving to the SCifi approach. 
- What happens if we don't set this defaulting value? Will SCiFi break the existing users?  I assume existing users (sts-merge) will have the default values updating their spec already. So those values won't be empty. 

- Some field seems to be an enum. Will the GCP assign some default value as well (maybe a different one like the proto `UNIMPLEMENTED_XXX`) ? If so, it seems to be more instructive to expose the GCP defaulting value to the users, so that users can better triage since no hidden step in the transient.  

-  If we do not set the default value, will the new SQL instance has any problems (assume the existing users are fine)? I would expect the GCP server gives users the instruction if an expected value is unset. so that we don't need to maintain and keep a version of the GCP server defaulting. 

Basically I'm trying to see if these defaulting problems come from the state-into-spec:merge and if has been resolved after we set the absent as the default.

jasonvigil: > What happens if we don't set this defaulting value? Will SCiFi break the existing users? I assume existing users (sts-merge) will have the default values updating their spec already. So those values won't be empty.

It is not necessarily true that existing users will have default values applied in their spec already (from sts-merge). Now that we have made the default sts absent, we have several releases where users may create SQLInstances where the state will not be merged into the spec.

> Some field seems to be an enum. Will the GCP assign some default value as well (maybe a different one like the proto UNIMPLEMENTED_XXX) ? If so, it seems to be more instructive to expose the GCP defaulting value to the users, so that users can better triage since no hidden step in the transient.

> If we do not set the default value, will the new SQL instance has any problems (assume the existing users are fine)? I would expect the GCP server gives users the instruction if an expected value is unset. so that we don't need to maintain and keep a version of the GCP server defaulting.

There will be issues if we do not add logic for GCP-applied default values in KCC. Here is an example:

Create minimal postgres instance:
```
{
  "name": "test-instance-postgres",
  "databaseVersion": "POSTGRES_16",
  "settings": {
    "tier": "db-perf-optimized-N-2"
  }
}
```

Here is the created instance (with all GCP API applied default values, and outputs included):
```
{
  "kind": "sql#instance",
  "state": "RUNNABLE",
  "databaseVersion": "POSTGRES_16",
  "settings": {
    "authorizedGaeApplications": [],
    "tier": "db-perf-optimized-N-2",
    "kind": "sql#settings",
    "availabilityType": "ZONAL",
    "pricingPlan": "PER_USE",
    "replicationType": "SYNCHRONOUS",
    "activationPolicy": "ALWAYS",
    "ipConfiguration": {
      "authorizedNetworks": [],
      "sslMode": "ALLOW_UNENCRYPTED_AND_ENCRYPTED",
      "ipv4Enabled": true,
      "requireSsl": false,
      "serverCaMode": "GOOGLE_MANAGED_INTERNAL_CA"
    },
    "locationPreference": {
      "zone": "us-central1-f",
      "kind": "sql#locationPreference"
    },
    "dataDiskType": "PD_SSD",
    "backupConfiguration": {
      "startTime": "22:00",
      "kind": "sql#backupConfiguration",
      "backupRetentionSettings": {
        "retentionUnit": "COUNT",
        "retainedBackups": 15
      },
      "enabled": true,
      "replicationLogArchivingEnabled": true,
      "pointInTimeRecoveryEnabled": true,
      "transactionLogRetentionDays": 14,
      "transactionalLogStorageState": "CLOUD_STORAGE"
    },
    "edition": "ENTERPRISE_PLUS",
    "connectorEnforcement": "NOT_REQUIRED",
    "dataCacheConfig": {
      "dataCacheEnabled": true
    },
    "settingsVersion": "1",
    "storageAutoResizeLimit": "0",
    "storageAutoResize": true,
    "dataDiskSizeGb": "10",
    "deletionProtectionEnabled": false
  },
  "etag": "1d7fb28918c9f8529bd86a82e2d3a1bd729031bd5b21e23184be683587b74a50",
  "ipAddresses": [
    {
      "type": "PRIMARY",
      "ipAddress": "<redacted>"
    },
    {
      "type": "OUTGOING",
      "ipAddress": "<redacted>"
    }
  ],
  "serverCaCert": {
    "kind": "sql#sslCert",
    "certSerialNumber": "0",
    "cert": "<redacted>",
    "commonName": "<redacted>",
    "sha1Fingerprint": "<redacted>",
    "instance": "test-instance-postgres",
    "createTime": "2024-10-18T21:41:52.804Z",
    "expirationTime": "2034-10-16T21:42:52.804Z"
  },
  "instanceType": "CLOUD_SQL_INSTANCE",
  "project": "<redacted>",
  "serviceAccountEmailAddress": "<redacted>",
  "backendType": "SECOND_GEN",
  "selfLink": "https://content-sqladmin.googleapis.com/sql/v1beta4/projects/my-project/instances/test-instance-postgres",
  "connectionName": "my-project:us-central1:test-instance-postgres",
  "name": "test-instance-postgres",
  "region": "us-central1",
  "gceZone": "us-central1-f",
  "databaseInstalledVersion": "POSTGRES_16_4",
  "maintenanceVersion": "POSTGRES_16_4.R20240910.01_04",
  "geminiConfig": {
    "entitled": false,
    "googleVacuumMgmtEnabled": false,
    "oomSessionCancelEnabled": false,
    "activeQueryEnabled": false,
    "indexAdvisorEnabled": false
  },
  "createTime": "2024-10-18T21:42:01.122Z",
  "sqlNetworkArchitecture": "NEW_NETWORK_ARCHITECTURE",
  "satisfiesPzi": false
}
```

On next resync, the controller will detect a diff with desired state. There are many differences between desired and actual state that cause the controller to interpret a diff. One example is: the desired state specifies `"availabilityType": nil`, but GCP API has applied a default value of `"availabilityType": "ZONAL"`

Therefore, on resync, the controller will try to send an update with the desired state:
```
{
  "name": "test-instance-postgres",
  "databaseVersion": "POSTGRES_16",
  "settings": {
    "tier": "db-perf-optimized-N-2",
    "settingsVersion": 2
  }
}
```

And, the API returns an error:
```
Invalid request: Invalid Tier (db-perf-optimized-N-2) for (EDITION_UNSPECIFIED) Edition.
```

I think this happens because the API only applies the default value for edition: `"edition": "ENTERPRISE_PLUS"` on create, not on update.

The end result is that on resync, postgres instances that do not specify an edition will enter an error state.

yuwenma: For the `"edition": "ENTERPRISE_PLUS" case, the Cloud SQL server does the right thing that its
CREATE call allows empty field and returns the defaulting value to give a full sqlinstance payload, and then request the UPDATE to give full payload. 

In such case, shall we record those GCP defaulting value in `status.observedState` and merge into the object field if not specified in the UPDATE call? In such case, KCC does not need to keep track of the actual default value.

jasonvigil: I do not think that is a good behavior. That would men we are basically unmanaging the edition. My opinion is that we should manage all the SQLInstance fields, and assume the user desired state matches the GCP-assigned defaults.

yuwenma: I'd like to further discuss the statement that "this means we are bascially unmanaging the edition".  I don't think we are unmanaging the field, but rather reflecting the user expectations in the form of assigning default without interrupting users intends.

But this is more of a generic discussion for an implementation detail, and this does not affect the cloudSQL API.  I'll let this PR goes in to unblock the resource.

yuwenma (commented):
/lgtm
/approve

PR #2949, opened 2024-10-18
feat:refactor: kind aware cmd add

acpana:
This work comes back to look at the controller builder cmd.

* it reverts the per package work as we're moving away from that direction to having a structure like:
```
resourceA_controller.go
resourceB_controller.go
```
* For any datastructure or function that needs a resource specific name, I added the necessary changes in the template.

Been testing this with the `BigQueryAnalyticsHubDataExchange` and `BigQueryAnalyticsHubListing`

### DIFF
--- a/dev/tools/controllerbuilder/template/controller/controller.go
+++ b/dev/tools/controllerbuilder/template/controller/controller.go
@@ -78,20 +78,20 @@ const (
 )
 
 func init() {
-	registry.RegisterModel(krm.{{.Kind}}GVK, NewModel)
+	registry.RegisterModel(krm.{{.Kind}}GVK, New{{.Kind}}Model)
 }
 
-func NewModel(ctx context.Context, config *config.ControllerConfig) (directbase.Model, error) {
-	return &model{config: *config}, nil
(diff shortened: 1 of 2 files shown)


dev/tools/controllerbuilder/template/controller/controller.go
```diff
@@ -78,20 +78,20 @@ const (
...
 
-var _ directbase.Model = &model{}
+var _ directbase.Model = &model{{.Kind}}{}
 
-type model struct {
+type model{{.Kind}} struct {
 	config config.ControllerConfig
 }
 
-func (m *model) client(ctx context.Context) (*gcp.Client, error) {
```
jingyih, line 94: nit: `clientForKind` seems a bit verbose? Given that the function already has a pointer receiver of `modelKind`.

jingyih (commented):
/lgtm

jingyih (approved):
/lgtm
/approve

/hold
Feel free to remove when PR is ready to merge.

yuwenma:
Is this for using the mapping?

PR #2954, opened 2024-10-19
controllerbuilder: don't add log name

justinsb:
It should be set in the context already.


yuwenma (commented):
/lgtm
/approve

Thanks

PR #2955, opened 2024-10-19
chore: refactor SecureSourceManagerInstance to work in proto

justinsb:
We can keep the desired state in proto.


yuwenma (commented):
/lgtm
/approve

Sounds good!

PR #2956, opened 2024-10-19
refactor: move reference resolver into shared package

justinsb:


yuwenma (commented):
/lgtm
/approve
/hold

feel like we can either merge these two files to a `visitreferences.go` or a sub package `reference` for easy discoverable purpose. What do you think? Feel free to unhold.

justinsb:
The reason I kept them separate is because I think we'll want the "walk a proto" functionality in other places also.  e.g. implementing update_mask

justinsb:
/assign @yuwenma 

Can we move this forwards?  Would be good to standardize reference resolving

pkg/controller/direct/common/refs.go
yuwenma, line 1: Do you want to move the MonitoringAlertPolicy specific functions out of this file?

yuwenma (commented):
/approve
/lgtm

/hold 

One suggestion. Not a blocker. Thank you for this!

pkg/controller/direct/common/refs.go
justinsb: I think not yet ... I read a tweet the other day: https://x.com/KentBeck/status/250733358307500032?lang=en

So this is making the change easy, but not yet making the easy change.  The easy changes are to unify all the resolvers, including picking up the Reference interface :-)

justinsb:
Ah, but I've lost LGTM in a rebase anyway :-)

yuwenma:
/lgtm


> Ah, but I've lost LGTM in a rebase anyway :-)

No, you got it in seconds.

PR #2962, opened 2024-10-21
tests: fix env var to be ONLY_TEST_APIGROUPS

justinsb:
It was using ONLY_TEST_APIGROUP, which is the old form.


yuwenma (commented):
/lgtm
/approve

PR #2963, opened 2024-10-21
chore: cleanup a few lint problems

justinsb:
Some small non-functional changes to avoid the IDE giving us false-positive problem warnings.


yuwenma (commented):
/lgtm
/approve

Thanks!

PR #2966, opened 2024-10-21
Add field mapper for attached clusters direct controller migration.

hankfreund:
Followed step 3 of the Direct Controller migration guide.
Manually implemented functions which the generator was unable to handle on its own.
Since this is a migration for an existing resource, some `NOTYET`s are in the code for new, unimplemented fields.

Fix the change that made the required `ProjectRef` optional in the previous PR.

### DIFF
--- a/apis/containerattached/v1beta1/containerattachedcluster_types.go
+++ b/apis/containerattached/v1beta1/containerattachedcluster_types.go
@@ -25,9 +25,8 @@ var ContainerAttachedClusterGVK = GroupVersion.WithKind("ContainerAttachedCluste
 // ContainerAttachedClusterSpec defines the desired state of ContainerAttachedCluster
 // +kcc:proto=google.cloud.gkemulticloud.v1.AttachedCluster
 type ContainerAttachedClusterSpec struct {
-	/* The ID of the project in which the resource belongs. If it is not provided, the provider project is used. */
-	// +optional
-	ProjectRef *refs.ProjectRef `json:"projectRef,omitempty"`
+	/* The ID of the project in which the resource belongs.*/
+	ProjectRef *refs.ProjectRef `json:"projectRef"`
 
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="ResourceID field is immutable"
 	// Immutable, Optional.
--- a/pkg/controller/direct/containerattached/containerattachedcluster_mappings.go
+++ b/pkg/controller/direct/containerattached/containerattachedcluster_mappings.go
@@ -1,0 +1,189 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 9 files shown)


pkg/controller/direct/containerattached/containerattachedcluster_mappings.go
```diff
@@ -0,0 +1,189 @@
...
+		return nil
+	}
+	out := &krm.LoggingComponentConfig{}
+	out.EnableComponents = slice_FromProto_two(mapCtx, in.EnableComponents, func(mapCtx *direct.MapContext, p pb.LoggingComponentConfig_Component) *string {
+		return direct.Enum_FromProto(mapCtx, p)
+	})
+	return out
+}
+
+func slice_FromProto_two[T, U any](mapCtx *direct.MapContext, in []T, mapper func(mapCtx *direct.MapContext, in T) *U) []U {
```
acpana, line 153: Is this because the current Slice_FromProto has a pointer in the mapping function? If so, I think it would be worth adding this in the `direct` package to be reused. That can happen as a follow on PR. Let's rename this to something like `slice_FromProto` for now if so.

apis/containerattached/v1beta1/containerattachedcluster_types.go
```diff
@@ -26,8 +26,7 @@ var ContainerAttachedClusterGVK = GroupVersion.WithKind("ContainerAttachedCluste
 // +kcc:proto=google.cloud.gkemulticloud.v1.AttachedCluster
 type ContainerAttachedClusterSpec struct {
 	/* The ID of the project in which the resource belongs. If it is not provided, the provider project is used. */
```
acpana, line 28: I think by making this a required field now we should update the docs:
```suggestion
	/* The ID of the project in which the resource belongs. */
```

apis/containerattached/v1beta1/containerattachedcluster_types.go
```diff
@@ -26,8 +26,7 @@ var ContainerAttachedClusterGVK = GroupVersion.WithKind("ContainerAttachedCluste
...
-	// +optional
-	ProjectRef *refs.ProjectRef `json:"projectRef,omitempty"`
+	ProjectRef *refs.ProjectRef `json:"projectRef"`
```
yuwenma, line 29: Mind elaborating this change?

hankfreund: In the existing tf-based resource, this field is required even though the comment implies it is optional.
In my previous PR I made this optional to match the comment, but you called this out as a change in behavior, so I said I'd fix it. This is just putting everything back the way it was.

pkg/controller/direct/containerattached/containerattachedcluster_mappings.go
hankfreund: That is correct.
And I did intend to rename it because this name is silly, I just forgot. My mistake!

apis/containerattached/v1beta1/containerattachedcluster_types.go
hankfreund: It was always a required field and the docs were misleading. I have updated them.

apis/containerattached/v1beta1/containerattachedcluster_types.go
yuwenma: Sorry I didn't make it clear. I mean the removal of the `omitempty`. I'm wondering why you make this change (because this could be an interesting topic!)

PR #2974, opened 2024-10-22
Remove map field if it is externally-managed using the powertool

maqiuyujoyce:
### Change description

This PR fixes the powertool behavior when the non-leaf object field is externally-managed: This means all the subfields are also externally-managed and should be removed after switching to `state-into-spec: absent`.

Fixes b/356413432

### Tests you have done

### DIFF
--- a/pkg/stateintospec/remove.go
+++ b/pkg/stateintospec/remove.go
@@ -67,7 +67,7 @@ func RemoveStateIntoSpecFields(ctx context.Context, u *unstructured.Unstructured
 					return
 				}
 				log.Info("removing field", "field", pathName)
-				errs = append(errs, removeFieldIfLeaf(ctx, u, fieldPath, warnings))
+				errs = append(errs, removeFieldIfNotArray(ctx, u, fieldPath, warnings))
 			case ".status", ".metadata":
 				// Never part of state-into-spec, ignore
 			default:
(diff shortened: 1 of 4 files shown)


maqiuyujoyce:
I did a quick experiment on KCC v1.119.0 and found that it's possible that the `spec.resourceID` field is still managed by `cnrm-controller-manager` after the field is added to the desired state (e.g. added to the manifest and got applied).
It needs further validation whether this is an edge case (as `spec.resourceID` is handled specially) or a common case. I'll need to understand whether an object field managed by `cnrm-controller-manager` is truly externally-managed (e.g. whether its manager becomes `kubectl` if the user "acquire"s subfields of this originally externally-managed field), and whether a field will become managed at all after being mentioned/updated in the manifest.

Here is the test I did:
1. `kubectl apply -f` the following yaml manifest:
    ```
    apiVersion: storage.cnrm.cloud.google.com/v1beta1
    kind: StorageBucket
    metadata:
      name: test-bucket-202411-[uniqueID]
      namespace: test
      annotations:
        cnrm.cloud.google.com/deletion-policy: "abandon"
    spec:
      versioning:
        enabled: false
      lifecycleRule:
        - action:
            type: Delete
          condition:
            age: 7
    ```
3. Described the state of the resource and verified `spec.resourceID` was managed by `cnrm-controller-manager`. Actually `spec.lifecycleRule` was also interpreted as being managed by `cnrm-controller-manager` so it looks like an object field will be marked as externally-managed when any of its subfield is externally-managed. This is kinda unexpected I think (?)
4. Updated the yaml manifest to below and ran `kubectl apply -f` again:
    ```
    apiVersion: storage.cnrm.cloud.google.com/v1beta1
    kind: StorageBucket
    metadata:
      name: test-bucket-202411-[uniqueID]
      namespace: test
      annotations:
        cnrm.cloud.google.com/deletion-policy: "abandon"
    spec:
      versioning:
        enabled: false
      lifecycleRule:
        - action:
            type: Delete
          condition:
            age: 7
      resourceID: test-bucket-202411-[uniqueID]
    ```
6. Described the state of the resource and found `spec.resourceID` is still managed by `cnrm-controller-manager`.

ldanielmadariaga:

PR #2977, opened 2024-10-22
chore: add script to download the e2e golden output from real-gcp

justinsb:
This is often faster than running the tests against real GCP directly.

Currently we don't expose the logs publicly in case they have
credentials in them (though they should be fine), so this script is
probably only usable by googlers.

### DIFF
--- a/dev/tasks/download-e2e-logs
+++ b/dev/tasks/download-e2e-logs
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


dev/tasks/download-e2e-logs
```diff
@@ -0,0 +1,48 @@
...
+
+if [[ -z "${TEST:-}" ]]; then
+  echo "Must specify TEST (e.g. TEST=e2e-service-pubsub)"
+  echo "Listing tests:"
+  gsutil ls gs://cnrm-prow/logs/ | sed -e "s@gs://cnrm-prow/logs/@@g" | sed -e "s@/@@g"
+  exit 1
+fi
+
+echo "Getting latest test run for ${TEST}"
+LATEST=$(gsutil cat gs://cnrm-prow/logs/${TEST}/latest-build.txt)
```
yuwenma, line 39: This seems to be not for developers. Shall we move it to somewhere else like `dev/ci`?

yuwenma (commented):
/lgtm
/approve
/hold

I have a clarification question about the use. Feel free to unhold if the script is for contributors to use.

dev/tasks/download-e2e-logs
justinsb: It is for developers!  If we have an e2e test, that has the golden logs from GCP, and it's much faster to download them than to run `hack/record-gcp`.  (Of course this doesn't work for brand new test fixtures)

One downside is that we don't know yet if we can expose this to the public Internet, for now it's only the KCC eng team at google that has access to the bucket I think.

PR #2978, opened 2024-10-22
tests: Only load the CRDs we are using

justinsb:
Speeds up tests a little, but also keeps the logs much more focused.


yuwenma (commented):
/lgtm
/approve

PR #2972, opened 2024-10-21
Skip ignored CRDs when listing CRDs in the recorder

maqiuyujoyce:
### Change description

Fixes #2874

### Tests you have done

- [N/A] Run `make ready-pr` to ensure this PR is ready for review.
- [X] Perform necessary E2E testing for changed resources.

Built the change locally and verified that orphaned CRDs covered by the IgnoredCRDList won't cause the issue any more.

Issue #2874, opened 2024-10-08
CRDs listed in IgnoredCRDList are causing RBAC errors in cnrm-resource-stats-recorder


dwilliams782:
### Checklist

- [X] I did not find a related open issue.
- [X] I did not find a solution in the troubleshooting guide: (https://cloud.google.com/config-connector/docs/troubleshooting)
- [X] If this issue is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Bug Description

Apologies if this is a duplicate but I couldn't find it.

We've upgraded to 1.123.1, and we get the following RBAC logs in cnrm-resource-stats-recorder:

```
E1008 12:19:54.881132       1 main.go:183] "error recording metrics for CRD" err="error listing objects for gameservices.cnrm.cloud.google.com/v1beta1, Kind=GameServicesRealm: error listing objects: gameservicesrealms.gameservices.cnrm.cloud.google.com is forbidden: User \"system:serviceaccount:cnrm-system:cnrm-resource-stats-recorder\" cannot list resource \"gameservicesrealms\" in API group \"gameservices.cnrm.cloud.google.com\" at the cluster scope" gvk="gameservices.cnrm.cloud.google.com/v1beta1, Kind=GameServicesRealm"

E1008 12:19:54.881242       1 main.go:183] "error recording metrics for CRD" err="error listing objects for core.cnrm.cloud.google.com/v1alpha1, Kind=ServiceMapping: error listing objects: servicemappings.core.cnrm.cloud.google.com is forbidden: User \"system:serviceaccount:cnrm-system:cnrm-resource-stats-recorder\" cannot list resource \"servicemappings\" in API group \"core.cnrm.cloud.google.com\" at the cluster scope" gvk="core.cnrm.cloud.google.com/v1alpha1, Kind=ServiceMapping"
```

Both of these CRDs are listed in [IgnoredCRDList](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/operator/pkg/k8s/constants.go#L66C2-L66C16), but we still get the RBAC errors as above. I've since deleted these CRDs from the cluster to silence the logs, but I wanted to share for awareness.

### Additional Diagnostic Information

n/a

### Kubernetes Cluster Version

v1.30.3-gke.1969001

### Config Connector Version

1.123.1

### Config Connector Mode

cluster mode

### Steps to reproduce the issue

[...]

PR #2980, opened 2024-10-22
Add warning log to clean up orphaned CRDs

maqiuyujoyce:
### Change description

Follow up on #2972 .

### Tests you have done

Verified that the error showed up in the log when orphaned CRD existed, and disappeared after the orphaned CRD was deleted.

### DIFF
--- a/cmd/recorder/main.go
+++ b/cmd/recorder/main.go
@@ -192,11 +192,14 @@ func doRecord(ctx context.Context, c client.Client, gvks []schema.GroupVersionKi
 	return nil
 }
 
-func forEach(c client.Client, gvk schema.GroupVersionKind, listOptions *client.ListOptions, fn func(unstructured.Unstructured) error) error {
+func forEach(ctx context.Context, c client.Client, gvk schema.GroupVersionKind, listOptions *client.ListOptions, fn func(unstructured.Unstructured) error) error {
+	logger := klog.FromContext(ctx)
 	for ok := true; ok; ok = listOptions.Continue != "" {
 		crdName := gvkToCRDName(gvk)
 		if _, ok := opk8s.IgnoredCRDList[crdName]; ok {
-			logger.Info(fmt.Sprintf("Orphaned CRD %s found, you should delete it using 'kubectl delete crd %s'", crdName, crdName))
+			logger.Error(fmt.Errorf("unexpected CRD %s", crdName),
+				fmt.Sprintf("Orphaned CRD %s found, you should delete "+
+					"it using 'kubectl delete crd %s'", crdName, crdName))
 			continue
 		}
 		list := unstructured.UnstructuredList{}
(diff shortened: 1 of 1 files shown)


cmd/recorder/main.go
```diff
@@ -194,7 +194,9 @@ func doRecord(ctx context.Context, c client.Client, gvks []schema.GroupVersionKi
 
 func forEach(c client.Client, gvk schema.GroupVersionKind, listOptions *client.ListOptions, fn func(unstructured.Unstructured) error) error {
 	for ok := true; ok; ok = listOptions.Continue != "" {
-		if _, ok := opk8s.IgnoredCRDList[gvkToCRDName(gvk)]; ok {
+		crdName := gvkToCRDName(gvk)
+		if _, ok := opk8s.IgnoredCRDList[crdName]; ok {
+			logger.Info(fmt.Sprintf("Orphaned CRD %s found, you should delete it using 'kubectl delete crd %s'", crdName, crdName))
```
yuwenma, line 199: Logger.Error?

maqiuyujoyce: I think it is not an error to keep orphaned CRDs. If we use logger.Error() here, then the recorder will act the same as what reported in https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/2874 except for a more actionable message.

yuwenma: >  then the recorder will act the same as what reported in #2874 except for a more actionable message.

Right, we expect the users to fix their stale CRD by giving clear guidance and not break their usage. 

Also suggest adding the `logger := klog.FromContext(ctx)`.

maqiuyujoyce: Done.

cmd/recorder/main.go
```diff
@@ -192,9 +192,14 @@ func doRecord(ctx context.Context, c client.Client, gvks []schema.GroupVersionKi
...
-func forEach(c client.Client, gvk schema.GroupVersionKind, listOptions *client.ListOptions, fn func(unstructured.Unstructured) error) error {
+func forEach(ctx context.Context, c client.Client, gvk schema.GroupVersionKind, listOptions *client.ListOptions, fn func(unstructured.Unstructured) error) error {
+	logger := klog.FromContext(ctx)
 	for ok := true; ok; ok = listOptions.Continue != "" {
-		if _, ok := opk8s.IgnoredCRDList[gvkToCRDName(gvk)]; ok {
+		crdName := gvkToCRDName(gvk)
+		if _, ok := opk8s.IgnoredCRDList[crdName]; ok {
+			logger.Error(fmt.Errorf("unexpected CRD %s", crdName),
+				fmt.Sprintf("Orphaned CRD %s found, you should delete "+
+					"it using 'kubectl delete crd %s'", crdName, crdName))
```
yuwenma, line 202: two nits:
1. go log does not use capitalized message. "Orphaned" --> "orphaned" 
2. the logger is a structured message. so it is better to be
```go
logger.Error( <error>, "please run `kubectl delete crd <crd_name>` to delete the orphaned CRD", "crd", crdName)
```

maqiuyujoyce: Done.

PR #2982, opened 2024-10-22
tests: warn, don't panic, if we can't read spec.resourceID

justinsb:
This shouldn't happen, but we need to know why it happens if it does.


yuwenma (commented):
/lgtm
/approve

Do you mind sharing more context? Do we have some issue?

PR #2986, opened 2024-10-23
build:add deploy commands

xiaoweim:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -185,12 +185,14 @@ repo to quickly set up a local dev environment.
 
         ```shell
         make deploy-kcc-standard
+        make install
         ```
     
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -185,12 +185,14 @@ repo to quickly set up a local dev environment.
...
         ```shell
         make deploy-kcc-standard
+        make install
         ```
     
         For autopilot clusters, please use the following command.
 
         ```shell
         make deploy-kcc-autopilot
+        make install
```
cheftako, line 195: install should probably be renamed install-crds

cheftako: If the operator will install the CRDs, do we need the explicit make install step?

PR #2994, opened 2024-10-23
fix: Assign default values for SQLInstance to fix sqlinstance-activationpolicy-direct

jasonvigil:
This fixes the failing periodic sqlinstance-activationpolicy-direct test, by assigning default values, so the controller does not send an unnecessary update call on re-reconcile.


yuwenma (commented):
/lgtm
/approve

PR #2995, opened 2024-10-23
tests: create generic fuzz tester

justinsb:
Fix some nits in the workstation cluster mapper also.

### DIFF
--- a/pkg/fuzztesting/fuzzkrm.go
+++ b/pkg/fuzztesting/fuzzkrm.go
@@ -1,0 +1,149 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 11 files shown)


yuwenma (commented):
/lgtm
/approve

Cool! Thank you

pkg/fuzztesting/fuzzkrm.go
acpana, line 1: This is really cool! I'm just catching up on it. 

I was looking to add something similar to this for these:

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/790c54b1bae78e405615273abbd7e54376180a7d/dev/tasks/find-missing-fields#L23-L46

Which gets ran in:

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/790c54b1bae78e405615273abbd7e54376180a7d/.github/workflows/fuzz.yaml#L31-L47

It seems we might've raced a bit here. Happy to clean up the old way of doing things and migrate away to the new one!

PR #3001, opened 2024-10-24
DiscoveryEngineDataStore: controller

justinsb:
- **DiscoveryEngineDataStore: controller**
- **DiscoveryEngineDataStore: basic test**


apis/discoveryengine/v1alpha1/datastore_reference.go
```diff
@@ -110,72 +124,76 @@ func NewDiscoveryEngineDataStoreRef(ctx context.Context, reader client.Reader, o
...
+			return nil, fmt.Errorf("cannot change object key after creation; status=%q, new=%q",
+				statusID.String(), id.String())
+		}
 	}
-	return nil, fmt.Errorf("DiscoveryEngineDataStoreRef not initialized from `NewDiscoveryEngineDataStoreRef` or `NormalizedExternal`")
+	return id, nil
 }
 
-type DiscoveryEngineDataStoreParent struct {
+// +k8s:deepcopy-gen=false
```
yuwenma, line 154: I'm a little bit get lost here. Is this because the parent is not used as  runtime.Object?

apis/discoveryengine/v1alpha1/discoveryenginedatastore_types.go
```diff
@@ -67,6 +67,10 @@ type DiscoveryEngineDataStoreSpec struct {
 	/* The ID of the project in which the resource belongs.*/
 	ProjectRef *refs.ProjectRef `json:"projectRef"`
 
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Collection field is immutable"
+	// Immutable. The collection for the DataStore.
+	Collection string `json:"collection"`
```
yuwenma, line 72: I realized that parent is not suitable for immutable CEL because this can be tricked, and we already validate the parent immutability from the `status.externalRef`

pkg/controller/direct/discoveryengine/datastore_controller.go
```diff
@@ -0,0 +1,280 @@
...
+	if err != nil {
+		return err
+	}
+
+	desired := direct.ProtoClone(a.desired)
+	desired.Name = a.id.String()
+
+	// TODO(user): Update the field if applicable.
+	updateMask := &fieldmaskpb.FieldMask{}
+	if !reflect.DeepEqual(a.desired.DisplayName, a.actual.DisplayName) {
```
yuwenma, line 200: Is `displayname` the only mutable field?

pkg/controller/direct/discoveryengine/datastore_controller.go
```diff
@@ -0,0 +1,280 @@
...
+	log := klog.FromContext(ctx)
+	log.V(2).Info("creating discoveryengine datastore", "name", a.id)
+
+	desired := direct.ProtoClone(a.desired)
+	desired.Name = a.id.String()
+
+	gcpClient, err := a.model.client(ctx, a.id.ProjectID)
+	if err != nil {
+		return err
+	}
```
yuwenma, line 159: I'm wondering why we initialize the gcpClient in each CRUD call rather than put it under `AdapterForObject`

yuwenma (commented):
/approve

A few clarification questions, no blockers.

apis/discoveryengine/v1alpha1/datastore_reference.go
justinsb: Right, we don't need the deepcopy.

I'll remove, it's pretty harmless having an extra DeepCopy method.

apis/discoveryengine/v1alpha1/discoveryenginedatastore_types.go
justinsb: Not sure what you mean.  Are you saying we should not have the immutability CEL rule here?

pkg/controller/direct/discoveryengine/datastore_controller.go
justinsb: I don't really know ... it's the only supported mutable field right now, I guess.  I do think we need more consistent treatment of mutability.

apis/discoveryengine/v1alpha1/discoveryenginedatastore_types.go
yuwenma: Yes.

pkg/controller/direct/discoveryengine/datastore_controller.go
justinsb: It's because the need for the projectID because the API doesn't auto-infer the project.  But you're right, we have the projectId in both AdapterForObject and AdapterForURL so ... moved it.

apis/discoveryengine/v1alpha1/datastore_reference.go
```diff
@@ -98,8 +104,16 @@ func NewDiscoveryEngineDataStoreRef(ctx context.Context, reader client.Reader, o
 	if projectID == "" {
 		return nil, fmt.Errorf("cannot resolve project")
 	}
+
 	location := obj.Spec.Location
-	id.parent = &DiscoveryEngineDataStoreParent{ProjectID: projectID, Location: location}
+	if location == "" {
```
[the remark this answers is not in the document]
justinsb: It already was a required field, but added the `// +required` annotation to be explicit

apis/discoveryengine/v1alpha1/datastore_reference.go
```diff
@@ -98,8 +104,16 @@ func NewDiscoveryEngineDataStoreRef(ctx context.Context, reader client.Reader, o
...
+		return nil, fmt.Errorf("cannot resolve location")
+	}
+
+	collectionID := obj.Spec.Collection
+	if collectionID == "" {
```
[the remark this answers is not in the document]
justinsb: Added the required annotation!

apis/discoveryengine/v1alpha1/datastore_reference.go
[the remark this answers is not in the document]
yuwenma: Sg. Thanks

PR #3008, opened 2024-10-25
fix: Fix SQLInstance authorizednetworks periodic

jasonvigil:
Turns out, we need to set a default value for the IpConfiguration fields for Ipv4Enabled and SslMode, even if some of the IpConfiguration is specified. And, we do not need to set a default for ServerCaMode.

Also, SslMode appears to have a different default value depending on which type of database is used.


yuwenma (commented):
/lgtm
/approve

PR #3015, opened 2024-10-28
tests: don't accidentally pick up "verbs" in URLs

justinsb:
For example, we don't want ":move" in

https://cloudresourcemanager.googleapis.com/v3/folders/${folderID}:move?alt=json&prettyPrint=false


yuwenma (commented):
/lgtm
/approve

Thanks!

PR #3020, opened 2024-10-29
test: fix dynamic tests for BigQueryDataTransferConfig

jingyih:
This PR includes:

 - Skipped the "drift correction test" for this resource because it has a service-generated ID.
 - Added functionality in the dynamic e2e test to ignore sensitive fields (masked by the GCP API) during "update test".

Tested
```
$ go test -v -tags=integration ./pkg/controller/dynamic/ -test.run TestCreateNoChangeUpdateDelete -run-tests bigquerydatatransferconfig-salesforce

$ go test -v -tags=integration ./pkg/controller/dynamic/ -test.run TestCreateNoChangeUpdateDelete -run-tests bigquerydatatransferconfig-scheduledquery
```

For reference, the original error in "update test":
```
    dynamic_controller_integration_test.go:820: unexpected diff:   string(
        -       "AKfZ4pamHD_459mErG-pofG1xFFX1F3JRWRCh8nULWvFVHa1U3ZG1vFywgzDKR8",
        +       "client-id2",
          )
```

### DIFF
--- a/pkg/controller/dynamic/dynamic_controller_integration_test.go
+++ b/pkg/controller/dynamic/dynamic_controller_integration_test.go
@@ -452,8 +452,9 @@ func testUpdate(ctx context.Context, t *testing.T, testContext testrunner.TestCo
 		if gcpUnstruct.Object["spec"] == nil {
 			t.Fatalf("GCP resource has a nil spec even though it was created using a resource with a non-nil spec")
 		}
-		changedFields := getChangedFields(initialUnstruct.Object, reconciledUnstruct.Object, "spec")
-		assertObjectContains(t, gcpUnstruct.Object["spec"].(map[string]interface{}), changedFields)
+		changedSpecFields := getChangedFields(initialUnstruct.Object, reconciledUnstruct.Object, "spec")
+		removeSensitiveFeilds(reconciledUnstruct, changedSpecFields) // remove sensitive fields which are reacted by the GCP API
+		assertObjectContains(t, gcpUnstruct.Object["spec"].(map[string]interface{}), changedSpecFields)
 	}
 
 	// Check that an "Updating" event was recorded, indicating that the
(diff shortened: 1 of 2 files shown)


pkg/controller/dynamic/dynamic_controller_integration_test.go
```diff
@@ -452,8 +452,9 @@ func testUpdate(ctx context.Context, t *testing.T, testContext testrunner.TestCo
 		if gcpUnstruct.Object["spec"] == nil {
 			t.Fatalf("GCP resource has a nil spec even though it was created using a resource with a non-nil spec")
 		}
-		changedFields := getChangedFields(initialUnstruct.Object, reconciledUnstruct.Object, "spec")
-		assertObjectContains(t, gcpUnstruct.Object["spec"].(map[string]interface{}), changedFields)
+		changedSpecFields := getChangedFields(initialUnstruct.Object, reconciledUnstruct.Object, "spec")
+		removeSensitiveFeilds(reconciledUnstruct, changedSpecFields) // remove sensitive fields which are reacted by the GCP API
```
maqiuyujoyce, line 456: IIUC, those sensitive fields are unreadable, right? Is there any design in direct controller that captures unreadable fields and handle them differently?

jingyih: > IIUC, those sensitive fields are unreadable, right? Is there any design in direct controller that captures unreadable fields and handle them differently?

The API returns these fields with some random value. I have included an example test failure in the PR description.

jingyih:
Hi @yuwenma, during an offline discussion with @maqiuyujoyce, one question came up: For direct resources with service-generated IDs, what should be the expected behavior if the underlying GCP resource is gone (e.g., accidentally deleted by a user)? Currently, the controller calls `adapter.Find()`, which returns "not found" (but no error), then it calls `adapter.Create()` to create a new GCP resource which will have a different service-generated ID. Should the `adapter.Find()` return an error in this case?

pkg/controller/dynamic/dynamic_controller_integration_test.go
maqiuyujoyce: Ah, thank you for the pointer. This API design looks a bit weird. We might want to come up with a better wrap if it happens again...

PR #3021, opened 2024-10-29
Fix hanging deletion for PrivilegedAccessManagerEntitlement

maqiuyujoyce:
### Change description

Deletion is hanging for PrivilegedAccessManagerEntitlement with the following error:

```
ERROR 2024-10-29T06:58:03.310877291Z [resource.labels.containerName: deletiondefender] ... "error":"error with 
update call to API server: admission webhook "deny-immutable-field-updates.cnrm.cloud.google.com" denied the 
request: couldn't get ResourceConfig for kind PrivilegedAccessManagerEntitlement: error getting service mapping: unable 
to get service mapping: no mapping with name 'privilegedaccessmanager.cnrm.cloud.google.com' found"...
```

The root cause is our immutable fields validator cannot identify direct resources.

Simply `adding the ResourceConfig for kind PrivilegedAccessManagerEntitlement` as a workaround doesn't work because it'll try to load the corresponding TF resource from ./third_party and `google_privileged_access_manager_entitlement` doesn't exist there.

To fix the issue, a customized immutability check is added.

I also think it may not be a best design to store all the immutability checks in the immutable fields validator. We're going to support hundreds of direct resources, and adding all the immutability checks in one file will make the file unreadable.

This issue can be detected if we add samples and run the samples test. So I think supporting samples test should be a blocker to v1alpha1 resources even if we don't generate docs.

### Tests you have done

Did a local build and verified that PrivilegedAccessManagerEntitlement can be deleted.


maqiuyujoyce:
@gouthamkrishna FYI

maqiuyujoyce:
More context about the original issue:

1. While we have guide that direct resources should use [other options for immutability checks](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/v1.124.0/docs/develop-resources/api-conventions/validations.md#rule-2-immutable-field), we currently still [register all resources to the immutable fields validator](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/v1.124.0/pkg/webhook/register.go#L97) including the direct resources.
2. The immutable fields validator is triggered whenever there is a mutating operation against the CR [NOT instantiated by Controller Manager SA](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/v1.124.0/pkg/webhook/immutable_fields_validator.go#L81), e.g. `system:serviceaccount:cnrm-system:cnrm-controller-manager`.
3. When a resource is deleted, the deletion defender will first update the finalizers, which will cause a mutation of the CR instantiated by `system:serviceaccount:cnrm-system:cnrm-deletiondefender`. This triggers the logic to validate immutable fields in the resource in [immutable_fields_validator.go](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/v1.124.0/pkg/webhook/immutable_fields_validator.go).
4. Direct resources will by default go through the [validateImmutableFieldsForTFBasedResource() route](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/v1.124.0/pkg/webhook/immutable_fields_validator.go#L130) to validate if the immutable fields are changed unexpectedly.
5. For PrivilegedAccessManagerEntitlement, because it doesn't have `ResourceConfig` configured, which is required for a TF-based resource, the mutation and deletion got stuck by the immutable fields validator.

About the workaround:

Unfortunately, configuring the ResourceConfig for PrivilegedAccessManagerEntitlement won't work around this issue because the validateImmutableFieldsForTFBasedResource() route will also try to fetch the TF schema of the corresponding TF resource specified in the ResourceConfig from [./third_party](https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/v1.124.0/third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services), and the corresponding TF resource, `google_privileged_access_manager_entitlement` is not available and will result in a Not Found error. Because we don't plan to use the immutable fields validator to validate immutability of the resource, we can simply work around the issue by always allowing any mutation in PrivilegedAccessManagerEntitlement.

About the solution:

We should decide which resources should not be registered for the immutable fields validator (this should include PrivilegedAccessManagerEntitlement), and fix the issue by unregistering them. Due to the priority, I prefer we work around the original issue now and do the actual solution later.

maqiuyujoyce:
> More context about the original issue:
> 
> 1. While we have guide that direct resources should use [other options for immutability checks](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/v1.124.0/docs/develop-resources/api-conventions/validations.md#rule-2-immutable-field), we currently still [register all resources to the immutable fields validator](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/v1.124.0/pkg/webhook/register.go#L97) including the direct resources.
> 2. The immutable fields validator is triggered whenever there is a mutating operation against the CR [NOT instantiated by Controller Manager SA](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/v1.124.0/pkg/webhook/immutable_fields_validator.go#L81), e.g. `system:serviceaccount:cnrm-system:cnrm-controller-manager`.
> 3. When a resource is deleted, the deletion defender will first update the finalizers, which will cause a mutation of the CR instantiated by `system:serviceaccount:cnrm-system:cnrm-deletiondefender`. This triggers the logic to validate immutable fields in the resource in [immutable_fields_validator.go](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/v1.124.0/pkg/webhook/immutable_fields_validator.go).
> 4. Direct resources will by default go through the [validateImmutableFieldsForTFBasedResource() route](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/v1.124.0/pkg/webhook/immutable_fields_validator.go#L130) to validate if the immutable fields are changed unexpectedly.
> 5. For PrivilegedAccessManagerEntitlement, because it doesn't have `ResourceConfig` configured, which is required for a TF-based resource, the mutation and deletion got stuck by the immutable fields validator.
> 
> About the workaround:
> 
> Unfortunately, configuring the ResourceConfig for PrivilegedAccessManagerEntitlement won't work around this issue because the validateImmutableFieldsForTFBasedResource() route will also try to fetch the TF schema of the corresponding TF resource specified in the ResourceConfig from [./third_party](https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/v1.124.0/third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services), and the corresponding TF resource, `google_privileged_access_manager_entitlement` is not available and will result in a Not Found error. Because we don't plan to use the immutable fields validator to validate immutability of the resource, we can simply work around the issue by always allowing any mutation in PrivilegedAccessManagerEntitlement.
> 
> About the solution:
> 
> We should decide which resources should not be registered for the immutable fields validator (this should include PrivilegedAccessManagerEntitlement), and fix the issue by unregistering them. Due to the priority, I prefer we work around the original issue now and do the actual solution later.

yuwenma:
> Discussed with @yuwenma offline that we want to avoid immutable field validation in the webhook for all direct resources except for SQLInstance.
>
More information can be found [in the guide](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/develop-resources/api-conventions/validations.md#option-2). "Some Config Connector resources use webhook to validate immutable fields. We can continue using that, but for external contributors CEL is much easier for self-learning."


> This will cause behavior change to other direct resources for immutability checks during CR mutation but we are okay with such change.

We may want to be more specific to avoid confusing contributors/users. We change the immutable check to a different component (from webhook --> CRD or controller), this does not necessarily will change the behavior.

pkg/webhook/immutable_fields_validator.go
```diff
@@ -294,6 +294,9 @@ func validateImmutableFieldsForTFBasedResource(obj, oldObj *unstructured.Unstruc
 		return admission.Errored(http.StatusBadRequest,
 			fmt.Errorf("couldn't get ResourceConfig for kind %v: %w", obj.GetKind(), err))
 	}
+	if rc.Direct && rc.Name != "google_sql_database_instance" {
```
yuwenma, line 297: cc @jasonvigil  Just FYI, we skip the immutable webhook check for other direct resources, and leave the sql there to make sure its behavior is unchanged.

yuwenma:
/approve
/lgtm 
/hold

@gemmahou could you take a look at the change, to make sure if the  webhook change will affect ComputeForwardingRule (and other Compute resources you work on) once we move them to the direct controller?

gemmahou:
> /approve /lgtm /hold
> 
> @gemmahou could you take a look at the change, to make sure if the webhook change will affect ComputeForwardingRule (and other Compute resources you work on) once we move them to the direct controller?

IIUC, we want to move the immutability check from webhook to CRD(by adding the CEL rule([guide](immutability)))
Some thoughts for ComputeForwardingRule: 
1. this should not affect TF based ComputeForwardingRule.
2. will change behavior for users who opt-in direct controller, but won't "break" anything because the underlying API will fail those requests to update immutable fields.

I need to have a follow-up PR to add CEL rule on **immutable, and required fields**.  But that should not be a blocker(see point 2 above). My concern is, there are several **immutable, but optional fields**, and seems like they cannot be handled by CEL rule. Without immutability webhook, how do we want to handle these fields?

maqiuyujoyce:
> 1. this should not affect TF based resource.

The change in this PR doesn't affect any TF-based resource that doesn't have `direct: true`.

> 2. will change behavior for users who opt-in direct controller, but won't "break" anything because the underlying API will fail those requests to update immutable fields.

Opt-in only direct resources don't have `direct: true` in their resource config, so they will still be using the immutable fields validator like before.

> My concern is, there are several immutable, but optional fields, and seems like they cannot be handled by CEL rule. Without immutability webhook, how do we want to handle these fields?

IIUC, we don't plan to cover all the immutability checks in our CRD, but do it at our best effort. If some immutability checks can't be done within Config Connector, we'll let the GCP API return the error. So for the migration, immutability checks are also on a best efforts basis. @yuwenma is this understanding correct?

gemmahou:
> > 1. this should not affect TF based resource.
> 
> The change in this PR doesn't affect any TF-based resource that doesn't have `direct: true`.
> 
> > 2. will change behavior for users who opt-in direct controller, but won't "break" anything because the underlying API will fail those requests to update immutable fields.
> 
> Opt-in only direct resources don't have `direct: true` in their resource config, so they will still be using the immutable fields validator like before.
> 
> > My concern is, there are several immutable, but optional fields, and seems like they cannot be handled by CEL rule. Without immutability webhook, how do we want to handle these fields?
> 
> IIUC, we don't plan to cover all the immutability checks in our CRD, but do it at our best effort. If some immutability checks can't be done within Config Connector, we'll let the GCP API return the error. So for the migration, immutability checks are also on a best efforts basis. @yuwenma is this understanding correct?

They might be handled in direct controller, I feel. I'm curious if handling ALL immutable fields should be a requirement, or something "nice to have".

maqiuyujoyce:
> They might be handled in direct controller, I feel. I'm curious if handling ALL immutable fields should be a requirement, or something "nice to have".

In some edge case situations, the immutability checks may need to be done by the controllers. I'd suggest we avoid it when possible. Regarding what kind of immutable checks should be covered and what can be optional, we should follow the latest guide (https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/develop-resources/api-conventions/validations.md#rule-2-immutable-field) and meanwhile avoid doing immutability checks on behalf of the API.

And I think for this PR, the goal is to avoid using the immutable field validator for the direct only resources. @gemmahou do you have any concern about it?

PR #3028, opened 2024-10-29
mockgcp: don't depend on "root" module

justinsb:
Introduce mocks/pointerutils.go to avoid depending on the "root" module.

Depending on the root creates a circular dependency, because the root
depends on the mockgcp module.  This in turn breaks `go mod tidy` and
friends, because the root module contains an import override.


acpana:
/lgtm
/approve

great catch! We (I) should write a rule against these kind of imports.

justinsb:
> We (I) should write a rule against these kind of imports.
I think just run `go mod tidy` in mockgcp catches it ... suggesting we probably aren't doing that today

PR #3030, opened 2024-10-29
tests: wait for the webhook server to start

justinsb:
Because we start the manager in a goroutine, there is a race where the
webhook may not have started yet.  We were observing some test
failures because of this.


yuwenma (commented):
/lgtm
/approve

Thank you!

PR #3031, opened 2024-10-29
fix: cbwp GCP server request/response project in different format

yuwenma:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/apis/refs/v1beta1/projectref.go
+++ b/apis/refs/v1beta1/projectref.go
@@ -168,3 +168,16 @@ func ResolveProjectID(ctx context.Context, reader client.Reader, obj *unstructur
 
 	return "", fmt.Errorf("cannot find project id for %v %v/%v", obj.GetKind(), obj.GetNamespace(), obj.GetName())
 }
+
+type ProjectIDAndNum struct {
+	ID     string
+	Number string
+}
+
+func ProjectIDToNumber(p *ProjectIDAndNum, valWithProjectID string) (valWithProjectNumber string) {
+	return strings.Replace(valWithProjectID, p.ID, p.Number, -1)
+}
+
+func ProjectNumberToID(p *ProjectIDAndNum, valWithProjectNum string) (valWithProjectID string) {
+	return strings.Replace(valWithProjectNum, p.Number, p.ID, -1)
+}
--- a/apis/refs/v1beta1/projectref.go
+++ b/apis/refs/v1beta1/projectref.go
@@ -168,24 +168,3 @@ func ResolveProjectID(ctx context.Context, reader client.Reader, obj *unstructur
 
 	return "", fmt.Errorf("cannot find project id for %v %v/%v", obj.GetKind(), obj.GetNamespace(), obj.GetName())
 }
-
-type ProjectIDAndNumer struct {
-	ID     string
-	Number string
-}
-
-func ProjectIDToNumber(p *ProjectIDAndNumer, valWithProjectID string) (string, error) {
-	replaced := strings.Replace(valWithProjectID, "projects/"+p.ID, "projects/"+p.Number, -1)
-	if replaced == valWithProjectID {
-		return "", fmt.Errorf("project ID was not found in %q", valWithProjectID)
-	}
-	return replaced, nil
-}
-
-func ProjectNumberToID(p *ProjectIDAndNumer, valWithProjectNum string) (string, error) {
-	replaced := strings.Replace(valWithProjectNum, "projects/"+p.Number, "projects/"+p.ID, -1)
-	if replaced == valWithProjectNum {
-		return "", fmt.Errorf("project Number was not found in %q", valWithProjectNum)
-	}
-	return replaced, nil
-}
--- a/apis/refs/v1beta1/projectref.go
+++ b/apis/refs/v1beta1/projectref.go
@@ -169,15 +169,23 @@ func ResolveProjectID(ctx context.Context, reader client.Reader, obj *unstructur
 	return "", fmt.Errorf("cannot find project id for %v %v/%v", obj.GetKind(), obj.GetNamespace(), obj.GetName())
 }
 
-type ProjectIDAndNum struct {
+type ProjectIDAndNumer struct {
 	ID     string
 	Number string
 }
 
-func ProjectIDToNumber(p *ProjectIDAndNum, valWithProjectID string) (valWithProjectNumber string) {
-	return strings.Replace(valWithProjectID, p.ID, p.Number, -1)
+func ProjectIDToNumber(p *ProjectIDAndNumer, valWithProjectID string) (string, error) {
+	replaced := strings.Replace(valWithProjectID, "projects/"+p.ID, "projects/"+p.Number, -1)
+	if replaced == valWithProjectID {
+		return "", fmt.Errorf("project ID was not found in %q", valWithProjectID)
+	}
+	return replaced, nil
 }
 
-func ProjectNumberToID(p *ProjectIDAndNum, valWithProjectNum string) (valWithProjectID string) {
-	return strings.Replace(valWithProjectNum, p.Number, p.ID, -1)
+func ProjectNumberToID(p *ProjectIDAndNumer, valWithProjectNum string) (string, error) {
+	replaced := strings.Replace(valWithProjectNum, "projects/"+p.Number, "projects/"+p.ID, -1)
+	if replaced == valWithProjectNum {
+		return "", fmt.Errorf("project Number was not found in %q", valWithProjectNum)
+	}
+	return replaced, nil
 }
--- a/mockgcp/mockcloudbuild/workerpool.go
+++ b/mockgcp/mockcloudbuild/workerpool.go
@@ -16,7 +16,6 @@ package mockcloudbuild
 
 import (
 	"context"
-	"reflect"
 	"strings"
 
 	"cloud.google.com/go/longrunning/autogen/longrunningpb"
@@ -64,6 +63,9 @@ func (s *CloudBuildV1) CreateWorkerPool(ctx context.Context, req *pb.CreateWorke
 
 	obj := proto.Clone(req.GetWorkerPool()).(*pb.WorkerPool)
 	obj.Name = fqn
+	obj.CreateTime = now
+
+	populateDefaultsForWorkerPool(obj)
 	if err := s.storage.Create(ctx, fqn, obj); err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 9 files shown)


pkg/test/resourcefixture/testdata/basic/cloudbuild/v1beta1/cloudbuildworkerpool/_http.log
yuwenma, line 1: I fixed several mock gcp issues and the log is very like the real GCP. The git-diff is not for CBWP resources but from the dependencies. See [diff](https://github.com/GoogleCloudPlatform/k8s-config-connector/compare/f1dfa03e1d07cd6951101fabce35d476a4b34cf5..0f6e3a986820c1692e15f80036fd0c88566b094e#diff-cc4c2efbec60c3989bcb6d62cfabd0ad967cd2b80b8680129da69b99189c656f) between real GCP and mock gcp (

pkg/test/resourcefixture/testdata/basic/cloudbuild/v1beta1/cloudbuildworkerpool/_generated_object_cloudbuildworkerpool.golden.yaml
```diff
@@ -38,10 +38,10 @@ status:
...
-      egressOption: NO_PUBLIC_EGRESS
-      peeredNetwork: projects/${projectId}/global/networks/computenetwork-${uniqueId}
+      egressOption: PUBLIC_EGRESS
+      peeredNetwork: projects/${projectNumber}/global/networks/computenetwork-${uniqueId}
       peeredNetworkIPRange: /29
     updateTime: "1970-01-01T00:00:00Z"
     workerConfig:
-      diskSizeGb: 100
-      machineType: e2-medium
+      diskSizeGb: 150
```
yuwenma, line 46: Those git-diff reflects the update.yaml (previous missing in the mock GCP). See [update.yaml](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/test/resourcefixture/testdata/basic/cloudbuild/v1beta1/cloudbuildworkerpool/update.yaml#L25)

pkg/controller/direct/cloudbuild/workerpool_controller.go
```diff
@@ -241,6 +252,14 @@ func (a *Adapter) Update(ctx context.Context, updateOp *directbase.UpdateOperati
 	}
 	wp.Name = a.id.FullyQualifiedName()
 	wp.Etag = a.actual.Etag
+	// the peered_network has a different HTTP request and response format.
+	// The HTTP request uses ProjectID, in the form of /projects/<projectID>/global/networks/<network>
```
justinsb, line 256: Can we avoid this by passing in the project number?

apis/refs/v1beta1/projectref.go
```diff
@@ -168,3 +168,16 @@ func ResolveProjectID(ctx context.Context, reader client.Reader, obj *unstructur
 
 	return "", fmt.Errorf("cannot find project id for %v %v/%v", obj.GetKind(), obj.GetNamespace(), obj.GetName())
 }
+
+type ProjectIDAndNum struct {
```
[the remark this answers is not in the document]
yuwenma: yeah.. my bad

pkg/controller/direct/cloudbuild/workerpool_controller.go
yuwenma: we want to use the compareProto.

apis/refs/v1beta1/projectref.go
```diff
@@ -168,3 +168,16 @@ func ResolveProjectID(ctx context.Context, reader client.Reader, obj *unstructur
...
+	ID     string
+	Number string
+}
+
+func ProjectIDToNumber(p *ProjectIDAndNum, valWithProjectID string) (valWithProjectNumber string) {
+	return strings.Replace(valWithProjectID, p.ID, p.Number, -1)
```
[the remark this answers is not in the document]
yuwenma: yeah, I"m not sure if I should put the functions here or just a cbwp internal function. 
For ` /projects/%s/` --> `/projects/%d/`, do you mean regex? I was in the impression that we just don't use regex in kcc.

pkg/controller/direct/cloudbuild/workerpool_controller.go
```diff
@@ -135,10 +135,20 @@ func (m *model) AdapterForObject(ctx context.Context, reader client.Reader, u *u
 	if err != nil {
 		return nil, err
 	}
+
+	projectIDAndNum := &refs.ProjectIDAndNum{ID: projectRef.ProjectID}
+	network, found, err := unstructured.NestedString(u.Object, "status", "observedState", "networkConfig", "peeredNetwork")
+	if err == nil || found {
+		tokens := strings.Split(network, "/")
```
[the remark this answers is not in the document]
yuwenma: I don't understand this comment, mind elaborating a little bit more? We use the computeNetworkRef in the mapper to find the compute network's external value. The problem is the GCP server converts the projectID (in request) to projectNum (in response), so the  `spec` contains projectID and the `status` contains projectNum. Then the next reconciliation `compareProto`  always raise error because it detects diff in this immutable field.

apis/refs/v1beta1/projectref.go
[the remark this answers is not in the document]
justinsb: So we can use regex, but regex tends to be very hard to read and thus require more tests than we save with most regexs.  At least in my experience!

I was thinking:

```
replaced := strings.Replace(valWithProjectID, "projects/" + p.ID + "/", "projects/" + strconv.Itoa(p.Number), -1)
if replaced == valWithProjectID {
return "", fmt.Errorf("project ID was not found in %q", valWithProjectID)
}
return replaced, nil
```

pkg/controller/direct/cloudbuild/workerpool_controller.go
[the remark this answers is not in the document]
justinsb: So IIUC what we're doing here is parsing a reference to a network resource.  That should be parseable using the ComputeNetworkRef functions, because I would hope the format is the same as the format used in an `external` ComputeNetworkRef.

pkg/controller/direct/cloudbuild/workerpool_controller.go
justinsb: No, I mean that peeredNetworkRef is a ComputeNetworkRef https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/86da25427f1a05f46a9cf7a8078d8ad61409140f/apis/cloudbuild/v1beta1/workerpool_types.go#L79

So what I'm suggesting is that we use something like LinkWithProjectNumber() instead of String() here: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/86da25427f1a05f46a9cf7a8078d8ad61409140f/pkg/controller/direct/cloudbuild/workerpool_controller.go#L130

The downside of doing this is that we also want to work towards resolving references using a generic visitor, and that's going to be harder if sometimes we want link-with-project-id and sometimes link-with-project-number, but I think we want to have correctness first and then make the resolving code elegant.

pkg/controller/direct/cloudbuild/workerpool_controller.go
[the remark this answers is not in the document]
yuwenma: Unfortunately, no. The computeNetworkRef does not support parsing a single string yet. We should migrate it to the new reference function format.

pkg/controller/direct/cloudbuild/workerpool_controller.go
yuwenma: This project ID/Num specialty seems to be less relevant to `ComputeNetwork` but more a CBWP problem. I looked at the ComputeNetwork object, it does not use ProjectNumber , see ` gcloud compute networks describe default`.  We can add a `LinkWithProjectID` method, in case other resource does the same thing. But I want to keep it as a special handling inside CBWP than adding the solution in ComputeNetwork. Because the computeNetwork has its own issues that we need to fix.

apis/refs/v1beta1/projectref.go
```diff
@@ -168,3 +168,24 @@ func ResolveProjectID(ctx context.Context, reader client.Reader, obj *unstructur
 
 	return "", fmt.Errorf("cannot find project id for %v %v/%v", obj.GetKind(), obj.GetNamespace(), obj.GetName())
 }
+
+type ProjectIDAndNumer struct {
+	ID     string
+	Number string
```
yuwenma, line 174: Ideally this ProjectNumber should be int64. But then the mockgcp is overly complicated because we want to normalize the ProjectNumber int64 to `${projectNumber}`, but this break the actual reconciliation because it parses`${projectNumber}`  to int64.

pkg/controller/direct/cloudbuild/workerpool_controller.go
```diff
@@ -135,10 +135,23 @@ func (m *model) AdapterForObject(ctx context.Context, reader client.Reader, u *u
...
 		return nil, err
 	}
+
+	projectIDAndNumer := &refs.ProjectIDAndNumer{ID: projectRef.ProjectID}
+	network, found, err := unstructured.NestedString(u.Object, "status", "observedState", "networkConfig", "peeredNetwork")
+	if err == nil && found {
+		tokens := strings.Split(network, "/")
+		if len(tokens) == 5 && tokens[0] == "projects" {
+			projectIDAndNumer.Number = tokens[1]
+			if err != nil {
```
justinsb, line 145: I think this is left over from parsing the number. Hopefully we can do that again if we change the mock.

pkg/controller/direct/cloudbuild/workerpool_controller.go
```diff
@@ -135,10 +135,23 @@ func (m *model) AdapterForObject(ctx context.Context, reader client.Reader, u *u
...
+		}
+	}
 	return &Adapter{
-		id:        id,
-		gcpClient: gcpClient,
-		desired:   obj,
+		id:              id,
+		gcpClient:       gcpClient,
+		desired:         obj,
+		projectIDAndNum: projectIDAndNumer,
```
justinsb, line 154: This could get confusing because the project ID is the ID of the cloud build worker pool, but the project number is the number of the peered network

mockgcp/mockcloudbuild/workerpool.go
```diff
@@ -64,74 +64,65 @@ func (s *CloudBuildV1) CreateWorkerPool(ctx context.Context, req *pb.CreateWorke
...
-				f.SetString(stringVal)
-			}
-
-		}
+	if err := fields.UpdateByFieldMask(obj, req.WorkerPool, req.UpdateMask.Paths); err != nil {
+		return nil, err
 	}
+	now := timestamppb.Now()
+	populateDefaultsForWorkerPool(obj)
+	obj.CreateTime = obj.GetCreateTime()
```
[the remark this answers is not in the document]
yuwenma: cleaned up. (Seems to be a left over from resolving the merge conflict)

pkg/controller/direct/cloudbuild/workerpool_controller.go
yuwenma: Good point! Let me give a better solution. I think this project ID <--> Number deserves a more careful handling.

yuwenma: I think the current `ProjectIDAndNumber` is not good. Mostly because I think we want to refactor both the `ProjectRef` and `ComputeNetwork` to a better format. Right now, none of the function can be reused.

justinsb (approved):

PR #3034, opened 2024-10-30
Fix SQL direct periodics (replica, encryptionkey, postgres, mysql)

jasonvigil:
See commit message for details

### DIFF
--- a/pkg/controller/direct/sql/sqlinstance_controller.go
+++ b/pkg/controller/direct/sql/sqlinstance_controller.go
@@ -347,6 +347,9 @@ func (a *sqlInstanceAdapter) Delete(ctx context.Context, deleteOp *directbase.De
 	if err != nil {
 		return false, fmt.Errorf("deleting SQLInstance %s failed: %w", a.resourceID, err)
 	}
+	if err := a.pollForLROCompletion(ctx, op, "delete"); err != nil {
+		return false, err
+	}
 
 	log.V(2).Info("deleted SQLInstance", "op", op)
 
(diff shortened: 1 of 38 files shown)


yuwenma:
/lgtm
/approve

Make sense

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -347,6 +347,9 @@ func (a *sqlInstanceAdapter) Delete(ctx context.Context, deleteOp *directbase.De
 	if err != nil {
 		return false, fmt.Errorf("deleting SQLInstance %s failed: %w", a.resourceID, err)
 	}
+	if err := a.pollForLROCompletion(ctx, op, "delete"); err != nil {
```
justinsb, line 350: I would have expected us to pass `deleteOp` into `pollForLROCompletion`

justinsb (approved):

pkg/controller/direct/sql/sqlinstance_controller.go
jasonvigil: We need a handle to the LRO to poll / wait. `deleteOp` does not provide that (currently). Probably some way to improve it in the future, but that would likely be a separate effort. We'd need to decide how to make the interface work for services that do not use LROs, and for each service's separate LRO implementation.

yuwenma: Thanks. Please update this once you have the follow up LRO improvements

PR #3038, opened 2024-10-30
chore: remove "generate-go-client" from presubmit check

### LINKED ISSUE #3037: Clean up the duplicated (and problematic) "generate-go-crd-clients"
### Checklist

*This issue is for tracking purpose*

This originated from triaging a bug that block [promoting a KMS Alpha resource to Beta](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3012) (direct), where the KMS services have both TF and direct resources in both Alpha and Beta versions, and the autogenerated client cannot import the autogenerated API (alpha), when promoting alpha to beta and keep the alpha CRD. It turns out that the bug comes from `generate-go-crd-clients` where it cannot support auto-generating APIs with multiple CRD versions. Instead, it deletes the autogenerated alpha API.

The real problem is that the  `generate-go-crd-clients`  generates the api and go client lib using go template, rather than relying on the sig-k8s its own solution (controller-gen). 

As we migrating to the [new "direct" approach](https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/master/docs/develop-resources), we are using the controller-gen to build the api client and can now support multiple CRD versions. However, the  `generate-go-crd-clients`  is still there in the presubmit infra, generating duplicate code and blocks the new direct resources. 

We want to  remove it from the presubmit infra immediately to unblock new resources. But we cannot delete the existing auto-generate code because OSS users may still need it.  

### Additional information

_No response_

### Importance

_No response_

yuwenma:
### Change description

See  full context in #3037 

Fixes #3037

### Tests you have done


justinsb:
I don't think we should remove big functionality areas lightly.  That said, I do think we would recommend to people to use the dynamic client anyway.  Can we fix this functionality instead?

PR #3039, opened 2024-10-30
tests: switch placeholder from testFolderId -> folderID

justinsb:
Reducing the size of the bigger refactor to how replacements happen.


yuwenma:
/lgtm
/approve

kk. Thanks for the PR description!

PR #3042, opened 2024-10-30
fix dynamic test cloudspannerconnectionbasic

yuwenma:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/mockgcp/mockbigqueryconnection/connection.go
+++ b/mockgcp/mockbigqueryconnection/connection.go
@@ -201,18 +201,8 @@ func (s *ConnectionV1) UpdateConnection(ctx context.Context, req *pb.UpdateConne
 	if err := s.storage.Get(ctx, fqn, obj); err != nil {
 		return nil, err
 	}
-
-	paths := req.GetUpdateMask().GetPaths()
-	for _, path := range paths {
-		switch path {
-		case "friendlyName":
-			obj.FriendlyName = req.GetConnection().GetFriendlyName()
-		case "description":
-			obj.Description = req.GetConnection().GetDescription()
-		default:
-			return nil, status.Errorf(codes.InvalidArgument, "update_mask path %q not valid", path)
-		}
-	}
+	obj.FriendlyName = req.GetConnection().GetFriendlyName()
+	obj.Description = req.GetConnection().GetDescription()
 	obj.LastModifiedTime = now.Unix()
 
 	if _, ok := (req.Connection.Properties).(*pb.Connection_Aws); ok {
--- a/mockgcp/mockbigqueryconnection/connection.go
+++ b/mockgcp/mockbigqueryconnection/connection.go
@@ -28,6 +28,7 @@ import (
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/emptypb"
 
+	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/common/fields"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/common/projects"
 	pb "github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/generated/mockgcp/cloud/bigquery/connection/v1"
 )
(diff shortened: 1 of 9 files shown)


pkg/controller/direct/bigqueryconnection/connection_controller.go
```diff
@@ -223,26 +224,24 @@ func (a *Adapter) Update(ctx context.Context, updateOp *directbase.UpdateOperati
...
+		log.V(2).Info("no field needs update", "name", a.id.External)
+		return nil
+	}
 	fqn := a.id.External
 	req := &bigqueryconnectionpb.UpdateConnectionRequest{
 		Name:       fqn,
-		Connection: resource,
-		UpdateMask: updateMask,
+		Connection: connection,
+		UpdateMask: &fieldmaskpb.FieldMask{Paths: sets.List(paths)},
```
justinsb, line 244: Oh I like this ... I would expect the API to return an error here if we try to update an immutable field (or at least hope so), which makes it easier for us to write controllers!

justinsb (approved):

pkg/test/resourcefixture/testdata/basic/bigqueryconnection/v1alpha1/bigqueryconnectionconnection/awsconnectionbasic/_http.log
```diff
@@ -66,7 +66,7 @@ X-Xss-Protection: 0
 
 ---
 
-PATCH https://bigqueryconnection.googleapis.com/v1/projects/${projectId}/locations/aws-us-east-1/connections/71389360-831c-431d-8975-837aee2153be?%24alt=json%3Benum-encoding%3Dint&updateMask=friendlyName%2Cdescription
```
jingyih, line 69: Not sure why "name" is included in the updateMask?

mockgcp/mockbigqueryconnection/connection.go
```diff
@@ -201,18 +201,8 @@ func (s *ConnectionV1) UpdateConnection(ctx context.Context, req *pb.UpdateConne
...
-		switch path {
-		case "friendlyName":
-			obj.FriendlyName = req.GetConnection().GetFriendlyName()
-		case "description":
-			obj.Description = req.GetConnection().GetDescription()
-		default:
-			return nil, status.Errorf(codes.InvalidArgument, "update_mask path %q not valid", path)
-		}
-	}
+	obj.FriendlyName = req.GetConnection().GetFriendlyName()
```
jingyih, line 204: Not sure I follow the change. In the new mock implementation we are always setting these 2 fields?

Is this behavior specific to the real API? or do we plan to improve the mock in the future?

yuwenma: Are you asking why not checking the update field_mask? This is an easy hack because the BQCC controller now uses the `CompareProtoMessage`!

More context:
In general, if the GCP server supports update field mask,  the MockGCP needs to support each field mask as well to tell if a field can be changed or not. However, this requires a lot of engineering effort (boring, repeated code). I made some improvements recently that the controller can determine the field immutable in the `CompareProtoMessage` function via the proto, so that contributors no longer need to repeatedly write `reflect.DeepEqual` ... `if fieldPath[i] == "display_name" in either MockGCP and scifi controller. Corresponding code (https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/common/compare.go#L77C13-L77C28),

jingyih: The helper library looks very cool! Removing repeated code and effort sounds exciting. My only concern is that it might affect the mock's ability to guard the controller's behavior in presubmit tests?

pkg/test/resourcefixture/testdata/basic/bigqueryconnection/v1alpha1/bigqueryconnectionconnection/awsconnectionbasic/_http.log
yuwenma: Good catch. Looks like the auto mapper does not map `name` between proto and krm (I think that's right because this is a service-generated ID). I updated the code and golden log.

mockgcp/mockbigqueryconnection/connection.go
justinsb: I think we only want to copy the values if the field path is specified in the update field_mask

yuwenma: Updated. Also added a handler to only update the field based on field_mask

pkg/test/resourcefixture/testdata/basic/bigqueryconnection/v1alpha1/bigqueryconnectionconnection/cloudspannerconnectionbasic/update.yaml
```diff
@@ -24,6 +24,6 @@ spec:
     databaseRef:
       name: spannerdb-${uniqueId}
     maxParallelism: 150
-    useDataBoost: True
-    useParallelism: True
-    databaseRole: "admin"
+    useDataBoost: False
+    useParallelism: False
+    databaseRole: "user"
```
yuwenma, line 29: the previous update.yaml is the same as the create.yaml. This change verifies the MockGCP update now works.

pkg/controller/direct/bigqueryconnection/connection_controller.go
justinsb: Ah OK.  I was actually with sending invalid requests to GCP if the user tells us to (i.e. we can tackle our backoff strategy)

If compareProtoMessage does all this maybe we give it a more important sounding name - something like "ConstructUpdateRequest"?

mockgcp/mockbigqueryconnection/connection.go
```diff
@@ -202,35 +203,10 @@ func (s *ConnectionV1) UpdateConnection(ctx context.Context, req *pb.UpdateConne
...
-		switch path {
-		case "friendlyName":
-			obj.FriendlyName = req.GetConnection().GetFriendlyName()
-		case "description":
-			obj.Description = req.GetConnection().GetDescription()
-		default:
-			return nil, status.Errorf(codes.InvalidArgument, "update_mask path %q not valid", path)
-		}
+	if err := fields.UpdateByFieldMask(obj, req.Connection, req.UpdateMask.Paths); err != nil {
+		return nil, fmt.Errorf("update field_mask.paths: %w", err)
```
justinsb, line 207: Nit: arguably if we're using one of our functions we can just `return nil, err` here.  If we're not adding information that is specific to how we are calling UpdateByFieldMask, we might as well put it into the UpdateByFieldMask errors!

justinsb (approved):

justinsb (approved):

PR #3043, opened 2024-10-30
fix: Specify missing fields in clone backup configuration

jasonvigil:
This way, the test will not need to re-reconcile after cloning in order to update the backup configuration.


yuwenma (commented):
/lgtm
/approve

PR #3044, opened 2024-10-30
chore: update client generation to generate multiple CRD versions

justinsb:


yuwenma:
/lgtm
/approve

Thank you, @justinsb ! Hope we can unblock the kms auotkey upgrade!

PR #3046, opened 2024-10-30
Sort the AccessControlEntry slice in PrivilegedAccessManagerEntitlement before diffing

maqiuyujoyce:
### Change description

AccessControlEntry.Principals slices in PrivilegedAccessManagerEntitlement will be returned with an unexpected, random order, and reflect.DeepEqual() will take entry order into consideration when doing comparison. So we need to sort the AccessControlEntry.Principals slices before compare the actual state and the desired state of a PrivilegedAccessManagerEntitlement.

### Tests you have done

Verified that the change of order of principals in EligibleUsers and ApprovalWorkflow.ManualApprovals.Steps.Approvers won't trigger an update call to the API.

### DIFF
--- a/pkg/controller/direct/privilegedaccessmanager/entitlement_controller.go
+++ b/pkg/controller/direct/privilegedaccessmanager/entitlement_controller.go
@@ -308,36 +308,36 @@ func (a *Adapter) Update(ctx context.Context, updateOp *directbase.UpdateOperati
 
 	updateMask := &fieldmaskpb.FieldMask{}
 
-	parsedAndSortedActual := PrivilegedAccessManagerEntitlementSpec_FromProto(mapCtx, a.actual)
+	parsedActual := PrivilegedAccessManagerEntitlementSpec_FromProto(mapCtx, a.actual)
 	if mapCtx.Err() != nil {
 		return fmt.Errorf("error generating update mask: %w", mapCtx.Err())
 	}
-	sortArrayFieldsInSpec(parsedAndSortedActual)
-	sortedDesired := a.desired.DeepCopy()
-	sortArrayFieldsInSpec(&sortedDesired.Spec)
+	sortArrayFieldsInSpec(parsedActual)
+	parsedDesired := a.desired.DeepCopy()
+	sortArrayFieldsInSpec(&parsedDesired.Spec)
 
-	if !reflect.DeepEqual(parsedAndSortedActual.AdditionalNotificationTargets, sortedDesired.Spec.AdditionalNotificationTargets) {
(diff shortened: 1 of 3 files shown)


maqiuyujoyce:
@gouthamkrishna FYI I updated the AccessControlEntry slice comparison logic to avoid updates caused by reordering the items.

pkg/controller/direct/privilegedaccessmanager/entitlement_controller.go
yuwenma, line 1: Suggest not changing the name, because it is some subfields inside the spec that are being sorted, not the spec itself

pkg/controller/direct/privilegedaccessmanager/util.go
yuwenma, line 1: suggest using	"k8s.io/apimachinery/pkg/util/sets"

pkg/controller/direct/privilegedaccessmanager/entitlement_controller.go
maqiuyujoyce: Done.

pkg/controller/direct/privilegedaccessmanager/util.go
maqiuyujoyce: The `principal` list can contain multiple string entries of the same value. I checked `k8s.io/apimachinery/pkg/util/sets` and looks like this only supports true sets without duplicate values. So I feel this may not be a good match. Or did I miss anything?

maqiuyujoyce: Experiment result can be found here: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3077 Looks like sublists in each item in the set won't be sorted before comparison.

maqiuyujoyce: @yuwenma I did some further digging. Though not sure if that's what you suggested.

Using "k8s.io/apimachinery/pkg/util/sets" to establish a new set meaning the original type needs to be a comparable. If we compare at the `eligibleUsers`/`approvers` level, because `AccessControlEntry` contains field `principals`, which is a list of strings, AccessControlEntry can never be a comparable and can't be used to construct an apimachinary set.

If it is done at the `principals` level, converting it to apimachinary set is also a bunch of code to loop through the `eligibleUsers`/`approvers` list and parse the nested string slice, which is actually a lot of comparison code: fb26e8c1665e18458df2d945711d115a3a866299

So I'd suggest we do sorting/canonicalizing before DeepEqual() instead of using special handling for equaling.

yuwenma (commented):
/lgtm
/approve

Thanks. One suggestion. Non blocker

pkg/controller/direct/privilegedaccessmanager/util_test.go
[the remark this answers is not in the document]
maqiuyujoyce: I think I covered all the test cases here in the unit test... or are you suggesting something like a scenario test?

PR #3055, opened 2024-10-31
tests: rename tests with same name

justinsb:
The IAMPolicy samples and IAMPartialPolicy samples had the same name,
causing our test results to be ambiguous.

Rename to deduplicate.


yuwenma (commented):
/lgtm
/approve

PR #3056, opened 2024-10-31
tests: normalize addresses for RedisInstance

justinsb:
This avoids the non-determinism


yuwenma (commented):
/lgtm
/approve

Cool!

PR #3058, opened 2024-10-31
chore: download logs from golden tests also

justinsb:


dev/tasks/download-e2e-logs
```diff
@@ -46,3 +46,6 @@ gsutil cp gs://cnrm-prow/logs/${TEST}/${LATEST}/artifacts/golden.zip .build/gold
 
 echo "Expanding golden output into testdata directory"
 unzip -o .build/golden.zip "pkg/test/resourcefixture/testdata/*"
+
+echo "Downloading test log to e2e.log"
+gsutil cp gs://cnrm-prow/logs/${TEST}/${LATEST}/build-log.txt e2e.log
```
yuwenma, line 51: suggest adding a timestamp to the downloaded file name.

yuwenma (commented):
/approve
/lgtm
/hold 

Thank you! One minor suggestion, but all defer to you.

dev/tasks/download-e2e-logs
justinsb: The nice thing about going to the same file is that we can grep it easily from other scripts, and that we don't end up with dozens of e2e log files because they overwrite each other.

I think also because the file still exists on GCS, we're not going to lose data we care about by overwriting.

justinsb:
Let's try it this way and see how it goes, we can add a timestamp later if needed.

/hold cancel

dev/tasks/download-e2e-logs
yuwenma: fair enough.

PR #3059, opened 2024-10-31
tests: log test failures with FAIL

justinsb:
The go formatting of t.Fatalf and t.Errorf is not very obvious.

Manually change the messages so that they stand out more.

### DIFF
--- a/pkg/test/utils.go
+++ b/pkg/test/utils.go
@@ -83,7 +83,7 @@ func CompareGoldenFile(t *testing.T, p string, got string, normalizers ...func(s
 			// Golden file won't be generated if the result is empty.
 			return
 		} else {
-			t.Fatalf("failed to read golden file %q: %v", p, err)
+			t.Fatalf("FAIL: failed to read golden file %q: %v", p, err)
 		}
 	}
 	want := string(wantBytes)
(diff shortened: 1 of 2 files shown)


yuwenma (commented):
Can I ask why we need this change? golang linter suggests starting with lower case, and FATAL/Error itself already contains the error so the convention is normally not including FAILURE/ERROR in the log itself.

yuwenma (commented):
Can I ask why we need this change? golang linter suggests starting with lower case, and FATAL/Error itself already contains the error so the convention is normally not including FAILURE/ERROR in the log itself.

justinsb:
> Can I ask why we need this change? golang linter suggests starting with lower case, and FATAL/Error itself already contains the error so the convention is normally not including FAILURE/ERROR in the log itself.

Sure - the problem is that in our logs, it's very difficult to see the error.  The error tends to be printed only with a line number.  In theory we could create a replacement formatter for the logs, but we aren't using the pretty formatter very widely AFAICT, and so this seems way easier.

If there's a different marker I should use to find the actual error for a test failure, LMK!

yuwenma:
> > Can I ask why we need this change? golang linter suggests starting with lower case, and FATAL/Error itself already contains the error so the convention is normally not including FAILURE/ERROR in the log itself.
> 
> Sure - the problem is that in our logs, it's very difficult to see the error. The error tends to be printed only with a line number. In theory we could create a replacement formatter for the logs, but we aren't using the pretty formatter very widely AFAICT, and so this seems way easier.
> 
> If there's a different marker I should use to find the actual error for a test failure, LMK!

How about colorizing the log output? That's more straightforward and benefit other contributors who don't know which keywords to search for. Something like grc https://github.com/garabik/grc 

My main concern for `t.Fatalf("FAIL: failed ` is the triplicated "failure" makes it less readable. But I'm fine if you find that's easier to read the log.

justinsb:
/hold cancel

Agree we might want to do something better and clean up the messages, but that is a separate effort - I think anh was doing that?

PR #3063, opened 2024-10-31
feat: Make SQLInstance use direct controller by default

jasonvigil:
Allow the user to "opt-out" of using the direct controller and fall back to using the legacy controller by setting the annotation:
```
alpha.cnrm.cloud.google.com/reconciler: legacy
```


jasonvigil:
Per offline discussion, we decided against the "opt out" approach for now

PR #3068, opened 2024-11-01
feat: simplify the mockgcp UPDATE based on field-mask

yuwenma:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/mockgcp/common/fields/updatemask.go
+++ b/mockgcp/common/fields/updatemask.go
@@ -23,58 +23,70 @@ import (
 	"google.golang.org/protobuf/reflect/protoreflect"
 )
 
-func UpdateByFieldMask(old, new proto.Message, updatePaths []string) error {
+// UpdateByFieldMask updates the `original` Message with the `update` Message value in the given `updatePaths` fields
+func UpdateByFieldMask(original, update proto.Message, updatePaths []string) error {
 	var errs []error
 	for _, path := range updatePaths {
-		if err := walk(old, new, path); err != nil {
+		if err := walk(original, update, path); err != nil {
 			errs = append(errs, err)
 		}
 	}
 	return errors.Join(errs...)
 }
 
-func walk(old, new proto.Message, path string) error {
-	segments := strings.Split(path, ".")
+func walk(original, update proto.Message, path string) error {
+	originalRm := original.ProtoReflect()
+	updateRm := update.ProtoReflect()
+	segments := strings.SplitN(path, ".", 2)
 	if len(segments) == 1 {
-		return replace(old, new, segments[0])
+		return replace(originalRm, updateRm, segments[0])
 	}
-
-	subPath := strings.TrimPrefix(path, segments[0]+".")
-	return walk(childMessage(old, segments[0]), childMessage(new, segments[0]), subPath)
+	return walk(originalChildMessage(originalRm, segments[0]), updateChildMessage(updateRm, segments[0]), segments[1])
 }
 
-func replace(old, new proto.Message, fieldName string) error {
-	oldFd := old.ProtoReflect().Descriptor().Fields().ByJSONName(fieldName)
-	oldVal := old.ProtoReflect().Get(oldFd)
-	newFd := new.ProtoReflect().Descriptor().Fields().ByJSONName(fieldName)
-	newVal := new.ProtoReflect().Get(newFd)
+func replace(original, update protoreflect.Message, fieldName string) error {
+	originalFd := original.Descriptor().Fields().ByJSONName(fieldName)
+	originalVal := original.Get(originalFd)
+	updateFd := update.Descriptor().Fields().ByJSONName(fieldName)
+	updateVal := update.Get(updateFd)
 
-	if oldFd.Kind() != protoreflect.MessageKind {
-		m := old.ProtoReflect()
-		if !m.IsValid() {
+	if originalFd.Kind() != protoreflect.MessageKind {
+		if !original.IsValid() {
 			return fmt.Errorf("%s is read-only or empty", fieldName)
 		}
-		m.Set(newFd, newVal)
+		original.Set(updateFd, updateVal)
+		return nil
 	}
 	// Update Map
-	if oldFd.IsMap() {
-		newVal.Map().Range(func(k protoreflect.MapKey, v protoreflect.Value) bool {
-			oldVal.Map().Set(k, v)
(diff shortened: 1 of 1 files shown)


mockgcp/common/fields/updatemask.go
```diff
@@ -0,0 +1,80 @@
...
+	segments := strings.Split(path, ".")
+	if len(segments) == 1 {
+		return replace(old, new, segments[0])
+	}
+
+	subPath := strings.TrimPrefix(path, segments[0]+".")
+	return walk(childMessage(old, segments[0]), childMessage(new, segments[0]), subPath)
+}
+
+func replace(old, new proto.Message, fieldName string) error {
```
justinsb, line 46: Does this code get simpler if we accept `protoreflect.Message` instead of `proto.Message`?  (may save a lot of calls to ProtoReflect)

mockgcp/common/fields/updatemask.go
```diff
@@ -0,0 +1,80 @@
...
+	newFd := new.ProtoReflect().Descriptor().Fields().ByJSONName(fieldName)
+	newVal := new.ProtoReflect().Get(newFd)
+
+	if oldFd.Kind() != protoreflect.MessageKind {
+		m := old.ProtoReflect()
+		if !m.IsValid() {
+			return fmt.Errorf("%s is read-only or empty", fieldName)
+		}
+		m.Set(newFd, newVal)
+	}
```
justinsb, line 58: `return nil` here?

mockgcp/common/fields/updatemask.go
```diff
@@ -0,0 +1,80 @@
...
+		})
+	}
+	// Update List
+	if oldFd.IsList() {
+		oldVal.List().Truncate(0)
+		for i := 0; i < newVal.List().Len(); i++ {
+			oldVal.List().Append(newVal.List().Get(i))
+		}
+	}
+	return nil
```
justinsb, line 73: `return fmt.Errorf("unhandled type for field %v", oldFd)` or something like that?

mockgcp/common/fields/updatemask.go
```diff
@@ -0,0 +1,80 @@
...
+	if oldFd.Kind() != protoreflect.MessageKind {
+		m := old.ProtoReflect()
+		if !m.IsValid() {
+			return fmt.Errorf("%s is read-only or empty", fieldName)
+		}
+		m.Set(newFd, newVal)
+	}
+	// Update Map
+	if oldFd.IsMap() {
+		newVal.Map().Range(func(k protoreflect.MapKey, v protoreflect.Value) bool {
```
justinsb, line 61: The semantics aren't well defined for maps AFAIK.  i.e. do we clear the map first?

Might be easier to leave this until we have an example...  Unless we have an example, in which case, great!

mockgcp/common/fields/updatemask.go
```diff
@@ -0,0 +1,80 @@
...
+		oldVal.List().Truncate(0)
+		for i := 0; i < newVal.List().Len(); i++ {
+			oldVal.List().Append(newVal.List().Get(i))
+		}
+	}
+	return nil
+}
+
+func childMessage(m proto.Message, fieldName string) proto.Message {
+	fd := m.ProtoReflect().Descriptor().Fields().ByJSONName(fieldName)
```
justinsb, line 77: Again, might be easier to pass in the value of `ProtoReflect()`

mockgcp/common/fields/updatemask.go
```diff
@@ -0,0 +1,80 @@
...
+	v := m.ProtoReflect().Get(fd)
```
justinsb, line 78: We want to auto-populate here, I think there's a method like `Mutable` Arguably we don't want to mutate the request, so _strictly_ we should have a different method for `old` vs `new`, but I suspect it's harmless in practice.  It does limit our ability to reuse the code though if we start saying "well we know this bug is harmless because of how it is called", so maybe it's better to just handle it.

mockgcp/common/fields/updatemask.go
```diff
@@ -0,0 +1,80 @@
...
+	return errors.Join(errs...)
+}
+
+func walk(old, new proto.Message, path string) error {
+	segments := strings.Split(path, ".")
+	if len(segments) == 1 {
+		return replace(old, new, segments[0])
+	}
+
+	subPath := strings.TrimPrefix(path, segments[0]+".")
```
justinsb, line 42: I like this "iterative" approach to parsing the field path, because it might one day be able to cope with more complex paths (e.g. `foo.bars[1]`).  👍 

You might consider `subPath := strings.Join(segments[1:], ".")`.

Or once I saw that, I realized you can probably also do this:

```
segments := strings.SplitN(path, ".", 2)
if len(segments) == 1 {
		return replace(old, new, segments[0])
	}

subPath := segments[1]
```

yuwenma: Thank you! I like it.

mockgcp/common/fields/updatemask.go
yuwenma: We could. But I also realize it is more maintainable if the passed-in args are of the same type (like `proto.Message` as the main function). So helper functions can be reused more easily and not much refactoring if we want to add some new features.
 
That's said, I'm happy to change to  `protoreflect.Message` as well, if  `protoreflect.Message` is the preferred args in those helper functions.

mockgcp/common/fields/updatemask.go
yuwenma: ~Hmmm, do you mean the readability is better? there's no difference in logic~ returned err at the end. Thanks!

mockgcp/common/fields/updatemask.go
yuwenma: Good catch! Yes, I think the GCP Patch for `Map` normally means a replace rather than an update.

mockgcp/common/fields/updatemask.go
yuwenma: ok. Yeah, now the logic has a difference.

mockgcp/common/fields/updatemask.go
yuwenma: Thank you, this is very helpful. I used two functions, and renamed `old`, `new` to `original` and `update` to distinguish.

justinsb (approved):

justinsb (approved):

PR #3073, opened 2024-11-01
feat: A solution for SecretRef (BasicAuthRef )

yuwenma:
For sensitive data in direct resource API, we want to reference a k8s Secret. 

This PR is diff base from #3042.


pkg/controller/direct/bigqueryconnection/connection_mapping.go
```diff
@@ -289,3 +288,26 @@ func BigQueryConnectionConnectionSpec_ToProto(mapCtx *direct.MapContext, in *krm
...
 	return out
 }
+
+func CloudSqlCredential_FromProto(mapCtx *direct.MapContext, in *pb.CloudSqlCredential) *krm.CloudSqlCredential {
+	if in == nil {
+		return nil
+	}
+	out := &krm.CloudSqlCredential{}
+	// This is sensitive data, input-only field.
+	// out.Username = direct.LazyPtr(in.GetUsername())
```
justinsb, line 298: Does this cause the round-trip tests to fail?

pkg/test/resourcefixture/testdata/basic/bigqueryconnection/v1alpha1/bigqueryconnectionconnection/awsconnectionbasic/_http.log
```diff
@@ -93,8 +94,7 @@ X-Xss-Protection: 0
 {
   "aws": {
     "accessRole": {
-      "iamRoleId": "arn:aws:iam::199999999999:role/basicrole",
-      "identity": "048077221682493034546"
```
justinsb, line 97: What happened to this field?

justinsb:
Some suggestions but I don't think they are blocking (they don't impact external behaviour).

/approve
/lgtm

/hold for suggestions

pkg/controller/direct/bigqueryconnection/connection_controller.go
```diff
@@ -76,24 +76,49 @@ func (m *model) AdapterForObject(ctx context.Context, reader client.Reader, u *u
...
+	return &Adapter{
+		id:        connectionRef,
+		gcpClient: gcpClient,
+		desired:   obj,
+		reader:    reader,
+		namespace: obj.Namespace,
+	}, nil
+}
 
+func (a *Adapter) NormalizeReference(ctx context.Context) error {
```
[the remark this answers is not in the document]
yuwenma: +1. "Maybe it should be part of the interface though," Before it happens, let me change this to private. Thanks!

apis/refs/v1beta1/secret/basicauth.go
```diff
@@ -0,0 +1,60 @@
...
+
+type BasicAuthSecret struct {
+	// +required
+	// The `metadata.name` field of a Kubernetes `Secret`
+	Name string `json:"name,omitempty"`
+	// The `metadata.namespace` field of a Kubernetes `Secret`.
+	Namespace string `json:"namespace,omitempty"`
+
+	// The public field with json:"-" tag is to skip the field
+	// in the CRD, and bypass "the unexported field error"
```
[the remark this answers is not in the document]
yuwenma: Agree to not export (that's also what I did at the beginning). But when this object is referenced and parsed in `runtime.DefaultUnstructuredConverter.FromUnstructured`, it panics. Here's the related k/k issue.  
Here's the related issue https://github.com/kubernetes/kubernetes/issues/124154

pkg/controller/direct/bigqueryconnection/connection_mapping.go
yuwenma: We don't have roundtrip test for this resource yet. I think we cannot not map input-only field back to KRM any ways, can we?

pkg/test/resourcefixture/testdata/basic/bigqueryconnection/v1alpha1/bigqueryconnectionconnection/awsconnectionbasic/_http.log
yuwenma: This PR based on a pending PR. #3042 (implements the update field mask in MockGCP)
I'll rebase once that other one merges. This should go away

apis/refs/v1beta1/secret/basicauth.go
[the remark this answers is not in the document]
justinsb: Ah - good to know.  Disappointing that we haven't fixed that!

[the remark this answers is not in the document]
justinsb: I guess we can image that we _might_ let users set the raw values directly, and then having these fields makes sense.

We don't want to let users set the raw values, because we want to nudge them to using a secret.  But if we get push-back, we could change our minds here and allow the values to be set via the json tag.  And then if we did that, this structure and normalization _would_ make sense.

pkg/controller/direct/bigqueryconnection/connection_mapping.go
justinsb: We'll have to figure out to do here.  We'll probably just have to skip the field.  That's acceptable I think.

pkg/test/resourcefixture/testdata/basic/bigqueryconnection/v1alpha1/bigqueryconnectionconnection/cloudsqlconnectionbasic/_http.log
```diff
@@ -778,7 +778,7 @@ X-Xss-Protection: 0
       "instance": "sqlinstance-sample-${uniqueId}",
       "kind": "sql#user",
       "name": "sqluser-${uniqueId}",
-      "password": "password",
+      "password": "cGFzc3dvcmQ=",
```
justinsb, line 781: Uhoh ... should the password be provided in plaintext or base64?

Edit: nevermind, I see why, but I suggest using a password that doesn't cause confusion (like "password" or "secret")

pkg/test/resourcefixture/testdata/basic/bigqueryconnection/v1alpha1/bigqueryconnectionconnection/cloudsqlconnectionbasic/dependencies.yaml
```diff
@@ -11,12 +11,13 @@
...
-
 apiVersion: v1
 kind: Secret
 metadata:
   name: secret-${uniqueId}
-data:
+type: kubernetes.io/basic-auth
+stringData:
+  username: sqluser-${uniqueId}
   password: cGFzc3dvcmQ=
```
justinsb, line 21: Nit: maybe make this `password: password` because otherwise people (OK, me!) will think you are forgetting to base64 the secret :-)

justinsb (approved):

justinsb:
BTW I added the hold but looks like you've looked at the issues, so I think this is good to go @yuwenma

yuwenma:
/hold cancel

Thank you for the reply @justinsb I'll resolve issues in the follow up PRs

PR #3074, opened 2024-11-01
docs: Update 5-releases.md

acpana:
Some more link anchoring


maqiuyujoyce:
/lgtm
/approve

I wondered if we'd prefer absolute paths here if the doc structure is updated again...

PR #3079, opened 2024-11-02
Refactor CloudNetworkRef to use Ref / ID split pattern

justinsb:
- **ComputeNetworkRef: split into ref and ID**
- **mockgcp: CloudBuildWorkerPool: normalize peered network link to use project number**

### DIFF
--- a/apis/refs/v1beta1/computerefs.go
+++ b/apis/refs/v1beta1/computerefs.go
@@ -17,8 +17,11 @@ package v1beta1
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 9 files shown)


justinsb:
WDYT @yuwenma ?  We do have to call the project lookup fairly often, but I think we can add some caching here (I don't think a project ID can be reused within a few days, so we could easily cache with a 1 hour TTL)

I proposed turning on e2e for cloudbuild so we can capture some real-world logs though and make sure we our simulations are accurate!  That is #3081

PR #3080, opened 2024-11-02
feat: Check if CRD contains sensitive data

yuwenma:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/tests/apichecks/crds_test.go
+++ b/tests/apichecks/crds_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/crd/crdloader"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test"
 	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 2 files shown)


tests/apichecks/crds_test.go
```diff
@@ -233,6 +233,93 @@ func TestCRDsAcronyms(t *testing.T) {
...
+			if len(skepticalfieldPaths) == 0 {
+				continue
+			}
+			if len(skepticalfieldPaths) == 1 {
+				if filter(skepticalfieldPaths[0]) {
+					errs = append(errs, fmt.Sprintf("crd=%s version=%v: field %q is sensitive data, should use secretRef", crd.Name, version.Name, skepticalfieldPaths[0]))
+				}
+				continue
+			}
+			sort.Strings(skepticalfieldPaths)
```
justinsb, line 286: I think using a set is probably easier?

tests/apichecks/crds_test.go
```diff
@@ -233,6 +233,93 @@ func TestCRDsAcronyms(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("error loading crds: %v", err)
+	}
+
+	var errs []string
+
+	sensitiveKeywords := []string{
+		"password",
+	}
+	filter := func(fieldPath string) bool {
```
jingyih, line 248: nit: recommend using a function name more specific to its functionality, something like "hasSensitiveSuffix" or "endWithSensitiveKeywords"?

tests/apichecks/crds_test.go
```diff
@@ -233,6 +233,93 @@ func TestCRDsAcronyms(t *testing.T) {
...
+		isSensitive := false
+		fieldPath = strings.ToLower(fieldPath)
+		for _, sensitiveWord := range sensitiveKeywords {
+			if strings.HasSuffix(fieldPath, sensitiveWord) {
+				isSensitive = true
```
jingyih, line 253: nit: can just `return true` here. (no need for "isSensitive")

jingyih (commented):
/lgtm

tests/apichecks/crds_test.go
yuwenma: Used `sets.Set` as Justin suggested. It looks much nicer. Could you take another look.

tests/apichecks/crds_test.go
yuwenma: Thank you very much! I really like the new code style!

jingyih (commented):
/lgtm

tests/apichecks/crds_test.go
```diff
@@ -233,6 +234,56 @@ func TestCRDsAcronyms(t *testing.T) {
...
+// Avoid passing sensitive data as plain text in the CRD
+func TestNoSensitiveField(t *testing.T) {
+	crds, err := crdloader.LoadAllCRDs()
+	if err != nil {
+		t.Fatalf("error loading crds: %v", err)
+	}
+
+	var errs []string
+
+	sensitiveKeywords := []string{
```
cheftako, line 246: Might be worth pulling this up to a constant list at the top rather than embedding inside a function at line 246. This seems like something we are going to want to easily grow as we find other field names which suggest a problem. (passphrase, key, private, pemCertificate, ....)

PR #3087, opened 2024-11-04
mockgcp: support int32 in update_mask

justinsb:
It just wasn't in the initial list.


yuwenma (commented):
/lgtm
/approve
Thanks

PR #3091, opened 2024-11-04
feat: Add docs for enabling direct resources

jasonvigil:


docs/features/optin.md
```diff
@@ -0,0 +1,56 @@
...
+> `Feature State`: `alpha` as of version v1.125
+
+KCC can be configured to enable alpha version of controllers (referred to as
+"direct" controllers) for reconciling specific resources. This will allow users
+to test the new direct controller code for bug fixes or features not available
+in legacy controllers. It is possible to enable the direct controller code on a
+per-resource basis, by setting an annotation for each custom resource.
+
+
+## Enabling
```
yuwenma, line 12: Could you also explain what resource can apply this annotation (has both TF/DCL and direct supports) ?

docs/features/optin.md
```diff
@@ -0,0 +1,56 @@
+# Opt in to Use Alpha Version Controllers
+
+> `Feature State`: `alpha` as of version v1.125
```
yuwenma, line 3: IIRC, this feature is enabled in 1.124.0. Could you double check?

docs/features/optin.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Enabling
+
+To enable the direct controller for a specific resource, update the resource and
+specify the annotation:
+
+```yaml
+metadata:
+  annotations:
+    alpha.cnrm.cloud.google.com/reconciler: "direct"
```
yuwenma, line 20: One question users may have is that if this annotation accept other values, like "terraform" or "dcl". Could you add some explanation to that?

docs/features/optin.md
jasonvigil: Would prefer to specify 1.125 here, so the user does not enable direct controller for sqlinstance on 1.124 (which does not have fixes from this last iteratino).

yuwenma (commented):
/approve

docs/features/optin.md
jasonvigil: ook. updated

docs/features/optin.md
jasonvigil: Ok, added a section "Applicability" for this

docs/features/optin.md
yuwenma: It worths noting that we already revealed this annotation in [v1.123.1](https://github.com/GoogleCloudPlatform/k8s-config-connector/releases/tag/v1.123.1) for DataflowFlexTemplateJob resource. So this mismatch could confuse users.

docs/features/optin.md
```diff
@@ -0,0 +1,104 @@
...
+is only available in the direct controller.
+
+If you're unsure if a particular resource has an experimental direct controller
+available and want to try out the new code wherever possible, read on.
+
+To find out if an experimental direct controller is available for a resource,
+check in the `apis/` directory for the service `*_types.go` file. If the file
+specifies either of the following legacy controller labels, then it is likely
+possible to enable the experimental direct controller for resources of that
+kind.
```
yuwenma, line 80: We want to suggest users to check their installed CRDs in the clusters. Because they cannot find the correct CRD version under `./apis` by git -tag (it is broken).

docs/features/optin.md
```diff
@@ -0,0 +1,104 @@
...
+logs will show "starting reconcile" for the resource instead of "Running reconcile".
+
+> {"severity":"info","timestamp":"2024-10-30T23:47:49.726Z","logger":"sqlinstance-controller","msg":"starting reconcile","resource":{"name":"my-sqlinstance","namespace":"default"}}
+
+
+## Applicability
+
+We recommend only enabling the experimental direct controller if it is
+necessary to work around a legacy controller bug or enable a new feature that
+is only available in the direct controller.
```
yuwenma, line 71: I like this statement. Can we move it to the top before "enabling", in case users turn it on unexpected before seeing the statement.

docs/features/optin.md
jasonvigil: Ok, I updated it to say v1.123.1. I am still worried that the users with sqlinstance issues will interpret this to mean their problems will be solved by enabling the direct controller annotation for sqlinstance resources in v1.123.1 / v1.124.0 (which have known bugs for sqlinstance direct controller).

docs/features/optin.md
```diff
@@ -0,0 +1,104 @@
+# Opt in to Enable Experimental Direct Controllers
+
+> `Feature State`: `alpha` as of version v1.123.1
+
+KCC can be configured to enable experimental versions of direct controllers for
```
acpana, line 5:
```suggestion
Config Connector can be configured to enable experimental versions of direct controllers for
```

docs/features/optin.md
```diff
@@ -0,0 +1,104 @@
...
+annotation will have an experimental direct controller available. This is
+because not all resources in the apis/ directory have a direct controller
+implemented yet. To confirm if there is an experimental direct controller
+available for one of the types in the apis/ dir, check to make sure there is
+an implementation of the controller for that type in `pkg/controller/direct`.
+The controllers are found in `*_controller.go` [[example]](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/sql/sqlinstance_controller.go).
+Also, verify that the direct controller is registered in `pkg/controller/direct/register/register.go`
+[[example]](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/f485d486553f1fc939eb8e487ccabaf2f2f032ed/pkg/controller/direct/register/register.go#L43).
+
+At this point, if there is both a type in the apis/ dir with a legacy controller
```
acpana, line 100: Do we need this paragraph? To my mind, the common case is that the direct ctrl is enabled if we are a type in the apis dir for a resource or will soon be enabled as the resource is being converted.

docs/features/optin.md
jasonvigil: Curious, why is this change suggestion necessary?

docs/features/optin.md
yuwenma: Understood. How about we give a more dedicated, per-resource release version guide, like sqlinstance 1.125, dataflowflexTempalteJob 1.123.1, ...and we can also get rid of the long explanation about how to find the right versions to apply this labels.

docs/features/optin.md
jasonvigil: I just remembered -- this is our guidance for product naming (as mentioned by @maqiuyujoyce in a previous PR) https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2393#discussion_r1704740437

docs/features/optin.md
jasonvigil: After modifying the wording based on @yuwenma's suggestion in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3091#discussion_r1828556840, it's definitely necessary to include this. Because now we are checking the CRDs (not apis/ dir), there are many CRDs with the legacy label that do not yet have a direct controller implemented.

docs/features/optin.md
jasonvigil: Done

docs/features/optin.md
jasonvigil: Ok, updated to instruct users to check CRDs (not `./apis`)

docs/features/optin.md
jasonvigil: Fixed

docs/features/optin.md
jasonvigil: That could work also. Though, I'd rather not have to maintain such a list. Hopefully we will soon switch over to direct only controller for dataflowflextemplatejob / sqlinstance. At that point, we'd have to update this doc again.

jasonvigil: Maybe the concern in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3091#discussion_r1828560321 is ok -- it is experimental code after all.

docs/features/optin.md
```diff
@@ -0,0 +1,106 @@
+# Opt in to Enable Experimental Direct Controllers
```
maqiuyujoyce, line 1: Thank you so much for adding the doc, @acpana ! It's fairly comprehensive and actionable.

I think this (though a less technical version) should also be on our public doc if we plan to advertise this feature. @yuwenma do we plan to do that?

docs/features/optin.md
```diff
@@ -0,0 +1,106 @@
...
+  databaseVersion: POSTGRES_16
+  region: us-central1
+  settings:
+    tier: db-custom-1-3840
+```
+
+The only supported value for the annotation is "direct", to enable the
+direct controller.
+
+It's worth noting that it's (relatively) harmless to add the annotation to enable
```
cheftako, line 46: Not sure why we want to suggest adding the annotation for resources which don't yet support it. It seems like we are suggesting a mechanism to lay a land mine. 6 months later this seems likely to cause unexpected behavior when that controller is created and everyone has forgotten that a particular record had this annotation added.

cheftako: Maybe something more like "While adding an annotation to enable experimental direct controllers for resources that do not yet support experimental direct controllers should be harmless, we recommend against doing so. Adding such an annotation can be confusing and surprising when the functionality is added later."

docs/features/optin.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Applicability
+
+If you're unsure if a particular resource has an experimental direct controller
+available and want to try out the new code wherever possible (or if you're just
+curious to learn more about the internals of this feature in Config Connector),
+read on.
+
+To find out if an experimental direct controller is available for a resource,
+check CRD for the resource. If the CRD specifies either of the following legacy
```
cheftako, line 84: Would be good to give instructions for checking the CRD. We can suggest either something like `kubectl get crd alloydbinstances.alloydb.cnrm.cloud.google.com -ojson | jq '.metadata.labels' ` OR inspecting https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_alloydbinstances.alloydb.cnrm.cloud.google.com.yaml.

docs/features/optin.md
jasonvigil: Good point, fixed!

docs/features/optin.md
jasonvigil: Alright, added some explicit instructions for checking the CRDs.

PR #3093, opened 2024-11-05
Configure pprof on controller-manager.

cheftako:
Generated file.
Adding cluster mode changes.
Updated version for cluster mode.
Skipping version check on dev mode controllers.
Make pprof support case insensitive.

### Change description

Allows you to enable pprof and use it as follows.

apiVersion: customize.core.cnrm.cloud.google.com/v1alpha1
kind: NamespacedControllerReconciler
metadata:
  name: cnrm-controller-manager 
  namespace: kcc-test
spec:
  pprof:
    support: all
    port: 1234

kubectl port-forward cnrm-controller-manager-<id> -n cnrm-system 1234

curl http://localhost:1234/debug/pprof/heap > heap.outheap.out

Fixes #

### Tests you have done

### DIFF
--- a/operator/config/crd/bases/customize.core.cnrm.cloud.google.com_controllerreconcilers.yaml
+++ b/operator/config/crd/bases/customize.core.cnrm.cloud.google.com_controllerreconcilers.yaml
@@ -41,6 +41,20 @@ spec:
           spec:
             description: ControllerReconcilerSpec is the specification of ControllerReconciler.
             properties:
+              pprof:
+                description: Configures the debug endpoint on the service.
+                properties:
+                  port:
+                    description: The port that the pprof server binds to if enabled
+                    type: integer
+                  support:
+                    description: Control if pprof should be turned on and which types
+                      should be enabled.
+                    enum:
+                    - none
+                    - all
+                    type: string
+                type: object
               rateLimit:
                 description: |-
                   RateLimit configures the token bucket rate limit to the kubernetes client used
--- a/operator/config/crd/bases/customize.core.cnrm.cloud.google.com_namespacedcontrollerreconcilers.yaml
+++ b/operator/config/crd/bases/customize.core.cnrm.cloud.google.com_namespacedcontrollerreconcilers.yaml
@@ -41,6 +41,20 @@ spec:
           spec:
             description: NamespacedControllerReconciler is the specification of NamespacedControllerReconciler.
             properties:
+              pprof:
+                description: Configures the debug endpoint on the service.
+                properties:
+                  port:
+                    description: The port that the pprof server binds to if enabled
+                    type: integer
+                  support:
+                    description: Control if pprof should be turned on and which types
+                      should be enabled.
+                    enum:
+                    - none
+                    - all
+                    type: string
+                type: object
               rateLimit:
                 description: |-
                   RateLimit configures the token bucket rate limit to the kubernetes client used
--- a/operator/pkg/apis/core/customize/v1alpha1/controllerreconciler_types.go
+++ b/operator/pkg/apis/core/customize/v1alpha1/controllerreconciler_types.go
@@ -41,6 +41,9 @@ type NamespacedControllerReconcilerSpec struct {
 	// If not specified, the default will be Token Bucket with qps 20, burst 30.
 	// +optional
 	RateLimit *RateLimit `json:"rateLimit,omitempty"`
+	// Configures the debug endpoint on the service.
+	// +optional
+	Pprof *PprofConfig `json:"pprof,omitempty"`
 }
 
 type RateLimit struct {
@@ -50,6 +53,16 @@ type RateLimit struct {
 	// The burst of the token bucket rate limit for all the requests to the kubernetes client.
 	// +optional
 	Burst int `json:"burst,omitempty"`
+}
+
+type PprofConfig struct {
+	// Control if pprof should be turned on and which types should be enabled.
+	// +kubebuilder:validation:Enum=none;all
+	// +optional
+	Support string `json:"support,omitempty"`
+	// The port that the pprof server binds to if enabled
+	// +optional
+	Port int `json:"port,omitempty"`
 }
 
 // NamespacedControllerReconcilerStatus defines the observed state of NamespacedControllerReconciler.
@@ -92,6 +105,9 @@ type ControllerReconcilerSpec struct {
 	// If not specified, the default will be Token Bucket with qps 20, burst 30.
 	// +optional
 	RateLimit *RateLimit `json:"rateLimit,omitempty"`
+	// Configures the debug endpoint on the service.
+	// +optional
+	Pprof *PprofConfig `json:"pprof,omitempty"`
 }
 
 // ControllerReconcilerStatus defines the observed state of ControllerReconciler.
@@ -113,6 +129,10 @@ type ControllerReconcilerList struct {
 }
 
 var ValidRateLimitControllers = []string{
+	"cnrm-controller-manager",
+}
+
+var SupportedPprofControllers = []string{
 	"cnrm-controller-manager",
 }
 
--- a/operator/pkg/controllers/utils.go
+++ b/operator/pkg/controllers/utils.go
@@ -530,6 +530,7 @@ func ApplyContainerRateLimit(m *manifest.Objects, targetControllerName string, r
 			targetControllerName, strings.Join(customizev1alpha1.ValidRateLimitControllers, ", "))
 	}
 
+	count := 0
 	for _, item := range m.Items {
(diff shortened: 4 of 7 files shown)


go.work
```diff
@@ -3,8 +3,8 @@ go 1.22.5
 use (
 	.
 	./dev/tools/controllerbuilder
-	./experiments/kompanion
-	./experiments/kubectl-plan
+//	./experiments/kompanion
```
justinsb, line 6: Looks unrelated - split into another PR maybe?

operator/config/crd/bases/customize.core.cnrm.cloud.google.com_controllerreconcilers.yaml
```diff
@@ -41,6 +41,20 @@ spec:
           spec:
             description: ControllerReconcilerSpec is the specification of ControllerReconciler.
             properties:
+              pprofConfig:
```
justinsb, line 44: Naming nits: it's all config, so I would suggest `pprofConfig` -> `pprof` or `profiling`

And then if we use `pprof`, `pprofPort` can just be `port`, and `pprofSupport` can just be `support` (or `supportedModes`, I originally thought pprofSupport was really `enabled`

If we do `profiling`, `pprofPort` and `pprofSupport` (or `pprofSupportedModes` or `pprofModes`) make sense, because then we don't have pprof duplicated (and we allow for other means of profiling in future)

justinsb: BTW I also wonder whether otel might be more useful than pprof.  pprof might be useful for tracking down memory usage, and might be useful for tracking down CPU usage, but I don't _think_ either of those are our biggest problems?  Not a blocker to adding pprof support first though :-)

operator/config/crd/bases/customize.core.cnrm.cloud.google.com_namespacedcontrollerreconcilers.yaml
```diff
@@ -41,6 +41,20 @@ spec:
           spec:
             description: NamespacedControllerReconciler is the specification of NamespacedControllerReconciler.
             properties:
+              pprofConfig:
+                description: PprofConfigures the debug endpoint on the service.
```
justinsb, line 45: Also nit, should the comment should probably be "Configures the debug endpoint on the service" or "<fieldName> configures the debug endpoint on the service"

operator/pkg/controllers/utils.go
```diff
@@ -586,3 +586,86 @@ func applyRateLimitToContainerArg(container map[string]interface{}, rateLimit *c
...
+		return fmt.Errorf("pprof config customization for %s is not supported. "+
+			"Supported controllers: %s",
+			targetControllerName, strings.Join(customizev1alpha1.SupportedPprofControllers, ", "))
+	}
+
+	for _, item := range m.Items {
+		if item.GroupVersionKind() != targetControllerGVK {
+			continue
+		}
+		if !strings.HasPrefix(item.GetName(), targetControllerName) {
```
justinsb, line 617: It doesn't matter, but ... why the prefix match?

operator/pkg/controllers/utils.go
```diff
@@ -586,3 +586,86 @@ func applyRateLimitToContainerArg(container map[string]interface{}, rateLimit *c
...
+			continue
+		}
+		if err := item.MutateContainers(customizePprofConfigFn(targetContainerName, pprofConfig)); err != nil {
+			return err
+		}
+		break // we already found the matching controller, no need to keep looking.
```
justinsb, line 623: My 2c: the small savings in CPU are not worth breaking.  I think it's better to return nil here and making it an error if we don't find a match (i.e. return error at the end of the function)

Another option is to count the number of containers changed and return an error if there was not exactly one changed.

operator/pkg/controllers/utils.go
```diff
@@ -586,3 +586,86 @@ func applyRateLimitToContainerArg(container map[string]interface{}, rateLimit *c
...
+
+func applyPprofConfigToContainerArg(container map[string]interface{}, pprofConfig *customizev1alpha1.PprofConfig) error {
+	if pprofConfig == nil {
+		return nil
+	}
+	origArgs, found, err := unstructured.NestedStringSlice(container, "args")
+	if err != nil {
+		return fmt.Errorf("error getting args in container: %w", err)
+	}
+	wantArgs := []string{}
```
justinsb, line 649: LGTM.  Somewhere we have a similar function to replace flags, but I don't know which codebase that is in!

justinsb:
Looks good, my only real quibble was with the field naming

operator/config/crd/bases/customize.core.cnrm.cloud.google.com_controllerreconcilers.yaml
cheftako: This is really last mile wiring. We already have pprof capability in the binaries themselves. We were just missing a way to turn it on. All for adding otel but this seemed like a quick win.

operator/pkg/apis/core/customize/v1alpha1/controllerreconciler_types.go
```diff
@@ -41,6 +41,9 @@ type NamespacedControllerReconcilerSpec struct {
 	// If not specified, the default will be Token Bucket with qps 20, burst 30.
 	// +optional
 	RateLimit *RateLimit `json:"rateLimit,omitempty"`
+	// PprofConfigures the debug endpoint on the service.
+	// +optional
+	PprofConfig *PprofConfig `json:"pprofConfig,omitempty"`
```
acpana, line 46: you can probably argue YAGNI against this but how about we sub class the Pprof config under something like `debugConfig` and we can add more debugging controls like otel, etc.

operator/config/crd/bases/customize.core.cnrm.cloud.google.com_controllerreconcilers.yaml
cheftako: Also this was inspired by the LRO resources where the controller went silent. Figured this would be useful for confirming that the controller was stack waiting on the LRO and not that something else was wrong.

operator/pkg/controllers/utils.go
cheftako: I believe the manifest name here is for the Pod(s) and would be something like "cnrm-controller-manager-okwcikx5o4yw4ly-0". 
However that needs to match the name of the kind: NamespacedControllerReconciler which in this case is just "cnrm-controller-manager".
Might be better to use the cnrm.cloud.google.com/component annotation but I would rather make that a broad fix across reconcilers.

operator/pkg/controllers/utils.go
cheftako: Going with 2 but also changing ApplyContainerRateLimit to match.

go.work
cheftako: I think I was having a problem with the linter or make command.. Let me see what happens if I just drop this change.

cheftako: Fixed for now :D

operator/pkg/apis/core/customize/v1alpha1/controllerreconciler_types.go
cheftako: My personal take is if we think we will have many of these (say more than 5) then this organization would make a lot of sense.
Right now we have 1 and we'll see if we feel the need for the second.
It its just 2 it won't hurt but I'm not convinced its worth the effort of changing.
I may come to regret that.

jingyih: Since this is an alpha API, we can always make improvements in the future if needed.

operator/config/crd/bases/customize.core.cnrm.cloud.google.com_controllerreconcilers.yaml
```diff
@@ -41,6 +41,20 @@ spec:
           spec:
             description: ControllerReconcilerSpec is the specification of ControllerReconciler.
             properties:
+              pprof:
+                description: Configures the debug endpoint on the service.
+                properties:
+                  pprofPort:
```
yuwenma, line 47: nit: `pprof` --> `port`

yuwenma (commented):
/lgtm
/approve
/hold 

hold in case you want to rename the pprof naem. Feel free to unhold

yuwenma (commented):
/lgtm

operator/pkg/controllers/utils.go
justinsb: Ah, makes sense, thank you!  It shouldn't have the `-0`, but I had forgotten about the "namespace suffix"

justinsb (approved):

PR #3096, opened 2024-11-05
feat: add gemini prompting to controllerbuilder

justinsb:
Splitting off from the CSV PR ( #2276 ) so that we can get this onto the main branch


yuwenma (commented):
/lgtm
/approve

+1000. Was trying to ask if we can merge that PR.

PR #3101, opened 2024-11-05
tests: fuzz DataExchange spec

acpana:
Adds a fuzz test for DataExchange.


acpana:
/hold

this is the old way of doing the fuzzers; I should use the new way!

acpana:
Took a different approach in #3119

PR #3106, opened 2024-11-05
fix: Use actual sqlinstance location preference if not specified

jasonvigil:
Otherwise, the controller will always detect a diff and try to update on each re-reconcile.


yuwenma (commented):
/lgtm
/approve

PR #3107, opened 2024-11-05
chore: remove a sample for bigquerydatatransferconfig

jingyih:
We will likely introduce backward-compatible changes to the API to improve the handling of sensitive fields in `BigQueryDataTransferConfig`. In the meantime, let's remove the "salesforce" sample for this resource, as we don't want to promote usage where users need to include plain credentials in their YAML configurations.


yuwenma:
/lgtm
/approve 
Love it! Thanks

PR #3113, opened 2024-11-06
fix: bqcc aws identity output-only and required in update

yuwenma:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/pkg/controller/direct/bigqueryconnection/connection_controller.go
+++ b/pkg/controller/direct/bigqueryconnection/connection_controller.go
@@ -21,6 +21,7 @@ import (
 
 	gcp "cloud.google.com/go/bigquery/connection/apiv1"
 	bigqueryconnectionpb "cloud.google.com/go/bigquery/connection/apiv1/connectionpb"
+	pb "cloud.google.com/go/bigquery/connection/apiv1/connectionpb"
 	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/bigqueryconnection/v1beta1"
 	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
 	refsv1beta1secret "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1/secret"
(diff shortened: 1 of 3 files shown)


pkg/controller/direct/bigqueryconnection/connection_controller.go
```diff
@@ -242,8 +254,9 @@ func (a *Adapter) Update(ctx context.Context, updateOp *directbase.UpdateOperati
 		return err
 	}
 	mapCtx := &direct.MapContext{}
-	desired := a.desired.DeepCopy()
-	connection := BigQueryConnectionConnectionSpec_ToProto(mapCtx, &desired.Spec)
+	connection := BigQueryConnectionConnectionSpec_ToProto(mapCtx, &a.desired.Spec)
+	connection = merge3(connection, a.desired.Status.ObservedState)
```
yuwenma, line 258: This is a quick fix for GCP error "Error 400: Only 'aws.access_role.iam_role_id' field can be updated." where the `iam_role_id` is a output-only field, that the `CompareProtoMessage` considers the field needs to be changed.

I think the final solution should be a 3 way merge among the desired (`.spec`), previously applied (`.status.observedState`) and the actual.

pkg/controller/direct/bigqueryconnection/connection_controller.go
```diff
@@ -232,6 +233,17 @@ func (a *Adapter) Create(ctx context.Context, createOp *directbase.CreateOperati
 	return setStatus(u, status)
 }
 
+// aws.accessRole.Identity is a output-only field in CREATE, it is required in the UPDATE.
+func merge3(desired *pb.Connection, previouslyApplied *krm.BigQueryConnectionConnectionObservedState) *pb.Connection {
+	if previouslyApplied != nil {
+		if previouslyApplied.Aws != nil {
```
jingyih, line 239: nit: `if previouslyApplied != nil && previouslyApplied.Aws != nil {`

jingyih (commented):
/lgtm

pkg/controller/direct/bigqueryconnection/connection_controller.go
yuwenma: Sure, done.
(I was thinking we may want to deal with other property merge, like Azure, SQL, Spanner, Spark, Salesforce)

jingyih (commented):
/lgtm

pkg/controller/direct/bigqueryconnection/connection_controller.go
```diff
@@ -21,6 +21,7 @@ import (
 
 	gcp "cloud.google.com/go/bigquery/connection/apiv1"
 	bigqueryconnectionpb "cloud.google.com/go/bigquery/connection/apiv1/connectionpb"
+	pb "cloud.google.com/go/bigquery/connection/apiv1/connectionpb"
```
cheftako, line 24: This seems redundant with line 23. Why do we want 2 aliases for this package?

pkg/controller/direct/bigqueryconnection/connection_controller.go
```diff
@@ -253,6 +254,14 @@ func (a *Adapter) Create(ctx context.Context, createOp *directbase.CreateOperati
 	return setStatus(u, status)
 }
 
+// aws.accessRole.Identity is a output-only field in CREATE, it is required in the UPDATE.
+func merge3(desired *pb.Connection, previouslyApplied *krm.BigQueryConnectionConnectionObservedState) *pb.Connection {
```
cheftako, line 258: The function name seems confusing. Its not (currently) a 3 way merge. I would also expect something named  3 way merge to be a generic 3 way merge. This seems like a awsIdentityStateMerge() function. Can we rename it?

PR #3116, opened 2024-11-06
Release 1.125.0-rc.1

yuwenma:


operator/config/gke-addon/image_configmap.yaml
```diff
@@ -1,11 +1,6 @@
 apiVersion: v1
 data:
-  cnrm.controller: gcr.io/gke-release/cnrm/controller:7a86865
-  cnrm.deletiondefender: gcr.io/gke-release/cnrm/deletiondefender:7a86865
-  cnrm.recorder: gcr.io/gke-release/cnrm/recorder:7a86865
-  cnrm.unmanageddetector: gcr.io/gke-release/cnrm/unmanageddetector:7a86865
-  cnrm.webhook: gcr.io/gke-release/cnrm/webhook:7a86865
-  prom-to-sd: gcr.io/gke-release/prometheus-to-sd:v0.11.12-gke.11
+  cnrm.controller: gcr.io/gke-release/cnrm/controller:1.125.0-rc.1
```
yuwenma, line 3: @justinsb does this look right to you?

justinsb: I think so.  This is where we will eventually promote the image to, I believe!

justinsb: Oh I see - you mean that most of the images are missing ... something has gone wrong.  I'll try to pull and see if I can understand why!

PR #3117, opened 2024-11-07
docs: update release note with BigQueryDataset direct controller

jingyih:
I'm not sure if we should officially announce the direct reconciler for the `BigQueryDataset` resource at this stage. Announcing it early could provide valuable soak time to uncover potential bugs before we enable it by default.

cc @yuwenma @xiaoweim


yuwenma:
I understood and like the indentions. But we don't want to recommend users to try out this annotation right now mostly because we don't yet have a complete stories to pretend malicious behavior or misuse. @jasonvigil is working on the document about when and how we recommend this annotation. https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3091

jingyih:
Thanks! Jason's doc provides users with a pointer to see which direct controllers are available, so announcing it in the release notes may not be necessary.

/close

PR #3122, opened 2024-11-07
fix: don't include unmanageddetector in base

justinsb:
This causes it to be included twice, which is an error.

### DIFF
--- a/config/installbundle/base/kustomization.yaml
+++ b/config/installbundle/base/kustomization.yaml
@@ -22,5 +22,6 @@ resources:
   - ../components/roles
   - ../components/rolebindings
   - ../components/recorder
-  - ../components/unmanageddetector
   - ../components/webhook
+  # Note: unmanageddetector is only used in namespaced mode
+  #- ../components/unmanageddetector


justinsb:
/assign @yuwenma 

This is the one that blocks the new release process

config/installbundle/base/kustomization.yaml
```diff
@@ -22,5 +22,6 @@ resources:
   - ../components/roles
   - ../components/rolebindings
   - ../components/recorder
-  - ../components/unmanageddetector
-  - ../components/webhook
\ No newline at end of file
+  - ../components/webhook
+  # Note: unmanageddetector is only used in namespaced mode
+  #- ../components/unmanageddetector
```
yuwenma, line 27: I think we want to update [this code snippet](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/dev/tasks/build-release-bundle#L68) as well, otherwise the propose-build-tag won't work either.

yuwenma (commented):
/lgtm
/approve
/hold

One more finding. just in case you want to fix the other code in the same PR.

config/installbundle/base/kustomization.yaml
yuwenma: Never mind. I was confused (again) by the mix use of kustomize and bash on the same file.

yuwenma (commented):
/hold cancel

PR #3125, opened 2024-11-07
migrate bigquerydataset to new API

xiaoweim:
The [New BigQuery API](https://pkg.go.dev/cloud.google.com/go/bigquery#DatasetMetadata) does not support the field `isCaseSensitive`.

Update: Filed a feature request with the bigquery team and the field has been added. Reference PR: https://github.com/googleapis/google-cloud-go/pull/11216. Reference BigQuery version: https://github.com/googleapis/google-cloud-go/releases/tag/bigquery/v1.65.0


pkg/controller/direct/bigquerydataset/bigquerydataset_mappings.go
```diff
@@ -21,210 +21,309 @@
...
+func BigQueryDatasetStatus_FromProtoMetadata(mapCtx *direct.MapContext, in *pb.DatasetMetadata) *krm.BigQueryDatasetStatus {
 	if in == nil {
 		return nil
 	}
 	out := &krm.BigQueryDatasetStatus{}
-	out.Etag = direct.LazyPtr(in.Etag)
-	out.CreationTime = direct.LazyPtr(in.CreationTime)
-	out.LastModifiedTime = direct.LazyPtr(in.LastModifiedTime)
-	out.SelfLink = direct.LazyPtr(in.SelfLink)
-	out.ObservedState = &krm.BigQueryDatasetObservedState{Location: direct.LazyPtr(in.Location)}
```
jingyih, line 91: Why is `ObservedState` removed?

pkg/controller/direct/bigquerydataset/bigquerydataset_mappings.go
```diff
@@ -21,210 +21,309 @@
...
-	out.DefaultPartitionExpirationMs = direct.ValueOf(in.DefaultPartitionExpirationMs)
-	out.DefaultTableExpirationMs = direct.ValueOf(in.DefaultTableExpirationMs)
-	out.DefaultEncryptionConfiguration = EncryptionConfiguration_ToAPI(mapCtx, in.DefaultEncryptionConfiguration)
+	out.DefaultPartitionExpiration = time.Duration(direct.ValueOf(in.DefaultPartitionExpirationMs)) * time.Millisecond
+	out.DefaultTableExpiration = time.Duration(direct.ValueOf(in.DefaultTableExpirationMs)) * time.Millisecond
+	out.DefaultEncryptionConfig = EncryptionConfiguration_ToProto(mapCtx, in.DefaultEncryptionConfiguration)
 	out.Description = direct.ValueOf(in.Description)
-	out.FriendlyName = direct.ValueOf(in.FriendlyName)
-	out.DatasetReference = DatasetReference_ToAPI(mapCtx, in, name)
+	out.ExternalDatasetReference = DatasetSpec_ToExternalDatasetReference(mapCtx, in)
```
jingyih, line 51: If currently we don't have this field in KRM go types, maybe we just remove this line for now. When we add this field in the future, we need to add the Go types first.

pkg/controller/direct/bigquerydataset/bigquerydataset_mappings.go
```diff
@@ -21,210 +21,309 @@
...
-	out.IsCaseInsensitive = direct.LazyPtr(in.IsCaseInsensitive)
-	maxTime := strconv.FormatInt(in.MaxTimeTravelHours, 10)
-	out.MaxTimeTravelHours = direct.LazyPtr(maxTime)
+	maxTimeInHours := fmt.Sprintf("%v", in.MaxTimeTravel.Hours())
+	out.MaxTimeTravelHours = &maxTimeInHours
 	out.StorageBillingModel = direct.LazyPtr(in.StorageBillingModel)
 	return out
 }
-func BigQueryDatasetStatus_FromAPI(mapCtx *direct.MapContext, in *api.Dataset) *krm.BigQueryDatasetStatus {
+func BigQueryDatasetStatus_FromProtoMetadata(mapCtx *direct.MapContext, in *pb.DatasetMetadata) *krm.BigQueryDatasetStatus {
```
jingyih, line 84: nit: rename to `BigQueryDatasetStatus_FromProtoMetadata`. I understand the intention, but I don't think we have a pattern of suffixing the mapping function with part of the proto message name.

pkg/controller/direct/bigquerydataset/dataset_controller.go
```diff
@@ -253,15 +269,16 @@ func (a *Adapter) Update(ctx context.Context, updateOp *directbase.UpdateOperati
 	if desired.Access == nil || len(desired.Access) == 0 {
 		resource.Access = a.actual.Access
 	}
-	updateDatasetCall := a.gcpService.Datasets.Update(parent.ProjectID, datasetId, resource)
-	updated, err := updateDatasetCall.Do()
+	datasetMetadataToUpdate := BigQueryDatasetMetadata_ToMetadataToUpdate(mapCtx, resource)
```
jingyih, line 272: Not sure if it is possible/easier to convert directly from `krm.BigQueryDatasetSpec` to `pb.DatasetMetadataToUpdate` in "Update". I feel it might be easier to read and maintain in the future.

pkg/controller/direct/bigquerydataset/dataset_controller.go
```diff
@@ -303,9 +320,8 @@ func (a *Adapter) Delete(ctx context.Context, deleteOp *directbase.DeleteOperati
 	if err != nil {
 		return false, fmt.Errorf("failed to parse bigquery dataset full name, %w", err)
 	}
-	deleteDatasetCall := a.gcpService.Datasets.Delete(parent.ProjectID, datasetId)
-	err = deleteDatasetCall.Do()
-	if err != nil {
+	dsHandler := a.gcpService.DatasetInProject(parent.ProjectID, datasetId)
+	if err := dsHandler.Delete(ctx); err != nil {
```
jingyih, line 324: It seems the old behavior is equivalent to calling `DeleteWithContents`?

tests/e2e/unified_test.go
```diff
@@ -606,6 +606,11 @@ func runScenario(ctx context.Context, t *testing.T, testPause bool, fixture reso
 
 					// Specific to BigQuery
 					addSetStringReplacement(".access[].userByEmail", "user@google.com")
+					for _, event := range events {
+						if event.Request.Method == "PATCH" && strings.Contains(event.Request.URL, "bigquery.googleapis.com/bigquery/v2/") {
+							event.Request.ReplaceHeader("If-Match", "1111111")
```
jingyih, line 611: IIUC, it is more readable to set this value to "abcdef0123A=", which is the Etag from previous read.

pkg/controller/direct/bigquerydataset/bigquerydataset_mappings.go
```diff
@@ -21,210 +21,309 @@
...
-	out.DefaultEncryptionConfiguration = EncryptionConfiguration_ToAPI(mapCtx, in.DefaultEncryptionConfiguration)
+	out.DefaultPartitionExpiration = time.Duration(direct.ValueOf(in.DefaultPartitionExpirationMs)) * time.Millisecond
+	out.DefaultTableExpiration = time.Duration(direct.ValueOf(in.DefaultTableExpirationMs)) * time.Millisecond
+	out.DefaultEncryptionConfig = EncryptionConfiguration_ToProto(mapCtx, in.DefaultEncryptionConfiguration)
 	out.Description = direct.ValueOf(in.Description)
-	out.FriendlyName = direct.ValueOf(in.FriendlyName)
-	out.DatasetReference = DatasetReference_ToAPI(mapCtx, in, name)
+	out.ExternalDatasetReference = DatasetSpec_ToExternalDatasetReference(mapCtx, in)
 	out.Location = direct.ValueOf(in.Location)
-	out.IsCaseInsensitive = direct.ValueOf(in.IsCaseInsensitive)
```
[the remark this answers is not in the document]
xiaoweim: I can only find out the usage of the resource, but not specific field :(. I just a YAQS question to the bigquery team, hopefully I can get an answer. If this field is too important, we might have to hold off migrating to the recommended API.

pkg/controller/direct/bigquerydataset/bigquerydataset_mappings.go
xiaoweim: Ah it should be back after I rebased the PR.

pkg/controller/direct/bigquerydataset/dataset_controller.go
xiaoweim: This update does not support `updateMask`, so I would still need to go through all the non-immutable fields and update the clone of the `a.actual` object.

I could convert the `desired` object directly from `krm.BigQueryDatasetSpec` to `pb.DatasetMetadataToUpdate`. But I think the problem is that `a.actual` is of type `pb.DatasetMetadata`. After I updated the `a.actual` object with new values from the `desired` object, I would still need to convert it to `pb.DatasetMetadataToUpdate`.

pkg/controller/direct/bigquerydataset/dataset_controller.go
xiaoweim: You are right.

pkg/controller/direct/bigquerydataset/dataset_controller.go
```diff
@@ -25,9 +25,9 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
+	"github.com/huandu/go-clone"
```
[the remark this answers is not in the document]
xiaoweim: `bigquery.DatasetMetadata` is a go struct, and proto.Clone does not work on the resource. That's why I searched for another clone method.

I understand your concern, I could manually clone the actual object during update.

pkg/controller/direct/bigquerydataset/dataset_controller.go
```diff
@@ -99,11 +108,12 @@ func (m *model) AdapterForURL(ctx context.Context, url string) (directbase.Adapt
...
-	gcpService *api.Service
-	desired    *krm.BigQueryDataset
-	actual     *api.Dataset
-	reader     client.Reader
+	id                         *krm.BigQueryDatasetRef
+	gcpService                 *bigquery.Client
+	desired                    *krm.BigQueryDataset
+	actual                     *bigquery.DatasetMetadata
+	reader                     client.Reader
+	hasUpdateIsCaseInsensitive bool
```
jingyih, line 116: Not sure I follow, what does `hasUpdateIsCaseInsensitive` do?

xiaoweim: In go, explicitly set a bool value to false and not setting it (it defaults to false), there is not a method to tell if the difference. I added it here to tell if the update has explicitly set the value to false, therefore I should update this field. Or it only defaults to false, where I should not be updating it.

jingyih:
Github diff tool is not working correctly. The actual diff on http log:

[_http5.log](https://www.diffchecker.com/pUGT7j86/)
[_http6.log](https://www.diffchecker.com/cQf2JlOD/)
[_http7.log](https://www.diffchecker.com/bWl41j2u/)

jingyih (commented):
Sorry I forgot to send out my review...

pkg/controller/direct/bigquerydataset/dataset_controller.go
```diff
@@ -177,91 +195,96 @@ func (a *Adapter) Update(ctx context.Context, updateOp *directbase.UpdateOperati
...
+	datasetMetadataToUpdate := BigQueryDataset_ToMetadataToUpdate(mapCtx, resource, updateMask.Paths)
+	for k, v := range a.desired.GetObjectMeta().GetLabels() {
+		datasetMetadataToUpdate.SetLabel(k, v)
 	}
-	updateDatasetCall := a.gcpService.Datasets.Update(parent.ProjectID, datasetId, resource)
-	updated, err := updateDatasetCall.Do()
+	datasetMetadataToUpdate.SetLabel("managed-by-cnrm", "true")
+	// Call update
+	dsHandler := a.gcpService.DatasetInProject(parent.ProjectID, datasetId)
+	updated, err := dsHandler.Update(ctx, *datasetMetadataToUpdate, resource.ETag)
```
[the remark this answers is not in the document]
yuwenma: Good point. This etag usage (both KCC and GCP sides) is very skeptical.  I have several guesses

@xiaoweim Does the GCP service actually update the changed fields? I cannot find it in the log. Basically I'm looking for the case where a field is changed in KCC spec, and we pass in the old etag. I want to know after the Update finishes, if the GCP Get shows the field is updated.

What happens if you use a invalid etag or stale etag? Will it return errors or no-op?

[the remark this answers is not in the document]
jingyih: I think to match the original behavior, we just pass the empty string for etag for a "blind write" that will always succeed.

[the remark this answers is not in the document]
xiaoweim: Yes we can pass an empty string to do a blind write https://pkg.go.dev/cloud.google.com/go/bigquery#Dataset.Update. "To perform a read-modify-write that protects against intervening reads, set the etag argument to the DatasetMetadata.ETag field from the read"

pkg/controller/direct/bigquerydataset/dataset_controller.go
xiaoweim: Yes it worked, thank you!

tests/e2e/testdata/scenarios/fields/management/bigquery/dataset/set_unset/_object05.yaml
```diff
@@ -28,7 +28,6 @@ spec:
   maxTimeTravelHours: "72"
   projectRef:
     external: ${projectId}
-  resourceID: bigquerydataset${uniqueId}
```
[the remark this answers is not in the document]
jingyih: Ping on this issue again. This change could potentially break existing customers who depend on the `spec.resourceID` field.

jingyih:
This isn’t directly related to this PR but is relevant to the BigQueryDataset migration. I just realized that this resource has a special annotation, `cnrm.cloud.google.com/delete-contents-on-destroy` [1], which we should continue supporting in the direct controller. The corresponding REST API field and its description is here [2], and it seems we just need to call either `Delete` or `DeleteWithContents` in the BigQuery client library, depending on the annotation.

[1] https://cloud.google.com/config-connector/docs/reference/resource-docs/bigquery/bigquerydataset#annotations
[2] https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets/delete#query-parameters

xiaoweim:
> This isn’t directly related to this PR but is relevant to the BigQueryDataset migration. I just realized that this resource has a special annotation, `cnrm.cloud.google.com/delete-contents-on-destroy` [1], which we should continue supporting in the direct controller. The corresponding REST API field and its description is here [2], and it seems we just need to call either `Delete` or `DeleteWithContents` in the BigQuery client library, depending on the annotation.
> 
> [1] https://cloud.google.com/config-connector/docs/reference/resource-docs/bigquery/bigquerydataset#annotations [2] https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets/delete#query-parameters

Thanks for pointing it out! I added a separate commit to address this.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/bigquerydatasetaccessblock-direct/_http.log
```diff
@@ -151,8 +151,9 @@ X-Xss-Protection: 0
 
 ---
 
-PUT https://bigquery.googleapis.com/bigquery/v2/projects/${projectId}/datasets/${datasetID}?alt=json&prettyPrint=false
+PATCH https://bigquery.googleapis.com/bigquery/v2/projects/${projectId}/datasets/${datasetID}?alt=json&prettyPrint=false
 Content-Type: application/json
+If-Match: abcdef0123A=
```
jingyih, line 156: I am surprised that `If-Match` still exist here. According to [1], updating with an empty etag should not result in this line?

[1] https://github.com/googleapis/google-cloud-go/blob/f3483963999b7fb7b9ac6364f3a6392d5a27d392/bigquery/dataset.go#L447

xiaoweim: The `If-Match` still exist because the etag in the patch request header was normalized without checking if it exists.

jingyih (commented):
/lgtm

PR #3129, opened 2024-11-07
fix: mask SQLInstance deletion_policy field

justinsb:
It's only relevant during deletion, and causes spurious diffs
at other times.


justinsb:
Actually, I think I'll send this with a test.

/close

PR #3130, opened 2024-11-07
Change BigqueryDataset Location to Optional

xiaoweim:


apis/bigquery/v1beta1/dataset_types.go
```diff
@@ -136,11 +135,17 @@ type BigQueryDatasetStatus struct {
...
+	ObservedState *BigQueryDatasetObservedState `json:"observedState,omitempty"`
 }
 
-// BigQueryDatasetSpec defines the desired state of BigQueryDataset
+// BigQueryDatasetObservedState defines the desired state of BigQueryDataset
 // +kcc:proto=google.cloud.bigquery.v2.dataset
 type BigQueryDatasetObservedState struct {
+
+	// Optional. If the location is not specified in the spec, the GCP server defaults to a location and will be captured here.
+	Location *string `json:"location,omitempty"`
```
jingyih, line 148: Not sure we want to add location to observedState at this point. The default controller is still Terraform-based, and the direct controller doesn't use this field.

xiaoweim: Yes the default controller is still TF. I added this field with the intention to use it in our update call. The idea is that if `spec.Location` is not specified upon creation, I want to use the location value in the `observedState` during an UPDATE call and compared that with actual. I have updated the code logic.

pkg/controller/direct/bigquerydataset/dataset_controller.go
```diff
@@ -184,6 +184,20 @@ func (a *Adapter) Update(ctx context.Context, updateOp *directbase.UpdateOperati
 
 	resource := clone.Clone(a.actual).(*api.Dataset)
 
+	// If the location is not specified upon creation, merge the desired with the previously applied location defaulting.
+	presviouslyApplied, _, _ := unstructured.NestedMap(u.Object, "status")
```
jingyih, line 188: nit: does `unstructured.NestedString(u.Object, "status", "observedState", "location")` not working?

jingyih (commented):
/lgtm

apis/bigquery/v1beta1/dataset_reference.go
```diff
@@ -125,15 +124,15 @@ func NewBigQueryDatasetRef(ctx context.Context, reader client.Reader, obj *BigQu
 	if actualParent.ProjectID != projectID {
 		return nil, fmt.Errorf("spec.projectRef changed, expect %s, got %s", actualParent.ProjectID, projectID)
 	}
-	if actualParent.Location != valueOf(location) {
-		return nil, fmt.Errorf("spec.location changed, expect %s, got %s", actualParent.Location, valueOf(location))
-	}
+	// if actualParent.Location != valueOf(location) {
+	// 	return nil, fmt.Errorf("spec.location changed, expect %s, got %s", actualParent.Location, valueOf(location))
+	// }
```
yuwenma, line 129: nit: clean up

jingyih (commented):
/lgtm

PR #3134, opened 2024-11-07
Fix the unexpected diff in runservicemulticontainer test

maqiuyujoyce:
### Change description

Fixes the failed runservicemulticontainer test (b/375678794).

When getting the underlying Cloud Run service resource in runservicemulticontainer test, `launchStage` value was updated to `GA` in the GET response. (https://cloud.google.com/run/docs/reference/rest/v2/projects.locations.services) So during the reconciliation of the testNoChange step(s) in dynamic test, the value of `launchStage` is `GA` in the live state and is `BETA` in the desired state, and this caused an unexpected diff and triggers an unexpected API call.

This behavior comes from the underlying Cloud Run API so we can't really fix it. To resolve the failed test, the workaround is to match the `launchStage` in the test case with the `launchStage` returned by the Cloud Run API. 

### Tests you have done

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/run/v1beta1/runservice/runservicemulticontainer/create.yaml
+++ b/pkg/test/resourcefixture/testdata/basic/run/v1beta1/runservice/runservicemulticontainer/create.yaml
@@ -18,7 +18,7 @@ metadata:
   name: runservice-${uniqueId}
 spec:
   ingress: "INGRESS_TRAFFIC_ALL"
-  launchStage: "BETA"
+  launchStage: "GA"
   location: "us-central1"
   projectRef:
     external: "projects/${projectId}"
--- a/pkg/test/resourcefixture/testdata/basic/run/v1beta1/runservice/runservicemulticontainer/update.yaml
+++ b/pkg/test/resourcefixture/testdata/basic/run/v1beta1/runservice/runservicemulticontainer/update.yaml
@@ -18,7 +18,7 @@ metadata:
   name: runservice-${uniqueId}
 spec:
   ingress: "INGRESS_TRAFFIC_ALL"
-  launchStage: "BETA"
+  launchStage: "GA"
   location: "us-central1"
   projectRef:
     external: "projects/${projectId}"


gemmahou:
Shall we add a comment in the test yaml or update our doc, saying that "set launchStage to GA to avoid unexpected diff".

I'm actually confused why the API changes the value. I feel this is **unaccepted**, because it alters user input and may affect their desired behavior. (this is different with the project id->project number change we found in some compute resources; project number still points to the same project, but BETA->GA alters the desired launch stage)

maqiuyujoyce:
> Shall we add a comment in the test yaml or update our doc, saying that "set launchStage to GA to avoid unexpected diff".

It depends. Not all the Cloud Run service behave this way. Using the hard-coded launchStage `GA` might be confusing. And `GA` may not be the desired value of the users'.

> I'm actually confused why the API changes the value. I feel this is unacceptable(and we probably should ask them to fix it), because it alters user input and may affect their desired behavior. (this is different with the project id->project number change we found in some compute resources; project number still points to the same project, but BETA->GA alters the desired launch stage)

It was a design decision made a while ago before RunService was supported in Config Connector. To keep backwards compatibility, the service team can't really do too much change.
Please note that in the API design, the returned launchStage is not longer the desired launchStage. The API changes the semantics of this field IMO.
Please check b/375678794 for more details. We can discuss offline if you have any further questions.

gemmahou:
> > Shall we add a comment in the test yaml or update our doc, saying that "set launchStage to GA to avoid unexpected diff".
> 
> It depends. Not all the Cloud Run service behave this way. Using the hard-coded launchStage `GA` might be confusing. And `GA` may not be the desired value of the users'.
> 
> > I'm actually confused why the API changes the value. I feel this is unacceptable(and we probably should ask them to fix it), because it alters user input and may affect their desired behavior. (this is different with the project id->project number change we found in some compute resources; project number still points to the same project, but BETA->GA alters the desired launch stage)
> 
> It was a design decision made a while ago before RunService was supported in Config Connector. To keep backwards compatibility, the service team can't really do too much change. Please note that in the API design, the returned launchStage is not longer the desired launchStage. The API changes the semantics of this field IMO. Please check b/375678794 for more details. We can discuss offline if you have any further questions.

Sounds good! I do not have the approver permission, feel free to ask one of the approvers to approve this PR: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/OWNERS

yuwenma (commented):
/lgtm
/approve

Thanks!

PR #3136, opened 2024-11-08
feat: Generate and fixup WorkstationConfig mapping functions

jasonvigil:
Forgot to rename some of the shared mapping functions previously (had meant to do that in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3159). So, renamed and fixed the shared mapping function names in this commit also.

Also, for now, using `_Alpha` to distinguish between shared mapping functions that act on the v1alpha1 API vs the "normal" shared mapping functions that act on the v1beta1 API, but they are functionally equivalent. Will delete these functions and switch over to the "normal", non-`_Alpha` (v1beta1) version ones upon promoting WorkstationConfig to v1beta1.

### DIFF
--- a/dev/tools/controllerbuilder/generate.sh
+++ b/dev/tools/controllerbuilder/generate.sh
@@ -203,6 +203,13 @@ go run . generate-types \
     --output-api ${APIS_DIR} \
     --resource WorkstationCluster:WorkstationCluster
 
+go run . generate-types \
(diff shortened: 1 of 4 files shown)


yuwenma:
/lgtm
/approve

Thanks! (I seems to have reviewed similar annotation/labels somewhere 😄 )

PR #3141, opened 2024-11-08
fix: Remove whitespace from docs command

jasonvigil:
Super minor, but this caused an issue for me when copy/pasting the command into my terminal, editing, and attempting to run it.

The error I got was:
```
`--proto-resource` is required
exit status 1
```


yuwenma (commented):
/lgtm
/approve

Thanks!

PR #3143, opened 2024-11-09
fix: Remove need for parent field in resource references

jasonvigil:
The unexported parent field currently causes the apimachinery unstructured converter to panic:
https://github.com/kubernetes/kubernetes/issues/124154.

This fix is a way to work around the issue.

### DIFF
--- a/dev/tools/controllerbuilder/template/apis/refs.go
+++ b/dev/tools/controllerbuilder/template/apis/refs.go
@@ -43,8 +43,6 @@ type {{.Kind}}Ref struct {
 
 	// The namespace of a {{.Kind}} resource.
 	Namespace string ` + "`" + `json:"namespace,omitempty"` + "`" + `
-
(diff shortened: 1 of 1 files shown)


dev/tools/controllerbuilder/template/apis/refs.go
acpana, line 1: Thanks for looking into this 💯  -- do we want to also tackle any direct resources which followed this pattern? We can do that outside of this PR.

yuwenma:
/hold

Thank you for fixing this. However, this will break the controller tool (where parent is called) if used as a bundle. I want to further refactor the _reference.go file to redefine the parent, reference and status externalRef. Can we hold this PR?

jasonvigil:
Discussed offline w/ @yuwenma. There are already some plans for fixing this in a different way, which include a larger refactor to the reference / controller generation code.

PR #3144, opened 2024-11-09
tests: increase timeout for lint job

justinsb:
We seem to be hitting the timeout, particularly with the go update.


yuwenma (commented):
/lgtm
/approve

Or we can run the subdirectory, like ignoring the auto-generated client go lib

PR #3145, opened 2024-11-11
feat: change UUID to user-specified ID in bqcc

yuwenma:
This change overrides the BQCC server behavior a little bit, that it no longer uses the service-generated UUID, instead, it always uses a user-specified ID. Another option that is more close to the BQC server behavior is that only `spec.resourceID`  configures the user created connectionID, if not given, KCC lets BQC assigns a default connection ID (UUID). The downside is that user may be confused due to the slightly different use of this resource's `metadata.name` and `spec.resourceID` compared to other resources. @jingyih  WDYT?


yuwenma:
The fix has been checked in

PR #3147, opened 2024-11-11
fix: Remove service prefix from WorkstationCluster

jasonvigil:
Can refactor the ref code into apis/ later, for now just a quick fix to use the correct format


yuwenma (commented):
/lgtm
/approve

PR #3159, opened 2024-11-12
Cleanup: Rename WorkstationCluster files for readability and uniqueness

jasonvigil:
Some minor tweaks. About to add WorkstationConfig code, so need to rename WorsktationCluster types / methods for uniqueness.

Also, rename the controller files to match pattern set in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3011.


yuwenma (commented):
/lgtm
/approve

Good catch. Can you update the controller template as well?

PR #3161, opened 2024-11-12
feat: Add support for nested arrays to gapic converter

jasonvigil:
Also, fixes the comments, so they are now applied.

### DIFF
--- a/mockgcp/apis/mockgcp/cloud/ids/v1/service.proto
+++ b/mockgcp/apis/mockgcp/cloud/ids/v1/service.proto
@@ -13,180 +13,277 @@
 // limitations under the License.
 
 syntax = "proto2";
-package google.cloud.ids.v1;
+package mockgcp.cloud.ids.v1;
 import "google/api/annotations.proto";
 import "google/longrunning/operations.proto";
 import "google/protobuf/any.proto";
 import "google/protobuf/timestamp.proto";
 option go_package = "cloud.google.com/go/ids/apiv1/idspb;idspb";
+
 service ProjectsServer {
 }
+
 service ProjectsLocationsServer {
-  rpc GetProjectsLocation(GetProjectsLocationRequest) returns (Location){
-  option (google.api.http) = {
-    get:  "/v1/{name=projects/*/locations/*}"
+
+  // Gets information about a location.
+  rpc GetProjectsLocation(GetProjectsLocationRequest) returns (Location) {
+    option (google.api.http) = {
+      get: "/v1/{name=projects/*/locations/*}"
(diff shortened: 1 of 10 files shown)


mockgcp/apis/mockgcp/cloud/ids/v1/service.proto
```diff
@@ -13,187 +13,285 @@
...
 message ListEndpointsResponse {
+  // The list of endpoints response.
   repeated Endpoint endpoints = 1 [json_name="endpoints"];
+  // A token, which can be sent as `page_token` to retrieve the next page. If this field is omitted, there are no subsequent pages.
   optional string next_page_token = 2 [json_name="nextPageToken"];
+  // Locations that could not be reached.
   repeated string unreachable = 3 [json_name="unreachable"];
 }
+
+  // The response message for Locations.ListLocations.
```
yuwenma, line 170: Can we fix the indentation for top tier message?

yuwenma (commented):
/approve

Thank you! Left one comment.

mockgcp/apis/mockgcp/cloud/ids/v1/service.proto
jasonvigil: Good catch, fixed

mockgcp/apis/mockgcp/cloud/ids/v1/service.proto
```diff
@@ -13,187 +13,285 @@
 // limitations under the License.
 
 syntax = "proto2";
-package google.cloud.ids.v1;
+package mockgcp.cloud.ids.v1;
```
yuwenma, line 16: Do you know what happened here?

jasonvigil: I think this proto was originally generated before https://github.com/GoogleCloudPlatform/k8s-config-connector/commit/bfe519eddf03983d049084ccd32abb072d552477, which retroactively changed the package, but did not re-generate the protos.

jasonvigil: Actually just realized -- maybe this means that this PR didn't really fix the comments, it was just the first time in a while that we've re-generated the protos. Hopefully this doesn't cause issues for the services with regenerated protos.

yuwenma: I was wondering the same, but all the presubmit test passes. I spot checked some proto code diff, seems to be all backward compatible (which makes sense because KCC only add GA-ed resources)

PR #3167, opened 2024-11-13
mock: fix service account ID in connection mock

jingyih:


jingyih:
/assign @yuwenma 

Not release blocking. Just a bug in mock.

yuwenma (commented):
/lgtm
/approve

Good call!

Issue #3162, opened 2024-11-13
Implement retry in the dynamic test


ziyue-101:
### Checklist

- [X] I did not find a related open enhancement request.
- [X] I understand that enhancement requests filed in the GitHub repository are by default low priority.
- [X] If this request is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Describe the feature or resource

The dynamic test doesn't retry on failed operations including (create/update/delete) see https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/dynamic/dynamic_controller_integration_test.go#L499C28-L499C34

This increase the flakiness of test runs,  some example failures are due to concurrent update to the GKE Hub Feature

```
    dynamic_controller_integration_test.go:500: error deleting: error deleting the resource x62xgcdenatqqii/gkehubfeaturemembership-basic-acm-xxxxxxxxi: googleapi: Error 409: unable to queue the operation
        Details:
        [
          {
            "@type": "type.googleapis.com/google.rpc.ResourceInfo",
            "resourceName": "projects/cnrm-test-xxxxxx/locations/global/features/configmanagement"
          }
        ]

```

ziyue-101:
/cc @acpana 

Feel free to triage

ldanielmadariaga:
This item has been open for >12 months. As part of scheduled repository maintenance, this Issue/PR will be closed. If this is still a critical ask, please re-open and work with your Google Cloud Account team to prioritize it with GCP KCC Product Management.

PR #3172, opened 2024-11-13
tests: simple backoff on direct dynamic

acpana:
Going off of #3162 , let's add a simple exp backoff to the delete. The underlying 1P API calls  may benefit from this. Ideally we'd get the retries for free from the testing framework but starting off here for now.

### DIFF
--- a/pkg/test/resourcefixture/contexts/register.go
+++ b/pkg/test/resourcefixture/contexts/register.go
@@ -447,20 +447,38 @@ func directCreate(ctx context.Context, u *unstructured.Unstructured, c client.Cl
 }
 
 func directDelete(ctx context.Context, u *unstructured.Unstructured, c client.Client) error {
+	maxRetries := 3
+	baseDelay := 100 * time.Millisecond // starting delay
+	var err error
+
 	a, err := getAdapter(ctx, u, c)
 	if err != nil {
 		return err
 	}
+	// the very first call to find can be true but subsequent calls may return false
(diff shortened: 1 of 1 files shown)


pkg/test/resourcefixture/contexts/register.go
```diff
@@ -447,20 +447,30 @@ func directCreate(ctx context.Context, u *unstructured.Unstructured, c client.Cl
...
-		return err
-	}
-	if !found {
-		return fmt.Errorf("GVK %s '%v' is not found", u.GroupVersionKind(), u.GetName())
+	for i := 0; i < maxRetries; i++ {
+		found, err := a.Find(ctx)
+		if err == nil && found {
+			op := directbase.NewDeleteOperation(c, u)
+			_, err = a.Delete(ctx, op)
+			if err == nil {
```
maqiuyujoyce, line 464: Is it possible that err != nil but the deletion is still completed? I remember some resources may return an error during creation but still create the resource.

In that case we'll get err == nil && !found, and return an error in the second iteration, right? So maybe let only do err == nil && !found check for the first iteration?

acpana: Oh interesting I didn't know that -- I can adjust the code to function like that!

maqiuyujoyce: It may not be a common case. But I think NotFound in general can be considered fine when the resource was found in the first deletion.

maqiuyujoyce (commented):
/lgtm
/approve

PR #3176, opened 2024-11-14
chore: refactor the external reference for SecretManagerSecret

yuwenma:
Separate identity and reference to different go struct; Use struct and functions to define the meaning and usage of each key component.


apis/secretmanager/v1beta1/secret_identity.go
```diff
@@ -0,0 +1,87 @@
...
+func (r *SecretIdentity) Parent() *SecretParent {
+	return r.parent
+}
+
+func (r *SecretIdentity) ID() string {
+	return r.id
+}
+
+type SecretParent struct {
+	ProjectID string
```
cheftako, line 45: Are we planning to add to SecretParent? If not we could probably drop the struct as per https://go.dev/play/p/D8y1EF81zyF

cheftako:
/lgtm
/approve

I like that we have isolated the reference code to its own file. I think we should discuss centralizing it into a library. When we get to dealing with something like https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_iamauditconfigs.iam.cnrm.cloud.google.com.yaml#L89 I think we will want a central library.

PR #3182, opened 2024-11-14
docs:fix: misspell

acpana:
Fix some misspellings! We/ I should write a linter for these!


yuwenma (commented):
/lgtm
/approve

Ah! Good catch

PR #3183, opened 2024-11-14
fix: Update apigee proto definition to fix naming conflict

jasonvigil:
When trying to generate protos, got the following error
```
--grpc-gateway_out: ProjectsServer.ProvisionOrganizationProject: project
is a protobuf message type. Protobuf message types cannot be used as
path parameters, use a scalar value type (such as string) instead:
```

It seems to be due to a naming conflict around "project" which is allowed in the JSON API definition, but causes issues when converting. Probably the easiest fix is to manually make this update.


yuwenma (commented):
I guess this will be overridden if we update the apigee proto. Maybe you can try sometihng like this https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/mockgcp/apply-proto-patches.sh#L29

yuwenma (commented):
/lgtm
/approve

PR #3189, opened 2024-11-15
fix: Remove wait after SQLInstance periodic test steps

jasonvigil:
This is should longer be necessary after waiting for LRO completion: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3034/commits/d9baf0bcffb932e8c8516117aec591c54790d437

### DIFF
--- a/pkg/controller/dynamic/dynamic_controller_integration_test.go
+++ b/pkg/controller/dynamic/dynamic_controller_integration_test.go
@@ -503,7 +503,7 @@ func testDriftCorrection(ctx context.Context, t *testing.T, testContext testrunn
 
 	// Underlying APIs may not have strongly-consistent reads due to caching. Sleep before attempting a re-reconcile, to
 	// give the underlying system some time to propagate the deletion info.
-	time.Sleep(resourceContext.RecreateDelay)
+	time.Sleep(time.Second * 10)
 
 	// get the current state
 	t.Logf("reconcile with %v\r", testUnstruct)
--- a/pkg/test/resourcefixture/contexts/register.go
+++ b/pkg/test/resourcefixture/contexts/register.go
@@ -63,10 +63,6 @@ type ResourceContext struct {
 	// services in GCP that claim to be done with creating / updating the resource before it is actually available.
 	PostModifyDelay time.Duration
 
-	// Time to delay before recreating the resource as part of the drift detection test.
-	// The default wait time is 10 seconds. However, some resources appear to need to
-	// wait longer before recreating, so this value is customizable.
-	RecreateDelay time.Duration
 	// If true, skip drift detection test.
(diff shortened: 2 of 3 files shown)


yuwenma (commented):
/lgtm
/approve

Great!

jasonvigil:
@yuwenma could you pls re-review this? Turns out we no longer need `RecreateDelay`, SQLInstance was the only resource using it. So, also removed that field.

pkg/controller/dynamic/dynamic_controller_integration_test.go
```diff
@@ -503,7 +503,7 @@ func testDriftCorrection(ctx context.Context, t *testing.T, testContext testrunn
 
 	// Underlying APIs may not have strongly-consistent reads due to caching. Sleep before attempting a re-reconcile, to
 	// give the underlying system some time to propagate the deletion info.
-	time.Sleep(resourceContext.RecreateDelay)
+	time.Sleep(time.Second * 10)
```
yuwenma, line 506: Is this a revert to a previous const var? I like that we make this a configurable field. It is more manageable and can help future replace which needs additional delays.

jasonvigil: There are no known circumstances where we need to add additional delays.

yuwenma: "dead code should be cleaned up" --> Right, that makes the KCC code base much better.
I consider this not a dead code but a feature that has been approved working well. For dynamic test, we do not expect the contributors can be able to fix the source code when their resource need some special treatment, the only option for them is to explore the configurables in the resource-context file. Enriching the configurable pools is always appreciated, especially if the feature does not introduce performance regression to the current test.  Therefore, can we keep it?

yuwenma:
/lgtm

We want to keep the custom delay feature in dynamic test. Non blocker to this PR. You can always put it back in the next code change.

pkg/controller/dynamic/dynamic_controller_integration_test.go
jasonvigil: Aside from this code no longer being used, I also feel like this code is not really a desirable feature. It allowed me to miss an actual bug, which was that my direct controller was not waiting correctly for delete LRO's to complete. From this one example of a previous use, this configuration seems like more like a "foot-gun", rather than a desirable feature. If any contributors hit issues in the future where they feel like something like this is actually necessary, I think not having this configuration available would cause us to examine the situation more closely, rather than allow a "quick workaround" of extending the wait time and potentially miss a similar bug.

yuwenma: Ok. If it is not approved to be a desired feature, agree that we should drop it .

PR #3190, opened 2024-11-16
Update mockalloydb and the test golden files

maqiuyujoyce:
### Change description

Support mock for AlloyDBCluster and AlloyDBInstance.

### Tests you have done


pkg/test/resourcefixture/testdata/basic/alloydb/v1beta1/alloydbinstance/basicalloydbinstance/_http.log
```diff
@@ -1335,571 +871,8 @@ X-Frame-Options: SAMEORIGIN
...
-    "requestedCancellation": false,
-    "target": "projects/${projectId}/locations/europe-west1/clusters/alloydbcluster-${uniqueId}/instances/alloydbinstance-${uniqueId}",
-    "verb": "delete"
-  },
-  "name": "projects/${projectId}/locations/europe-west1/operations/${operationID}"
-}
-
----
-
-GET https://compute.googleapis.com/compute/v1/projects/${projectId}/global/networks/${networkID}?alt=json
```
maqiuyujoyce: @gemmahou the GET event for compute address and compute network somehow got deleted when it is running against mock. Wondered if you happen to have any ideas about it.

maqiuyujoyce: Discussed offline. The additional GET requests are triggered by by re-reconciliation happening every 10-ish minutes and the removal of them in the mock http log should not be a concern.

justinsb: I suggest we should set the reconciliation interval high for any objects in dependencies.yaml.

If we wanted to, we could do that automatically or just go through and add them.

mockgcp/Makefile
```diff
@@ -12,16 +12,15 @@ tools:
 	wget -N -O bin/protoc.zip https://github.com/protocolbuffers/protobuf/releases/download/v3.12.4/protoc-3.12.4-linux-x86_64.zip
 	cd bin; rm -rf protoc; mkdir protoc; cd protoc; unzip ../protoc.zip
 
-GOOGLEAPI_VERSION?=HEAD
 .PHONY: sync-repo
```
yuwenma, line 15: Any reasons to remove this feature?  Dedicating to a specific one is better for version control

mockgcp/apply-proto-patches.sh
```diff
@@ -50,17 +50,3 @@ cd tools/patch-proto
 #   // The maintenance update policy determines when to allow or deny updates.
 #   MaintenanceUpdatePolicy maintenance_update_policy = 32;
 # EOF
-
-# Use our proto patch tool to add the missing pscConfig field for AlloyDB cluster.
-go run . --file ${REPO_ROOT}/mockgcp/third_party/googleapis/google/cloud/alloydb/v1beta/resources.proto --message Cluster <<EOF
```
yuwenma, line 55: Oh, nice! Does this mean the AlloyDB proto now is up to date?

tests/e2e/unified_test.go
```diff
@@ -603,6 +603,73 @@ func runScenario(ctx context.Context, t *testing.T, testPause bool, fixture reso
...
+										val == "type.googleapis.com/google.cloud.alloydb.v1beta.Cluster" ||
+										val == "type.googleapis.com/google.cloud.alloydb.v1beta.Instance" ||
+										val == "type.googleapis.com/google.cloud.alloydb.v1beta.Backup" {
+										// Explicitly set `reconciling` in response to `false`.
+										if _, found, _ := unstructured.NestedBool(obj, "response", "reconciling"); !found {
+											if err := unstructured.SetNestedField(obj, false, "response", "reconciling"); err != nil {
+												t.Fatal(err)
+											}
+										}
+										// Explicitly set `automatedBackupPolicy.enabled` in response to `false`.
```
yuwenma, line 660: Can we do this in the Mock service level? My concern is that this could override other resources and make the triaging very hard.

pkg/test/resourcefixture/testdata/basic/alloydb/v1beta1/alloydbinstance/basicalloydbinstance/_http.log
maqiuyujoyce: Thanks for the suggestions, @justinsb ! Yes, increasing the reconciliation interval is how I verified this issue. I plan to do it after the mockgcp update is merged.

I actually think we should do it for all the CRs because the deletion of a dependent resource can take very long and result in the re-reconciliation of the target test resource. So I was thinking about patching this annotation to all the CRs in the test framework. Maybe that's too much automation?

mockgcp/apply-proto-patches.sh
maqiuyujoyce: Currently yes.

tests/e2e/unified_test.go
maqiuyujoyce: I tried and it didn't work. AFAIK, the empty values of the primitive typed fields are somehow dropped in the http log recording for mock. Possibly a gap in the dependent lib? We should look into it when it becomes more problematic. I think other folks ran into the same issue as well.

Regarding the concern about unexpected override, the code I added checks the returned resource name to ensure they are responses for AlloyDB resources. So it's unlikely to override non-AlloyDB resources.

mockgcp/Makefile
maqiuyujoyce: In my understanding, `make sync-repo` should be a target to sync the repo to remote master HEAD. But what `cd third_party/googleapis && git fetch --all && git reset --hard HEAD` does is not syncing the repo to match remote, instead, it simply checkout the local HEAD without rebasing the remote branch. So, not syncing at all.

Or, I can create a new target `checkout-head` to represent the old behavior of `sync-repo`.

pkg/test/resourcefixture/testdata/basic/alloydb/v1beta1/alloydbinstance/fullalloydbinstance/_generated_object_fullalloydbinstance.golden.yaml
```diff
@@ -35,6 +35,13 @@ status:
     status: "True"
     type: Ready
   createTime: "1970-01-01T00:00:00Z"
+  ipAddress: 10.1.2.3
   name: projects/${projectId}/locations/europe-north1/clusters/alloydbcluster${uniqueId}/instances/alloydbinstance${uniqueId}
   observedGeneration: 2
-  state: STATE_UNSPECIFIED
+  outboundPublicIpAddresses:
+  - 35.228.195.235
+  - 34.88.204.106
```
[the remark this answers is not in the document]
maqiuyujoyce: Ah, I got what's missing: these IP addresses were not normalized in the test framework. Done.

mockgcp/Makefile
maqiuyujoyce: Hmmm, so here is a reason I made the change:

I don't know what's the original purpose of `make sync-repo` is, but the functionality to sync local repo `./mockgcp/third_party/googleapis` to remote master HEAD is missing, and can be addressed by `git checkout -f origin/master` (creating a detached branch, but not actually updating the local branch) or `git reset --hard origin/master` (updating the local branch to match the remote master HEAD). (I picked the latter because I think we should update the local branch.)

In mockgcp doc (https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/master/mockgcp#adding-a-new-service), it said: `Add the proto to the Makefile and run make gen-proto (or just make).` While it didn't say what contributors should expect after running the command, I expected that the proto files for the exact resource I worked on under `./mockgcp/generated/mockgcp` would be updated and match the remote master HEAD.

However `make gen-proto` doesn't update `./mockgcp/generated/mockgcp` at all.

My change is to fix this.

maqiuyujoyce: Rebased on top of master and reverted such change. @justinsb seems to be improving the process right now.

tests/e2e/unified_test.go
justinsb: I think it is a proto3 explicit optional, so that's relatively unusual.

If we set it explicitly to false (`.Enabled = PtrTo(false)`), I would hope the protojson marshal code would understand but I guess it doesn't....

justinsb: I commented out both blocks that  `// Explicitly set `automatedBackupPolicy.enabled` in response to `false`.` (this one and the LRO one).  I then ran `hack/compare-mock fixtures/basicalloydbinstance` and it passed.  I also tried `hack/compare-mock fixtures/basicalloydbcluster` and was able to recreate the issue.  The thing is that the diff happens way down the log, and it actually happens after a patch to automatedBackupPolicy.

So I actually think that the problem is that we should default automatedBackupPolicy.enabled to PtrTo(false), and we aren't doing that after a patch.

Something like this:

```
~/kcc/k8s-config-connector  (202411-alloydb-mockgcp) > git diff
diff --git a/mockgcp/mockalloydb/cluster.go b/mockgcp/mockalloydb/cluster.go
index 362aa7847..fb230305a 100644
--- a/mockgcp/mockalloydb/cluster.go
+++ b/mockgcp/mockalloydb/cluster.go
@@ -65,11 +65,11 @@ func setClusterFields(name *clusterName, obj *pb.Cluster) {
                        BackupWindow: direct.PtrTo(duration.Duration{Seconds: 3600}),
                        // Defaults to true when unset, which is different from the value in
                        // the default policy.
-                       Enabled:  direct.PtrTo(false),
+                       Enabled:  PtrTo(false),
                        Location: name.Location,
                        Retention: &pb.AutomatedBackupPolicy_TimeBasedRetention_{
                                TimeBasedRetention: &pb.AutomatedBackupPolicy_TimeBasedRetention{
-                                       RetentionPeriod: direct.PtrTo(duration.Duration{Seconds: 1209600}),
+                                       RetentionPeriod: PtrTo(duration.Duration{Seconds: 1209600}),
                                },
                        },
                        Schedule: &pb.AutomatedBackupPolicy_WeeklySchedule_{
@@ -240,6 +240,10 @@ func (s *AlloyDBAdminV1) UpdateCluster(ctx context.Context, req *pb.UpdateCluste
                obj.ContinuousBackupInfo.EnabledTime = nil
        }
 
+       if obj.AutomatedBackupPolicy != nil && obj.AutomatedBackupPolicy.Enabled == nil {
+               obj.AutomatedBackupPolicy.Enabled = PtrTo(false)
+       }
+
        if err := s.storage.Update(ctx, fqn, obj); err != nil {
                return nil, err
        }

```


(We also should not be using direct.PtrTo from mockgcp, because then we introduce a circular dependency into the modules which makes updating them tricky.  In other services we've created a utils.go in each service.  We could always have a similar set of utils for mockgcp - we do have common so maybe that's a natural place)

maqiuyujoyce: Thank you for all the details, @justinsb ! Yes, the optional fields (e.g. `automatedBackupPolicy.enabled`) should have their values recorded in the http log as is, and the missing part here is to set the value of `automatedBackupPolicy.enabled` explicitly again in the mock Update.

@yuwenma Didn't realize you are calling out this specific field. Should be updated now. My original comment is about other fields that need to have `false` explicitly set in the normalizer.

yuwenma (commented):
/lgtm
/approve

PR #3193, opened 2024-11-18
Use CEL rule to validate field immutability for direct Compute resources

gemmahou:
### Change description

b/377531379

Checked all three direct compute resources(forwarding rule, firewall policy rule, target tcp proxy), added CEL rule on the immutable and required fields, following https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/develop-resources/api-conventions/validations.md#option-2

Note: We haven't turned on direct reconciliation for those compute resources yet, instead, we use the "alpha.cnrm.cloud.google.com/reconciler: "direct"" annotation to opt-in. So they are still protected by the immutable webhook for now. 

### Tests you have done

### DIFF
--- a/apis/compute/v1beta1/firewallpolicyrule_types.go
+++ b/apis/compute/v1beta1/firewallpolicyrule_types.go
@@ -102,12 +102,14 @@ type ComputeFirewallPolicyRuleSpec struct {
 	// +optional
 	EnableLogging *bool `json:"enableLogging,omitempty"`
 
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="the field is immutable"
 	/* Immutable. */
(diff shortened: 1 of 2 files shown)


apis/compute/v1beta1/firewallpolicyrule_types.go
```diff
@@ -102,12 +102,14 @@ type ComputeFirewallPolicyRuleSpec struct {
 	// +optional
 	EnableLogging *bool `json:"enableLogging,omitempty"`
 
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="the field is immutable"
```
yuwenma, line 105: These fields are related to identity. We do not need the immutable check, but set this as required.

gemmahou:
No immutability checks on fields are needed.

PR #3199, opened 2024-11-19
fix: Remove underscores from generated type names

jasonvigil:
In general, go names should not contain underscores. Also, the names
(with underscores) start to look a bit strange when we start to use them
in conjunction with the "ObservedState" suffix. For example, this is not
ideal: `WorkstationConfig_HostObservedState`. Would prefer to just avoid
underscores altogether.

### DIFF
--- a/dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
+++ b/dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
@@ -262,7 +262,7 @@ func goNameForProtoMessage(msg protoreflect.MessageDescriptor) string {
 
 	fullName = strings.TrimPrefix(fullName, string(msg.ParentFile().FullName()))
 	fullName = strings.TrimPrefix(fullName, ".")
-	fullName = strings.ReplaceAll(fullName, ".", "_")
+	fullName = strings.ReplaceAll(fullName, ".", "")
 	return fullName
 }
 


dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
```diff
@@ -262,7 +262,7 @@ func goNameForProtoMessage(msg protoreflect.MessageDescriptor) string {
 
 	fullName = strings.TrimPrefix(fullName, string(msg.ParentFile().FullName()))
 	fullName = strings.TrimPrefix(fullName, ".")
-	fullName = strings.ReplaceAll(fullName, ".", "_")
+	fullName = strings.ReplaceAll(fullName, ".", "")
```
jingyih, line 265: Is there a particular reason to remove the underscore? I think it helps with readability. As an random example:
https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/064601c7e765e8f9b9c2dce0b41c78ba98d7f926/apis/spanner/v1beta1/types.generated.go#L48-L49

jasonvigil: Ah, I tried to give some reasoning in the commit message:
> In general, go names should not contain underscores. Also, the names
> (with underscores) start to look a bit strange when we start to use them
> in conjunction with the "ObservedState" suffix. For example, this is not
> ideal: `WorkstationConfig_HostObservedState`. Would prefer to just avoid
> underscores altogether.

jingyih: Oh, I just noticed your commit message. This seems like a matter of convention. We could go either way, but if we decide to remove the underscores, we might need to update all the existing generated types accordingly.

jasonvigil: Would we really need to update all the existing generated types? I thought the tooling used the `// +kcc:proto=xyz` comments to find the corresponding types, not their names.

yuwenma: +1 to Jingyi that underscore helps with readability. Also, this is a convention in the sig k8s client-go for generated functions.

jasonvigil: @yuwenma can you point to an example of this convention? I thought underscores were discouraged in go naming: https://google.github.io/styleguide/go/guide#mixedcaps

jasonvigil: Ok, I think I found where underscores are allowed in Kubernetes. They allow underscores for autogenerated conversion function names https://github.com/kubernetes/kubernetes/blob/bf70d289fb3b7deb9b121bbe10e278f12840cfcb/hack/golangci.yaml#L55. However, it does not look like Kubernets allows underscores in any type names. At least, the linter is not configured to allow it.

PR #3200, opened 2024-11-19
refactor:tests: field management scenario tests

acpana:
A very small refactor to organize the scenario tests around field management.


tests/e2e/testdata/scenarios/fields/management/README.md
```diff
@@ -0,0 +1,20 @@
+This test folder contains scenarios for setting and unsetting a field.
+
+Unsetting a native field (int, bool, string) refers to setting that field to the
```
maqiuyujoyce, line 3: This is always a confusing concept for users. And while I understand our functionality, I could understand what this sentence means exactly. It'll be hard to summarize this functionality, so could you add more details when explaining it? E.g.

```
Unsetting a native field (int, bool, string) in the underlying GCP resource that a Config Connector resource manages can be achieved by setting the corresponding field in the Config Connector resource to its 0 value. This means setting an int field to `0`, a bool field to `false`,  a string field to `""` (and etc).
```

(I kinda want to avoid mentioning Go in case it confuses people more.)

maqiuyujoyce (commented):
Thank you, @acpana ! Just one comment about the doc.

tests/e2e/testdata/scenarios/fields/management/README.md
```diff
@@ -0,0 +1,20 @@
...
+0 value in Go.
```
acpana, line 4:
```suggestion
Unsetting a native field (int, bool, string) in the underlying GCP resource that a Config Connector resource manages can be achieved by setting the corresponding field in the Config Connector resource to its 0 value. This means setting an int field to `0`, a bool field to `false`,  a string field to `""` (and etc).
```

maqiuyujoyce (commented):
/lgtm
/approve

PR #3203, opened 2024-11-19
fix: Add known acronyms for type generator

jasonvigil:
These were used in WorkstationConfig.

### DIFF
--- a/dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
+++ b/dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
@@ -348,6 +348,20 @@ func isAcronym(s string) bool {
 		return true
 	case "http", "https", "ssh":
 		return true
+	case "ip":
+		return true
+	case "gb":
+		return true
+	case "fs":
+		return true
+	case "pd":
+		return true
+	case "kms":
+		return true
+	case "gce":
+		return true
+	case "vtpm":
+		return true
 	default:
 		return false
 	}


jingyih (commented):
/lgtm

dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
```diff
@@ -348,6 +348,20 @@ func isAcronym(s string) bool {
...
+	case "fs":
+		return true
+	case "pd":
+		return true
+	case "kms":
+		return true
+	case "gce":
+		return true
+	case "vtpm":
+		return true
```
yuwenma, line 364: Cool, can I ask where you get this list?

dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
```diff
@@ -348,6 +348,20 @@ func isAcronym(s string) bool {
...
+	case "fs":
+		return true
+	case "pd":
+		return true
+	case "kms":
+		return true
+	case "gce":
+		return true
+	case "vtpm":
```
yuwenma, line 363: Shall this be "vTPM}"?

dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
jasonvigil: WorkstationCluster

dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
jasonvigil: That would mean the variable ends up named `EnablevTPM`, which is not ideal. https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/603f8c85cdd15ab4c605eb1034d22d4d06b4fa14/apis/workstations/v1alpha1/config_types.go#L145

yuwenma: Fair enough.

PR #2953, opened 2024-10-18
api: generate types of VPCAccessConnector

justinsb:

PR #3212, opened 2024-11-20
chore: improvements around Direct identity and external  

yuwenma:
Improvements:
- Separate resource Identity and external Ref
- Added in-line comments to guide the contributors on the functionality and expected behaviors. 
- Make the variable names shorter. e.g. `SecretManagerSecretVersion` --> `SecretVersion`
- Applied some best practice like introducing the CompareProtoMessage

Fixes:
- Remove `parent` from the reference API to avoid the json tag hack, and potential client-go unstructured parsing issues.  
- Parent and Identity should be read-only for adapter.

### DIFF
--- a/dev/tools/controllerbuilder/pkg/commands/generatetypes/generatetypescommand.go
+++ b/dev/tools/controllerbuilder/pkg/commands/generatetypes/generatetypescommand.go
@@ -189,6 +189,14 @@ func RunGenerateCRD(ctx context.Context, o *GenerateCRDOptions) error {
 				return fmt.Errorf("add refs file %s: %w", scaffolder.PathToRefsFile(kind, resource.ProtoName), err)
 			}
 		}
+		if scaffolder.IdentityFileExist(kind, resource.ProtoName) {
+			fmt.Printf("file %s already exists, skipping\n", scaffolder.PathToIdentityFile(kind, resource.ProtoName))
+		} else {
+			err := scaffolder.AddIdentityFile(kind, resource.ProtoName)
+			if err != nil {
+				return fmt.Errorf("add identity file %s: %w", scaffolder.PathToIdentityFile(kind, resource.ProtoName), err)
+			}
+		}
 	}
 
 	if err := typeGenerator.WriteVisitedMessages(); err != nil {
--- a/dev/tools/controllerbuilder/template/apis/identity.go
+++ b/dev/tools/controllerbuilder/template/apis/identity.go
@@ -1,0 +1,124 @@
+package apis
+
+const IdentityTemplate = `// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package {{ .Version }}
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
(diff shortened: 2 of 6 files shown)


dev/tools/controllerbuilder/pkg/commands/generatetypes/generatetypescommand.go
```diff
@@ -189,6 +189,14 @@ func RunGenerateCRD(ctx context.Context, o *GenerateCRDOptions) error {
 				return fmt.Errorf("add refs file %s: %w", scaffolder.PathToRefsFile(kind, resource.ProtoName), err)
 			}
 		}
+		if scaffolder.IdentityFileExist(kind, resource.ProtoName) {
```
justinsb, line 192: Not sure it needs to be another file, but ... we can see!

dev/tools/controllerbuilder/pkg/commands/generatetypes/generatetypescommand.go
```diff
@@ -189,6 +189,14 @@ func RunGenerateCRD(ctx context.Context, o *GenerateCRDOptions) error {
...
+			fmt.Printf("file %s already exists, skipping\n", scaffolder.PathToIdentityFile(kind, resource.ProtoName))
+		} else {
+			err := scaffolder.AddIdentityFile(kind, resource.ProtoName)
+			if err != nil {
+				return fmt.Errorf("add refs file %s: %w", scaffolder.PathToIdentityFile(kind, resource.ProtoName), err)
```
justinsb, line 197: Nit: "adding identity file %q: %w"

justinsb:
A few thoughts but nothing blocking.

/approve
/lgtm

justinsb:
BTW I tried using your suggested approach for the ID in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2953 and I like it.

I personally find it easier to write the code first, and then update the template, but not a problem!

dev/tools/controllerbuilder/pkg/commands/generatetypes/generatetypescommand.go
yuwenma: I grouped them by logical unit. The identity file is for the resource own usage. The reference file is for other resources. For contributors using reference, they don't need to know and read the identity logic.

dev/tools/controllerbuilder/template/apis/identity.go
```diff
@@ -0,0 +1,124 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// {{.ProtoResource}}Identity defines the resource reference to {{.Kind}}, which "External" field
+// holds the GCP identifier for the KRM object.
+type {{.ProtoResource}}Identity struct {
+	parent *{{.ProtoResource}}Parent
```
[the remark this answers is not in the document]
yuwenma: Yes, it is more than a json tag. So I want to do that in a follow up PR.

dev/tools/controllerbuilder/template/apis/identity.go
```diff
@@ -0,0 +1,124 @@
...
+	// Use approved External
+	externalRef := common.ValueOf(obj.Status.ExternalRef)
+	if externalRef != "" {
+		// Validate desired with actual
+		actualParent, actualResourceID, err := Parse{{.ProtoResource}}External(externalRef)
+		if err != nil {
+			return nil, err
+		}
+		if actualParent.ProjectID != projectID {
+			return nil, fmt.Errorf("spec.projectRef changed, expect %s, got %s", actualParent.ProjectID, projectID)
```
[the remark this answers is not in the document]
yuwenma: This code is not only for string validation, but helping users to write correct KCC config as well.
https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/develop-resources/api-conventions/validations.md#rule-3-parent

But I will change this code anyways in the parent PR 😄

justinsb (approved):

PR #3222, opened 2024-11-20
chore: controllergen uses pinned googleapi version

justinsb:
This ensures that we have one version of the googleapis, which would otherwise get confusing,  for example if the mocks had  fields that the clients did not.


jingyih (commented):
/lgtm

Cool! This is helpful for detecting future API changes!

yuwenma (commented):
/lgtm
/approve

Thanks

PR #3228, opened 2024-11-20
feat: Generate "ObservedState" structs in type generator

jasonvigil:
The ObservedState structs contain the output-only fields of the original messages, and are intended to be used in the KRM status. The types intended for the KRM spec no longer contain the output-only fields.

This PR depends on https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3337

### DIFF
--- a/apis/workstations/v1alpha1/types.generated.go
+++ b/apis/workstations/v1alpha1/types.generated.go
@@ -111,41 +111,28 @@ type WorkstationCluster_PrivateClusterConfig struct {
 // +kcc:proto=google.cloud.workstations.v1.WorkstationConfig
 type WorkstationConfig struct {
 	// Full name of this workstation configuration.
+	// +kcc:proto:field=google.cloud.workstations.v1.WorkstationConfig.name
 	Name *string `json:"name,omitempty"`
 
 	// Optional. Human-readable name for this workstation configuration.
+	// +kcc:proto:field=google.cloud.workstations.v1.WorkstationConfig.display_name
 	DisplayName *string `json:"displayName,omitempty"`
 
-	// Output only. A system-assigned unique identifier for this workstation
-	//  configuration.
-	Uid *string `json:"uid,omitempty"`
-
-	// Output only. Indicates whether this workstation configuration is currently
-	//  being updated to match its intended state.
-	Reconciling *bool `json:"reconciling,omitempty"`
-
 	// Optional. Client-specified annotations.
+	// +kcc:proto:field=google.cloud.workstations.v1.WorkstationConfig.annotations
 	Annotations map[string]string `json:"annotations,omitempty"`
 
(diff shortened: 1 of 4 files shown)


apis/workstations/v1alpha1/types.generated.go
yuwenma, line 1: Looks good and the approach works for me. Thank you for the contribution.

jasonvigil:
After offline discussion w/ @jingyih, I am planning to restructure this code a bit and create a separate function for determining output fields (rather than re-use the "find dependency" code path). Marking WIP for now.

jingyih (commented):
Thank you for fixing this issue! I haven't tried this, so I'm not entirely sure it's feasible, but my suggestion would be to maintain the existing logic for finding dependencies and introduce a post-processing step to handle nested output fields before writing the messages to the file. This approach could result in cleaner code that is easier to debug and maintain. The only potential downside might be related to performance, but since this is an offline process, performance concerns are not a priority.

jasonvigil:
> Thank you for fixing this issue! I haven't tried this, so I'm not entirely sure it's feasible, but my suggestion would be to maintain the existing logic for finding dependencies and introduce a post-processing step to handle nested output fields before writing the messages to the file. This approach could result in cleaner code that is easier to debug and maintain. The only potential downside might be related to performance, but since this is an offline process, performance concerns are not a priority.

Ok, I just did this. Moved the logic for finding outputs to a separate function, and created a separate code path for it. Also, included a diff for the generated types of WorkstationConfig to illustrate this change.

dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
```diff
@@ -436,3 +541,91 @@ func RemoveNotMappedToGoStruct(msgs map[string]protoreflect.MessageDescriptor) {
...
+
+	res := []*OutputMessageDetails{}
+	for _, msgDetails := range outputDeps {
+		res = append(res, msgDetails)
+	}
+	return res, nil
+}
+
+// FindDependenciesForField recursively explores the dependent proto messages of the given field.
+func FindOutputsForField(field protoreflect.FieldDescriptor, seen map[string]bool, outputDeps map[string]*OutputMessageDetails, ignoredFields sets.String) bool {
```
jingyih, line 575: nit: add a description of the function return? Something like:  function returns true if the current field or any of its nested field is marked as an output field.

jingyih (commented):
/lgtm

dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
jasonvigil: Done

jingyih (commented):
/lgtm

Thanks!

PR #3231, opened 2024-11-21
feat: support referencing BQCC service account ID in IAM resources

### LINKED ISSUE #2243: Retrieve serviceaccount from bigquery connection to be referenced in IamPolicyMember
### Checklist

b/339349950

Regarding connections: the name of the service account (SA) in a connection cannot be known ahead of time. It is generated randomly. Today users must create a connection first, then retrieve the created connection to get the SA name.

KCC to support retrieval of the service account from Bigquery external connections to be referenced in iamPolicyMember.

iamPolicyMember has the field "spec.memberFrom" which can be used to reference the SA generated by a cloud resource. Today it only supports a few resource types such as "sqlinstanceRef" as described in [b/242856558](https://buganizer.corp.google.com/issues/242856558)

A valid solution could be to extend this behaviour to support "bigqueryconnection"

```yaml
memberForm:
    bigqueryconnectionRef:
        name: <connection-id>
        namespace: <namespace>
```

### Additional information

_No response_

### Importance

_No response_

jingyih:
Fixes #2243

TESTED

```
$ go test -v -tags=integration pkg/controller/iam/iamclient/iamclient_integration_test.go -test.run=TestAllGetSetDeletePolicyMemberWithMemberReference -run-tests bigqueryconnectionconnectionref

$ GOLDEN_REQUEST_CHECKS=1 GOLDEN_OBJECT_CHECKS=1 WRITE_GOLDEN_OUTPUT=1 E2E_KUBE_TARGET=envtest RUN_E2E=1 E2E_GCP_TARGET=mock go test -test.count=1 -v ./tests/e2e -run TestAllInSeries/fixture/bigqueryconnectionconnectionref
```


pkg/apis/iam/v1beta1/krm_types.go
```diff
@@ -54,6 +59,10 @@ type MemberSource struct {
 	// The ServiceIdentity whose service account (i.e., its
 	// 'status.email') is to be bound to the role.
 	ServiceIdentityRef *MemberReference `json:"serviceIdentityRef,omitempty"`
+
+	// BigQueryConnectionConnection whose service account in
+	// `status.observedState.cloudSQL.serviceAccountID` is to be bound to the role.
+	BigQueryConnectionConnectionCloudSQLRef *MemberReference `json:"bigQueryConnectionConnectionCloudSQLRef,omitempty"`
```
jingyih, line 65: There are different ways to implement this. The API will generate a service account ID and write it to one of three possible field, depending on the connection configuration.

Option 1: Add three dedicated fields under `memberFrom`, each corresponding to one of the possible fields mentioned above.
```
memberFrom:
    bigQueryConnectionConnectionCloudSQLRef:
        name: <connection-id>
        namespace: <namespace>
```
```
memberFrom:
    bigQueryConnectionConnectionCloudResourceRef:
        name: <connection-id>
        namespace: <namespace>
```
```
memberFrom:
    bigQueryConnectionConnectionSparkRef:
        name: <connection-id>
        namespace: <namespace>
```
Option 2. Use a single field under `memberFrom`, where the controller dynamically determines which of the three possible fields is non-empty and retrieves the service account ID accordingly.
```
memberFrom:
    bigQueryConnectionConnectionRef:
        name: <connection-id>
        namespace: <namespace>
```

600lyy (approved):
I think option 2 is better to have a generic field to hold the service account created by BigqueryConnectionConnection. There might be more connection types to be added in BQCC, thus having a generic field makes it consistent for user experience

pkg/apis/iam/v1beta1/krm_types.go
yuwenma: Have we considered building this inside ServiceAccountRef ? For example,
```yaml
spec:
   memberFrom:
        serviceAccountRef: # Existing already
             valueFrom:  # New field 
                  BigQueryConnectionConnectionRef:
                         name: 
                         namespace:
                         externalRef:
```

apis/bigqueryconnection/v1beta1/connection_reference.go
```diff
@@ -149,8 +150,32 @@ func (r *BigQueryConnectionConnectionRef) NormalizedExternal(ctx context.Context
...
+	if err := reader.Get(ctx, key, u); err != nil {
+		if apierrors.IsNotFound(err) {
+			return "", k8s.NewReferenceNotFoundError(u.GroupVersionKind(), key)
+		}
+		return "", fmt.Errorf("reading referenced %s %s: %w", BigQueryConnectionConnectionGVK, key, err)
+
+	}
+	sa, found, err := unstructured.NestedString(u.Object, "status", "observedState", "cloudSQL", "serviceAccountID") // TODO: a better way to provide info on which field to read?
+	if err != nil {
+		return "", err
```
yuwenma, line 175: Would GCP service return empty SAID? If so, this would be found=true, err=nil, and sa=""

apis/bigqueryconnection/v1beta1/connection_reference.go
```diff
@@ -149,8 +150,32 @@ func (r *BigQueryConnectionConnectionRef) NormalizedExternal(ctx context.Context
...
+	}
+	if !found {
+		return "", fmt.Errorf("no value found under `status.observedState.cloudSQL.serviceAccountID` (maybe the BigQueryConnectionConnection resource is not ready yet)")
```
yuwenma, line 178: nit: This will show up in the IAM controller which makes it confusing because the IAM does not have `status.observedState.cloudSQL.serviceAccountID`

How about 

if !found || err != nil {
   return "", fmt.Errorf("unable to resolve BigQueryConnectionConnection `status.observedState.cloudSQL.serviceAccountID` for %s. found=%t: %w", key. found,  err)
}

apis/bigqueryconnection/v1beta1/connection_reference.go
jingyih: During my testing, I did not encounter an empty service account. Based on the golden object files generated from our existing tests, service accounts are always present for connections of type Spark, CloudSQL, and CloudResource. Did you encounter some corner case?

pkg/apis/iam/v1beta1/krm_types.go
jingyih: We decided to go with option 3 in the doc.
```yaml
spec:
  memberFrom:
    bigQueryConnectionConnectionRef:
      type: "cloudSQL"  # Allowed values: "cloudSQL", "cloudResource", "spark"
      name:
      namespace:
```

jingyih:
Updated PR to use the following API.

```yaml
spec:
  memberFrom:
    bigQueryConnectionConnectionRef:
      type: "cloudSQL"  # Allowed values: "cloudSQL", "cloudResource", "spark"
      name:
      namespace:
```

apis/bigqueryconnection/v1beta1/connection_reference.go
```diff
@@ -149,8 +150,49 @@ func (r *BigQueryConnectionConnectionRef) NormalizedExternal(ctx context.Context
...
+	sa, found, err := unstructured.NestedString(u.Object, path...)
+	if err != nil {
+		return "", fmt.Errorf("failed to access serviceAccountID field in BigQueryConnection %s/%s: %w", key.Namespace, key.Name, err)
+	}
+	if !found {
+		pathStr := strings.Join(path, ".")
+		return "", fmt.Errorf("BigQueryConnection %s/%s is not ready - field '%s' is missing", key.Namespace, key.Name, pathStr)
+	}
+
+	return "serviceAccount:" + sa, nil // TODO: a better way to provide format/template
```
yuwenma, line 197: Suggest returning "sa" only, and let the caller to deal with the format

apis/bigqueryconnection/v1beta1/connection_reference.go
```diff
@@ -149,8 +150,49 @@ func (r *BigQueryConnectionConnectionRef) NormalizedExternal(ctx context.Context
...
+		return "", fmt.Errorf("invalid connection type '%s' for BigQueryConnection for IAM reference. Supported types are: cloudSQL, spark, cloudResource", ref.Type)
+	}
+
+	sa, found, err := unstructured.NestedString(u.Object, path...)
+	if err != nil {
+		return "", fmt.Errorf("failed to access serviceAccountID field in BigQueryConnection %s/%s: %w", key.Namespace, key.Name, err)
+	}
+	if !found {
+		pathStr := strings.Join(path, ".")
+		return "", fmt.Errorf("BigQueryConnection %s/%s is not ready - field '%s' is missing", key.Namespace, key.Name, pathStr)
```
yuwenma, line 194: nit: the `types.NamespacedName` has a string which combines name and namespace

pkg/controller/iam/iamclient/helpers.go
```diff
@@ -54,6 +56,26 @@ func ResolveMemberIdentity(ctx context.Context, member v1beta1.Member,
 		return string(member), nil
 	}
 
+	if err := memberFrom.Validate(); err != nil {
+		return id, err
+	}
+
+	if id, err := tryResolveTFMemberReference(ctx, memberFrom, namespace, tfIAMClient); err != nil {
```
yuwenma, line 63: I like the name, though it is a little bit misleading. For example, the tryResolveTFMemberRefer contains sqlinstance  which could be a direct resource, so it's not "successfully resolved member reference from TF-based resources".

yuwenma (commented):
/approve

Some nits, overall looks great!

yuwenma (commented):
/approve

Some nits, overall looks great!

pkg/controller/iam/iamclient/helpers.go
jingyih: Hmm... I named the function this way because all 4 resources referenced within this function depend on the TF-based service mapping to determine the service account. As an example, for SQLInstance it uses this configuration.
https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/3530c83a5e0d331640ec2160675d80336fad9c53/config/servicemappings/sql.yaml#L56-L58

/cc @jasonvigil When SQLInstance is switched to a direct resource, do we still depend on service mapping when the SQLInstance is referenced by an IAM resource? e.g.
```
apiVersion: iam.cnrm.cloud.google.com/v1beta1
kind: IAMPolicyMember
metadata:
  name: iampolicymember-sample-memberref
spec:
  memberFrom:
    sqlInstanceRef:
      name: xxx
```

apis/bigqueryconnection/v1beta1/connection_reference.go
jingyih: Cool! Updated.

apis/bigqueryconnection/v1beta1/connection_reference.go
```diff
@@ -149,8 +150,49 @@ func (r *BigQueryConnectionConnectionRef) NormalizedExternal(ctx context.Context
...
 	}
 	if actualExternalRef == "" {
-		return "", fmt.Errorf("BigQueryConnectionConnection is not ready yet.")
+		return "", fmt.Errorf("BigQueryConnectionConnection is not ready yet")
 	}
 	r.External = actualExternalRef
 	return r.External, nil
 }
+
+func ResolveServiceAccountID(ctx context.Context, reader client.Reader, namespace string, ref *iamv1beta1.BigQueryConnectionConnectionMemberReference) (string, error) {
```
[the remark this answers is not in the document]
jingyih: I added a comment. Referencing between resources probably needs a better design but it is outside the scope of this PR.

pkg/apis/iam/v1beta1/krm_types.go
```diff
@@ -62,6 +74,14 @@ type MemberReference struct {
 	Name      string `json:"name"`
 }
 
+type BigQueryConnectionConnectionMemberReference struct {
```
yuwenma, line 77: Suggest defining this struct in BQCC API to avoid cyclic dependencies and make it reusable.

yuwenma (commented):
One last comment otherwise looks good

jingyih:
Rebased to pick up the fix for the broken HEAD.

pkg/apis/iam/v1beta1/krm_types.go
jingyih: Done!

PR #3234, opened 2024-11-21
feat: A new parent interface

yuwenma:
### Change description

Fixes #

### Tests you have done


apis/common/parent/projectlocation.go
```diff
@@ -0,0 +1,88 @@
...
+	"context"
+	"fmt"
+	"strings"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ Parent = &ProjectAndLocationParent{}
+
+type ProjectAndLocationParent struct {
```
yuwenma, line 26: We can merge `ProjectAndLocationParent` and `ProjectAndLocationRef`, it makes this file shorter. But it adds additional complexity to the controller on error handling,  and the order of the function calls to set the `external`.

apis/bigqueryconnection/v1beta1/connection_reference.go
```diff
@@ -28,49 +29,47 @@ import (
...
 		if err != nil {
 			return nil, err
 		}
-		// Validate spec parent and resourceID field if the resource is already reconcilied with a GCP Connection resource.
-		if projectID != actualProject {
-			return nil, fmt.Errorf("spec.projectRef changed, expect %s, got %s", projectID, actualProject)
-		}
-		if location != actualLocation {
-			return nil, fmt.Errorf("spec.location changed, expect %s, got %s", location, actualLocation)
+		if err := desiredParent.MatchActual(actualParent); err != nil {
```
justinsb, line 58: Suggest just comparing `String()` of the ID

apis/bigqueryconnection/v1beta1/connection_reference.go
```diff
@@ -88,36 +87,13 @@ type BigQueryConnectionConnectionRef struct {
 	Namespace string `json:"namespace,omitempty"`
 }
 
-func parseExternal(external string) (string, string, string, error) {
+func ParseConnectionExternal(external string) (*parent.ProjectAndLocationParent, string, error) {
```
justinsb, line 90: Suggest this should return `(*ConnectionIdentity,error)`  (and not the parent split out)

apis/bigqueryconnection/v1beta1/connection_types.go
```diff
@@ -23,20 +24,10 @@ import (
...
-	// Immutable.
-	// +required
-	Location string `json:"location"`
-}
-
 // BigQueryConnectionConnectionSpec defines the desired state to connect BigQuery to external resources
 // +kcc:proto=google.cloud.bigquery.connection.v1.Connection
 type BigQueryConnectionConnectionSpec struct {
-	Parent `json:",inline"`
+	*parent.ProjectAndLocationRef `json:",inline"`
```
justinsb, line 30: I thought you wanted to embed `ConnectionIdentity` here instead

apis/common/parent/interface.go
```diff
@@ -0,0 +1,37 @@
...
+	// Verify that the desired parent (from .spec) matches the actual parent in .status.externalRef.
+	// This ensures the parent remains unchanged.
+	// We currently don't enforce parent immutability using a webhook or CRD CEL due to legacy reasons.
+	MatchActual(Parent) error
+}
+
+// ParentBuilder builds a Parent object from a ParentRef.
+// - ParentRef is the Config Connector API reference for identifying a resource's logical parent.
+// - The Parent object provides helper functions for parent-related logic in direct reconciliation.
+type ParentBuilder interface {
```
justinsb, line 34: Maybe the krm object should implement `ResolveIdentity(ctx context.Context, reader client.Reader) (Identity, error)`

If we want Identity to be typed, it would have to be something like this:

```
type ObjectWithIdentity[T] interface {
  ResolveIdentity(ctx contetx.Context, reader client.Reader) (*T, error)
}
```

apis/common/parent/project.go
```diff
@@ -0,0 +1,132 @@
...
+	/* The `name` field of a `Project` resource. */
+	Name string `json:"name,omitempty"`
+	/* The `namespace` field of a `Project` resource. */
+	Namespace string `json:"namespace,omitempty"`
+}
+
+// Builds a the ProjectParent from ProjectRef.
+// If `projectRef.external` is given, parse projectID from External, otherwise find the ConfigConnector project
+// according to `projectRef.name` and `projectRef.namespace`.
+func (p *ProjectRef) Build(ctx context.Context, reader client.Reader, othernamespace string, parent Parent) error {
```
justinsb, line 50: I suspect we don't need this function.  We need a function that resolves a ProjectRef to a ProjectIdentity, and ProjectIdentity should have a function that can return the ProjectID.  It might be rename this function to map from ProjectRef -> ProjectIdentity.

We would have 3 concepts: Foo, FooRef and FooIdentity

FooIdentity is essentially the fully-qualified name, I don't know if we should name it FooLink or something like that instead, but I do think that just using these 3 concepts will simplify our code a lot.

apis/common/parent/project.go
```diff
@@ -0,0 +1,132 @@
...
+		return fmt.Errorf("reading spec.resourceID from %v %v/%v: %w", project.GroupVersionKind().Kind, p.Namespace, p.Name, err)
+	}
+	if projectID == "" {
+		projectID = project.GetName()
+	}
+	projectParent.ProjectID = projectID
+	return nil
+}
+
+func ParseProjectParent(external string) (*ProjectParent, error) {
```
justinsb, line 108: We _could_ make this a method on FooIdentity, something like `ParseExternal() error`

apis/common/parent/projectlocation.go
justinsb: I think we want Foo, FooRef and FooIdentity.  I think this is FooIdentity, and ProjectAndLocationRef is FooRef

apis/common/parent/projectlocation.go
```diff
@@ -0,0 +1,88 @@
...
+		return fmt.Errorf("spec.location changed, desired %s, actual %s", p.Location, actual.Location)
+	}
+	return nil
+}
+
+var _ ParentBuilder = &ProjectAndLocationRef{}
+
+// ProjectAndLocationParent specifies the resource's GCP hierarchy (Project/Folder/Organization) and its geographical location.
+// +kubebuilder:object:generate:=true
+type ProjectAndLocationRef struct {
```
justinsb, line 53: I think this is actually ProjectAndLocationIdentity, because we want to embed this into FooIdentity, which gets embedded into Foo

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_bigqueryconnectionconnections.bigqueryconnection.cnrm.cloud.google.com.yaml
```diff
@@ -293,10 +289,6 @@ spec:
                     description: The `projectID` field of a project, when not managed
                       by Config Connector.
                     type: string
-                  kind:
```
justinsb, line 296: Sadly we can't do this, this is a breaking API change.  We can't break resources that specify `kind`, even though we (I) introduced it by mistake.

pkg/controller/direct/bigqueryconnection/connection_controller.go
```diff
@@ -76,7 +75,7 @@ func (m *model) AdapterForObject(ctx context.Context, reader client.Reader, u *u
 	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &obj); err != nil {
 		return nil, fmt.Errorf("error converting to %T: %w", obj, err)
 	}
-	connectionRef, err := krm.NewBigQueryConnectionConnectionRef(ctx, reader, obj)
+	connectionIdentity, err := krm.NewConnectionIdentity(ctx, reader, obj)
```
justinsb, line 78: This is the one where _maybe_ we could call `obj.ResolveIdentity` ... maybe :-)

pkg/controller/direct/bigqueryconnection/connection_controller.go
```diff
@@ -211,29 +204,16 @@ func (a *Adapter) Create(ctx context.Context, createOp *directbase.CreateOperati
...
 		return mapCtx.Err()
 	}
-
-	parent, err := a.id.Parent()
-	if err != nil {
-		return err
-	}
 	req := &pb.CreateConnectionRequest{
-		Parent:     parent,
+		Parent:     a.id.Parent.String(),
```
justinsb, line 208: I like this approach, but one thing we will have to be careful of is that we have a "partial ID" here, ResourceID is empty, the parent fields are set.  There are going to be bugs because of this.

It might be easier to resolve the parent directly here, perhaps with a method on the embedded ProjectAndLocationID in BigQueryConnectionID

pkg/controller/direct/bigqueryconnection/connection_controller.go
```diff
@@ -243,14 +223,14 @@ func (a *Adapter) Create(ctx context.Context, createOp *directbase.CreateOperati
 		return mapCtx.Err()
 	}
 
-	tokens := strings.Split(created.Name, "/")
-	parent, err = a.id.Parent()
+	_, name, err := krm.ParseConnectionExternal(created.Name)
```
justinsb, line 226: Looking at this, I think it would read well if it was a method on BigQueryConnectionID, like this:

```
name := &BigQueryConnectionID{}
if err := name.ParseSelfLink(created.Name); err != nil {
....
```

pkg/controller/direct/bigqueryconnection/connection_controller.go
```diff
@@ -315,22 +292,14 @@ func (a *Adapter) Export(ctx context.Context) (*unstructured.Unstructured, error
...
-	}
-	if parent != "" {
-		tokens := strings.Split(parent, "/")
-		if len(tokens) == 4 && tokens[0] == "projects" && tokens[2] == "locations" {
-			obj.Spec.ProjectRef = &refs.ProjectRef{Name: tokens[1]}
-			obj.Spec.Location = tokens[3]
-		}
+		return nil, err
 	}
+	obj.Spec.ResourceID = &id
```
justinsb, line 300: If we embed ID into Spec, I think these three lines will just be `obj.Spec.ID = id`

apis/bigqueryconnection/v1beta1/connection_reference.go
yuwenma: This `MatchActual` is more for KCC users. 
My main concern about comparing `String` is that it cannot give users enough information to help them fix their bad config.
For example, the parent is build from `spec.ProjectRef.External` and `spec.Location` (for some other resources it may also have `spec. FooRef.External` `spec.BarRef.External`), it is unclear for users to know which field is misconfigured by just seeing something like`expected format projects/<projectID>/location/<lcoation>/foos/<fooid>/bars/<bar>, got ....`

apis/bigqueryconnection/v1beta1/connection_reference.go
yuwenma: Sure. I don't have a preference. It used to be handy to split them because we don't have the identity.  😄

apis/bigqueryconnection/v1beta1/connection_reference.go
justinsb: Well I think I would make the message "Can't change object identity; existing identity is %q, new identity is %q" so it would be pretty easy to figure out what has changed, but ... I can send a PR on top of this - let's get this in!

apis/bigqueryconnection/v1beta1/connection_reference.go
```diff
@@ -28,49 +29,47 @@ import (
...
 	}
-	projectID := projectRef.ProjectID
-	if projectID == "" {
-		return nil, fmt.Errorf("cannot resolve project")
-	}
-	// Get location
-	location := obj.Spec.Location
 
 	// Get desired connection ID from spec
 	desiredID := direct.ValueOf(obj.Spec.ResourceID)
```
justinsb, line 49: Do we need to fallback to name here?

justinsb:
So I think two blockers, and everything else we can tackle in follow-up PRs:

* I think we need a fallback to name?
* I don't think you can remove the kind field

yuwenma:
> So I think two blockers, and everything else we can tackle in follow-up PRs:
> 
> * I think we need a fallback to name?
> * I don't think you can remove the kind field

I reverted the changes to BQCC. So this PR only contains the interface and the implemented `ProjectParent` and `ProjectAndLocationParent`

justinsb (approved):

justinsb (approved):

justinsb (approved):

PR #3242, opened 2024-11-22
chore: check generated files under /operator as part of presubmit validation

jingyih:


yuwenma (commented):
/lgtm
/approve

Thanks!

PR #3245, opened 2024-11-22
Support env var to only compare http log events with given URL prefix

maqiuyujoyce:
### Change description

When there are a lot of dependent non-direct resources in the test case, the comparison between real GCP http log and mock GCP http log can be messy. With the new env var `ONLY_COMPARE_URL_PREFIX`, we can focus on the comparison of the target resource/API in the http log events.

### Tests you have done

### DIFF
--- a/pkg/test/utils.go
+++ b/pkg/test/utils.go
@@ -86,30 +86,30 @@ func extractEventsWithURLPrefix(allEvents, urlPrefix string) string {
 }
 
 // CompareGoldenFile performs a file comparison for a golden test.
-func CompareGoldenFile(t *testing.T, path, got string, normalizers ...func(s string) string) {
+func CompareGoldenFile(t *testing.T, p, fullGot string, normalizers ...func(s string) string) {
 	writeGoldenOutput := os.Getenv("WRITE_GOLDEN_OUTPUT") != ""
 
 	for _, normalizer := range normalizers {
-		got = normalizer(got)
+		fullGot = normalizer(fullGot)
 	}
-	gotForComp := got
+	got := fullGot
 
-	wantBytes, err := os.ReadFile(path)
+	wantBytes, err := os.ReadFile(p)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/test/utils.go
```diff
@@ -65,15 +66,35 @@ func MustReadFile(t *testing.T, p string) []byte {
...
+
 // CompareGoldenFile performs a file comparison for a golden test.
-func CompareGoldenFile(t *testing.T, p string, got string, normalizers ...func(s string) string) {
+func CompareGoldenFile(t *testing.T, path, got string, normalizers ...func(s string) string) {
 	writeGoldenOutput := os.Getenv("WRITE_GOLDEN_OUTPUT") != ""
 
 	for _, normalizer := range normalizers {
 		got = normalizer(got)
 	}
+	gotForComp := got
```
justinsb, line 95: Nit: `gotForComparison`.  (And maybe easier to reuse `got`, because comparing `got` and `want` is a well-known pattern.  So easier to stash the `got` value e.g. `fullGot := got`)

pkg/test/utils.go
```diff
@@ -65,15 +66,35 @@ func MustReadFile(t *testing.T, p string) []byte {
...
 
-	wantBytes, err := os.ReadFile(p)
+	wantBytes, err := os.ReadFile(path)
```
justinsb, line 97: Amusing counterexample to the "no abbreviations" guideline :-)  We tend to use `p` instead of `path` so we don't mask the `path` package.  But normally we should be using `filepath` instead, so maybe I should follow the "no abbreviations" rule for `path` also :-)

maqiuyujoyce: Good point! Will change it back.

pkg/test/utils.go
maqiuyujoyce: Yep, using `fullGot` sounds like a cleaner and simpler option. Will update it!

PR #3249, opened 2024-11-23
mockgcp: pin googleapis for GRPC generation also

justinsb:
- **mockgcp: ping googleapis for GRPC generation also**
- **mockgcp: update generated GRPC bindings**


yuwenma (commented):
/lgtm
/approve

PR #3227, opened 2024-11-20
feat: Turn on ComputeTCPProxy direct controller

gemmahou:
### Change description

Turn on ComputeTCPProxy direct controller

### Tests you have done

samples test passed:
```
go test -v -tags=integration ./config/tests/samples/create -test.run TestAll -run-tests global-target-tcp-proxy
...
ok      github.com/GoogleCloudPlatform/k8s-config-connector/config/tests/samples/create 126.238s
```

- [x] Run `make ready-pr` to ensure this PR is ready for review.
- [x] Perform necessary E2E testing for changed resources.

PR #3252, opened 2024-11-25
fix: Log error when update is not supported for regional TCP Proxy

gemmahou:
### Change description

Fixes b/380362493

Address https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3236#discussion_r1853057354

Log error when update is not supported(by API), but fields get updated by user. However, if this happens, the resource will remain in error state. The only way to fix it might be to delete the resource and recreate a new one. Users won't be able to just revert the change.

I should also add a warning message that in our public doc, this will be done in #3227 

As mentioned in above discussion, there're several other ways to handle this:
1. GCP approach: do nothing on our side, and let the GCP API to handle the potential error(Users won't be able to revert the change too)
2. TF approach: trigger a `forceNew` feature to delete the old resource and create a new one with updated fields
3. Immutable check: Mark all fields in resource as immutable if the API doesn't support update. But it won't work for this case because global and regional share the CRD, global resource can be updated

### Tests you have done

Tried to update `proxyHeader` field in regional tcp proxy and the error looks like:
```
{"severity":"info","timestamp":"2024-11-25T21:26:49.428Z","msg":"resource is not ready","kind":"ComputeTargetTCPProxy","name":"computetargettcpproxy-hcmu7thn7lt2pmq","conditions":[{"lastTransitionTime":"2024-11-25T21:26:40Z","message":"Update call failed: error updating: update operation not supported for resource ComputeTargetTCPProxy hcmu7thn7lt2pmq/computetargettcpproxy-hcmu7thn7lt2pmq: projects/mock-project/regions/europe-west4/targetTcpProxies/computetargettcpproxy-hcmu7thn7lt2pmq","reason":"UpdateFailed","status":"False","type":"Ready"}]}
```

### DIFF
--- a/pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
+++ b/pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
@@ -20,6 +20,8 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
+
 	"google.golang.org/api/option"
 
 	gcp "cloud.google.com/go/compute/apiv1"
@@ -111,7 +113,7 @@ func (m *targetTCPProxyModel) AdapterForObject(ctx context.Context, reader clien
 	if err != nil {
 		return nil, fmt.Errorf("get ComputeTargetTCPProxyAdapter parent %s: %w", computeTargetTCPProxyRef.External, err)
 	}
-	location := parent.Region
+	location := parent.Location
 
 	// Handle API/TF default values
 	if obj.Spec.ProxyBind != nil && *obj.Spec.ProxyBind == false {
@@ -181,13 +183,13 @@ func (a *targetTCPProxyAdapter) Create(ctx context.Context, createOp *directbase
 	if err != nil {
 		return fmt.Errorf("get ComputeTargetTCPProxy parent %s: %w", a.id.External, err)
 	}
-	region := parent.Region
+	location := parent.Location
 
 	tokens := strings.Split(a.id.External, "/")
 	targetTCPProxy.Name = direct.LazyPtr(tokens[len(tokens)-1])
 
 	op := &gcp.Operation{}
-	if region == "global" {
+	if location == "global" {
 		req := &computepb.InsertTargetTcpProxyRequest{
 			Project:                parent.ProjectID,
 			TargetTcpProxyResource: targetTCPProxy,
@@ -196,7 +198,7 @@ func (a *targetTCPProxyAdapter) Create(ctx context.Context, createOp *directbase
 	} else {
 		req := &computepb.InsertRegionTargetTcpProxyRequest{
 			Project:                parent.ProjectID,
-			Region:                 region,
+			Region:                 location,
 			TargetTcpProxyResource: targetTCPProxy,
 		}
 		op, err = a.regionalTargetTcpProxiesClient.Insert(ctx, req)
@@ -258,13 +260,18 @@ func (a *targetTCPProxyAdapter) Update(ctx context.Context, updateOp *directbase
 	if err != nil {
 		return fmt.Errorf("get ComputeTargetTCPProxy parent %s: %w", a.id.External, err)
 	}
-	region := parent.Region
+	location := parent.Location
+
+	// Regional API does not support Update
+	if location != "global" {
+		return fmt.Errorf("update operation not supported for resource %v %v",
+			a.desired.GroupVersionKind(), k8s.GetNamespacedName(a.desired))
+	}
 
 	tokens := strings.Split(a.id.External, "/")
 	targetTCPProxy.Name = direct.LazyPtr(tokens[len(tokens)-1])
 
-	// Regional API does not support Update
-	if !reflect.DeepEqual(targetTCPProxy.ProxyHeader, a.actual.ProxyHeader) && region == "global" {
+	if !reflect.DeepEqual(targetTCPProxy.ProxyHeader, a.actual.ProxyHeader) {
 		setProxyHeaderReq := &computepb.SetProxyHeaderTargetTcpProxyRequest{
 			Project: parent.ProjectID,
 			TargetTcpProxiesSetProxyHeaderRequestResource: &computepb.TargetTcpProxiesSetProxyHeaderRequest{ProxyHeader: targetTCPProxy.ProxyHeader},
@@ -283,7 +290,7 @@ func (a *targetTCPProxyAdapter) Update(ctx context.Context, updateOp *directbase
 		log.V(2).Info("successfully updated ComputeTargetTCPProxy proxy header", "name", a.id.External)
 	}
 
-	if !reflect.DeepEqual(targetTCPProxy.Service, a.actual.Service) && region == "global" {
+	if !reflect.DeepEqual(targetTCPProxy.Service, a.actual.Service) {
 		setBackendServiceReq := &computepb.SetBackendServiceTargetTcpProxyRequest{
 			Project: parent.ProjectID,
 			TargetTcpProxiesSetBackendServiceRequestResource: &computepb.TargetTcpProxiesSetBackendServiceRequest{Service: targetTCPProxy.Service},
@@ -347,11 +354,11 @@ func (a *targetTCPProxyAdapter) Delete(ctx context.Context, deleteOp *directbase
 	if err != nil {
 		return false, fmt.Errorf("get ComputeTargetTcpProxy parent %s: %w", a.id.External, err)
 	}
-	region := parent.Region
+	location := parent.Location
 
 	op := &gcp.Operation{}
 	tokens := strings.Split(a.id.External, "/")
-	if region == "global" {
+	if location == "global" {
 		delReq := &computepb.DeleteTargetTcpProxyRequest{
 			Project:        parent.ProjectID,
 			TargetTcpProxy: tokens[len(tokens)-1],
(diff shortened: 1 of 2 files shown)


maqiuyujoyce:
Thank you for the update, @gemmahou ! Could you also share the resource status of the resource (e.g. output of `kubectl describe`) after running the update?

gemmahou:
> Thank you for the update, @gemmahou ! Could you also share the resource status of the resource (e.g. output of `kubectl describe`) after running the update?

Here's the output: https://screenshot.googleplex.com/AW5HajtWGGwkUbK.png

pkg/k8s/errors.go
```diff
@@ -233,3 +233,18 @@ func (e *ImmutableFieldsMutationError) Error() string {
 func NewImmutableFieldsMutationError(immutableFields []string) *ImmutableFieldsMutationError {
 	return &ImmutableFieldsMutationError{immutableFields}
 }
+
+type UpdateNotSupportedError struct {
+	resourceGVK schema.GroupVersionKind
+	namespace   types.NamespacedName
```
maqiuyujoyce, line 239: `namespacedName`?
This is the NamespacedName in the format of `<namespace>/<resourceName>`. It's not just the namespace. In your test log, it is `hcmu7thn7lt2pmq/computetargettcpproxy-hcmu7thn7lt2pmq`.

pkg/k8s/errors.go
```diff
@@ -233,3 +233,18 @@ func (e *ImmutableFieldsMutationError) Error() string {
...
+	resource    string
```
maqiuyujoyce, line 240: In the implementation, this field contains the relative resource name of the GCP resource. I'm hesitant to mention it in the error message because this error is at the KCC level, not the GCP level. If one day the API team starts supporting update for the resource, the error message might become confusing because KCC is saying that an updatable GCP resource cannot be updated.

maqiuyujoyce (commented):
@gemmahou thank you for supporting and testing this edge case. Have you reached out to the API team to understand context about the unsupported Update? In the PR description, you mentioned multiple options, but you didn't mention why we landed on this option. I'm not super sure if blocking it at KCC level is the right solution without more context.

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
```diff
@@ -260,11 +262,16 @@ func (a *targetTCPProxyAdapter) Update(ctx context.Context, updateOp *directbase
 	}
 	region := parent.Region
 
+	// Regional API does not support Update
+	if region != "global" {
```
yuwenma, line 266: This could be error prone and hard to maintain. Can we separate 'global' from region or use a different field?

gemmahou:
> @gemmahou thank you for supporting and testing this edge case. Have you reached out to the API team to understand context about the unsupported Update? In the PR description, you mentioned multiple options, but you didn't mention why we landed on this option. I'm not super sure if blocking it at KCC level is the right solution without more context.

Hmmm I do not think reaching API team is helpful and they have the design for a long time. One different from them is global and regional are separate resources. I think I want to use this PR to start the discussion about how do we want to handle the case when Update is not supported, this is a common case I think. Logging an error is straight forward but may not be what we'd like...

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
gemmahou: Can we separate 'global' from 'region'

Do you mean separate them into two resources? I think the reason we decided to support them in one CRD is because it's align with other regional resources, and easier for user adoption. In addition, since the resource will be used as a reference in CFR, we do not need to add another ref resource type.

or use a different field

Not sure why that would help? The shared and global mutable fields are "service" and "proxyHead".

maqiuyujoyce:
> > @gemmahou thank you for supporting and testing this edge case. Have you reached out to the API team to understand context about the unsupported Update? In the PR description, you mentioned multiple options, but you didn't mention why we landed on this option. I'm not super sure if blocking it at KCC level is the right solution without more context.
> 
> Hmmm I do not think reaching API team is helpful and they have the design for a long time. One different from them is global and regional are separate resources. I want to use this PR to start the discussion about how do we want to handle the case when Update is not supported, this is a common case I think. Logging an error is straight forward but may not be what we'd like...

If the API team plans to support the Update in the near future (which is something I'd want to check with API team about), then let it fail at the API level might be the cleaner and cheaper solution because we don't have to adjust based on the API team's change.

gemmahou:
Chatted with @maqiuyujoyce in our 1:1 offline.

the "reversibility"(revert to original configuration without delete and re-create) isn't a primary concern at this time, because we need to re-create KCC-managed objects that in an error state. 

I'll investigate how the API handles update to regional resources and see whether we can leverage the API for error handling.  This will be easier for maintenance, in case API behavior changes later. Additionally, I'll add a warning message to our documentation.

pkg/k8s/errors.go
gemmahou: Done

pkg/k8s/errors.go
gemmahou: Done

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
gemmahou: Done, use `location` instead.

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
```diff
@@ -258,13 +260,18 @@ func (a *targetTCPProxyAdapter) Update(ctx context.Context, updateOp *directbase
 	if err != nil {
 		return fmt.Errorf("get ComputeTargetTCPProxy parent %s: %w", a.id.External, err)
 	}
-	region := parent.Region
+	location := parent.Location
+
+	// Regional API does not support Update
+	if location != "global" {
+		return k8s.NewUpdateNotSupportedError(a.desired.GroupVersionKind(),
```
yuwenma, line 267: Can I ask why we want to define a new structed type?

yuwenma (commented):
some code need update after the renaming.

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
gemmahou: I feel that could be a common thing we want to handle as we migrate more resources to direct. Did a quick [code search](https://github.com/search?q=repo%3AGoogleCloudPlatform%2Fk8s-config-connector%20skipUpdate&type=code) and found several "skipUpdate" due to no API support. We used "immutable fields" to prevent update, but since we removed immutability webhook for direct resources, the CEL rule has limit guarantee on only required fields. I feel might be better to create a new error type, in case we have more similar cases in the future. But I'm open to suggestions.

gemmahou: Also worth mentioning that regional TCP Proxy is a new direct resource, so luckily we won't need to worry about backward compatibility. I look forward to getting users feedback from this resource, and explore how to handle "skipUpdate" for other resources.  (For existing resources, we'll need to consider backward compatibility carefully.)

maqiuyujoyce:
/lgtm
/approve
/hold

Hold because Yuwen has another comment.

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
yuwenma: We could, but maybe not in this PR. Because this new error in k8s does not affect the reconcile lifecycle. If you want to use a generic error, we should not put it inside k8s. Besides, the error message needs to be more carefully designed since this is user facing and reflects the reconciliation difference.

yuwenma (commented):
/approve

One nit to revert the k8s errors. Other than that, everything looks great!

pkg/k8s/errors.go
```diff
@@ -233,3 +233,17 @@ func (e *ImmutableFieldsMutationError) Error() string {
 func NewImmutableFieldsMutationError(immutableFields []string) *ImmutableFieldsMutationError {
 	return &ImmutableFieldsMutationError{immutableFields}
 }
+
+type UpdateNotSupportedError struct {
+	resourceGVK    schema.GroupVersionKind
```
yuwenma, line 238: nit: GVK may not be appropriate (maybe GK?), but it really refers to how we want to handle this error in the reconcile. Suggest adding this error in k8s only when we plan to handle this error.

pkg/k8s/errors.go
```diff
@@ -233,3 +233,17 @@ func (e *ImmutableFieldsMutationError) Error() string {
...
+	namespacedName types.NamespacedName
+}
+
+func (e *UpdateNotSupportedError) Error() string {
```
yuwenma, line 242: We need code level guarantee that this is only called in `Update`. Otherwise, I suggest we rename it to something else.

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
```diff
@@ -258,13 +260,18 @@ func (a *targetTCPProxyAdapter) Update(ctx context.Context, updateOp *directbase
 	if err != nil {
 		return fmt.Errorf("get ComputeTargetTCPProxy parent %s: %w", a.id.External, err)
 	}
-	region := parent.Region
+	location := parent.Location
+
+	// Regional API does not support Update
+	if location != "global" {
+		return fmt.Errorf("update operation not supported for resource %v %v",
+			a.desired.GroupVersionKind().Kind, k8s.GetNamespacedName(a.desired))
```
yuwenma, line 268: nit: I think a.desired.GroupVersionKind() has a `String()` builtin.

PR #3253, opened 2024-11-26
tool: simplify controller builder commands by removing redundant flags

jingyih:
- Set a new default for the `--proto-source-path` flag.
 - Remove redundant flags with default values for simpler commands.
 - Update the contributor guide to reflect these changes.

### DIFF
--- a/docs/develop-resources/deep-dives/2-define-apis.md
+++ b/docs/develop-resources/deep-dives/2-define-apis.md
@@ -4,17 +4,17 @@ Config Connector builds the API using the Google Cloud Client proto.
 
 ## 2.1 Build the Google Cloud Proto (one-off step)
 
-This step generates the Google Cloud Client proto in a single file under ./dev/tools/proto-to-mapper/build. This should be a one-off command only when you need to update the proto or the first time you develop the direct resource.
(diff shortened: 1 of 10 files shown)


docs/develop-resources/deep-dives/2-define-apis.md
```diff
@@ -37,11 +36,6 @@ go run main.go generate-types \
 
 The proto name of the GCP service, you can find them in [https://github.com/googleapis/googleapis.git](https://github.com/googleapis/googleapis.git). For example, the SQL service is [https://github.com/googleapis/googleapis/tree/master/google/cloud/sql/v1beta4](https://github.com/googleapis/googleapis/tree/master/google/cloud/sql/v1beta4). The `–service` should be `google.cloud.sql.v1beta4`
 
-
-* `--proto-source-path`
```
yuwenma, line 41: Is this flag removed? Maybe mark it as a optional? But non blocker

yuwenma (commented):
/lgtm
/approve

Thank you! This is much cleaner

docs/develop-resources/deep-dives/2-define-apis.md
jingyih: It is not removed; it's optional to developers because the tool provides the correct default value for it.

PR #3255, opened 2024-11-26
Ignore NotFound error when deleting PrivilegedAccessManagerEntitlement

maqiuyujoyce:
### Change description

The deletion of PrivilegedAccessManagerEntitlement resource is stuck if the underlying GCP resource doesn't exist. But this deletion should go through successfully. Ignoring the resource not found error in the deletion fixes this error.

### Tests you have done

### DIFF
--- a/pkg/controller/direct/privilegedaccessmanager/entitlement_controller.go
+++ b/pkg/controller/direct/privilegedaccessmanager/entitlement_controller.go
@@ -424,6 +424,9 @@ func (a *Adapter) Delete(ctx context.Context, deleteOp *directbase.DeleteOperati
 	req := &privilegedaccessmanagerpb.DeleteEntitlementRequest{Name: a.id.FullyQualifiedName()}
 	op, err := a.gcpClient.DeleteEntitlement(ctx, req)
 	if err != nil {
+		if direct.IsNotFound(err) {
+			return false, nil
+		}
 		return false, fmt.Errorf("error deleting PrivilegedAccessManagerEntitlement %s: %w", a.id.FullyQualifiedName(), err)
 	}
 


maqiuyujoyce:
A test step that deletes the underlying GCP resource first and then delete the KRM resource will help us detect such error.

yuwenma:
What is the real PAM service behavior if the DELETE request is on a non-exist resource? 
 
> ... is stuck if the underlying GCP resource doesn't exist.
Is this from the dynamic test? Do you know why the resource does not exist?

maqiuyujoyce:
> What is the real PAM service behavior if the DELETE request is on a non-exist resource?
> 
> > ... is stuck if the underlying GCP resource doesn't exist.
> > Is this from the dynamic test? Do you know why the resource does not exist?

It'll return a NOT_FOUND error:
```
$ gcloud beta pam entitlements delete pam-3zpy4jkelr4buo --location global
Parsed [entitlement] resource: projects/[myproject]/locations/global/entitlements/pam-3zpy4jkelr4buo
All grants associated with the entitlement are also deleted upon deleting the entitlement.

Do you want to continue (Y/n)?  Y

ERROR: (gcloud.beta.pam.entitlements.delete) NOT_FOUND: Resource 'projects/[myproject]/locations/global/entitlements/pam-3zpy4jkelr4buo' was not found. This command is authenticated as [user] which is the active account specified by the [core/account] property
```

yuwenma:
> > What is the real PAM service behavior if the DELETE request is on a non-exist resource?
> > > ... is stuck if the underlying GCP resource doesn't exist.
> > > Is this from the dynamic test? Do you know why the resource does not exist?
> 
> It'll return a NOT_FOUND error:
> 
> ```
> $ gcloud beta pam entitlements delete pam-3zpy4jkelr4buo --location global
> Parsed [entitlement] resource: projects/[myproject]/locations/global/entitlements/pam-3zpy4jkelr4buo
> All grants associated with the entitlement are also deleted upon deleting the entitlement.
> 
> Do you want to continue (Y/n)?  Y
> 
> ERROR: (gcloud.beta.pam.entitlements.delete) NOT_FOUND: Resource 'projects/[myproject]/locations/global/entitlements/pam-3zpy4jkelr4buo' was not found. This command is authenticated as [user] which is the active account specified by the [core/account] property
> ```

Could you elaborate more on this change? If the GCP server returns a not found error, the MockGCP should not silo that .

maqiuyujoyce:
@yuwenma this change is not about mockGCP (sorry for the confusing communication offline), but is about the direct controller.

Deletion of a KRM object should be successful when the underlying GCP resource is gone. If the underlying GCP resource never exists, then the deletion should go through successfully. So we should ignore the NOT_FOUND error returned by the GCP API.

PR #3259, opened 2024-11-26
Update policy-least-privilege.yaml

tarynlucas:
minor change to update incorrect role name, also remove the square brackets as they cause issues when using variables on cgc docs

### DIFF
--- a/config/samples/tutorials/hardening-your-cluster/policy-least-privilege.yaml
+++ b/config/samples/tutorials/hardening-your-cluster/policy-least-privilege.yaml
@@ -17,9 +17,9 @@ kind: IAMPolicyMember
 metadata:
   name: policy-least-privilege
 spec:
-  member: serviceAccount:[SA_NAME]@[PROJECT_ID].iam.gserviceaccount.com
-  role: roles/container.nodeServiceAccount
+  member: serviceAccount:SA_NAME@PROJECT_ID.iam.gserviceaccount.com
+  role: roles/container.defaultNodeServiceAccount
   resourceRef:
     kind: Project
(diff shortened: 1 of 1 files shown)


yuwenma:
/lgtm
/approve
/hold

Thank you, @tarynlucas ! This looks good to me. I'd like to have @maqiuyujoyce take another look to make sure if the square bracket is introduced for some legacy reasons, because this sample runs as a source against the presubmit check.

maqiuyujoyce:
The change should be fine.

I don't have context about this sample so did some research:
1. It is not tested by our integration test.
2. It is not referenced by any public doc (based on some quick search).
3. There is no activity about this sample since it was added in 2022.

My question is: Do we still need it? Can we just delete the sample to reduce our maintenance burden?

yuwenma:
> The change should be fine.
> 
> I don't have context about this sample so did some research:
> 
> 1. It is not tested by our integration test.
> 2. It is not referenced by any public doc (based on some quick searches but I might miss something).
> 3. There is no activity about this sample since it was added in 2022.
> 
> My question is: Do we still need it? Can we just delete the sample to reduce our maintenance burden?

Good to know this is not used. Please go ahead and remove them.

maqiuyujoyce:
Did some further searches, and it turns out that these tutorials are referenced by GKE doc: https://cloud.google.com/kubernetes-engine/docs/how-to/hardening-your-cluster#config-connector_1

Let's keep it as is for now.

In the long term, I suggest we don't maintain samples we don't test (at least I am not aware of any testing of these samples but I might be missing something).

PR #3261, opened 2024-11-26
wip:fix: don't assign external

acpana:
The canonical example for reference handling is that we either pass in:

* the external format OR
* the name of the reference

Consider this example

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/bcf1e3e3fcce891e4ba56104533acb77e61ea0b3/pkg/test/resourcefixture/testdata/basic/bigqueryanalyticshub/v1alpha1/bigqueryanalyticshublisting-base/create.yaml#L23-L25

If the NormlizedExternal checks that only one of `r.External` or `r.Name` is set BUT assigns `r.External` at the end, future calls to this function will error out. IOW, the function is not idempotent and I think it should be because we may be calling this function repeatedly throughout the reconciliation of a resource.

### DIFF
--- a/dev/tools/controllerbuilder/template/apis/refs.go
+++ b/dev/tools/controllerbuilder/template/apis/refs.go
@@ -80,7 +80,7 @@ func (r *{{.ProtoResource}}Ref) NormalizedExternal(ctx context.Context, reader c
 	if actualExternalRef == "" {
 		return "", k8s.NewReferenceNotReadyError(u.GroupVersionKind(), key)
 	}
-	r.External = actualExternalRef
-	return r.External, nil
+
+	return actualExternalRef, nil
 }
 `
--- a/apis/bigquery/v1beta1/dataset_reference.go
+++ b/apis/bigquery/v1beta1/dataset_reference.go
@@ -81,8 +81,7 @@ func (r *BigQueryDatasetRef) NormalizedExternal(ctx context.Context, reader clie
 	if actualExternalRef == "" {
 		return "", k8s.NewReferenceNotReadyError(u.GroupVersionKind(), key)
 	}
-	r.External = actualExternalRef
-	return r.External, nil
+	return actualExternalRef, nil
 }
 
 // New builds a BigQueryDatasetRef from the Config Connector BigQueryDataset object.
--- a/apis/bigqueryanalyticshub/v1alpha1/dataexchange_reference.go
+++ b/apis/bigqueryanalyticshub/v1alpha1/dataexchange_reference.go
@@ -81,8 +81,7 @@ func (r *BigQueryAnalyticsHubDataExchangeRef) NormalizedExternal(ctx context.Con
 	if actualExternalRef == "" {
 		return "", k8s.NewReferenceNotReadyError(u.GroupVersionKind(), key)
 	}
-	r.External = actualExternalRef
-	return r.External, nil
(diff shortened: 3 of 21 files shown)


dev/tools/controllerbuilder/template/apis/refs.go
acpana, line 1: cc @yuwenma / @gemmahou 

Let me know what you think! The `apis` changes were mechanical so let's not index on them for now.

gemmahou:
>If the NormlizedExternal checks that only one of r.External or r.Name is set BUT assigns r.External at the end, future calls to this function will error out. 

Could you explain more on the error and maybe share the error message? I think it's intended to resolve and assign to r.External.

I'm working on some refactoring on the resource_references.go code. New function looks like:
```
func (r *ComputeBackendServiceRef) ResolveExternal(ctx context.Context, reader client.Reader, otherNamespace string) error {
	// Get value from spec.ComputeBackendServiceRef.external
	if r.External != "" {
		if r.Name != "" {
			return "", fmt.Errorf("cannot specify both name and external on reference")
		}
		return nil
	}

	if r.Name == "" {
		return fmt.Errorf("must specify either name or external on reference")
	}

	// Get value from the Config Connector object
	if r.Namespace == "" {
		r.Namespace = otherNamespace
	}
	key := types.NamespacedName{Name: r.Name, Namespace: r.Namespace}
	u := &unstructured.Unstructured{}
	u.SetGroupVersionKind(ComputeBackendServiceGVK)
	if err := reader.Get(ctx, key, u); err != nil {
		if apierrors.IsNotFound(err) {
			return k8s.NewReferenceNotFoundError(u.GroupVersionKind(), key)
		}
		return fmt.Errorf("reading referenced %s %s: %w", ComputeBackendServiceGVK, key, err)
	}

	// Get value from object's status.externalRef(only support for objects created by direct controller). 
	// This is the most trustworthy place.
	actualExternalRef, _, err := unstructured.NestedString(u.Object, "status", "externalRef")
	if err != nil {
		return fmt.Errorf("reading status.externalRef: %w", err)
	}
	if actualExternalRef == "" {
		return k8s.NewReferenceNotReadyError(u.GroupVersionKind(), key)
	}
	r.External = actualExternalRef
	return nil
}
```

dev/tools/controllerbuilder/template/apis/refs.go
acpana: @gemmahou let's use this comment thread. The problem persists in the new refactor too.

For instance, if a code block (or control flow) calls the function twice for a reference defined with `r.Name`:

```
ref.ResolveExternal(...) // this one assigns the External value
ref.ResolveExternal(...) // this one sees the External assigned and the Name used so errors out...
```

The error will be `cannot specify both name and external on reference`

gemmahou: Gotcha!

I'm thinking why we don't encounter this error during direct reconciliation of current resources: Since we always use a brand new object([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/04beef2d5588f628c0c8dd35967268679ab41611/pkg/controller/direct/bigqueryanalyticshub/listing_controller.go#L72)) when calling the resolver, and only resolve it once. It seems unnecessary to resolve it again unless there's a reason I'm not aware of?

ldanielmadariaga:

PR #3262, opened 2024-11-26
fix: Codify Apigee proto manual fixes

jasonvigil:
Following up on suggested change in
https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3183#pullrequestreview-2447281185

### DIFF
--- a/mockgcp/apply-proto-patches.sh
+++ b/mockgcp/apply-proto-patches.sh
@@ -67,3 +67,7 @@ cd tools/patch-proto
 #   // Optional. The configuration for Private Service Connect (PSC) for the cluster.
 #   PscConfig psc_config = 31 [(google.api.field_behavior) = OPTIONAL];
 # EOF
+
+# Fix issue in generated apigee service proto
+# Ref: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3183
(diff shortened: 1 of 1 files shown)


mockgcp/apply-proto-patches.sh
```diff
@@ -67,3 +67,7 @@ cd tools/patch-proto
 #   // Optional. The configuration for Private Service Connect (PSC) for the cluster.
 #   PscConfig psc_config = 31 [(google.api.field_behavior) = OPTIONAL];
 # EOF
+
+# Fix issue in generated apigee service proto
+# Ref: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3183
+sed -i 's/^      post: "\/v1\/{project=projects\/\*}:provisionOrganization"$/      post: "\/v1\/{name=projects\/\*}:provisionOrganization"/g' ${REPO_ROOT}/mockgcp/apis/mockgcp/cloud/apigee/v1/service.proto
```
yuwenma, line 73: This will break Mac users on adding MockGCP. 

Can we use `mockgcp/tools/patch-proto` (like extending the tool to find-and-replace)?

jasonvigil: Ok, I implemented this change in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3351. Going to close this PR in favor of that one.

PR #2657, opened 2024-09-09
Add management field for gkehubfeaturemembership

ziyue-101:
### Change description

Add a new field to support CS auto upgrade
#2654
### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #3263, opened 2024-11-27
Improve update method to avoid permanent diff

ziyue-101:
### Change description

Given that the server defaults values if certain fields are not set, the PR  Improves update method to avoid permanent diff

### DIFF
--- a/pkg/controller/direct/gkehub/featuremembership_controller.go
+++ b/pkg/controller/direct/gkehub/featuremembership_controller.go
@@ -17,7 +17,6 @@ package gkehub
 import (
 	"context"
 	"fmt"
-	"reflect"
 	"strings"
 	"time"
 
@@ -62,7 +61,7 @@ type gkeHubAdapter struct {
 	projectID    string
 	location     string
 
-	desired *featureapi.MembershipFeatureSpec
+	desired *krm.GKEHubFeatureMembership
 	actual  *featureapi.Feature
 
 	hubClient *gkeHubClient
@@ -108,16 +107,12 @@ func (m *gkeHubModel) AdapterForObject(ctx context.Context, reader client.Reader
 	if err := resolveIAMReferences(ctx, reader, obj); err != nil {
 		return nil, err
 	}
-	apiObj, err := featureMembershipSpecKRMtoMembershipFeatureSpecAPI(&obj.Spec)
-	if err != nil {
-		return nil, err
-	}
 	return &gkeHubAdapter{
 		membershipID: membership.id,
 		featureID:    feature.id,
 		projectID:    projectID,
 		location:     obj.Spec.Location,
-		desired:      apiObj,
+		desired:      obj,
 		hubClient:    hubClient,
 	}, nil
 }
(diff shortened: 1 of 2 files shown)


acpana:
/lgtm
/approve

thanks for adding these!

cc @maqiuyujoyce would you mind approving too 🙏🏼 ?

The spelled out comparison is necessary since these protos don't implement the message interfaces which would allow us to use the common tooling that yuwen wrote. Note that since there's no changes to http logs this means these don't cause any extra diffs.

Once we merge this Ziyue will look at #2657

pkg/controller/direct/gkehub/featuremembership_controller.go
```diff
@@ -249,7 +248,7 @@ func (a *gkeHubAdapter) Update(ctx context.Context, updateOp *directbase.UpdateO
 	log.V(2).Info("updating object", "u", u)
 	actual := a.actual.MembershipSpecs[a.membershipID]
 	//  There are no output fields in the api Object, so we can compare the desired and the actaul directly.
```
maqiuyujoyce, line 250: Just curious, is this comment still valid?

pkg/controller/direct/gkehub/featuremembership_controller.go
```diff
@@ -249,7 +248,7 @@ func (a *gkeHubAdapter) Update(ctx context.Context, updateOp *directbase.UpdateO
...
-	if !reflect.DeepEqual(a.desired.Configmanagement, actual.Configmanagement) || !reflect.DeepEqual(a.desired.Policycontroller, actual.Policycontroller) || !reflect.DeepEqual(a.desired.Mesh, actual.Mesh) {
```
maqiuyujoyce, line 252: What's the problem of DeepEqual here? I think we use it a lot in diffing for other resources. If there are any particular issues, we should identify them here and avoid them for future implementations.

maqiuyujoyce (commented):
/approve

I have a couple of comments and both comments can be addressed offline.

pkg/controller/direct/gkehub/featuremembership_controller.go
ziyue-101: the `desired `is converted from KRM representation. The `actual` is actually fetched from the gcp server. If there are server defaults values, there will be permanent diff in the controller.

pkg/controller/direct/gkehub/featuremembership_controller.go
ziyue-101: It still holds in some sense. but we need to rephrase it a bit, will send out a separate PR

maqiuyujoyce:
@ziyue-101 Looking at the diffing functions again, it seems that the diffing works if we support externally-managed fields (i.e. desired state is nil and actual state is non-nil). But in direct controller, I think we want the desired state to be the source of truth, so if the desired state is nil, the actual state should be validated as well: it should be nil or the default value (depending on the API behavior). @yuwenma just want to double check, is my understanding correct?

ziyue-101:
While waiting on yuwen's response, I am wondering if KCC doesn't want to support externally-managed fields, how should the controller handle server populated fields such as defaults? A way could be to record a copy of fields that have defaults, but that would make it harder to keep that copy up-to-date.

It would also be great to update https://cloud.google.com/config-connector/docs/concepts/managing-fields-externally to reflect the latest KCC status.

PR #3265, opened 2024-11-27
feat: Implement WorkstationConfig direct controller and record GCP

jasonvigil:
Also made a few fixes to generated types (based on https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3228), and updated mappings accordingly.


pkg/controller/direct/workstations/config_controller.go
```diff
@@ -0,0 +1,264 @@
...
+	resource.Name = a.id.External
+	resource.Etag = a.actual.Etag
+
+	paths, err := common.CompareProtoMessage(resource, a.actual, common.BasicDiff)
+	if err != nil {
+		return err
+	}
+	if len(paths) == 0 {
+		log.V(2).Info("no field needs update", "name", a.id.External)
+		return nil
```
yuwenma, line 187: Ah! Looks like the template is wrong. If this is acquiring a resource, the GCP resource may already in-sync with KCC  but KCC `status` is empty if not updated.

pkg/controller/direct/workstations/config_controller.go
```diff
@@ -0,0 +1,264 @@
...
+	req := &pb.DeleteWorkstationConfigRequest{Name: a.id.External}
+	op, err := a.gcpClient.DeleteWorkstationConfig(ctx, req)
+	if err != nil {
+		return false, fmt.Errorf("deleting WorkstationConfig %s: %w", a.id.External, err)
+	}
+	log.V(2).Info("successfully deleted WorkstationConfig", "name", a.id.External)
+
+	_, err = op.Wait(ctx)
+	if err != nil {
+		// todo (b/368419476): Workstation service does not provide a valid response on success.
```
yuwenma, line 257: Nice! Thank you for figuring it out.

pkg/controller/direct/workstations/config_defaults.go
```diff
@@ -0,0 +1,46 @@
...
+
+package workstations
+
+import (
+	pb "cloud.google.com/go/workstations/apiv1/workstationspb"
+	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/workstations/v1alpha1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
+)
+
+func ApplyWorkstationConfigGCPDefaults(mapCtx *direct.MapContext, in *krm.WorkstationConfigSpec, out *pb.WorkstationConfig, actual *pb.WorkstationConfig) {
```
yuwenma, line 23: Will the Workstation service changes the default value or makes a difference between the fields that are unset (what user intend) or set to a default value (what KCC does).

pkg/controller/direct/workstations/config_mappings.go
```diff
@@ -26,15 +26,32 @@ func WorkstationConfigObservedState_FromProto(mapCtx *direct.MapContext, in *pb.
...
-	// MISSING: Reconciling
+	out.UID = direct.LazyPtr(in.GetUid())
 	out.CreateTime = direct.StringTimestamp_FromProto(mapCtx, in.GetCreateTime())
 	out.UpdateTime = direct.StringTimestamp_FromProto(mapCtx, in.GetUpdateTime())
 	out.DeleteTime = direct.StringTimestamp_FromProto(mapCtx, in.GetDeleteTime())
 	out.Etag = direct.LazyPtr(in.GetEtag())
+	out.Host = WorkstationConfig_HostObservedState_FromProto(mapCtx, in.Host)
 	out.Degraded = direct.LazyPtr(in.GetDegraded())
-	// MISSING: Conditions
+	out.GCPConditions = WorkstationGCPConditions_FromProto_Alpha(mapCtx, in.Conditions)
```
yuwenma, line 36: Just out of curiosity, why is this called `_alpha`?

pkg/controller/direct/workstations/config_defaults.go
yuwenma: I'm ok with the current approach, ideally we want to separate the create and update handling because logically they are different. 
You may also want to double check the `ReplicaZones`, if it is set to empty slice (rather than unset), will the GCP service gives a default value, reject the call or unset the zones. (I recalled it uses the default value, which means we may want to deal `len(in.ReplicaZones) == 0` as well.

apis/workstations/v1alpha1/config_reference.go
```diff
@@ -148,6 +150,8 @@ func NewWorkstationConfigRef(ctx context.Context, reader client.Reader, obj *Wor
 		return nil, fmt.Errorf("cannot reset `metadata.name` or `spec.resourceID` to %s, since it has already assigned to %s",
 			resourceID, actualResourceID)
 	}
+	id.Name = resourceID
+	id.Namespace = obj.GetNamespace()
```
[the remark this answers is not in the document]
jasonvigil: They are used in the `NormalizedExternal` function. Ref: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/bc85425a3f65f771296e8406702000b2056ba758/apis/workstations/v1alpha1/config_reference.go#L49

That function is currently unused, but it will probably be used soon by `WorkstationsWorkstation`.

pkg/controller/direct/workstations/config_controller.go
jasonvigil: Will you make this fix for the template? I will update the code in this controller to set status.

pkg/controller/direct/workstations/config_mappings.go
jasonvigil: There is a difference in type between v1beta1 / v1alpha1. I will remove the `_Alpha` functions once I promote this resource (and other Workstation resources) to v1beta1.

pkg/controller/direct/workstations/config_controller.go
```diff
@@ -0,0 +1,264 @@
...
+		return nil, fmt.Errorf("error converting to %T: %w", obj, err)
+	}
+
+	id, err := krm.NewWorkstationConfigRef(ctx, reader, obj)
+	if err != nil {
+		return nil, err
+	}
+
+	// Resolve references
+	if err := ResolveWorkstationConfigRefs(ctx, reader, obj); err != nil {
```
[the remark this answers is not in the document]
jasonvigil: There is no k8s client available in `Create` or `Update`. Any thoughts / opinions on how to manage this?

pkg/controller/direct/workstations/config_defaults.go
jasonvigil: It is impossible to update the `ReplicaZones` field (it's immutable).

apis/workstations/v1alpha1/config_reference.go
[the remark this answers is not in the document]
yuwenma: This might be a better example on how the `NewWorkstationConfigRef` and `NormalizedExternal` could be used. 
[NormalizedExternal
](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/secretmanager/v1beta1/secret_reference.go)
https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/secretmanager/v1beta1/secret_identity.go#L52

In summary, they should be in different struct (and we have fixed that in the api template), and when using the `WorkstationConfig` as a reference, the `name` or `external` is assigned from the API, and we call `normalizeExternal`.

pkg/controller/direct/workstations/config_controller.go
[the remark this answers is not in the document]
yuwenma: Sure, the k8s client can be a  attribute in `WorkstationConfigAdapter` (we should do that in the template as well), and the obj is already stored in `adapter.desired`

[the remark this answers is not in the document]
jasonvigil: Ok, I moved the ref resolve function calls to `Create` and `Update`

apis/workstations/v1alpha1/config_reference.go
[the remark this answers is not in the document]
jasonvigil: Ok, updated to use the latest revision of the ref resolving template. Also, modified the KRM type definition to remove the unnecessary `location` and `projectRef` fields.

apis/workstations/v1alpha1/config_identity.go
```diff
@@ -0,0 +1,132 @@
...
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/apis/common"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// WorkstationConfigIdentity defines the resource reference to WorkstationConfig, which "External" field
+// holds the GCP identifier for the KRM object.
```
yuwenma, line 27: nit: remove "which "External" field holds the GCP identifier for the KRM object."

apis/workstations/v1alpha1/config_identity.go
```diff
@@ -0,0 +1,132 @@
...
+
+func (i *WorkstationConfigIdentity) ID() string {
+	return i.id
+}
+
+func (i *WorkstationConfigIdentity) Parent() *WorkstationConfigParent {
+	return i.parent
+}
+
+type WorkstationConfigParent struct {
```
yuwenma, line 45: Some thoughts: This can use the "WorkstationClusterIdentity", so shows the direct relationship between a child resource identity and its parent resource identity, especially if you want to only define the project/location in parent.

apis/workstations/v1alpha1/config_identity.go
```diff
@@ -0,0 +1,132 @@
...
+	// Get Parent
+	clusterRef := obj.Spec.Parent
+	if clusterRef == nil {
+		return nil, fmt.Errorf("no parent cluster")
+	}
+	clusterExternal, err := clusterRef.NormalizedExternal(ctx, reader, obj.Namespace)
+	if err != nil {
+		return nil, fmt.Errorf("cannot resolve cluster: %w", err)
+	}
+	clusterParent, cluster, err := parseWorkstationClusterExternal(clusterExternal)
```
yuwenma, line 66: Personally I like it, we might end up add back the `spec.projectRef` and `spec.location` to give user a consistent experience compared to the other resources. If that's the case, what you did here would be a valuable thing and a hidden gem for those KCC lovers to explore and discover.

yuwenma (commented):
/lgtm
/approve

Looks great! Left some thoughts, non blocker

PR #3266, opened 2024-11-27
feat: Add mockgcp for WorkstationConfig

jasonvigil:
This PR depends on https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3265, would suggest reviewing by commit.

Also, there are a couple of minor quirks with this resource:
- Seems impossible to remove annotations / labels, so adjusted the test to make the updates additive.
- allowedPorts field is not yet supported in the proto.

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/workstations/workstationconfig/workstationconfig-full/_generated_object_workstationconfig-full.golden.yaml
+++ b/pkg/test/resourcefixture/testdata/basic/workstations/workstationconfig/workstationconfig-full/_generated_object_workstationconfig-full.golden.yaml
@@ -11,6 +11,8 @@ metadata:
   namespace: ${uniqueId}
 spec:
   annotations:
+  - key: a-key1
+    value: a-value1
   - key: a-key2
     value: a-value2
   container:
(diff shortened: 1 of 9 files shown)


yuwenma (commented):
Could you make sure the PR contains the git-commits that can reflect the following info  
- HTTP Log difference between mockgcp and real-gcp. (you can run `record-gcp` and git-commit, and then run `compare-mock` and then git-commit so we can see the diff)
- mutable fields are changed. You can remove updated.yaml and run `compare-mock`, and then git-commit, then add back the update.yaml and run it again.

yuwenma:
/approve

Reviewed the git-commit for mockGCP, waiting for the diff-base PR to merge

pkg/test/resourcefixture/testdata/basic/workstations/workstationconfig/workstationconfig-full/_generated_object_workstationconfig-full.golden.yaml
```diff
@@ -11,6 +11,8 @@ metadata:
   namespace: ${uniqueId}
 spec:
   annotations:
+  - key: a-key1
+    value: a-value1
```
yuwenma, line 15: qq: is there a reason to not using the yaml dict? like 
```yaml
annotations:
    "a-key1": "a-value1"
    "a-key2": "a-value2"
```

PR #3272, opened 2024-12-02
[POC] feat: support selective core CRD installation in different release channels

jingyih:
This PR introduces support for installing distinct sets of CRDs in the GKE add-on release compared to the default and Autopilot releases. This enables us to exclude certain beta customization CRDs and any future alpha customization CRDs from the GKE add-on.

Alternatively, using the same mechanism, we could align the CRDs in the GKE add-on release with the default/autopilot releases but exclude alpha core CRDs across all release channels.


yuwenma:
I think the PR is in the right direction, especially the huge improvements on kustomize use. However it may break the current release, could you run `make upgrade-kcc-local` to see if it passes?

jingyih:
This PR is not immediately needed at the moment.

/hold

jingyih:
This is superseded by #4645.

/close

PR #3277, opened 2024-12-02
refactor: table => tableRef

acpana:
* Adds the `source` to `ToProto` mapping
* Rename `table` to `tableRef`


pkg/controller/direct/bigqueryanalyticshub/mapper.go
```diff
@@ -239,7 +239,41 @@ func BigQueryAnalyticsHubListingSpec_ToProto(mapCtx *direct.MapContext, in *krmv
...
+	}
+
+	out := &pb.Listing_BigqueryDataset{}
+	if in.Dataset != nil {
+		out.BigqueryDataset = &pb.Listing_BigQueryDatasetSource{
+			Dataset: in.Dataset.External,
+		}
+
+		out.BigqueryDataset.RestrictedExportPolicy = &pb.Listing_BigQueryDatasetSource_RestrictedExportPolicy{}
+		if in.RestrictedExportPolicy != nil {
```
jingyih, line 258: What happens if we send `BigQueryDatasetSource` with empty Dataset but the other 2 fields are configured?

pkg/controller/direct/bigqueryanalyticshub/mapper.go
```diff
@@ -239,7 +239,41 @@ func BigQueryAnalyticsHubListingSpec_ToProto(mapCtx *direct.MapContext, in *krmv
 	dtype := direct.Enum_ToProto[pb.DiscoveryType](mapCtx, in.DiscoveryType)
 	out.DiscoveryType = &dtype
 
-	// todo acpana do source
+	out.Source = Listing_BigQueryDatasetSource_ToProto(mapCtx, in.Source.BigQueryDatasetSource)
```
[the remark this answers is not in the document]
acpana: for now yes it is the only source; I can added some comments

pkg/controller/direct/bigqueryanalyticshub/mapper.go
acpana: I'd say let's let the underlying cloud api let error out rather than try to pre-optimize here

jingyih: The current code only send the other 2 fields if the field `Dataset` is configured, which is why I asked the first question. I agree we can delegate to the underlying API.

yuwenma:
/approve

Defer to @jingyih for lgtm vote. Thanks for changing the API field name, it looks much better!

jingyih (commented):
/lgtm

PR #3278, opened 2024-12-03
Add an example command to generate a controller template

maqiuyujoyce:
### Change description

Took me a while to figure out the exact values to fill in the command so I think it'll be helpful to have an example.

### Tests you have done

### DIFF
--- a/docs/develop-resources/deep-dives/4-add-controller.md
+++ b/docs/develop-resources/deep-dives/4-add-controller.md
@@ -5,6 +5,8 @@ Run the following command to generate a controller template
 ```
 cd dev/tools/controllerbuilder
 go run main.go generate-controller --service <YOUR_SERVICE> --api-version <VERSION> --kind <YOUR_RESOURCE> --proto-resource <PROTO_RESOURCE>
+# For example,
+# go run main.go generate-controller --service alloydb.v1beta --api-version alloydb.cnrm.cloud.google.com/v1beta1 --kind AlloyDBInstance --proto-resource instance
 ```
 
 Fix the generated code to make your SciFi running!


docs/develop-resources/deep-dives/4-add-controller.md
```diff
@@ -5,6 +5,8 @@ Run the following command to generate a controller template
 ```
 cd dev/tools/controllerbuilder
 go run main.go generate-controller --service <YOUR_SERVICE> --api-version <VERSION> --kind <YOUR_RESOURCE> --proto-resource <PROTO_RESOURCE>
+# For example,
+# go run main.go generate-controller --service alloydb.v1beta --api-version alloydb.cnrm.cloud.google.com/v1beta1 --kind AlloyDBInstance --proto-resource instance
```
jingyih, line 9: The flag values should align with those used in other tools (as detailed in Step 2). If there is any discrepancy, we should get it fixed.

cc @yuwenma maybe we should combine `--kind` and `--proto-resource` into one flag, similar to what we did for the type generator.

jingyih (commented):
/lgtm

docs/develop-resources/deep-dives/4-add-controller.md
yuwenma: nit: move the last line in its own bash code snippet.

yuwenma: > The flag values should align with those used in other tools (as detailed in Step 2). If there is any discrepancy, we should get it fixed.
> 
> cc @yuwenma maybe we should combine `--kind` and `--proto-resource` into one flag, similar to what we did for the type generator.

+1. I thought the flag aligned already (?)

PR #3279, opened 2024-12-03
Fix nit in error message in direct controller template

maqiuyujoyce:
### Tests you have done

### DIFF
--- a/dev/tools/controllerbuilder/template/controller/controller.go
+++ b/dev/tools/controllerbuilder/template/controller/controller.go
@@ -95,7 +95,7 @@ func (m *model{{.ProtoResource}}) client(ctx context.Context) (*gcp.Client, erro
 	}
 	gcpClient, err := gcp.NewRESTClient(ctx, opts...)
 	if err != nil {
-		return nil, fmt.Errorf("building {{.ProtoResource}} client: %w", err)
+		return nil, fmt.Errorf("error building {{.ProtoResource}} client: %w", err)
 	}
 	return gcpClient, err
 }


dev/tools/controllerbuilder/template/controller/controller.go
```diff
@@ -95,7 +95,7 @@ func (m *model{{.ProtoResource}}) client(ctx context.Context) (*gcp.Client, erro
 	}
 	gcpClient, err := gcp.NewRESTClient(ctx, opts...)
 	if err != nil {
-		return nil, fmt.Errorf("building {{.ProtoResource}} client: %w", err)
+		return nil, fmt.Errorf("error building {{.ProtoResource}} client: %w", err)
```
jingyih, line 98: This is a matter of preference, I think adding "error" is better. If we decide to update this error string, we should probably update the others as well for consistency. (e.g. line 154)
cc @yuwenma

maqiuyujoyce: Good point! I'm doing incremental updates based on my findings while cleaning up the generated controller. Agreed the error messages look inconsistent over the controller. We should finalize on one "format". I'd prefer we make the `error` explicitly in the message.

yuwenma: This has been raised up multiple times. Like Jingyi suggested, the root problem is to give a unified form to highlight the errors so that we can search more easily. If we want to change them, I'd suggest we align the errors inside controller. 
nit: "error" is less accurate than "unable" "invalid"

PR #3282, opened 2024-12-03
feat: add secretManager version alias

yuwenma:
### Change description

Fixes #

### Tests you have done


tests/e2e/export.go
```diff
@@ -103,17 +106,24 @@ func exportResource(h *create.Harness, obj *unstructured.Unstructured, expectati
...
 	outputDir := h.TempDir()
 	outputPath := filepath.Join(outputDir, "export.yaml")
 	exportParams.Output = outputPath
 	exportParams.URI = exportURI
-	if err := export.Execute(h.Ctx, &exportParams); err != nil {
-		h.Errorf("error from export.Execute: %v", err)
-		return ""
+	switch gvk.Kind {
+	case "SecretManagerSecretVersion":
+		// Skip run export.Execute because the SecretVersion servicemappings has a broken marker
```
justinsb, line 118: Can we only export if using direct?  Or just move to direct export everywhere?

justinsb: I guess in practice it's not very useful to export the SecretManagerSecretVersion, and potentially risky (exposing secret material)

pkg/controller/direct/secretmanager/secret_mapping.go
```diff
@@ -23,7 +25,12 @@ import (
 )
 
 func SecretManagerSecretStatusObservedState_FromProto(mapCtx *direct.MapContext, in *pb.Secret) *krm.SecretManagerSecretObservedState {
-	return &krm.SecretManagerSecretObservedState{}
+	if in == nil {
+		return nil
+	}
+	out := &krm.SecretManagerSecretObservedState{}
+	out.VersionAliases = MapStringInt64_ToMapStringString(mapCtx, in.VersionAliases)
```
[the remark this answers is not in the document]
yuwenma: I was thinking of the same thing. But this use here is not typical enough and IMO more like a unclear SecretManager API decision.  I'd prefer adding it later if we encounter another case.

pkg/controller/direct/secretmanager/secret_mapping.go
```diff
@@ -96,8 +103,7 @@ func SecretManagerSecretSpec_ToProto(mapCtx *direct.MapContext, in *krm.SecretMa
 	}
 	// MISSING: Etag
 	out.Rotation = Rotation_ToProto(mapCtx, in.Rotation)
-	// MISSING: VersionAliases
-	// out.VersionAliases = in.VersionAliases
+	out.VersionAliases = MapStringString_ToMapStringInt64(mapCtx, in.VersionAliases)
```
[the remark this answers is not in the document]
yuwenma: Very interesting case! I can do a follow up PR to improve this. This is less likely to happen for SecretManager because the int64 is used a version counts starting from 1.

scripts/generate-google3-docs/resource-reference/generated/resource-docs/secretmanager/secretmanagersecret.md
```diff
@@ -555,7 +555,9 @@ conditions:
 externalRef: string
 name: string
 observedGeneration: integer
-observedState: {}
+observedState:
+  versionAliases:
+    string: string
```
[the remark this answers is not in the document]
yuwenma: `status.versionAliases` is a new field I added in this PR. otherwise, it is very unclear whether the versionAliases succeeds or not. 
The `spec.versionAliases` are already there.

tests/e2e/export.go
yuwenma: Agree. And agree. I hit an error around the scenario, and explored on what happened in the code. Lots of improvements are needed for Export.

pkg/controller/direct/secretmanager/secret_controller.go
```diff
@@ -203,7 +204,14 @@ func (a *Adapter) Create(ctx context.Context, op *directbase.CreateOperation) er
 	external := a.id.String()
 	status.ExternalRef = &external
 	status.Name = created.Name
-	return op.UpdateStatus(ctx, status, nil)
+	if err = op.UpdateStatus(ctx, status, nil); err != nil {
```
justinsb, line 207:
```suggestion
	if err := op.UpdateStatus(ctx, status, nil); err != nil {
```

scripts/generate-google3-docs/resource-reference/generated/resource-docs/secretmanager/secretmanagersecret.md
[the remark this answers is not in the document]
justinsb: Yeah, I was confused by github's presentation of the diff.  map[string]string makes sense.  It would be nice if it was map[string]int64 in KRM and proto, but it may encounter some bugs with big values.

justinsb:
LGTM, will hold so you can decide when you want to merge

/approve
/lgtm
/hold

pkg/controller/direct/secretmanager/secret_controller.go
yuwenma: Let me fix that in follow up PRs. have some other comments and nits want to resolve as well.

PR #3286, opened 2024-12-03
tests: csau test

ziyue-101:
### Change description

Remove version filed if auto-upgrade is turned on.

### Tests you have done

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/gkehub/v1beta1/gkehubfeaturemembership/basicacmgkehubfeaturemembership/update.yaml
+++ b/pkg/test/resourcefixture/testdata/basic/gkehub/v1beta1/gkehubfeaturemembership/basicacmgkehubfeaturemembership/update.yaml
@@ -38,6 +38,5 @@ spec:
         syncWaitSecs: "20"
         syncRev: "HEAD"
         secretType: "none"
-    version: "1.18.1"
-    management: "MANAGEMENT_AUTOMATIC"
+    version: "1.18.2"
 
--- a/pkg/test/resourcefixture/testdata/basic/gkehub/v1beta1/gkehubfeaturemembership/basiccsaugkehubfeaturemembership/_http.log
+++ b/pkg/test/resourcefixture/testdata/basic/gkehub/v1beta1/gkehubfeaturemembership/basiccsaugkehubfeaturemembership/_http.log
@@ -1,0 +1,2201 @@
+GET https://cloudresourcemanager.googleapis.com/v1/projects/example-project-01?alt=json&prettyPrint=false
+User-Agent: google-api-go-client/0.5 Terraform/ (+https://www.terraform.io) Terraform-Plugin-SDK/2.10.1 terraform-provider-google-beta/kcc/controller-manager
+
+403 Forbidden
+Cache-Control: private
+Content-Type: application/json; charset=UTF-8
+Server: ESF
+Vary: Origin
+Vary: X-Origin
+Vary: Referer
+X-Content-Type-Options: nosniff
+X-Frame-Options: SAMEORIGIN
+X-Xss-Protection: 0
+
+{
+  "error": {
+    "code": 403,
+    "message": "The caller does not have permission",
+    "status": "PERMISSION_DENIED"
+  }
+}
+
+---
+
+POST https://cloudbilling.googleapis.com/v1/billingAccounts/${billingAccountID}:testIamPermissions?alt=json&prettyPrint=false
+Content-Type: application/json
+User-Agent: google-api-go-client/0.5 Terraform/ (+https://www.terraform.io) Terraform-Plugin-SDK/2.10.1 terraform-provider-google-beta/kcc/controller-manager
+
+{
+  "permissions": [
+    "billing.resourceAssociations.create"
+  ]
+}
+
+200 OK
+Cache-Control: private
+Content-Type: application/json; charset=UTF-8
+Server: ESF
+Vary: Origin
+Vary: X-Origin
+Vary: Referer
+X-Content-Type-Options: nosniff
+X-Frame-Options: SAMEORIGIN
+X-Xss-Protection: 0
+
+{
+  "permissions": [
+    "billing.resourceAssociations.create"
+  ]
+}
+
+---
+
+POST https://cloudresourcemanager.googleapis.com/v1/projects?alt=json&prettyPrint=false
+Content-Type: application/json
+User-Agent: google-api-go-client/0.5 Terraform/ (+https://www.terraform.io) Terraform-Plugin-SDK/2.10.1 terraform-provider-google-beta/kcc/controller-manager
+
+{
+  "labels": {
+    "cnrm-test": "true",
+    "managed-by-cnrm": "true"
+  },
+  "name": "Dependent Project",
+  "parent": {
+    "id": "${organizationID}",
+    "type": "organization"
+  },
+  "projectId": "example-project-01"
+}
+
+200 OK
+Cache-Control: private
+Content-Type: application/json; charset=UTF-8
+Server: ESF
+Vary: Origin
+Vary: X-Origin
+Vary: Referer
+X-Content-Type-Options: nosniff
+X-Frame-Options: SAMEORIGIN
+X-Xss-Protection: 0
+
+{
+  "name": "operations/${operationID}"
+}
+
+---
+
+GET https://cloudresourcemanager.googleapis.com/v1/operations/${operationID}?alt=json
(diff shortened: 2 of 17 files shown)


pkg/test/resourcefixture/testdata/basic/gkehub/v1beta1/gkehubfeaturemembership/basicacmgkehubfeaturemembership/update.yaml
```diff
@@ -38,6 +38,5 @@ spec:
         syncWaitSecs: "20"
         syncRev: "HEAD"
         secretType: "none"
-    version: "1.18.1"
     management: "MANAGEMENT_AUTOMATIC"
```
acpana, line 42: I was chatting with @maqiuyujoyce offline and we were thinking it would be best if we kept this test how it was before -- without adding the `management` field. Then, we could have a separate test called "csmanagement-base` and `csmanagement-full` that showcases the mutual exclusivity of the `management` field and the `version`. Could you please make those changes 🙏🏼 ?

- remove `management` field from these two tests
- regen logs
- add a new test suite that this one without the `version` field but with the `management` field set.

acpana:
/lgtm
/approve

thanks for splitting the tests!

cc @maqiuyujoyce

pkg/test/resourcefixture/testdata/basic/gkehub/v1beta1/gkehubfeaturemembership/fullcsaugkehubfeaturemembership/create.yaml
```diff
@@ -0,0 +1,42 @@
...
+        gcpServiceAccountRef:
+          name: sa-${uniqueId} 
+        httpsProxy: "https://github.com"
+        syncRepo: "https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit"
+        syncBranch: "master"
+        policyDir: "config-connector"
+        syncWaitSecs: "20"
+        syncRev: "HEAD"
+        secretType: "none"
+    version: "1.18.1"
```
maqiuyujoyce, line 41: Could you make the test case be updating from one management mode (e.g. automatic) to another management mode (e.g. manual)? We want to test the update of this field as well.

pkg/test/resourcefixture/testdata/basic/gkehub/v1beta1/gkehubfeaturemembership/basiccsaugkehubfeaturemembership/_http.log
```diff
@@ -0,0 +1,2206 @@
...
+            "secretType": "none",
+            "syncBranch": "master",
+            "syncRepo": "https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit",
+            "syncRev": "HEAD",
+            "syncWaitSecs": "20"
+          },
+          "sourceFormat": "hierarchy"
+        },
+        "management": "MANAGEMENT_AUTOMATIC",
+        "version": "1.18.1"
```
maqiuyujoyce, line 1578: If I understand it correctly, "management" and "version" are conflicting fields. Could you help me understand why do they both exist in the PATCH request?

maqiuyujoyce:
I suggest we don't update from using `version` to using `management` (yet) due to this [test gap](https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/3344#issuecomment-2533330955).

yuwenma:
For the `management` and `version`, suggest making sure the direct controller can do the following?
1. If using CSAU, the GCP service returned version is recorded in `status.observedState.version` (or `status.version` if it already exists)
2. If using CSAU, and passing in the version in the HTTP request, does GCP service give errors? Or no-op? or take into account the version as well? 
 - if it is error. the current KCC API is good. users know how to fix their config if given both management AUTOMATIC and version.
 - If it is a no-op, the direct controller should add additional logic that `if   management= AUTOMATIC, unset version` before sending the HTTP request.
 - If GCP service take into account the version, KCC can treat it as a no-op.

maqiuyujoyce:
> If using CSAU, the GCP service returned version is recorded in status.observedState.version (or status.version if it already exists)

Good point.

> If using CSAU, and passing in the version in the HTTP request, does GCP service give errors? Or no-op? or take into account the version as well?

Note that KCC pass in the version in the HTTP request because version is also specified in the KRM object. Check https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/3344#issuecomment-2533330955 for more details. So basically this means user specified both `management` and `version`.

> If it is a no-op, the direct controller should add additional logic that if   management= AUTOMATIC, unset version before sending the HTTP request.

If no-op, then it means the live state is considered SoT and the desired intention is ignored. We should find the diff in between the desired state and live state and return an error accordingly if there is a diff.

yuwenma:
> Note that KCC pass in the version in the HTTP request because version is also specified in the KRM object. Check https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/3344#issuecomment-2533330955 for more details. So basically this means user specified both management and version.

Right. Passing in both version and management could be okay, it defers to how the GKEHub service responses. For the 3 possible results I gave, KCC only need additional logic if GKEHub service allows both values and ignore the version.

ziyue-101:
> I suggest we don't update from using `version` to using `management` (yet) due to this [test gap](https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/3344#issuecomment-2533330955).

Thanks for the digging. I don't think this PR should be blocked on the test gap. I can remove the version field in the created.yaml

ziyue-101:
@maqiuyujoyce the gap is in test, I wonder what's the behavior for base controller when doing update, does it `patch` or `update`?

ziyue-101:
Answer
> If using CSAU, the GCP service returned version is recorded in status.observedState.version (or status.version if it already exists)

If this is desired, this means the status.version field should be optional because it should only be populated when csau is enabled. This PR only adds test cases, I think we can address the status.version in followups.

ziyue-101:
removed the version field in the create.yaml due to the [test gap](https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/3344#issuecomment-2533330955)

maqiuyujoyce (commented):
/lgtm
/approve

/hold
Do you have any other concerns, @yuwenma ?

yuwenma:
> Answer
> 
> > If using CSAU, the GCP service returned version is recorded in status.observedState.version (or status.version if it already exists)
> 
> If this is desired, this means the status.version field should be optional because it should only be populated when csau is enabled. This PR only adds test cases, I think we can address the status.version in followups.

Do you infer the version is required now and not in `status`? If so, yes it should be changed from required to optional (this is common when the GCP service adds automatic support for some previously configurable only fields).

yuwenma:
/hold cancel 

Looks good on my side

ziyue-101:
> > Answer
> > > If using CSAU, the GCP service returned version is recorded in status.observedState.version (or status.version if it already exists)
> > 
> > 
> > If this is desired, this means the status.version field should be optional because it should only be populated when csau is enabled. This PR only adds test cases, I think we can address the status.version in followups.
> 
> Do you infer the version is required now and not in `status`? If so, yes it should be changed from required to optional (this is common when the GCP service adds automatic support for some previously configurable only fields).

My bad, I corrected my previous statement. The `version` field is prohibited when `management` field is set as automatic, it is optional If `management` field is set with other values. The GCP sever has validation logic to deny any request with version field and csau field set together. I suggest we delegate the validation to the GCP server and don't emulate GCP server behavior in the direct controller because that could change over time.

yuwenma:
> > > Answer
> > > > If using CSAU, the GCP service returned version is recorded in status.observedState.version (or status.version if it already exists)
> > > 
> > > 
> > > If this is desired, this means the status.version field should be optional because it should only be populated when csau is enabled. This PR only adds test cases, I think we can address the status.version in followups.
> > 
> > 
> > Do you infer the version is required now and not in `status`? If so, yes it should be changed from required to optional (this is common when the GCP service adds automatic support for some previously configurable only fields).
> 
> My bad, I corrected my previous statement. The `version` field is prohibited when `management` field is set as automatic, it is optional If `management` field is set with other values. The GCP sever has validation logic to deny any request with version field and csau field set together. I think we should delegate the validation to the GCP server and don't emulate GCP server behavior in the direct controller because that could change over time.

Right, that's option 3 I mentioned. If If GCP service take into account the version, KCC can treat it as a no-op, and let users take actions according to the GCP error  
Can you add a scenario test to record the real GCP behavior? Step 1. version configured Step 2 version and managed (auto) configured 3. version unset and only managed(auto) configured.

PR #3299, opened 2024-12-03
fix: Update docs for ApigeeEnvironment conditional IAM support

jasonvigil:
ApigeeEnvironment does not support conditional IAM.

### DIFF
--- a/scripts/generate-google3-docs/resource-reference/main.go
+++ b/scripts/generate-google3-docs/resource-reference/main.go
@@ -372,6 +372,11 @@ func handleAnnotationsAndIAMSettingsForDCLBasedResource(r *resource, gvk schema.
 		SupportsAuditConfigs:     false, // No DCL-based resources support AuditConfigs.
 		ExternalReferenceFormats: []string{externalReferenceFormat},
 	}
+	// Apigee Environment does not support conditional IAM permissions
+	// Ref: https://b.corp.google.com/issues/378594862#comment6
+	if gvk.Group == "apigee.cnrm.cloud.google.com" && gvk.Kind == "ApigeeEnvironment" {
+		r.IAM.SupportsConditions = false
+	}
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


scripts/generate-google3-docs/resource-reference/main.go
```diff
@@ -372,6 +372,11 @@ func handleAnnotationsAndIAMSettingsForDCLBasedResource(r *resource, gvk schema.
 		SupportsAuditConfigs:     false, // No DCL-based resources support AuditConfigs.
 		ExternalReferenceFormats: []string{externalReferenceFormat},
 	}
+	// Apigee Environment does not support conditional IAM permissions
+	// Ref: https://b.corp.google.com/issues/378594862#comment6
+	if gvk.Group == "apigee.cnrm.cloud.google.com" && gvk.Kind == "ApigeeEnvironment" {
+		r.IAM.SupportsConditions = false
+	}
```
yuwenma, line 379: IIUC, the godoc loads the `supportsConditions`  from the TF service mapping. 
And `ApigeeEnvironment` already has the `SupportsConditions` sets to false https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/scripts/resource-autogen/generated/servicemappings/apigee.yaml#L93
Do you know why it shows as true in the reference doc? cc @maqiuyujoyce in case you have more context.

maqiuyujoyce: ApigeeEnvironment is a DCL-based resource. I think the service mapping won't take effect for this resource in reference doc generation.

maqiuyujoyce: That's part of the design because a lot of the configurations that were defined in service mapping were moved to DCL. So service mapping is specifically used for TF-based resources, and service metadata (and configuration of DCL types) is specifically used for DCL-based resources.

yuwenma: Non blocker of this PR. I think we can further discuss on how  to re-shape the service mapping and service metadata in the scope to better serve our current resource growing needs.

PR #3300, opened 2024-12-03
chore: do not import direct package in apis/

jingyih:


apis/bigqueryconnection/v1alpha1/connection_reference.go
```diff
@@ -20,7 +20,7 @@ import (
 	"strings"
 
 	refsv1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
-	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/apis/common"
```
acpana, line 23: we can write a custom linter to prevent these imports 😛

jingyih: Removing existing patterns in the codebase might be sufficient for now. A linter would be great, but it may not be a high priority.

yuwenma (commented):
/lgtm
/approve

apis/bigqueryconnection/v1alpha1/connection_reference.go
yuwenma: Yeah, I think this common is imported in MockGCP unexpectedly as well (several times)

PR #3233, opened 2024-11-21
chore: pass in just enough args to ResolveProject

yuwenma:
### Change description

Fixes #

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #3301, opened 2024-12-04
fix: update to use new ResolveProject func

acpana:
I think https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3233 wasn't synced to the most recent changes and we forgot one old reference to the an old style `ResolveProject`


jingyih (approved):
/lgtm

PR #3164, opened 2024-11-13
Update operator to 1.125.0

yuwenma:

PR #3304, opened 2024-12-04
fix: duplicate cnrm in image path

xiaoweim:
Fix duplicate cnrm/ in image path, reference: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3164/commits/295914cae9d5f087ec130df19ed4bdfc5b648e7b

More investigations and details are in the Issue link: https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/3336.

### DIFF
--- a/operator/scripts/update-kcc-manifest/main.go
+++ b/operator/scripts/update-kcc-manifest/main.go
@@ -459,7 +459,7 @@ func swapContainerRegistry(manifestPath string) error {
 		return fmt.Errorf("error reading manifestPath: %w", err)
 	}
 	manifest := string(content)
-	updatedManifest := strings.ReplaceAll(manifest, "gcr.io/cnrm-eap/", "gcr.io/gke-release/cnrm/")
+	updatedManifest := strings.ReplaceAll(manifest, "gcr.io/cnrm-eap/", "gcr.io/gke-release/")
 	fileMode := os.FileMode(0644) // -rw-r--r--
 	return ioutil.WriteFile(manifestPath, []byte(updatedManifest), fileMode)
 }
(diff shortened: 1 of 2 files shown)


dev/tasks/build-release-bundle
```diff
@@ -51,7 +51,7 @@ echo ${VERSION} > ${BUNDLE_DIR}/version
 CRDS_FILE=$(mktemp -t crds.XXXXXXXX.yaml)
 
 # Update container registry for kustomize
-IMAGE_PREFIX=${IMAGE_PREFIX:-"gcr.io/gke-release/cnrm/"}
+IMAGE_PREFIX=${IMAGE_PREFIX:-"gcr.io/gke-release/"}
```
yuwenma, line 54: Could you add a comment to tell how this is used and works in the release pipeline? We want to figure out where the problem comes from so we can determine how to avoid it for the future. In other words, we may need some additional checker and validator in the release pipeline to block this change. Before that, we want to know how it is called.

yuwenma: Please do not resolve this comment if we haven't had a clear answer yet. But if you do, feel free to add the investigation result  here.

operator/scripts/update-kcc-manifest/main.go
```diff
@@ -459,7 +459,7 @@ func swapContainerRegistry(manifestPath string) error {
 		return fmt.Errorf("error reading manifestPath: %w", err)
 	}
 	manifest := string(content)
-	updatedManifest := strings.ReplaceAll(manifest, "gcr.io/cnrm-eap/", "gcr.io/gke-release/cnrm/")
+	updatedManifest := strings.ReplaceAll(manifest, "gcr.io/cnrm-eap/", "gcr.io/gke-release/")
```
yuwenma, line 462: Could you elaborate on the triaging process? 

This line of code is not changed since it is open sourced 2 years ago. https://github.com/GoogleCloudPlatform/k8s-config-connector/blame/master/operator/scripts/update-kcc-manifest/main.go#L462 

The duplicate `cnrm` is recently failing. So I'm not sure if this is the root cause. 
Also, the propose-tag is the new release and approved working in 1.126. I suggest we do not change its release path.

yuwenma (commented):
/lgtm
/approve

PR #3307, opened 2024-12-04
Default requireApproverJustification to false before diffing

maqiuyujoyce:
### Change description

~Fixes the issue that there is always a diff when `spec.ApprovalWorkflow.ManualApprovals.RequireApproverJustification` is set to `false` in the KCC manifest. This issue happens because `false` is the default value for this field, and the default value is not returned by the PAM API.~ (The reason why I initially thought the default value was not returned by the PAM API is the default bool literal is not printed out when using fmt.Print() to print the protobuf.)

Correction:
There is always a diff when `spec.ApprovalWorkflow.ManualApprovals.RequireApproverJustification` is set to `false` in the KCC manifest.
A partial cause of the issue is `false` is the default value for this field in the protobuf, and the value is lost when parsing the bool literal to a pointer using `direct.LazyPtr()` in the proto to krm mapper. Fixed by updating it to `direct.PtrTo()`.

After the behavior above is fixed, the new issue became that when `spec.ApprovalWorkflow.ManualApprovals.RequireApproverJustification` is unset in the KCC manifest, but the other field under `spec.ApprovalWorkflow.ManualApprovals` is set, then there is always a diff because the actual state always has `require_approver_justification` set to `false` as the default.
This is fixed by skipping diff comparison when `spec.ApprovalWorkflow.ManualApprovals.RequireApproverJustification` is unset in the desired state and `spec.ApprovalWorkflow.ManualApprovals.RequireApproverJustification` is set to `false` in the converted actual state.

Supported no change scenario tests for PrivilegedAccessManagerEntitlement to verify these changes in mock test.

Also updated the testdata version for PAM from v1alpha1 to v1beta1.

### Tests you have done


yuwenma (commented):
Could you add a test to reflect this change?

maqiuyujoyce:
> Could you add a test to reflect this change?

Done!

pkg/controller/direct/privilegedaccessmanager/entitlement_controller.go
```diff
@@ -320,6 +320,8 @@ func (a *Adapter) Update(ctx context.Context, updateOp *directbase.UpdateOperati
 		log.V(2).Info("'spec.additionalNotificationTargets' field is updated (-old +new)", cmp.Diff(parsedActual.AdditionalNotificationTargets, parsedDesired.Spec.AdditionalNotificationTargets))
 		updateMask.Paths = append(updateMask.Paths, "additional_notification_targets")
 	}
+	defaultApprovalWorkflowManualApprovalsRequireApproverJustification(parsedActual)
```
yuwenma, line 323: We should not change the actual value (then it is no longer "actual").   What happens if we remove ln323, will the test fail? In other words, does the GCP service give a different default in the actual value?

maqiuyujoyce: > What happens if we remove ln323, will the test fail?

Yes, the test will fail if `requireApproverJustification` is explicitly set to `false` in the KRM object.

> In other words, does the GCP service give a different default in the actual value?

Hmmm, what do you mean by a "different" default value? The default value for `requireApproverJustification` is `false`. Even if `requireApproverJustification` is set to `false` explicitly in the POST/PATCH request, this field is not returned in the GET response. Basically, I think this means "the actual state returned by the GCP GET is not the full state of the underlying GCP resource". So I don't think we are changing the actual value here, we just make the actual state complete.

yuwenma: Ok, I think the confusion is mostly about the code: line 325 compares the `actual` and `desired` on the `ApprovalWorkflow` field, which are "hard-coded" to false forehead in line 323-324. These codes seem do nothing.

Looks like the original problem is that a field is required in POST/PATCH but not returned in GET. For that case, you can only set the default for the "`desired` object but not actual.

PR #3309, opened 2024-12-04
chore: add public sample for BigQueryConnection referenced by IAMPolicyMember

jingyih:
TESTED
```
$ go test -v -tags=integration ./config/tests/samples/create/ -run-tests policy-member-with-bigqueryconnection-reference
```


yuwenma (commented):
/lgtm
/approve

Thank you! looks great!

PR #3312, opened 2024-12-04
auto generated mapper

himanikh:
### Change description

Fixes #

### Tests you have done


pkg/controller/direct/memorystore/instance_controller.go
```diff
@@ -0,0 +1,282 @@
...
+	desired := a.desired.DeepCopy()
+	resource := MemorystoreInstanceSpec_ToProto(mapCtx, &desired.Spec)
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+
+	// TODO(contributor): Complete the gcp "CREATE" or "INSERT" request.
+	req := &memorystorepb.CreateInstanceRequest{
+		Instance:   resource,
+		InstanceId: a.id.ID(),
```
anhdle-sso, line 141: I believe you don't need to provide id during creat

pkg/controller/direct/memorystore/instance_controller.go
```diff
@@ -0,0 +1,282 @@
...
+	mapCtx := &direct.MapContext{}
+
+	desiredPb := MemorystoreInstanceSpec_ToProto(mapCtx, &a.desired.DeepCopy().Spec)
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+
+	paths := []string{}
+	{
+		if !reflect.DeepEqual(a.desired.Spec.ReplicaCount, a.actual.ReplicaCount) {
```
anhdle-sso, line 177: You don't need to wrap this chunk in {}

himanikh: done

pkg/controller/direct/memorystore/instance_controller.go
himanikh: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/redis/cluster/rediscluster_controller.go#L219 

I followed the pattern here

yuwenma: @himanikh How the Memorystore service handle the request if InstanceID is given or not given? 

If the given instanceID points to an already exist instance object, will the service returns error (409 or 400) or ignore? 
If the Id is not given, will it get the ID from the other `instance` field?

cypres:
Currently when I deploy a memorystore.cnrm.cloud.google.com/v1alpha1/MemorystoreInstance - nothing is provisioned.

If we try to modify/patch the resource, it fails with this message
> admission webhook "deny-immutable-field-updates.cnrm.cloud.google.com" denied the request: couldn't get ResourceConfig for kind MemorystoreInstance: error getting service mapping: unable to get service mapping: no mapping with name 'memorystore.cnrm.cloud.google.com' found (retried 5 times).

Will this PR fix these issues? Is it possible to provision a MemorystoreInstance atm?

Running 1.129.2

PR #3321, opened 2024-12-05
Generate proto file if it does not exist before

anhdle-sso:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/dev/tools/controllerbuilder/pkg/options/generateoptions.go
+++ b/dev/tools/controllerbuilder/pkg/options/generateoptions.go
@@ -15,6 +15,8 @@
 package options
 
 import (
+	"errors"
+	"os"
 	"os/exec"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


jingyih (commented):
https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/57c400166493f3736bedc5320fc5c370c53eaea8/dev/tools/controllerbuilder/generate.sh#L24

This script is added to generate the pb file in the new location. After adding this script, we updated the default path to point to the pb file's new location.

dev/tools/controllerbuilder/pkg/options/generateoptions.go
```diff
@@ -32,7 +34,10 @@ func (o *GenerateOptions) InitDefaults() error {
 	if err != nil {
 		return nil
 	}
-	o.ProtoSourcePath = root + "/.build/googleapis.pb"
+	o.ProtoSourcePath, err = ProtoSourcePath(root)
```
jingyih, line 37: nit: rename to `defaultProtoSourcePath`. We might also want to move this to a later stage where the user input is bound to the flag. Perhaps we could generate it when we attempt to read it and find that it doesn't exist.

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/7bb291b13175d8614a98d6acc539474fb332344f/dev/tools/controllerbuilder/pkg/protoapi/loader.go#L33

PR #3323, opened 2024-12-05
feat: Enable direct controller for regional tcp proxy

gemmahou:
### Change description

Enable direct controller when location is set for regional target tcp proxy

### Tests you have done

### DIFF
--- a/pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
+++ b/pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
@@ -31,6 +31,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
+	kccpredicate "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/predicate"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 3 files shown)


pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
```diff
@@ -38,13 +39,32 @@ import (
...
+func (r *TargetTCPProxyReconcileGate) ShouldReconcile(o *unstructured.Unstructured) bool {
+	if r.optIn.ShouldReconcile(o) {
+		return true
+	}
+	obj := &krm.ComputeTargetTCPProxy{}
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(o.Object, &obj); err != nil {
+		return false
+	}
+	// Run the direct reconciler only when spec.location is specified and not global
+	return obj.Spec.Location != nil && obj.Spec.Location != direct.PtrTo("global")
```
yuwenma, line 65: Does `obj.spec.location` optional and a empty field means it is global? if so, I see a potential user problem. 😢

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computetargettcpproxy/regionalcomputetargettcpproxy-direct/_generated_object_regionalcomputetargettcpproxy-direct.golden.yaml
yuwenma, line 1: can we add a test suite that have `alpha.cnrm.cloud.google.com/reconciler: direct` and no `spec.location` (defaulting to global)?

gemmahou: Yeah we have this covered: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computetargettcpproxy/globalcomputetargettcpproxy-basic-direct/create.yaml

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
gemmahou: Existing customers(global only) do not need to specify this field, so we have to make `obj.spec.location` optional, to avoid breaking them. Do you have an example for the user problem? 

empty field do you mean `obj.spec.location: ""`? then it will be handled by regional API and the error message looks like: googleapi: Error 400: Invalid value for field 'region': ''. Must be a match of regex '[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?'"

yuwenma (commented):
/approve

Thanks! I'd like to have @jasonvigil to take another pass on the reconcile gate code and grant LGTM

jasonvigil:
/lgtm

Though, do we have plans to switch over to direct controller to handle global tcp proxy also? Ideally we will completely migrate to direct, right?

gemmahou:
> /lgtm
> 
> Though, do we have plans to switch over to direct controller to handle global tcp proxy also? Ideally we will completely migrate to direct, right?

Right, I will enable it in #3227.

jasonvigil:
> > /lgtm
> > Though, do we have plans to switch over to direct controller to handle global tcp proxy also? Ideally we will completely migrate to direct, right?
> 
> Right, I will enable it in #3227.

Ok, we should be able to remove the reconcile gate in #3227 then, right?

PR #3324, opened 2024-12-05
fix: Specify that MockGCP will not pass due to API changes

jasonvigil:


docs/develop-resources/deep-dives/2-define-apis.md
```diff
@@ -101,5 +101,5 @@ According to the above principles, you can decide how to process the following.
 
 ### Exit Criteria
 
-* The API PRs shall pass the MockGCP tests. This requires a dirct controller or a Terraform/DCL based controller.
+* The API PRs shall pass the MockGCP tests. This requires a dirct controller or a Terraform/DCL based controller. If the API was modified (during migration of a Terraform/DCL-based Alpha to direct Alpha), MockGCP will not pass until the direct controller is implemented.
```
yuwenma, line 104: I'm not fully understand this. Do you mean "MockGCP *should* not pass"? This is exit criteria so it only talks about the requirements. If this is best practice, maybe move it to the content above.

jasonvigil: This exit criteria does not apply for migration of a Terraform/DCL-based Alpha to direct Alpha, if the API is modified. For example, in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3332, the fixtures are not passing. This is because the API was modified to use resource references, and the terraform controller does not support the new API. And, there is no direct controller implemented (yet).

yuwenma: ah, I see. Thanks!

yuwenma (commented):
/lgtm
/approve

PR #3325, opened 2024-12-05
Release 1.126.0

yuwenma:


operator/channels/packages/configconnector/1.126.0/namespaced/per-namespace-components.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   annotations:
-    cnrm.cloud.google.com/version: 1.123.1
```
cheftako, line 5: @justinsb why is this being copied from [1.123.1](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/operator/channels/packages/configconnector/1.123.1/namespaced/per-namespace-components.yaml) and not [1.125.0](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/operator/channels/packages/configconnector/1.125.0/namespaced/per-namespace-components.yaml)? That seems wrong.

cheftako: It seems like we should always be comparing to the greatest semver version of the file to ensure we are backward compatible with that.

justinsb: It's because renames are just a societal construct, man

PR #3328, opened 2024-12-06
tests: remove LRO false fields

justinsb:
For reasons unclear, operations emit `done: false`
and `metadata.cancelRequested: false`.

This seems to violate the normal behaviour of proto bool fields
with implicit presence.

Easiest just to normalize away the GCP responses that are very hard
to mock; these values should be semantically equal anyway.

### DIFF
--- a/tests/e2e/normalize.go
+++ b/tests/e2e/normalize.go
@@ -734,6 +734,26 @@ func normalizeHTTPResponses(t *testing.T, events test.LogEntries) {
 		return s
 	})
 
+	// Specific to LROs
+	{
+		// For reasons unclear, operations emit done: false and cancelRequested: false.
+		// This seems to violate the normal behaviour of proto bool fields with implicit presence.
+		// Easiest just to normalize away the GCP responses that are hard to produce!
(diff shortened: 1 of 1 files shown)


tests/e2e/normalize.go
```diff
@@ -734,6 +734,26 @@ func normalizeHTTPResponses(t *testing.T, events test.LogEntries) {
 		return s
 	})
 
+	// Specific to LROs
+	{
+		// For reasons unclear, operations emit done: false and cancelRequested: false.
+		// This seems to violate the normal behaviour of proto bool fields with implicit presence.
+		// Easiest just to normalize away the GCP responses that are hard to produce!
+		visitor.objectTransforms = append(visitor.objectTransforms, func(path string, m map[string]any) {
```
yuwenma, line 742: Does this happen for a specific service or a general behavior? IIRC, each GCP service implements their own  LRO.

justinsb: It seems to happen for a lot of services.  I'm not entirely sure why, I think it is an oddity of the C anypb implementation when converted to json.

PR #3330, opened 2024-12-06
Enable klog in dynamic test

maqiuyujoyce:
### Change description

Followed this guide to enable klog in dynamic test: https://github.com/kubernetes/klog?tab=readme-ov-file#coexisting-with-glog

Now we can see the logs from direct controllers with `-v=2`.

### Tests you have done


maqiuyujoyce:
Though still need to figure out how to enable the logs in the unified test.

jingyih:
Thank you for fixing the klog setup in this test!

go.mod
```diff
@@ -79,6 +79,7 @@ require (
 	k8s.io/apiextensions-apiserver v0.27.9
 	k8s.io/apimachinery v0.27.11
 	k8s.io/client-go v0.27.11
+	k8s.io/klog/examples v0.0.0-20241125110655-fa118f5f4770
```
yuwenma, line 82: Interesting. Why do we need to directly import this lib?

pkg/test/main/testmain.go
```diff
@@ -27,6 +28,8 @@ import (
 	testenvironment "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test/environment"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test/webhook"
 	cnrmwebhook "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/webhook"
+	"k8s.io/klog/examples/util/require"
```
yuwenma, line 31: I suggest we do not import the examples lib. It is used for sample purposes but not used as a library. If it changes, Config Connector will break.

go.mod
maqiuyujoyce: Removed.

pkg/test/main/testmain.go
maqiuyujoyce: Removed.

yuwenma (commented):
/lgtm
/approve

PR #3333, opened 2024-12-06
fix: improve listing

acpana:
A few improvements for the listing resource
* fix on update controller
* improve mock for existing update fields
* improve basic test suite w update.yaml
* fix reference resolution for old style refs


pkg/controller/direct/bigqueryanalyticshub/listing_controller.go
```diff
@@ -95,24 +95,21 @@ func (m *modelListing) AdapterForObject(ctx context.Context, reader client.Reade
...
 func resolveOptionalReferences(ctx context.Context, reader client.Reader, obj *krm.BigQueryAnalyticsHubListing) error {
-	if ref := obj.Spec.DataExchangeRef; ref != nil {
-		_, err := ref.NormalizedExternal(ctx, reader, obj.GetNamespace())
-		if err != nil {
-			return fmt.Errorf("failed to resolve optional DataExchangeRef: %w", err)
-		}
-	}
-
 	if obj.Spec.Source != nil && obj.Spec.Source.BigQueryDatasetSource != nil {
 		if ref := obj.Spec.Source.BigQueryDatasetSource.Dataset; ref != nil {
```
jingyih, line 99: nit: rename the internal go type to DatasetRef?

jingyih (approved):
/lgtm

jingyih (commented):
/lgtm

acpana:
cc @yuwenma these small changes make the alpha controller usable for this resource

pkg/controller/direct/bigqueryanalyticshub/listing_controller.go
```diff
@@ -95,24 +95,21 @@ func (m *modelListing) AdapterForObject(ctx context.Context, reader client.Reade
...
-
 	if obj.Spec.Source != nil && obj.Spec.Source.BigQueryDatasetSource != nil {
-		if ref := obj.Spec.Source.BigQueryDatasetSource.Dataset; ref != nil {
-			if _, err := refs.ResolveBigQueryDataset(ctx, reader, obj, ref); err != nil {
+		if ref := obj.Spec.Source.BigQueryDatasetSource.DatasetRef; ref != nil {
+			id, err := refs.ResolveBigQueryDataset(ctx, reader, obj, ref)
+			if err != nil {
 				return err
 			}
+			obj.Spec.Source.BigQueryDatasetSource.DatasetRef.External = id.String()
```
yuwenma, line 104: I don't quite follow here. Could you elaborate a little bit more on why using the listing resource ID as its reference resource' external? 

We would want to use the `datasetRef.NormalizeExternal`. I know the dataset is not fully migrated to direct yet, but the  `NormalizeExternal` does not depend on which controller the resource uses.

yuwenma: I can see the current dataset reference is blocked by the `parent`  https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/bigquery/v1beta1/dataset_reference.go 

We can use the new struct and it should be fixed. cc @xiaoweim 
- ref https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/bigquery/v1beta1/dataset_reference.go
- identity https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/dev/tools/controllerbuilder/template/apis/identity.go

acpana: the `id` on line 100 is the new "old" resolved dataset reference

yuwenma: ah, ok. Suggest renaming to `datasetID` in ln100 and `tableID` in line 108.

yuwenma (commented):
/approve
/lgtm

PR #3334, opened 2024-12-06
feat: Add direct controller support for GRPC logging

jasonvigil:


pkg/config/controllerconfig.go
```diff
@@ -85,23 +89,15 @@ func (c *ControllerConfig) GRPCClientOptions() ([]option.ClientOption, error) {
 	if c.UserAgent != "" {
 		opts = append(opts, option.WithUserAgent(c.UserAgent))
 	}
-	if c.HTTPClient != nil {
```
acpana, line 88: Just adding my own PoC experience here that we cannot configure http client options when adding GRPCDialOptions. So this looks good 👍🏼 

@jasonvigil it would be cool if you also added a (grpc) log capture with this PR to showcase these changes but I'll leave that up to you/ the approves.

pkg/config/controllerconfig.go
```diff
@@ -85,23 +89,15 @@ func (c *ControllerConfig) GRPCClientOptions() ([]option.ClientOption, error) {
...
-		opts = append(opts, option.WithHTTPClient(httpClient))
-	}
 	if c.UserProjectOverride && c.BillingProject != "" {
 		opts = append(opts, option.WithQuotaProject(c.BillingProject))
 	}
 	if c.GCPTokenSource != nil {
 		opts = append(opts, option.WithTokenSource(c.GCPTokenSource))
 	}
+	if c.GRPCUnaryClientInterceptor != nil {
+		opts = append(opts, option.WithGRPCDialOption(grpc.WithUnaryInterceptor(c.GRPCUnaryClientInterceptor)))
```
yuwenma, line 99: can we provide a default UnaryClientInterceptor?

yuwenma:
/lgtm
/approve
/hold

Thanks!

pkg/config/controllerconfig.go
jasonvigil: Do we need one? I thought it's fine to use an empty interceptor (and not intercept anything) for production workflow (not in tests).

yuwenma: Okay, as long as it proves working in k8s cluster against real GCP. IIRC, the MockGCP overrides this field for gRPC so the mock result may not be good to test this.

jasonvigil: Hmmm, I tested this with `hack/record-gcp`. Not sure actually if it will work in real cluster, without the interceptor.

jasonvigil: I think, based on https://github.com/grpc/grpc-go/blob/3f95b38ded016ebf32507fc7cb6baeb2f15aef59/interceptor.go#L28-L43, it should be optional.

pkg/config/controllerconfig.go
jasonvigil: @acpana as far as I know, there are no controllers that actually use this (yet). To test this, I used https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3251, with a bunch of hacks to get the controller running in my local test environment. Also, I also modified the containerattached client function as follows:
```
func (m *modelContainerAttachedCluster) client(ctx context.Context, endpoint string) (*gcp.AttachedClustersClient, error) {
    opts, err := m.config.GRPCClientOptions()
    if err != nil {
        return nil, err
    }
    opts = append(opts, option.WithEndpoint(endpoint))

    gcpClient, err := gcp.NewAttachedClustersClient(ctx, opts...)
    if err != nil {
        return nil, fmt.Errorf("building AttachedCluster client: %w", err)
    }
    return gcpClient, err
}
```

PR #3337, opened 2024-12-09
tool: correctly insert field into Spec or ObservedState.

jingyih:
- commit 1: Refactor the typeupdater tool and fix some minor bugs.
- commit 2: Ensure that when the parent proto message of the field to be inserted matches 2 Go structs (`Spec` and `ObservedState`), the field is inserted into the appropriate struct.

### DIFF
--- a/dev/tools/controllerbuilder/pkg/commands/updatetypes/updatetypescommand.go
+++ b/dev/tools/controllerbuilder/pkg/commands/updatetypes/updatetypescommand.go
@@ -15,34 +15,24 @@
 package updatetypes
 
 import (
-	"bytes"
+	"context"
 	"fmt"
 	"os"
-	"strings"
 
-	"github.com/GoogleCloudPlatform/k8s-config-connector/dev/tools/controllerbuilder/pkg/codegen"
-	"github.com/GoogleCloudPlatform/k8s-config-connector/dev/tools/controllerbuilder/pkg/gocode"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/dev/tools/controllerbuilder/pkg/options"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/dev/tools/controllerbuilder/pkg/typeupdater"
 
 	"github.com/spf13/cobra"
-	"google.golang.org/protobuf/proto"
-	"google.golang.org/protobuf/reflect/protodesc"
-	"google.golang.org/protobuf/reflect/protoreflect"
-	"google.golang.org/protobuf/types/descriptorpb"
-	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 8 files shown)


docs/develop-resources/scenarios/new-field.md
```diff
@@ -11,23 +11,22 @@ REPO_ROOT="$(git rev-parse --show-toplevel)"
...
 go run . update-types \
-    --parent-message-full-name "google.monitoring.dashboard.v1.Dashboard" \
-    --new-field "row_layout" \
+    --parent-message "google.monitoring.dashboard.v1.Dashboard" \
+    --field-to-insert "row_layout" \
     --api-dir ${REPO_ROOT}/apis/monitoring/v1beta1
 ```
 
-* `--parent-message-full-name`
+* `--parent-message`
```
yuwenma, line 19: Suggest using no more than 2 words for flag name. Otherwise it would be very hard to use.
how about `parent` and adding an example in the desription.

docs/develop-resources/scenarios/new-field.md
```diff
@@ -11,23 +11,22 @@ REPO_ROOT="$(git rev-parse --show-toplevel)"
...
 
 Fully qualified name of the proto message holding the new field.
 
-* `--new-field`
-
-Name of the new field.
+* `--field-to-insert`
```
yuwenma, line 23: May I ask why we make this change? I think `new-field` and `field-to-insert` mean the same thing but the latter is a less typical name.

dev/tools/controllerbuilder/pkg/commands/updatetypes/updatetypescommand.go
yuwenma, line 1: What is the test plan for the insert tool?

yuwenma (commented):
/lgtm
/approve

THanks!

docs/develop-resources/scenarios/new-field.md
jingyih: Yeah we can use `parent`.

dev/tools/controllerbuilder/pkg/commands/updatetypes/updatetypescommand.go
jingyih: I am manually updating some of the existing direct resources. When there is a new field detected, I will use this tool. Next step is to try to have AI call this function.

docs/develop-resources/scenarios/new-field.md
jingyih: We could revert back to `--new-field`, or call it `--insert-field`. In my mind, I was thinking maybe in the future we also want to have flags for removing field, update a field (new comment), etc.

PR #3340, opened 2024-12-09
feat: Add fuzz test for FirestoreDatabase

jasonvigil:
Seems there were quite a few fields added to FirestoreDatabase since it was originally implemented.

### DIFF
--- a/pkg/controller/direct/firestore/firestoredatabase_controller.go
+++ b/pkg/controller/direct/firestore/firestoredatabase_controller.go
@@ -25,10 +25,12 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/fuzztesting"
 
 	gcp "cloud.google.com/go/firestore/apiv1"
 	apiv1 "cloud.google.com/go/firestore/apiv1/admin"
 	firestorepb "cloud.google.com/go/firestore/apiv1/admin/adminpb"
+	pb "cloud.google.com/go/firestore/apiv1/admin/adminpb"
 	"google.golang.org/api/option"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/fieldmaskpb"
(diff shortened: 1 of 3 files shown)


yuwenma (commented):
/approve

cc @jingyih for another pass

pkg/fuzztesting/fuzzkrm.go
```diff
@@ -141,6 +141,9 @@ func (f *FuzzTest[ProtoT, KRMType]) Fuzz(t *testing.T, seed int64) {
 		t.Fatalf("error mapping from krm to proto: %v", ctx.Err())
 	}
 
+	// Remove any output only or known-unimplemented fields
+	fuzz.Visit("", p2.ProtoReflect(), nil, clearFields)
```
jingyih, line 145: Not sure why we need to clear fields again. cc @acpana

jasonvigil: This controller takes an opinion on the values that should be set for `type`, `app_engine_integration_mode`, and `deletion_protection_state`. Values for those fields are set in the `ToProto` method, regardless of what is passed in.

jingyih: This behavior seems resource-specific, but is currently applied library-wide, affecting all resources. Ideally we want to scope it down to individual resource fuzz tests. A potential solution is to add something like a `defaultFields` to the KRMTypedFuzzer struct. The `defaultFields` can be set at resource level, and be cleared in library after p2 is calculated.

jasonvigil: Just to clarify, those fields are not settable in the KRM resource. They are given hardcoded, default values in order to make the resource declarative. Not sure if this is a common scenario. Though, is there any reason we would not want this behavior (of clearing unimplemented fields) for all resources?

jingyih: > Though, is there any reason we would not want this behavior (of clearing unimplemented fields) for all resources?

My understanding is that it defeats the purpose of roundtrip. For example,
 - The KRM -> proto conversion is incorrectly adding data to ignored / unimplemented field.
 - or, the proto -> KRM conversion is converting data to the wrong field that then gets mapped back to ignored field (I am thinking about in a spec fuzz test where the mapper functions were incorrectly converting proto field to a status field).

Although this is likely a rare edge case, we should ideally catch it when it happens.

If a proper fix is not straightforward, we can merge this PR as is and add a TODO for later, to unblock other PRs.

jasonvigil: > My understanding is that it defeats the purpose of roundtrip. 

Got it, makes sense. Thanks for the examples. I will try to figure out a better solution for this.

jasonvigil: Ok, I have fixed this @jingyih. I moved the defaulting behavior out of the mapper and into a separate function, and marked all the "default value" fields as unimplemented (in the fuzz test). This organization is probably better for maintainability / readability anyway.

yuwenma (commented):
/lgtm
/approve

PR #3341, opened 2024-12-09
feat: Promote FirestoreDatabase to v1beta1

jasonvigil:


pkg/clients/generated/apis/firestore/v1alpha1/doc.go
yuwenma, line 1: Suggest dropping the change under `pkg/clients/generated` because of this https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/clients/generated/README.md

yuwenma (commented):
/approve

one nit, otherwise looks good!

pkg/clients/generated/apis/firestore/v1alpha1/doc.go
jasonvigil: I am trying to keep the generated clients updated for all these legacy resources that were created before that change. No reason why, other than just consistency. For all future direct controllers, will not generate the clients. Is it ok to update the generated client in this case, even though it isn't used?

yuwenma: I understand the consistency concern. But the way to generate these code introduces bug to our code base (and we won't fix them). For direct resources, the lib is already in `./apis`. So updating the generated clients lib is more than just introducing duplicate but buggy code

jasonvigil: After offline discussion, I now recognize these generated libs are for customer use, and we said they are no longer updated. Reverted these changes.

yuwenma:
Can you add this feature to the release note?

PR #3343, opened 2024-12-09
chore: turn on readOnlyRootFilesystem for GKE add-on release'

yuwenma:
This is required by GKE add-on

### DIFF
--- a/operator/config/gke-addon/manager_patch.yaml
+++ b/operator/config/gke-addon/manager_patch.yaml
@@ -30,3 +30,6 @@
     postStart:
       exec:
         command: ["./gke_addon_poststart"]
+- op: add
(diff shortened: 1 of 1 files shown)


operator/config/gke-addon/manager_patch.yaml
```diff
@@ -29,4 +29,7 @@
   value:
     postStart:
       exec:
-        command: ["./gke_addon_poststart"]
\ No newline at end of file
+        command: ["./gke_addon_poststart"]
+- op: add
+  path: /spec/template/spec/containers/0/securityContext/readOnlyRootFilesystem
```
yuwenma, line 34: We can also turn on the readOnlyRootFilesystem for all KCC release approaches (manual, CC, gke add-on). But I'd prefer do it step by step

PR #3345, opened 2024-12-10
fix: add version

acpana:
Super small fix to add the `Version` field to the FromProto conversion.


jingyih:
Do we have a Fuzz test for this resource?

acpana:
> Do we have a Fuzz test for this resource?

No because the `proto` for that resource doesn't actually implement the `protomessage` interface. But I am working on one here -- https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/acpana/krm-fuzzer-gkehubfm . That's how I found the need to add this change.

In the future, we will check in that fuzzer! But not now.

jingyih (commented):
/lgtm

jingyih:
Oh, but I couldn't find the `ConfigManagementMembershipSpec_ToProto` function, how do we test roundtrip?

jingyih:
> Oh, but I couldn't find the `ConfigManagementMembershipSpec_ToProto` function, how do we test roundtrip?

Chatted with Alex offline, the equivalent function is called `featureMembershipSpecKRMtoMembershipFeatureSpecAPI`.

yuwenma (commented):
/lgtm
/approve

Is this for export?

PR #3346, opened 2024-12-10
fix: only patch membership config for one membership

ziyue-101:
### Change description

Before, the controller patches the whole GKEHub Feature resource, which may contains multiple membershipSpecs.

This broad update is not a problem for the associated membership, but would be a problem for other non-associated ones(created by other means) when there are mutually exclusive fields like management and version. The two fields are mutually exclusive in the request, but not in the GET response. The gkehub server defaults the version field if management field is set as automatic(they are both present in the GET response).  Thus, during the patch, the controller sends the request with what it gets from the server for non-associated memberships, which would get rejected by the server because both field are present.

The gkehub server has special logic to prevent other memberships being removed from the feature resource. It will only remove memberships only if they are explicitly set the in the membershipspecs map and with an empty config.

The controller should not reconcile on other non-assoicated memberships. This PR limits the reconciliation to one membership.

The customer impact is if customers have other membership clusters with cs auto-upgrade but not created through KCC. the controller won't be able to reconcile because of the two conflicting fields during update.  But the existing feature(cs, mesh, poco) controllers are not affected.

### Tests you have done

### DIFF
--- a/pkg/controller/direct/gkehub/featuremembership_controller.go
+++ b/pkg/controller/direct/gkehub/featuremembership_controller.go
@@ -183,16 +183,15 @@ func (a *gkeHubAdapter) Delete(ctx context.Context, deleteOp *directbase.DeleteO
 
 func (a *gkeHubAdapter) patchMembershipSpec(ctx context.Context) ([]byte, error) {
 	feature := a.actual
-	mSpecs := feature.MembershipSpecs
-	if mSpecs == nil {
-		mSpecs = make(map[string]featureapi.MembershipFeatureSpec)
-	}
+	mSpecs := make(map[string]featureapi.MembershipFeatureSpec)
 	// only change the feature configuration for the associated membership
 	desiredApiObj, err := featureMembershipSpecKRMtoMembershipFeatureSpecAPI(&a.desired.Spec)
 	if err != nil {
 		return nil, err
 	}
 	mSpecs[a.membershipID] = *desiredApiObj
+	// MembershipSpecs is a map of membership spec. Here we only patch one membership.
+	// GKE Hub server doesn't patch other memberships if they are not present in the membershipSpecs map.
 	feature.MembershipSpecs = mSpecs
 	op, err := a.hubClient.featureClient.Patch(a.featureID, feature).UpdateMask("membershipSpecs").Context(ctx).Do()
 	if err != nil {
--- a/mockgcp/mockgkehub/gkehubfeature.go
+++ b/mockgcp/mockgkehub/gkehubfeature.go
@@ -116,7 +116,7 @@ func (s *GKEHubFeature) UpdateFeature(ctx context.Context, req *pb.UpdateFeature
 		case "spec":
 			obj.Spec = req.GetResource().Spec
 		case "membershipSpecs":
-			obj.MembershipSpecs = req.GetResource().GetMembershipSpecs()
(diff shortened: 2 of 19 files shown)


ziyue-101:
/assign @acpana @maqiuyujoyce 
/CC @yuwenma Would it possible to include the change in the 1.126 release?

ziyue-101:
Failed test `--- FAIL: TestAllInSeries/samples/bigquerydatatransferconfig-scheduledquery` is unrelated to the change

pkg/controller/direct/gkehub/featuremembership_controller.go
```diff
@@ -183,16 +183,15 @@ func (a *gkeHubAdapter) Delete(ctx context.Context, deleteOp *directbase.DeleteO
...
-	}
+	mSpecs := make(map[string]featureapi.MembershipFeatureSpec)
 	// only change the feature configuration for the associated membership
 	desiredApiObj, err := featureMembershipSpecKRMtoMembershipFeatureSpecAPI(&a.desired.Spec)
 	if err != nil {
 		return nil, err
 	}
 	mSpecs[a.membershipID] = *desiredApiObj
+	// MembershipSpecs is a map of membership spec. Here we only patch one membership.
+	// GKE Hub server doesn't not touch other memberships if they are not present in the membershipSpecs map.
```
acpana, line 194:
```suggestion
	// GKE Hub server doesn't paych other memberships if they are not present in the membershipSpecs map.
```

pkg/controller/direct/gkehub/featuremembership_controller.go
acpana, line 1: Thanks for digging into this 💯 and the change. I think this change can go into the next release.

maqiuyujoyce:
/lgtm

This looks good to me. Thank you so much for digging and fixing, @ziyue-101 !

Not approving because @yuwenma will want to manage the timeline for its merge.

yuwenma:
Thanks Ziyue! 
Could you upload a scenario test to verify this change?

ziyue-101:
> Thanks Ziyue! Could you upload a scenario test to verify this change?

Done

maqiuyujoyce (commented):
/lgtm
/approve

Thank you so much for adding the scenario test, @ziyue-101 !

/hold

@yuwenma do you have any additional comment?

pkg/controller/direct/gkehub/featuremembership_controller.go
```diff
@@ -183,16 +183,15 @@ func (a *gkeHubAdapter) Delete(ctx context.Context, deleteOp *directbase.DeleteO
...
-	}
+	mSpecs := make(map[string]featureapi.MembershipFeatureSpec)
 	// only change the feature configuration for the associated membership
 	desiredApiObj, err := featureMembershipSpecKRMtoMembershipFeatureSpecAPI(&a.desired.Spec)
 	if err != nil {
 		return nil, err
 	}
 	mSpecs[a.membershipID] = *desiredApiObj
+	// MembershipSpecs is a map of membership spec. Here we only patch one membership.
+	// GKE Hub server doesn't not patch other memberships if they are not present in the membershipSpecs map.
```
yuwenma, line 194: typo: "doesn't ~not~"

yuwenma (commented):
/lgtm
/approve
/hold
Thanks! one typo in case you want to fix in this PR

pkg/controller/direct/gkehub/featuremembership_controller.go
ziyue-101: done, thanks!

PR #3355, opened 2024-12-11
tool: allow `Spec` to map to `Status` or `ObservedState` in mapper generator

jingyih:
In our codebase, some direct resources split a proto message into `Spec` and `Status` (as opposed to `Spec` and `ObservedState`). Example:

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/dd2d99cfd42d275c6b6ace78469f9beef54d67ea/apis/bigqueryconnection/v1beta1/connection_types.go#L118-L123

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/dd2d99cfd42d275c6b6ace78469f9beef54d67ea/apis/bigqueryconnection/v1beta1/connection_types.go#L233-L236

We should allow `Spec` to map to `Status` or `ObservedState` in mapper generator.

### DIFF
--- a/dev/tools/controllerbuilder/pkg/codegen/mappergenerator.go
+++ b/dev/tools/controllerbuilder/pkg/codegen/mappergenerator.go
@@ -26,6 +26,7 @@ import (
 
 	"google.golang.org/protobuf/reflect/protoreflect"
 	"google.golang.org/protobuf/types/descriptorpb"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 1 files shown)


yuwenma:
I think the BQCC AWSPropertyStatus is manually added.

jingyih:
> I think the BQCC AWSPropertyStatus is manually added.

Yeah, in @jasonvigil's output field design doc, the proposed naming convention is to append `ObservedState`.

This PR introduces a small update to ensure the mapper generator is compatible with the legacy naming conventions. A few Go structs in BQCC currently end with the suffix `Status`, and other resources might have also followed this pattern.

jingyih:
When migrating an existing TF-based beta resource to direct reconciler, output-only fields often end up under `Status` instead of `ObservedState`. I believe this PR is needed for the mapper generator to correctly generate mapping functions for these fields.

/cc @yuwenma

jingyih:
This one is probably obsolete.

/close

PR #3356, opened 2024-12-11
refactor:tests: add overrides

acpana:
This change adds an overrides func to deal with fields that look like a string but may behave differently. I'm removing the custom logic for enums and we can push it all in the overrides. We can build custom overrides like string -> int or string -> time and have those live closer to the filler too.

### DIFF
--- a/pkg/test/fuzz/krmgen.go
+++ b/pkg/test/fuzz/krmgen.go
@@ -20,36 +20,46 @@ import (
 	"testing"
 )
 
+type OverrideFiller func(t *testing.T, fieldName string, field reflect.Value)
 type RandomFiller struct {
 	randStream *rand.Rand
 
-	// for iota based enums, defines the upper bound for a named enum type
-	intEnumAllowableValues map[string]int64
-	// for non iota based enums, holds the set of allowable values for a named enum type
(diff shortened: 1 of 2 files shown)


pkg/test/fuzz/krmgen.go
```diff
@@ -20,51 +20,54 @@ import (
...
-	intEnumAllowableValues map[string]int64
-	// for non iota based enums, holds the set of allowable values for a named enum type
-	stringEnumAllowableValues map[string][]interface{}
+	fieldOverrides map[string]OverrideFiller
 }
 
-func NewRandomFiller(seed int64, enumBoundsMap map[string]int64, enumValuesMap map[string][]interface{}) *RandomFiller {
+type FillerConfig struct {
+	Stream         *rand.Rand
+	FieldOverrides map[string]OverrideFiller
```
xiaoweim, line 32: I like this! :)

yuwenma:
/approve

I'd like to have @jingyih to take another pass.

jingyih (commented):
I'd like a quick chat to understand the context better. Since this only impacts fuzz testing, please feel free to merge this if it's blocking other PRs.

/lgtm
/hold

PR #3364, opened 2024-12-12
Add Spanner instance unset compute field test

anhdle-sso:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/spanner/v1beta1/spanneinstance-unset-compute-fields/_http.log
+++ b/pkg/test/resourcefixture/testdata/basic/spanner/v1beta1/spanneinstance-unset-compute-fields/_http.log
@@ -1,0 +1,339 @@
+GET https://spanner.googleapis.com/v1/projects/${projectId}/instances/spannerinstance-sample-${uniqueId}?alt=json
+Content-Type: application/json
+User-Agent: Terraform/ (+https://www.terraform.io) Terraform-Plugin-SDK/2.10.1 terraform-provider-google-beta/kcc/controller-manager
+
+404 Not Found
+Cache-Control: private
+Content-Type: application/json; charset=UTF-8
+Server: ESF
+Vary: Origin
+Vary: X-Origin
+Vary: Referer
+X-Frame-Options: SAMEORIGIN
+X-Xss-Protection: 0
+
+{
+  "error": {
+    "code": 404,
+    "message": "Instance not found: projects/${projectId}/instances/spannerinstance-sample-${uniqueId}",
+    "status": "NOT_FOUND"
+  }
+}
+
+---
+
+POST https://spanner.googleapis.com/v1/projects/${projectId}/instances?alt=json
+Content-Type: application/json
+User-Agent: Terraform/ (+https://www.terraform.io) Terraform-Plugin-SDK/2.10.1 terraform-provider-google-beta/kcc/controller-manager
+
+{
+  "instance": {
+    "config": "projects/${projectId}/instanceConfigs/regional-us-west1",
+    "displayName": "Spanner Instance Sample",
+    "labels": {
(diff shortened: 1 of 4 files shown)


pkg/test/resourcefixture/testdata/basic/spanner/v1beta1/spanneinstance-unset-compute-fields/create.yaml
```diff
@@ -0,0 +1,24 @@
...
+apiVersion: spanner.cnrm.cloud.google.com/v1beta1
+kind: SpannerInstance
+metadata:
+  labels:
+    label-one: "value-one"
+  name: spannerinstance-sample-${uniqueId}
+spec:
+  config: regional-us-west1
+  displayName: Spanner Instance Sample
+  numNodes: 2
```
yuwenma, line 24: Can we have both numNodes and processingUnit configured at the same time?

pkg/test/resourcefixture/testdata/basic/spanner/v1beta1/spanneinstance-unset-compute-fields/_http.log
```diff
@@ -0,0 +1,339 @@
...
+  "instance": {
+    "displayName": "New spanner Instance Sample",
+    "labels": {
+      "cnrm-test": "true",
+      "label-one": "value-one",
+      "managed-by-cnrm": "true"
+    },
+    "name": "projects/${projectId}/instances/%!s(\u003cnil\u003e)",
+    "nodeCount": 2,
+    "processingUnits": 2000
```
yuwenma, line 188: I see the TF-based considers unset as a no-op. I think that is not right (unless the GCP service assigns a default value in creation). If no defaulting, when migrating to the direct approach, we want to change this behavior to **require** user to configure one of the field (unless autoscaling is turned on).

pkg/test/resourcefixture/testdata/basic/spanner/v1beta1/spanneinstance-unset-compute-fields/create.yaml
anhdle-sso: No, API would give you an error when setting both fields.

pkg/test/resourcefixture/testdata/basic/spanner/v1beta1/spanneinstance-unset-compute-fields/_http.log
anhdle-sso: if there is no value at creation, TF assign 1 as a value for nodecount. I agree with the check to ask for at least 1 fields, which would keep the controller and API in sync.

anhdle-sso: I can update the test case in the direct controller PR.

pkg/test/resourcefixture/testdata/basic/spanner/v1beta1/spanneinstance-unset-compute-fields/create.yaml
yuwenma: I mean, I suggest we make both numNodes and processingUnit to be configurable. I understood the GCP service gives a no op, but it also contains the defaulting logic (the unset field will be apply to a default value). Thus, I suggest we make the KRM API behaves as
1. If both are sets, choose one (maybe the max?) and unset the other when sending the requests to GCP
2. If both fields and the autoscaling are set. Raise errors to customers because "autoscaling does not allow setting the other two fields", rather than "because those two fields are set". 
This is not a Beta release blocker, but more an enhancement to the beta API behavior.

yuwenma (commented):
/lgtm
/approve

PR #3370, opened 2024-12-12
fix: dataflowflextemplatejob specialize subnetwork format

yuwenma:
### Change description

Fixes b/383648533

### Tests you have done

### DIFF
--- a/pkg/controller/direct/dataflow/refs.go
+++ b/pkg/controller/direct/dataflow/refs.go
@@ -16,6 +16,8 @@ package dataflow
 
 import (
 	"context"
+	"fmt"
+	"strings"
 
 	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -59,7 +61,7 @@ func (r *refNormalizer) VisitField(path string, v any) error {
 	}
 
 	if subnetworkRef, ok := v.(*refs.ComputeSubnetworkRef); ok {
-		resolved, err := refs.ResolveComputeSubnetwork(r.ctx, r.kube, r.src, subnetworkRef)
+		resolved, err := RefineComputeSubnetworkRef(r.ctx, r.kube, r.src, subnetworkRef)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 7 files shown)


pkg/controller/direct/dataflow/refs.go
```diff
@@ -79,3 +81,41 @@ func (r *refNormalizer) VisitField(path string, v any) error {
...
 	return nil
 }
+
+// RefineComputeSubnetworkRef refine the subnetwork format because DataflowFlexTemplateJob has a specific format requirement:
+// "You can specify a subnetwork using either a complete URL or an abbreviated path.
+//
+//	Expected to be of the form "https://www.googleapis.com/compute/v1/projects/HOST_PROJECT_ID/regions/REGION/subnetworks/SUBNETWORK"
+//	or "regions/REGION/subnetworks/SUBNETWORK". If the subnetwork is located in a Shared VPC network,
+//	you must use the complete URL.
+func RefineComputeSubnetworkRef(ctx context.Context, reader client.Reader, src client.Object, ref *refs.ComputeSubnetworkRef) (*refs.ComputeSubnetworkRef, error) {
```
justinsb, line 91: So the challenge here is going to be that we want to move this resolving logic into shared code (no longer per-controller).  Not sure what we should do here....

yuwenma: Agree. The good (or bad?) news is that this format requirement is specific to the Dataflow service. I checked another resource which depends on this subnetwork, and it has a different format requirements. See [computeForwardingRule](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/regionalcomputeforwardingrulefull/_http.log#L563)

justinsb: Right, so I think we need a plan.  Some options: Do we want to move normalization into a common area?  Do we want the ability to override normalization (for a type/field) in a service?  Should we normalize and then post-normalize to deal with weird formats?  Do we want to use tags to specify which format we want?

yuwenma: I split the problem into two sections: what APIs users have, what formats KCC needs to validate.
- The API is always convention based, and accept well documented, resource specific exceptions. Like this one. So we support both  `project/<project>/instance/<instanceID>` (convention), and `https://compute/v1/project/<project>/instance/<instanceID>` (for `DataflowFlexTemplateJob` only).
- I feel like the format validation should be in its granular level and be more specific to a resource. So we always validate the conventional format in apis/<resource>/refs.go, and if custom validation is needed, it converts before and after the normalizeExternal to keep the dependent resource  the conventional format in its own code.

justinsb: We discussed and the plan is to move all this reference normalization under common/, but we can merge this first

justinsb (approved):

PR #3371, opened 2024-12-12
feat: SecretManagerSecret full test coverage & manual replication field

yuwenma:
Improve SecretManager with full test coverage. 
2nd git-commit is real gcp
3rd git-commit is mockgcp

Filed #3395 to capture the TTL issue


apis/secretmanager/v1beta1/secret_types.go
```diff
@@ -48,8 +48,7 @@ type SecretManagerSecretSpec struct {
 	//  This is always provided on output, regardless of what was sent on input.
 	ExpireTime *string `json:"expireTime,omitempty"`
 
-	// Input only. The TTL for the
-	//  [Secret][google.cloud.secretmanager.v1.Secret].
+	// Input only. A duration in seconds with up to nine fractional digits, ending with 's'. Example: "3.5s".
```
yuwenma, line 51: The TTL is a stale field that has causes issues (it uses "Now()" when the resource is created that users cannot control or predict)

dev/tasks/run-e2e
```diff
@@ -26,7 +26,7 @@ if [[ -z "${KUBEBUILDER_ASSETS:-}" ]]; then
 fi
 
 if [[ -z "${KCC_USE_DIRECT_RECONCILERS:-}" ]]; then
-  KCC_USE_DIRECT_RECONCILERS=ComputeForwardingRule,GKEHubFeatureMembership,SecretManagerSecret,SecretManagerSecretVersion
```
yuwenma, line 29: The test now covers both legacy controller and direct controller (via alpha direct annotation). we no longer need this

pkg/controller/direct/secretmanager/secret_controller.go
```diff
@@ -122,6 +122,18 @@ func normalizeExternal(ctx context.Context, reader client.Reader, src client.Obj
 				secret.Spec.Replication.LegacyAutomatic.CustomerManagedEncryption.KmsKeyRef = kmsKeyRef
 			}
 		}
+		if secret.Spec.Replication.UserManaged != nil {
+			for _, r := range secret.Spec.Replication.UserManaged.Replicas {
+				if r.CustomerManagedEncryption != nil {
+					kmsKeyRef := r.CustomerManagedEncryption.KmsKeyRef
```
yuwenma, line 128: I previously missed this field because I thought it is another field `spec.customerManagedEncryption`, which is not supported yet in the current released beta API.

pkg/test/resourcefixture/testdata/basic/secretmanager/v1beta1/fullsecretmanagersecret-auto-legacy/_http.log
```diff
@@ -635,9 +642,57 @@ X-Xss-Protection: 0
 
 ---
 
-DELETE https://secretmanager.googleapis.com/v1/projects/${projectId}/secrets/secretmanagersecret-${uniqueId}?alt=json
+GET https://secretmanager.googleapis.com/v1/projects/${projectId}/secrets/secretmanagersecret-${uniqueId}?%24alt=json%3Benum-encoding%3Dint
```
yuwenma, line 645: The extra GET comes from the golden Export file.

pkg/test/resourcefixture/testdata/basic/secretmanager/v1beta1/secretmanagersecretversion/create.yaml
```diff
@@ -17,6 +17,7 @@ kind: SecretManagerSecretVersion
 metadata:
   annotations:
     cnrm.cloud.google.com/project-id: ${projectId}
+    alpha.cnrm.cloud.google.com/reconciler: direct
```
jasonvigil, line 20: Now that the controller for this test is switched to direct, looks you will need to re-record VCR (following https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/tests/README.VCRTesting.md#record-mode). Or, maybe we can remove VCR coverage?

yuwenma: I regenerated the VCR log

apis/secretmanager/v1beta1/secret_types.go
justinsb: Ouch.  This doesn't feel like a spec field, because if I specify `ttl: 60s` what is the TTL value in 10 seconds?  It seems like it should be 60s, because that is my desired state.  But I suspect it will be 50s.

mockgcp/mocksecretmanager/secrets.go
```diff
@@ -84,6 +84,13 @@ func (s *SecretsV1) populateDefaultsForSecret(ctx context.Context, obj *pb.Secre
 			return fmt.Errorf("Aliases cannot be assigned to versions that don't exist")
 		}
 	}
+	// TTL and ExpireTime are OneOf, but the GCP service always converts TTL to expireTime before storing the object.
```
justinsb, line 87: This is a reasonable behaviour.  Maybe we should deprecate spec.ttl (and replace with `spec.initialTTL` - although if we are going to do that, we can just change the comment on the field)

mockgcp/mocksecretmanager/secrets.go
```diff
@@ -154,6 +161,10 @@ func (s *SecretsV1) UpdateSecret(ctx context.Context, req *pb.UpdateSecretReques
 			updated.Expiration = &pb.Secret_ExpireTime{
 				ExpireTime: req.Secret.GetExpireTime(),
 			}
+		case "ttl":
+			updated.Expiration = &pb.Secret_Ttl{
+				Ttl: req.Secret.GetTtl(),
```
justinsb, line 166: Nit: really?  I would have assumed it would set Expiration.ExpireTime, just like in create.  Still if this is what the logs tell us....

pkg/controller/direct/secretmanager/secret_controller.go
justinsb: Do we have a fuzz test :-)

mockgcp/mocksecretmanager/secrets.go
justinsb: Actually I don't know if we have test coverage of this in this PR (?).  It's not a big deal, because it's a mock (we can fix it when we need it), just that I would have guessed the other way!

In general I think the behaviour of this `ttl` field is very tricky.  Should we reconcile it (meaning secrets never expire)?  Should we only apply on create?  But if we do that, what happens when the user changes the TTL field in KRM?  What happens if they change a different field?

apis/secretmanager/v1beta1/secret_types.go
yuwenma: Yes, this is exactly the problem!

pkg/controller/direct/secretmanager/secret_controller.go
yuwenma: PR on the way.

mockgcp/mocksecretmanager/secrets.go
yuwenma: Yes... here's what the SecertManager server tells us  https://screenshot.googleplex.com/9giJXnHPt3nMSwm.png

justinsb: Oh I see.  We set ttl, and then in `populateDefaultsForSecret` we convert it.  I guess that works, because then we can use the shared fieldmask update code at some point in the future!

yuwenma:
Filed #3771 to track the TTL issue

pkg/controller/direct/secretmanager/secret_controller.go
```diff
@@ -246,7 +258,10 @@ func (a *Adapter) Update(ctx context.Context, op *directbase.UpdateOperation) er
 	if err != nil {
 		return err
 	}
-
+	if paths.Has("ttl") {
+		paths = paths.Delete("ttl")
+		resource.Expiration = a.actual.Expiration
```
justinsb, line 263: Nit: there should be a comment here, because I don't think this does anything unless `expiration` is set in `paths`

justinsb (approved):

PR #3377, opened 2024-12-12
fix: Update resource status even if no GCP fields are updated

jasonvigil:
Even when adopting resources, we want to update status.


yuwenma (commented):
/lgtm
/approve

Thanks!

PR #3381, opened 2024-12-13
tool: add annotation to the generated Go fields with their corresponding proto fields

jingyih:
This provides an easier way to track the mapping between generate Go fields and their corresponding proto fields. I think we should start tracking this information.

/cc @jasonvigil @yuwenma


yuwenma (commented):
/lgtm
/approve

PR #3314, opened 2024-12-04
tests:feat: enable adc

acpana:
Small but powerful change to enable us to use APIs that require a quota project to be set. IOW, with this change I can run 

```bash
$ WRITE_GOLDEN_OUTPUT=1 GOLDEN_REQUEST_CHECKS=1 \
GOLDEN_OBJECT_CHECKS=1 E2E_KUBE_TARGET=envtest RUN_E2E=1 \
E2E_GCP_TARGET=real \
 go test -test.count=1 -timeout 3600s -v ./tests/e2e -run TestAllInSeries/fixtures/bigqueryanalyticshublisting-base  2>&1
```
and have the direct controller in particular set the right header for `x-goog-user-project`.

PR #3386, opened 2024-12-13
tests:feat:integ: adc for direct

acpana:
This is a companion PR to https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3314 but for our integ (post submit) dynamic test suite. Tested out w a resource I am working on. Prior to this change, there's no plumbing for the `X-goog-user-project` header in the actual test setup.

### DIFF
--- a/pkg/test/controller/reconciler/testreconciler.go
+++ b/pkg/test/controller/reconciler/testreconciler.go
@@ -19,7 +19,9 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"os"
 	"regexp"
+	"strconv"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


maqiuyujoyce:
@acpana have you verified locally for all resources that this config works? I can't be sure if the following logic works for all non-direct resources, so could you use the code from master HEAD to create a baseline test result, and then run the test with this change to ensure the test result is the same (ish)?

```
	p := testgcp.GetDefaultProject(t)
	// technically this would be sufficient for TF and DCL controllers
	// to set the BillingProject as the resource project that is being used.
	kccConfig.UserProjectOverride = true
```

ldanielmadariaga:

PR #3392, opened 2024-12-16
fix:  golang non-constant format string

yuwenma:
Maybe introduced from recent golang change.


justinsb (approved):

maqiuyujoyce:
/lgtm

Just to confirm, is this the only place where string format is broken?

PR #3375, opened 2024-12-12
tool: support using config files in codegen

jingyih:
The `generate.sh` script is becoming increasingly difficult to maintain as we add more resources.

This PR adds support for YAML-based config file to gradually replace the script.

PR #3396, opened 2024-12-16
chore: add presubmit check for generated types for direct resources

jingyih:
Follow up of #3375.

This PR adds a presubmit check to prevent accidental manual edits to the generated types for direct resources, ensuring the tools can be safely rerun. (Diff in copyright year within the generated file is ignored)

To ***opt-in*** the check, create a configuration YAML file and place it in the `dev/tools/controllerbuilder/config` directory. Here’s an example configuration file:

```
service: google.cloud.bigquery.datatransfer.v1
apiVersion: bigquerydatatransfer.cnrm.cloud.google.com/v1beta1
generateMapper: true
resources:
  - kind: BigQueryDataTransferConfig
    protoName: TransferConfig
    skipScaffoldFiles: true # files were scaffolded using a previous template, making them incompatible with the new scaffolding template.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -400,3 +400,13 @@ operator-e2e-tests:
 	export TEST_ORG_ID=${ORG_ID}
 	export TEST_BILLING_ACCOUNT_ID=${BILLING_ACCOUNT}
 	cd operator/tests/e2e/ && go test --project-id=${PROJECT_ID}
+
+# Generate Go types for direct resources specified in the config files located under `dev/tools/controllerbuilder/config`.
+.PHONY: generate-types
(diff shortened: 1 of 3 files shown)


jingyih:
/cc @yuwenma @justinsb

Let me know your thoughts. I prefer that we add this check to ensure the tools can be safely rerun on the generated code. In the future, this could become an exit criteria for a direct beta resource.

yuwenma (commented):
/lgtm
/approve

Nice! (What I missed!)

jingyih:
Rebased to resolve merge conflict.

PR #3400, opened 2024-12-17
DiscoveryEngineEngine controller

justinsb:
We create the controller, a test and the mocks.  I guess we could split out the mocks into their own PR, but this feels like a reasonable "chunk"

### DIFF
--- a/apis/discoveryengine/v1alpha1/engine_reference.go
+++ b/apis/discoveryengine/v1alpha1/engine_reference.go
@@ -43,6 +43,12 @@ type DiscoveryEngineEngineRef struct {
 	Namespace string `json:"namespace,omitempty"`
 }
 
+// DiscoveryEngineEngineID is the resolved identifier for a DiscoveryEngineEngine
+type DiscoveryEngineEngineID struct {
+	*CollectionLink
+	Engine string
+}
+
 // NormalizedExternal provision the "External" value for other resource that depends on DiscoveryEngineEngine.
 // If the "External" is given in the other resource's spec.DiscoveryEngineEngineRef, the given value will be used.
 // Otherwise, the "Name" and "Namespace" will be used to query the actual DiscoveryEngineEngine object from the cluster.
@@ -52,7 +58,7 @@ func (r *DiscoveryEngineEngineRef) NormalizedExternal(ctx context.Context, reade
 	}
 	// From given External
 	if r.External != "" {
-		id, err := parseDiscoveryEngineEngineExternal(r.External)
+		id, err := ParseDiscoveryEngineEngineExternal(r.External)
 		if err != nil {
 			return "", err
 		}
@@ -61,10 +67,10 @@ func (r *DiscoveryEngineEngineRef) NormalizedExternal(ctx context.Context, reade
 	}
 
 	// From the Config Connector object
-	if r.Namespace == "" {
-		r.Namespace = otherNamespace
-	}
 	key := types.NamespacedName{Name: r.Name, Namespace: r.Namespace}
+	if key.Namespace == "" {
+		key.Namespace = otherNamespace
+	}
 	u := &unstructured.Unstructured{}
 	u.SetGroupVersionKind(DiscoveryEngineEngineGVK)
 	if err := reader.Get(ctx, key, u); err != nil {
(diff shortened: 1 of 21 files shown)


justinsb:
We also use the common functions to resolve references (the last commit), cc @yuwenma

pkg/controller/direct/discoveryengine/mapper.go
```diff
@@ -65,7 +68,16 @@ func DiscoveryEngineEngineSpec_FromProto(mapCtx *direct.MapContext, in *pb.Engin
 	out.DisableAnalytics = direct.LazyPtr(in.GetDisableAnalytics())
 
 	for _, dataStoreID := range in.DataStoreIds {
-		out.DataStoreRefs = append(out.DataStoreRefs, &krm.DiscoveryEngineDataStoreRef{External: dataStoreID})
+		tokens := strings.Split(in.Name, "/")
```
justinsb, line 71: This (and the similar function below) is the trickiest thing in this PR, IMO.  The GCP API expects only the data_store_id, and implicitly resolves the project / location / collection because (I guess) we can't go across projects / locations / collections.

mockgcp/mockdiscoveryengine/datastore.go
```diff
@@ -48,6 +48,11 @@ func (s *dataStoreService) CreateDataStore(ctx context.Context, req *pb.CreateDa
 	}
 	now := time.Now()
 
+	if name.Collection != "default_collection" {
+		collectionLink := fmt.Sprintf("projects/%d/locations/global/collections/%s", name.Project.Number, name.Collection)
+		return nil, status.Errorf(codes.NotFound, "Collection with name %s does not exist.", collectionLink)
+	}
+
```
yuwenma, line 55: Shall we change the `spec.connection` to a reference?

apis/discoveryengine/v1alpha1/engine_reference.go
```diff
@@ -43,16 +43,22 @@ type DiscoveryEngineEngineRef struct {
...
+type DiscoveryEngineEngineID struct {
+	*CollectionLink
+	Engine string
+}
+
 // NormalizedExternal provision the "External" value for other resource that depends on DiscoveryEngineEngine.
 // If the "External" is given in the other resource's spec.DiscoveryEngineEngineRef, the given value will be used.
 // Otherwise, the "Name" and "Namespace" will be used to query the actual DiscoveryEngineEngine object from the cluster.
-func (r *DiscoveryEngineEngineRef) NormalizedExternal(ctx context.Context, reader client.Reader, otherNamespace string) (string, error) {
+func (r *DiscoveryEngineEngineRef) NormalizedExternal(ctx context.Context, reader client.Reader, defaultNamespace string) (string, error) {
```
yuwenma, line 55: I'd prefer "other" over "default" because this is the referrer's namespace, which overrides the "default". 
e.g.
```yaml
kind: OtherResource
metadata:
   name: other
   namespace: otherNamespace
spec:
    discoveryEnginRef:
         name:  this 
--- 
kind: DiscoveryEngineEngine
metadata:
   name: this
```
The `DiscoveryEngineEngine`  will use namespace "otherNamespace"

apis/discoveryengine/v1alpha1/engine_reference.go
```diff
@@ -61,10 +67,10 @@ func (r *DiscoveryEngineEngineRef) NormalizedExternal(ctx context.Context, reade
 	}
 
 	// From the Config Connector object
-	if r.Namespace == "" {
-		r.Namespace = otherNamespace
-	}
 	key := types.NamespacedName{Name: r.Name, Namespace: r.Namespace}
+	if key.Namespace == "" {
+		key.Namespace = defaultNamespace
+	}
```
yuwenma, line 73: oh no, I thought this has been fixed in the template.

apis/discoveryengine/v1alpha1/engine_reference.go
justinsb: OK, I'll update

justinsb: My thought was that this is actually the default value for the Namespace of the Ref.  That's the only namespace the _ref_ has.

apis/discoveryengine/v1alpha1/engine_reference.go
justinsb: Probably, but we're only at December

mockgcp/mockdiscoveryengine/datastore.go
justinsb: This is in mockgcp.

discoveryengine doesn't have gcloud support, so we have to do mockgcp at this point.

apis/discoveryengine/v1alpha1/engine_reference.go
justinsb: Just pushed changed with otherNamespace.  We probably need to cleanup the NormalizedExternal contract anyway so maybe we document it there

ldanielmadariaga:

PR #3407, opened 2024-12-18
tests: lint shortnames

acpana:
This keeps happening so we need a linter to prevent CRDs from losing or not having their short names defined.

I'll work on cutting down the exceptions file in a separate PR.


yuwenma (commented):
/lgtm
/approve

Thank you so much! ❤️

PR #3410, opened 2024-12-18
feat: Add fuzzer for WorkstationConfig

jasonvigil:


pkg/controller/direct/workstations/config_controller.go
```diff
@@ -39,6 +40,40 @@ import (
 
 func init() {
 	registry.RegisterModel(krm.WorkstationConfigGVK, NewWorkstationConfigModel)
+	fuzztesting.RegisterKRMFuzzer(workstationConfigFuzzer())
```
yuwenma, line 43: Just notice this, could you elaborate how to run this fuzz test if add the init here but not [this file](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/dev/tasks/find-missing-fields)?

yuwenma:
One question about trigging the fuzz test.
cc @jingyih who's working on detecting new field.

pkg/controller/direct/workstations/config_controller.go
jasonvigil: This seems to be a new(er) method for setting up fuzz tests. I think it's a bit easier to use. The tests are run here https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/95b3225e0040b41db93bb05d26bd4d6007dab44b/.github/workflows/presubmit.yaml#L219, or manually with `go test -v ./pkg/fuzztesting/fuzztests/ -fuzz=FuzzAllMappers -fuzztime 60s`

pkg/controller/direct/workstations/config_mappings.go
```diff
@@ -127,8 +127,6 @@ func WorkstationConfigSpec_ToProto(mapCtx *direct.MapContext, in *krm.Workstatio
 	out.ReadinessChecks = direct.Slice_ToProto(mapCtx, in.ReadinessChecks, WorkstationConfig_ReadinessCheck_ToProto)
 	out.ReplicaZones = in.ReplicaZones
 
-	ApplyWorkstationConfigGCPDefaults(mapCtx, in, out, actual)
```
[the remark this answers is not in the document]
jasonvigil: This had to be moved out so that the mapping function would satisfy the interface required by the fuzz tests. Applying the defaults requires passing in the actual WorkstationConfig proto.

pkg/controller/direct/workstations/config_controller.go
jingyih: Yeah I believe we have a new framework for fuzzers now. "find-missing-fields" is probably deprecated.

@acpana do we have plan to migrate old-styled fuzz tests and remove the deprecated files to avoid confusion?

yuwenma:
/lgtm
/approve
Nice! Thank you for the context!

PR #3411, opened 2024-12-18
feat: Add fuzzer for Workstation

jasonvigil:


pkg/controller/direct/workstations/workstation_controller.go
```diff
@@ -39,6 +40,32 @@ import (
...
+}
+
+func workstationFuzzer() fuzztesting.KRMFuzzer {
+	f := fuzztesting.NewKRMTypedFuzzer(&workstationspb.Workstation{},
+		WorkstationSpec_FromProto, WorkstationSpec_ToProto,
+		WorkstationObservedState_FromProto, WorkstationObservedState_ToProto,
+	)
+
+	f.UnimplementedFields.Insert(".name")
+	f.UnimplementedFields.Insert(".reconciling")
```
yuwenma, line 53: Can I ask what is reconciling? I can see why we skip `name`.

jasonvigil: It's an output-only field that we decided to not include in the resource. Originally this came up in WorkstationConfig, here: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2819#discussion_r1783238928

yuwenma (commented):
/lgtm
/approve

PR #3413, opened 2024-12-18
chore: switch controllerbuilder to use vertexai gemini endpoint

justinsb:
This means we don't need an API Key (but we do need a GCP account)


jingyih (commented):
Cool!

Looks like the new file is missing the license head.

/lgtm

yuwenma (commented):
/approve

PR #3415, opened 2024-12-18
tooling: add hints when we can't match a field

justinsb:
If there's a field with the same name but different case,
add the near-miss to the output.


jingyih (commented):
/lgtm

By the way, we are adding the `+kcc:proto` annotation to KRM fields as well. This should make it easier to identify the correct mappings for newly generated resources.

yuwenma:
/approve

+1 to use a different go tag

jingyih:
/lgtm

Using a different tag sounds good! Practically does it make a difference? I think when parsing a Go file, we can always determine whether we're dealing with a Go struct or a field within a struct. However, using a distinct tag might help simplify the parsing process.

PR #3417, opened 2024-12-18
fix: Do not add management conflict annotaiton for direct resources

jasonvigil:


yuwenma:
/approve

Thank you!

pkg/gvks/supportedgvks/supportedgvks.go
```diff
@@ -144,3 +144,17 @@ func BasedOnDCL(serviceMetaLoader metadata.ServiceMetadataLoader) []schema.Group
 	}
 	return gvkList
 }
+
+func IsDirectByGVK(gvk schema.GroupVersionKind) bool {
```
yuwenma, line 148: It's better to pass in gk here

yuwenma (commented):
/hold cancel

pkg/gvks/supportedgvks/supportedgvks.go
jasonvigil: Can you please explain why? I thought GVK is more precise.

PR #3238, opened 2024-11-21
Add SecureSourceManagerInstance PrivateConfig field API

ericpang777:
### Change description

Fixes #

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #3419, opened 2024-12-18
Add PrivateConfig field to SecureSourceManagerInstance

ericpang777:
### Change description

Adds the PrivateConfig field to the controller

### Tests you have done


ericpang777:
Branched off https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3238


@yuwenma  Is it okay to let this through despite the example not working in real gcp due to a limitation of the KCC PrivateCACertificateAuthority? The KCC PrivateCACertificateAuthority doesn't enable the CA for some reason, but if you enable it manually, this example would work, so it's not an issue with the logic of SSM.

ericpang777:
> Branched off #3238
> 
> @yuwenma Is it okay to let this through despite the example not working in real gcp due to a limitation of the KCC PrivateCACertificateAuthority? The KCC PrivateCACertificateAuthority doesn't enable the CA for some reason, but if you enable it manually, this example would work, so it's not an issue with the logic of SSM.

@yuwenma oh and this one as well

pkg/test/resourcefixture/testdata/basic/securesourcemanager/securesourcemanagerinstance/securesourcemanagerinstanceprivate/create.yaml
```diff
@@ -0,0 +1,26 @@
...
+metadata:
+  name: ssminstance-${uniqueId}
+spec:
+  location: us-central1
+  projectRef:
+    external: ${projectId}
+  privateConfig:
+    isPrivate: true
+    caPoolRef:
+      name: privatecacapool-${uniqueId}
```
gemmahou, line 26: I have an alternative idea: could we create a working CA pool(enabled) for this test? We could then use `external: projects/{project_id}/locations/{region}/caPools/{caPool}` here to reference that CA pool resource. This would bypass the issue when CA pool created by KCC not being enabled successfully.

If we choose this way, I'll log a bug for the KCC CA pool enabling issue when it's referenced by another resource and add a comment here explaining why we need to use an external reference.

ericpang777: Yep that should work!

ericpang777: Edit: We can have a static CA Pool, but it also needs an existing CertificateAuthority that is enabled.

gemmahou: We can leave the test as is, but need to disable it [here](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/f86cf73a45a76f6bbeaff2547702181072402b4a/pkg/test/constants/constants.go#L25) until we fix the CA issue. Otherwise the periodic test will fail once it's added.

I created a ticket for this issue: b/391704025, you can link this ticket in comment to explain why we disable the test.

gemmahou:
Everything else looks good to me

/lgtm

pkg/test/resourcefixture/testdata/basic/securesourcemanager/securesourcemanagerinstance/securesourcemanagerinstanceprivate/create.yaml
ericpang777: Added

ericpang777:
The test failure seems to be a diff on the request headers on the privateca mock 

2025-01-23T01:39:36.6995385Z         - 	User-Agent: kcc/controller-manager DeclarativeClientLib/0.0.1
2025-01-23T01:39:36.6996625Z         + 	User-Agent: kcc/${kccVersion} (+https://github.com/GoogleCloudPlatform/k8s-config-connector) kcc/controller-manager/${kccVersion} 


Will fetch and rebase from `master` to see if that fixes it

gemmahou:
/lgtm

@yuwenma - this is the PR to add a new `PrivateConfig` field to the SMInstance resource. Test has been temporarily disabled because the referenced PrivateCACertificateAuthority cannot be enabled by KCC, as logged in b/391704025. @ericpang777 verified the test by manually enabling the CA after creation.

PR #3422, opened 2024-12-18
refactor: no refs for BigQueryDatasetRef

acpana:
This PR gets rid of the "old" refs. usage for BigQueryDatasetRef. It doesn't get rid of it just yet; that will be in a separate PR when I change the `ResolveBigQueryTable` behavior

### DIFF
--- a/apis/bigquery/v1beta1/dataset_reference.go
+++ b/apis/bigquery/v1beta1/dataset_reference.go
@@ -71,9 +71,23 @@ func (r *DatasetRef) NormalizedExternal(ctx context.Context, reader client.Reade
 		return "", fmt.Errorf("reading referenced %s %s: %w", BigQueryDatasetGVK, key, err)
 	}
 	// Get external from status.externalRef. This is the most trustworthy place.
-	actualExternalRef, _, err := unstructured.NestedString(u.Object, "status", "externalRef")
+	actualExternalRef, found, err := unstructured.NestedString(u.Object, "status", "externalRef")
 	if err != nil {
 		return "", fmt.Errorf("reading status.externalRef: %w", err)
+	}
+	if !found {
+		// BigQueryDataset is still TF based so we resolve the reference from the object
+		projectID, err := refsv1beta1.ResolveProjectID(ctx, reader, u)
+		if err != nil {
+			return "", err
+		}
+		datasetID, err := refsv1beta1.GetResourceID(u)
+		if err != nil {
+			return "", err
+		}
+		actualExternal := fmt.Sprintf("projects/%s/datasets/%s", projectID, datasetID)
+		r.External = actualExternal
+		return r.External, nil
 	}
 	if actualExternalRef == "" {
 		return "", k8s.NewReferenceNotReadyError(u.GroupVersionKind(), key)
--- a/apis/bigquery/v1beta1/dataset_reference.go
+++ b/apis/bigquery/v1beta1/dataset_reference.go
@@ -77,6 +77,15 @@ func (r *DatasetRef) NormalizedExternal(ctx context.Context, reader client.Reade
 	}
 	if !found {
 		// BigQueryDataset is still TF based so we resolve the reference from the object
+		// BUT only construct the reference if the resource is ready
+		resource, err := k8s.NewResource(u)
+		if err != nil {
+			return "", fmt.Errorf("error converting unstructured to resource: %w", err)
+		}
+		if !k8s.IsResourceReady(resource) {
+			return "", k8s.NewReferenceNotReadyError(u.GroupVersionKind(), key)
+		}
+
 		projectID, err := refsv1beta1.ResolveProjectID(ctx, reader, u)
 		if err != nil {
 			return "", err
--- a/pkg/controller/direct/bigqueryanalyticshub/listing_controller.go
+++ b/pkg/controller/direct/bigqueryanalyticshub/listing_controller.go
@@ -97,11 +97,10 @@ func (m *modelListing) AdapterForObject(ctx context.Context, reader client.Reade
 func resolveOptionalReferences(ctx context.Context, reader client.Reader, obj *krm.BigQueryAnalyticsHubListing) error {
 	if obj.Spec.Source != nil && obj.Spec.Source.BigQueryDatasetSource != nil {
 		if ref := obj.Spec.Source.BigQueryDatasetSource.DatasetRef; ref != nil {
-			datasetID, err := refs.ResolveBigQueryDataset(ctx, reader, obj, ref)
+			_, err := obj.Spec.Source.BigQueryDatasetSource.DatasetRef.NormalizedExternal(ctx, reader, obj.GetNamespace())
 			if err != nil {
 				return err
 			}
-			obj.Spec.Source.BigQueryDatasetSource.DatasetRef.External = datasetID.String()
 
 			for _, selectedResource := range obj.Spec.Source.BigQueryDatasetSource.SelectedResources {
 				if ref := selectedResource.TableRef; ref != nil {
--- a/pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
+++ b/pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"reflect"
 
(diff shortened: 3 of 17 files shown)


yuwenma:
The validation failure should have been fixed. Could you rebase?

pkg/controller/direct/bigqueryanalyticshub/listing_controller.go
```diff
@@ -97,11 +97,10 @@ func (m *modelListing) AdapterForObject(ctx context.Context, reader client.Reade
 func resolveOptionalReferences(ctx context.Context, reader client.Reader, obj *krm.BigQueryAnalyticsHubListing) error {
 	if obj.Spec.Source != nil && obj.Spec.Source.BigQueryDatasetSource != nil {
 		if ref := obj.Spec.Source.BigQueryDatasetSource.DatasetRef; ref != nil {
-			datasetID, err := refs.ResolveBigQueryDataset(ctx, reader, obj, ref)
+			_, err := obj.Spec.Source.BigQueryDatasetSource.DatasetRef.NormalizedExternal(ctx, reader, obj.GetNamespace())
```
jingyih, line 100: Why should NormalizedExternal return a string if it's already modifying the ref directly?

jingyih (commented):
/lgtm

pkg/controller/direct/bigqueryanalyticshub/listing_controller.go
gemmahou: I thought about this, right, the code refactoring would apply to the interface and impact all existing refs. Should be done in a separate PR and it's not urgent(I guess :D)

apis/bigquery/v1beta1/dataset_reference.go
```diff
@@ -71,10 +72,42 @@ func (r *DatasetRef) NormalizedExternal(ctx context.Context, reader client.Reade
...
 	// Get external from status.externalRef. This is the most trustworthy place.
-	actualExternalRef, _, err := unstructured.NestedString(u.Object, "status", "externalRef")
+	actualExternalRef, found, err := unstructured.NestedString(u.Object, "status", "externalRef")
 	if err != nil {
 		return "", fmt.Errorf("reading status.externalRef: %w", err)
 	}
+	if !found {
+		// BigQueryDataset is still TF based so we resolve the reference from the object
+		// Fetch BigQueryDataset object to construct the external form.
+		dataset := &unstructured.Unstructured{}
```
jingyih, line 82: Didn't we just read the dataset before this block? Can we reuse it?

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
```diff
@@ -111,11 +111,10 @@ func (m *model) AdapterForObject(ctx context.Context, reader client.Reader, u *u
 
 	// Resolve BigQueryDataSet Ref
 	if obj.Spec.DatasetRef != nil {
-		dataset, err := refv1beta1.ResolveBigQueryDataset(ctx, reader, obj, obj.Spec.DatasetRef)
```
jingyih, line 114: This is a behavior change. The original code uses the datasetID, not the fully qualified resource name. The change is also reflected in the HTTP log.

apis/bigquery/v1beta1/dataset_reference.go
```diff
@@ -71,10 +72,42 @@ func (r *DatasetRef) NormalizedExternal(ctx context.Context, reader client.Reade
 		return "", fmt.Errorf("reading referenced %s %s: %w", BigQueryDatasetGVK, key, err)
 	}
 	// Get external from status.externalRef. This is the most trustworthy place.
-	actualExternalRef, _, err := unstructured.NestedString(u.Object, "status", "externalRef")
+	actualExternalRef, found, err := unstructured.NestedString(u.Object, "status", "externalRef")
 	if err != nil {
 		return "", fmt.Errorf("reading status.externalRef: %w", err)
 	}
+	if !found {
```
jingyih, line 79: What happens if the KRM YAML is applied to the cluster, but the KCC controller has not yet reconciled it? In this scenario, the `externalRef` would be missing, and the BigQueryDataset GCP resource is not yet created.

apis/bigquery/v1beta1/dataset_reference.go
acpana: ah good call!

apis/bigquery/v1beta1/dataset_reference.go
acpana: Yes and that's okay because it will be found eventually! (Say a few reconciles later after)

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
acpana: Let's chat a bit more about this (maybe offline?). I'd say it's not behavior change! We don't modify any behavior or format expectations for the YAML itself. Yes the payload format is changed to follow the external format. But I think that's allowed!

Also, at this stage, it's not possible to follow the existing pattern (new new reference style as opposed to the old new reference style being used here) and shove the ID into the External field without reparsing the external field here. Which ... we could do if we had to offer a backwards compatibility promise on the format of the resolved datasetref.

cc @gemmahou

jingyih: I disagree. I don't believe the `destinationDatasetId` field in GCP accepts a fully qualified dataset name. You are right that the YAML remains unchanged, but the HTTP payload sent to the GCP API is different.

apis/bigquery/v1beta1/dataset_reference.go
jingyih: This actually represents a significant behavior change on the KCC side. In the scenario I described, there is no GCP resource yet, but we are sending an HTTP request that references it. According to KCC conventions, if a dependent resource is not ready (e.g., still in the process of being created), the KCC controller waits until it is ready. The new behavior, however, causes the KCC controller to continuously send HTTP requests to the GCP API until the dependent resource is created in GCP.

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
gemmahou: Here are my thoughts on this:

1. Eventually, I would expect the resolver functions to be replaced by normalizedExternal. and **all** resources, no matter direct or TF/DCL based, to have their _identity.go and _reference.go(with normalizedExternal function inside) to be referenced.
2. To ensure backward compatibility, and avoid format mismatching from normalizedExternal(full name) vs the current format, I proposed an extra function in individual controllers to handle that case by case, because the format of refs can be different. Example code in my pending PR: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/b63ff072ca520522cd238b9761402efe7755d63b/pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go#L403, also line 409 and 414. 

For BQ Dataset, what's the current format of the ref?

jingyih: To handle this type of issue more systematically, we probably need a dedicated place to define the "format template." This template would specify which fields to extract from the resource being referenced and how to structure the final value using those fields.  

In this specific case, `BigQueryDataTransferConfig` references `BigQueryDataset` and needs to extract the "dataset ID" (resource ID). In another case, an `IAM resource` references `BigQueryConnectionConnection`, but it needs to extract `status.observedState.cloudSQL.serviceAccountID`.

apis/bigquery/v1beta1/dataset_reference.go
acpana: I've parked the findings from an offline discussion here: https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/3619

I think the current PR can be merged as is (in particular with the fix for `IsResourceReady`).

jingyih (commented):
/lgtm

Let's also monitor the post-submit Prow tests. Currently, both `TestCreateNoChangeUpdateDelete/bigqueryanalyticshub` and `TestCreateNoChangeUpdateDelete/bigquerydatatransfer` are passing. If any unexpected regressions occur, we need to address them ASAP.

PR #3427, opened 2024-12-19
Remove the CEL rule for resourceID in the template

maqiuyujoyce:
Context: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3281#discussion_r1886826560

`spec.resourceID` is mutable because we want users to be able to:
1. Set the correct resource ID even if it was previously unset.
2. Unset the field when needed.
3. Update the value when it was set to an incorrect resource ID.


jasonvigil:
Would we want to update existing APIs also? https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3434 does the update for all existing APIs

maqiuyujoyce:
This PR is a duplicate of part of #3434. Closing.

PR #3430, opened 2024-12-19
Promote Workstation to v1beta1

jasonvigil:


yuwenma (commented):
/lgtm
/approve

Awesome! THanks

apis/workstations/v1beta1/workstation_identity.go
```diff
@@ -0,0 +1,141 @@
...
+
+func (i *WorkstationIdentity) ID() string {
+	return i.id
+}
+
+func (i *WorkstationIdentity) Parent() *WorkstationParent {
+	return i.parent
+}
+
+type WorkstationParent struct {
```
yuwenma, line 44: Or we can use `WorkstationClusterIdentity`.  But we need some code refactoring to allow (and only allow) the `WorkstationClusterIdentity` initialization to be done via `NewWorkstationClusterIdentity` or its child's `NewIdentity` function.

apis/workstations/v1beta1/workstation_identity.go
```diff
@@ -0,0 +1,141 @@
...
+			Cluster:   cluster,
+			Config:    config,
+		},
+		id: resourceID,
+	}, nil
+}
+
+func ParseWorkstationExternal(external string) (parent *WorkstationParent, resourceID string, err error) {
+	tokens := strings.Split(external, "/")
+	if len(tokens) != 10 || tokens[0] != "projects" || tokens[2] != "locations" || tokens[4] != "workstationClusters" || tokens[6] != "workstationConfigs" || tokens[8] != "workstations" {
```
yuwenma, line 130: Or something like 
```go
tokens := strings.SplitN(external, "/workstations", 2)
if len(tokens) != 2 {
   
}
parent := ParseWorkstationConfigExternal(tokens[[0])
id := tokens[1]
```

apis/workstations/v1beta1/workstation_types.go
```diff
@@ -0,0 +1,132 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var WorkstationGVK = GroupVersion.WithKind("Workstation")
+
+// WorkstationSpec defines the desired state of Workstation
+// +kcc:proto=google.cloud.workstations.v1.Workstation
+type WorkstationSpec struct {
+	// Parent is a reference to the parent WorkstationConfig for this Workstation.
+	Parent *WorkstationConfigRef `json:"parentRef"`
```
yuwenma, line 28: I also did some exploration previously about simplifying the parent ref to use existing code and use go code to tell contributors what Parent means, and how it should be used. These Workstations resources seem to be a interesting case to rethink how we can restructure the parent code. Just to share some ideas (the code is not fully checked in) https://github.com/GoogleCloudPlatform/k8s-config-connector/commit/46fb55e80e2c023add729028cdd29e2fc4bf1644#diff-1ec84c91cc52a0147c0b6b7f0f6ea167bd371db248f87d3c69274873c29719ad

PR #3434, opened 2024-12-19
fix: Make ResourceID field "mutable"

jasonvigil:
In practice, due to in-controller enforcement, the only valid change is to set from empty string -> actual resource ID.

### DIFF
--- a/apis/bigquerydatatransfer/v1alpha1/config_types.go
+++ b/apis/bigquerydatatransfer/v1alpha1/config_types.go
@@ -83,6 +83,7 @@ type BigQueryDataTransferConfigSpec struct {
 	Parent `json:",inline"`
 
 	// The BigQueryDataTransferConfig name. If not given, the metadata.name will be used.
+	// Cannot be changed after create, except for updating from empty string to the actual resource ID.
 	ResourceID *string `json:"resourceID,omitempty"`
 
 	// Data transfer schedule.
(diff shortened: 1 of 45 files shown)


maqiuyujoyce (commented):
Could you also remove the `Immutable.` from the field description?

jasonvigil:
> Could you also remove the `Immutable.` from the field description?

I think it is sort of "correct" to include the immutable in the field description. The only valid change is from empty string -> actual resource id. Otherwise it is immutable.

jasonvigil:
> > Could you also remove the `Immutable.` from the field description?
> 
> I think it is sort of "correct" to include the immutable in the field description. The only valid change is from empty string -> actual resource id. Otherwise it is immutable.

Though, maybe since it is removed from the template, we should remove from existing also. Will do.

maqiuyujoyce:
I think "Immutable" is a very specific concept. Something like `Must be the same resource ID once successfully reconciled` may be helpful?

yuwenma (commented):
/lgtm
/approve

Thanks!

jasonvigil:
> I think "Immutable" is a very specific concept. Something like `Must be the same resource ID once successfully reconciled` may be helpful?

Agreed it will be helpful to add some comment to help users understand the validation behavior. Just added (for template and all resources)

PR #3435, opened 2024-12-19
tests: record fields that are never set

acpana:
We should check that all fields for a CRD are set in the fixtures at some point (either in create.yaml or update.yaml. Further, doing that we automate a fair amount of PR reviewing away from checking the CRD definition and how it's tested in the resourcefixtures

( This is actually a better incarnation of https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2326 )


tests/apichecks/crds_test.go
```diff
@@ -390,3 +397,151 @@ func TestCRDShortNames(t *testing.T) {
 	want := strings.Join(errs, "\n")
 	test.CompareGoldenFile(t, "testdata/exceptions/shortnames.txt", want)
 }
+
+// Run this test with WRITE_GOLDEN_OUTPUT set to update the expections list.
+// Note adding to the exceptions list should not be done trivially.
```
justinsb, line 402: If you're fixing the spelling mistake here, maybe "adding to the exceptions list should primarily be used for backwards compatibility" ?

tests/apichecks/crds_test.go
```diff
@@ -390,3 +397,151 @@ func TestCRDShortNames(t *testing.T) {
...
+			visitCRDVersion(version, func(field *CRDField) {
+				fieldPath := field.FieldPath
+
+				// Only consider fields under `spec`
+				if !strings.HasPrefix(fieldPath, ".spec.") {
+					return
+				}
+
+				// skip the resource id field
+				if strings.HasSuffix(fieldPath, "resourceID") {
```
justinsb, line 428: Nit: `.resourceID` (not that it should actually matter in practice)

tests/apichecks/crds_test.go
```diff
@@ -390,3 +397,151 @@ func TestCRDShortNames(t *testing.T) {
...
+					if len(field.props.Properties) > 0 || field.props.Type == "object" {
+						return
+					}
+					if field.props.Type == "array" && field.props.Items != nil {
+						return // Skip the array itself; focus on its elements
+					}
+				}
+
+				// Any XYZRef field was already handled and handling the children will just double count
+				if strings.Contains(fieldPath, "Ref") {
```
justinsb, line 465: What's an example of this?  Because I think you've already checked `strings.HasSuffix(fieldPath, "Ref") {` above...

tests/apichecks/crds_test.go
```diff
@@ -23,7 +23,14 @@ import (
 
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/crd/crdloader"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test"
+	testcontroller "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test/controller"
+	testgcp "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test/gcp"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test/resourcefixture"
+	testvariable "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test/resourcefixture/variable"
+	"github.com/ghodss/yaml"
```
justinsb, line 30: Please (always) use `sigs.k8s.io/yaml`.  I know there are some places we don't, but I consider those problematic

tests/apichecks/crds_test.go
```diff
@@ -390,3 +397,151 @@ func TestCRDShortNames(t *testing.T) {
...
+
+				// Any XYZRef field was already handled and handling the children will just double count
+				if strings.Contains(fieldPath, "Ref") {
+					return
+				}
+
+				// Check if field exists in any unstructured object
+				missing := true
+				for _, obj := range unstructs {
+					if hasField(obj.Object, fieldPath) {
```
justinsb, line 472: Oooh, can we have another linter that checks for fields we hallucinated :-)  (I.e. checks the other way round)  There might be a refactor where we introduce something like `CollectFields(obj *unstructured.Unstructured) sets[string]`

tests/apichecks/crds_test.go
justinsb: Hmmm, there are a lot of them, so maybe just drop the note :-)

justinsb:
Looks good, a few nits but then I think we should merge

tests/apichecks/crds_test.go
acpana: Oh interesting so the check is there to make sure that the fields in the list are defined in the CRD, right ?

tests/apichecks/crds_test.go
acpana: Yes agreed; I tried using that one first but I was having issues unmarshalling to yaml;

i can investigate why some more and can probably take it out in places like these too: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/cd58b1e916a87a22686fa361f0166833f0fec7ac/pkg/test/unstructured.go#L20

tests/apichecks/crds_test.go
acpana: Yeah for references in particular I decided to call it set if the field ending in Ref specifies either `external` or `name`.

For instance, when we visit `projectRef` we run the special handling to determine if either is set.

but then in our depth first traversal we also visit the fields children again and can end up in a situation where we call out that external is not set when name is:

```
        +       [missing_field] crd=alloydbbackups.alloydb.cnrm.cloud.google.com version=v1beta1: field ".spec.clusterNameRef.external" is not set in unstructured objects
```

tests/apichecks/crds_test.go
acpana: Ah figured it out 🙃 ; will create an issue and fix it elsewhere too (eventaully)

tests/apichecks/crds_test.go
justinsb: Ah gotcha!
```suggestion
// Any fooRef field was already handled and handling the children will just double count
// when we visit  fooRef.namespace, fooRef.name and fooRef.external etc
				if strings.Contains(fieldPath, "Ref.") {
```

PR #3438, opened 2024-12-20
Mark 'spec.deletionPolicy' mutable-but-unreadable in SQLDatabase

maqiuyujoyce:
### Change description

Fixes b/375600924

`spec.deletionPolicy` is a mutable field but it cannot be read from the GCP API because this is a customized TF field. Marking it `mutable-but-unreadable` ensures the live state contains the correct value of `spec.deletionPolicy` during reconciliation.

### Tests you have done

Dynamic test passed.


yuwenma:
/approve
/hold 

Thank you @maqiuyujoyce! Hold in case you want others to take a look. cc @jasonvigil

PR #3440, opened 2024-12-20
Flex Template Composition for Dataflow

svetakvsundhar:
This change adds a composition to deploy Dataflow Flex Templates. Note that this can be used to deploy Beam YAML pipelines. 

Tested on DataflowRunner. 

Screenshot 2024-12-20 at 3 00 18 PM

Fixes #

### Tests you have done


svetakvsundhar:
R: @barney-s 
R: https://github.com/Polber/ (for a Beam YAML review).


A note: While we can pass in variables to the Beam YAML template via the `parameters: --jinja_variables` flag, I don't think this actually makes sense in practice. The Beam YAML template will already be staged in a GCS bucket, and it'll be an infra team executing KCC, and thus they shouldn't have access to modify Beam YAML pipeline parameters.

svetakvsundhar:
cc: https://github.com/hm98765

barney-s:
@svetakvsundhar Please look at KRO and migrate this sample there.
KRO is the new version of compositions: https://github.com/kro-run/kro/tree/main/examples/gcp

PR #3446, opened 2024-12-23
doc: 1.127 release note

yuwenma:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/docs/releasenotes/release-1.127.md
+++ b/docs/releasenotes/release-1.127.md
@@ -1,35 +1,49 @@
-# v1.126.0
-
-** This version is not yet released; this document is gathering release notes for the future release **
-
-* ...
-
-* Special shout-outs to ... for their contributions to this release.
-TODO: list contributors with `git log v1.126.0... | grep Merge | grep from | awk '{print $6}' | cut -d '/' -f 1 | sort | uniq`
+* Special shout-outs to @acpana, @anhdle-sso, @cheftako, @ericpang777, @gemmahou, @haiyanmeng, @jasonvigil, @jingyih, @justinsb, @maqiuyujoyce, @nb-goog, @xiaoweim, @yuwenma, @ziyue-101 for their contributions to this release.
 
 ## Announcement 
 
+
 ## New Beta Resources (Direct Reconciler):
 
-* `WorkstationConfig`
-* `Workstation`
-* `FirestoreDatabase`
+* [`BigQueryAnalyticsHubListing`](https://cloud.google.com/config-connector/docs/reference/resource-docs/bigqueryanalyticshub/bigqueryanalyticshublisting)
+
+* [`FirestoreDatabase`](https://cloud.google.com/config-connector/docs/reference/resource-docs/firestore/firestoredatabase)
+
+* [`WorkstationConfig`](https://cloud.google.com/config-connector/docs/reference/resource-docs/workstations/workstationconfig)
+
+* [`Workstation`](https://cloud.google.com/config-connector/docs/reference/resource-docs/workstations/workstation)
 
 ## New Fields:
 
-* `GkeHubFeatureMembership`
-  * Added `spec.configmanagement.management` field to enable Config Sync Auto Upgrade.
+* `BigQueryDataTransferConfig`
 
-* `Placeholder`
+  * Added `spec.scheduleOptionsV2` to customize the different types of data transfer schedule.
+  * Added `status.observedState.error` with detailed information about reason of the latest config failure.
+
+* `GkeHubFeatureMembership`
+
+  * Added `spec.configmanagement.management` to enable Config Sync Auto Upgrade. This is an opt-in feature and you need to turn on the `alpha.cnrm.cloud.google.com/reconciler: direct` annotation on the object.
+
 
 ## Modified Beta Reconciliation
 
-We migrated the following reconciliation from the TF-based or DCL-based controller to the new Direct controller to enhance the reliability and performance. The resource CRD is unchanged.
+We added the direct controller support for the following 3 resources to enhance the reliability and performance. Add `alpha.cnrm.cloud.google.com/reconciler: direct` annotation on the object to use the direct controller. CRD is backward compatible.
 
-* `GkeHubFeatureMembership` is now a direct resource.
+* `GkeHubFeatureMembership`
 
-* `Placeholder`
+  * Added Config Sync Auto-Upgrade support.
 
-## New Alpha Resources (Direct Reconciler):
+* `SecretManagerSecret`
 
-* `Placeholder`
+  * [#510](https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/510) Enhanced `spec.rotation.nextRotationTime` to use a fixed datetime value to avoid relative `now()` friction. 
+  * [#1081](https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/1081) Fixed the `spec.replication.auto` immutable issue
+  * [#3051](https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/3051) Fixed the `spec.rotation.rotationPeriod` immutable issue 
+  * Added the in-use version aliases in `status.observedState.versionAliases`
+  * Resolved update stalling issues. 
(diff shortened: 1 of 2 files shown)


docs/releasenotes/release-1.127.md
```diff
@@ -1,35 +1,53 @@
...
 
-* Special shout-outs to ... for their contributions to this release.
-TODO: list contributors with `git log v1.126.0... | grep Merge | grep from | awk '{print $6}' | cut -d '/' -f 1 | sort | uniq`
+## New Beta Resources (Direct Reconciler):
 
-## Announcement 
+* [`BigQueryAnalyticsHubListing`](https://cloud.google.com/config-connector/docs/reference/resource-docs/bigqueryanalyticshub/bigqueryanalyticshublisting)
 
-## New Beta Resources (Direct Reconciler):
+  * Manage the reference to a shared dataset that a publisher lists in a data exchange.
```
justinsb, line 10: Nit: I think it's easier to just list the fields we added. The information here feels redundant vs the field docs.  Ideally we would link them, but that seems less important than being precise about what we are adding.

justinsb: Oh, unless there are entirely new resources?  I would still just link to the resource (as you are doing) and not add the text.  "Manage the reference" made me think we were adding a ref field

docs/releasenotes/release-1.127.md
```diff
@@ -1,35 +1,53 @@
...
 ## Modified Beta Reconciliation
 
 We migrated the following reconciliation from the TF-based or DCL-based controller to the new Direct controller to enhance the reliability and performance. The resource CRD is unchanged.
 
-* `GkeHubFeatureMembership` is now a direct resource.
 
-* `Placeholder`
+* `GkeHubFeatureMembership`
+
+  * Added `spec.configmanagement.management` to enable Config Sync Auto Upgrade.
```
justinsb, line 38: Personally I would add to "New Fields" and to "modified reconciler".  They are telling the user different things: the new field is saying "there's new functionality available".  The "Modified Beta Reconciliation" tells the user that if they see a weird regression, it might be because of this.

docs/releasenotes/release-1.127.md
yuwenma: Could you be more specific about `list the fields we add`? This adds an entire new resource (beta), not a single field. Do you want me to list all the fields the resource contains? I added the link to the resource reference page which lists all the fields.

justinsb: Sorry, I was confused.  I don't think we should add the text description, because it duplicates the docs.  Better (but not required) to link to the docs.

justinsb:
Looks good, I don't know if we want to add the field/type summaries, vs linking to the docs.  @cheftako might have a different take though.

docs/releasenotes/release-1.127.md
yuwenma: > Oh, unless there are entirely new resources? I would still just link to the resource (as you are doing) and not add the text. "Manage the reference" made me think we were adding a ref field

So adding the text to briefly explain what this resource is is from a discussion with the tech writer. We think it is a better form to swift the focus of the release note from a developer focus to a user focus. I can remove it here but may still want to keep it in the google release note

cheftako: Not sure I want to add all new fields for a new resource. That seems too verbose. I like the idea of linking the Google resource docs from the Google release notes. As far as I know we don't have OSS resource docs (Maybe we should) but till we do not sure that we have something we should link from these release notes.

docs/releasenotes/release-1.127.md
cheftako: I think the field should be changed. I would keep the reconciliation warning here. Might be worth being more explicit. Eg Fixed bug where false value was being ignored.

yuwenma: Thanks! Added the item in both places.

docs/releasenotes/release-1.127.md
```diff
@@ -1,35 +1,49 @@
...
+## Announcement 
 
-* ...
 
-* Special shout-outs to ... for their contributions to this release.
-TODO: list contributors with `git log v1.126.0... | grep Merge | grep from | awk '{print $6}' | cut -d '/' -f 1 | sort | uniq`
+## New Beta Resources (Direct Reconciler):
 
-## Announcement 
+* [`BigQueryAnalyticsHubListing`](https://cloud.google.com/config-connector/docs/reference/resource-docs/bigqueryanalyticshub/bigqueryanalyticshublisting)
```
acpana, line 8: Thanks for adding this 🙏🏼 I was going to open a PR to add it too 😆 

FWIW, 

/lgtm

docs/releasenotes/release-1.127.md
```diff
@@ -1,35 +1,49 @@
...
-* `Placeholder`
+  * Added `spec.configmanagement.management` to enable Config Sync Auto Upgrade. This is an opt-in feature and you need to turn on the `alpha.cnrm.cloud.google.com/reconciler: direct` annotation on the object.
+
 
 ## Modified Beta Reconciliation
 
-We migrated the following reconciliation from the TF-based or DCL-based controller to the new Direct controller to enhance the reliability and performance. The resource CRD is unchanged.
+We added the direct controller support for the following 3 resources to enhance the reliability and performance. Add `alpha.cnrm.cloud.google.com/reconciler: direct` annotation on the object to use the direct controller. CRD is backward compatible.
+
+* `GkeHubFeatureMembership`
```
maqiuyujoyce, line 32: GKEHubFeatureMembership is now a pure direct resource and doesn't need the direct annotation afaik.

docs/releasenotes/release-1.127.md
```diff
@@ -1,35 +1,49 @@
...
 
-* `GkeHubFeatureMembership` is now a direct resource.
+  * Added Config Sync Auto-Upgrade support.
```
maqiuyujoyce, line 34: Feel like duplicate information as we already mentioned it in the New Fields section.

docs/releasenotes/release-1.127.md
```diff
@@ -1,35 +1,49 @@
...
+* [`Workstation`](https://cloud.google.com/config-connector/docs/reference/resource-docs/workstations/workstation)
 
 ## New Fields:
 
+* `BigQueryDataTransferConfig`
+
+  * Added `spec.scheduleOptionsV2` to customize the different types of data transfer schedule.
+  * Added `status.observedState.error` with detailed information about reason of the latest config failure.
+
 * `GkeHubFeatureMembership`
```
maqiuyujoyce, line 23: GKEHubFeatureMembership

docs/releasenotes/release-1.127.md
```diff
@@ -1,35 +1,49 @@
...
-  * Added `spec.configmanagement.management` field to enable Config Sync Auto Upgrade.
 
-* `Placeholder`
+  * Added `spec.configmanagement.management` to enable Config Sync Auto Upgrade. This is an opt-in feature and you need to turn on the `alpha.cnrm.cloud.google.com/reconciler: direct` annotation on the object.
```
maqiuyujoyce, line 25: I think this doesn't need to be opt-in.

docs/releasenotes/release-1.128.md
```diff
@@ -0,0 +1,54 @@
...
+
+We have added support for direct reconciliation to more resources, with opt-in behaviour.  The API is unchanged. To use the direct reconciler, add the `alpha.cnrm.cloud.google.com/reconciler: direct` annotation to the corresponding Config Connector object.  The following resources now have direct reconciliation support (and we list some of the issues that this fixes): 
+
+* [`SAMPLE_SQLInstance`](https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/master/pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance)  
+
+    * Fix the upgrade and downgrade issue between ENTERPRISE and ENTERPRISE_PLUS.
+    * Supports "creating from clone" via `spec.cloneSource`
+
+
+## New features:
```
maqiuyujoyce, line 46: `Features` (capitalized `F`)?

PR #3451, opened 2024-12-30
chore: fix makefile

jingyih:
It seems that there are two duplicate "generate" rules in the makefile.


yuwenma (commented):
/lgtm
/approve

Good catch!

PR #3456, opened 2025-01-06
fix:shortnames: worstationcluster

acpana:
Following up on https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3407 adding shortnames to WorkstationCluster.


yuwenma (commented):
/lgtm
/approve

Thanks for the fix! Do you think it worths the effort to add a CRD checker to require the  existence of the shortnames? (not mandatory and seems to be nice to have as a warning)

acpana:
> Thanks for the fix! Do you think it worths the effort to add a CRD checker to require the existence of the shortnames? (not mandatory and seems to be nice to have as a warning)

I think we have that already in the form of the API linter:

https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3407

Now we are just patching what we missed pre linter.

PR #3458, opened 2025-01-07
chore: sync CRDs inside "/crds" folder

jingyih:
The public documentation on [installing alpha CRDs](https://cloud.google.com/config-connector/docs/how-to/install-alpha-crds) still references the legacy "/crds" folder. Let's ensure it stays up to date.

```
$ dev/tasks/sync-crds-folder.sh 1.127.0
```

### DIFF
--- a/dev/tasks/sync-crds-folder.sh
+++ b/dev/tasks/sync-crds-folder.sh
@@ -1,0 +1,64 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+VERSION=${1:-1.127.0}  # Default to 1.127.0 if no version provided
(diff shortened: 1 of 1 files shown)


yuwenma (commented):
/lgtm
/approve
/hold

Thanks for the change. I like the direction. There is actually a bigger issue we should fix to resolve the the alpha compatibility issue: The ./crds should be git-tag versioned together with the release version, otherwise users cannot apply a reliable ./crds

jingyih:
> Thanks for the change. I like the direction. There is actually a bigger issue we should fix to resolve the the alpha compatibility issue: The ./crds should be git-tag versioned together with the release version, otherwise users cannot apply a reliable ./crds

That makes sense. I was thinking we could run `dev/tasks/sync-crds-folder.sh ${VERSION}` as part of the release process.

yuwenma:
cc @xiaoweim

jingyih:
@xiaoweim
Do we still need to merge `dev/tasks/sync-crds-folder.sh` into the codebase, or have you already included it in the new release process?

/cc @yuwenma

yuwenma:
> Do we still need to merge `dev/tasks/sync-crds-folder.sh` into the codebase, or have you already included it in the new release process?
>


We can merge this script. But it would be called in an internal process *right after*we push the release git tag

xiaoweim:
> > Do we still need to merge `dev/tasks/sync-crds-folder.sh` into the codebase, or have you already included it in the new release process?
> 
> We can merge this script. But it would be called in an internal process _right after_we push the release git tag. And that will be automated

Yes I agree, we can merge it. But eventually it will be a step in an automated process. I think it would be better to incorporate it into the propose-tag step, so there will only be one PR before we push release tag @yuwenma .

jingyih:
PR rebased. Dropped the second commit that ran the script to update the CRDs.

jingyih:
Rebased again, to pick up the fix for the broken HEAD.

PR #3461, opened 2025-01-07
lint: add depguard

acpana:
Using `depguard` to highlight the imports we don't want to use. I will `//nolint` existing usages and this will prevent future imports.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -23,6 +23,7 @@ issues:
 linters:
   disable-all: true
   enable:
+    - depguard
     - errcheck
     - errorlint
     # - exportloopref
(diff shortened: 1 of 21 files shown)


.golangci.yaml
```diff
@@ -60,3 +61,9 @@ linters-settings:
       - shadow
       - testinggoroutine
       - unusedwrite
+  depguard:
+    rules:
+      main:
+        deny:
+          - pkg: "github.com/ghodss/yaml"
+            desc: Use "sigs.k8s.io/yaml"
```
yuwenma, line 69: Right! We should use kyaml instead. https://pkg.go.dev/sigs.k8s.io/kustomize/kyaml

yuwenma (commented):
/lgtm
/approve

Thanks! I'm curious how you notice this issue. 👍

PR #3465, opened 2025-01-08
tests: zero filler

acpana:
Adding a zero filler (similar to the random filler) to the KRMgen. We will use this in tooling to create tests for set 0 scenarios.


pkg/test/fuzz/krmgen.go
```diff
@@ -112,3 +116,86 @@ func (rf *RandomFiller) fillWithRandom(t *testing.T, fieldName string, field ref
...
+
+	case reflect.Float32, reflect.Float64:
+		field.SetFloat(0.0)
+
+	case reflect.String:
+		field.SetString("")
+
+	case reflect.Slice: // undefined for now
+
+		// count := rf.randStream.Intn(10) + 1
```
jingyih, line 168: Should we silently skip slice and map fields, or should we explicitly fail? Alternatively, could we create empty slices and maps to use as the "zero value"?

jingyih (commented):
/lgtm

Had one question but not a blocker for now.

pkg/test/fuzz/krmgen.go
acpana: We can revisit this decision but I think ignoring more complex types for now is okay. As I wire this into the dev/ generation tools we will iterate here.

yuwenma (commented):
/approve

PR #3467, opened 2025-01-10
Spanner instance add autoscale and edition field

anhdle-sso:
### Change description

Fixes #

### Tests you have done


apis/spanner/v1beta1/instance_types.go
```diff
@@ -45,6 +45,18 @@ type SpannerInstanceSpec struct {
...
+	// Optional. The autoscaling configuration. Autoscaling is enabled if this
+	// field is set. When autoscaling is enabled, node_count and processing_units
+	// are treated as OUTPUT_ONLY fields and reflect the current compute capacity
+	// allocated to the instance.
+	// +optional
+	AutoscalingConfig *AutoscalingConfig `json:"autoscalingConfig,omitempty"`
+
+	// Optional. The `Edition` of the current instance.
+	// +optional
+	// +kubebuilder:validation:Enum=EDITION_UNSPECIFIED;STANDARD;ENTERPRISE;ENTERPRISE_PLUS
```
yuwenma, line 57: One potential issue about using CRD enum validation is that if the list changes on the GCP service side, KCC will be  broken. One potential improvements is to check the enum in the controller ( proto) level via `CompareProtoMessage`. 

Defer to you

tests/apichecks/testdata/exceptions/missingfields.txt
```diff
@@ -3752,6 +3752,8 @@
 [missing_field] crd=spannerdatabases.spanner.cnrm.cloud.google.com version=v1beta1: field ".spec.enableDropProtection" is not set in unstructured objects
 [missing_field] crd=spannerdatabases.spanner.cnrm.cloud.google.com version=v1beta1: field ".spec.encryptionConfig.kmsKeyRef" is not set; neither 'external' nor 'name' are set
 [missing_field] crd=spannerdatabases.spanner.cnrm.cloud.google.com version=v1beta1: field ".spec.versionRetentionPeriod" is not set in unstructured objects
+[missing_field] crd=spannerinstances.spanner.cnrm.cloud.google.com version=v1beta1: field ".spec.autoscalingConfig.autoscalingLimits.maxProcessingUnits" is not set in unstructured objects
+[missing_field] crd=spannerinstances.spanner.cnrm.cloud.google.com version=v1beta1: field ".spec.autoscalingConfig.autoscalingLimits.minProcessingUnits" is not set in unstructured objects
```
yuwenma, line 3756: why do we not adding these two fields? It seems to be not complicated to do

scripts/github-actions/tests-e2e-samples
```diff
@@ -33,6 +33,6 @@ MOCK_GROUP_EMAIL="test@group.com"
 RUN_E2E=1 \
 E2E_KUBE_TARGET=envtest \
 E2E_GCP_TARGET=mock \
-KCC_USE_DIRECT_RECONCILERS="ComputeForwardingRule" \
+KCC_USE_DIRECT_RECONCILERS="ComputeForwardingRule,SpannerInstance" \
```
yuwenma, line 36: We no longer need to flip the KCC_USE_DIRECT_RECONCILERS. If the presubmit passes (fixture passes), we can  assume the direct approach for sample is good as well. For fixtures, if we want to support direct approach, we can use the direct alpha annotation.

pkg/test/resourcefixture/testdata/externalref/externalwithsubresource/_http.log
```diff
@@ -352,8 +357,9 @@ X-Xss-Protection: 0
 
 {
   "config": "projects/${projectId}/instanceConfigs/regional-us-west1",
-  "createTime": "2024-04-01T12:34:56.123456Z",
```
yuwenma, line 355: Any reason to drop this createTime field?

pkg/test/resourcefixture/testdata/basic/spanner/v1beta1/spannerinstance-direct-full/update.yaml
```diff
@@ -0,0 +1,33 @@
...
+kind: SpannerInstance
+metadata:
+  labels:
+    label-one: "value-one"
+  name: spannerinstance-sample-${uniqueId}
+  annotations:
+    alpha.cnrm.cloud.google.com/reconciler: "direct"
+spec:
+  config: regional-us-west1
+  displayName: New spanner Instance Sample
```
yuwenma, line 25: ah, suggest in the future change the update.yaml with wording "updated" not "new", because normally "new" shows up in create.yaml

pkg/test/resourcefixture/testdata/externalref/externalwithsubresource/_http.log
anhdle-sso: deleted in the mock by accident. I am adding it back

apis/spanner/v1beta1/instance_types.go
anhdle-sso: ack, the CompareProtoMessage might not work well in this resource because of weird behavior between numnode, processing unit and autoscaling. The edition is unlikely to change; but I keep that in mind when we have a standard to fully deal with enum

PR #3468, opened 2025-01-10
tests: add mock tests that use gcloud

justinsb:
This can enable testing of mockgcp before we have the full controller etc.

PR #3469, opened 2025-01-10
mockgcp: add support for storage bucket list

justinsb:
- **tests: add mock tests that use gcloud**
- **mockgcp: add support for storage bucket list**

### DIFF
--- a/mockgcp/mockgcptests/proxy.go
+++ b/mockgcp/mockgcptests/proxy.go
@@ -1,0 +1,106 @@
+package mockgcptests
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"strconv"
+	"strings"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp"
+	"k8s.io/klog/v2"
+)
+
+type Proxy struct {
+	httpClient *http.Client
+}
+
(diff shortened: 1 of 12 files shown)


mockgcp/mockgcptests/proxy.go
```diff
@@ -89,16 +94,22 @@ func (p *Proxy) BuildGcloudConfig(proxyEndpoint *net.TCPAddr, mockgcp mockgcp.In
 	config.AddConfig("proxy/port", strconv.Itoa(proxyEndpoint.Port))
 
 	// We need to register these over http, to stop gcloud trying to use TUNNEL with our proxy
-	services := mockgcp.ListServices()
+
+	// We need a hard-coded list, because we don't always mockgcp available
+	services := []string{
```
yuwenma, line 99: Since we have more and more hard-coded lists in the MockGCP, we may want to start moving them to its own file `mockgcp/const` dir or at least to the top of this file, otherwise the maintenance overhead will continue grow dramatically.

yuwenma (commented):
/approve

Reviewed the last comment. Looks good! 
Do you want to merge #3468 first ? Or just on this ?

mockgcp/mockgcptests/proxy.go
justinsb: I sent a PR to just add all the services.  #3655   cc @cheftako

PR #3470, opened 2025-01-10
experiment: a command line interactive AI chatbot for KCC coding

justinsb:
Can we create a chatbot that is better than "generic" chatbots,
for coding KCC?


yuwenma (commented):
/lgtm
/approve

This is just so advanced!

PR #3466, opened 2025-01-09
fix: more shortnames

acpana:
Even more shortname fixes!

PR #3472, opened 2025-01-10
fix: last shortnames missing

acpana:
Missed these two in the big one https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3466


justinsb (approved):

justinsb:
It might be interesting to see whether we can add shortnames using AI :-)  i.e. shortnames.txt is a list of problems, can we feed them into AI one at a time?  Can we ask gemini to formulate a plan and then execute each step of the plan (and is there any reason to do so vs just having gemini "one shot" it)

PR #3475, opened 2025-01-12
feat: add  cmd to merge two go files (struct only)

yuwenma:
This go file merge cannot handle comments well because it uses ast which does not treat comment as first citizen (comment is inserted by its relative position in the file so refactoring the go file place all comments in the wrong place)  

This PR drops the comments in the updated file. I have some plans to add back the comments, so I keep the comment Pos code in this PR.


justinsb:
Looks purely additive so let's get this in!

/approve
/lgtm

PR #3478, opened 2025-01-14
mockgcp: update with dev/tasks/update-mockgcp

justinsb:
- **mockgcp: update with dev/tasks/update-mockgcp**
- **autogen: update mockgcp**


justinsb:
/assign @yuwenma 

This is blocking some contributors for newer resources

PR #3479, opened 2025-01-14
mockgcp: tolerate deprecated field in swagger

justinsb:
We're parsing "strictly" (errors for unknown fields),
and we hadn't encountered this field before.


justinsb:
/assign @yuwenma 

This was blocking some contributors for newer resources (though the mockgcp update is the better solution)

yuwenma (commented):
/approve
/lgtm

PR #3488, opened 2025-01-14
chore: update golangci-lint

justinsb:
Update everything to 1.63.4 (because go versions must match).

exportloopref is not needed; that go-tcha has been fixed in go.

Fix a few error message lint issues.


acpana:
nice thanks for doing that 🙏🏼 

/lgtm
/approve

PR #3490, opened 2025-01-15
fix: Fix autopilot deploy targets

jasonvigil:
Previously, deployments (via `make deploy-operator`) to autopilot clusters failed with the following error:
```
...
horizontalpodautoscaler.autoscaling/cnrm-webhook created              
Error from server (GKE Warden constraints violations): error when creating "STDIN": admission webhook "warden-validating.common-webhooks.networking.gke.io" denied the request: GKE Warden rejected the request because it violates one or more constraints.                                                                                                      
Violations details: {"[denied by autogke-disallow-hostnamespaces]":["enabling hostNetwork is not allowed in Autopilot."],"[denied by autogke-no-host-port]":["container recorder specifies host ports [48797], which are disallowed in Autopilot."]}                                                                                                              
Requested by user: 'jasonvigil@google.com', groups: 'system:authenticated'.
make: *** [Makefile:228: deploy-controller] Error 1
```

This PR:
1) Fixes the makefile targets so they use the correct patch directories for autopilot clusters
2) Includes a patch to remove the host port from the `cnrm-resource-stats-recorder` deployment spec. This patch was defined in the "config connector operator" manifests ([here](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/08b5ce7f6b1e1506b365be1ef57e729762237392/operator/config/autopilot-channels/cluster-workload-identity/recorder_remove_hostport_patch.yaml#L1)), but was omitted from the non-"config-connector operator" manifests (in `./config`).


jasonvigil:
CC @xiaoweim @justinsb is there a cleaner way to do this? Seems like duplicating the same patch in both `config/` and `operator/config` might not be necessary, but I am not sure.

yuwenma:
This PR is more than deploying the master code to your own autopilot cluster. It changes (probably fixes) the releases as well. Can we verify if the change works well on a non google (not influenced by overground) cluster

jasonvigil:
> This PR is more than deploying the master code to your own autopilot cluster. It changes (probably fixes) the releases as well. Can we verify if the change works well on a non google (not influenced by overground) cluster

I tested this change by installing KCC on a GKE autopilot cluster in an external org.

yuwenma:
/approve
/assign @xiaoweim 

Could you review this PR?

config/installbundle/releases/scopes/cluster/autopilot-withworkloadidentity/kustomization.yaml
```diff
@@ -17,4 +17,11 @@ resources:
   - ../withworkloadidentity
 patches:
   - path: hostnetwork_patch.yaml
-
+patchesJson6902:
```
yuwenma, line 20: Why do we have two separate patch files, can we add this to the existing hostnetwork_patch.yaml instead? Maybe a question to @xiaoweim

jasonvigil: @yuwenma -- I copied the new patch files from https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/08b5ce7f6b1e1506b365be1ef57e729762237392/operator/config/autopilot-channels. Now that we are making these patches in `config/installbundle/*`, I think we can potentially remove the patches from `operator/config`, because they will already get applied in `dev/tasks/propose-tag` by `dev/tasks/build-release-bundle`. Will update this PR

jasonvigil: Ok, updated to remove the (now unnecessary?) patches from `operator/config`. However, I am not sure how to combine the two patches into one for each of the `config/installbundle/releases/scopes` autopilot directores. The hostnetwork_patch is setting the `hostnetwork` field to false, while the `recorder_remove_hostport_patch` is removing the `hostPort` field completely. Is is even possible to combine the two different styles of `patches` and `patchesJson6902`?

xiaoweim: I think it might be possible to combine the two patches into one if we can convert the JSON patch to a merge patch for the `record_remove_hostport_patch.yaml`. Something like this:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: cnrm-resource-stats-recorder
  namespace: cnrm-system
spec:
  template:
    spec:
      hostNetwork: false
      containers:
      - name: 0
        ports:
        - hostPort: null
```

I am not sure if it will have the exact same effect, but it is not a blocker.

PR #3481, opened 2025-01-14
codebot: create a simple capture-the-flag test

justinsb:
- **experiment: a command line interactive AI chatbot for KCC coding**
- **codebot: create a simple capture-the-flag test**

PR #3496, opened 2025-01-17
codebot: add a simple UI with tview

justinsb:
- **codebot: create a simple capture-the-flag test**
- **codebot: add a simple UI with tview**


justinsb:
Builds on #3481 .  Adds a nicer UI, but what I'm really going for here with the nicer UI is auto-complete, so we can reference files or protos more easily

yuwenma (commented):
/lgtm
/approve

PR #3501, opened 2025-01-17
codebot: support more LLMs (e.g. gemma)

justinsb:
Helps us set up a nice abstraction, and useful for benchmarking.

### DIFF
--- a/dev/tools/controllerbuilder/go.mod
+++ b/dev/tools/controllerbuilder/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/GoogleCloudPlatform/k8s-config-connector v1.125.0
 	github.com/fatih/color v1.17.0
 	github.com/gdamore/tcell/v2 v2.7.1
+	github.com/google/generative-ai-go v0.19.0
 	github.com/rivo/tview v0.0.0-20241227133733-17b7edb88c57
 	github.com/spf13/cobra v1.8.0
 	github.com/spf13/pflag v1.0.5
@@ -24,6 +25,7 @@ require (
 
 require (
 	cloud.google.com/go v0.116.0 // indirect
+	cloud.google.com/go/ai v0.8.0 // indirect
 	cloud.google.com/go/aiplatform v1.68.0 // indirect
 	cloud.google.com/go/auth v0.9.9 // indirect
 	cloud.google.com/go/auth/oauth2adapt v0.2.4 // indirect
@@ -37,6 +39,7 @@ require (
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
 	github.com/google/s2a-go v0.1.8 // indirect
+	github.com/google/uuid v1.6.0 // indirect
 	github.com/googleapis/enterprise-certificate-proxy v0.3.4 // indirect
 	github.com/googleapis/gax-go/v2 v2.13.0 // indirect
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
--- a/dev/tools/controllerbuilder/pkg/codebot/chat.go
+++ b/dev/tools/controllerbuilder/pkg/codebot/chat.go
@@ -20,15 +20,14 @@ import (
 	"fmt"
 	"strings"
 
-	"cloud.google.com/go/vertexai/genai"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/dev/tools/controllerbuilder/pkg/codebot/ui"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/dev/tools/controllerbuilder/pkg/llm"
 	"k8s.io/klog/v2"
 )
 
 type Chat struct {
-	client  *genai.Client
-	session *genai.ChatSession
+	client  llm.Client
+	session llm.Chat
 	baseDir string
 	ui      ui.UI
 }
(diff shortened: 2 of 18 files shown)


dev/tools/controllerbuilder/pkg/llm/gemini.go
```diff
@@ -15,55 +15,223 @@
...
+func BuildGeminiClient(ctx context.Context) (Client, error) {
 	var opts []option.ClientOption
 
 	if s := os.Getenv("GEMINI_API_KEY"); s != "" {
 		opts = append(opts, option.WithAPIKey(s))
-	} else {
-		// Some account can not use GEMINI_API_KEY but requires stricter access control via OAuth.
-		creds, err := google.FindDefaultCredentials(ctx, "https://www.googleapis.com/auth/generative-language", "https://www.googleapis.com/auth/cloud-platform")
+	}
+	//  else {
```
yuwenma, line 34: Any reasons to comment out this?

dev/tools/controllerbuilder/pkg/llm/ollama.go
```diff
@@ -0,0 +1,296 @@
...
+
+	// options: additional model parameters listed in the documentation for the Modelfile such as temperature
+
+	// stream: if false the response will be returned as a single response object, rather than a stream of objects
+	Stream *bool `json:"stream,omitempty"`
+
+	// keep_alive: controls how long the model will stay loaded into memory following the request (default: 5m)
+}
+
+type chatResposne struct {
```
yuwenma, line 102: typo: chatResponse

dev/tools/controllerbuilder/pkg/llm/ollama.go
```diff
@@ -0,0 +1,296 @@
...
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+
+	"k8s.io/klog/v2"
+)
+
+// BuildOllamaClient builds a client for the Ollama API.
+func BuildOllamaClient(ctx context.Context) (*OllamaClient, error) {
```
yuwenma, line 31: What different features ollama gives us compared to vertextAI?

dev/tools/controllerbuilder/go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	cloud.google.com/go/vertexai v0.13.2
 	github.com/GoogleCloudPlatform/k8s-config-connector v1.125.0
 	github.com/fatih/color v1.17.0
+	github.com/google/generative-ai-go v0.19.0
```
yuwenma, line 11: I'm kind of getting lost. Are we recommended to use vertextAI?

dev/tools/controllerbuilder/pkg/llm/gemini.go
justinsb: So my understanding, and TBH I'm confused, is that gemini uses API Keys and not GCP Auth, and vertexai uses GCP Auth (which _can_ be api keys, but that is very rare).  So I removed the unexpected paths in both gemini and vertexai.  (And I moved from commenting out to deletion)

dev/tools/controllerbuilder/pkg/llm/ollama.go
justinsb: Thanks - fixed!

dev/tools/controllerbuilder/pkg/llm/ollama.go
justinsb: So ollama lets us easily use self-hosted models like gemma.  gemma doesn't currently support function calling (AFAICT) so we're actually somewhat limited in the models we can use here, but it lets us benchmark more models.  Also, as a library, we want to be able to support more models and ollama gets us a lot!

dev/tools/controllerbuilder/go.mod
justinsb: So we can now evaluate vertexai vs gemini, figure out which one is best.  I would not imagine a significant difference, but we can also evaluate gemma where I would imagine we would see significant underperformance because the gemma model is so much smaller.

The big thing we're getting here is the ability to switch between models, as part of evaluating/tracking our progress.

dev/tools/controllerbuilder/pkg/llm/gemini.go
```diff
@@ -15,55 +15,215 @@
...
+	chat := model.StartChat()
+
+	return &GeminiChat{
+		model: model,
+		chat:  chat,
+	}
+}
+
+type GeminiChat struct {
+	model *genai.GenerativeModel
```
yuwenma, line 87: But you cannot change the model after initiaizing the Chat. Wondering if we need to keep both attributes in this class?

dev/tools/controllerbuilder/pkg/llm/gemini.go
```diff
@@ -15,55 +15,215 @@
...
+	chat  *genai.ChatSession
```
yuwenma, line 88: If this is for the functionDefinition, it also needs be defined before the chat starts IIRC.

dev/tools/controllerbuilder/pkg/codebot/chat.go
```diff
@@ -193,75 +136,80 @@ func (c *Chat) Close() error {
...
-					Response: response,
-				})
-
-			} else {
-				klog.Infof("UNKNOWN: %T %+v", part, part)
 			}
 		}
 
-		if len(functionResponses) != 0 {
-			c.session.History = append(c.session.History, &genai.Content{
```
yuwenma, line 258: +1. We don't need to record the history

dev/tools/controllerbuilder/pkg/llm/gemini.go
yuwenma: Even if the vertextAI doc says it supports API keys, my test shows the authentication is rejected in the gcp server level saying oauth2 is required. So the GCP oauth2 is the only auth that works for vertextai (and it's pretty nit).

justinsb: OK, well good thing it's deleted then :-)

dev/tools/controllerbuilder/pkg/llm/gemini.go
justinsb: So it might not be the best API, but we can (and do) change the model... we call SetFunctionDefinitions

dev/tools/controllerbuilder/pkg/llm/gemini.go
justinsb: It looks like we can change functions right up until we send the first chat message

dev/tools/controllerbuilder/pkg/codebot/chat.go
justinsb: Right, some of the client libs do this for us

justinsb:
Rebased, but lost LGTM

yuwenma:
/lgtm
/approve

Great! (my script is out of date now)

PR #3508, opened 2025-01-20
codebot: add a benchmark to evaluate the prediction job about API generation

yuwenma:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/dev/tools/controllerbuilder/cmd/benchmark/type/main.go
+++ b/dev/tools/controllerbuilder/cmd/benchmark/type/main.go
@@ -1,0 +1,117 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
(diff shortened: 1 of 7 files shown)


dev/tools/controllerbuilder/cmd/benchmark/type/main.go
```diff
@@ -0,0 +1,117 @@
...
+
+		// Get ground truth
+		baseUrl, err := url.Parse("https://raw.githubusercontent.com/GoogleCloudPlatform/k8s-config-connector/refs/heads/master/apis/")
+		if err != nil {
+			return fmt.Errorf("url parse: %w", err)
+		}
+		rawUrl := baseUrl.ResolveReference(&url.URL{Path: resource.RawFile})
+		fmt.Println("URL:", rawUrl.String())
+
+		groundTruth, err := kccio.GetFromRemote(ctx, rawUrl.String())
```
justinsb, line 95: Any reason not to read from the repo?  Not a big deal, just wondering why we're taking a dependency on github

dev/tools/controllerbuilder/pkg/io/cache.go
```diff
@@ -17,22 +17,38 @@ package io
...
+	fi, err := os.Stat(dest)
+	if err != nil {
+		return "", err
+	}
+	if fi.Mode().IsRegular() {
+		return dest, os.WriteFile(dest, []byte(out), 0644)
+	}
+
+	// If dest is dir, then write to dir
+	if _, err := os.Stat(dest); os.IsNotExist(err) {
```
justinsb, line 41: I think you can just call `os.MkdirAll`, no need to check if dir exists

justinsb (approved):

dev/tools/controllerbuilder/pkg/io/cache.go
yuwenma: Good catch! Thanks

PR #3513, opened 2025-01-21
chore: replace slugify dependency

justinsb:
We should own this behaviour, so we can guarantee compatability.


yuwenma (commented):
/lgtm
/approve

... cool. How did you hit this?

justinsb:
> How did you hit this?

One of the dependabots was failing to create a PR to update it, so I bumped the version (using gemini), and then looked at whether we should be using it at all

PR #2297, opened 2024-07-16
mockgcp: support for ApigeeOrganization

justinsb:
- **mockgcp: update service enablement APIs for fidelity with GCP**
- **tests: update golden output for service tests**
- **mockgcp: updates for generating apigee**
- **autogen: generated proto code for apigee**
- **tests: fix apigee organization sample**
- **mockgcp: support for ApigeeOrganization**
- **tests: default TEST_BILLING_ACCOUNT_ID in record-gcp**
- **chore: add golden data for apigeeorganization test**

PR #3517, opened 2025-01-21
mockgcp: apigeeorganization

acpana:
Mostly rips #2297 so I can use it in `ApigeeEnvironment`s so I can use it in `ApigeeEnvgroupAttachment`s

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/apigee/v1beta1/apigeeorganization/create.yaml
+++ b/pkg/test/resourcefixture/testdata/basic/apigee/v1beta1/apigeeorganization/create.yaml
@@ -28,11 +28,11 @@ spec:
   authorizedNetworkRef:
     name: "default"
   runtimeType: "CLOUD"
-  addonsConfig:
-    advancedApiOpsConfig:
-      enabled: true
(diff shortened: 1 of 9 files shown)


pkg/test/resourcefixture/testdata/basic/apigee/v1beta1/apigeeorganization/create.yaml
```diff
@@ -28,11 +28,11 @@ spec:
...
     name: "default"
   runtimeType: "CLOUD"
-  addonsConfig:
-    advancedApiOpsConfig:
-      enabled: true
-    integrationConfig:
-      enabled: false
-    monetizationConfig:
-      enabled: false
+  # addonsConfig:
```
acpana, line 31: Setting addons seems to be nondeterministic (my experience) for this DCL resource. I'd say we convert it to direct then we won't have an issue stemming from the underlying DCL. Till then I commented these out.

jasonvigil: Awesome, thank you @acpana !

PR #3538, opened 2025-01-26
controllerbuilder: fix to use proto field name

justinsb:
We were mistakenly using the KRM field name


jingyih (commented):
/lgtm

It seems there may be additional places that need fixing, e.g. line 304 and line 608?

yuwenma (commented):
/approve

Thanks!

PR #3540, opened 2025-01-27
controllerbuilder: strict mode to catch inconsistent training examples

justinsb:
- **mockgcp: clean up the tool annotations**
- **controllerbuilder: strict mode to catch inconsistent training examples**
- **controllerbuilder: precompute proto service definitions**


yuwenma (commented):
/lgtm
/approve

It only about filtering out the right examples, but not to pass in more accurate prompt message to Gemini, right? I tried using wrong/typo/unaligned marker message, and I don't see a difference in the response.

PR #3543, opened 2025-01-28
Fix regionalcomputeforwardingrule-direct integration test

gemmahou:
### Change description

Fix TestCreateNoChangeUpdateDelete/compute/basic-regionalcomputeforwardingrule-direct
Error message: 
```
googleapi: Error 400: Invalid value for field 'resource.IPAddress': ''. A reserved IP should be specified for forwarding rule with target type TARGET_VPN_GATEWAY
```

### Tests you have done

 go test -v -tags=integration ./pkg/controller/dynamic/ -test.run TestCreateNoChangeUpdateDelete -run-tests regionalcomputeforwardingrule-direct  -timeout 900s

--- PASS: TestCreateNoChangeUpdateDelete (0.17s)
    --- PASS: TestCreateNoChangeUpdateDelete/compute (0.00s)
        --- PASS: TestCreateNoChangeUpdateDelete/compute/basic-regionalcomputeforwardingrule-direct (160.30s)
PASS


maqiuyujoyce:
/lgtm
/approve

Regarding this test case, can other fields be updated as well? Looks like currently only the labels are updated, which doesn't seem to make the best use of the test.

gemmahou:
> /lgtm /approve
> 
> Regarding this test case, can other fields be updated as well? Looks like currently only the labels are updated, which doesn't seem to make the best use of the test.

The target VPNGateway is also undateable, updated the test yaml.

PR #3554, opened 2025-01-29
feat: Add Apigee instance controller

jasonvigil:
This PR adds the mappers, controller, and records GCP interaction for ApigeeInstance. MockGCP tests will come in a separate PR.

### DIFF
--- a/pkg/controller/direct/apigee/instance_controller.go
+++ b/pkg/controller/direct/apigee/instance_controller.go
@@ -1,0 +1,261 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package apigee
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+	"strings"
+
+	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/apigee/v1alpha1"
+	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1alpha1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/config"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
+
+	api "google.golang.org/api/apigee/v1"
+	"google.golang.org/protobuf/types/known/fieldmaskpb"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func init() {
+	registry.RegisterModel(krm.ApigeeInstanceGVK, NewApigeeInstanceModel)
+}
+
+func NewApigeeInstanceModel(ctx context.Context, config *config.ControllerConfig) (directbase.Model, error) {
+	return &modelApigeeInstance{config: config}, nil
+}
+
+var _ directbase.Model = &modelApigeeInstance{}
+
+type modelApigeeInstance struct {
+	config *config.ControllerConfig
+}
+
+func (m *modelApigeeInstance) AdapterForObject(ctx context.Context, reader client.Reader, u *unstructured.Unstructured) (directbase.Adapter, error) {
+	obj := &krm.ApigeeInstance{}
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &obj); err != nil {
+		return nil, fmt.Errorf("error converting to %T: %w", obj, err)
+	}
+
+	id, err := krm.NewApigeeInstanceIdentity(ctx, reader, obj)
+	if err != nil {
+		return nil, err
+	}
+
+	// Get apigee GCP client
+	gcpClient, err := newGCPClient(ctx, m.config)
+	if err != nil {
+		return nil, err
+	}
+	return &ApigeeInstanceAdapter{
+		id:               id,
+		k8sClient:        reader,
+		instancesClient:  gcpClient.instancesClient(),
+		operationsClient: gcpClient.operationsClient(),
+		desired:          obj,
+	}, nil
+}
+
+func (m *modelApigeeInstance) AdapterForURL(ctx context.Context, url string) (directbase.Adapter, error) {
+	// TODO: Support URLs
+	return nil, nil
+}
+
+type ApigeeInstanceAdapter struct {
+	id               *krm.InstanceIdentity
+	k8sClient        client.Reader
+	instancesClient  *api.OrganizationsInstancesService
+	operationsClient *api.OrganizationsOperationsService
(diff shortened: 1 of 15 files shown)


pkg/controller/direct/apigee/instance_controller.go
```diff
@@ -0,0 +1,264 @@
...
+	log.V(2).Info("updating ApigeeInstance", "name", a.id)
+	mapCtx := &direct.MapContext{}
+	updateMask := fieldmaskpb.FieldMask{}
+
+	resource := ApigeeInstanceSpec_ToAPI(mapCtx, &a.desired.DeepCopy().Spec)
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+
+	if resource.AccessLoggingConfig != nil && !reflect.DeepEqual(resource.AccessLoggingConfig, a.actual.AccessLoggingConfig) {
```
yuwenma, line 162: Can we use this function to sort and the compare? https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/apigee/envgroup_controller.go#L278

pkg/controller/direct/apigee/instance_controller.go
```diff
@@ -0,0 +1,264 @@
...
+
+// Delete the resource from GCP service when the corresponding Config Connector resource is deleted.
+func (a *ApigeeInstanceAdapter) Delete(ctx context.Context, deleteOp *directbase.DeleteOperation) (bool, error) {
+	log := klog.FromContext(ctx)
+	log.V(2).Info("deleting ApigeeInstance", "name", a.id)
+
+	op, err := a.instancesClient.Delete(a.id.String()).Context(ctx).Do()
+	if err != nil {
+		if direct.IsNotFound(err) {
+			// Return success if not found (assume it was already deleted)
```
yuwenma, line 236: Suggest adding a log?

pkg/controller/direct/apigee/instance_controller.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+	log.V(2).Info("successfully deleted ApigeeInstance", "name", a.id)
+
+	if err := a.waitForOp(ctx, op); err != nil {
+		return false, fmt.Errorf("waiting delete ApigeeInstance %s: %w", a.id, err)
+	}
+	return true, nil
+}
+
+func (a *ApigeeInstanceAdapter) waitForOp(ctx context.Context, op *api.GoogleLongrunningOperation) error {
```
yuwenma, line 249: This can be an individual function. And suggest moving it to the controller/direct/common lib for sharing purpose.

pkg/controller/direct/apigee/instance_mappings.go
```diff
@@ -0,0 +1,142 @@
...
+}
+func ApigeeInstanceSpec_ToAPI(mapCtx *direct.MapContext, in *krm.ApigeeInstanceSpec) *pb.GoogleCloudApigeeV1Instance {
+	if in == nil {
+		return nil
+	}
+	out := &pb.GoogleCloudApigeeV1Instance{}
+	out.AccessLoggingConfig = AccessLoggingConfig_ToAPI(mapCtx, in.AccessLoggingConfig)
+	out.ConsumerAcceptList = in.ConsumerAcceptList
+	out.Description = direct.ValueOf(in.Description)
+	// MISSING: DiskEncryptionKeyName
```
yuwenma, line 135: Just to double check, do you have a NOTYET comment for this field?

dev/tools/controllerbuilder/generate.sh
```diff
@@ -209,5 +209,11 @@ go run . generate-types \
     --api-version apigee.cnrm.cloud.google.com/v1alpha1 \
     --resource ApigeeInstance:GoogleCloudApigeeV1Instance
 
+go run . generate-controller \
+    --service mockgcp.cloud.apigee.v1 \
+    --api-version apigee.cnrm.cloud.google.com/v1alpha1 \
+    --kind ApigeeInstance \
+    --proto-resource GoogleCloudApigeeV1Instance
+
```
yuwenma, line 217: We don't need to update this file. It is just an example (or is it a requirement now?)

yuwenma (commented):
/approve
/assign @jingyih 

A few nits but overall looks great on my side. Can I have @jingyih to take another pass?

pkg/controller/direct/apigee/instance_controller.go
```diff
@@ -0,0 +1,264 @@
...
+	for {
```
jingyih, line 250: We can probably create a helper function which
 - respect the ctx timeout
 - take a input timeout
 
Something like this (AI generated)
 ```
 func waitUntil(ctx context.Context, timeout time.Duration, condition func() bool) error {
    ctx, cancel := context.WithTimeout(context.Background(), timeout)
    defer cancel()

    ticker := time.NewTicker(100 * time.Millisecond)
    defer ticker.Stop()

    for {
        select {
        case <-ctx.Done():
            if err := ctx.Err(); err == context.DeadlineExceeded {
                return fmt.Errorf("timeout exceeded")
            }
            return ctx.Err()
        case <-ticker.C:
            if condition() {
                return nil
            }
        }
    }
}
```

File: dev/tools/controllerbuilder/generate.sh (hunk 5 above)
Comment (jingyih) [reply dev/tools/controllerbuilder/generate.sh:217]: Yeah let's not add generate-controller command.

File: pkg/controller/direct/apigee/instance_mappings.go [hunk 7]
```diff
@@ -0,0 +1,142 @@
...
+	out.Description = direct.LazyPtr(in.Description)
+	// MISSING: DiskEncryptionKeyName
+	out.DisplayName = direct.LazyPtr(in.DisplayName)
+	out.IPRange = direct.LazyPtr(in.IpRange)
+	out.Location = direct.LazyPtr(in.Location)
+	// MISSING: Name
+	out.PeeringCIDRRange = direct.LazyPtr(in.PeeringCidrRange)
+	return out
+}
+func ApigeeInstanceSpec_ToAPI(mapCtx *direct.MapContext, in *krm.ApigeeInstanceSpec) *pb.GoogleCloudApigeeV1Instance {
```
jingyih, line 127: where is the mapping function for `diskEncryptionKMSCryptoKeyRef`?

pkg/controller/direct/apigee/instance_controller.go
jasonvigil: Is this a new idea / pattern to add a log here? Should we update the template / existing controllers for this?

pkg/controller/direct/apigee/instance_controller.go
jasonvigil: The waitForOp respects the context also, no? In the `operationsClient.Get(...).Context(ctx).Do()`

pkg/controller/direct/apigee/instance_mappings.go
jasonvigil: Ah, not implemented yet. Will add it.

pkg/controller/direct/apigee/instance_mappings.go
jasonvigil: The mapper for this needs to be implemented. Will fix.

pkg/controller/direct/apigee/instance_controller.go
jingyih: Yeah, the current implementation of `waitForOp` does respect the context's timeout and cancellation. If we were to create a helper function that accepts an arbitrary input function (not just `"a.operationsClient.Get"`), that function may or may not respect the context. Not saying we need to create one—just an idea.

jasonvigil: I ended up going with a variant of this idea. Thanks for the recommendation.

pkg/controller/direct/apigee/instance_mappings.go
jasonvigil: I added this mapper, good catch

jingyih (commented):
/lgtm

pkg/controller/direct/apigee/instance_controller.go
jasonvigil: I have created a PR to address this here: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3656

PR #3443, opened 2024-12-21
Clean up 'Direct' usage in resource config

maqiuyujoyce:
### Change description

* Removed the dummy resource configs.
* Added the test case to ensure `Direct` field is configured for migration only.
* Stopped depending on `resourceConfig` when possible.
  * Updated the reference doc generation logic so that it works for direct resources without resource config.
* Stopped using `resourceConfig.Direct` when it can be decided via function supportedgvks.IsDirectByGVK().
  * Updated the webhook.
* Added clarification comments to the usage of `resourceConfig.Direct`.
* Removed direct only resources from autogen allowlist.

### Tests you have done

- [X] Run `make ready-pr` to ensure this PR is ready for review.
- [X] Perform necessary E2E testing for changed resources.
   See go/cnrm-r/71860.

PR #3567, opened 2025-01-30
Remove the step to add dummy resource config

maqiuyujoyce:
This is to match the change in #3443.


yuwenma (commented):
/lgtm
/approve

Lovely!

PR #3577, opened 2025-01-30
codebot: create examples for generating mockgcp

justinsb:


jingyih (commented):
/lgtm

The demo was super cool. Thank you!

justinsb:
I introduced a "top level" doc and substitution for the first command... we can obviously iterate.

yuwenma:
> I introduced a "top level" doc and substitution for the first command... we can obviously iterate.

It may be hard to discover mockgcp/generating-a-mockgcp-resource.md from the other files when PR gets merged. But I don't see any blocker reasons.

PR #3521, opened 2025-01-22
Fix format validation for SecretManagerSecretRef and SecretManagerSecretVersionRef

jingyih:
Fix error
```
{"severity":"error","timestamp":"2025-01-22T07:00:49.994Z","msg":"Reconciler error","controller":"dataformrepository-controller","controllerGroup":"dataform.cnrm.cloud.google.com","controllerKind":"DataformRepository","DataformRepository":{"name":"dataformrepository-2tmq7a3ku34xkma","namespace":"2tmq7a3ku34xkma"},"namespace":"2tmq7a3ku34xkma","name":"dataformrepository-2tmq7a3ku34xkma","reconcileID":"645ec9cc-8246-42bb-ad57-b3ded07f6cd1","error":"Update call failed: failed to resolve resource references format of secretManagerSecretVersionRef external=\"projects/518915279/secrets/secretmanagersecret-2tmq7a3ku34xkma/versions/1\" was not known (use projects/<projectId>/secrets/<secretID>/versions/<versionID>)"}
```

PR #3578, opened 2025-01-30
chore: fix format validation for MonitoringAlertPolicyRef

jingyih:
Similar bug to #3521.


yuwenma:
/lgtm
/approve

Thanks for fixing this, @jingyih! The entire file can be moved to its resource controller dir.

Issue #3050, opened 2024-10-31
Empty status events after RedisCluster is created.


maqiuyujoyce:
### Checklist

- [X] I did not find a related open enhancement request.
- [X] I understand that enhancement requests filed in the GitHub repository are by default low priority.
- [X] If this request is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Describe the feature or resource

After RedisCluster was created, the status events kept to be `<none>` for 10+ minutes without providing any information to the users.

The expected behavior should be status events show that the resource is during the event `Updating` right after the controller starts reconcile the resource.

gemmahou:
This can be a more generic issue not limited to RedisCluster. 

With direct reconciliation, we currently only record `UpToDate` and `UpdateFailed` events. We should also consider recording `Updating` event, especially for long running resources like RedisCluster, to indicate that reconciliation is running.

strimpak:
I am seeing the same with `RedisCluster` while with other resources, like `RedisInstance` I don't face such issues

ldanielmadariaga:
This item has been open for >12 months. As part of scheduled repository maintenance, this Issue/PR will be closed. If this is still a critical ask, please re-open and work with your Google Cloud Account team to prioritize it with GCP KCC Product Management.

PR #3582, opened 2025-01-31
test: Fix regionalcomputetargettcpproxy-direct integration test

gemmahou:
### Change description

Fixes TestCreateNoChangeUpdateDelete/compute/basic-regionalcomputetargettcpproxy-direct integration test
Error message: 
```
reconcile.go:191: error was not considered transient; chain is [[*fmt.wrapError: Update call failed: error updating: update operation not supported for resource compute.cnrm.cloud.google.com/v1beta1, Kind=ComputeTargetTCPProxy v376okdfsjnwkki/computetargettcpproxy-v376okdfsjnwkki] [*fmt.wrapError: error updating: update operation not supported for resource compute.cnrm.cloud.google.com/v1beta1, Kind=ComputeTargetTCPProxy v376okdfsjnwkki/computetargettcpproxy-v376okdfsjnwkki] [*errors.errorString: update operation not supported for resource compute.cnrm.cloud.google.com/v1beta1, Kind=ComputeTargetTCPProxy v376okdfsjnwkki/computetargettcpproxy-v376okdfsjnwkki]]
```

Update operation is not supported for regional TCP Proxy. Usually we use an immutable webhook or CEL rule to prevent updates to fields. However, global TCP Proxy does support Update operation, so we cannot mark fields as immutable. 

The previous fix (#3252) returned an error in the Update function, caused reconciliation to fail even there's no update. To address this, we should only return an error when changes are detected.

### Tests you have done

Inspect the resource after trying to update a field:
```
Status:
  Conditions:
    Last Transition Time:  2025-02-01T00:46:06Z
    Message:               Update call failed: error updating: Update operation not supported for regional ComputeTargetTCPProxy
    Reason:                UpdateFailed
    Status:                False
    Type:                  Ready
  Creation Timestamp:      2025-01-31T16:45:46.445-08:00
  External Ref:            projects/cnrm-yuhou/regions/europe-west4/targetTcpProxies/computetargettcpproxy-yuhou0131
  Observed Generation:     2
  Proxy Id:                5132983695651688000
  Self Link:               https://www.googleapis.com/compute/v1/projects/cnrm-yuhou/regions/europe-west4/targetTcpProxies/computetargettcpproxy-yuhou0131
Events:
  Type     Reason        Age               From                              Message
  ----     ------        ----              ----                              -------
  Normal   UpToDate      31s               computetargettcpproxy-controller  The resource is up to date
  Warning  UpdateFailed  5s (x3 over 12s)  computetargettcpproxy-controller  Update call failed: error updating: Update operation not supported for regional ComputeTargetTCPProxy
```

Local test passed
```
--- PASS: TestCreateNoChangeUpdateDelete (0.18s)
    --- PASS: TestCreateNoChangeUpdateDelete/compute (0.00s)
        --- PASS: TestCreateNoChangeUpdateDelete/compute/basic-regionalcomputetargettcpproxy-direct (135.75s)
PASS
```


pkg/test/resourcefixture/contexts/compute_context.go
```diff
@@ -189,33 +189,6 @@ func init() {
...
-	}
-
-	resourceContextMap["regionalcomputeforwardingrule"] = ResourceContext{
-		ResourceKind: "ComputeForwardingRule",
-		SkipUpdate:   false,
-		// TestCreateNoChangeUpdateDelete/basic-regionalcomputeforwardingrule: dynamic_controller_integration_test.go:149:
-		//   value mismatch for label with key 'label-one': got 'value-two', want 'value-one'
-		SkipDriftDetection: true,
-	}
-
```
gemmahou, line 218: Those test names are out-dated. Also we now check for the existence of `update.yaml` to determine whether to skip the update: #2601

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
```diff
@@ -266,10 +268,18 @@ func (a *targetTCPProxyAdapter) Update(ctx context.Context, updateOp *directbase
...
 	parent := a.id.Parent()
 
+	status := &krm.ComputeTargetTCPProxyStatus{}
 	// Regional API does not support Update
 	if parent.Location != "global" {
-		return fmt.Errorf("update operation not supported for resource %v %v",
-			a.desired.GroupVersionKind(), k8s.GetNamespacedName(a.desired))
+		message := fmt.Sprintf("Update operation not supported for regional ComputeTargetTCPProxy")
+		updateOp.RecordEvent(v1.EventTypeWarning, k8s.UpdateFailed, message)
+		readyCondition := &v1alpha1.Condition{
```
yuwenma, line 276: Returning the error should be the same. Could you double check?
 `return fmt.Errorf(message)`

maqiuyujoyce: Yes, I'm also curious if there is anything special about this error, or is the direct controller not handling event update properly. Similar issue: https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/3050

gemmahou: > Returning the error should be the same. Could you double check? `return fmt.Errorf(message)`

During reconciliation, direct update operation is triggered([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/8319f918b9d5114eebfc4c760c0b99674486784e/pkg/controller/direct/directbase/directbase_controller.go#L308)). 

Return an error will cause reconciliation to fail([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/8319f918b9d5114eebfc4c760c0b99674486784e/pkg/controller/direct/directbase/directbase_controller.go#L313)); Recored failure event and update status will save the error in KRM object but the reconciliation will go through without errors. Both approach, the KRM object will remain error status after reconciliation.

I feel we should only error out when changes on certain fields are detected, either by returning an error or setting status to error. Please see the latest commit.

gemmahou: > Yes, I'm also curious if there is anything special about this error, or is the direct controller not handling event update properly. Similar issue: #3050

In direct controller, we handle the update operation but we don't currently record `Updating` event, and we skip the check([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/8319f918b9d5114eebfc4c760c0b99674486784e/pkg/controller/dynamic/dynamic_controller_integration_test.go#L269-L276)) in dynamic test. 

Direct controller only record event when there's error([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/lifecyclehandler/handler.go#L315)), or when it's UpToDate([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/lifecyclehandler/handler.go#L228)) for now.

What I did here is I recorded the event in line275: `updateOp.RecordEvent(v1.EventTypeWarning, k8s.UpdateFailed, message)`, and saved error into the object status. To fix #3050, we need to record the Updating event during direct reconciliation.

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
```diff
@@ -265,17 +263,16 @@ func (a *targetTCPProxyAdapter) Update(ctx context.Context, updateOp *directbase
...
-			a.desired.GroupVersionKind(), k8s.GetNamespacedName(a.desired))
-	}
-
 	tokens := strings.Split(a.id.String(), "/")
 	targetTCPProxy.Name = direct.LazyPtr(tokens[len(tokens)-1])
 
+	// Regional API does not support Update
+	errorMessage := fmt.Sprintf("Update operation not supported for regional ComputeTargetTCPProxy")
 	if !reflect.DeepEqual(targetTCPProxy.ProxyHeader, a.actual.ProxyHeader) {
+		updateOp.RecordUpdatingEvent()
```
yuwenma, line 272: Could you elaborate why we need to specifically record the event?

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
```diff
@@ -295,6 +292,10 @@ func (a *targetTCPProxyAdapter) Update(ctx context.Context, updateOp *directbase
 	}
 
 	if !reflect.DeepEqual(targetTCPProxy.Service, a.actual.Service) {
+		updateOp.RecordUpdatingEvent()
+		if parent.Location != "global" {
+			return fmt.Errorf("%s", errorMessage)
+		}
```
yuwenma, line 298: Does this mean the regional service now can support Update and cannot change the service field? I don' think we need to check this logic in KCC. Or is it related to the dynamic test? How it fixes the dynamic test?

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
maqiuyujoyce: Looks like in the direct controller flow, we completely dropped the `Updating` (and `Deleting` maybe) event? I personally prefer the "WIP" event to be recorded, but should that be handled at directbase level?

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
gemmahou: We record `Updating` events in legacy controller so that we can verify the reconciliation is triggered. In direct controller we currently don't handle Updating events, so that the https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/3050 occurs. 

Ideally we should record Updating events when the API call is triggered, but maybe not record it here. That should be handled more generically by the direct controller as Joyce suggested in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3582#discussion_r1941954301. I removed this code, and we can handle Updating in the future(not urgent).

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
gemmahou: Please correct me if I'm wrong: In direct controller, the `Update function` does not mean the KCC resource can be updated. It means the KCC resource is reconciling and only make Update API calls when certain condition is met. Every KCC resource needs to support `Update` to be able to reconcile.

If I log an error under `Update function`, I completely block the reconcile on that regional KCC resource, that's where the dynamic test error came from. To fix that, I need to log the error when changes are detected, ensure that we cannot change the Service or ProxyHeader fields in regional KCC resource. Other immutable fields will still be guarded by the CEL rule.

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
```diff
@@ -265,16 +263,18 @@ func (a *targetTCPProxyAdapter) Update(ctx context.Context, updateOp *directbase
...
-	}
-
 	tokens := strings.Split(a.id.String(), "/")
 	targetTCPProxy.Name = direct.LazyPtr(tokens[len(tokens)-1])
 
+	desiredSpec := &desired.Spec
+	actualSpec := ComputeTargetTCPProxySpec_FromProto(mapCtx, a.actual)
+	// Changes on resource spec are detected
+	// todo(yuhou): Can we have a more general way that can be applied to other controllers or base controller?
+	if !reflect.DeepEqual(desiredSpec, actualSpec) && parent.Location != "global" {
```
maqiuyujoyce, line 273: Could you create the nochange scenario test to verify that when there is no change, PATCH API is not triggered?

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
gemmahou: By saying `If I log an error under Update function, I completely block the reconciliation on the regional KCC resource`, I mean something like my original implementation:
```
func (a *targetTCPProxyAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOperation) error {
        (code)
	if parent.Location != "global" {
		return fmt.Errorf("update operation not supported for regional ComputeTargetTCPProxy")
	}
        (code)
```
With above code, when resource is regional, the reconciliation([Adapter.Update](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/8319f918b9d5114eebfc4c760c0b99674486784e/pkg/controller/direct/directbase/directbase_controller.go#L308)) will fail, even though there's no field change. 

The key point will be only error out when changes are detected. Like your comment: `you can return error if users are trying to update some fields in spec or return nil if the GCP does not support update and there is no field changed`. I implemented this way like:
```
func (a *targetTCPProxyAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOperation) error {
        (code)
	if **!reflect.DeepEqual(desiredSpec, actualSpec)** && parent.Location != "global" {
		return fmt.Errorf("update operation not supported for regional ComputeTargetTCPProxy")
	}
        (code)
```

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
gemmahou: Added scenario tests for nochange. Also added scenario tests for update(even it's covered by mock test) to compare with nochange tests scenario, so that we can see the differences in logs.

pkg/controller/direct/compute/targettcpproxy/refs.go
```diff
@@ -95,6 +95,9 @@ func resolveDependencies(ctx context.Context, reader client.Reader, obj *krm.Com
 
 		}
 		obj.Spec.BackendServiceRef.External = backendServiceRef.External
+		// Remove Name field as it's already resolved to External
+		// todo(yuhou): Shall we apply this to other resolvers?
+		obj.Spec.BackendServiceRef.Name = ""
```
maqiuyujoyce, line 100: As discussed offline, this is not required if we use CompareProtoMessage() to compare the diff during the Update().

Though I'm also curious, have we agreed on a common pattern for handling references here and why?

gemmahou: I'd say the whole resolvers is a temp solution and our goal is to use NormalizeExternal to resolve refs. Regarding if we want to keep "Name" or not, I don't think we've ever have discussion on this, because we haven't seen issues. And I remember Alex encountered an issue before where both "Name" and "External" exist when the resource got resolved twice.

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
```diff
@@ -265,15 +265,33 @@ func (a *targetTCPProxyAdapter) Update(ctx context.Context, updateOp *directbase
...
-	if parent.Location != "global" {
-		return fmt.Errorf("update operation not supported for resource %v %v",
-			a.desired.GroupVersionKind(), k8s.GetNamespacedName(a.desired))
+	// Assign API output-only values
+	targetTCPProxy.CreationTimestamp = a.actual.CreationTimestamp
+	targetTCPProxy.Id = a.actual.Id
+	targetTCPProxy.SelfLink = a.actual.SelfLink
+	targetTCPProxy.Kind = a.actual.Kind
+	// Convert `europe-west4` to `https://www.googleapis.com/compute/v1/projects/projectId/regions/europe-west4`
+	parts := strings.Split(a.actual.GetRegion(), "/")
```
maqiuyujoyce, line 276: Not a blocker: It may be clearer if we turn it into a util function. I think other Compute resources could leverage this function as well.

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
```diff
@@ -265,15 +265,33 @@ func (a *targetTCPProxyAdapter) Update(ctx context.Context, updateOp *directbase
 	updated := &computepb.TargetTcpProxy{}
```
maqiuyujoyce, line 265: Not a blocker: Maybe declare `op` and `updated` closer to their usage? I think they don't need to be declared explicitly, right?

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
```diff
@@ -265,15 +265,33 @@ func (a *targetTCPProxyAdapter) Update(ctx context.Context, updateOp *directbase
...
+	targetTCPProxy.Id = a.actual.Id
+	targetTCPProxy.SelfLink = a.actual.SelfLink
+	targetTCPProxy.Kind = a.actual.Kind
+	// Convert `europe-west4` to `https://www.googleapis.com/compute/v1/projects/projectId/regions/europe-west4`
+	parts := strings.Split(a.actual.GetRegion(), "/")
+	parts[len(parts)-1] = targetTCPProxy.GetRegion()
+	targetTCPProxy.Region = direct.LazyPtr(strings.Join(parts, "/"))
+	targetTCPProxy.Name = direct.LazyPtr(a.id.ID())
+
+	paths, err := common.CompareProtoMessage(targetTCPProxy, a.actual, common.BasicDiff)
```
maqiuyujoyce, line 281: Nice! Have you verified it with the no change scenarios test?

maqiuyujoyce (commented):
/lgtm

with some non-blocking suggestions and questions.

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
gemmahou: I can actually get the region string from `a.id.parent`, to avoid the complex string replacement. This way I think it's fine to leave all assign value functions there. See the latest commit.

pkg/controller/direct/compute/targettcpproxy/refs.go
gemmahou: I refactored BackenServiceRef in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3729 and removed this refs.go file.

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
gemmahou: I'm not going to address "record updating event" in this PR and the code has been removed.

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
gemmahou: I'm not going to address "record updating event" in this PR and the code has been removed.

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
```diff
@@ -261,63 +261,80 @@ func (a *targetTCPProxyAdapter) Update(ctx context.Context, updateOp *directbase
...
+		// Changes on resource spec are detected
+		if parent.Location != "global" {
+			// Regional ComputeTargetTCPProxy API does not support Update
+			return fmt.Errorf("update operation not supported for regional ComputeTargetTCPProxy")
 		}
-		if !op.Done() {
-			err = op.Wait(ctx)
+
+		op := &gcp.Operation{}
+		if !reflect.DeepEqual(targetTCPProxy.ProxyHeader, a.actual.ProxyHeader) {
```
maqiuyujoyce, line 295: Looks like we should error out for global resources as well when the updated path are neither proxy_header nor service?

maqiuyujoyce (commented):
Thank you for the thorough testing. I just have one small comment.

pkg/controller/direct/compute/targettcpproxy/targettcpproxy_controller.go
gemmahou: I thought I was using the CEL rule to prevent changes to immutable fields: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/develop-resources/api-conventions/validations.md#option-2

While after reading the guide again, I feel I mistakenly applied CEL rules to optional fields and that might cause issues. I will remove the CEL and check immutable fields in the controller in a followup PR. Thanks for pointing out this.

gemmahou: In the guide it mentions `If the field is optional, you should not use this approach, because the CRD validation disallows assigning a previous unset field if adding the immutable CEL rule. For example, setting the optional spec.preferredLocation from "" to "us-central1" violates this CEL rule.`. However, I feel the CEL rule here is correct(at least for this resource): assigning a previous unset field, i.e. optional field `description` **should** be blocked, and it's also unsupported by the API.

PR #3584, opened 2025-02-01
mockgcp: support for workflows

justinsb:
- **mockgcp: add test case for workflows workflow**
- **mockgcp: support for workflows**
- **mockgcp: generate bindings for workflows.v1**
- **mockgcp: capture golden output for workflow test**
- **tests: don't require header on any testdata files**


mockgcp/mockgcptests/proxy.go
```diff
@@ -121,6 +121,8 @@ func (p *Proxy) BuildGcloudConfig(proxyEndpoint *net.TCPAddr, mockgcp mockgcp.In
 		"storage.googleapis.com",
 		"cloudresourcemanager.googleapis.com",
 		"serviceusage.googleapis.com",
+		"workflows.googleapis.com",
+		"workflowexecutions.googleapis.com",
```
yuwenma, line 125: Cool! Is this added by AI via some hints like "update the service in proxy.go" or manually added?

yuwenma (commented):
/lgtm
/approve
/hold 

two questions, non blocker.

mockgcp/mockworkflows/testdata/workflow/crud/workflow-v2.yaml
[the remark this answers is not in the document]
justinsb: Yes, they are referenced from `script.yaml`

mockgcp/mockgcptests/proxy.go
justinsb: I think I manually added these, but yes it would be easy enough to ask AI to add them automatically.  I guess one gotcha is that we don't actually know the service names at this stage (until we've run the test).

Maybe we should just add all the hostnames we know of (or at least the ones that gcloud knows of!)

There might also be a better way to intercept gcloud traffic, I just haven't found one yet!

justinsb:
/assign @yuwenma 

We need this one so we don't need copyright headers on script.yaml

PR #3585, opened 2025-02-01
feat: add ManagedKafkaTopic as a direct resource

jingyih:


apis/managedkafka/v1alpha1/topic_types.go
```diff
@@ -0,0 +1,117 @@
...
+
+// ManagedKafkaTopicSpec defines the desired state of ManagedKafkaTopic
+// +kcc:proto=google.cloud.managedkafka.v1.Topic
+// ManagedKafkaTopicObservedState is the state of the ManagedKafkaTopic resource as most recently observed in GCP.
+type ManagedKafkaTopicObservedState struct {
+	// Identifier. The name of the topic. The `topic` segment is used when
+	//  connecting directly to the cluster. Structured like:
+	//  projects/{project}/locations/{location}/clusters/{cluster}/topics/{topic}
+	// +kcc:proto:field=google.cloud.managedkafka.v1.Topic.name
+	Name *string `json:"name,omitempty"`
```
jasonvigil, line 82: Is this the same as the externalRef?

apis/managedkafka/v1alpha1/topic_types.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// TODO(user): make sure the pluralizaiton below is correct
```
jasonvigil, line 87: Nit: Remove TODO comment

pkg/controller/direct/managedkafka/topic_controller.go
```diff
@@ -0,0 +1,251 @@
...
+	log := klog.FromContext(ctx)
+	log.V(2).Info("updating Topic", "name", a.id)
+	mapCtx := &direct.MapContext{}
+
+	desiredPb := ManagedKafkaTopicSpec_ToProto(mapCtx, &a.desired.DeepCopy().Spec)
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+
+	// cannot use common.CompareProtoMessage because some proto fields are not correctly labeled as output only, which means the path may include a output only field
```
jasonvigil, line 168: Which fields are incorrectly labeled? Seems the `name` field is the only output field. Reason I ask is because name is common to a lot of resources. Should we be treating name as an output field?

jasonvigil (commented):
Looks pretty good, just a couple questions, and a small nit.

pkg/controller/direct/managedkafka/topic_controller.go
jingyih: IIRC, the `name` field is not explicitly labeled as "output only." This means the YAML does not include `name` in `spec` since it isn't a spec field, but the actual GCP resource has a computed `name` field. As a result, the function detects a diff in the `name` field.  

However, for this resource, the `name` field is indeed an output-only field. In the Create API, the "topic_id" is required, while the `name` field is explicitly ignored.

https://github.com/googleapis/google-cloud-go/blob/0ddf33831c5f0bed7072d988795de353956f22db/managedkafka/apiv1/managedkafkapb/managed_kafka.pb.go#L683-L690

> Should we be treating name as an output field?

I believe it depends on the resource. If a resource ignores the "name" field during creation and computes its value afterward, then it should be considered an output-only field.

apis/managedkafka/v1alpha1/topic_types.go
jasonvigil: I have not been adding this field to observedState for my resources, since it is the same as externalRef.

pkg/controller/direct/managedkafka/topic_controller.go
jasonvigil: > As a result, the function detects a diff in the name field.

To work around this issue, I have been explicitly setting the name field in the proto object (after mapping ToProto), before comparing. Here is an example: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/c036074e5aec3c1907835a83b61ffc62172485d1/pkg/controller/direct/workstations/workstation_controller.go#L202-L209

jingyih: Sound good. I will give it a try. Thanks!

apis/managedkafka/v1alpha1/topic_types.go
jingyih: I added TODOs to remove the field in a followup PR.

pkg/controller/direct/managedkafka/topic_controller.go
jingyih: PR updated.

apis/managedkafka/v1alpha1/topic_types.go
```diff
@@ -0,0 +1,116 @@
...
+
+var ManagedKafkaTopicGVK = GroupVersion.WithKind("ManagedKafkaTopic")
+
+// ManagedKafkaTopicSpec defines the desired state of ManagedKafkaTopic
+// +kcc:proto=google.cloud.managedkafka.v1.Topic
+type ManagedKafkaTopicSpec struct {
+	commonv1alpha1.CommonSpec `json:",inline"`
+
+	// +required
+	Location string `json:"location"`
```
yuwenma, line 31: Suggest adding descriptions for Location and ClusterRef, because these comments are the source to generate the reference doc, which is the most straightforward (if not only) way for users to configure the KCC object

apis/managedkafka/v1alpha1/topic_types.go
```diff
@@ -0,0 +1,116 @@
...
+
+// ManagedKafkaTopicSpec defines the desired state of ManagedKafkaTopic
+// +kcc:proto=google.cloud.managedkafka.v1.Topic
+// ManagedKafkaTopicObservedState is the state of the ManagedKafkaTopic resource as most recently observed in GCP.
+type ManagedKafkaTopicObservedState struct {
+	// Identifier. The name of the topic. The `topic` segment is used when
+	//  connecting directly to the cluster. Structured like:
+	//  projects/{project}/locations/{location}/clusters/{cluster}/topics/{topic}
+	// +kcc:proto:field=google.cloud.managedkafka.v1.Topic.name
+	Name *string `json:"name,omitempty"` // TODO(jingyih): remove this field as it is the same as externalRef
```
yuwenma, line 82: Wait, any reasons to add this field (with a cleanup TODO) if this is a pure new Alpha resource?

yuwenma:
/approve
/hold

Left a few  nits. Non blockers. Hold in case you can take a look

apis/managedkafka/v1alpha1/topic_types.go
jingyih: Yeah this is a new resource. I am removing the field in a followup PR.

apis/managedkafka/v1alpha1/topic_types.go
jingyih: Done! [c6056a0](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3585/commits/c6056a0cdb1644a46d9ea09463c5cccb57071c34)

jingyih:
PR rebased to pick up the fix for broken HEAD.

PR #3592, opened 2025-02-03
codebot: require two empty lines, so that we can use newlines in prompts

justinsb:


cheftako:
/lgtm
/approve

Thanks for the fix Justin

PR #3595, opened 2025-02-04
[WIP]Add a workflow to automate config connector release bundle creation

xiaoweim:
### Change description

Fixes #

### Tests you have done


.github/workflows/release-automation.yaml
```diff
@@ -0,0 +1,35 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Config Connector Release Automation
+
+on:
+  push:
+    tags:
+      - 'v*'
```
yuwenma, line 20: can we change this to only match semver2?

dev/tasks/deploy-release-manifest
```diff
@@ -0,0 +1,72 @@
...
+echo "Added autopilot release manifest to ${MANUAL_DIR}/operator-system/autopilot-configconnector-operator.yaml"
+
+cp -rf operator/config/samples ${MANUAL_DIR}/
+
+tar -czvf ${BUILD_DIR}/release-bundle.tar.gz -C ${MANUAL_DIR}/ .
+
+echo "Generated ${BUILD_DIR}/release-bundle.tar.gz for manual installation."
+
+gsutil cp ${REPO_ROOT}/.build/release-bundle.tar.gz gs://configconnector-operator/${LATEST_VERSION}/
+gsutil cp ${REPO_ROOT}/.build/release-bundle.tar.gz gs://configconnector-operator/latest
```
yuwenma, line 70: Pushing a new tag to github needs less permission and pushing a release bundle. Thus, this change may cause safety hazards since it gives the release permission to a broader group.  Maybe it is easier to move the gs push step to internal (gob)?

config/installbundle/release-manifests/standard/kustomization.yaml
```diff
@@ -18,6 +18,6 @@ commonLabels:
 commonAnnotations:
   cnrm.cloud.google.com/operator-version: "1.128.0"
 resources:
-  - crds.yaml
```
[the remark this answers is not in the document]
xiaoweim: For the release script, I have added the correct operator image when generating the `manager_image_patch_template.yaml`, so the image in the statefulset is updated to the right version.

dev/tasks/deploy-release-manifest
```diff
@@ -0,0 +1,72 @@
...
+# List all directories in the specified path and store them in an array
+VERSIONS=($(ls -d "$FOLDER_PATH"/*/ 2>/dev/null | xargs -n 1 basename | sort -V))
+
+if [ ${#VERSIONS[@]} -eq 0 ]; then
+    echo "Error: No version directories found in $FOLDER_PATH"
+    exit 1
+fi
+    
+LATEST_VERSION="${VERSIONS[-1]}"
+OPERATOR_IMG=gcr.io/gke-release/cnrm/operator:${LATEST_VERSION}
```
xiaoweim, line 49: @yuwenma This is the place where the operator image is specified before we update the `manager_image_patch_template.yaml` below.

dev/tasks/deploy-release-manifest
xiaoweim: I want to make sure that I am understanding this correctly. Do you mean that we still build this release bundle in OSS, wait until it is copied over to internal and then push it. Or that we both build the release bundle and then push internally?

.github/workflows/release-automation.yaml
xiaoweim: Sure, I can add expression format for standard versions. I think we do not have to trigger release bundle build for pre-release versions like `v1.0.0-alpha.1`, `v2.0.0-beta.1`, `v1.0.0-rc.1`.

dev/tasks/deploy-release-manifest
```diff
@@ -0,0 +1,74 @@
...
+
+echo "Updating kustomize image patch file for manager."
+cp ${REPO_ROOT}/operator/config/manager/manager_image_patch_template.yaml ${REPO_ROOT}/operator/config/manager/manager_image_patch.yaml
+sed -i'' -e 's@image: .*@image: '"${OPERATOR_IMG}"'@' ${REPO_ROOT}/operator/config/manager/manager_image_patch.yaml
+cp -f ${REPO_ROOT}/operator/config/autopilot-manager/manager_image_patch_template.yaml ${REPO_ROOT}/operator/config/autopilot-manager/manager_image_patch.yaml
+sed -i'' -e 's@image: .*@image: '"${OPERATOR_IMG}"'@' ${REPO_ROOT}/operator/config/autopilot-manager/manager_image_patch.yaml
+
```
yuwenma, line 56: We should use kustomize to replace the image tag.

xiaoweim:
Could you please also update the test log and rebase to master head? Thank you!

yuwenma (commented):
/hold

The code itself looks good but we cannot merge it now because everyone has permission to "create the release bundle" by pushing a trigger (git tag). @xiaoweim Could you share the overall plan on how to deal with this issue?

PR #3596, opened 2025-02-04
chore: fix test file name for managedkafkacluster-basic

jingyih:
The filename should be updated when renaming the test folder, but this was overlooked in #3576. Since mock support is still being added, the PR was not properly gated.


yuwenma (commented):
/lgtm
/approve

PR #3597, opened 2025-02-04
tool: use LLM to generate fuzzer

jingyih:
This PR:  
- Enhances the "prompt" tool by incorporating matching Go structs. For example, when generating the fuzzer for the `ManagedKafkaCluster` resource, the enhancer will identify the corresponding Go structs, `ManagedKafkaClusterSpec` and `ManagedKafkaClusterObservedState`, and include them in the prompt.
- Adds the `// +tool:fuzz-gen` annotation to existing fuzzers in the codebase as data points.  
- Adds default values to the "prompt" command flags for developer convenience, accommodating their various source code locations.  
- Generates the fuzzer for `ManagedKafkaCluster` using the following command, with **no** manual edits (besides updating the year in the license).

```
$ controllerbuilder prompt <<EOF > cluster_fuzzer.go
// +tool:fuzz-gen
// proto.message: google.cloud.managedkafka.v1.Cluster
// krm.kind: ManagedKafkaCluster
EOF
```

### DIFF
--- a/dev/tools/controllerbuilder/pkg/commands/exportcsv/prompt.go
+++ b/dev/tools/controllerbuilder/pkg/commands/exportcsv/prompt.go
@@ -45,10 +45,7 @@ type PromptOptions struct {
 }
 
 func (o *PromptOptions) InitDefaults() error {
-	root, err := options.RepoRoot()
-	if err != nil {
-		return err
-	}
+	root := os.Getenv("REPO_ROOT")
 	o.SrcDir = root
 	o.ProtoDir = root + "/.build/third_party/googleapis/"
 	return nil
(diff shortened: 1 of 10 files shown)


yuwenma (commented):
/lgtm
/approve

Awesome!!!

dev/tools/controllerbuilder/pkg/commands/exportcsv/prompt.go
```diff
@@ -44,6 +44,16 @@ type PromptOptions struct {
 	StrictInputColumnKeys []string
 }
 
+func (o *PromptOptions) InitDefaults() error {
```
yuwenma, line 47: Suggest expose this defaulting, like in cmd flag, make rules, or bash script. A common pain point we have is that some variables are modified all over the place especially in the release pipeline. Say if we later on change the proto to somehwere like `pkg/internal/third_party/googleapis/1.123/` (for versioning), it wold be hard to discover this function.

jingyih:
Rebased to pick up the fix for broken HEAD.

PR #3598, opened 2025-02-04
Support direct migration in type generator

maqiuyujoyce:
### Change description

First trial to match type comment with field description from the existing CRD for direct migration.

### Tests you have done

### DIFF
--- a/dev/tools/controllerbuilder/pkg/codegen/common.go
+++ b/dev/tools/controllerbuilder/pkg/codegen/common.go
@@ -39,7 +39,8 @@ var protoMessagesNotMappedToGoStruct = map[string]string{
 
 var Acronyms = []string{
 	"ID", "HTML", "URL", "URI", "HTTP", "HTTPS", "SSH",
-	"IP", "GB", "FS", "PD", "KMS", "GCE", "VTPM",
+	"IP", "GB", "FS", "PD", "KMS", "GCE", "VTPM", "SSL",
+	"CA", "CPU", "SQL", "PSC", "DNS",
 }
 
 // IsAcronym returns true if the given string is an acronym
--- a/dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
+++ b/dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
@@ -24,20 +24,23 @@ import (
 	"strings"
 
 	"github.com/GoogleCloudPlatform/k8s-config-connector/dev/tools/controllerbuilder/pkg/protoapi"
-	"k8s.io/apimachinery/pkg/util/sets"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/text"
 
 	"google.golang.org/genproto/googleapis/api/annotations"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/reflect/protoreflect"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 )
 
 type TypeGenerator struct {
 	generatorBase
-	api             *protoapi.Proto
-	goPackage       string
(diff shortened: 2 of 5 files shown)


dev/tools/controllerbuilder/pkg/codegen/common.go
```diff
@@ -39,7 +39,8 @@ var protoMessagesNotMappedToGoStruct = map[string]string{
 
 var Acronyms = []string{
 	"ID", "HTML", "URL", "URI", "HTTP", "HTTPS", "SSH",
-	"IP", "GB", "FS", "PD", "KMS", "GCE", "VTPM",
+	"IP", "GB", "FS", "PD", "KMS", "GCE", "VTPM", "SSL",
+	"CA", "CPU", "SQL", "PSC", "DNS",
```
yuwenma, line 43: Others looks good but I'm not sure about CA and PSC.

yuwenma (commented):
/assign @jingyih

jingyih:
I’d like to better understand the use case and context. Why not simply update the field description to the newest? I noticed that for many resources migrated to direct resource, the field descriptions have been updated to align with the latest descriptions from the proto.

maqiuyujoyce:
I am under the impression that we want the migration to be transparent to users. They don't need to be aware of the migration, they don't need to opt-in, and they don't need to adapt to the description updates. That's what I did for AlloyDBInstance.

If the team agrees that we'll update the field descriptions to match the proto in the migration, then I'm okay with that too.

jingyih:
As I understand it, we can update CRD fields over time (adding new fields or updating comments on existing ones). We're not stuck with the same field comments. These migrations remain transparent to the user - they only notice that field comments have been updated during a release.

WDYT @yuwenma

maqiuyujoyce:
To improve velocity, I think it's a good idea to not try to match to existing field descriptions.
Though looking at the diff for AlloyDBCluster (the resource I'm working on), about 20+ spec fields have description changes within one version, and ~10 of them have actual change, e.g. stating `required` in the existing CRD and `optional` in the new CRD.

I think it could be a bit confusing if we update the description when we still use the TF-based controller for the resource. But I'm okay that we manually pick out the ones that can be confusing.

dev/tools/controllerbuilder/pkg/codegen/common.go
maqiuyujoyce: Context about PSC and CA:
* PSC: https://cloud.google.com/vpc/docs/private-service-connect, https://medium.com/google-cloud/gcp-private-service-connect-psc-service-publication-62eaf1d58651
* CA: https://cloud.google.com/security/products/certificate-authority-service, https://medium.com/google-cloud/gcp-ca-service-how-to-get-started-6eba056c9e12

yuwenma:
> As I understand it, we can update CRD fields over time (adding new fields or updating comments on existing ones). We're not stuck with the same field comments. These migrations remain transparent to the user - they only notice that field comments have been updated during a release.
> 

Right. field description can be updated as stated here (2.4-3 #2  "The go comment of the field can be changed.") And I think it is better to be because the GCP service is also improving their APIs, like immutability, required/option which can be valuable information for the users.

maqiuyujoyce:
Okay, so I think for a migration, we should use the latest comments from protobuf if it makes sense and backwards compatible.

In the guideline for migration (https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/develop-resources/scenarios/migrate-tf-resource-beta.md#add-the-backward-compatible-api), it mentioned that `make sure the CRD are the same (comment changes are acceptable)`. This doesn't tell contributors to prefer using the the newly generated comments. I think this should be `make sure the CRD are the same. Comment changes to match the latest protobuf are recommended as long as the content doesn't contain breaking changes.`

maqiuyujoyce:
Created #3623.

Issue #3571, opened 2025-01-30
TestCRDFieldPresenceInUnstructured does not detect list fields


jasonvigil:
### Checklist

- [x] I did not find a related open issue.
- [x] I did not find a solution in the troubleshooting guide: (https://cloud.google.com/config-connector/docs/troubleshooting)
- [x] If this issue is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Bug Description

List fields are incorrectly reported by the test as not set, for all resources.

### Additional Diagnostic Information

For example, in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3544#discussion_r1932372371 we noticed the error
```
[missing_field] crd=apigeeenvgroups.apigee.cnrm.cloud.google.com version=v1beta1: field ".spec.hostnames[]" is not set in unstructured objects
```
This should not fail, because the field is set in the test fixture [here](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/49660b4db0fe572c246b28ae3d69fdbcbb7b9612/pkg/test/resourcefixture/testdata/basic/apigee/v1beta1/apigeeenvgroup/apigeeenvgroup-basic/create.yaml#L20-L21)

### Kubernetes Cluster Version

n/a

### Config Connector Version

n/a

### Config Connector Mode

cluster mode

### Steps to reproduce the issue

Run the test in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3435

### YAML snippets

```YAML

```

jingyih:
I believe that line needs to be manually deleted.

jingyih:
Oh I misunderstood earlier. This looks like a bug in the test.

/cc the test author @acpana

jasonvigil:
Fixed in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3651

PR #3599, opened 2025-02-04
Promote ApigeeInstance to v1beta1

jasonvigil:


apis/apigee/v1alpha1/environmentgroup_types.go
```diff
@@ -24,7 +24,7 @@ var ApigeeEnvgroupGVK = GroupVersion.WithKind("ApigeeEnvgroup")
 
 type Parent struct {
 	// +required
-	OrganizationRef *refv1alpha1.ApigeeOrganizationRef `json:"organizationRef"`
+	OrganizationRef *refs.ApigeeOrganizationRef `json:"organizationRef"`
```
yuwenma, line 27: We should move the ApigeeOrganizationRef to apis/apigee? So version promotion do not need to change the import, and we can have different versions for ApigeeOrganizationRef

apis/apigee/v1beta1/environmentgroup_identity.go
```diff
@@ -19,7 +19,7 @@ import (
 	"fmt"
 	"strings"
 
-	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1alpha1"
+	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
```
yuwenma, line 22: ditto

pkg/test/resourcefixture/sets.go
```diff
@@ -89,6 +89,7 @@ func addResourceConfig(t *testing.T, smLoader *servicemappingloader.ServiceMappi
 func IsPureDirectResource(gk schema.GroupKind) bool {
```
yuwenma, line 89: @maqiuyujoyce Could you confirm if we still need this hack with the removal of the service mapping ?

tests/apichecks/testdata/exceptions/missingfields.txt
```diff
@@ -65,6 +65,7 @@
 [missing_field] crd=alloydbinstances.alloydb.cnrm.cloud.google.com version=v1beta1: field ".spec.networkConfig.authorizedExternalNetworks[].cidrRange" is not set in unstructured objects
 [missing_field] crd=alloydbusers.alloydb.cnrm.cloud.google.com version=v1beta1: field ".spec.databaseRoles[]" is not set in unstructured objects
 [missing_field] crd=apigeeenvgroups.apigee.cnrm.cloud.google.com version=v1beta1: field ".spec.hostnames[]" is not set in unstructured objects
+[missing_field] crd=apigeeinstances.apigee.cnrm.cloud.google.com version=v1beta1: field ".spec.consumerAcceptList[]" is not set in unstructured objects
```
yuwenma, line 68: This will leave the field uncovered  (and does not meet the exit criteria). Can we fix the code to make the CRD check pass? cc @acpana

pkg/controller/direct/apigee/instance_controller.go
yuwenma, line 1: Update method does not contain 5 fields. Some like `Description`, `DisplayName`, `KMSKey` seems to be mutable. Is there any reasons for that?

It would be a better UX to check all changed fields, and (either add logic in controller or from GCP response to) give user the guidance on which resource is immutable. Otherwise, they have no way to tell.

yuwenma:
/approve

A few nits for better UX and maintenance.

tests/apichecks/testdata/exceptions/missingfields.txt
jasonvigil: +1, this test appears to be broken https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/3571. @acpana can you please help fix this?

apis/apigee/v1beta1/environmentgroup_identity.go
jasonvigil: Will do, in a follow-up PR.

apis/apigee/v1alpha1/environmentgroup_types.go
jasonvigil: Will do, in a follow-up PR.

pkg/controller/direct/apigee/instance_controller.go
yuwenma: Sorry, I didn't make it clear. We can either "allow user to attempt to change them anyway, and return the error from GCP" or add the check in the controller. The advantage for the former is that the users can get the exact error message from GCP and once GCP changes the behavior KCC does not need to update the code. The disadvantage is repeated reconcile failures (which may not be too bad).

jasonvigil: Sounds good, I agree. Will make the change

pkg/test/resourcefixture/sets.go
jasonvigil: I created https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/3614 to track this improvement

maqiuyujoyce: My change in #3443 didn't touch this hack so yes, let's still use it to avoid potential issues.

@jasonvigil filed an issue #3614 (thank you, Jason) and I'll take a look at it later.

pkg/controller/direct/apigee/instance_controller.go
jasonvigil: ok updated

apis/refs/v1beta1/apigeeorganizationref.go
```diff
@@ -12,14 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package v1alpha1
+package v1beta1
```
acpana, line 15: Ah I think this bit me in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3605

apis/apigee/v1alpha1/environmentgroup_types.go
jasonvigil: Done in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3639

apis/apigee/v1beta1/environmentgroup_identity.go
jasonvigil: See https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3599#discussion_r1951633615, this is done in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3639

PR #3591, opened 2025-02-03
feat: Add CloudIdentityGroup types

gemmahou:
### Change description

Added CloudIdentityGroup types.
Reference PR: #2804 

### Tests you have done

- [x] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #3603, opened 2025-02-04
feat: Add CloudIdentityGroup mappers

gemmahou:
### Change description

Based on #3591

### Tests you have done


pkg/controller/direct/cloudidentity/roundtrip_test.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test/fuzz"
+	"github.com/google/go-cmp/cmp"
+	"google.golang.org/protobuf/encoding/prototext"
+	"google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/testing/protocmp"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+func FuzzCloudIdentityGroupSpec(f *testing.F) {
```
jingyih, line 32: We’ve migrated to the new fuzz library for fuzz testing. Check out an example here: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/f313b00c52f09c4a52a2eb5fe2c15fa4b30a05fd/pkg/controller/direct/discoveryengine/fuzzers.go#L26-L47

You can also try using the LLM tool to generate the fuzz for you:)
https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/develop-resources/deep-dives/3-add-mapper.md#33-optional-add-fuzzer-test

pkg/controller/direct/cloudidentity/roundtrip_test.go
```diff
@@ -0,0 +1,140 @@
...
+	f.Fuzz(func(t *testing.T, seed int64) {
+		randStream := rand.New(rand.NewSource(seed))
+
+		p1 := &pb.Group{}
+		fuzz.FillWithRandom(t, randStream, p1)
+
+		// A few fields are not implemented yet in KRM, don't test them
+		unimplementedFields := sets.New[string](
```
jingyih, line 40: /cc @yuwenma 
Just want to double check that we're okay with adding these new fields later. i.e., the resource migration isn’t because the customer has specifically asked for the new fields.

pkg/controller/direct/cloudidentity/roundtrip_test.go
jingyih: Oh, actually, the LLM tool won't work for this resource because the Go structs aren't annotated with their corresponding proto messages.

pkg/controller/direct/cloudidentity/mapper.go
```diff
@@ -0,0 +1,101 @@
...
+	if in == nil {
+		return nil
+	}
+	out := &krm.CloudIdentityGroupStatus{}
+	out.CreateTime = direct.StringTimestamp_FromProto(mapCtx, in.GetCreateTime())
+	out.Name = in.Name
+	out.UpdateTime = direct.StringTimestamp_FromProto(mapCtx, in.GetUpdateTime())
+	return out
+}
+func CloudIdentityGroupStatus_ToProto(mapCtx *direct.MapContext, in *krm.CloudIdentityGroupStatus) *pb.Group {
```
jingyih, line 74: The mapper functions for `Status` should also include the mapper for `ObservedState`?

pkg/controller/direct/cloudidentity/roundtrip_test.go
gemmahou: I think we should be able to support new fields into Beta resources, would be a feature request.

This migration is not related to a FR on new fields: b/https://buganizer.corp.google.com/issues/367975895#comment37

pkg/controller/direct/cloudidentity/mapper.go
gemmahou: I have ObservedState here: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3603/files#diff-4638cbf1c68cb7c4e608879bab2c68e716623d1ff4278191902a4d44c85be472R24-R39

jingyih: Sorry for the confusion! What I meant was: Since `ObservedState` is a subfield of `Status`, shouldn't its mapping function be included within the mapping function of `Status`?

gemmahou: Hmmm I do not think so? Our mappers(To/From proto) only does the conversion between krm fields and proto fields. `ObservedState` is not a proto field, so we have a separate function for it to convert its subfield To/From proto.

jingyih: Ah, I see. This is a special case where we split the "output-only" fields into two Go structures (`Status` and `Status.ObservedState`) for backward compatibility reasons. While `ObservedState` itself is not a proto field, its subfield `AdditionalGroupKeys` is. That said, the current implementation works.

jingyih (commented):
/lgtm

pkg/controller/direct/cloudidentity/mapper.go
gemmahou: QQ: Shall we move `name`, `createTimestamp` and `updateTimestamp`(currently under `Status`) to `ObservedState`? Asking because that would introduce changes to existing Beta CRD... but since they are also output only fields they should be `ObservedState`? `Status` should only contain `[CommonStatus](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/4cbb076b4dd8f47572b0edf0714c880f1e0e3ddc/pkg/apis/common/v1alpha1/types.go#L32)` and `ObservedState`?

jingyih: Removing them from `Status` would be a breaking change, which we need to avoid. Optionally, we could copy those output fields to `Status.ObservedState`, but that would introduce redundancy.

gemmahou: > Ah, I see. This is a special case where we split the "output-only" fields into two Go structures (`Status` and `Status.ObservedState`) for backward compatibility reasons. While `ObservedState` itself is not a proto field, its subfield `AdditionalGroupKeys` is. That said, the current implementation works.

Per [discussion](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3647#issuecomment-2648816479) I added the `observedState` as subfield of `Status` in mappers.

pkg/controller/direct/cloudidentity/group_mapper.go
```diff
@@ -0,0 +1,103 @@
...
+}
+func CloudIdentityGroupStatus_ToProto(mapCtx *direct.MapContext, in *krm.CloudIdentityGroupStatus) *pb.Group {
+	if in == nil {
+		return nil
+	}
+	out := &pb.Group{}
+	out.CreateTime = direct.StringTimestamp_ToProto(mapCtx, in.CreateTime)
+	out.Name = in.Name
+	out.UpdateTime = direct.StringTimestamp_ToProto(mapCtx, in.UpdateTime)
+	out.AdditionalGroupKeys = CloudIdentityGroupObservedState_ToProto(mapCtx, in.ObservedState).AdditionalGroupKeys
```
jingyih, line 83: check if `in.ObservedState` is nil here?

gemmahou: We have nil check in CloudIdentityGroupObservedState_ToProto(line 33), and I updated the code to check if `CloudIdentityGroupObservedState_ToProto` is nil-that would cause issue when we call `.AdditionalGroupKeys` if it's nil.

jingyih (commented):
/lgtm

maqiuyujoyce (commented):
/approve

PR #3604, opened 2025-02-05
fix: fix the update diff in ManagedKaflaCluster

jingyih:
Fix the update diff on the `name` field in `ManagedKaflaCluster`, similar to how it's handled for `ManagedKaflaTopic` in #3585.


yuwenma (commented):
/approve
Could you try running this in real gcp as well?

PR #3398, opened 2024-12-17
fix: Fix g3doc for refs

gemmahou:
### Change description

We are using `projects/<projectID>/global/networks/<network>` in template but that will cause issues in our g3doc, screenshot: https://screenshot.googleplex.com/5Hg3goroSATbXDF.png. I changed the template to escape braces `{{` and `}}` during Golang Template rendering: `projects/{{projectID}}/global/networks/{{network}}`.

I noticed there are other refs using `[[` and `]]`. Other than consistency I don't see doc issues, so I skip them for now.

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #3605, opened 2025-02-05
apis: ApigeeEnvgroupAttachment

acpana:
Adds direct style APIs for the `ApigeeEnvgroupAttachment`. This is an alpha resource, still running over the TF controller.

Next PR:
- add a direct controoler
- ~change envgroup and environment fields to references!~ [ did it in this PR ]


apis/apigee/v1alpha1/envgroupattachment_reference.go
```diff
@@ -0,0 +1,86 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ refsv1beta1.ExternalNormalizer = &ApigeeEnvgroupAttachmentRef{}
+
+// ApigeeEnvgroupAttachmentRef defines the resource reference to ApigeeEnvgroupAttachment, which "External" field
+// holds the GCP identifier for the KRM object.
+type ApigeeEnvgroupAttachmentRef struct {
+	// A reference to an externally managed ApigeeEnvgroupAttachment resource.
+	// Should be in the format "projects/*/locations/global/envgroups/*/attachments/*".
```
jasonvigil, line 35: This format looks incorrect to me, shouldn't it be in the pattern of `organizations/{{organizationID}}/envgroups/{{envgroupID}}/attachments/{{attachmentID}}` ?

apis/apigee/v1alpha1/envgroupattachment_identity.go
```diff
@@ -0,0 +1,128 @@
...
+			Envgroup:     obj.Spec.Envgroup,
+		},
+		id: resourceID,
+	}, nil
+}
+
+func ParseApigeeEnvgroupAttachmentExternalRef(external string) (parent *EnvgroupAttachmentParent, resourceID string, err error) {
+	tokens := strings.Split(external, "/")
+	if len(tokens) != 6 {
+		return nil, "", fmt.Errorf("invalid external format: %s, expecting organizations/*/envgroups/*/attachments/*", external)
```
jasonvigil, line 119: For consistency with https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3398, I think we should use the same pattern (using `{{xyzID}}` instead of `*`)

apis/apigee/v1alpha1/envgroupattachment_identity.go
```diff
@@ -0,0 +1,128 @@
...
+		return nil, fmt.Errorf("cannot resolve resource ID")
+	}
+
+	externalRef := common.ValueOf(obj.Status.ExternalRef)
+	if externalRef != "" {
+		actualParent, actualResourceID, err := ParseApigeeEnvgroupAttachmentExternalRef(externalRef)
+		if err != nil {
+			return nil, err
+		}
+		if actualParent.Envgroup != obj.Spec.Envgroup {
```
jasonvigil, line 95: tiny nit, feels a bit weird to compare these fields out-of-order. i.e. envgroup -> org -> id instead of org -> envgroup -> id

apis/apigee/v1alpha1/envgroupattachment_types.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+var EnvgroupAttachmentGVK = GroupVersion.WithKind("ApigeeEnvgroupAttachment")
+
+// ApigeeEnvgroupAttachmentSpec defines the desired state of ApigeeEnvgroupAttachment
+type ApigeeEnvgroupAttachmentSpec struct {
+	// +required
+	OrganizationRef *refv1alpha1.ApigeeOrganizationRef `json:"organizationRef"`
+
+	// Immutable. The Apigee environment to attach to.
```
jasonvigil, line 32: I thought this resource attaches an environment to an envgroup. Not the other way around. Where did this comment come from? Was it originally from the API? In any case, I think we should update this.

apis/apigee/v1alpha1/envgroupattachment_types.go
```diff
@@ -0,0 +1,89 @@
...
+	Environment string `json:"environment"`
+
+	// Immutable. The Apigee environment group which will host the environment.
```
jasonvigil, line 35: Nit, would put this field before Environment, because Envgroup is part of the parent.

apis/apigee/v1alpha1/envgroupattachment_types.go
```diff
@@ -0,0 +1,89 @@
...
+	Envgroup string `json:"envgroup"`
+
+	// The ApigeeEnvgroupAttachment name. If not given, the metadata.name will be used.
+	ResourceID *string `json:"resourceID,omitempty"`
+}
+
+// ApigeeEnvgroupAttachmentStatus defines the observed state of ApigeeEnvgroupAttachment
+type ApigeeEnvgroupAttachmentStatus struct {
```
jasonvigil, line 43: There should be an observedState.createdAt field? https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/e127b9658a2e042cf45980f5b0e1409734a7c2f5/mockgcp/apis/mockgcp/cloud/apigee/v1/service.proto#L3977

apis/apigee/v1alpha1/environment_reference.go
```diff
@@ -0,0 +1,101 @@
...
+
+var _ refsv1beta1.ExternalNormalizer = &ApigeeEnvironmentRef{}
+
+var EnvironmentGVK = GroupVersion.WithKind("ApigeeEnvironment") // todo acpana house this in environment_types.go eventually
+
+// ApigeeEnvironmentRef defines the resource reference to ApigeeEnvironment, which "External" field
+// holds the GCP identifier for the KRM object.
+type ApigeeEnvironmentRef struct {
+	// A reference to an externally managed ApigeeEnvironment resource.
+	// Should be in the format "projects/*/environments/*".
```
jasonvigil, line 38: This should have format starting with `organizations/{{organizationID}}/...`, no?

apis/apigee/v1alpha1/environment_reference.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+		return "", fmt.Errorf("reading referenced %s %s: %w", EnvironmentGVK, key, err)
+	}
+	// Get external from status.externalRef. This is the most trustworthy place.
+	actualExternalRef, found, err := unstructured.NestedString(u.Object, "status", "externalRef")
+	if err != nil {
+		return "", fmt.Errorf("reading status.externalRef: %w", err)
+	}
+	if !found {
+		// todo acpana careful to deal with this
```
jasonvigil, line 82: In case you are not the next person who hits this issue, could you please put a more descriptive comment? Would suggest linking to bug / ticket and including some details around expected issue behavior, resolution hints, etc.

apis/apigee/v1alpha1/environment_reference.go
```diff
@@ -0,0 +1,101 @@
...
+		return "", fmt.Errorf("status.externalRef is not found")
+	}
+	if actualExternalRef == "" {
+		return "", k8s.NewReferenceNotReadyError(u.GroupVersionKind(), key)
+	}
+	r.External = actualExternalRef
+	return r.External, nil
+}
+
+// todo acpana this should be in a _identity.go file eventually
```
jasonvigil, line 92: Can we do this now?

apis/apigee/v1alpha1/environment_reference.go
```diff
@@ -0,0 +1,101 @@
...
+func ParseApigeeEnvironmentExternalRef(external string) (org string, environment string, err error) {
+	tokens := strings.Split(external, "/")
+
+	if len(tokens) != 4 || tokens[0] != "organizations" || tokens[2] != "environments" {
+		return "", "", fmt.Errorf("external should be organizations/{organization}/environments/{environment}, got: %s", external)
```
jasonvigil, line 97: For consistency with other resources, we should use `{{organizationID}}` (double curly)

apis/apigee/v1alpha1/envgroupattachment_types.go
```diff
@@ -0,0 +1,89 @@
...
+
+var EnvgroupAttachmentGVK = GroupVersion.WithKind("ApigeeEnvgroupAttachment")
+
+// ApigeeEnvgroupAttachmentSpec defines the desired state of ApigeeEnvgroupAttachment
+type ApigeeEnvgroupAttachmentSpec struct {
+	// +required
+	OrganizationRef *refv1alpha1.ApigeeOrganizationRef `json:"organizationRef"`
+
+	// Immutable. The Apigee environment to attach to.
+	Environment string `json:"environment"`
```
jasonvigil, line 33: We should use a reference type here

apis/apigee/v1alpha1/envgroupattachment_types.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// Immutable. The Apigee environment group which will host the environment.
+	Envgroup string `json:"envgroup"`
```
jasonvigil, line 36: Same, this should be a reference type, right?

apis/apigee/v1alpha1/envgroupattachment_types.go
acpana: Let's convert these to reference types in the next PR. If we do this now we have to have a direct controller so we need the direct controller first

apis/apigee/v1alpha1/envgroupattachment_types.go
acpana: I think the resource docs are a little wonky. They mark the `envgroup` as output only but we clearly rely on it for creation so ... 🤷🏼 

Updated the comment and will probably investigate the docs issue some more in the background.

apis/apigee/v1alpha1/envgroupattachment_types.go
acpana: Let's add this in once we have the direct controller!

apis/apigee/v1alpha1/envgroupattachment_types.go
jasonvigil: I think we have already broken backwards compatibility with the TF controller in a couple of ways:
* spec.envgroupId -> spec.envgroup
* spec.organizationRef is now a required field
If we want to maintain the property of having a working controller, we'll need to hold this PR anyway, until the direct controller is implemented.

apis/apigee/v1alpha1/envgroupattachment_types.go
jasonvigil: Curious why would we wait to add this field? The API is already not compatible with the TF controller.

apis/apigee/v1alpha1/environment_reference.go
jasonvigil: Actually I just did this in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3639

apis/apigee/v1alpha1/environment_reference.go
```diff
@@ -0,0 +1,103 @@
...
+	u := &unstructured.Unstructured{}
+	u.SetGroupVersionKind(EnvironmentGVK)
+	if err := reader.Get(ctx, key, u); err != nil {
+		if apierrors.IsNotFound(err) {
+			return "", k8s.NewReferenceNotFoundError(u.GroupVersionKind(), key)
+		}
+		return "", fmt.Errorf("reading referenced %s %s: %w", EnvironmentGVK, key, err)
+	}
+	// Get external from status.externalRef. This is the most trustworthy place.
+	actualExternalRef, found, err := unstructured.NestedString(u.Object, "status", "externalRef")
```
jasonvigil, line 77: Just realized this will not work actually, there is no status.externalRef field in the existing ApigeeEnvironment type. I created a PR to define types/refs for ApigeeEnvironment (and ApigeeOrganization), you will need to rebase to use updated code.

acpana: Yup that's why I have my TODO lower. There's been chatter about how to fix this in a separate PR (for a different resource). I don't think we should block on this for now. TL;DR - for TF or DCL resources this field will not be populated. But I think we have a path forward.

Can we close on this PR so I can focus on the TODO (and the remaining work items) ?

yuwenma (commented):
/approve
/assign @jasonvigil 

Defer to Jason for the final vote.

PR #3607, opened 2025-02-05
dev:tools: mockgcp, api contexts

acpana:
A bit of a rougher PR:

- adds context around mockgcp files (add separates this addition from other files)
- ditto for api files

A lot of this work revolves still around prompt engineering but I figured it's worth showcasing it and adding in some of the changes I made. We can continue down this road a bit by adding "notable examples" beyond just raw numbers for mockgco or apis that we think are the gold standard.

Also check out the rules. We can enforce some of our codestyle in prompt. Another direction is to run the crds_test.go or other important linters and make the codebot iterate on its output accordingly.


acpana:
this is now superseded I think

PR #3608, opened 2025-02-05
tool: fix the `types.go` template

jingyih:
Not sure if there was a merge error, but I noticed this while generating the new resources.

### DIFF
--- a/apis/iap/v1alpha1/iapsettings_types.go
+++ b/apis/iap/v1alpha1/iapsettings_types.go
@@ -70,9 +70,8 @@ type IAPSettingsStatus struct {
 	// ObservedState *IAPSettingsObservedState `json:"observedState,omitempty"`
 }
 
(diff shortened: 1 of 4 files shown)


yuwenma (commented):
/lgtm
/approve

Suggest to add a flag to skip adding the `observedState`. Some resources have specialties and this could cause confusions

jingyih:
> Suggest to add a flag to skip adding the `observedState`. Some resources have specialties and this could cause confusions

From my experience, I typically decide whether to keep or remove `observedState` AFTER the types have been generated. Instead of using a flag, it might be simpler to just delete or comment out `observedState` afterward. Using a flag complicates the workflow: first generate the types -> realize `observedState` isn't needed -> manually delete the generated files and regenerate them.

jingyih:
Rebased to resolve a merge conflict.

PR #3617, opened 2025-02-05
codebot: do KCC specific tasks like a developer

yuwenma:
Let codebot to do some contributor work like  `go test ./tests/apichecks` `make ready-pr`, and it will try to fix from the result.

### DIFF
--- a/dev/tools/controllerbuilder/pkg/codebot/tool_run_cmd.go
+++ b/dev/tools/controllerbuilder/pkg/codebot/tool_run_cmd.go
@@ -69,13 +69,21 @@ func (t *RunTerminalCommand) Run(ctx context.Context, c *Chat, args map[string]a
 	cmd.Dir = c.baseDir
 	cmd.Args = append(tokens, strings.Split(t.Args, " ")...)
 
-	stdout, stderr := cmd.CombinedOutput()
-	if stderr != nil {
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+
+		if exitError, ok := err.(*exec.ExitError); ok {
+			result.Error = fmt.Sprintf("command failed with exit code %d: %s", exitError.ExitCode(), string(exitError.Stderr))
+		} else {
+			result.Error = fmt.Sprintf("command failed: %s", err.Error())
+		}
 		result.Success = false
-		result.Error = stderr.Error()
+	} else {
+		result.Success = true
 	}
-	result.Success = true
-	result.Output = string(stdout)
+	if len(output) > 0 {
+		result.Output = string(output)
+	}
 	return result, nil
 }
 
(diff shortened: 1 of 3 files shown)


dev/tools/controllerbuilder/pkg/codebot/tools.go
```diff
@@ -70,6 +70,13 @@ func (c *Chat) runFunctionCall(ctx context.Context, functionCall llm.FunctionCal
 			Response: result,
 			Error:    err,
 		}, nil
+	case "RunTerminalCommand":
```
jingyih, line 73: Oh this is nice! We already have a lot of tools that can be used by the codebot.

dev/tools/controllerbuilder/pkg/codebot/tool_run_cmd.go
```diff
@@ -0,0 +1,117 @@
...
+		cmd = exec.CommandContext(ctx, "bash", "-c", t.Command)
+	}
+	cmd.Dir = c.baseDir
+	cmd.Args = append(tokens, strings.Split(t.Args, " ")...)
+
+	stdout, stderr := cmd.CombinedOutput()
+	if stderr != nil {
+		result.Success = false
+		result.Error = stderr.Error()
+	}
```
justinsb, line 76: This probably isn't what you mean, it always falls through

go.mod
```diff
@@ -82,6 +82,7 @@ require (
 	k8s.io/apiextensions-apiserver v0.27.9
 	k8s.io/apimachinery v0.27.11
 	k8s.io/client-go v0.27.11
+	k8s.io/klog v1.0.0
```
justinsb, line 85: Odd ... we shouldn't be depending on klog v1

dev/tools/controllerbuilder/pkg/codebot/tools.go
yuwenma: This helped me catch and (kind of) fixed https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3613/commits/0846f87c4c1d421f1c1c10159bb67250ee73a0d5

go.mod
yuwenma: good catch. I have some local file not checked in to this PR. That may be why this lib is introduced.

dev/tools/controllerbuilder/pkg/codebot/tool_run_cmd.go
yuwenma: Ah! You are right. I messed it with the other error when tracing how this tool function is called.

dev/tools/controllerbuilder/pkg/codebot/tools.go
jingyih: If the codebot is running the same command line tools as human, how did it catch and fix it?

justinsb (approved):

dev/tools/controllerbuilder/pkg/codebot/tools.go
yuwenma: From the response `RunTerminalCommandResults`, we tell gemini the stdout/sterr which is the input to the next chat message.

Issue #3620, opened 2025-02-06
management-conflict-prevention-policy is not supported in pure direct resources


maqiuyujoyce:
### Checklist

- [x] I did not find a related open enhancement request.
- [x] I understand that enhancement requests filed in the GitHub repository are by default low priority.
- [x] If this request is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Describe the feature or resource

Looks like [resource conflict management](https://cloud.google.com/config-connector/docs/concepts/managing-conflicts) is not supported for pure direct resources.

We probably should wait and see if it is still a feature gap for Config Connector resource management based on user feedback after we have more direct resources.

ldanielmadariaga:
This item has been open for >12 months. As part of scheduled repository maintenance, this Issue/PR will be closed. If this is still a critical ask, please re-open and work with your Google Cloud Account team to prioritize it with GCP KCC Product Management.

PR #3621, opened 2025-02-06
Switch AlloyDBInstance to pure direct resource

maqiuyujoyce:
### Change description

Switched AlloyDBInstance to pure direct resource and updated test cases to reflect the change.

Noticed that we actually don't support resource conflict management so filed an issue #3620.

### Tests you have done

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/alloydb/v1beta1/alloydbcluster/basicsecondaryalloydbcluster/_http.log
+++ b/pkg/test/resourcefixture/testdata/basic/alloydb/v1beta1/alloydbcluster/basicsecondaryalloydbcluster/_http.log
@@ -765,9 +765,10 @@ X-Xss-Protection: 0
 
 ---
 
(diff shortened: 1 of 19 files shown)


pkg/test/resourcefixture/testdata/basic/alloydb/v1beta1/alloydbcluster/basicsecondaryalloydbcluster/_http.log
```diff
@@ -913,73 +916,6 @@ X-Xss-Protection: 0
 
 ---
 
-GET https://alloydb.googleapis.com/v1beta/projects/${projectId}/locations/us-west1/clusters/alloydbcluster-1-${uniqueId}/instances/alloydbinstance-${uniqueId}?alt=json
```
maqiuyujoyce, line 916: This GET is no longer there because in the TF-based controller, a GET of the target resource is always done after an operation is completed while in the direct controller we don't do it.

jasonvigil (commented):
/lgtm

PR #3622, opened 2025-02-06
feat: Enable direct reconciler for SQLInstance

jasonvigil:
No longer fall back to TF-based reconciler for resources that do not have spec.cloneSource set. Instead, use direct reconciler for all resources.

This will help with maintenance / debugging issues going forward. Now, it is simpler to reason about which controller is active, and which code path is being used.

Mockgcp http log changes are due to slight behavior differences between direct controller and terraform controller, but (I believe) should not actually cause any functional differences. 

As part of this change, I also removed SQLInstance from the list of resources that support state-into-spec-merge. This is because direct controllers do not support this feature.

Also, disabled VCR test for sqluser resource. We can use the real/mock gcp log based tests instead.


pkg/controller/direct/sql/sqlinstance_controller.go
yuwenma, line 1: is there any disadvantage on keeping this code? Switching to default direct should only change the tf2crd annotation.

jasonvigil: This code path is now dead. The only purpose of having the reconcile gate is to choose between using the direct controller or a legacy controller. Now that we no longer support the legacy controller, this code should be removed.

jasonvigil:
> We should remove those test suites if they are a duplication to the `*-direct` tests. It is confusing (and risky) that switching to direct involves controller change, and many golden log changes.

Ok, removed the duplicate tests. Some tests (which did not previously have duplicates that enabled direct controller) still show some diffs / changes. These tests are from other suites, for resources that depend on SQLInstance.

pkg/controller/direct/sql/sqlinstance_controller.go
yuwenma: In general it is better to not change the controller logic because it is a reverse of what opt-in trying to do: test the controller in production before fully switch to the default. When changing the reconcile gate, the controller is back to the not fully tested stage. Can we leave the code as it is if there's no obvious disadvantage?

yuwenma (commented):
/lgtm
/approve
/hold

My main concern is about changing the controller code (which put the sqlinstance back to the situation that "the controller is not fully tested in prod before switching to default") . But I'll defer it to you.

pkg/controller/direct/sql/sqlinstance_controller.go
jasonvigil: The reconcile gate will not affect controller logic. Removing it is part of disabling the terraform controller. I think leaving it in is more risky, because it now means the SQLInstance controller will potentially have different behavior than every other direct resource when deciding which reconciler to use.

jasonvigil:
@yuwenma I performed some manual tests to double-check this PR works as intended.

1. Deploy current (master branch) KCC controller, and current version of SQLInstance CRD (with tf2crd label present).
2. Apply some test resources:

A basic SQLInstance (using TF-based controller):
```yaml
---
apiVersion: sql.cnrm.cloud.google.com/v1beta1
kind: SQLInstance
metadata:
  name: test-basic-tf
spec:
  databaseVersion: POSTGRES_16
  region: us-west4
  rootPassword:
    valueFrom:
      secretKeyRef:
        key: password
        name: sqlinstance-password
  settings:
    tier: db-custom-1-3840
```

A basic SQLInstance (using direct controller, via opt-in annotation):
```yaml
---
apiVersion: sql.cnrm.cloud.google.com/v1beta1
kind: SQLInstance
metadata:
  name: test-basic-direct
  annotations:
spec:
  databaseVersion: POSTGRES_16
  region: us-west4
  rootPassword:
    valueFrom:
      secretKeyRef:
        key: password
        name: sqlinstance-password
  settings:
    tier: db-custom-1-3840
```

A "source" SQLInstance (for testing clone feature, this one will use TF-based controller):
```yaml
---
apiVersion: sql.cnrm.cloud.google.com/v1beta1
kind: SQLInstance
metadata:
  name: test-source
spec:
  databaseVersion: POSTGRES_16
  region: us-west4
  rootPassword:
    valueFrom:
      secretKeyRef:
        key: password
        name: sqlinstance-password
  settings:
    backupConfiguration:
      enabled: true
      pointInTimeRecoveryEnabled: true
    tier: db-custom-1-3840
```

At this point, I added some data into the `test-source` instance.

A "clone" SQLInstance (for testing clone feature, this one will use direct controller due to reconcile gate):
```yaml
---
apiVersion: sql.cnrm.cloud.google.com/v1beta1
kind: SQLInstance
metadata:
  name: test-clone
spec:
  cloneSource:
    sqlInstanceRef:
      name: test-source
  settings:
    backupConfiguration:
      enabled: true
      pointInTimeRecoveryEnabled: true
    tier: db-custom-1-3840
```

Here, I verified the data from `test-source` was loaded into the `test-clone`.

3. Apply new version of SQLInstance CRD (with tf2crd label removed).
4. Upgrade KCC controller to the version in this PR.
5. Verify all previously created SQLInstances were adopted successfully, and are now all using the direct controller for reconciliation.

Based on this test (and all of our other resourcefixture coverage for SQLInstance), I feel reasonably confident in switching over to use the SQLInstance direct controller for all resources.

jasonvigil:
/hold cancel

Discussed offline.

yuwenma:
Umbrella issue: https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/4007

PR #3623, opened 2025-02-06
Update the instruction for the migration scenario

maqiuyujoyce:
### Change description

Reformatted the file to be more readable.
Updated the file according to https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3598#issuecomment-2638638411.

### Tests you have done

### DIFF
--- a/docs/develop-resources/scenarios/migrate-tf-resource-beta.md
+++ b/docs/develop-resources/scenarios/migrate-tf-resource-beta.md
@@ -1,88 +1,159 @@
 # Migrate Terraform/DCL-based Beta to direct Beta
- 
- Config Connector requires the Beta resource to be backward compatible. Thus, we have a higher bar on migrating the Beta resource compared to migrating an Alpha resource. 
- - In the MockGCP step, we require the Mock GCP to reflect the real GCP cases where the current Terraform or DCL based approach uses or depend on to make sure the migration is backward compatible (i.e. assigning defaulting value to a field, same HTTP errors for matching, etc)
- - In the API and mapper step, we require the migration to happen in two stages: the first stage is to make sure all APIs and their behavior are exactly the same after migration (verifying via the golden logs). The second stage is to add new fields to make the resource up-to-date with what the current GCP service supports.
- - In the controller step, we do not require the direct controller and the Terraform/DCL based controller to behave exactly the same, but fixing the legacy or non Kubernetes-native problems like merging config back to `spec` (`state-into-spec:merge`), setting defaulting on-behalf of users. 
 
-To help the developer better manage the migration, we design the following steps to be self-validating and not restricted by the Config Connector release cycles. In other words, it is safe for a migration to happen across multiple Config Connector releases.
+Config Connector requires the Beta resource to be backward compatible. Thus, we
+have a higher bar on migrating the Beta resource compared to migrating an Alpha
+resource. - In the MockGCP step, we require the Mock GCP to reflect the real GCP
+cases where the current Terraform or DCL based approach uses or depend on to
+make sure the migration is backward compatible (i.e. assigning defaulting value
+to a field, same HTTP errors for matching, etc) - In the API and mapper step, we
+require the migration to happen in two stages: the first stage is to make sure
+all APIs and their behavior are exactly the same after migration (verifying via
+the golden logs). The second stage is to add new fields to make the resource
+up-to-date with what the current GCP service supports. - In the controller step,
+we do not require the direct controller and the Terraform/DCL based controller
+to behave exactly the same, but fixing the legacy or non Kubernetes-native
+problems like merging config back to `spec` (`state-into-spec:merge`), setting
+defaulting on-behalf of users.
+
+To help the developer better manage the migration, we design the following steps
+to be self-validating and not restricted by the Config Connector release cycles.
+In other words, it is safe for a migration to happen across multiple Config
+Connector releases.
 
 ## Add MockGCP tests
 
-Follow [deep-dives Step 1](../deep-dives/1-add-mockgcp-tests.md). We use the Terraform or DCL based controller to help develop the Mock GCP.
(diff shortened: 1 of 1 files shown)


docs/develop-resources/scenarios/migrate-tf-resource-beta.md
```diff
@@ -1,88 +1,159 @@
...
+up-to-date with what the current GCP service supports. - In the controller step,
+we do not require the direct controller and the Terraform/DCL based controller
+to behave exactly the same, but fixing the legacy or non Kubernetes-native
+problems like merging config back to `spec` (`state-into-spec:merge`), setting
+defaulting on-behalf of users.
+
+To help the developer better manage the migration, we design the following steps
+to be self-validating and not restricted by the Config Connector release cycles.
+In other words, it is safe for a migration to happen across multiple Config
+Connector releases.
```
acpana, line 21: I would rewrite this a bit to say:

```suggestion
Config Connector requires the Beta resource to be backward compatible.  We require the Mock GCP to reflect the real GCP
cases where the current Terraform or DCL based approach uses or depend on to
make sure the migration is backward compatible. Example things to look out for:

* assigning defaulting value
to a field
* having thesame HTTP errors for matching

In the API and mapper step, we
require the migration to happen in two stages: 

- the first stage is to make sure all APIs and their behavior are exactly the same after migration (verifying via
the golden logs)
- The second stage is to add new fields to make the resource up-to-date with what the current GCP service supports. 

In the controller step, we do not require the direct controller and the Terraform/DCL based controller
to behave exactly the same, but fixing the legacy or non Kubernetes-native
problems like merging config back to `spec` (`state-into-spec:merge`), setting
defaulting on-behalf of users.

To help the developer better manage the migration, we design the following steps
to be self-validating and not restricted by the Config Connector release cycles.
In other words, it is safe for a migration to happen across multiple Config
Connector releases.
```

docs/develop-resources/scenarios/migrate-tf-resource-beta.md
```diff
@@ -1,88 +1,159 @@
...
+    conventions). You can run `dev/tasks/generate-crds` (repeatedly) to make
+    sure the CRD are the same. Comment changes to match the latest protobuf are
+    recommended as long as the content doesn't contain breaking changes. An
+    example of a breaking change would be the comment explicitly says a
+    previously optional field `required`.
 
-* Add `cnrm.cloud.google.com/dcl2crd: "true"` or `cnrm.cloud.google.com/tf2crd: "true"` to the API tag [example](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/0bbac86ace6ab2f4051b574f026d5fe47fa05b75/pkg/controller/direct/redis/cluster/roundtrip_test.go#L92), to continue using DCL-based or Terraform-based controllers.
+*   Add `cnrm.cloud.google.com/dcl2crd: "true"` or
+    `cnrm.cloud.google.com/tf2crd: "true"` to the API tag
+    [example](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/0bbac86ace6ab2f4051b574f026d5fe47fa05b75/pkg/controller/direct/redis/cluster/roundtrip_test.go#L92),
```
acpana, line 81: this is not the right example line right?

docs/develop-resources/scenarios/migrate-tf-resource-beta.md
```diff
@@ -1,88 +1,159 @@
...
+    to continue using DCL-based or Terraform-based controllers.
 
-* You may see some new fields added to the CRD. These are expected since the Terraform/DCL based resources could be out of date (and users are looking forward to these new fields!). You **shall comment out** those new fields using `/*NOTYET ..*/` in this PR if they are not supported in the Terraform-based controller yet (we will add them later).
+*   You may see some new fields added to the CRD. These are expected since the
+    Terraform/DCL based resources could be out of date (and users are looking
+    forward to these new fields!). You **shall comment out** those new fields
+    using `/*NOTYET ..*/` in this PR if they are not supported in the
+    Terraform-based controller yet (we will add them later).
```
acpana, line 88: I think some changes here and in other places are just formatting right?

acpana (commented):
I think overall this LGTM! Just a few questions; will defer to Yuwen for final approval for the instructions.

/lgtm

docs/develop-resources/scenarios/migrate-tf-resource-beta.md
maqiuyujoyce: Thank you for the rewrite suggestions. I didn't make any changes to this part of the doc so will copy your change over.

docs/develop-resources/scenarios/migrate-tf-resource-beta.md
maqiuyujoyce: Nope. Will find a correct example.

docs/develop-resources/scenarios/migrate-tf-resource-beta.md
maqiuyujoyce: Yes, the change at most places are about formatting. The only real change is:
```
... Comment changes to match the latest protobuf are
recommended as long as the content doesn't contain breaking changes. An
example of a breaking change would be the comment explicitly says a
previously optional field `required`.
```

docs/develop-resources/scenarios/migrate-tf-resource-beta.md
maqiuyujoyce: Done.

docs/develop-resources/scenarios/migrate-tf-resource-beta.md
maqiuyujoyce: Done.

acpana:
/lgtm 💯 

cc @yuwenma  for approval 🙏🏼

jingyih (commented):
I'm not sure if we relaxed some of the requirements, but let's go ahead and merge this first.

/lgtm
/approve

PR #3624, opened 2025-02-06
Support acronyms in type names during type generation

maqiuyujoyce:
### Change description

Extracted out acronym related work from #3598 so that:
1. More acronyms are supported.
2. Type names also support acronyms.

### Tests you have done

Verified by running the type generator.


dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
```diff
@@ -362,6 +363,22 @@ func GoNameForProtoMessage(msg protoreflect.MessageDescriptor) string {
...
+			if IsAcronym(token) {
+				token = strings.ToUpper(token)
+			} else {
+				token = strings.Title(token)
+			}
+			tokens[j] = token
+		}
+		parts[i] = strings.Join(tokens, "")
+	}
+	fullName = strings.Join(parts, ".")
```
jingyih, line 381: nit: it seems the last 3 lines can be combined into `return strings.Join(parts, "_")`.

jingyih (commented):
/lgtm

dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
maqiuyujoyce: Done.

jingyih (commented):
/lgtm

dev/tools/controllerbuilder/pkg/codegen/common.go
```diff
@@ -39,7 +39,8 @@ var protoMessagesNotMappedToGoStruct = map[string]string{
 
 var Acronyms = []string{
 	"ID", "HTML", "URL", "URI", "HTTP", "HTTPS", "SSH",
-	"IP", "GB", "FS", "PD", "KMS", "GCE", "VTPM",
+	"IP", "GB", "FS", "PD", "KMS", "GCE", "VTPM", "SSL",
+	"CA", "CPU", "SQL", "PSC", "DNS",
```
yuwenma, line 43: I'm a little bit lost. I think we have reviewed this code and discussed the potential risk about CA. Why do we have a separate PR?

maqiuyujoyce: This PR is extracted out from https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3598 where I included a lot of more functionality in it.

Original comment is at https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3598#discussion_r1941378023
I personally don't have any concern about it. @jingyih any thoughts?

jingyih: I noticed a discrepancy between this and our test. I suggest using the same source of truth for acronyms. If there's any uncertainty about an acronym, we can check which field is currently failing to help our decision.

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/ad509808d96b5ff98bd12a78089019c8d8a2128b/tests/apichecks/crds_test.go#L166

maqiuyujoyce: Good point. I think both lists contain special acronyms the other list doesn't have so I'm going to create a union of them.

maqiuyujoyce: Done.

tests/apichecks/testdata/exceptions/acronyms.txt
```diff
@@ -33,7 +33,11 @@
 [acronyms] crd=appengineflexibleappversions.appengine.cnrm.cloud.google.com version=v1alpha1: field ".spec.endpointsApiService.name" should be ".spec.endpointsAPIService.name"
 [acronyms] crd=appengineflexibleappversions.appengine.cnrm.cloud.google.com version=v1alpha1: field ".spec.endpointsApiService.rolloutStrategy" should be ".spec.endpointsAPIService.rolloutStrategy"
 [acronyms] crd=appengineflexibleappversions.appengine.cnrm.cloud.google.com version=v1alpha1: field ".spec.handlers[].redirectHttpResponseCode" should be ".spec.handlers[].redirectHTTPResponseCode"
+[acronyms] crd=appengineflexibleappversions.appengine.cnrm.cloud.google.com version=v1alpha1: field ".spec.resources.diskGb" should be ".spec.resources.diskGB"
```
jingyih, line 36: /cc @jasonvigil Can we fix these while the resources are still alpha?

jasonvigil: +1, agreed that we should fix these API issues as long as we haven't already released the v1beta1 API (and therefore need to maintain backwards compatibility and cannot change the API). For Apigee (and other resources), we have been making these fixes during the migrations to direct controllers. Curious though why did you tag me on this? I am not working on appengine currently (as far as I know)?

tests/apichecks/testdata/exceptions/acronyms.txt
```diff
@@ -116,13 +130,17 @@
 [acronyms] crd=computedisks.compute.cnrm.cloud.google.com version=v1beta1: field ".status.sourceImageId" should be ".status.sourceImageID"
 [acronyms] crd=computedisks.compute.cnrm.cloud.google.com version=v1beta1: field ".status.sourceSnapshotId" should be ".status.sourceSnapshotID"
 [acronyms] crd=computefirewallpolicies.compute.cnrm.cloud.google.com version=v1beta1: field ".status.selfLinkWithId" should be ".status.selfLinkWithID"
+[acronyms] crd=computeforwardingrules.compute.cnrm.cloud.google.com version=v1beta1: field ".spec.allowPscGlobalAccess" should be ".spec.allowPSCGlobalAccess"
```
jingyih, line 133: I am not sure which acronyms should be included, so I asked Gemini which is better `allowPscGlobalAccess` vs. `allowPSCGlobalAccess`. Gemini prefers `allowPscGlobalAccess`.

```
There isn't a strict, formalized rule in Kubernetes API conventions or Go naming guidelines that mandates a specific approach for acronyms in all cases. However, the general preference leans towards improved readability and consistency, which often leads to treating acronyms as separate words, particularly if they're not universally known.

Here's a more nuanced view:

Short, Very Common Acronyms (Exceptions): Acronyms that are extremely common and well-known within the context of Kubernetes or a specific domain can sometimes be used without separation. Examples might include:

API: apiVersion (very common, usually treated as a single word)

ID: resourceID (very common)

URL: webhookURL (also quite common)

However, even with these, you'll often see ApiVersion, ResourceId, and WebhookUrl used as well.

Less Common or Longer Acronyms (Separate Words): For acronyms that are:

Not universally understood

Longer (more than 2-3 letters)

Specific to a niche domain

It's generally better to treat them as separate words for readability. That's why allowPscGlobalAccess is preferred over allowPSCGlobalAccess.

Context Matters: As previously mentioned, the context where the acronym is used is crucial. If the acronym is highly relevant to your CRD, you can treat it as separate words.

Go Struct Tags: When defining your Go structs for your CRD, the struct tags (used for JSON serialization) also influence the final names that appear in your YAML or JSON resources. Consider this when deciding on your field names and struct tags.

Example Scenarios:

Good (using separate words): privateServiceConnectEndpoint, pscEndpointAddress (if PSC is less universally known)

Acceptable (common acronym): webhookURL (if "URL" is obviously URL)

Less Ideal (especially if acronym is uncommon): privateServiceConnectEndpointAddress (too long and potentially hard to read)

Key Principles to Remember:

Readability First: Always prioritize the readability and understandability of your field names for the users of your CRD.

Consistency: Strive for consistency in your naming conventions throughout your CRD. If you use one style for acronyms in one place, use it consistently.

Clarity: If there's any doubt about whether an acronym is well-known enough, err on the side of spelling it out or treating it as separate words.

Kubernetes/Go Style: Follow Kubernetes/Go common practices as much as possible to integrate smoothly with the ecosystem.

In summary, while there's no hard rule against using combined acronyms, the general trend in Kubernetes API design and Go is to prioritize readability and treat acronyms as separate words, especially if they are not universally recognized or if they make the field name significantly less readable. Using allowPscGlobalAccess is generally the better practice in your case.
```

jasonvigil: The Google go style guide prefers all caps for exported acronyms: https://google.github.io/styleguide/go/decisions#initialisms

tests/apichecks/testdata/exceptions/acronyms.txt
jingyih: Oops, I totally misread that and thought these belong to Apigee...

jasonvigil: Ah no worries! Yeah we are trying to fix the apigee ones, but unfortunately a couple of the resources were already released v1beta1, so we cannot change at this point

tests/apichecks/testdata/exceptions/acronyms.txt
maqiuyujoyce: I googled "google cloud psc" and "gcp psc", the top hits only contains `PSC`s (all capitalized). (Not sure how much the result is customized for me though.) I think it should not be super confusing for us to just use `PSC` in the field names and kind names.

maqiuyujoyce:
@yuwenma @jingyih do you have any further comments about this PR?

Regarding the acronyms in the list, they are picked based on public branding, common acronyms and legacy decisions. And I think we should just capitalize all of them.

jingyih (commented):
I'm fine with merging this. We can always revisit the acronyms later if we find they're negatively affecting readability. 

/lgtm

PR #3625, opened 2025-02-06
test: generate fuzzer for ManagedKafkaTopic

jingyih:
Had to manually fix this resource because it doesn't have "ObservedState." Hopefully, the results improve as more similar examples are added.

I also updated the documentation since the tool was enhanced to prompt with the matching go type, making the "// krm.kind" marker no longer necessary.


jingyih:
/hold

Trying a different idea.

jingyih:
Superseded by #3661

/close

PR #3633, opened 2025-02-07
feat: Add ApigeeInstanceAttachment v1alpha1 reconciler

jasonvigil:


pkg/controller/direct/apigee/instanceattachment_controller.go
```diff
@@ -0,0 +1,297 @@
...
+				return false, nil
+			}
+			return false, fmt.Errorf("getting ApigeeInstanceAttachment %q: %w", a.id, err)
+		}
+		a.actual = attachment
+		return true, nil
+	} else {
+		// Otherwise, we must list all the attachments for the instance, and search for the one that contains the specified environment.
+		// This code path will only run on adoption, and will not run after the resource has already been reconciled.
+		// Therefore, it is ok (and necessary) to attempt to resolve the ApigeeEnvironment reference here.
```
yuwenma, line 115: For "adoption", do you mean the "acquire" mode?  Is this because the  attachment and environment does not have a relationship built in the URL?

pkg/controller/direct/apigee/instanceattachment_controller.go
```diff
@@ -0,0 +1,297 @@
...
+		if err := ResolveApigeeInstanceAttachmentRefs(ctx, a.k8sClient, a.desired); err != nil {
+			return false, err
+		}
+
+		// HACK: Environment required by GCP API is name-only, not fully-qualified ID.
+		// So, we fix this by editing the environment to be name-only.
+		environmentID := &apigeev1beta1.ApigeeEnvironmentIdentity{}
```
yuwenma, line 122: nit: suggest change to a shared function among Find, Create and Delete

pkg/controller/direct/apigee/instanceattachment_controller.go
```diff
@@ -0,0 +1,297 @@
...
+// Update updates the resource in GCP based on `spec` and update the Config Connector object `status` based on the GCP response.
+func (a *ApigeeInstanceAttachmentAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOperation) error {
+	log := klog.FromContext(ctx)
+	log.V(2).Info("updating ApigeeInstanceAttachment", "id", a.id)
+	mapCtx := &direct.MapContext{}
+
+	// There are no fields in the GCP ApigeeInstanceAttachment API that can be updated.
+	// So, we will only update the KRM status.
+	status := &krm.ApigeeInstanceAttachmentStatus{}
+	status.ObservedState = ApigeeInstanceAttachmentObservedState_FromAPI(mapCtx, a.actual)
```
yuwenma, line 232: Some future improvements: use etag or a snapshot to decide if we can skip status update.

pkg/controller/direct/apigee/instanceattachment_controller.go
jasonvigil: Hmm, it already uses a shared function: `environmentID.FromExternal()`. Do you mean wrapping that in another shared function? I am not sure really how useful that will be. This logic is only required in two places, find and create. Export uses a different logic.

pkg/controller/direct/apigee/instanceattachment_controller.go
jasonvigil: I updated the comment to use the correct verb:
"adopt" -> "acquire"

pkg/controller/direct/apigee/instanceattachment_controller.go
jasonvigil: I created a helper function that gets the name of the environment.

yuwenma (commented):
/lgtm
/approve

Thanks! great job!

PR #3634, opened 2025-02-07
chore: combine two fuzz presubmits

jingyih:
Just realized we have two presubmits for the fuzzer. Should we combine them to avoid confusion?  (This PR picked the workflow over the presubmit, but we could go the other way instead)

Also, we should migrate individual fuzz tests to use the fuzztesting library (with help from LLM).


jingyih:
/assign @acpana
I'd love to hear your thoughts!

yuwenma (commented):
/lgtm
/approve

acpana:
Awesome work jingyi!

I think this fixes https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/3102

PR #3636, opened 2025-02-07
tests:fix: native type arrays

acpana:
Correctly address terminal array fields (they will look like spec.foo.bar.xyz[])

addresses https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/3571

### DIFF
--- a/tests/apichecks/crds_test.go
+++ b/tests/apichecks/crds_test.go
@@ -468,7 +468,7 @@ func TestCRDFieldPresenceInUnstructured(t *testing.T) {
 				// Check if field exists in any unstructured object
 				missing := true
 				for _, obj := range unstructs {
-					if hasField(obj.Object, fieldPath) {
+					if hasField(obj.Object, strings.TrimSuffix(fieldPath, "[]")) {
 						missing = false
 						break
 					}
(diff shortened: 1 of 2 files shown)


jasonvigil:
Hmm, this does not appear to work for arrays of objects.

For example, looks like it still throws an error for
```
[missing_field] crd=workstations.workstations.cnrm.cloud.google.com version=v1beta1: field ".spec.annotations[].value" is not set in unstructured objects
```
, even though that field is being set in https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/test/resourcefixture/testdata/basic/workstations/workstation/workstation-full/create.yaml

Is there any way to implement a more general fix for arrays of all types? Not just native type arrays. IMO, this does not actually fix the original issue.

acpana:
> For example, looks like it still throws an error for

Good catch! I think we can keep the issue open as we investigate that case.

I think this is a good fix for native types for now and we can follow up for other array cases.

jingyih:
Oh, I just saw this PR. I sent #3651 to unblock my other PRs, and I believe it also fixes the "arrays of objects" issue.

acpana:
Closing in favor of https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3651

Good job Jingyi !!

PR #3642, opened 2025-02-07
chore: run generate-controller anywhere

yuwenma:
with this PR I can run `generate-controller` outside of the repo.

### DIFF
--- a/dev/tools/controllerbuilder/scaffold/controller.go
+++ b/dev/tools/controllerbuilder/scaffold/controller.go
@@ -23,7 +23,6 @@ import (
 	"go/token"
 	"os"
 	"path/filepath"
-	"runtime/debug"
 	"strings"
 	"text/template"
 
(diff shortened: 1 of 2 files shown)


dev/tools/controllerbuilder/scaffold/controller.go
```diff
@@ -42,30 +41,33 @@ var funcMap = template.FuncMap{
...
 
-	// Get main model name
-	bi, ok := debug.ReadBuildInfo()
-	if !ok {
-		return fmt.Errorf("could not read build info")
-	}
-	modelPath := strings.TrimSuffix(bi.Main.Path, currRelPath)
-
-	importPath := filepath.Join(modelPath, directControllerRelPath, service)
+	importPath := filepath.Join("github.com/GoogleCloudPlatform/k8s-config-connector", directControllerRelPath, c.service)
```
yuwenma, line 67: I used a fixed import path because the debug.ReadBuildInfo cannot find the module if running in a different git repo.

justinsb: sgtm.  You can always make it a flag in future (and pass down the Options struct)

justinsb (approved):

PR #3644, opened 2025-02-07
codebot: add support for completion method

justinsb:
This can give more consistent results than the chat method.

PR #3645, opened 2025-02-08
tool: some improvements to `controllerbuilder prompt"

jingyih:
- Make 'LLM model name" configurable"
 - Trim code block marker if the prompt output is code


jingyih:
Ah model is made configurable in #3644. I am going to close this for now.

/close

PR #3646, opened 2025-02-08
chore: add a missing proto annotation

jingyih:
Although this Go struct isn’t auto generated, let’s add the proto annotation to help various LLM tools.


maqiuyujoyce (commented):
/lgtm
/approve

PR #3647, opened 2025-02-08
test: Improve fuzzer test

gemmahou:
### Change description

Currently fuzzer test requires both Spec and ObservedState exist([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/7cd1b21e1d1e790d0c17108eb9b435a80332666a/pkg/fuzztesting/fuzzkrm.go#L72-L73)), and does not cover the Status fields.

This [function](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/7cd1b21e1d1e790d0c17108eb9b435a80332666a/pkg/fuzztesting/fuzzkrm.go#L105) is added for object that only has Spec fields, no Status or ObservedState fields.

For some resources(i.e. CloudIdentityGroup), it's possible that the object has Spec, Status and ObservedState fields, and some resources they only have Spec fields. Improve the fuzzer test to be able to test Spec, Status, and ObservedState Individually. 

### Tests you have done

Verified locally, test failed when I comment out certain fields in mapper.

### DIFF
--- a/pkg/controller/direct/discoveryengine/datastoretargetsite_fuzzer.go
+++ b/pkg/controller/direct/discoveryengine/datastoretargetsite_fuzzer.go
@@ -20,30 +20,39 @@ package discoveryengine
 import (
 	pb "cloud.google.com/go/discoveryengine/apiv1/discoveryenginepb"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/fuzztesting"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 func init() {
-	fuzztesting.RegisterKRMFuzzer(dataStoreTargetSiteFuzzer())
+	fuzztesting.RegisterFuzzer(dataStoreTargetSiteSpecFuzzer().FuzzSpec)
+	fuzztesting.RegisterFuzzer(dataStoreTargetSiteObservedStateFuzzer().FuzzObservedState)
 }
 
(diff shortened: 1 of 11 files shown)


jingyih:
My first question is KCC design related: If we place multiple output-only fields under `.status` for backward compatibility, should we keep all output-only fields there? Or should we split them between `.status` and `.status.observedState`?

If we decide to split the output fields, I suggested in the other [PR](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3603) that we might need to invoke the mapping function for `.status.observedState` within the mapping function for `.status`. Since `observedState` is technically a subfield of `status` and it contains fields that map to proto fields. This way, the fields under `.status.observedState` would be included when we fuzz testing `.status`.

gemmahou:
> My first question is KCC design related: If we place multiple output-only fields under `.status` for backward compatibility, should we keep all output-only fields there? Or should we split them between `.status` and `.status.observedState`?
> 
> If we decide to split the output fields, I suggested in the other [PR](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3603) that we might need to invoke the mapping function for `.status.observedState` within the mapping function for `.status`. Since `observedState` is technically a subfield of `status` and it contains fields that map to proto fields. This way, the fields under `.status.observedState` would be included when we fuzz testing `.status`.

I can double-check on that - for resource migration how do we want to manage fields under `Status` and `.Status.ObservedState`. I'll put this on hold for now. 

/hold

PR #3655, opened 2025-02-11
mockgcp: proxy all the services

justinsb:
Until we get a better approach, having all the services proxies feels easiest.


yuwenma (commented):
/lgtm
/approve

PR #3657, opened 2025-02-11
fix: Remove deprecated SQLInstance failover configuration

jasonvigil:
As of January 13, 2025, the SQLInstance API deprecated the previous method of failover configuration. This fixes the test case to no longer use the deprecated configuration method.

Ref: https://cloud.google.com/sql/docs/mysql/configure-legacy-ha


yuwenma:
/lgtm
/approve

Great catch! Thanks!

PR #3664, opened 2025-02-13
fix: CreateFile to create its parent dir

yuwenma:
otherwise, say I want to create a new file under "mockXXX/testdata/YYY/script.yaml", if  mockXXX or mockXXX/testdata does not exist, the codebot will fail

### DIFF
--- a/dev/tools/controllerbuilder/pkg/codebot/tool_create_file.go
+++ b/dev/tools/controllerbuilder/pkg/codebot/tool_create_file.go
@@ -56,14 +56,22 @@ func (t *CreateFile) Run(ctx context.Context, c *Chat, args map[string]any) (*Cr
 	p := filepath.Join(c.baseDir, t.Filename)
 	if _, err := os.Stat(p); err == nil {
 		return nil, fmt.Errorf("file %q already exists", t.Filename)
+	} else {
+		err = os.MkdirAll(filepath.Dir(p), 0755)
+		if err != nil {
+			return nil, fmt.Errorf("creating dir %s: %w", filepath.Dir(p), err)
+		}
 	}
+	f, err := os.Create(p)
+	if err != nil {
+		return nil, fmt.Errorf("creating file %s: %w", p, err)
+	}
+	defer f.Close()
 
 	if t.Contents == "" {
 		return nil, fmt.Errorf("the contents argument is requiremnt")
 	}
-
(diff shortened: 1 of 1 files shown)


dev/tools/controllerbuilder/pkg/codebot/tool_create_file.go
```diff
@@ -56,14 +56,22 @@ func (t *CreateFile) Run(ctx context.Context, c *Chat, args map[string]any) (*Cr
 	p := filepath.Join(c.baseDir, t.Filename)
 	if _, err := os.Stat(p); err == nil {
 		return nil, fmt.Errorf("file %q already exists", t.Filename)
+	} else {
+		err = os.MkdirAll(filepath.Dir(p), 0755)
```
justinsb, line 60: Nit: `if err := os.MkdirAll(); err != nil {` ..

dev/tools/controllerbuilder/pkg/codebot/tool_create_file.go
```diff
@@ -56,14 +56,22 @@ func (t *CreateFile) Run(ctx context.Context, c *Chat, args map[string]any) (*Cr
...
+		if err != nil {
+			return nil, fmt.Errorf("creating dir %s: %w", filepath.Dir(p), err)
+		}
 	}
+	f, err := os.Create(p)
```
justinsb, line 65: Not sure why you switched from os.WriteFile ?

justinsb (approved):

dev/tools/controllerbuilder/pkg/codebot/tool_create_file.go
yuwenma: That will change the expected behavior of this function. From the description (for codebot), CreateFile should only be called if there is no such file. I also like this check since it provides a way to have codebot self-correct.

dev/tools/controllerbuilder/pkg/codebot/tool_create_file.go
yuwenma: Ah, because I did a little bit more when adding this function  and need to make multiple `WriteString` calls. Bascially, I'm checking if we can validate the LLM generated `t.Contents` before writing to a function. Seems to be more efficient for smaller cases (lint, validation) than sending an entire `go build` call afterwords.

PR #3666, opened 2025-02-13
feat: Refactor Apigee resource identity and reference code

jasonvigil:
Deduplicate and simplify the identity logic. Use the parent resource identity in the child resource identity. Create a general pattern for building new identities from external string format.

When getting the identity of an existing object, use a method defined on the object's type. This is a bit more ergonomic. In order to simplify the code, we no longer check individual fields, so error messages only show if the identity changes overall.

With this change, it's no longer necessary to refer to all parent resources in the spec of a child resource. It's only necessary to refer to the direct parent resource.

Also, updated ApigeeEnvgroupAttachment types to refer only to parent resource, and removed output field of parent resource id.


apis/apigee/v1alpha1/envgroupattachment_types.go
```diff
@@ -61,9 +60,6 @@ type EnvgroupAttachmentObservedState struct {
 	// Output only. The time at which the environment group attachment
 	// was created as milliseconds since epoch.
 	CreatedAt *string `json:"createdAt,omitempty"`
-
-	// Output only. ID of the environment group.
-	EnvironmentGroupID *string `json:"environmentGroupID,omitempty"`
```
acpana, line 66: Are we sure we want to remove this?

https://cloud.google.com/apigee/docs/reference/apis/apigee/rest/v1/organizations.envgroups.attachments#EnvironmentGroupAttachment

And if so, it may be best done as a separate commit/ PR.

jasonvigil: It is the ID of the parent resource, which is provided in spec. Seems redundant to output it. This is the same as outputting the name of the resource, something we previously decided against doing.

Issue #3618, opened 2025-02-05
Empty types should not be generated in type generation


maqiuyujoyce:
### Checklist

- [x] I did not find a related open enhancement request.
- [x] I understand that enhancement requests filed in the GitHub repository are by default low priority.
- [x] If this request is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Describe the feature or resource

When I tried to generate the types for AlloyDBCluster using the following command:
```
go run main.go generate-types \
     --service google.cloud.alloydb.v1beta \
     --api-version "alloydb.cnrm.cloud.google.com/v1beta1" \
     --resource AlloyDBCluster:Cluster \
     --resource AlloyDBInstance:Instance
```
I found that some of the generated types are actually empty, e.g. [`EncryptionInfo`](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/34248c110a178915dfe330deceb337ec6fe7fbf3/apis/alloydb/v1beta1/types.generated.go#L159).

It might be better for us not to write empty types in the generated files.

ldanielmadariaga:
This item has been open for >12 months. As part of scheduled repository maintenance, this Issue/PR will be closed. If this is still a critical ask, please re-open and work with your Google Cloud Account team to prioritize it with GCP KCC Product Management.

PR #3667, opened 2025-02-13
tool: improve types generator

jingyih:
- Avoid generating a Go struct if all its fields are output-only and therefore should already be included in the `ObservedState` Go struct.  
- Improve the logic to correctly detect existing Go structs, as a proto message can now be split into two Go structs.  
- Made a minor adjustment to ensure a leading newline is added appropriately when writing fields.


jingyih:
Fixes #3618

/assign @jasonvigil
/cc @maqiuyujoyce

apis/alloydb/v1beta1/cluster_types.go
```diff
@@ -23,6 +23,68 @@ import (
 
 var AlloyDBClusterGVK = GroupVersion.WithKind("AlloyDBCluster")
 
+// +kcc:proto=google.cloud.alloydb.v1beta.Cluster.PscConfig
+type Cluster_PSCConfig struct {
```
maqiuyujoyce, line 27: Hmmm, I assume the change is made because of the acronym capitalization issue? Can I get an /lgtm for https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3624? With this change, you shouldn't need to do manual modifications.

jingyih: Right. I'm not sure if we reached a consensus on which acronym to include in that PR, so I manually moved the Go structs in this PR to unblock the fix.

dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
```diff
@@ -120,6 +120,11 @@ func (g *TypeGenerator) WriteVisitedMessages() error {
 			continue
 		}
 
+		if hasOnlyOutputFields(msg) {
+			klog.Infof("skipping generation of %q as it only contains output fields", msg.FullName())
```
yuwenma, line 124: I see why you want to do this. But it will cause rabbit holes if the existence of the field itself is configurable. PAM entitlement has one case where a field is structured but has no sub fields . I believe this function will skip that field.

yuwenma (commented):
My gut feeling is that this is going toward a magic machine.  A lot of API logic is determined by the type generator, which itself determines the behavior based on best practice. I'm afraid this limits KCC ability to adjust the field in each resource level. e.g. the PAM entitlement exception.

jingyih:
> My gut feeling is that this is going toward a magic machine. A lot of API logic is determined by the type generator, which itself determines the behavior based on best practice. I'm afraid this limits KCC ability to adjust the field in each resource level. e.g. the PAM entitlement exception.

To correctly generate Go types from proto, I think some kind of magic machine is necessary—either explicitly building one or using an LLM. However, I don’t believe LLMs are yet producing Go types at a high enough quality, meaning fixing the generated code might take more time than just using this tool. Therefore I think there is value in fixing the bugs we found in the tool so it can be used on most resources.

For cases like the PAM entitlement you mentioned, we can decide whether to make the magic machine more complicated or just manually fix the generated code.

PR #3675, opened 2025-02-13
chore: add a dedicated presubmit workflow to validate generated types

jingyih:
#3396 introduced a step to validate the generated Go types for direct resources within `scripts/validate-prereqs.sh`. However, since this script is used by multiple systems, it may lead to unexpected errors. Let's separate the validation into its own script and presubmit workflow.


maqiuyujoyce (commented):
/lgtm
/approve

PR #3676, opened 2025-02-13
fix: Revert cloudidentity proto to match realGCP and add generation commands

gemmahou:
### Change description

In #3591 I changed the kcc proto path from `google/apps` to `mockgcp/cloud`, as I though CloudIdentity is a cloud resource. After checking the openAPI proto and the realGCP log I realized that `google/apps` should be the correct path and consist with other generated protos. I changed the path and added generate commands to file. 

### Tests you have done


jingyih (commented):
/lgtm

PR #3677, opened 2025-02-13
refactor: pretty prints

acpana:
Factors out pretty prints.

### DIFF
--- a/pkg/test/utils.go
+++ b/pkg/test/utils.go
@@ -15,6 +15,7 @@
 package test
 
 import (
+	"encoding/json"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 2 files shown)


pkg/test/utils.go
```diff
@@ -242,3 +243,32 @@ func CompareGoldenObject(t *testing.T, p string, got []byte) {
...
+		t.Fatalf("error encoding to json: %v", err)
+	}
+
+	otherK := new(T)
+	err = json.Unmarshal(encoded, &otherK)
+	if err != nil {
+		t.Fatalf("error decoding from json: %v", err)
+	}
+
+	yEncoded, err := yaml.Marshal(otherK)
```
maqiuyujoyce, line 268: Just curious: Why can't we just do the marshaling here?

acpana: I think JSON serialization as an intermediate step is a good idea for normalization and type safety.

maqiuyujoyce (commented):
/lgtm
/approve

PR #3678, opened 2025-02-14
CloudIdentityMembership - 2. Add the backward compatible API

barney-s:
### Change description

Add APIs to CloudIdentityMembership

### Tests you have done


apis/cloudidentity/v1beta1/membership_types.go
```diff
@@ -0,0 +1,185 @@
...
+
+	// The `MembershipRole`s that apply to the `Membership`. If unspecified, defaults to a single `MembershipRole` with `name` `MEMBER`. Must not contain duplicate `MembershipRole`s with the same `name`.
+	// +required
+	Roles []MembershipRoles `json:"roles"`
+}
+
+// CloudIdentityMembershipStatus defines the config connector machine state of CloudIdentityMembership
+type CloudIdentityMembershipStatus struct {
+	/* Conditions represent the latest available observation of the resource's current state. */
+	Conditions []v1alpha1.Condition `json:"conditions,omitempty"`
```
gemmahou, line 50: nit: replace conditions, ObservedGeneration and ExternalRef with commonv1alpha1.CommonStatus: https://github.com/gemmahou/k8s-config-connector/blob/92d0cc46249a21ae095964dbbf4339df45b2e8c5/apis/cloudidentity/v1beta1/group_types.go#L67.

apis/refs/v1beta1/groupref.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/clients/generated/apis/k8s/v1alpha1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// GroupRef is a reference to a Group resource.
+type GroupRef struct {
```
gemmahou, line 31: Here's the code I'd like to share in our call: https://github.com/gemmahou/k8s-config-connector/blob/92d0cc46249a21ae095964dbbf4339df45b2e8c5/apis/cloudidentity/v1beta1/group_reference.go#L33
We can use the GroupRef and normalizer defined in group_reference.go. And also update the field in group_types, i.e.
GroupRef *GroupRef `json:"groupRef"`

So this file is no longer needed.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_cloudidentitymemberships.cloudidentity.cnrm.cloud.google.com.yaml
```diff
@@ -152,7 +156,6 @@ spec:
                         expireTime:
                           description: The time at which the `MembershipRole` will
                             expire.
-                          format: date-time
```
gemmahou, line 155: @jingyih do you have idea on why this line is missing in CRD? I don't see changes in the expireTime or createTime field... not sure if this is acceptable.

apis/cloudidentity/v1beta1/membership_types.go
```diff
@@ -0,0 +1,185 @@
...
+	/* The ID of the entity. For Google-managed entities, the `id` must be the email address of an existing group or user. For external-identity-mapped entities, the `id` must be a string conforming to the Identity Source's requirements. Must be unique within a `namespace`. */
+	Id *string `json:"id,omitempty"`
+
+	/* The namespace in which the entity exists. If not specified, the `EntityKey` represents a Google-managed entity such as a Google user or a Google Group. If specified, the `EntityKey` represents an external-identity-mapped group. The namespace must correspond to an identity source created in Admin Console and must be in the form of `identitysources/{identity_source_id}`. */
+	Namespace *string `json:"namespace,omitempty"`
+}
+
+type MembershipMemberRestrictionEvaluation struct {
+	/* Output only. The current state of the restriction Possible values: ENCRYPTION_STATE_UNSPECIFIED, UNSUPPORTED_BY_DEVICE, ENCRYPTED, NOT_ENCRYPTED */
+	State *string `json:"state,omitempty"`
```
gemmahou, line 114: @jingyi .spec.Roles.RestrictionEvaluations.State is an output-only field, but it's also a nested field... unsure if we should move this to observedState?

apis/refs/v1beta1/groupref.go
barney-s: I was following the generated code which seemed to align with the project ref and based on that implemented groupref.go

Looking at group_reference.go that you have linked, i see minor differences in what they return.

group_reference.go:NormalizeExternal() returns: 
  externalref:"group/<id>" OR  
  Group{name,namespace}.status.externalRef

groupref.go:ResolveGroup() returns:
  externalref:"<id>" OR
  Group{name,namespace}.spec.resourceID

Is there a preferred approach ?

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_cloudidentitymemberships.cloudidentity.cnrm.cloud.google.com.yaml
jingyih: This should be fine since we're essentially removing a CRD validation. However, to maintain the exact same API, we should probably add `// +kubebuilder:validation:Format=date-time` to the types file.

apis/cloudidentity/v1beta1/types.generated.go
```diff
@@ -0,0 +1,86 @@
...
+
+// +kcc:proto=mockgcp.cloud.cloudidentity.groups.v1beta1.ExpiryDetail
+type ExpiryDetail struct {
+	// The time at which the `MembershipRole` will expire.
+	// +kcc:proto:field=mockgcp.cloud.cloudidentity.groups.v1beta1.ExpiryDetail.expire_time
+	ExpireTime *string `json:"expireTime,omitempty"`
+}
+
+// +kcc:proto=mockgcp.cloud.cloudidentity.groups.v1beta1.Membership
+type Membership struct {
```
jingyih, line 25: Could you rerun the types generation command to clean up this file? We should add to the guide that after manually editing the types, the command needs to be rerun to update the generated code.

apis/refs/v1beta1/groupref.go
jingyih: I believe the current decision is to deprecate the `/refs` folder and move all references to their respective API folders (e.g., `apis/cloudidentity/v1beta1`). So, if `CloudIdentityMembership` needs to reference `CloudIdentityGroup`, we should refer to `apis/cloudidentity/v1beta1/group_reference.go`.

apis/cloudidentity/v1beta1/membership_types.go
```diff
@@ -0,0 +1,178 @@
...
+	// Output only. The time when the `Membership` was last updated.
+	UpdateTime *string `json:"updateTime,omitempty"`
+}
+
+// CloudIdentityMembershipObservedState is the state of the CloudIdentityMembership resource as most recently observed in GCP.
+// +kcc:proto=mockgcp.cloud.cloudidentity.groups.v1beta1.Membership
+type CloudIdentityMembershipObservedState struct {
+	// The name of the Membership resource in GCP. Server generated.
+	// Shall be of the form `groups/{group_id}/memberships/{membership_id}`.
+	Name string `json:"name,omitempty"`
```
jingyih, line 77: Is this field equivalent to `externalRef`? If so, let's remove it. It is probably not clearly documented in the guide, but the latest chat room discussion seems to be leaning in this direction. And if we removed the "name" field, we should just remove the "ObservedState" and place the `// +kcc:proto=mockgcp.cloud.cloudidentity.groups.v1beta1.Membership` annotation to the "Status", it will help the mapper generation.

apis/cloudidentity/v1beta1/membership_types.go
```diff
@@ -0,0 +1,178 @@
...
+	// Immutable.
+	// +required
+	GroupRef refsv1beta1.GroupRef `json:"groupRef"`
+
+	// Immutable. The `EntityKey` of the member. Either `member_key` or `preferred_member_key` must be set when calling MembershipsService.CreateMembership but not both; both shall be set when returned.
+	MemberKey *MembershipMemberKey `json:"memberKey,omitempty"`
+
+	// Immutable. Required. Immutable. The `EntityKey` of the member.
+	// +required
+	PreferredMemberKey MembershipPreferredMemberKey `json:"preferredMemberKey"`
```
jingyih, line 40: nit: when copying and pasting code from the "types.generated.go" file, we should keep the `// +kcc:proto:field` annotations. While they don't affect the final CRD rendering, they are useful for our internal tooling.

apis/cloudidentity/v1beta1/membership_types.go
jingyih: The latest types generator code should support cases where a message contains both configurable and output-only fields. This issue might stem from the API’s proto being converted from OpenAPI using our tooling, which could have resulted in a missing field annotation in the proto file. In such cases, we need to manually replicate what the types generator would typically do—create a new Go struct, `MembershipRolesObservedState`, and move the "RestrictionEvaluations" field into it. However, this could be addressed in a follow-up PR.

apis/cloudidentity/v1beta1/membership_types.go
barney-s: It wasnt in the originally generated file : https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/422c59d957ec31e714681ec9c9f5c965ad3952ac/apis/cloudidentity/v1beta1/membership_types.go

apis/cloudidentity/v1beta1/types.generated.go
barney-s: did you mean this command:
```
cd dev/tools/controllerbuilder ; go run . generate-types      --service mockgcp.cloud.cloudidentity.groups.v1beta1      --resource CloudIdentityMembership:Membership      --api-version 'cloudidentity.cnrm.cloud.google.com/v1beta1'
```

apis/cloudidentity/v1beta1/membership_types.go
barney-s: Its the server generated ID and is called name in the proto. Should'nt we have something similar here ?

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_cloudidentitymemberships.cloudidentity.cnrm.cloud.google.com.yaml
barney-s: done.

apis/cloudidentity/v1beta1/types.generated.go
barney-s: done.

apis/cloudidentity/v1beta1/membership_types.go
barney-s: removed.

dev/tools/controllerbuilder/generate.sh
```diff
@@ -220,5 +220,15 @@ go run . generate-mapper \
      --service google.apps.cloudidentity.groups.v1beta1 \
      --api-version cloudidentity.cnrm.cloud.google.com/v1beta1
 
+# CloudIdentity : Membership
+go run . generate-types \
```
jingyih, line 224: nit: having a separate `generate-types` command for a resource within the same API only works if there is no existing `types.generate.go` file. The intended usage is to add the resource to the existing command, for example:  

```
go run . generate-types \
     --service google.apps.cloudidentity.groups.v1beta1 \
     --resource CloudIdentityGroup:Group \
     --resource CloudIdentityMembership:Membership \
     --api-version "cloudidentity.cnrm.cloud.google.com/v1beta1"
```

Also @gemmahou updated the "service" recently in #3676, so the command we are adding here no longer works with HEAD.

jingyih (commented):
/lgtm

PR #3692, opened 2025-02-16
feat: codebot bash ui & run in another GCP  project.

yuwenma:


dev/tools/controllerbuilder/pkg/codebot/ui/bashui.go
```diff
@@ -0,0 +1,75 @@
...
+func (u *BashUI) Run() error {
+	scanner := bufio.NewScanner(os.Stdin)
+	totalText := []string{}
+	// Read and process lines from stdin
+	lastLine := ""
+	text := ""
+	for scanner.Scan() {
+		line := scanner.Text()
+		if line == "\n" && lastLine == "\n" {
+			totalText = append(totalText, text)
```
justinsb, line 43: Any reason to parse stdin at all in bash mode, vs just sending whatever we `io.ReadAll` ?

justinsb (approved):

justinsb:
One question but LGTM

dev/tools/controllerbuilder/pkg/codebot/ui/bashui.go
yuwenma: Yes, I like the idea about using double empty lines (triple "Enter") to represent a new LLM session message. This uses the similar idea that if the stdin instruction has double empty lines, we split that into different LLM session message.  This is different from starting a new LLM chat session, because it  contains the session history.

yuwenma: Well, yes, I can also `io.ReadAll` for this.

PR #3700, opened 2025-02-18
refactor: rename

acpana:
These fields should be ending in `Ref`!

We can probably write a checker that errs out if the gostruct field ends in Ref but not the CRD field def.


jasonvigil:
/lgtm
/assign yuwenma
nice

apis/apigee/v1alpha1/envgroupattachment_types.go
```diff
@@ -29,10 +29,10 @@ type ApigeeEnvgroupAttachmentSpec struct {
 	OrganizationRef *apigeev1beta1.OrganizationRef `json:"organizationRef"`
 
 	// Immutable. The Apigee environment group which will host the environment.
-	EnvgroupRef *EnvironmentGroupRef `json:"envgroup"`
+	EnvgroupRef *EnvironmentGroupRef `json:"envgroupRef"`
```
yuwenma, line 32: qq: shall this be `envGroupRef`?  Defer to @jasonvigil

acpana: We're not capitalizing the `g` anywhere else 😝 ; including the envgroup resource itself

yuwenma: 😞

acpana: To clarify, I'm happy to take on a capitalization effort for this resource (and others) if that's the right thing to do 🔨

PR #3703, opened 2025-02-18
fix: apigee: wait for ref to be ready

acpana:
Technically we want to wait for an object's reference to be ready before sending actual requests for the object to the cloud provider. The understanding is that those requests will fail bc the reference is not ready.

(probably need to rebase on #3701 once merged)


acpana:
cc @jasonvigil @yuwenma maybe we can add this fix while we wait for the big refactor that jason is doing?

jasonvigil:
Have you observed this as a problem while testing @acpana? My thinking was that `status.projectId` would not be filled in unless the resource was already reconciled.

acpana:
Closing as obsolete (not necessarily not needed anymore)

PR #3704, opened 2025-02-18
fix: wait for ref to be ready

acpana:
Technically we want to wait for an object's reference to be ready before sending actual requests for the object to the cloud provider. The understanding is that those requests will fail bc the reference is not ready.

This PR adds fixes for the "old new" (intermediate) style refs.

(Will need rebasing on #3701)


acpana:
cc @gemmahou @jasonvigil @maqiuyujoyce @yuwenma

ldanielmadariaga:

PR #3706, opened 2025-02-19
CloudIdentityMembership: duplicate output-only field to observed state

gemmahou:
### Change description

Follow up on https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3678

### Tests you have done


config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_cloudidentitymemberships.cloudidentity.cnrm.cloud.google.com.yaml
```diff
@@ -166,20 +166,6 @@ spec:
                       description: The name of the `MembershipRole`. Must be one of
                         `OWNER`, `MANAGER`, `MEMBER`.
                       type: string
-                    restrictionEvaluations:
```
jingyih, line 169: This was already beta, we probably want to duplicate the "state" field here.

dev/tools/controllerbuilder/generate.sh
```diff
@@ -222,12 +222,12 @@ go run . generate-mapper \
 
 # CloudIdentity : Membership
 go run . generate-types \
-     --service mockgcp.cloud.cloudidentity.groups.v1beta1 \
+     --service google.apps.cloudidentity.groups.v1beta1 \
```
jingyih, line 225: Let's just combine the two resource into a single command. It is more correct.

```
go run . generate-types \
     --service google.apps.cloudidentity.groups.v1beta1 \
     --resource CloudIdentityGroup:Group \
     --resource CloudIdentityMembership:Membership \
     --api-version "cloudidentity.cnrm.cloud.google.com/v1beta1"
```

jingyih (commented):
I added a non-blocking comment.

/lgtm

apis/cloudidentity/v1beta1/membership_types.go
```diff
@@ -91,46 +94,54 @@ type CloudIdentityMembershipObservedState struct {
 	// DeliverySetting
 
 	// The state output field is in a list
-	// .spec.[]roles.restrictionEvaluations.memberRestrictionEvaluationstate
-	// Not sure if we need to build an array copying everything from the spec here.
-	// []MembershipMemberRestrictionEvaluation.State
+	// .spec.[]roles.restrictionEvaluations.memberRestrictionEvaluation.state
+
+	Roles []*MembershipRolesObservedState `json:"roles,omitempty"`
```
maqiuyujoyce, line 99: Just curious, does it not have a field description, and mapping to a protobuf field?

gemmahou: It's unnecessary to have +kcc:proto tag to each subfields, as long as we have the tag in line 83. 

For the Roles field, I've copied the first sentence from the Spec description: 'The MembershipRoles that apply to the Membership.' As it's output-only, the field requirements are unnecessary.

jingyih: Good question! If the resource has a standard proto (not one generated from OpenAPI), I believe this field will be automatically generated with the `// +kcc:proto:field=` annotation. However, for this specific resource, splitting the field into "ObservedState" is done manually. I don’t think we should require developers to track which annotations are missing—this should be handled by the tool. If the tool doesn’t do it, I suppose that’s acceptable.

jingyih (commented):
/lgtm

maqiuyujoyce (commented):
After some thinking, now I remember why we didn't put the `restrictionEvaluations` under `status` when we supporedt this resource.

`restrictionEvaluations`'s parent, `roles`, is a list of object that doesn't have a clear identifier for each entry (maybe there is an ID for each entry that I'm not aware of), so only listing out `restrictionEvaluations` in the `roles` will be confusing for users because they don't know how entries under `status` map to entries under `spec`.
APIs don't guarantee the returned list contains the entries of the same order, and even so, comparing and checking the values from both `spec` and `status` don't feel like a good UX.

I suggest we use the same struct for MembershipRoles and MembershipRolesObservedState structs. What do you think?

apis/cloudidentity/v1beta1/membership_types.go
jingyih: Yeah I would agree. For manual fields, it is still better to add the annotation, because it is simply additional information for the tools. But it should not be required.

yuwenma:
I don't quite understand the goal of this PR, like why we don't add it in the adding-types step, why a status field (for observability purpose) is needed,  could you elaborate?

jingyih:
It is an output-only field that technically should live in ObservedState. The field is duplicated in Spec due to backward compatibility concerns.

gemmahou:
> I don't quite understand the goal of this PR, like why we don't add it in the adding-types step, why a status field (for observability purpose) is needed, could you elaborate?

This is a follow up of Barni's types PR, since he's out this week, and I'm not able to modify his PR directly.
This PR is to address Jingyi's [comment](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3678#discussion_r1957220488). spec.[]roles.restrictionEvaluations contains both spec and output-only fields, and we want to duplicate the output-only fields to ObservedState to align with our requirments.

gemmahou:
> After some thinking, now I remember why we didn't put the `restrictionEvaluations` under `status` when we supporedt this resource.
> 
> `restrictionEvaluations`'s parent, `roles`, is a list of object that doesn't have a clear identifier for each entry (maybe there is an ID for each entry that I'm not aware of), so only listing out `restrictionEvaluations` in the `roles` will be confusing for users because they don't know how entries under `status` map to entries under `spec`. APIs don't guarantee the returned list contains the entries of the same order, and even so, comparing and checking the values from both `spec` and `status` don't feel like a good UX.
> 
> I suggest we use the same struct for MembershipRoles and MembershipRolesObservedState structs. What do you think?

Ah good point! The order of the list is essential to take into consideration. 
Two thoughts: a) Following the past exception where we did not move fields to Status, keep the output-only fields in Spec to maintain the same UX. or b) Duplicate the entire Roles field from Spec to ObservedState, and use the MembershipRoles struct(so we no longer need the MembershipRolesObservedState, similar to your suggestion).

jingyih:
Essentially, we are treating the entire "MembershipRoles" field as both configurable and an output field, which sounds good to me.  

Alternatively, we could keep this field only in Spec. If a customer requests it to be displayed in `.status.observedState` as an output field, we can add it then.

maqiuyujoyce:
@jingyih @gemmahou both work for me. I prefer we do the right thing to have the entire list of membership roles under `status.observedState`. @yuwenma what do you think?

yuwenma (commented):
/hold

Do we have the direct controller for this resource? If not, we should have the controller ready then add new field. Otherwise, it surfaces the change to users with no reasons.
Here's the steps for [Beta migrations](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/develop-resources/scenarios/migrate-tf-resource-beta.md)

ldanielmadariaga:

PR #3708, opened 2025-02-19
llm: compilable AdvisorynotificationsSettings reconciler

yuwenma:
- **[llm] fix gcp imports for AdvisorynotificationsSettings**
- **[llm] fix gobuilds for AdvisorynotificationsSettings**
- **add basic AdvisorynotificationsSettings reconciler**
- **make read-pr**


config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_advisorynotificationssettings.advisorynotifications.cnrm.cloud.google.com.yaml
```diff
@@ -0,0 +1,163 @@
...
+  labels:
+    cnrm.cloud.google.com/managed-by-kcc: "true"
+    cnrm.cloud.google.com/system: "true"
+  name: advisorynotificationssettings.advisorynotifications.cnrm.cloud.google.com
+spec:
+  group: advisorynotifications.cnrm.cloud.google.com
+  names:
+    categories:
+    - gcp
+    kind: AdvisorynotificationsSettings
```
justinsb, line 16: I think this and the name of the group are the only "one way door" decisions here.

Group name looks good (I assume this is our first resource in this group, otherwise we would have to match what came before)

I think the kind should be AdvisoryNotificationSettings (capital N) for consistency with how we have named kinds elsewhere.  At least that's my 2c :-)

yuwenma: Sg! I can change the way of composing the Kind name and make it all words capitalized.  should not be a hard change

PR #3722, opened 2025-02-19
codebot: improve EditFile tool

justinsb:
We try to match lines of context, and accept some fuzzy matching.

This is what the LLM keeps sending us, so we should probably work with it!


yuwenma (commented):
/lgtm
/approve

Thank you!

PR #3741, opened 2025-02-20
mock: add support for ManagedKafkaTopic

jingyih:

PR #3742, opened 2025-02-20
feat: promote ManagedKafkaTopic to v1beta1

jingyih:


jingyih:
/hold
Need to rebase to #3741

config/samples/resources/managedkafkatopic/managedkafka_v1beta1_managedkafkatopic.yaml
```diff
@@ -0,0 +1,29 @@
...
+  name: managedkafkatopic-sample
+spec:
+  projectRef:
+    # Replace ${PROJECT_ID?} with your project ID
+    external: ${PROJECT_ID?}
+  location: us-central1
+  clusterRef:
+    name: managedkafkatopic-dep
+  partitionCount: 2
+  replicationFactor: 3
```
yuwenma, line 27: Does MKT give default value to these two fields?

pkg/test/resourcefixture/sets.go
```diff
@@ -108,6 +108,7 @@ func IsPureDirectResource(gk schema.GroupKind) bool {
 		"SecureSourceManagerInstance",
 		"SecureSourceManagerRepository",
 		"ManagedKafkaCluster",
+		"ManagedKafkaTopic",
```
yuwenma, line 111: @maqiuyujoyce Could you confirm if we need this hack after you removed direct from the service mapping path?

pkg/controller/direct/managedkafka/topic_controller.go
yuwenma, line 1: Could you elaborate why we did a special immutable check in line 174? 
`paths.Has("replication_factor")` I was expecting either the `CompareProtoMessage` or GCP service do the validation but not direct controller, especially if the direct controller has no additional handling for this field.

yuwenma (commented):
/approve
/hold 

One question about `replication_factor`, otherwise looks great

pkg/controller/direct/managedkafka/topic_controller.go
jingyih: I'm okay with removing the check. We can defer to GCP to verify the immutability of this field.

config/samples/resources/managedkafkatopic/managedkafka_v1beta1_managedkafkatopic.yaml
jingyih: Both fields are required.

jingyih: Marked both fields as "required" in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3742/commits/bff9ffac6a6c3263cb1df6b3c72813ed60eab157

PR #3745, opened 2025-02-20
fix: update groupversion_info.go

acpana:
This should probably be `v1beta1`. Yet another thing we can write a linter for 😝

### DIFF
--- a/apis/apigee/v1beta1/groupversion_info.go
+++ b/apis/apigee/v1beta1/groupversion_info.go
@@ -23,7 +23,7 @@ import (
 
 var (
 	// GroupVersion is group version used to register these objects
-	GroupVersion = schema.GroupVersion{Group: "apigee.cnrm.cloud.google.com", Version: "v1alpha1"}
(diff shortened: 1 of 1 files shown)


yuwenma (commented):
I assume this should be captured when the resource is promoted to beta: the test suite has a unknown v1beta1 versioning so the test should fail. Could you investigate and see why the presubmit test could not catch it?

apis/apigee/v1beta1/groupversion_info.go
acpana, line 1: @yuwenma let me answer your question here. If I had to guess controller gen uses the package name as the version.

I think that's true:

https://github.com/kubernetes-sigs/controller-tools/blob/a57ec68d4aca081c0ead223796f18b7c6c3a61c1/pkg/crd/parser.go#L135-L145

So everything gets generated right: the CRD, the tests, etc. The one thing that won't work is things that will actually reference this GroupVersion (like the ApigeeEnvironmentGVK). I found this out while creating the tests for EnvgroupAttachment.

yuwenma: Oh, this is a reference? That make sense now because KCC only matches KNN but not GVKNN

PR #3746, opened 2025-02-20
Separate SecureSourceManager instance fields into input and observed structs

ericpang777:
### Change description

- Separate hostConfig and privateConfig into the spec and observed structs.
- Add createTimestamp, updateTimestamp, and labels fields.
- Update the instance mock for privateConfig.

### Tests you have done


ericpang777:
@gemmahou

ericpang777:
Test failure looks unrelated

apis/securesourcemanager/v1alpha1/types.generated.go
```diff
@@ -16,20 +16,41 @@ package v1alpha1
 
 // +kcc:proto=google.cloud.securesourcemanager.v1.Instance.HostConfig
 type Instance_HostConfig struct {
```
gemmahou, line 18: remove empty struct?

pkg/controller/direct/securesourcemanager/mapper.go
```diff
@@ -62,32 +61,7 @@ func OperationMetadata_EndTime_ToProto(mapCtx *direct.MapContext, in *string) *t
 	mapCtx.Errorf("OperationMetadata_EndTime_ToProto not implemented")
 	return nil
 }
-func Instance_PrivateConfig_FromProto(mapCtx *direct.MapContext, in *pb.Instance_PrivateConfig) *krm.Instance_PrivateConfig {
```
gemmahou, line 65: Why thefunctions are removed?

ericpang777: I realized the generator correctly generates it now. It essentially got "moved" to mapper.generated.go

apis/securesourcemanager/v1alpha1/instance_types.go
```diff
@@ -85,25 +85,19 @@ type SecureSourceManagerInstanceObservedState struct {
 	StateNote *string `json:"stateNote,omitempty"`
 
 	// Output only. A list of hostnames for this instance.
-	HostConfig *Instance_HostConfig `json:"hostConfig,omitempty"`
+	HostConfig *Instance_HostConfigObservedState `json:"hostConfig,omitempty"`
```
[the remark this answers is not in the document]
ericpang777: Ahh I forgot to add this in, I thought I did but that must've been in another branch.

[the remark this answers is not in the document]
ericpang777: Done

apis/securesourcemanager/v1alpha1/types.generated.go
ericpang777: This was autogenerated though, but I can just remove it. Done

apis/securesourcemanager/v1alpha1/types.generated.go
```diff
@@ -14,10 +14,6 @@
 
 package v1alpha1
 
-// +kcc:proto=google.cloud.securesourcemanager.v1.Instance.HostConfig
-type Instance_HostConfig struct {
```
jingyih, line 18: cc @yuwenma this looks like another example that we need this bug fix.
https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3667

apis/securesourcemanager/v1alpha1/types.generated.go
jingyih: Yeah, this should be addressed in #3667, but there's some debate about whether fixing the bug is worth adding complexity to the tool.

gemmahou: Thanks, I will keep an eye on the PR. 

While I feel even we do not plan to clean empty struct by the generator, we should clean it up manually(it's very rare we run the generator again once the resource is landed). So I prefer removing this here to unblock the PR.

jingyih: > it's very rare we run the generator again once the resource is landed

This is the current reality, but I don’t believe it is our intention. We avoid rerunning because the tool isn’t stable enough (e.g. it will create empty go structs). However, not rerunning prevents the resource from being updated when there are changes in GCP API. 

For this PR, I agree that we should manually remove the empty struct.

gemmahou:
Looks good. Pending updates on roundtrip_test.go and _http.log

ericpang777:
> Need to update the http log of `securesourcemanagerinstanceprivate`, same comment to the repository PR.

Done, also updated the private instance mock

yuwenma (commented):
/approve

PR #3756, opened 2025-02-22
Support AlloyDBCluster direct controller

maqiuyujoyce:
### Change description

Migrate AlloyDBCluster to direct controller. Following are some behavioral changes when using the direct controller:

1. Mutable but unreadable fields are no longer preserved in the object. This means that if `spec.initialUser` or `spec.displayName` is configured in the manifest, then a PATCH request will be triggered during each reconciliation no matter there is a user change or not.
2. Diff of the network field is ignored because we don't have an easy way to compare project ID and project number right now. This is a blocker if we want to turn AlloyDBCluster as a pure direct resource.

### Tests you have done

### DIFF
--- a/pkg/controller/direct/alloydb/cluster_controller.go
+++ b/pkg/controller/direct/alloydb/cluster_controller.go
@@ -254,29 +254,13 @@ func (a *ClusterAdapter) resolveKRMDefaultsForUpdate() {
 // TODO: Scenario test case: Update initialUser.password from `value` to `valueFrom` and vise versa.
 func (a *ClusterAdapter) resolveInitialUserField(ctx context.Context) error {
 	obj := a.desired
-	if obj.Spec.InitialUser != nil && obj.Spec.InitialUser.Password != nil {
-		value := obj.Spec.InitialUser.Password.Value
-		valueFrom := obj.Spec.InitialUser.Password.ValueFrom
-		if value != nil && valueFrom != nil {
-			return fmt.Errorf("only one of 'spec.initialUser.password.value' " +
-				"and 'spec.initialUser.password.valueFrom' should be configured: " +
-				"both are configured")
-		}
-		if value != nil {
-			return nil
-		}
-		if valueFrom != nil {
-			if valueFrom.SecretKeyRef == nil {
-				return fmt.Errorf("'spec.initialUser.password.valueFrom.secretRef' " +
-					"should be configured")
-			}
-			secretValue, _, err := direct.GetSecret(ctx, valueFrom.SecretKeyRef, obj.Namespace, a.reader)
-			if err != nil {
-				return err
-			}
-			obj.Spec.InitialUser.Password.Value = direct.PtrTo(secretValue)
-			return nil
-		}
+	if obj.Spec.InitialUser == nil || obj.Spec.InitialUser.Password == nil {
+		return nil
+	}
+
+	// Resolve sensitive field 'spec.initialUser.password' when it is set.
+	if err := direct.ResolveSensitiveField(ctx, obj.Spec.InitialUser.Password, "spec.initialUser.password", obj.Namespace, a.reader); err != nil {
+		return err
 	}
 
 	return nil
@@ -637,8 +621,6 @@ func (a *ClusterAdapter) Delete(ctx context.Context, deleteOp *directbase.Delete
 	log := klog.FromContext(ctx)
 	log.V(2).Info("deleting Cluster", "name", a.id)
 
-	// TODO Test: Force deletion and non force deletion
-	// TODO: Test: Non force deletion succeeded and failed
 	req := &alloydbpb.DeleteClusterRequest{
 		Name:  a.id.String(),
 		Force: direct.ValueOf(a.desired.Spec.DeletionPolicy) == "FORCE",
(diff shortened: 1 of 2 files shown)


pkg/controller/direct/alloydb/cluster_controller.go
```diff
@@ -17,339 +17,644 @@ package alloydb
...
 
-// Delete implements the Adapter interface.
-func (a *clusterAdapter) Delete(ctx context.Context, deleteOp *directbase.DeleteOperation) (bool, error) {
-	// Already deleted
-	if a.resourceID == "" {
-		return false, nil
+// TODO: Scenario test cases: both networkConfig.networkRef and networkRef set; none set.
+func (a *ClusterAdapter) resolveNetworkRef(ctx context.Context) error {
+	obj := a.desired
+	if obj.Spec.NetworkRef == nil && obj.Spec.NetworkConfig == nil {
```
jingyih, line 134: Have we considered adding CEL rules for these checks, which allows us to shift the validation earlier in the process.

pkg/controller/direct/alloydb/cluster_controller.go
```diff
@@ -17,339 +17,644 @@ package alloydb
...
+}
 
-	created, err := a.client.Projects.Locations.Clusters.Get(a.fullyQualifiedName()).Context(ctx).Do()
-	if err != nil {
-		return fmt.Errorf("getting created cluster: %w", err)
+// TODO: Scenario test case: Update initialUser.password from `value` to `valueFrom` and vise versa.
+func (a *ClusterAdapter) resolveInitialUserField(ctx context.Context) error {
+	obj := a.desired
+	if obj.Spec.InitialUser != nil && obj.Spec.InitialUser.Password != nil {
+		value := obj.Spec.InitialUser.Password.Value
```
jingyih, line 258: Consider packaging this logic into a helper function to resolve the password from "refsv1beta1secret.Legacy"?

pkg/controller/direct/alloydb/cluster_controller.go
```diff
@@ -17,339 +17,644 @@ package alloydb
...
+	desiredPb.Labels = make(map[string]string)
+	for k, v := range a.desired.GetObjectMeta().GetLabels() {
+		desiredPb.Labels[k] = v
+	}
+	desiredPb.Labels["managed-by-cnrm"] = "true"
+
+	// 6. Set resource name. This step is not needed for other operations.
+	desiredPb.Name = a.id.String()
+	// 7. Handle default values for fields not yet supported in KRM types.
+	a.resolveGCPDefaults(desiredPb, a.actual)
```
jingyih, line 533: Do we need to set default values for fields that are not included in the updateMask?

maqiuyujoyce: Hmmm, the fields will be included in the updateMask if we don't set default values.
We are using `CompareProtoMessage()` function to compare the diff, and any differences in between the actual proto object and the desired proto object will be reflected. However the actual state contains many fields (and they have default values) that are not supported in the KRM types. So we need to match desired state with actual state about those fields to avoid unexpected diffs.

pkg/controller/direct/alloydb/cluster_controller.go
maqiuyujoyce: Good point. Will do.

pkg/controller/direct/alloydb/cluster_controller.go
maqiuyujoyce: I guess what you asked basically is this TODO: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/7d1940e97255c6d7df7f4e23cc8418e1b47a55f1/pkg/controller/direct/alloydb/cluster_controller.go#L567?

pkg/controller/direct/directbase/operations.go
```diff
@@ -82,15 +82,22 @@ var _ Operation = &CreateOperation{}
...
-func NewCreateOperation(client client.Client, object *unstructured.Unstructured) *CreateOperation {
+func NewCreateOperation(lifecycleHandler lifecyclehandler.LifecycleHandler, client client.Client, object *unstructured.Unstructured) *CreateOperation {
 	op := &CreateOperation{}
+	op.lifecycleHandler = lifecycleHandler
 	op.client = client
 	op.object = object
 	return op
 }
 
+func (o *CreateOperation) RecordUpdatingEvent() {
```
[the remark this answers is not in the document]
maqiuyujoyce: Yes. It's used to record the `Updating` state during resource creation. Direct controller previously didn't support it. I added it to match the existing behavior in the TF-based resource.

pkg/controller/direct/alloydb/cluster_controller.go
maqiuyujoyce: Done.

pkg/controller/direct/alloydb/cluster_controller.go
jingyih: Oh, I see. I usually use `spec.someField != nil` to conditionally trigger the diff, which means I don't use `CompareProtoMessage()`. It turns out the amount of code is about the same, so I suppose there's no real workaround.

pkg/controller/direct/alloydb/cluster_controller.go
jingyih: Yeah probably not. Even if it can be done, the logic is probably more readable if implemented here.

jingyih (commented):
/lgtm
/approve

PR #3760, opened 2025-02-22
chore: move scenarios-tests github action to follow pattern

justinsb:
* github actions should map to a script in dev/ci/
* scripts should be runnable directly
* tests that verify golden output should write the golden output

### DIFF
--- a/dev/ci/scenarios-tests
+++ b/dev/ci/scenarios-tests
@@ -18,11 +18,13 @@ set -o nounset
 set -o pipefail
 
 REPO_ROOT="$(git rev-parse --show-toplevel)"
-
 cd ${REPO_ROOT}/
 
 echo "Downloading envtest assets..."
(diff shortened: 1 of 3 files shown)


justinsb:
By putting the scripts in one place, and making sure that they all write output, we can easily regenerate the golden output after a mock change.

dev/ci/scenarios-tests
```diff
@@ -18,12 +18,14 @@ set -o nounset
...
 
 REPO_ROOT="$(git rev-parse --show-toplevel)"
-
 cd ${REPO_ROOT}/
 
 echo "Downloading envtest assets..."
 export KUBEBUILDER_ASSETS=$(go run sigs.k8s.io/controller-runtime/tools/setup-envtest@latest use -p path)
 
+# Always write golden output
+export WRITE_GOLDEN_OUTPUT=1
```
yuwenma, line 27: Are we going to run this not only in the CI action but somewhere else?

yuwenma (commented):
/lgtm
/approve

PR #3761, opened 2025-02-23
tests: split out some more services

justinsb:
The fixtures test time is getting big again, so splitting out a few more services.


yuwenma (commented):
/lgtm
/approve

 +100. Thank you!

PR #3764, opened 2025-02-23
chore: Scripts to create the expected 3 PR structure

justinsb:


experiments/conductor/cmd/runner/scripts/crd_mapper_fuzzer/01-write-generator-script.sh
```diff
@@ -0,0 +1,117 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+cd ${REPO_ROOT}/dev/tools/controllerbuilder
```
yuwenma, line 76: Maybe we can ship the controllerbuilder (so this will be `go install`). It is easy to use the wrong path,

experiments/conductor/cmd/runner/scripts/crd_mapper_fuzzer/02-generate-spec-and-status.sh
```diff
@@ -0,0 +1,65 @@
...
+git co master
+git co ${BRANCH_NAME}
+
+cd apis/${SERVICE}/${CRD_VERSION}
+
+# TODO: We should be able to in-place replace here
+# TODO: Generate spec and status separately
+# TODO: lower case (but it's nice right now because it always fails compilation)
+# TODO: Auto-exclude any examples with the same proto??
+controllerbuilder prompt --src-dir ~/kcc/k8s-config-connector --proto-dir ~/kcc/k8s-config-connector/.build/third_party/googleapis/ <<EOF >> ${RESOURCE}_types.go
```
yuwenma, line 55: nit: This "RESOURCE" is not the KIND, but the PROTO_RESOURCE of lower case.

yuwenma (commented):
/lgtm
/approve

Users may need to adjust some small places to make it work in their working directory . e.g. ~kcc and the controllerbuilder path. But I guess that's part of the goals

experiments/conductor/cmd/runner/scripts/crd_mapper_fuzzer/02-generate-spec-and-status.sh
```diff
@@ -0,0 +1,65 @@
...
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+cd ${REPO_ROOT}
+
+git co master
+git co ${BRANCH_NAME}
+
+cd apis/${SERVICE}/${CRD_VERSION}
+
+# TODO: We should be able to in-place replace here
+# TODO: Generate spec and status separately
```
jingyih, line 52: FYI, I had a type-generator-v2 which does this. I'm not working on it at the moment, but when I tested it on `ManagedKafkaCluster`, the generated types required no manual edits.

https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3748

PR #3775, opened 2025-02-24
chore: import DCL as git subtree

justinsb:
- **chore: import DCL as subtree**
- **Squashed 'third_party/github.com/GoogleCloudPlatform/declarative-resource-client-library/' content from commit a0e375a780**


justinsb:
This will let us carry patches against DCL, until we can replace them or get them upstreamed.

justinsb:
/assign @yuwenma 

cc @maqiuyujoyce

PR #3778, opened 2025-02-25
feat: add resource Spannerbackupschedule

xiaoweim:


yuwenma:
Could you send the first PR for Mock + gcloud first?

xiaoweim:
> Could you send the first PR for Mock + gcloud first?

Sure! This is the PR for mock + gcloud https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3793

apis/spannerbackupschedules/v1alpha1/backupschedule_types.go
```diff
@@ -0,0 +1,156 @@
...
+	//  * For multi-regional instance configs of type GOOGLE_MANAGED,
+	//  either specify a multi-regional location KMS key or multiple regional
+	//  location KMS keys that cover all regions in the instance config.
+	//  * For an instance config of type USER_MANAGED, please specify only
+	//  regional location KMS keys to cover each region in the instance config.
+	//  Multi-regional location KMS keys are not supported for USER_MANAGED
+	//  instance configs.
+	// +kcc:proto:field=google.spanner.admin.database.v1.CreateBackupEncryptionConfig.kms_key_names
+	KMSKeyNameRefs []*refs.KMSCryptoKeyRef `json:"kmsKeyNameRefs,omitempty"`
+}
```
yuwenma, line 91: Wondering if we should merge these into kmsKeyNameRefs. But I don't have preference.

apis/spannerbackupschedules/v1alpha1/backupschedule_types.go
```diff
@@ -0,0 +1,156 @@
...
+	UpdateTime *string `json:"updateTime,omitempty"`
+
+	// Cron style schedule specification.
+	// +kcc:proto:field=google.spanner.admin.database.v1.BackupScheduleSpec.cron_spec
+	CronSpec *CrontabSpecObservedState `json:"cronSpec,omitempty"`
+}
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// TODO(user): make sure the pluralizaiton below is correct
```
yuwenma, line 126: nit: remove

apis/spannerbackupschedules/v1alpha1/groupversion_info.go
yuwenma, line 1: Mind elaborating why the service is `spannerbackupschedules` but not `spanner`? Mine shows as this service is "spanner".

apis/spannerbackupschedules/v1alpha1/backupschedule_types.go
xiaoweim: The [rest API](https://cloud.google.com/spanner/docs/reference/rest/v1/projects.instances.databases.backupSchedules#EncryptionType) has these two fields, the difference between these two fields are the encryption type. 

I think if we want to merge the two into one in our KRM CRD, then our controller/mapper needs to take care of populating the right field in the request based on the encryption type. I would prefer to keep it this way now.

apis/spannerbackupschedules/v1alpha1/backupschedule_identity.go
yuwenma, line 1: We also need to update the directory "api/spannerbackupschedules" --> "apis/spanner". Here and else where

apis/spanner/v1alpha1/backupschedule_types.go
```diff
@@ -0,0 +1,155 @@
...
+	// +kcc:proto:field=google.spanner.admin.database.v1.CreateBackupEncryptionConfig.encryption_type
+	EncryptionType *string `json:"encryptionType,omitempty"`
+
+	// Optional. The Cloud KMS key that will be used to protect the backup.
+	//  This field should be set only when
+	//  [encryption_type][google.spanner.admin.database.v1.CreateBackupEncryptionConfig.encryption_type]
+	//  is `CUSTOMER_MANAGED_ENCRYPTION`. Values are of the form
+	//  `projects/<project>/locations/<location>/keyRings/<key_ring>/cryptoKeys/<kms_key_name>`.
+	// +kcc:proto:field=google.spanner.admin.database.v1.CreateBackupEncryptionConfig.kms_key_name
+	KMSKeyNameRef *refs.KMSCryptoKeyRef `json:"kmsKeyNameRef,omitempty"`
```
yuwenma, line 72: Suggest renaming this to KMSKeyRef (ditto to json name)

apis/spanner/v1alpha1/backupschedule_types.go
```diff
@@ -0,0 +1,155 @@
...
+	//  location KMS key.
+	//  * For multi-regional instance configs of type GOOGLE_MANAGED,
+	//  either specify a multi-regional location KMS key or multiple regional
+	//  location KMS keys that cover all regions in the instance config.
+	//  * For an instance config of type USER_MANAGED, please specify only
+	//  regional location KMS keys to cover each region in the instance config.
+	//  Multi-regional location KMS keys are not supported for USER_MANAGED
+	//  instance configs.
+	// +kcc:proto:field=google.spanner.admin.database.v1.CreateBackupEncryptionConfig.kms_key_names
+	KMSKeyNameRefs []*refs.KMSCryptoKeyRef `json:"kmsKeyNameRefs,omitempty"`
```
yuwenma, line 90: ditto

mockgcp/mockserviceusage/serviceusagev1beta1.go
```diff
@@ -75,6 +75,9 @@ func (s *ServiceUsageV1Beta1) GenerateServiceIdentity(ctx context.Context, req *
 		identity.Email = "service-" + strconv.FormatInt(name.Project.Number, 10) + "@gcp-sa-managedkafka.iam.gserviceaccount.com"
 	case "workflows.googleapis.com":
 		identity.Email = "service-" + strconv.FormatInt(name.Project.Number, 10) + "@gcp-sa-worksflow.iam.gserviceaccount.com"
+	case "spanner.googleapis.com":
+		identity.Email = "service-" + strconv.FormatInt(name.Project.Number, 10) + "@gcp-sa-spanner.iam.gserviceaccount.com"
+		identity.UniqueId = "123456789009"
```
yuwenma, line 80: @anhdle-sso Could you double check on this? I'm wondering why SpannerInstance does not need this.

apis/spanner/v1alpha1/backupschedule_types.go
```diff
@@ -0,0 +1,155 @@
...
+	// +kcc:proto:field=google.spanner.admin.database.v1.BackupSchedule.encryption_config
+	EncryptionConfig *CreateBackupEncryptionConfig `json:"encryptionConfig,omitempty"`
+
+	// The schedule creates only full backups.
+	// +kcc:proto:field=google.spanner.admin.database.v1.BackupSchedule.full_backup_spec
+	FullBackupSpec *FullBackupSpec `json:"fullBackupSpec,omitempty"`
+
+	// The schedule creates incremental backup chains.
+	// +kcc:proto:field=google.spanner.admin.database.v1.BackupSchedule.incremental_backup_spec
+	IncrementalBackupSpec *IncrementalBackupSpec `json:"incrementalBackupSpec,omitempty"`
```
yuwenma, line 51: Not a blocker, but cannot add a TODO about confirming if incrementalBackupSpec and fullBackupSpec should be a reference or not (maybe to SpannerBackup)? Since this is an alpha resource, we may need to re-iterate the APIs to clear the TODOs before promoting to beta.

mockgcp/mockserviceusage/serviceusagev1beta1.go
anhdle-sso: You don't have encryption on Instance level so we don't have KMS key field for that resource. DB and backup would have encryption which requires this change for mock.

apis/spanner/v1alpha1/backupschedule_types.go
xiaoweim: I don't think these fields are references. From the REST API [FullBackup](https://cloud.google.com/spanner/docs/reference/rest/v1/projects.instances.databases.backupSchedules#fullbackupspec) and [IncrementalBackup](https://cloud.google.com/spanner/docs/reference/rest/v1/projects.instances.databases.backupSchedules#incrementalbackupspec) specify the type of spanner backup. For example, if a backup schedule is created with type set to incremental backup, it will create multiple backups with the first one being a full backup, so I don't think it will map to any specific backup. It is more like telling the schedule which type of backup to create. 

But I can add a TODO just to confirm before promoting to beta.

yuwenma: Thanks for the confirmation. This is  exactly what I wanted to know from the TODO. 😄

yuwenma (commented):
/lgtm
/approve

PR #3781, opened 2025-02-25
feat: multi-parent for IAPSettings

jingyih:
Rather than using the `name` string field, construct the name based on parent Refs.


apis/iap/v1alpha1/iapsettings_identity.go
```diff
@@ -52,14 +53,102 @@ func (i *IAPSettingsIdentity) ID() string {
 // New builds a IAPSettingsIdentity from the Config Connector IAPSettings object.
 func NewIAPSettingsIdentity(ctx context.Context, reader client.Reader, obj *IAPSettings) (*IAPSettingsIdentity, error) {
 	// Get desired ID
-	// Note that we cannot use `metadata.name` as resourceID since the supported resource ID formats are not valid Kubernetes names.
 	resourceID := common.ValueOf(obj.Spec.ResourceID)
 	if resourceID == "" {
```
yuwenma, line 57: Suggest casting the "parent" or "identity" interface to struct, so the code can self explain and reflects a One Of relationship.   Otherwise, it would be error prone and hard to maintain the multi-parent relationship.

yuwenma (commented):
/lgtm
/approve

PR #3786, opened 2025-02-25
tests: capture kube-apiserver logs into artifacts

justinsb:
Only artifacts, because we keep our golden log for GCP.

This gives a more complete picture of what is happening.


yuwenma (commented):
/lgtm
/approve

Did you see anything interesting from the kube-apiserver that we want to look at?

PR #3791, opened 2025-02-25
mockgcp: bigquerybiglake

acpana:
LLM assisted mock for BigLakeTable.

Note there's not a simple gcloud command for this resource (https://cloud.google.com/bigquery/docs/reference/biglake/rest/v1/projects.locations.catalogs.databases.tables).

I'm hoping we can check this in and enhance the mock once we have a controller in hand.

Note: see the commits for manual edits made.

PR #3792, opened 2025-02-25
refactor: small changes

acpana:
Some of the fixes that I added to generate https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3791

### DIFF
--- a/dev/tools/controllerbuilder/pkg/toolbot/csv.go
+++ b/dev/tools/controllerbuilder/pkg/toolbot/csv.go
@@ -93,6 +93,10 @@ func (x *CSVExporter) VisitCodeDir(ctx context.Context, srcDir string) error {
 		}
 		// klog.Infof("%v", p)
 		if err := x.visitGoFile(ctx, p); err != nil {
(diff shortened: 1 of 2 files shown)


dev/tools/controllerbuilder/pkg/toolbot/csv.go
```diff
@@ -93,6 +93,10 @@ func (x *CSVExporter) VisitCodeDir(ctx context.Context, srcDir string) error {
 		}
 		// klog.Infof("%v", p)
 		if err := x.visitGoFile(ctx, p); err != nil {
+			if strings.HasSuffix(p, "cmd/runner/mock_commands.go") {
+				 klog.Infof("Skipping file: %v", p)
+				return nil
+			}
```
xiaoweim, line 99: I used the same trick to resolve the same error!

jingyih: Oh, I hadn't seen this PR before sending #3799. I believe we should skip the entire `experiments/` folder since it doesn't contain any examples.

PR #3799, opened 2025-02-26
codebot: skip experiments folder when extracting data points

jingyih:
Skip `experiements/` folder when extracting data points. This avoid incorrectly picking up strings in files such as 

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/fdb3f32b63d8424d95f844416e3c6f99e978dfd2/experiments/conductor/cmd/runner/mock_commands.go#L366-L374


yuwenma (commented):
/lgtm
/approve

THanks

PR #3802, opened 2025-02-26
feat: add types, mappers and fuzzer for NotebooksEnvironment

jingyih:
The v1alpha implementation did not check for reference fields.


jingyih:
Merged as part of https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3802

/close

PR #3557, opened 2025-01-29
feat: add DocumentAIProcessor as a direct resource

jingyih:

PR #3804, opened 2025-02-26
mockGCP: documentai processor and version

gemmahou:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/mockgcp/mockdocumentai/processors.go
+++ b/mockgcp/mockdocumentai/processors.go
@@ -1,0 +1,120 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mockdocumentai
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"google.golang.org/protobuf/proto"
+
+	longrunningpb "google.golang.org/genproto/googleapis/longrunning"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/common/projects"
+	pb "github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/generated/mockgcp/cloud/documentai/v1"
+)
+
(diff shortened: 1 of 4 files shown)


yuwenma (commented):
/lgtm
/approve

Did you get the golden log of this resource?

mockgcp/mockdocumentai/processors.go
```diff
@@ -0,0 +1,120 @@
...
+
+	fqn := name.String()
+
+	processorVersion := proto.Clone(req.GetProcessor()).(*pb.ProcessorVersion)
+	processorVersion.Name = fqn
+	now := time.Now()
+	req.Processor.CreateTime = timestamppb.New(now)
+	req.Processor.State = pb.Processor_ENABLED
+
+	if err := s.storage.Create(ctx, fqn, req.Processor); err != nil {
```
jingyih, line 65: Not sure how much fidelity we want the mocks to be. But my impression is that creating a Processor will also automatically create a few ProcessorVersions?

jingyih: The test was documented here:
https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3557#issue-2817365159

gemmahou:
> /lgtm /approve
> 
> Did you get the golden log of this resource?

Not able to test mockGCP at this stage-no gcloud, generated new test script but the recorder currently cannot capture REST logs

I only have the generated mockGCP in this PR. I have direct controller in #3817, I use the direct controller to test against mockGCP, capture logs, and modify the mockGCP. See https://chat.google.com/room/AAAA1cEHHoU/1Hru75aeB2Y/oK_bGA0q1J8?cls=10

mockgcp/mockdocumentai/processors.go
gemmahou: The code in this PR is autogenerated(see my reply to Yuwen's comment).
I fixed the ProcessorVersion mock in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3817/commits/a713eaa0040925a794c3ae3cd6c1036b1852a39d using direct controller. The fix of Processor mock could be done in a separate PR.

jingyih: The 3 PR structure may not be suitable in this case. We've already merged the controller for `DocumentAIProcessor`, including tests and real GCP logs. If we now merge a generated mock for `DocumentAIProcessor`, wouldn’t that break HEAD?

gemmahou: This will not break head because the mock Processor test is not enabled, once it's done we can enable it [here](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/bb23ee9164b38ac2b5287dee529f02e20f8a8689/config/tests/samples/create/harness.go#L736). This mock won't affect ProcessorVersion test as well since [I used external reference](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/a713eaa0040925a794c3ae3cd6c1036b1852a39d/pkg/test/resourcefixture/testdata/basic/documentai/v1alpha1/documentaiversion-basic/create.yaml#L22) for now.

jingyih: Sounds good. Let's address in a separate PR.

jingyih (commented):
/lgtm

PR #3806, opened 2025-02-26
fix: Drop apgia name from status, use spec.resourceID to acquire

jasonvigil:
The status.observedState.name field contains duplicate information that is already specified in status.externalRef.


yuwenma (commented):
/lgtm
/approve

Thanks!

PR #3816, opened 2025-02-28
feat: BigLakeTable controller

acpana:
Adds the direct controller and needed changes for `BigLakeTable`. An alpha direct resource!


apis/bigquerybiglake/v1alpha1/table_identity.go
```diff
@@ -69,6 +71,38 @@ func NewTableIdentity(ctx context.Context, reader client.Reader, obj *BigLakeTab
 		return nil, fmt.Errorf("cannot resolve location")
 	}
 
+	catalogRef := obj.Spec.CatalogRef
+	if catalogRef == nil {
+		return nil, fmt.Errorf("cannot resolve catalog")
+	}
```
yuwenma, line 77: Suggest making this a API level validation  ("required") rather than controller level https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/develop-resources/api-conventions/validations.md#basic-rules

apis/bigquerybiglake/v1alpha1/table_identity.go
```diff
@@ -92,29 +126,44 @@ func NewTableIdentity(ctx context.Context, reader client.Reader, obj *BigLakeTab
...
+			ProjectID:  projectID,
+			Location:   location,
+			CatalogID:  catalogID,
+			DatabaseID: databaseID,
 		},
 		id: resourceID,
 	}, nil
 }
 
 func ParseTableExternal(external string) (parent *TableParent, resourceID string, err error) {
```
yuwenma, line 151: Suggest moving the parent logic to the parent identity, otherwise we are duplicating the logic among databases, catalogs, and tables. You can look at Apigee for example.

pkg/controller/direct/bigquerybiglake/table_controller.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+
+	paths := make(sets.Set[string])
+	{
+		var err error
+		paths, err = common.CompareProtoMessage(desiredPb, a.actual, common.BasicDiff)
+		if err != nil {
+			return err
+		}
+	}
```
yuwenma, line 176: I like using "{}" to group logic as well, but it is a not recommended golang coding style. Thus I suggest removing line 170 and 176 to keep the code convention.

yuwenma:
Gentle ping on this PR @acpana

yuwenma:
/approve
/lgtm
/hold

I left some suggestions for better LLM work, but they are not blocker.

apis/bigquerybiglake/v1alpha1/table_identity.go
acpana: This is required today!

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/ad92d08fa0661a42382f9508596076249230fcc5/apis/bigquerybiglake/v1alpha1/table_types.go#L32-L34

But I always get a little defensive when coding and handling references. Do we want to just rely on the CRD validation?

apis/bigquerybiglake/v1alpha1/table_identity.go
acpana: Happy to male the template changes and follow that pattern! I used it for ApigeeEnvgroupIntance.

yuwenma:
Could you share the git compare between real and mock GCP for this resource?

apis/bigquerybiglake/v1alpha1/table_identity.go
yuwenma: We handle reference in controller (Adaptor initialization for required field, and Adaptor Create/Update for optional). 

But my previous comment is for required/optional, and not for reference.  Required and optional could be done via CRD validation unless it has additional logic (e.g. 3 fields have a anyOf requirement)

yuwenma (commented):
/lgtm

acpana:
/hold cancel

chatted offline to revisit real vs mock logs once we add the 2 dependent resources:

- Catalog
- Database

PR #3797, opened 2025-02-26
Add DocumentAIProcessorVersion types, mappers, CRD and fuzzer

gemmahou:
### Change description

Fixes #

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #3817, opened 2025-02-28
feat: DocumentAIVersion direct controller

gemmahou:
### Change description

Based on #3804 and #3797 

#3804 is not tested yet since there's no gcloud test script. I use the direct controller and the create.yaml added in this PR to test against mockGCP and fix the mock service, add normalizers.

### Tests you have done


yuwenma:
@gemmahou Could you resolve the merge conflict ? The dependent resource (documentAI processor) has been merged. 🥂 (kudos to @jingyih )

pkg/test/resourcefixture/testdata/basic/documentai/v1alpha1/documentaiversion-basic/create.yaml
```diff
@@ -0,0 +1,28 @@
...
+apiVersion: documentai.cnrm.cloud.google.com/v1alpha1
+kind: DocumentAIProcessorVersion
+metadata:
+  name: documentaiversion-${uniqueId}
+spec:
+  displayName: Test Document AI Version Name
+  processorRef:
+    # Use pre-created processor for version training.
+    # A processor requires manual setup after creation to allow for the training of a new version.
+    external: projects/${projectNumber}/locations/us/processors/7f8f177e3b9cc6d9
```
gemmahou, line 24: @yuwenma this is another case where we have to acquire a pre-created resource with generatedID. Similar to the test case in #3965. 

I'm not able to reference a KCC-managed resource, because after resource creation, it required manually setup dataset(I've done this in Pantheon) to be able to train a version.

yuwenma:
can you rebase from master HEAD to fix the presubmit?

pkg/test/resourcefixture/testdata/basic/documentai/v1alpha1/documentaiversion-basic/create.yaml
```diff
@@ -0,0 +1,28 @@
...
+  projectRef:
```
jingyih, line 25: Why do we need `projectRef` and `location` in the API? I thought we only need `processorRef` which will include project and location?

gemmahou: done

pkg/controller/direct/documentai/processorversion_controller.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	created := &documentaipb.ProcessorVersion{}
+	getReq := &documentaipb.GetProcessorVersionRequest{Name: metadata.GetCommonMetadata().GetResource()}
+
+	// if state is creating, keep retrieving until creation failed
+	// todo: currently we expect resource state to be FAILED after creation as there's no training data
+	// handle other successful state, i.e. undeployed after creation
+	done := false
+	for !done {
```
jingyih, line 158: Is there a reason we cannot use `op.Wait()`? If so, can we document it as code comments?

Also, this loop will just keep sending RPC calls until the GCP quota is hit?

pkg/controller/direct/documentai/processorversion_controller.go
```diff
@@ -0,0 +1,235 @@
...
+		return mapCtx.Err()
+	}
+	externalRef := created.GetName()
+	status.ExternalRef = direct.LazyPtr(externalRef)
+	return createOp.UpdateStatus(ctx, status, nil)
+}
+
+// Update updates the resource in GCP based on `spec` and update the Config Connector object `status` based on the GCP response.
+func (a *ProcessorVersionAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOperation) error {
+	// no update
```
jingyih, line 182: I think we still want to update the KRM status (in the case of acquiring a resource)

pkg/controller/direct/documentai/processorversion_controller.go
```diff
@@ -0,0 +1,235 @@
...
+
+// Find retrieves the GCP resource.
+// Return true means the object is found. This triggers Adapter `Update` call.
+// Return false means the object is not found. This triggers Adapter `Create` call.
+// Return a non-nil error requeues the requests.
+func (a *ProcessorVersionAdapter) Find(ctx context.Context) (bool, error) {
+	log := klog.FromContext(ctx)
+	log.V(2).Info("getting ProcessorVersion", "name", a.id)
+
+	if !a.id.HasKnownID() {
```
jingyih, line 107: What happens if we don't check this? Find() will return "false, nil" if resource is not found.

gemmahou: For this resource, resourceID id server-generated(i.e. we cannot use metadata.name as resourceID). Without this check, a.id is empty, and the error will be 400/403(invalid request) instead of 404.

pkg/controller/direct/documentai/processorversion_controller.go
gemmahou: Ah this is a bit tricky and I should add some comments.

`op.wait()` returns once the version resource exists, but the version's state can still be "creating." It takes a few seconds to reach a stable state like "failed" or "undeployed." That's why I have to explicitly retrieve the resource and check its state to determine if the operation has finished.

jingyih: This could be a helper function (not sure if it already exist in our codebase), but essentially we need some interval between the polls, and a timeout.

pkg/controller/direct/documentai/processorversion_controller.go
jingyih: Is this the same as checking `desiredResourceID == ""`? If so, we could avoid having an extra boolean, which is not a typical pattern in our controller code.

pkg/controller/direct/documentai/processorversion_controller.go
jingyih: Did a quick search. This is the closest I can find. Maybe we can create a helper function similar to this but with configurable poll interval.

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/045222bd5162376b1c853d1b04ea6756a79705bf/mockgcp/common/operations/wait.go#L28

pkg/controller/direct/documentai/processorversion_controller.go
gemmahou: Good point! I updated the code in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3817/commits/a12a7d8f2b6755c3ef5ae2b1e8d99d253caa998d

pkg/controller/direct/documentai/processorversion_controller.go
gemmahou: fixed in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3817/commits/2104a4f91d446caa3198a9f73e86613a4d7d5faf

pkg/controller/direct/documentai/processorversion_controller.go
gemmahou: I added a helper function in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3817/commits/60efeff4875bf5e652be9cc2521d59992cf7e10e, get created resource by checking its state, with poll interval and timeout.

jingyih (commented):
/lgtm
/approve

PR #3818, opened 2025-02-28
Added a runner for runmockgcptests against mock.

cheftako:
### Change description

Fixed a few preconditions.
Cleaned up my test sample metadata.
Added missing help commands.
Cleaned up log messages.

### Tests you have done

./bin/conductor runner --branch-repo=/usr/local/google/home/wfender/go/src/github.com/cheftako/k8s-config-connector_branches --branch-conf=branches.yaml --logging-dir=./logs --command=9

### DIFF
--- a/experiments/conductor/branches-all.yaml
+++ b/experiments/conductor/branches-all.yaml
@@ -17218,11 +17218,11 @@ branches:
       controller: terraform
       kind: ""
       package: ""
-      proto: ""
-      proto-path: ""
-      proto-service: ""
-      proto-msg: ""
-      host-name: ""
+      proto: BackendService
+      proto-path: google/cloud/compute/v1/compute.proto
+      proto-service: google.cloud.compute.v1.BackendServices
+      proto-msg: google.cloud.compute.v1.BackendService
+      host-name: compute.googleapis.com
       notes: []
     - name: compute-computebackendservicesignedurlkeys
       local: resource-compute-computebackendservicesignedurlkeys
--- a/experiments/conductor/cmd/runner/mock_commands.go
+++ b/experiments/conductor/cmd/runner/mock_commands.go
@@ -66,7 +66,7 @@ func createScriptYaml(opts *RunnerOptions, branch Branch) {
 	close := setLoggingWriter(opts, branch)
 	defer close()
 	if branch.Command == "" {
-		log.Printf("SKIPPING %s, no gcloud command\r\n", branch.Name)
+		log.Printf("SKIPPING %s, no gcloud command", branch.Name)
 		return
 	}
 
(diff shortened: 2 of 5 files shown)


experiments/conductor/cmd/runner/runner.go
```diff
@@ -205,6 +210,10 @@ func printHelp() {
 	log.Println("\t3 - Delete the local github branches from the metadata")
 	log.Println("\t4 - Create script.yaml for mock gcp generation in each github branch")
 	log.Println("\t5 - Create _http.log for mock gcp generation in each github branch")
+	log.Println("\t6 - Generate mock Service and Resource go files in each github branch")
+	log.Println("\t7 - Add service to mock_http_roundtrip.go in each github branch")
+	log.Println("\t8 - Add proto to makefile in each github branch")
```
yuwenma, line 215: Which proto are you referring here? Just some thoughts: maybe we can add all protos ahead to save this step.

yuwenma (commented):
/lgtm
/approve

experiments/conductor/branches-all.yaml
yuwenma, line 1: what's the difference between branches.yaml and branches-all.yaml?

barney-s: branches-all is the reference. All resources. 
branches is a subset and should be updated and used by some one using this tool.

experiments/conductor/cmd/runner/mock_commands.go
```diff
@@ -436,3 +438,41 @@ func addProtoToMakfile(opts *RunnerOptions, branch Branch) {
...
+	var out strings.Builder
+	checkoutBranch(branch, workDir, &out)
+
+	// Check to see if the http log file already exists
+	logFile := fmt.Sprintf("mock%s/testdata/%s/crud/_http.log", branch.Group, branch.Resource)
+	logFullPath := filepath.Join(workDir, logFile)
+
+	// Run the test against the generated mocks to determine quality
+	start := time.Now()
+	log.Printf("COMMAND: go test ./mockgcptests -v -run TestScripts/mock%s/testdata/%s/crud", branch.Group, branch.Resource)
```
barney-s, line 456: Would be more useful to comment this way
WRITE_GOLDEN_OUTPUT=1 E2E_GCP_TARGET=mock go test .....

experiments/conductor/cmd/runner/mock_commands.go
```diff
@@ -436,3 +438,41 @@ func addProtoToMakfile(opts *RunnerOptions, branch Branch) {
...
+
+func runMockgcpTests(opts *RunnerOptions, branch Branch) {
+	close := setLoggingWriter(opts, branch)
+	defer close()
+	workDir := filepath.Join(opts.branchRepoDir, "mockgcp")
+
+	var out strings.Builder
+	checkoutBranch(branch, workDir, &out)
+
+	// Check to see if the http log file already exists
```
barney-s, line 450: may be better to check for resource or service go here

barney-s (approved):
minor comments. none blocking.

experiments/conductor/cmd/runner/mock_commands.go
cheftako: I'm note sure, there are two parts to this one, the command log and the actual command.

For the log I'm not sure if its a good idea to represent the logged command differently than the execute command.

For the actual command I think probably not. 
"A=B command" is a shell trick and I'm not sure that shell trick will work here.
Eg. "egrep foo *.txt" does not work in exec because the * expansion is done bu the shell not the command. 
Also I think environment variables should be passed in through the environment API.
I think it would be better to change mockgcptests to take anything non-sensitive as a command line flag.

experiments/conductor/cmd/runner/mock_commands.go
cheftako: LOL the comment is wrong. 

I will say it probably does make sense to add a resource/service check here.
Feel free :D

PR #3819, opened 2025-02-28
feat: add VertexAIFeaturestore as a direct resource

jingyih:
This PR includes mock, types/mappers/CRD, and controller implementation.


apis/vertexai/v1alpha1/featurestore_types.go
```diff
@@ -0,0 +1,196 @@
...
+	OnlineServingConfig *Featurestore_OnlineServingConfig `json:"onlineServingConfig,omitempty"`
+
+	// Optional. TTL in days for feature values that will be stored in online
+	//  serving storage. The Feature Store online storage periodically removes
+	//  obsolete feature values older than `online_storage_ttl_days` since the
+	//  feature generation time. Note that `online_storage_ttl_days` should be less
+	//  than or equal to `offline_storage_ttl_days` for each EntityType under a
+	//  featurestore. If not set, default to 4000 days
+	// +kcc:proto:field=google.cloud.aiplatform.v1beta1.Featurestore.online_storage_ttl_days
+	OnlineStorageTtlDays *int32 `json:"onlineStorageTtlDays,omitempty"`
```
yuwenma, line 85: nit: "TTL"

apis/vertexai/v1alpha1/featurestore_types.go
```diff
@@ -0,0 +1,196 @@
...
+
+	// Optional. Customer-managed encryption key spec for data storage. If set,
+	//  both of the online and offline data storage will be secured by this key.
+	// +kcc:proto:field=google.cloud.aiplatform.v1beta1.Featurestore.encryption_spec
+	EncryptionSpec *EncryptionSpec `json:"encryptionSpec,omitempty"`
```
yuwenma, line 90: does this support other encryption format?

yuwenma (commented):
/approve

apis/vertexai/v1alpha1/featurestore_types.go
jingyih: Not sure I fully understand the question, do mean what happens if `EncryptionSpec` is not set? If unset, a "Google-owned and Google-managed encryption key" will be used as the default encryption. This is pretty standard across the GCP resources.

yuwenma: sorry for the confusion, I should add the comment in `type EncryptionSpec struct`. It looks like a one-of and currently only support KMS key. I'm wondering if it supports other secret management approaches.

jingyih: I did not see any oneof? It is the same struct shared by many VertexAI resources.

PR #3822, opened 2025-02-28
WIP feat: add CRD for Vertex Notebooks

barney-s:

PR #3823, opened 2025-02-28
Add conductor commands to generate Types, Mapper, CRD and fuzzer

barney-s:
CL that is using this tooling: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3822/commits


experiments/conductor/cmd/runner/crd_generator_commands.go
```diff
@@ -0,0 +1,198 @@
...
+set -o nounset
+set -o pipefail
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+cd ${REPO_ROOT}/dev/tools/controllerbuilder
+
+go run . generate-types \
+    --service <PROTO_PACKAGE> \
+    --api-version <CRD_GROUP>/<CRD_VERSION> \
+    --resource <CRD_KIND>:<PROTO_RESOURCE>
```
jingyih, line 51: Note that this will no longer work if a second resource is being generated under the same `--api-version`. This is a broader discussion, but I think we should have one git branch per `--api-version`, rather than per `--resource`.

experiments/conductor/branches.yaml
```diff
@@ -13,49 +13,31 @@
...
+      command: gcloud notebooks instances
+      group: notebooks
+      resource: instance
+      controller: Unknown
+      kind: NotebooksInstance
+      package: google.cloud.notebooks.v1
+      proto: Instance
+      proto-path: google/cloud/notebooks/v1/service.proto
+      proto-service: "google.cloud.notebooks.v1.NotebookService"
+      proto-msg: google.cloud.notebooks.v1.Instance
```
jingyih, line 28: Not related to this PR, but is `proto` the same as `proto-msg`, with the only difference being that `proto-msg` is the fully qualified name (i.e., `package` + `proto`)?

experiments/conductor/cmd/runner/crd_generator_commands.go
barney-s: i noticed it here: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3822/files#diff-dd75f3353e0a712412b061f3565c24dca0d5379d59d78f37235e8a28d9f638d6 

My question is do we run this script only once ?
Or do we run this script often.

Concern: If we make changes in the types.go for say Environment. Then i add instance it would run both. Would it revert the changes for environtment_types.go ?

barney-s: alternatively i will change the code to create generate_<type>.sh ?

jingyih: It will not revert environtment_types.go but it will regenerate types.generated.go. If we have a separate generate.sh per resource branch, they are going to fight with each other.

jingyih: With the current tooling and code structure, we need to generate "CRD/mapper/types" per `--api-version`. i.e., one git branch per `--api-version`.

barney-s: ok we may need to use llm to update existing file or create new one.

barney-s: done. works https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3822/files#diff-dd75f3353e0a712412b061f3565c24dca0d5379d59d78f37235e8a28d9f638d6

experiments/conductor/branches.yaml
```diff
@@ -13,49 +13,31 @@
 # limitations under the License.
 
 branches:
-  - name: accessapproval-approvalrequest
```
cheftako, line 16: Probably should try avoiding stomping on each others branches.yaml. Maybe add it to the .gitignore?

barney-s: I added it to .gitignore, but it still shows up as diff (may be because the file exists in the repo).

experiments/conductor/branches.yaml
cheftako: These are used to drive the automation. name and local are used for git operations and ids.
Interestingly proto-service, proto_path and proto-msg are used in templates.
I believe proto and remote are both essentially unused today.

experiments/conductor/cmd/runner/crd_generator_commands.go
```diff
@@ -0,0 +1,227 @@
...
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+cd ${REPO_ROOT}/dev/tools/controllerbuilder
+
+go run . generate-types \
```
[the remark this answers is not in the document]
barney-s: refactored the code to run the commands inline in go-code.

experiments/conductor/cmd/runner/crd_generator_commands.go
```diff
@@ -0,0 +1,320 @@
...
+		fmt.Sprintf("apis/%s/v1alpha1/", branch.Group),
+		fmt.Sprintf("pkg/controller/direct/%s/", branch.Group),
+		"config/crds/resources/")
+	gitCommit(workDir, &out, fmt.Sprintf("autogenerated types and CRDs using %s", scriptRelativePath))
+}
+
+func generateTypesAndMapper(opts *RunnerOptions, branch Branch) {
+	close := setLoggingWriter(opts, branch)
+	defer close()
+	workDir := filepath.Join(opts.branchRepoDir)
```
cheftako, line 149: filepath.Join is about getting the slashes correct. If your not concatenating directories/files, you don't need to call it.

experiments/conductor/cmd/runner/crd_generator_commands.go
```diff
@@ -0,0 +1,320 @@
...
+
+	var out strings.Builder
+	checkoutBranch(branch, workDir, &out)
+
+	// Change to controllerbuilder directory
+	controllerBuilderDir := filepath.Join(workDir, "dev/tools/controllerbuilder")
```
cheftako, line 155: This should be 
filepath.Join(workDir, "dev", "tools", "controllerbuilder")

experiments/conductor/cmd/runner/crd_generator_commands.go
```diff
@@ -0,0 +1,320 @@
...
+			"run", ".",
+			"generate-types",
+			"--service", branch.Package,
+			"--api-version", fmt.Sprintf("%s.cnrm.cloud.google.com/v1alpha1", branch.Group),
+			"--resource", fmt.Sprintf("%s:%s", branch.Kind, branch.Proto),
+		},
+		WorkDir: controllerBuilderDir,
+	}
+	if err := executeCommand(cfg, &out); err != nil {
+		log.Fatal(err)
```
cheftako, line 171: Please make sure to log the error & out just in case before calling fatal.

experiments/conductor/cmd/runner/crd_generator_commands.go
```diff
@@ -0,0 +1,320 @@
...
+		Args: []string{
+			"run", ".",
+			"generate-mapper",
+			"--service", branch.Package,
+			"--api-version", fmt.Sprintf("%s.cnrm.cloud.google.com/v1alpha1", branch.Group),
+		},
+		WorkDir: controllerBuilderDir,
+	}
+	if err := executeCommand(cfg, &out); err != nil {
+		log.Fatal(err)
```
cheftako, line 187: Ditto

experiments/conductor/cmd/runner/utilities.go
```diff
@@ -222,3 +235,71 @@ func setLoggingWriter(opts *RunnerOptions, branch Branch) closer {
...
+type CommandConfig struct {
+	Name    string        // Name of the command for logging
+	Cmd     string        // The command to run
+	Args    []string      // Command arguments
+	WorkDir string        // Working directory
+	Stdin   io.Reader     // Optional stdin
+	Timeout time.Duration // Timeout duration (default 5m)
+}
+
+// Helper function to execute a command with timing and logging
```
cheftako, line 268: :D

experiments/conductor/cmd/runner/crd_generator_commands.go
```diff
@@ -0,0 +1,320 @@
...
+			Args:    []string{"--ui-type=prompt", "--prompt=mockgcp/crdgen_prompt.txt"},
+			WorkDir: workDir,
+		}
+		if err := executeCommand(cfg, &out); err != nil {
+			log.Fatal(err)
+		}
+	}
+
+	// Add and commit changes
+	scriptRelativePath := fmt.Sprintf("apis/%s/v1alpha1/generate.sh", branch.Group)
```
cheftako, line 124: paths should be contructed using filepath.Join, at least for the slashes.

cheftako (commented):
Good start.

experiments/conductor/cmd/runner/crd_generator_commands.go
barney-s: executeCommand logs it.

experiments/conductor/cmd/runner/crd_generator_commands.go
barney-s: executeCOmmand logs it.

experiments/conductor/cmd/runner/crd_generator_commands.go
barney-s: of course. will change.

experiments/conductor/cmd/runner/crd_generator_commands.go
barney-s: done.

experiments/conductor/cmd/runner/crd_generator_commands.go
barney-s: fixed.

experiments/conductor/cmd/runner/crd_generator_commands.go
```diff
@@ -0,0 +1,320 @@
...
+	checkoutBranch(branch, workDir, &out)
+
+	// Run controllerbuilder to generate spec and status
+	log.Printf("Generating spec and status for %s", branch.Name)
+	stdinInput := fmt.Sprintf("// +kcc:proto=%s.%s\n", branch.ProtoSvc, branch.Proto)
+
+	cfg := CommandConfig{
+		Name:    "Spec/Status generation",
+		Cmd:     "controllerbuilder",
+		Args:    []string{"prompt", "--src-dir", workDir, "--proto-dir", filepath.Join(workDir, ".build/third_party/googleapis/")},
```
jingyih, line 251: Do you have an example commit generated by this command? My experience is that this step currently does not work?

experiments/conductor/cmd/runner/utilities.go
```diff
@@ -222,3 +235,96 @@ func setLoggingWriter(opts *RunnerOptions, branch Branch) closer {
...
+
+// CommandConfig holds the configuration for executing a command
+type CommandConfig struct {
+	Name    string            // Name of the command for logging
+	Cmd     string            // The command to run
+	Args    []string          // Command arguments
+	WorkDir string            // Working directory
+	Stdin   io.Reader         // Optional stdin
+	Env     map[string]string // Optional environment variables
+	Timeout time.Duration     // Timeout duration (default 5m)
```
acpana, line 266: big fan!

PR #3825, opened 2025-03-01
fix: handle already destroyed secret versions in `Delete`

jingyih:


yuwenma:
/lgtm
/approve

Good catch . Basically the SecretManager version is still retrievable even if destroyed.

Issue #3727, opened 2025-02-19
Handle deletion when the underlying GCP resource does not exist


gemmahou:
### Checklist

- [x] I did not find a related open enhancement request.
- [x] I understand that enhancement requests filed in the GitHub repository are by default low priority.
- [x] If this request is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Describe the feature or resource

If the underlying GCP resource does not exist, the krm resource should be able to be deleted too even if the deletion request returns 404 error. 
Here's the problem:
1. A resource fails to create due to a configuration error, GCP API rejects the request
2. Try to update the configuration, but the error field is immutable, requiring to delete and re-apply
3. The delete operation fails as well with a 404 error from the API, as the resource never exists
This results in a resource that cannot be deleted or updated, becoming stuck.

We have a recent improvement to address this https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3656, in apigee, firestore, sql and workstation controllers and the controller template.

Another option is to reuse the `existAlready` flag([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/d5fdcf11cf9d21c4bbc0b60f25051f686e38c70f/pkg/controller/direct/directbase/directbase_controller.go#L245)) to avoid making deletion request to GCP if we know the underlying resource does not exist.

gemmahou:
PR to address this: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4621

PR #3826, opened 2025-03-01
chore: Handle deletion when the underlying GCP resource does not exist

gemmahou:
### Change description

Fixes https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/3727

### Tests you have done

### DIFF
--- a/pkg/controller/direct/directbase/interfaces.go
+++ b/pkg/controller/direct/directbase/interfaces.go
@@ -36,9 +36,8 @@ type Model interface {
 type Adapter interface {
 	// Delete removes the GCP object.
 	// This can be called without calling Find.
-	// It returns (true, nil) if the object was deleted,
-	// and (false, nil) if the object was not found but should be presumed deleted.
-	// In an error, the state is not fully determined - a delete might be in progress.
+	// It returns (true, nil) if the object was deleted without error.
+	// In an error (false, err), the state is not fully determined - a delete might be in progress.
 	Delete(ctx context.Context, op *DeleteOperation) (deleted bool, err error)
 
 	// Find must be called as the first operation (unless we are deleting).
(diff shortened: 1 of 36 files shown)


pkg/controller/direct/directbase/interfaces.go
```diff
@@ -36,9 +36,8 @@ type Model interface {
 type Adapter interface {
 	// Delete removes the GCP object.
 	// This can be called without calling Find.
-	// It returns (true, nil) if the object was deleted,
-	// and (false, nil) if the object was not found but should be presumed deleted.
-	// In an error, the state is not fully determined - a delete might be in progress.
+	// It returns (true, nil) if the object was deleted without error.
+	// In an error (false, err), the state is not fully determined - a delete might be in progress.
 	Delete(ctx context.Context, op *DeleteOperation) (deleted bool, err error)
```
gemmahou, line 41: I feel the returned `deleted` might be unnecessary:
1. it's not used anywhere for now: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/bb23ee9164b38ac2b5287dee529f02e20f8a8689/pkg/controller/direct/directbase/directbase_controller.go#L267
2. the situation `return (false, nil) if the object was not found but should be presumed deleted`, we could call `Find` to determine the existence of the GCP obj, and avoid unnecessary Delete calls if obj does not exist. Also We don't have specific handling for the case of (false, nil); instead, we treat it the same as (true, nil).

`This can be called without calling Find.` Unsure if there's a use case of it I haven't considered...

pkg/test/resourcefixture/testdata/basic/secretmanager/v1beta1/secretmanagersecretversion/_vcr_cassettes/nontf.yaml
```diff
@@ -457,15 +457,25 @@ interactions:
...
+            }
         headers:
             Content-Type:
                 - application/json; charset=UTF-8
-        status: 404 Not Found
-        code: 404
-        duration: 232.507687ms
+        status: 200 OK
+        code: 200
+        duration: 245.142672ms
```
gemmahou, line 478: While idk why this changed, but I think the destroy version operation should return 200 instead of 404

gemmahou:
Discussed with Joyce offline, I'll handle this issue in #4621

PR #3763, opened 2025-02-23
chore: create script to generate github actions

justinsb:
- **chore: create script to generate github actions**
- **chore: remove generated tests from manually curated github actions file**
- **chore: rename validate-generated-types.sh to match pattern**

PR #3829, opened 2025-03-03
test: include apigee, secretmanager and vertexai in presubmit

jingyih:
They were split into separate test scripts, but are not run by the presubmit.


justinsb:
FYI https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3763

yuwenma (commented):
/lgtm
/approve

good catch!

PR #3820, opened 2025-02-28
1.129-release note

yuwenma:
### Change description

Fixes #

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #3838, opened 2025-03-03
doc: update release notes

jingyih:


docs/releasenotes/release-1.129.md
```diff
@@ -54,4 +54,4 @@ TODO: list contributors with `git log v1.128.0... | grep Merge | grep from | awk
 
 ## Bug Fixes:
 
-* [SAMPLE_Issue 3007](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3007) ComputeBackendService cannot refer clientTLSPolicy due to invalid format (This is a sample, your actual release note should not contain `SAMPLE_`, otherwise it will be deleted)
\ No newline at end of file
+* The `datasetRef` field in the `BigQueryDataTransferConfig` resource should be [optional](https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/3777) rather than required.
```
yuwenma, line 57: So we have fixed many issues in every release, but we are not aiming at including them all in the release note. Also, I don't think changing this field from required to optional is a bug, it is more of a drift from the rest API, which causes zero regression to Config Connector behavior.  What's more, the 1.129 release note has been published in Google, so it is harder to update  it.

docs/releasenotes/release-1.130.md
```diff
@@ -52,4 +50,4 @@ We have added support for direct reconciliation to more resources, with opt-in b
 
 ## Bug Fixes:
 
-* [SAMPLE_Issue 3007](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3007) ComputeBackendService cannot refer clientTLSPolicy due to invalid format (This is a sample, your actual release note should not contain `SAMPLE_`, otherwise it will be deleted)
\ No newline at end of file
+* Ensure proper [handling](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3825) of `SecretManagerSecretVersion` deletion when the associated GCP resource is already marked as "DESTROYED".
```
yuwenma, line 53: ditto, suggest not adding this bug fix (it causes more user confusion BTW, because SecretVersion is a opt-in feature)

yuwenma (commented):
/approve

Suggest removing two bug fix issues. Otherwise looks good. Could you review https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3820 which is prior to the new release?

docs/releasenotes/release-1.130.md
yuwenma: hmmm, you remind me, it looks like the direct resource has been proved by the customers, maybe it's time to switch it to direct as default. 😄

docs/releasenotes/release-1.129.md
jingyih: Done

docs/releasenotes/release-1.130.md
jingyih: Removed the bug fix entry.

We need more customer like this!

PR #3836, opened 2025-03-03
Refactor: Move KMSCryptoKeyRef to apis folder

gemmahou:
### Change description

Will gradually move other refs and pkg/controller/direct/.../refs to apis folder and follow the identity&reference pattern, so they can be shared by other services.

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #3839, opened 2025-03-04
chore: add a TODO to resolve reference fields in VertexAIFeaturestore

jingyih:
The `KMSKeyRef` need be resolved before making CREATE and UPDATE calls. This was overlooked in the previous PR because the test did not include this field. Let's address this after the KMSRefs refactoring PR (#3836) is merged.


gemmahou:
/lgtm

Thanks!

PR #3848, opened 2025-03-04
Add cmd to add api-enbalement metadata and refactor exec.Command

barney-s:
- refactor fixMetadata to take a function that modifies branchdata
- implement addEnableAPIsModifier branch modifier
- add cmd -3 to invoke fixMetadata with addEnableAPIsModifier
- REFACTOR replace exec.Command calls to use executeCommand util function
   - more modular code
   - consistent logging of inputs, output, err
   - consistent timeout support
 - Add global timeout option
 - Add cmd=4 to enable GCP APIs
 - Moved cmd numbers over to accommodate enable APIs cmd

## Using new command
```shell
cd experiments/conductor
make build

## copy over your branches to branches-user.yaml

## Run the conductor cmd=-3 to create branches-new.yaml with api data
./bin/conductor runner --branch-repo=/usr/local/google/home/barni/workspace/src/github.com/barney-s/vertex-notebook/k8s-config-connector  --branch-conf=branches-user.yaml --logging-dir=logs --command=-3

## review branches-new.yaml and move it to your working file
diff branches-new.yaml branches-user.yaml
mv branches-new.yaml branches-user.yaml
```

## TIPS

* re-run the above command to work on resources that weren't populated in the first run
* It is safe to re-run since we ignore entries that already have apis populated.
* If you see no diff between user and new file, check gcp project in your config. One of the runs changed the project in my configuration
* Inspect logs for cases where you did not see any api being detected.

### DIFF
--- a/experiments/conductor/cmd/runner/mock_commands.go
+++ b/experiments/conductor/cmd/runner/mock_commands.go
@@ -91,7 +91,7 @@ func createScriptYaml(opts *RunnerOptions, branch Branch) {
 		Args:    []string{"--ui-type=prompt", "--prompt=prompt.txt"},
 		WorkDir: workDir,
 	}
-	_, _, err := executeCommand(cfg)
+	_, _, err := executeCommand(opts, cfg)
 	if err != nil {
 		log.Fatal(err)
 	}
@@ -147,7 +147,7 @@ func enableAPIs(opts *RunnerOptions, branch Branch) error {
 			Args:    []string{"services", "enable", api},
 			WorkDir: workDir,
 		}
-		_, _, err := executeCommand(cfg)
+		_, _, err := executeCommand(opts, cfg)
 		if err != nil {
 			return fmt.Errorf("failed to enable API %s: %w", api, err)
 		}
@@ -198,7 +198,7 @@ func captureHttpLog(opts *RunnerOptions, branch Branch) {
 		WorkDir: workDir,
 		Env:     map[string]string{"WRITE_GOLDEN_OUTPUT": "1", "E2E_GCP_TARGET": "real"},
 	}
-	_, _, err := executeCommand(cfg)
+	_, _, err := executeCommand(opts, cfg)
 	if err != nil {
 		log.Printf("TEST GENERATE error: %q\n", err)
 		// Currently ignoring error and just basing on if the _http.log was generated.
@@ -267,7 +267,7 @@ func generateMockGo(opts *RunnerOptions, branch Branch) {
 			},
 			WorkDir: workDir,
 		}
-		output, _, err := executeCommand(cfg)
+		output, _, err := executeCommand(opts, cfg)
 		if err != nil {
 			log.Printf("MOCK SERVICE GENERATE error: %q\n", err)
 			// Currently ignoring error and just basing on if the _http.log was generated.
@@ -306,7 +306,7 @@ func generateMockGo(opts *RunnerOptions, branch Branch) {
 			},
 			WorkDir: workDir,
 		}
-		output, _, err := executeCommand(cfg)
+		output, _, err := executeCommand(opts, cfg)
 		if err != nil {
 			log.Printf("MOCK RESOURCE GENERATE error: %q\n", err)
 			// Currently ignoring error and just basing on if the _http.log was generated.
@@ -371,7 +371,7 @@ func addServiceToRoundTrip(opts *RunnerOptions, branch Branch) {
 		Args:    []string{"--ui-type=prompt", "--prompt=roundtrip_prompt.txt"},
 		WorkDir: workDir,
 	}
-	_, _, err := executeCommand(cfg)
+	_, _, err := executeCommand(opts, cfg)
 	if err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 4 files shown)


experiments/conductor/cmd/runner/mock_commands.go
```diff
@@ -147,6 +134,27 @@ If you have problems, please output a JSON result like this:
...
+	for _, api := range branch.ApisEnabled {
+		cfg := CommandConfig{
+			Name:    fmt.Sprintf("Enable API %s", api),
+			Cmd:     "gcloud",
+			Args:    []string{"services", "enable", api},
+			WorkDir: workDir,
+		}
+		_, _, err := executeCommand(cfg)
+		if err != nil {
+			return fmt.Errorf("failed to enable API %s: %v", api, err)
```
cheftako, line 152: Should be %w not %v when using fmt.Errorf

experiments/conductor/cmd/runner/utilities.go
```diff
@@ -267,7 +268,7 @@ type CommandConfig struct {
 }
 
 // Helper function to execute a command with timing and logging
-func executeCommand(cfg CommandConfig, out *strings.Builder, stderr *strings.Builder) error {
+func executeCommand(cfg CommandConfig) (string, string, error) {
```
cheftako, line 271: If we want to make the timeout configurable, it would probably be more concise to just pass Options in to the executeCommand function and let it (here) determine the over-ride. Right now we are passing timeout through the CommandConfig struct, which would require us to extract the field at every call site and set it in the CommandConfig before calling us.

experiments/conductor/cmd/runner/mock_commands.go
```diff
@@ -171,28 +179,31 @@ func captureHttpLog(opts *RunnerOptions, branch Branch) {
 		return
 	}
 
+	// Enable required APIs before running tests
+	if err := enableAPIs(opts, branch); err != nil {
```
cheftako, line 183: Would it make more sense to just make enableAPIs a top level command? Then you call it once per project rather than every time?

experiments/conductor/cmd/runner/runner.go
```diff
@@ -123,9 +123,13 @@ type Branch struct {
 	ProtoMsg  string `yaml:"proto-msg"`     // google.ai.generativelanguage.v1beta.Model
 	HostName  string `yaml:"host-name"`     // generativelanguage.googleapis.com
 
-	Notes []string `yaml:"notes"` // Observation goes here
+	Notes       []string `yaml:"notes"`        // Observation goes here
+	ApisEnabled []string `yaml:"apis-enabled"` // GCP APIs that need to be enabled
```
cheftako, line 127: when possible can we add a concrete example for typed data? Notes is more free form so its not a great example. The other fields have actual field values to make it more clear.

experiments/conductor/cmd/runner/runner.go
```diff
@@ -156,10 +160,12 @@ func RunRunner(ctx context.Context, opts *RunnerOptions) error {
 	}
 
 	switch opts.command {
+	case -3:
```
cheftako, line 163: Is the plan to get each person generating a group of mocks to run this on their work set?

experiments/conductor/cmd/runner/runner.go
```diff
@@ -465,9 +469,71 @@ func fixMetadata(opts *RunnerOptions, branches Branches) {
...
+	if branch.ApisEnabled != nil && len(branch.ApisEnabled) > 0 {
+		return branch
+	}
+
+	cfg := CommandConfig{
+		Name:    "API Discovery",
+		Cmd:     "codebot",
+		Args:    []string{"--prompt=/dev/stdin"},
+		WorkDir: workDir,
+		Stdin: strings.NewReader(fmt.Sprintf(`Given the gcloud command %q, what Google Cloud APIs need to be enabled to use this command?
```
cheftako, line 495: :)

experiments/conductor/cmd/runner/utilities.go
barney-s: done. Added a second commit to set global timeout and pass it thorough.

experiments/conductor/cmd/runner/runner.go
barney-s: Yes delegate updating their copy of branch-user.yaml 
The tool generates branch-new.yaml which can be inspected and renamed as the branch-user.yaml

experiments/conductor/cmd/runner/mock_commands.go
barney-s: done added a command to do that. 
Had to move commands over to new numbers.

experiments/conductor/cmd/runner/runner.go
barney-s: done

experiments/conductor/cmd/runner/runner.go
```diff
@@ -46,15 +47,16 @@ conductor runner --branch-repo=/usr/local/google/home/wfender/go/src/github.com/
...
-	cmdCreateScriptYaml    = 4
-	cmdCaptureHttpLog      = 5
-	cmdGenerateMockGo      = 6
-	cmdAddServiceRoundTrip = 7
-	cmdAddProtoMakefile    = 8
-	cmdRunMockTests        = 9
-	cmdGenerateTypes       = 10
-	cmdGenerateCRD         = 11
-	cmdGenerateFuzzer      = 12
+	cmdEnableGCPAPIs       = 4
```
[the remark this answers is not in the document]
barney-s: Will send a PSA once the merge is done.

PR #3849, opened 2025-03-04
tool: run conductor within a single workspace

jingyih:
Enable the conductor tool to run within a single workspace.

- Build and install dependent tools in `$(GOPATH)/bin/`. 
- Set default flag values to eliminate the need for manual specification when running.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -163,3 +163,4 @@ log
 # experiments/conductor
 experiments/conductor/logs
 experiments/conductor/branches.yaml
+experiments/conductor/bin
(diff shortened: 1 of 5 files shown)


yuwenma (commented):
/lgtm
/approve

Nice!

jingyih:
Just had a quick offline chat with @cheftako, and we discussed the pros and cons of using two workspaces versus one. Having two workspaces makes it easier to modify tools while working on resources. My personal take is that, for contributors (especially new contributors) who primarily use the tool rather than modify it, a single workspace setup is more straightforward. In that case, this PR could be beneficial.

/hold

jingyih:
/close

Most people (including myself) now runs conductor using multiple workspaces.

PR #3853, opened 2025-03-05
mockgcp: aigenerativelanguage

acpana:
Parking the mock for aigenerativelanguage.

This was mostly done with conductor.


mockgcp/mockaigenerativelanguage/model.go
```diff
@@ -0,0 +1,76 @@
...
+		if status.Code(err) == codes.NotFound {
+			return nil, status.Errorf(codes.NotFound, "Model %v not found.", name)
+		}
+		return nil, err
+	}
+	return obj, nil
+}
+
+// TODO acpana there is not Delete or Create in the pb
+
```
acpana, line 56: there is no Delete/ CreateModel but somehow the gcloud commands do use something! I think the paths may not be exposed in the protos

PR #3863, opened 2025-03-05
Add skip field to branches

barney-s:


maqiuyujoyce (commented):
/lgtm
/approve

experiments/conductor/cmd/runner/runner.go
```diff
@@ -114,6 +114,7 @@ type Branch struct {
 	Group      string `yaml:"group"`      // ai
 	Resource   string `yaml:"resource"`   // model
 	Controller string `yaml:"controller"` // Unknown
+	Skip       bool   `yaml:"skip"`       // Skip this branch in processing
```
acpana, line 117: I was thinking if we wanted to put a reason field or insist that a notes field be populated in this case!

barney-s: If for documenting purpose, we can add an yam comment. `// some reason`. If we want tooling to use the reason (say for logging), we need what you are suggesting.

PR #3864, opened 2025-03-05
Conductor: Add support for retries for failed commands

barney-s:
Add support for retries for failed commands.
- each command can cap the retry count using MaxRetries. This is useful for deterministic commands.
- global retry flag (default: 3) is used for all commands.

### DIFF
--- a/experiments/conductor/cmd/runner/runner.go
+++ b/experiments/conductor/cmd/runner/runner.go
@@ -92,6 +92,8 @@ func BuildRunnerCmd() *cobra.Command {
 		"", "", "dedicated directory for logging, empty for stdout.")
 	cmd.Flags().DurationVarP(&opts.timeout, "timeout",
 		"t", 5*time.Minute, "Global timeout for commands.")
+	cmd.Flags().IntVarP(&opts.defaultRetries, "retries",
+		"r", 10, "Default number of retries for failed commands.")
 
 	return cmd
 }
@@ -103,9 +105,13 @@ type RunnerOptions struct {
 	loggingDir     string
 	timeout        time.Duration
 	readFileType   string
+	defaultRetries int // Default number of retries for commands
 }
 
 func (opts *RunnerOptions) validateFlags() error {
+	if opts.defaultRetries < 0 {
+		return fmt.Errorf("retries flag cannot be negative, got %d", opts.defaultRetries)
+	}
 	return nil
 }
 
@@ -559,6 +565,7 @@ If you are using gcloud to create something, make sure you are deleting it after
 Print the list of APIs, one on each line in this format api-required: <api-name>
 Only include APIs that are directly needed by this command.
 `, branch.Command)),
+		RetryBackoff: GenerativeCommandRetryBackoff,
 	}
 
 	output, _, err := executeCommand(opts, cfg)
@@ -645,10 +652,11 @@ func inferProtoPath(opts *RunnerOptions, branch Branch, workDir string) string {
 	}
 
 	cfg := CommandConfig{
-		Name:    "Search for service",
-		Cmd:     "egrep",
-		Args:    args,
-		WorkDir: workDir,
+		Name:       "Search for service",
+		Cmd:        "egrep",
+		Args:       args,
+		WorkDir:    workDir,
+		MaxRetries: 2,
 	}
 	output, errOutput, err := executeCommand(opts, cfg)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


maqiuyujoyce (commented):
Thank you for adding the retries, @barney-s !

I have a few thoughts/questions:

1. The place where I found retries are needed is when the communication with LLM hangs/terminates unexpectedly. We talk to codebot for steps like script generation, API identification (right?), mockgcp generation. So I'm thinking we may want to focus on retries about these steps. And in this PR, I didn't find the retry for creating scripts. As I shared, I had to go through 12 retries to finish generating scripts for 40 resources.
2. I think with max retries of 2, the retry may not be that useful, especially you set the default retry to be 3. I'd probably start the max retries to be at least 10 considering we are doing batch work.
3. Have you tried to verify the retry works for steps involving using codebot for more than 50 resources?

barney-s:
> 1. The place where I found retries are needed is when the communication with LLM hangs/terminates unexpectedly. We talk to codebot for steps like script generation, API identification (right?), mockgcp generation. So I'm thinking we may want to focus on retries about these steps. And in this PR, I didn't find the retry for creating scripts. As I shared, I had to go through 12 retries to finish generating scripts for 40 resources.

We can bump the default retries to 10 ? For LLM steps i have not set maxRetries. So it would take the user passed/default retries count for those.

> 2. I think with max retries of 2, the retry may not be that useful, especially you set the default retry to be 3. I'd probably start the max retries to be at least 10 considering we are doing batch work.

maxRetries is an optional per command max that is hardcoded. The goal is not to retry certain deterministic commands N times and instead cap them to maxRetries. Re your suggestion 2 is as good as 1. I tend to agree. Just that it tries once more. 

> 3. Have you tried to verify the retry works for steps involving using codebot for more than 50 resources?

Have not yet. Will do and update here.

experiments/conductor/cmd/runner/utilities.go
```diff
@@ -296,61 +298,90 @@ func executeCommand(opts *RunnerOptions, cfg CommandConfig) (string, string, err
...
+		printCommandOutput(output)
+		if errBuf.Len() > 0 {
+			log.Printf("[%s] stderr output:", cfg.Name)
+			printCommandOutput(errOutput)
+		}
 
-	output := outBuf.String()
-	errOutput := errBuf.String()
+		retryCount++
+		if maxRetries >= 0 && retryCount > maxRetries {
```
maqiuyujoyce, line 377: Nit: Looks like the maxRetries is either opts.defaultRetries or the hard coded maxRetries, so `if retryCount > maxRetries {` seems to be sufficient. Otherwise when opts.defaultRetries is set to a negative value, then this command will never stop by itself.

experiments/conductor/cmd/runner/utilities.go
```diff
@@ -296,61 +298,90 @@ func executeCommand(opts *RunnerOptions, cfg CommandConfig) (string, string, err
 		}
 	}
 
-	log.Printf("Starting command step: %s", cfg.Name)
-	log.Printf("[%s] working directory: %s", cfg.Name, cfg.WorkDir)
-	log.Printf("[%s] command: %s %s", cfg.Name, cfg.Cmd, strings.Join(cfg.Args, " "))
-	if cfg.Stdin != nil {
-		log.Printf("[%s] stdin: %s", cfg.Name, cfg.Stdin)
+	if cfg.RetryBackoff == 0 {
+		cfg.RetryBackoff = time.Second
```
maqiuyujoyce, line 302: Nit: If this is related to quota then I guess one second of backoff time may not be sufficient. But we can improve it later.

experiments/conductor/cmd/runner/runner.go
```diff
@@ -82,6 +82,8 @@ func BuildRunnerCmd() *cobra.Command {
 		"", "", "dedicated directory for logging, empty for stdout.")
 	cmd.Flags().DurationVarP(&opts.timeout, "timeout",
 		"t", 5*time.Minute, "Global timeout for commands.")
+	cmd.Flags().IntVarP(&opts.defaultRetries, "retries",
+		"r", 10, "Default number of retries for failed commands.")
```
maqiuyujoyce, line 86: Nit: We should suggest increasing timeout when the number of retries here is large.

maqiuyujoyce (commented):
/lgtm
/approve

experiments/conductor/cmd/runner/utilities.go
barney-s: changed it to 30s for generative commands.

experiments/conductor/cmd/runner/utilities.go
barney-s: added a check to make sure defaultretries is not -ve

PR #3870, opened 2025-03-06
fix: Fix SQLInstance replica periodic

jasonvigil:
This test has the same issue as the other SQLInstance backup tests, because backup configuration is required to configure a replica.


jingyih (commented):
/lgtm
/approve

jingyih (commented):
/lgtm

PR #3875, opened 2025-03-06
fix: Add organization dependencies for Apigee tests

jasonvigil:
This should allow the e2e tests to provision their organization resources for each run, making it possible to run the apigee integration tests.  However, this will require running each Apigee test case in a separate project (so that each test can provision its own organization, as there is only one organization allowed per project).


jasonvigil:
/hold

This is untested in actual periodic tests. Not sure if this will work completely, but it does capture the work that I have done to define the resources needed to manually deploy an ApigeeOrganization.

PR #3876, opened 2025-03-06
Remove log.Fatal from conductor cmds

anhdle-sso:
### Change description

Make a few conductor cmd to ignore error and move to the next branch

### Tests you have done

### DIFF
--- a/experiments/conductor/cmd/runner/utilities.go
+++ b/experiments/conductor/cmd/runner/utilities.go
@@ -15,6 +15,7 @@
 package runner
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


experiments/conductor/cmd/runner/utilities.go
```diff
@@ -191,6 +192,25 @@ func gitCommit(workDir string, out *strings.Builder, msg string) {
 	out.Reset()
 }
 
+func gitFileHasChange(workDir string, filePath string) bool {
```
barney-s, line 195: Should we move this to github_commands.go 
and use startBash() ?

anhdle-sso: This is a git command not github. I am using the cmd exec to keep it similar to other git specific cmd such as gitAdd.

PR #3880, opened 2025-03-06
chore: mark mockgcptests as parallel-safe

justinsb:
We use env vars and proxies on unique ports


yuwenma (commented):
/approve

Shall we also set the `parallel` flag?   https://github.com/golang/go/blob/182bdbb1e1d62fe53f8386e7b02c7738898d373b/src/testing/testing.go#L180

PR #3884, opened 2025-03-06
mockGCP: Resource bigtable appprofile

gemmahou:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/tests/e2e/httplog.go
+++ b/tests/e2e/httplog.go
@@ -213,6 +213,7 @@ func (x *Normalizer) Render(events test.LogEntries) string {
 	// Remove headers that just aren't very relevant to testing
 	// Remove headers in request.
 	events.RemoveHTTPRequestHeader("X-Goog-Api-Client")
+	events.RemoveHTTPRequestHeader("X-Goog-User-Project")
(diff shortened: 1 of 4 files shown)


tests/e2e/httplog.go
```diff
@@ -213,6 +213,7 @@ func (x *Normalizer) Render(events test.LogEntries) string {
 	// Remove headers that just aren't very relevant to testing
 	// Remove headers in request.
 	events.RemoveHTTPRequestHeader("X-Goog-Api-Client")
+	events.RemoveHTTPRequestHeader("X-Goog-User-Project")
```
yuwenma, line 216: mind sharing more context about this change? My main question is why does the app profile resource requires this header (which is specifically for quota and billing)

yuwenma (commented):
/approve

One question

tests/e2e/httplog.go
gemmahou: I do not think this header is required for golden log comparison so I removed it. This is also fixed in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3937/files

PR #3885, opened 2025-03-06
conductor: script yaml generation improvements

barney-s:
- tweak the prompt to use correct projectId variable
- update readFiles for script yaml to produce structured output under logs/all_scripts.yaml [BULK READ]
- The user can inspect and manually edit all_scripts.yaml [MANUAL EDIT]
- add writeFiles to write the contents from all_scripts.yaml to each branch. [BULK_WRITE]
- If there are no changes it skips the update
- Update proxy to allow wildcard googleapis. This avoid a TLS warning in the logs

### DIFF
--- a/experiments/conductor/cmd/runner/mock_commands.go
+++ b/experiments/conductor/cmd/runner/mock_commands.go
@@ -95,7 +95,7 @@ func createScriptYaml(opts *RunnerOptions, branch Branch) {
 	// Check to see if the script file already exists
 	scriptFile := fmt.Sprintf("mock%s/testdata/%s/crud/script.yaml", branch.Group, branch.Resource)
 	scriptFullPath := filepath.Join(opts.branchRepoDir, "mockgcp", scriptFile)
-	if _, err := os.Stat(scriptFullPath); !errors.Is(err, os.ErrNotExist) {
+	if _, err := os.Stat(scriptFullPath); !errors.Is(err, os.ErrNotExist) && !opts.force {
 		log.Printf("SKIPPING %s, %s already exists", branch.Name, scriptFullPath)
 		return
 	}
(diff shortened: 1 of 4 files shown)


experiments/conductor/cmd/runner/mock_commands.go
```diff
@@ -160,25 +178,181 @@ func readScriptYaml(opts *RunnerOptions, branch Branch) {
...
+
+	for _, line := range scriptContent {
+		if cmd, ok := line["exec"]; ok && strings.Contains(cmd, "gcloud") {
+			if strings.Contains(cmd, " create ") {
+				hasCreate = true
+			}
+			if strings.Contains(cmd, " delete ") {
+				hasDelete = true
+			}
+			if strings.Contains(cmd, " list ") {
```
maqiuyujoyce, line 210: Nit: It's possible that the generated list command is followed by no parameter, so can we check `" list"` (no space after `list`) instead?

experiments/conductor/cmd/runner/mock_commands.go
```diff
@@ -160,25 +178,181 @@ func readScriptYaml(opts *RunnerOptions, branch Branch) {
...
-	if strings.Contains(dataStr, " list") {
-		fmt.Println("WARNING: Script contains a LIST command")
+	if hasList {
+		warning := "List found"
+		warnings = append(warnings, warning)
+		log.Printf("WARNING: %s in %s", warning, scriptFile)
 	}
-	fmt.Println(dataStr)
+
+	// Read existing all_scripts.yaml or create new ScriptData
```
maqiuyujoyce, line 232: I think when I read, I only want the latest information based on the list of metadata entries. I do not want to read information other than the entries I provided in the branch yaml. Otherwise I'd add those entries. So I think overriding the entire file may be a more intuitive option here?

experiments/conductor/cmd/runner/mock_commands.go
barney-s: done.

experiments/conductor/cmd/runner/mock_commands.go
barney-s: iam thinking with the --for-resource, this may not be true anymore ? 
Because we filter in only those resources. Iam ok to change this. May be we will try these out and update afterwards.

maqiuyujoyce (commented):
/lgtm
/approve

PR #3899, opened 2025-03-07
Support pre and post steps in gcloud script yaml

barney-s:
Support pre and post steps in gcloud script yaml.
We dont use proxy for pre and post commands. 

We dont yet sort the commands. We want to run them as they appear in the script in order. 

```
        - pre: gcloud pubsub topics create test-topic-${uniqueId} --project=${projectId}
        - exec: gcloud asset feeds create test-${uniqueId} --pubsub-topic=projects/${projectId}/topics/test-topic-${uniqueId} --project=${projectId}
        - exec: gcloud asset feeds describe test-${uniqueId} --project=${projectId}
        - exec: gcloud asset feeds update test-${uniqueId} --project=${projectId} --content-type=resource
        - exec: gcloud asset feeds delete test-${uniqueId} --project=${projectId}
        - post: gcloud pubsub topics delete test-topic-${uniqueId} --project=${projectId}
```  

This is a backward compatible change. Existing scripts continue to work. While we enhance LLM generation to add pre and post steps. At which point the code would start using those.

### DIFF
--- a/mockgcp/mockgcptests/e2e_test.go
+++ b/mockgcp/mockgcptests/e2e_test.go
@@ -74,8 +74,22 @@ func TestScripts(t *testing.T) {
 			h.StartProxy(ctx)
 
 			for _, step := range script.Steps {
-				if step.Exec != "" {
-					cmd := exec.CommandContext(ctx, "bash", "-c", step.Exec)
+				stepCmd := ""
+				stepType := ""
+				needProxy := false
+				if step.Pre != "" {
+					stepCmd = step.Pre
+					stepType = "pre"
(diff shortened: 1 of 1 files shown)


anhdle-sso:
/lgtm

Discussed offline with Barni, the feature is for readability for now. It can be useful so we can make sure the clean up (post) always happen.

yuwenma (commented):
/hold

My concern is that breaks the LLM reproduction on generating the `script.sh`. For example, how does LLM know which `exec` should use proxy or not? Also, how does engineer know that?

yuwenma:
Concern resovled in #3901

PR #3902, opened 2025-03-07
[WIP] Relanding spannerinstace forcedelete

anhdle-sso:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/pkg/controller/direct/spanner/instance_controller.go
+++ b/pkg/controller/direct/spanner/instance_controller.go
@@ -16,6 +16,7 @@ package spanner
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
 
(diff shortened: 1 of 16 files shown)


pkg/controller/direct/spanner/instance_controller.go
```diff
@@ -283,6 +293,14 @@ func (a *SpannerInstanceAdapter) Export(ctx context.Context) (*unstructured.Unst
 func (a *SpannerInstanceAdapter) Delete(ctx context.Context, deleteOp *directbase.DeleteOperation) (bool, error) {
 	log := klog.FromContext(ctx).WithName(ctrlName)
 	log.V(2).Info("deleting Instance", "name", a.id)
+	v, ok := a.desired.ObjectMeta.Annotations["cnrm.cloud.google.com/force-destroy"]
```
yuwenma, line 296: First of all, we want to deprecate those cnrm annotations when migrating to the direct approach, including force destroy. 

And including deleting another resource here will cause conflict. I suggest we report back errors and ask the users to manually delete the backup here.

anhdle-sso: I agree with all of that. This is the change that we agree to do to unblock spanner. This change has landed before but got revert due to github action breaking the HEAD. We can remove this change when the Backup resource is landed.

anhdle-sso:
/hold

Discussed offline, we will keep the standard of one controller to one resource. Will return delete error instead.

PR #3904, opened 2025-03-07
TPUVirtualMachine: controller

justinsb:
- **TPUVirtualMachine: Create controller client**
- **TPUVirtualMachine: Create controller**
- **TPUVirtualMachine: Create controller identity**
- **TPUVirtualMachine: Create minimal test**
- **TPUVirtualMachine: Support for testing with mockgcp**
- **chore: fix compilation error in controllerbuilder CTF**
- **mockgcp: don't remap tpu v2 to mockgcp**
- **chore: generate gcp clients for tpu.v2**
- **chore: use generated gcp clients for tpu.v2**
- **tests: golden output for tpuvirtualmachine-minimal**


justinsb:
/assign @yuwenma 

(But I suspect I'll have to do a few loops before this passes testing)

The most interesting thing here is that we have the proto, but not the client.  So to unblock ourselves, we compile the proto ourselves and build our own simple GRPC client  This could be a good trick e.g. if we want to change the protos.

pkg/controller/direct/tpu/tpuvirtualmachine_controller.go
```diff
@@ -0,0 +1,338 @@
...
+	if err != nil {
+		return nil, err
+	}
+	tpuClient, err := gcpClient.newClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	mapCtx := &direct.MapContext{}
+	desiredProto := TPUVirtualMachineSpec_ToProto(mapCtx, &obj.Spec)
```
yuwenma, line 89: Suggest moving this call to Create and Update to preserve the actual KRM value for some key information like defaulting, unset/unmanaged, and "OUTPUT-ONLY" name that is required in Create.

pkg/controller/direct/tpu/tpuvirtualmachine_controller.go
```diff
@@ -0,0 +1,338 @@
...
+// Return a non-nil error requeues the requests.
+func (a *TPUVirtualMachineAdapter) Find(ctx context.Context) (bool, error) {
+	log := klog.FromContext(ctx)
+	log.V(2).Info("getting TPUVirtualMachine", "name", a.id)
+
+	req := &pb.GetNodeRequest{
+		Name: a.id.String(),
+	}
+
+	// Because we are using a hand-coded GRPC client, we need to add the headers per https://google.aip.dev/client-libraries/4222
```
yuwenma, line 148: 👍

pkg/controller/direct/tpu/tpuvirtualmachine_controller.go
```diff
@@ -0,0 +1,338 @@
...
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+	status.ExternalRef = direct.LazyPtr(a.id.String())
+
+	var readyCondition *kcc.Condition
+	switch created.GetState() {
+	case pb.Node_CREATING:
+		// We're not actually done yet!
+		// Should we set readyCondition?
```
yuwenma, line 216: Yes, how about "Updating"?

yuwenma (commented):
/approve

yuwenma:
> /assign @yuwenma
> 
> (But I suspect I'll have to do a few loops before this passes testing)
> 
> The most interesting thing here is that we have the proto, but not the client. So to unblock ourselves, we compile the proto ourselves and build our own simple GRPC client This could be a good trick e.g. if we want to change the protos.

This is fantastic! (But I hope we don't need to do this a lot)

ldanielmadariaga:

PR #3918, opened 2025-03-08
conductor: always log to stderr, even when also logging to file

justinsb:


yuwenma (commented):
/lgtm
/approve

Good to learn this! Thanks!

PR #3923, opened 2025-03-09
mockgcp tests: only capture events from "normal" steps

justinsb:
We still need the proxy for pre and post, we just don't want to capture the events.


barney-s:
I will quickly verify if this works for my case (metastore service).

yuwenma (commented):
/lgtm
/approve

Thanks!
cc @barney-s

PR #3934, opened 2025-03-09
mockgcp: stop using api overrides, rely on CONNECT proxy

justinsb:
The API overrides are not implemented everywhere.

Also we have observed some oddities with gcloud behaving differently
with api_endpoint_overrides, such as the URLs in the compute APIS.

(For example, gcloud compute instance-templates create foo)

### DIFF
--- a/mockgcp/mockgcptests/proxy.go
+++ b/mockgcp/mockgcptests/proxy.go
@@ -217,15 +217,6 @@ func (p *Proxy) runRequest(req *http.Request) (*http.Response, error) {
 		return nil, fmt.Errorf("reading request body: %v", err)
 	}
 
-	// HACK: fix malformed network URLs in the request body
-	if strings.Contains(string(body), "\"network\": \"http://compute.googleapis.com/") {
-		bodyStr := string(body)
-		bodyStr = strings.Replace(bodyStr,
-			"\"network\": \"http://compute.googleapis.com/",
-			"\"network\": \"", -1)
-		body = []byte(bodyStr)
-	}
-
 	u := req.URL
 	u.Scheme = "https"
 	if u.Host == "" {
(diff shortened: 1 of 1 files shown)


justinsb:
I would not expect it to.  But OTOH the tests are currently not passing, so ...

mockgcp/mockbackupdr/testdata/managementserver/crud/_http.log
```diff
@@ -112,7 +112,7 @@ Content-Type: application/json
   "address": "8.8.8.8",
   "addressType": "INTERNAL",
   "name": "google-managed-services-${uniqueId}",
-  "network": "projects/${projectId}/global/networks/test-network-${uniqueId}",
+  "network": "https://compute.googleapis.com/compute/v1/projects/${projectId}/global/networks/test-network-${uniqueId}",
```
jingyih, line 115: I think removing the hack will cause the HTTP request to fail because the "network" field is incorrectly formatted?

yuwenma:
The presubmit failure comes from an issue that has been fixed in ComposerEnvironment change. Could you rebase from master HEAD?

mockgcp/mockbackupdr/testdata/managementserver/crud/_http.log
jingyih: I'm not sure why gcloud behaves differently. When testing locally (e.g., in my terminal), the HTTP request sent by the gcloud command is correct, with `"network": "projects/${projectId}/global/networks/test-network-${uniqueId}"`. However, in our `mockgcptests/e2e_test`, gcloud sends an incorrect HTTP request (after proxied) where `"network"` appears as `"https://compute.googleapis.com/compute/v1/projects/${projectId}/global/networks/test-network-${uniqueId}"`. This format is rejected by the GCP API as invalid.

justinsb: Ah so I think this is actually what I'm trying to fix then!  There are two problems with api overrides: 1) not all services implement it, so we need a CONNECT proxy.  But also and more oddly (2) I noticed something similar where gcloud was constructing a URL incorrectly ... I think it was also a compute URL.  There is some fancy logic happening in gcloud to convert URLs (it appears), and I think that api_overrides changes that logic, which isn't what we want in a mock.

The easy fix therefore is to switch to the CONNECT proxy, which (hopefully) should not change gcloud behaviour because it is at the transport layer.  It is much more complicated to implement (and my initial implementation leaves much to be desired), but once we have it it seems more reliable!

yuwenma:
/lgtm
/hold cancel 

Thank you!

PR #3935, opened 2025-03-09
mockgcp: for api quotapreference

barney-s:


mockgcp/mockapi/quotapreference.go
```diff
@@ -0,0 +1,129 @@
...
+	obj := &pb.QuotaPreference{}
+	if err := s.storage.Get(ctx, fqn, obj); err != nil {
+		return nil, err
+	}
+	paths := req.GetUpdateMask().GetPaths()
+	if len(paths) == 0 {
+		return nil, status.Errorf(codes.InvalidArgument, "update_mask must be provided")
+	}
+
+	now := time.Now()
```
justinsb, line 91: Nit: I prefer to loop over the paths ... the more examples we have that "do the wrong thing" (I.e. proto.Merge), the more confused the LLM will be

mockgcp/mockapi/quotapreference.go
```diff
@@ -0,0 +1,129 @@
...
+
+// parseQuotaPreferenceName parses a string into a quotaPreferenceName.
+// The expected form is `projects/*/locations/*/quotaPreferences/*`.
+func (s *MockService) parseQuotaPreferenceName(name string) (*quotaPreferenceName, error) {
+	tokens := strings.Split(name, "/")
+
+	if len(tokens) == 6 && tokens[2] == "locations" && tokens[4] == "quotaPreferences" {
+		if tokens[0] == "projects" || tokens[0] == "folders" || tokens[0] == "organizations" {
+			name := &quotaPreferenceName{
+				Parent:              strings.Join(tokens[0:4], "/"),
```
justinsb, line 120: Interesting ... did you code this, or did the LLM?  I don't know if we have examples of this, but it looks right 👍  (Assuming we don't want to verify the project/folder/org, which seems reasonable for a mock)

justinsb:
Looks correct, assuming that the real and mock output is the same (or close enough)

mockgcp/mockapi/quotapreference.go
barney-s: LLM :)

mockgcp/mockapi/quotapreference.go
barney-s: fixing.

PR #3937, opened 2025-03-09
mockgcp tests: ignore x-goog-user-project header

justinsb:
This header is (always) sent by gcloud if a quota project is set,
so this header reflects configuration, not the actual protocol.


barney-s:
/approve
/lgtm

seeing similar issue with cloudquota

gemmahou:
+1 seeing the same issue in bigtable appprofile

yuwenma:
/approve

Thanks for the comment. I raised the question in some PR reviews. Now I see where this comes from.

PR #3946, opened 2025-03-10
Set `ReturnPartialSuccess=true` in GKE Hub GetFeature call to ensure partial results are returned when a region is down

### LINKED ISSUE #3968: Enable gkehub GetFeature to return partial results when a cloud region is unreachable
### Checklist

The `ReturnPartialSuccess` flag allows the gkehub GetFeature call to return partial results when a cloud region is unreachable, instead of returning an internal error. 

We should set this flag in featuremembership_controller to allow the call to success even when a region is down

### Additional information

_No response_

### Importance

_No response_

yufan-su:
### Change description

Fixes #3968

### Tests you have done


yuwenma:
/assign @acpana 

cc @haiyanmeng 
Could you share more context about this?

yufan-su:
yes added the issue in #3968

pkg/controller/direct/gkehub/featuremembership_controller.go
```diff
@@ -145,7 +145,7 @@ func (a *gkeHubAdapter) Find(ctx context.Context) (bool, error) {
 	if a.membershipID == "" || a.featureID == "" {
 		return false, nil
 	}
-	feature, err := a.hubClient.featureClient.Get(a.featureID).Context(ctx).Do()
+	feature, err := a.hubClient.featureClient.Get(a.featureID).Context(ctx).ReturnPartialSuccess(true).Do()
```
acpana, line 148: Two non blocking thoughts

- we might want to give a shout out about this in the release notes
- (less likely) it would be nice if we could port this over to the mockgcp implementation

On that last note, let's coordinate once this PR lands to make sure that post submits are not upset internally.

acpana:
/lgtm
/approve

/hold

@yuwenma for final approval

pkg/controller/direct/gkehub/featuremembership_controller.go
yufan-su: sounds good, yes I will mention this change in the release note

PR #3948, opened 2025-03-11
Update spanner instance edition logic for direct controller

anhdle-sso:
### Change description

- Default Edition to STANDARD as EDITION_UNSPECIFIED no longer allowed.
- Add logic for Edition downgrade:
    - Make a separate request for non edition fields (if there is any).
    - Make a final call for edition downgrade, this can fail because more premium features are being used. The error will be communicated back to the user (which feature needs to be disable).

### Tests you have done

### DIFF
--- a/apis/spanner/v1beta1/instance_types.go
+++ b/apis/spanner/v1beta1/instance_types.go
@@ -55,7 +55,6 @@ type SpannerInstanceSpec struct {
 	// Optional. The `Edition` of the current instance.
 	// Currently accepted values are STANDARD, ENTERPRISE, ENTERPRISE_PLUS
 	// If edition is unspecified, the default value is STANDARD.
-	// Downgrading edition is not supported for the moment.
 	// +optional
 	Edition *string `json:"edition,omitempty"`
 
--- a/apis/spanner/v1beta1/instance_types.go
+++ b/apis/spanner/v1beta1/instance_types.go
@@ -53,8 +53,8 @@ type SpannerInstanceSpec struct {
 	AutoscalingConfig *AutoscalingConfig `json:"autoscalingConfig,omitempty"`
 
 	// Optional. The `Edition` of the current instance.
-	// Currently accepted values are STANDARD, ENTERPRISE, ENTERPRISE_PLUS
-	// If edition is unspecified, the default value is STANDARD.
+	// Currently accepted values are STANDARD, ENTERPRISE, ENTERPRISE_PLUS.
+	// If edition is unspecified, it has automatically upgraded to the lowest edition that matches your usage pattern.
 	// +optional
 	Edition *string `json:"edition,omitempty"`
 
--- a/apis/spanner/v1beta1/instance_types.go
+++ b/apis/spanner/v1beta1/instance_types.go
@@ -53,7 +53,9 @@ type SpannerInstanceSpec struct {
 	AutoscalingConfig *AutoscalingConfig `json:"autoscalingConfig,omitempty"`
 
 	// Optional. The `Edition` of the current instance.
-	// Currently accepted values are EDITION_UNSPECIFIED, STANDARD, ENTERPRISE, ENTERPRISE_PLUS
+	// Currently accepted values are STANDARD, ENTERPRISE, ENTERPRISE_PLUS
+	// If edition is unspecified, the default value is STANDARD.
+	// Downgrading edition is not supported for the moment.
 	// +optional
 	Edition *string `json:"edition,omitempty"`
 
--- a/pkg/controller/direct/spanner/instance_controller.go
+++ b/pkg/controller/direct/spanner/instance_controller.go
@@ -151,6 +151,14 @@ func (a *SpannerInstanceAdapter) Create(ctx context.Context, createOp *directbas
 		return err
 	}
 	resource := SpannerInstanceSpec_ToProto(mapCtx, &desired.Spec, a.id.SpannerInstanceConfigPrefix())
+
+	// if edition is unspecified, Spanner API default edition to STANDARD.
+	// Spanner API bans downgrade edition from STANDARD to unspecifed,
+	// which cause infinite reconcile loop.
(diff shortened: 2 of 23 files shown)


pkg/controller/direct/spanner/spannerinstace_mapper.go
```diff
@@ -93,6 +93,12 @@ func SpannerInstanceSpec_ToProto(mapCtx *direct.MapContext, in *krm.SpannerInsta
 	out.DisplayName = in.DisplayName
 	out.NodeCount = direct.ValueOf(in.NumNodes)
 	out.ProcessingUnits = direct.ValueOf(in.ProcessingUnits)
+	// if edition is unspecified, Spanner API default edition to STANDARD.
+	// Spanner API bans downgrade edition from STANDARD to unspecifed,
+	// which cause infinite reconcile loop.
```
yuwenma, line 98: Good to know.
Is "unspecified" allowed? In other words, does the server give a default value when creating a SpannerInstance without the edition? Is it allowed to downgrade edition to other values?

yuwenma (commented):
/approve

Looks good. One question about whether we should consider to allow unsetting the field.

pkg/controller/direct/spanner/spannerinstace_mapper.go
anhdle-sso: Unspecicied value is no longer allowed. STANDARD is the new default. Downgrading is technically possible but I plan to block it due to the constrains (from other resources).

anhdle-sso:
> /approve
> 
> Looks good. One question about whether we should consider to allow unsetting the field.

anhdle-sso:
> /approve
> 
> Looks good. One question about whether we should consider to allow unsetting the field.

I added the logic where we blocked downgrade edition. If users do not specify edition, we default to standard in mapping but won't update the edition (because it is treated as downgrade). Effectively, the field become unmanaged.

apis/spanner/v1beta1/instance_types.go
```diff
@@ -53,7 +53,8 @@ type SpannerInstanceSpec struct {
 	AutoscalingConfig *AutoscalingConfig `json:"autoscalingConfig,omitempty"`
 
 	// Optional. The `Edition` of the current instance.
-	// Currently accepted values are EDITION_UNSPECIFIED, STANDARD, ENTERPRISE, ENTERPRISE_PLUS
+	// Currently accepted values are STANDARD, ENTERPRISE, ENTERPRISE_PLUS
+	// If edition is unspecified, the default value is STANDARD.
```
yuwenma, line 57: From the doc, it seems that there's a chance this default value could be different than standard
https://cloud.google.com/spanner/docs/editions-overview
How about "If edition is unspecified, Spanner server will use the lowest edition that matches your usage pattern to avoid workload disruption."

pkg/controller/direct/spanner/instance_controller.go
```diff
@@ -151,6 +152,14 @@ func (a *SpannerInstanceAdapter) Create(ctx context.Context, createOp *directbas
...
 	}
 	resource := SpannerInstanceSpec_ToProto(mapCtx, &desired.Spec, a.id.SpannerInstanceConfigPrefix())
+
+	// if edition is unspecified, Spanner API default edition to STANDARD.
+	// Spanner API bans downgrade edition from STANDARD to unspecifed,
+	// which cause infinite reconcile loop.
+	if desired.Spec.Edition == nil {
+		resource.Edition = spannerpb.Instance_STANDARD
+	}
+
```
yuwenma, line 162: Suggest leaving the GCP `edition` as unset, hopefully  the spanner server will use the STANDARD value, but it looks like there's a chance it could be another value.

pkg/controller/direct/spanner/instance_controller.go
```diff
@@ -221,30 +230,64 @@ func (a *SpannerInstanceAdapter) Update(ctx context.Context, updateOp *directbas
...
+		if resource.Edition > a.actual.Edition {
+			updateMask.Paths = append(updateMask.Paths, "edition")
+		} else {
+			editionDowngrade = true
+		}
 	}
 
-	if len(updateMask.Paths) == 0 {
+	if len(updateMask.Paths) == 0 && !editionDowngrade {
 		log.V(2).Info("no field needs update", "name", a.id)
```
yuwenma, line 253: Before returning `nil`, we want to update the KRM status. Becase this could be a first time acquire where the KCC SpannerInstance has empty status while the spec and GCP match.  If not updating the KRM status, users will consider it as KCC hasn't reconciled the object.

pkg/controller/direct/spanner/instance_controller.go
```diff
@@ -221,30 +230,64 @@ func (a *SpannerInstanceAdapter) Update(ctx context.Context, updateOp *directbas
...
+	// This will fail if higher tier's features are not disabled.
+	if editionDowngrade {
+		log.V(2).Info("Upgrading Edition to lower tier", "name", a.id)
+		req := &spannerpb.UpdateInstanceRequest{
+			Instance: resource,
+		}
+		req.FieldMask = &fieldmaskpb.FieldMask{
+			Paths: []string{"edition"},
+		}
+		op, err := a.gcpClient.UpdateInstance(ctx, req)
```
yuwenma, line 283: Does it mean the edition downgrade share the same `Update` endpoint but only allow a single field `edition` in the fieldMask? If so, suggest adding a comment for the future maintainers to be aware of this special behavior. (I feel like this could be changed from the GCP server side)

yuwenma:
/approve

Some nits, overall looks great!

pkg/controller/direct/spanner/instance_controller.go
anhdle-sso: Our coding pattern in all controllers is that when there is no update filemask; we just return nil. For acquiring, I think the empty status is done behind the scence.

yuwenma (commented):
/lgmt

PR #3952, opened 2025-03-11
conductor: fix go format for the generated types and mappers

jingyih:
- run `${REPO_ROOT}/dev/tasks/fix-gofmt` before committing the generated types and mappers
 - include `zz_generated.deepcopy.go` in the generate CRD commit

### DIFF
--- a/experiments/conductor/cmd/runner/crd_generator_commands.go
+++ b/experiments/conductor/cmd/runner/crd_generator_commands.go
@@ -174,7 +174,8 @@ func generateTypesAndMapper(opts *RunnerOptions, branch Branch) {
 
 	// Change to controllerbuilder directory
 	controllerBuilderDir := filepath.Join(workDir, "dev", "tools", "controllerbuilder")
-	hasChange := false
+	hasAPIChange := false
+	hasMapperChange := false
 
 	// Generate types
 	apisDir := filepath.Join(opts.branchRepoDir, "apis", branch.Group, "v1alpha1", string(filepath.Separator))
@@ -196,8 +197,7 @@ func generateTypesAndMapper(opts *RunnerOptions, branch Branch) {
 		if err != nil {
 			log.Fatal(err)
 		}
-		gitAdd(ctx, workDir, apisDir)
-		hasChange = true
+		hasAPIChange = true
 	} else {
 		log.Printf("SKIPPING generating apis, %s already exists", apisDir)
 	}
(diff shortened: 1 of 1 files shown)


experiments/conductor/cmd/runner/crd_generator_commands.go
```diff
@@ -221,32 +221,47 @@ func generateTypesAndMapper(opts *RunnerOptions, branch Branch) {
...
-			Name: "Format generated code",
-			Cmd:  "go",
-			Args: []string{
-				"run", "-mod=readonly",
-				"golang.org/x/tools/cmd/goimports@latest",
-				"-w",
-				mapperDir,
-			},
+	// Run fix-gofmt script if any changes were made
+	if hasAPIChange || hasMapperChange {
```
yuwenma, line 230: A cheaper solution could be for any step that may introduce code changes, git commit relevant directory, so that the function can be shared, and we don't need to re-evaluate or reimplement it every time the steps are merged, split, changed.

barney-s: This step is deterministic (vs llm generation steps). Having these specific changes here is ok. I will scan the code and see if we can generalize these steps and extract them into common patterns. For now lets get this in since most folks are at this stage of execution.

experiments/conductor/cmd/runner/crd_generator_commands.go
```diff
@@ -276,6 +291,7 @@ func generateCRD(opts *RunnerOptions, branch Branch) {
 
 	// Stage the changed files
 	gitAdd(ctx, workDir, "config/crds/resources/")
+	gitAdd(ctx, workDir, "apis/") // stage the zz_generated.deepcopy.go file
```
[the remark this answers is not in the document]
barney-s: There is no LLM involved in the step. Though i dont like broad path apis/* it may be ok here.

experiments/conductor/cmd/runner/crd_generator_commands.go
yuwenma: If this is about unblocking others, we should be fine because this is covered in a later step via make ready-pr. 

I have two general questions: 1. why do we want to distinguish type change and mapper change. 2. can we use `make fmt` instead (in a later step)

experiments/conductor/cmd/runner/crd_generator_commands.go
```diff
@@ -221,32 +221,47 @@ func generateTypesAndMapper(opts *RunnerOptions, branch Branch) {
...
-				"-w",
-				mapperDir,
-			},
+	// Run fix-gofmt script if any changes were made
+	if hasAPIChange || hasMapperChange {
+		// Run goimports on the generated code
+		cfg := CommandConfig{
+			Name:       "Fix imports",
+			Cmd:        "go",
+			Args:       []string{"run", "-mod=readonly", "golang.org/x/tools/cmd/goimports@latest", "-w", mapperDir},
```
acpana, line 235: should we really on `make fmt`? or the existing `verify-goimports` ?

jingyih: Sorry could you elaborate?

experiments/conductor/cmd/runner/crd_generator_commands.go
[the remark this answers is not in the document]
jingyih: Currently, Conductor does not maintain a record of the "go package" (e.g., `vertexai.cnrm.cloud.google.com/v1alpha1`) used by the type generator when calculating the apis folder path. This makes it less straightforward to determine the exact API location for a specific resource. Besides, if any prior commands generated files outside the target folder, they would have already been added by `gitAdd(ctx, workDir, apisDir)`, which is the behavior of the existing code. I don’t believe this concern should block this PR.

jingyih:
/close
Superseded by #3967.

PR #3943, opened 2025-03-10
TagsTagKey: CRD/Mapper/Fuzzer

justinsb:
- **autogen: generate-script for gcloud resource-manager tags keys**
- **autogen: Capture golden output for mockresourcemanager/testdata/tagkey/crud**
- **mockgcp: update service for resourcemanager tagkey**
- **tests: fix confusion around unusual search syntax**
- **conductor: "Generated types and mapper for TagsTagKey"**
- **conductor: "Generated CRD for TagsTagKey"**
- **TagsTagKey: fuzzer**

PR #3954, opened 2025-03-11
apilinter: improve refs check

justinsb:
Find more likely-references, and exempt a few non-refs (e.g. spec.location)


justinsb:
In #3943, I noticed that our ref-detection heuristics were just too simplistic

PR #3766, opened 2025-02-23
mockgcp: support for resource-manager tagkey

justinsb:
- **autogen: generate-script for gcloud resource-manager tags keys**
- **autogen: Capture golden output for mockresourcemanager/testdata/tagkey/crud**
- **mockgcp: update service for resourcemanager tagkey**
- **tests: fix confusion around unusual search syntax**

PR #3955, opened 2025-03-11
conductor: update metadata for tags

justinsb:


experiments/conductor/branches-all.yaml
yuwenma, line 1: We don't have any beta `Tags` resources yet. I'm wondering why we start from v1beta1 here and change the group [from `resourcemanager` to `tags`](https://cloud.google.com/resource-manager/docs/tags/tags-creating-and-managing). Do you mind sharing more information?

yuwenma: I see. I didn't realize there's such a v1beta1 resource with alpha stability level and no reference doc. Considering that its group is incorrect (at least not following the convention), it's not stable like other Beta resource, and it is not that commonly used (no reference doc ), how about we treat this as a new direct alpha resource, and use the right group `resourcemanager`? So we don't need to dicover the TF-based special handlings and keep it backward compatible.

justinsb: I'm not sure I see the upside?

yuwenma: The advantage is that we can fix the group name and save the TF triaging time. 
If we keep the group the same, then you may have to add full test coverage for the TF-based TagsTagKey before adding the scifi controller (basically adds `basic-` and `full-` test suites in https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/master/pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagkey  besides #3766)

PR #3956, opened 2025-03-11
conductor: allow crd-version to be specified

justinsb:


yuwenma (commented):
I thought about this as well, but I think we are not ready to open the gate for beta resources. Mostly because LLM generated code is not guaranteed to have those signals that we used for reviewing Beta resources so this heavily relies on the reviewers' notices and reviewing a beta resource basically reviews every line of the code. This is not feasible yet considering the current test coverage and the needs for fast PR reviews. What we missed here is 1. the tool to run presubmit check against real and mock and do the comparison, 2 the full CRD checkers for different API problems, detecting reference resource and validate parent URL 3. the full scenario tests for acquire, service-generated-id, multi-updates, create/update/delete only use case.

I see you need this for beta resource migration, could you run the `generate-types` `generate-mappers` separately? Those steps do not require LLM.

justinsb:
> I think we are not ready to open the gate for beta resources. 
I don't think we're opening the gate for beta resources here.  When we have to do beta resources, we're just saying that people should use the tooling.

I'm currently trying the tooling, I think it's pretty close to the manual processes.

ldanielmadariaga:

PR #3957, opened 2025-03-11
conductor: echo stdout/stderr even when capturing

justinsb:
This makes it easier to see what conductor is doing.


yuwenma (commented):
/lgtm
/approve

Yes!

PR #3959, opened 2025-03-11
conductor: require logging dir

justinsb:
We really should always be capturing logs (and writing status to stdout also),
and this lets us move prompt files out of the "code repo".


yuwenma:
/lgtm
/approve

Agree!

PR #3960, opened 2025-03-11
conductor: run any script

acpana:
Add the ability to run any script! on the branches.

Sometimes you just want to `git reset HEAD^ --hard` or `make fmt` or something else!

```bash
$ make build && ./bin/conductor runner --branch-repo=<REPO_PATH> --branch-conf=branches.yaml --logging-dir=./logs --command=-5 --script=script.sh
```

Went with `-5` bc this is a bit of a footgun.

### DIFF
--- a/experiments/conductor/cmd/runner/script_commands.go
+++ b/experiments/conductor/cmd/runner/script_commands.go
@@ -1,0 +1,81 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


experiments/conductor/cmd/runner/script_commands.go
```diff
@@ -0,0 +1,81 @@
...
+// isExecutable checks if the given file path is executable.
+func isExecutable(info os.FileInfo) bool {
+	// Check if it's a regular file and has at least one execute permission bit set
+	mode := info.Mode()
+	return mode.IsRegular() && (mode.Perm()&0111 != 0)
+}
+
+// executeScript runs the script located at scriptPath
+func executeScript(opts *RunnerOptions, branch Branch) error {
+	ctx := context.TODO()
```
cheftako, line 34: Especially for a foot gun like this you need to set up the logging.
	defer close()
	if branch.Command == "" {
		log.Printf("SKIPPING %s, no gcloud command", branch.Name)
		return
	}

cheftako:
Clearly need to fix the logging.
Other concerns
- Its a foot gun
- Having tested and working commands which have been through code review is safer on multiple fronts including security
- Your examples  `git reset HEAD^ --hard` or `make fmt` are great but I would rather these are common shared and supported commands. Having them as scripts discourages sharing these solutions.

PR #3963, opened 2025-03-11
conductor: dont retry runMockgcpTests

acpana:
I don't think we should retry the deterministic steps, like running the mockgcp tests since an error is terminal usually in that case.

### DIFF
--- a/experiments/conductor/cmd/runner/mock_commands.go
+++ b/experiments/conductor/cmd/runner/mock_commands.go
@@ -795,6 +795,9 @@ func addProtoToMakefile(opts *RunnerOptions, branch Branch) {
 }
 
 func runMockgcpTests(opts *RunnerOptions, branch Branch) {
+    if opts.defaultRetries > 0 {
+        log.Printf("Command does not support retries")
+    }
 	ctx := context.TODO()
 
 	close := setLoggingWriter(opts, branch)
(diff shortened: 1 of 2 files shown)


experiments/conductor/cmd/runner/mock_commands.go
```diff
@@ -818,7 +821,7 @@ func runMockgcpTests(opts *RunnerOptions, branch Branch) {
 		},
 		WorkDir:    workDir,
 		Env:        map[string]string{"WRITE_GOLDEN_OUTPUT": "1", "E2E_GCP_TARGET": "mock"},
-		MaxRetries: 2,
+        MaxRetries: -1, // no retries, even if user defined
```
barney-s, line 824: wont setting this to 1 be sufficient ?

barney-s: we can rename this to maxattempts instead of retries if its more appropriate

yuwenma: +1 to @barney-s  and the naming suggestion as well

acpana: Closing in favor of https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4062

PR #3965, opened 2025-03-11
fix: Special handling when resolving FirewallPolicyRef

gemmahou:
### Change description

Fixes b/398901413

### Tests you have done


maqiuyujoyce (commented):
/lgtm
/approve

/hold

LGTM, hold in case @yuwenma has any additional thoughts.

apis/refs/v1beta1/computerefs.go
```diff
@@ -418,6 +418,19 @@ func ResolveComputeFirewallPolicy(ctx context.Context, reader client.Reader, src
...
+	// but user specified resourceID is firewallPolicies/${firewallPolicyID}. DCL controller accepts both format.
+	// That caused resource identifier was constructed incorrectly, resulted in an error when parsing the resource name
+	// and validation the format. See b/398901413 for the detail of the customer issue.
+	// DCL has the similar format conversion: https://github.com/GoogleCloudPlatform/declarative-resource-client-library/blob/main/services/google/compute/alpha/firewall_policy_internal.go#L610
+	// todo: handle discrepancy between user specified resourceID and server generated resourceID during firewall policy direct controller migration.
+	tokens := strings.Split(resourceID, "/")
+	if len(tokens) == 2 && tokens[0] == "firewallPolicies" {
+		resourceID = tokens[1]
+	}
+
```
yuwenma, line 433: A better way to solve the problem is to not get resourceID from `spec`, but from the `status`. Please use `status.selfLink` (DCL-based) instead.

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computefirewallpolicyrule/computefirewallpolicyrule-minimal-direct/_generated_object_computefirewallpolicyrule-minimal-direct.golden.yaml
yuwenma, line 1: Please add a test suite to cover the issue you want to fix and verify it works. 
1. add back dcl2crd label, run dcl-controller with the `spec.resourceID: firewallPolicies/${policyID}`. run real-gcp. 2nd git-commit
2. add `alpha.cnrm.cloud.google.com/reconciler: "direct"` annotation, run real-gcp. 3rd git-commit
3. Remove dcl2crd label, remove `alpha.cnrm.cloud.google.com/reconciler: "direct"` annotation, run mock gcp. 4th git-commit.

gemmahou: Done, please see the updated commits. 
This use case occurs when the customer attempts to acquire an existing dependency resource by specifying its resourceID in the configuration yaml:  https://screenshot.googleplex.com/8Bh8PvBGG8bwCJa.png
I added test cases for acquisition, please advice if additional testing steps are needed. Thanks!

apis/refs/v1beta1/computerefs.go
```diff
@@ -413,11 +413,18 @@ func ResolveComputeFirewallPolicy(ctx context.Context, reader client.Reader, src
 		return nil, fmt.Errorf("error reading referenced ComputeFirewallPolicy %v: %w", key, err)
 	}
 
-	resourceID, err := GetResourceID(computeFirewallPolicy)
-	if err != nil {
-		return nil, err
+	// Read status.selfLink to parse external reference ID. This will need to be updated once we migrate this resource
+	// to direct controller, which uses status.externalRef.
```
yuwenma, line 417: There's no need to wait for the migration. You can implement the status.externalRef here via unstructured. (The current approach does not guarantee this comment would be addressed during migration, so I suggest we move forward to implement it directly).

mockgcp/mockcompute/firewallpoliciesv1.go
```diff
@@ -41,13 +41,40 @@ func (s *FirewallPoliciesV1) Get(ctx context.Context, req *pb.GetFirewallPolicyR
...
 
 	obj := &pb.FirewallPolicy{}
-	if err := s.storage.Get(ctx, fqn, obj); err != nil {
+	if err = s.storage.Get(ctx, fqn, obj); err != nil {
 		if status.Code(err) == codes.NotFound {
+			// A hacky way to test acquiring pre-crated policy 1059732409893
+			if fqn == "locations/global/firewallPolicies/1059732409893" {
+				// Create a mock object with the exactly same values of the existing policy 1059732409893
+				return s.createMockPolicy(ctx, obj, name)
+			}
```
yuwenma, line 50: Suggest remove this. It is okay to mock the service-generated ID, but it is better not in the Get logic (unless the GCP server does a GET+ CREATE in its GET call, but I don't think that's the case here)

apis/refs/v1beta1/computerefs.go
gemmahou: Here I resolve policyRef from status.selfLink, as suggested in your previous comment: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3965#discussion_r1996072013, because policy is still DCL-based and does not have externalRef field yet. That's why I left this comment for future reference.

Do you mean we should resolve policyRef by status.externalRef?

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computefirewallpolicyrule/computefirewallpolicyrule-minimal-direct/_generated_object_computefirewallpolicyrule-minimal-direct.golden.yaml
gemmahou: Done

gemmahou:
I do see many GET events got added/removed in the diff between commits.
In unified_test, we [removed repeated GET event](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/d0327cc69fb9648fc7af89b81183e56e89fd8d6b/tests/e2e/unified_test.go#L941-L958), while this is not yet implemented in scenario_test. I think it's safe to ignore the GET diffs for now.

apis/refs/v1beta1/computerefs.go
```diff
@@ -413,11 +413,15 @@ func ResolveComputeFirewallPolicy(ctx context.Context, reader client.Reader, src
...
+	selfLink, _, _ := unstructured.NestedString(computeFirewallPolicy.Object, "status", "selfLink")
+	if selfLink == "" {
+		return nil, k8s.NewReferenceNotFoundError(computeFirewallPolicy.GroupVersionKind(), key)
 	}
```
yuwenma, line 421: Discussed offline, suggest adding the externalRef logic here otherwise this code will leave a bug that is hard to discover for the future. 

```go
externalRef, _, _ := unstructured.NestedString(computeFirewallPolicy.Object, "status", "externalRef")
if externalRef != "" {
      ...               
}

selfLink, _, _ := unstructured.NestedString(computeFirewallPolicy.Object, "status", "selfLink")
If selfLink != "" {
       ... 
}

return nil, k8s.NewReferenceNotFoundError()
```

pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagkey/tagkeyorgautogen/_generated_object_tagkeyorgautogen.golden.yaml
```diff
@@ -26,6 +26,6 @@ status:
     type: Ready
   createTime: "1970-01-01T00:00:00Z"
   name: ${uniqueId}
-  namespacedName: 123450001/keyname${uniqueId}
+  namespacedName: ${organizationID}/keyname${uniqueId}
```
yuwenma, line 29: Can I ask why this test needs to be changed?

pkg/test/resourcefixture/testdata/iamexternalonlyref/organizationiampolicymember/_generated_object_organizationiampolicymember.golden.yaml
yuwenma, line 1: Can I ask why this test needs to be changed?

yuwenma (commented):
Some ditto, overall looks good

apis/refs/v1beta1/computerefs.go
gemmahou: I don't think externalRef is critical right now as this resource is still within the TF/DCL framework. I don't feel it's impacting the immediate customer issue. Could you provide more details about the potential bug you foresee? It would be necessary to add externalRef after direct controller migration, so I left a comment here for future reference.

Also this implementation is applied in many other places, for example: https://github.com/GoogleCloudPlatform/k8s-config-connector/commit/1f02b1f13677fb33dc82013b4eeb85b0dff35e1a in ResolveComputeServiceAttachmentRef. If that would cause any risk I feel we should address in a separate PR, as well as in other resources.

pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagkey/tagkeyorgautogen/_generated_object_tagkeyorgautogen.golden.yaml
gemmahou: I added a new replacement rule to replace org id to a placeholder: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3965/commits/f29067a2176f6abc3a108068b6a3e3ab52247e77#diff-2f4802ffe175095be8935b2f116b72354eccdc34f3e433132b61898776b48941R83, that affects other tests that still have the org id in log

apis/refs/v1beta1/computerefs.go
yuwenma: All compute resources are more heavily referenced than others. Its references already break more often than others and this is another case that would easily break. 
If the `apis/refs/<compute>` has been  migrated to `apis/compute`, maybe we can make this exceptional case and make the change later. But I don't see that's the case. 

If you want to make the externalRef change later, please add test to make sure all compute references supports both the TF-based format and the scifi format.

gemmahou: Done

yuwenma (commented):
/lgtm
/hold cancel

PR #3967, opened 2025-03-11
conductor: Refactor to simplify each branch function

barney-s:
conductor: Refactor to simplify each branch function

processBranches calls branchFunction for each branch and handles common:
* logging setup
* branchCheckout
* git staging and commits
* make ready-pr (and other checks in future)

### DIFF
--- a/experiments/conductor/cmd/runner/crd_generator_commands.go
+++ b/experiments/conductor/cmd/runner/crd_generator_commands.go
@@ -150,13 +150,10 @@ func generateCRDFromScripts(opts *RunnerOptions, branch Branch) {
 }
 */
 
-func generateTypesAndMapper(opts *RunnerOptions, branch Branch) {
-	ctx := context.TODO()
-
-	close := setLoggingWriter(opts, branch)
-	defer close()
-	workDir := opts.branchRepoDir
-
+func generateTypesAndMapper(ctx context.Context, opts *RunnerOptions, branch Branch) ([]string, error) {
+	var affectedPaths []string
+	generatedCode := false
+	// Check parameters
 	if branch.Kind == "" || branch.Proto == "" || branch.Group == "" {
 		if branch.Kind == "" {
 			log.Printf("SKIPPING %s, missing Kind", branch.Name)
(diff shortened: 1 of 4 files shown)


experiments/conductor/cmd/runner/utilities.go
```diff
@@ -432,3 +434,106 @@ func executeCommand(opts *RunnerOptions, cfg CommandConfig) (string, string, err
...
+		// Process the branch
+		affectedPaths, err := processor(ctx, opts, branch)
+		if err != nil {
+			log.Printf("failed to process branch %s: %v", branch.Name, err)
+			close()
+			continue
+		}
+
+		// Run make ready-pr verification
+		if err := checkMakeReadyPR(opts); err != nil {
```
jingyih, line 527: I'm still not convinced that we should run "make ready-pr" after every step. First, the code might not compile at certain command. For example, after generating the types but before running the generate-crds script to create the deepcopy file. Second, doing so could add a lot of unnecessary overhead.

experiments/conductor/cmd/runner/crd_generator_commands.go
```diff
@@ -214,119 +207,77 @@ func generateTypesAndMapper(opts *RunnerOptions, branch Branch) {
...
-		log.Fatal(err)
-	}
-
-	// Stage the changed files
-	gitAdd(ctx, workDir, "config/crds/resources/")
 
-	// Commit the changes
-	gitCommit(ctx, workDir, fmt.Sprintf("Generated CRD for %s", branch.Kind))
+	_, _, err := executeCommand(opts, cfg)
+	return affectedPaths, err
```
jingyih, line 243: Add `apiDirPathRelative` to affectedPaths (for the deepcopy file).

experiments/conductor/cmd/runner/utilities.go
jingyih: We could instead run "go fmt" after each step.

barney-s: added a runLinters for each step. Currently we run go fmt. We can add to it.

experiments/conductor/cmd/runner/crd_generator_commands.go
barney-s: done.

experiments/conductor/cmd/runner/utilities.go
```diff
@@ -432,3 +434,126 @@ func executeCommand(opts *RunnerOptions, cfg CommandConfig) (string, string, err
...
+			continue
+		}
+
+		close := setLoggingWriter(opts, branch)
+		checkoutBranch(ctx, branch, opts.branchRepoDir)
+		// Process the branch
+		affectedPaths, err := processor(ctx, opts, branch)
+		if err != nil {
+			log.Printf("failed to process branch %s: %v", branch.Name, err)
+			close()
```
jingyih, line 539: nit: probably have a perBranch function so that we can just defer close()?

jingyih (commented):
/lgtm
/approve

jingyih:
Looks like there are couple linter failures.

jingyih (commented):
/lgtm

PR #3973, opened 2025-03-12
feat: add types, mappers and fuzzer for GKEBackupBackupPlan

jingyih:


yuwenma:
Could you fix the missing ref conflict?

jingyih:
Rebased.

apis/container/v1beta1/containercluster_reference.go
```diff
@@ -0,0 +1,30 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1beta1
+
+type ContainerClusterRef struct {
```
yuwenma, line 17: Suggest implementing the ExternalNormalizer interface, so the GKEBackupPlan (and other resources) can just call `NormalizeExternal` without caring about whether the reference supports the KCC  n/n or not. You would need that function in GKEBackupPlan controller  anyways, otherwise once this ref is extended to support KCC n/n the GKEBackupPlan could be broken.

jingyih: Not sure I follow this. Is this a blocking comment on the GKEBackupBackupPlan types and mappers? If not, can we follow up?

yuwenma: Sorry for the confusion. Yes, this is a blocking change. Because reference a KCC ContainerCluster  (TF-based beta) should be supported  when defining the reference APIs.

jingyih: Uploaded a commit to normalize the ContainerClusterRef
https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3973/commits/725a66d6a9cf86e63c02bc715add72025c3fe356

PR #3976, opened 2025-03-12
controllerbuilder: allow specification of proto messages from different proto packages

justinsb:


yuwenma (commented):
/lgtm
/approve

Thanks! We may want to do the same for Mapper and Controller generators.

justinsb:
> We may want to do the same for Mapper and Controller generators.
Yup, that's coming :-)

PR #3977, opened 2025-03-12
discoveryengine: create generate script, move files to updated locations

justinsb:


yuwenma (commented):
/lgtm
/approve

The presubmit failure looks like a flake

PR #3978, opened 2025-03-12
conductor: add cmdAdjustTypes

barney-s:


experiments/conductor/cmd/runner/utilities.go
```diff
@@ -543,16 +552,20 @@ func processBranch(ctx context.Context, opts *RunnerOptions, branch Branch, proc
 	}
 
 	// Run basic linting
+	LintFailure := false
```
justinsb, line 555: Nit: lintFailure (or `var lintErrors []error`)

justinsb (approved):

justinsb:
So this would be additive except that our steps are sequential.

Can we make it step 25 or something?  (And then maybe switch the tasks to be named?)

justinsb (approved):

PR #3999, opened 2025-03-13
controllerbuilder: fix default value for proto-dir flag

justinsb:


yuwenma (commented):
/lgtm
/approve

+100! I also fetched some other googleapis repos under .build/third_party/googleapis/ This can really help reducing the irrelevant data analyzed.

PR #4000, opened 2025-03-13
chore: remove debug in generate-crds

justinsb:


yuwenma (commented):
/lgtm
/approve

+100. Yeah, we don't want to feed this to gemini (if that's what this change for)

PR #4003, opened 2025-03-13
feat: generate API, CRD, mockgcp tests and controller for BQReservati…

600lyy:
### Change description

Adding a new KCC resource `BigQueryReservationAssignment`
1) Generated the API and CRD
2) Added resource mappings between KRM and API 
3) Generated the direct controller

### Tests you have done
`KCC_USE_DIRECT_RECONCILERS=BigQueryReservationAssignment hack/record-gcp fixture/projectassignment`


jingyih:
Is this PR ready for review?

apis/bigqueryreservation/v1alpha1/assignment_types.go
```diff
@@ -0,0 +1,125 @@
...
+	// +required
+	ProjectRef *refv1beta1.ProjectRef `json:"projectRef"`
+
+	// Note: changing this field will move the assignment from one reservation to another reservation.
+	// +required
+	Location *string `json:"location"`
+
+	// Note: changing this field will move the assignment from one reservation to another reservation.
+	// +required
+	ReservationName *string `json:"reservationName"`
```
jingyih, line 38: I think instead of using "projectRef, Location and ReservationName", we just use "ReservationRef". It will make it clear that `BigQueryReservation` is the parent resource.

Here is a recent example where `VMwareEngineNetworkPolicy` is the parent of `VMwareEngineExternalAccessRule`.
https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/vmwareengine/v1alpha1/externalaccessrule_types.go

apis/bigqueryreservation/v1alpha1/assignment_types.go
```diff
@@ -0,0 +1,125 @@
...
+
+// BigQueryReservationAssignmentObservedState is the state of the BigQueryReservationAssignment resource as most recently observed in GCP.
+// +kcc:proto=google.cloud.bigquery.reservation.v1.Assignment
+type BigQueryReservationAssignmentObservedState struct {
+	// Name of the resource. E.g.:
+	//  `projects/myproject/locations/US/reservations/team1-prod/assignments/123`.
+	//  The assignment_id must only contain lower case alphanumeric characters or
+	//  dashes and the max length is 64 characters.
+	// +kcc:proto:field=google.cloud.bigquery.reservation.v1.Assignment.name
+	Name *string `json:"name,omitempty"`
```
jingyih, line 86: Let's comment out this field if it is the same as "externalRef".

apis/bigqueryreservation/v1alpha1/assignment_types.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// State of the assignment.
+	// +kcc:proto:field=google.cloud.bigquery.reservation.v1.Assignment.state
+	State *string `json:"state,omitempty"`
+}
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// TODO(user): make sure the pluralizaiton below is correct
```
jingyih, line 95: nit: remove TODO

(we recently removed this line from the code template, so it will no longer exist in the generated code)

apis/bigqueryreservation/v1alpha1/assignment_types.go
```diff
@@ -0,0 +1,125 @@
...
+	// +required
+	ReservationName *string `json:"reservationName"`
+
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="assignee field is immutable"
+	// Immutable.
+	// The resource which will use the reservation. E.g.
+	//  `projects/myproject`, `folders/123`, or `organizations/456`.
+	// +kcc:proto:field=google.cloud.bigquery.reservation.v1.Assignment.assignee
+	// +required
+	Assignee *string `json:"assignee,omitempty"`
```
jingyih, line 46: This looks like a reference field. We probably need to make this field referencing other KCC resources. Do you know if there is other resource types that can be assigned to this field besides project, folder and organization?

jingyih: We probably need something like 

```
type Assignee struct {
    ProjectRef *refs.ProjectRef
    FolderRef *refs.FolderRef
    ...
}   

type BigQueryReservationAssignmentSpec struct {
    ...
    // add a CEL rule to ensure only one reference is set.
    Assignee *Assignee
    ...
}
```

Here is a recent example: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/ce76ead4fe9dbbcfff7682c56ba7c25a5cf964af/apis/vmwareengine/v1alpha1/networkpeering_types.go#L49

apis/bigqueryreservation/v1alpha1/assignment_types.go
600lyy: Sounds like a good idea!

apis/bigqueryreservation/v1alpha1/assignment_types.go
600lyy: No, only project, folder and organization can be assigned to a reservation.

Do we still need a reference field in this case?

jingyih: Yes I think this field should be a reference field.

pkg/controller/direct/bigqueryreservation/assignment_controller.go
```diff
@@ -0,0 +1,320 @@
...
+	if obj.Spec.Assignee.OrganizationRef != nil && obj.Spec.Assignee.OrganizationRef.External == "" {
+		obj.Spec.Assignee.OrganizationRef.External, err = refsv1beta1.ResolveOrganizationID(ctx, reader, u)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	var reservationName string
+	// Get the reservation name to move the assignment to
+	if obj.Spec.ReservationRef.External != "" {
```
jingyih, line 106: No need to check for `obj.Spec.ReservationRef.External != ""`? This is logic should already be inside "NormalizedExternal".

pkg/controller/direct/bigqueryreservation/assignment_controller.go
```diff
@@ -0,0 +1,320 @@
...
+		reservationName = obj.Spec.ReservationRef.External
+	} else {
+		reservationName, err = obj.Spec.ReservationRef.NormalizedExternal(ctx, reader, obj.GetNamespace())
+		if err != nil {
+			return nil, err
+		}
+	}
+	p, name, err := krm.ParseReservationExternal(reservationName)
```
jingyih, line 114: IMO this is unnecessary, technically we just need: 
```
type BQReservation struct {
     BQReservation string // format is `projects/{{projectID}}/locations/{{location}}/reservations/{{reservationName}}`
}
````

File: mockgcp/mockbigqueryreservation/reservation.go [hunk 7]
```diff
@@ -219,3 +220,145 @@ func setFieldNonfailoverFailover(obj *pb.Reservation, location string) *pb.Reser
...
+	if err := s.storage.Create(ctx, fqn, obj); err != nil {
+		return nil, err
+	}
+
+	return obj, nil
+}
+
+// Updates an existing assignment.
+func (s *ReservationV1) UpdateAssignment(ctx context.Context, req *pb.UpdateAssignmentRequest) (*pb.Assignment, error) {
+	// TODO
```
jingyih, line 286: Implement the Update?

apis/bigqueryreservation/v1alpha1/assignment_identity.go
```diff
@@ -0,0 +1,138 @@
...
+
+	parent := &BQReservation{
+		ProjectID:       reservationParent.ProjectID,
+		Location:        reservationParent.Location,
+		ReservationName: reservationID,
+	}
+
+	// Get desired assignment resourceID
+	// Only works for resource acquisition
+	resourceID := common.ValueOf(obj.Spec.ResourceID)
```
jingyih, line 90: if `obj.Spec.ResourceID` is set (acquisition), use it as `resourceID`. Otherwise we should check if `status.externalRef` is populated. If it is populated (meaning the resource was already created, and we are in the subsequent reconciliations), use the ID parsed from `status.externalRef`.

pkg/controller/direct/bigqueryreservation/assignment_controller.go
```diff
@@ -0,0 +1,320 @@
...
+	log.V(2).Info("creating Assignment", "name", a.id.String())
+	mapCtx := &direct.MapContext{}
+
+	desired := a.desired.DeepCopy()
+	assignment := BigqueryReservationAssignmentSpec_ToProto(mapCtx, &desired.Spec)
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+
+	req := &pb.CreateAssignmentRequest{
```
jingyih, line 184: The proto does not suggest that the resource ID is always service generated?
```
  // The optional assignment ID. Assignment name will be generated automatically
  // if this field is empty.
  // This field must only contain lower case alphanumeric characters or dashes.
  // Max length is 64 characters.
  string assignment_id = 4;
  ```

File: pkg/controller/direct/bigqueryreservation/assignment_controller.go [hunk 10]
```diff
@@ -0,0 +1,320 @@
...
+		return mapCtx.Err()
+	}
+
+	// update the externalRef in the KRM resoruce
+	status.ExternalRef = direct.LazyPtr(created.GetName())
+	return createOp.UpdateStatus(ctx, status, nil)
+}
+
+// Update updates the resource in GCP based on `spec` and update the Config Connector object `status` based on the GCP response.
+func (a *AssignmentAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOperation) error {
```
jingyih, line 207: Not sure what this function is trying to do. I would expect that the controller calls "UpdateAssignment" RPC.

pkg/controller/direct/bigqueryreservation/assignment_controller.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+
+	// Get bigqueryreservation GCP client
+	gcpClient, err := m.client(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// NormalizeAssignee
+	if obj.Spec.Assignee.ProjectRef != nil && obj.Spec.Assignee.ProjectRef.External == "" {
```
jingyih, line 85: No need to check for `"obj.Spec.Assignee.ProjectRef.External == ""`?

pkg/controller/direct/bigqueryreservation/assignment_controller.go
600lyy: The purpose is to run `NormalizedExternal ` only once as it will set both `r.External` and `r.Name` if `r.External` is empty. 

Running `NormalizedExternal ` more than once will return this error "cannot specify both name and external on BigQueryReservationReservation"

jingyih: Exactly, `NormalizedExternal` should only be called once during a reconciliation. If you see this error, find and remove the second call.

pkg/controller/direct/bigqueryreservation/assignment_controller.go
600lyy: It is doing moving of an assignment from one reservation to another, when a new reservationRef is specified.

`a.gcpClient.MoveAssignment`

jingyih: Can an assignment fields be updated without moving to a different reservation? I think we also need to support it.

pkg/controller/direct/bigqueryreservation/assignment_controller.go
600lyy: ok. What the proto is actually saying is:

1) when create a new assignment, it is always service generated.
2) when moving an assignment to a different reservation, you can choose either to keep the name or system will generate a new one for you.

In the API doc, it also says: 
`Output only. Name of the resource. E.g.: projects/myproject/locations/US/reservations/team1-prod/assignments/123.`
https://cloud.google.com/bigquery/docs/reference/reservations/rest/v1/projects.locations.reservations.assignments

jingyih: The `name` field is an output field commonly found in GCP resources. It essentially serves the same purpose as `status.externalRef` in KCC. When creating the resource, the API takes an `assignmentId` [1], which becomes the ID of the newly created resource. Setting this field consistently when calling the API simplifies the controller logic.

[1] https://cloud.google.com/bigquery/docs/reference/reservations/rest/v1/projects.locations.reservations.assignments/create#query-parameters

pkg/controller/direct/bigqueryreservation/assignment_controller.go
600lyy: Currently, no fields can be updated once the assignment is created.

pkg/controller/direct/bigqueryreservation/assignment_controller.go
600lyy: You're right! Thanks for the additional info.
Fixed

mockgcp/mockbigqueryreservation/reservation.go
600lyy: As said, no fields can be updated once the assignment is created.

pkg/controller/direct/bigqueryreservation/assignment_controller.go
600lyy: Yes, that's right!

pkg/controller/direct/bigqueryreservation/assignment_controller.go
```diff
@@ -0,0 +1,307 @@
...
+		return mapCtx.Err()
+	}
+
+	// update the externalRef in the KRM resoruce
+	status.ExternalRef = direct.LazyPtr(created.GetName())
+	return createOp.UpdateStatus(ctx, status, nil)
+}
+
+// Update updates the resource in GCP based on `spec` and update the Config Connector object `status` based on the GCP response.
+func (a *AssignmentAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOperation) error {
```
jingyih, line 193: I suggest we have separate code path for "update an assignment" and "move an assignment". You are correct that the googleapis git version we pulled does not have an mutable field, but GCP just added one. Given that the `UpdateAssignment` exist in the API, we should be able to call it. So the suggestion is that, if parent does not change, call `UpdateAssignment`; otherwise call `MoveAssignment`. It is clearer this way.

https://github.com/googleapis/googleapis/blob/d2dcfcf40ff8176b330697f021f3e7db9ed3dcf7/google/cloud/bigquery/reservation/v1/reservation.proto#L1013

600lyy: Are you suggesting like this?

`
func (a *AssignmentAdapter) moveAssignment(){....}
func  (a *AssignmentAdapter) updateAssignment() {....}

func (a *AssignmentAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOperation) error {

//  parent changed, move assignment to another reservation.
moveAssignment()...

// Otherwise, handle fields update
updateAssignment()...
}
`

And do we need to update the proto version at this time?

600lyy: SGTM!

jingyih (commented):
/lgtm
/approve

PR #4010, opened 2025-03-13
feat: add types, mappers and fuzzer for DatastreamPrivateConnection

jingyih:


apis/datastream/v1alpha1/privateconnection_types.go
```diff
@@ -0,0 +1,166 @@
...
+// +kcc:proto=google.cloud.datastream.v1.VpcPeeringConfig
+type VpcPeeringConfig struct {
+	// Required. Fully qualified name of the VPC that Datastream will peer to.
+	//  Format: `projects/{project}/global/{networks}/{name}`
+	// +kcc:proto:field=google.cloud.datastream.v1.VpcPeeringConfig.vpc
+	NetworkRef *refsv1beta1.ComputeNetworkRef `json:"networkRef,omitempty"`
+
+	// Required. A free subnet for peering. (CIDR of /29)
+	// +kcc:proto:field=google.cloud.datastream.v1.VpcPeeringConfig.subnet
+	Subnet *string `json:"subnet,omitempty"`
```
yuwenma, line 165: Could you check if this should be a ref to https://cloud.google.com/config-connector/docs/reference/resource-docs/compute/computesubnetwork

yuwenma (commented):
/approve

One question, otherwise looks great!

apis/datastream/v1alpha1/privateconnection_types.go
jingyih: As far as I can tell, `Subnet` is not a reference field.

```
vpcPeeringConfig:
    networkRef:
        xxx
    subnet: "10.0.0.0/29"
```

jingyih: According to [1]:

> Identify an available IP range (with a CIDR block of /29) on the VPC network. This can't be an IP range that already exists as a subnet, a private services access pre-allocated IP range, or any route (other than the default 0.0.0.0 route) that includes the IP range. Datastream uses this IP range to create a subnet so that it can communicate with the source database. The following table describes valid IP ranges.

[1] https://cloud.google.com/datastream/docs/create-a-private-connectivity-configuration#before-you-begin

jingyih:
Rebase to resolve a conflict in go.mod

PR #4011, opened 2025-03-13
Add CRD and Mappers for resource apphub discoveredworkload

xiaoweim:


yuwenma (commented):
/approve
/lgtm

Does this resource have no fields?

xiaoweim:
> Does this resource have no fields?

My understanding from the REST API [documents](https://cloud.google.com/app-hub/docs/reference/rest/v1/projects.locations.discoveredWorkloads) is that it only contains a name identifier and the rest of the related information are all output only fields.

PR #4016, opened 2025-03-13
feat: add types, mappers and fuzzer for BackupDRBackupVault

jingyih:


yuwenma (commented):
/lgtm
/approve

(I have some question about the Effective time, but it might be a question for the econtroller.  )

PR #4019, opened 2025-03-14
chore: remove duplicated resource

jingyih:
It looks like we have two alpha CRDs for the same underlying GCP resource.

 - https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_cloudtasksqueues.cloudtasks.cnrm.cloud.google.com.yaml

 - https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_tasksqueues.tasks.cnrm.cloud.google.com.yaml


jingyih:
@maqiuyujoyce I manually removed the auto-generated one. Not sure if this is the right way to do it.

/cc @xiaoweim

maqiuyujoyce:
Yes this should work.

/lgtm
/approve

PR #4026, opened 2025-03-14
Add bigtable authorizedview types, mappers, CRD and fuzzer 

gemmahou:
### Change description

Fixes #

### Tests you have done


yuwenma:
Please fix the presubmit failure

gemmahou:
/assign @yuwenma 
This is ready for review.

apis/bigtable/v1alpha1/types.generated.go
```diff
@@ -0,0 +1,66 @@
...
+	// +kcc:proto:field=google.bigtable.admin.v2.AuthorizedView.FamilySubsets.qualifier_prefixes
+	QualifierPrefixes [][]byte `json:"qualifierPrefixes,omitempty"`
+}
+
+// +kcc:proto=google.bigtable.admin.v2.AuthorizedView.SubsetView
+type AuthorizedView_SubsetView struct {
+	// Row prefixes to be included in the AuthorizedView.
+	//  To provide access to all rows, include the empty string as a prefix ("").
+	// +kcc:proto:field=google.bigtable.admin.v2.AuthorizedView.SubsetView.row_prefixes
+	RowPrefixes [][]byte `json:"rowPrefixes,omitempty"`
```
yuwenma, line 62: Could you confirm if this should be string or []byte? It seems to be very user unfriendly to use []byte

pkg/test/fuzz/generate.go
```diff
@@ -312,6 +319,15 @@ func Visit(msgPath string, msg protoreflect.Message, setter func(v protoreflect.
 					visitor.VisitPrimitive(path+"[]", el, setter)
 				}
 
+			case protoreflect.BytesKind:
```
yuwenma, line 322: Defer to the question about if we should change []byte to string, we may not want this type extension

yuwenma (commented):
/approve

apis/bigtable/v1alpha1/types.generated.go
gemmahou: This is generated and the proto uses [][]byte.
Some thoughts here: 1) we do support []byte type in other resource. I guess it's fine to support a list of []byte as well. 2) the API doc, the field should be a "base64-encoded string of bytes". If change to string I'm afraid user would mistakenly configure this field as regular string, but we can always leverage the underlying API to validate values&process the request.

WDYT?

yuwenma: It is very hard to configure [][]byte. This seems to be a bad API design if the purpose is for base64-encryption

yuwenma:
/lgtm

please fix the merge conflict

PR #4034, opened 2025-03-15
chore: remove TODO for pluralization checks in resource types template

jingyih:


config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_iapsettings.iap.cnrm.cloud.google.com.yaml
```diff
@@ -18,7 +18,6 @@ spec:
     plural: iapsettings
     shortNames:
     - gcpiapsettings
-    - gcpiapsettingss
```
yuwenma, line 21: I thinkwe may wnat to use `gcpiapsetting` and `gcpiapsettings` as short name. And change the singular to `iapsetting` (but this would be a breaking change)

yuwenma (commented):
/lgtm
/approve

Good catch and agree to remove the TODO

yuwenma:
As a follow up, can we add a CRD checker for the shortname and its plural form? Some ideas for plural check https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/tests/apichecks/crds_test.go#L218

PR #4041, opened 2025-03-15
Master broken: One of the earlier commits broke script a test

barney-s:
One of the earlier commits broke script execution by removing execution by removing pre/post for mocks.

* marking the pre/post in this case as exec. it is safe to do so since there are no gcloud commands.
* commit that broke this: 'https://github.com/GoogleCloudPlatform/k8s-config-connector/commit/0c90028af4c52ab35a602714ca827f6cdfcf7ce3#diff-be525b45d767496986e663eb33d460ae34834d69f780e874dd000ea2fcd9f0f8'
* we need to discuss if skipping is required for mocks and why so ?


jingyih (commented):
/lgtm
/approve

Let's fix HEAD to unblock other PRs.

PR #4043, opened 2025-03-16
Add CRD and Mapper for resource storage folder

xiaoweim:


pkg/controller/direct/tasks/queue_fuzzer.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// +tool:fuzz-gen
+// +tool:fuzz-no
```
yuwenma, line 15: Mind explaining what this change is for?

xiaoweim: It should be fuzz-gen! I think I might copied over from other branch as a typo and did not change it.

PR #4056, opened 2025-03-17
Conductor: Improve Type generation

barney-s:
#### Changes for command 21
* Previous versions produced incomplete commits or no changes for the steps in cmd 21 > 50% of the time
* On debugging the logs over several iterations (tuning prompt), discovered that gemini was producing whole file changes which were good but it did not translate to Edit tool calls
* So added an option to CreateFileTool to overwrite existing files. And used that in the prompt.
* With these changes the instances of skipped steps (no change) reduced dramatically and now has only ~ 10% failure rate.
* Batch run: https://github.com/GoogleCloudPlatform/k8s-config-connector/pulls?q=is%3Apr+is%3Aopen+author%3Abarney-s+WIP%3A+Types+and+Mapper
  * mapper run is failing ignore the last commit. Rest are from steps 20, 21 (bulk), 22

#### Add Command 9
* git push origin <branch> --force
* (future planning to add make ready-pr and commit changes)

### DIFF
--- a/dev/tools/controllerbuilder/pkg/codebot/tool_create_file.go
+++ b/dev/tools/controllerbuilder/pkg/codebot/tool_create_file.go
@@ -30,8 +30,9 @@ func init() {
 }
 
 type CreateFile struct {
-	Contents string `json:"contents"`
-	Filename string `json:"filename"`
+	Contents  string `json:"contents"`
+	Filename  string `json:"filename"`
+	Overwrite bool   `json:"overwrite"`
 }
 
 type CreateFileResults struct {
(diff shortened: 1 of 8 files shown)


dev/tools/controllerbuilder/pkg/codebot/tool_create_file.go
yuwenma, line 1: There's a edit_file function. Shall we instead using that? Also, I think we need a delete_file function.

barney-s: we are using that too. Sometimes gemini likes to return an entire file and it is easy to just use createfile. The prompt suggests using both edit and create.

yuwenma (commented):
/lgtm
/approve

Thanks. Is the prompt message the one you mentioned have been used in Justin's example? My question is that we need a way to verify instruction changes before merging.

PR #4059, opened 2025-03-17
WIP: ci: generate apis and mappers where we can

justinsb:
- **ci: generate apis and mappers where we can**
- **Remove broken? generate script for documentai**
- **chore: regenerate output**
- **WIP: extract more tool markers for generated code**


dev/ci/presubmits/validate-generated
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
jingyih, line 1: We should just remove `dev/ci/presubmits/validate-generated-types` since the new check already covers type validation.

dev/tasks/generate-all
```diff
@@ -0,0 +1,26 @@
...
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+cd ${REPO_ROOT}
+
+dev/tasks/generate-apis
```
jingyih, line 23: Not sure what is the best way to gather all the direct resources for validation. Previously, I tried to parse everything in the `dev/tools/controllerbuilder/generate.sh` script but found that many resources were missing their corresponding type generation commands. Additionally, the conductor doesn't generate a `generate.sh` script for new direct resources. Not sure we want to reverse-engineer all the Go types in the `/apis` folder...

jingyih: Ah, I think I found the answer in #4060.

ldanielmadariaga:

PR #4060, opened 2025-03-17
chore: write markers from generate-mapper to support regeneration

justinsb:
- **chore: refactor tool marker extractions to support generated code**
- **chore: emit generated annotations from generate-mapper**

### DIFF
--- a/dev/tools/controllerbuilder/pkg/commands/generatemapper/generatemappercommand.go
+++ b/dev/tools/controllerbuilder/pkg/commands/generatemapper/generatemappercommand.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/GoogleCloudPlatform/k8s-config-connector/dev/tools/controllerbuilder/pkg/annotations"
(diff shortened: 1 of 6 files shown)


yuwenma:
/approve
/assign @jingyih
 
Defer lgtm to @jingyih in case you can take a look. 

I like the idea! Basically the "// +tool:" is different kinds of  "functions" we don't know, we give LLM tons of input (metadata) and output (gocode), and want it to predict the new input.

dev/tools/controllerbuilder/pkg/commands/generatemapper/generatemappercommand.go
```diff
@@ -120,7 +121,17 @@ func RunGenerateMapper(ctx context.Context, o *GenerateMapperOptions) error {
 
 		return goPackage, true
 	}
-	mapperGenerator := codegen.NewMapperGenerator(pathForMessage, o.OutputMapperDirectory)
+
+	generatedFileAnnotation := &annotations.FileAnnotation{
```
jingyih, line 125: Can we also add file annotation to the type generator?

jingyih (commented):
/lgtm

PR #4065, opened 2025-03-17
mockgcp: NetAppBackupVault

acpana:
Mostly `conductor` driven; one small caveat regarding `gcloud` using the `v1beta1` API.


mockgcp/mocknetapp/testdata/backupvault/crud/script.yaml
```diff
@@ -1,4 +0,0 @@
-
-- exec: gcloud beta netapp backup-vaults create test-${uniqueId} --location=us-central1
-- exec: gcloud beta netapp backup-vaults describe test-${uniqueId} --location=us-central1
```
acpana: Not ideal, but I can follow up on this PR with curl commands or we can just rely on the controller-drive approach, as it will use v1.

I also tried re-writing the API paths but the v1beta1 and v1 differ in LRO metadata (i.e. not compatible).

yuwenma: gotcha. Let;'s move forward to the controller part (3rd PR).

PR #4069, opened 2025-03-18
Add bigtable cluster types, mappers, CRD and fuzzer

gemmahou:
### Change description

Fixes #

### Tests you have done


apis/bigtable/v1alpha1/cluster_types.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// The BigtableCluster name. If not given, the metadata.name will be used.
+	ResourceID *string `json:"resourceID,omitempty"`
+
+	// Immutable. The location where this cluster's nodes and storage reside. For
+	//  best performance, clients should be located as close as possible to this
+	//  cluster. Currently only zones are supported, so values should be of the
+	//  form `projects/{project}/locations/{zone}`.
+	// +kcc:proto:field=google.bigtable.admin.v2.Cluster.location
+	Location *string `json:"location,omitempty"`
```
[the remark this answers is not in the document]
gemmahou: checking the API doc, cluster's parent is `projects/*/instances/*`: https://cloud.google.com/bigtable/docs/reference/admin/rest/v2/projects.instances.clusters/create#path-parameters

apis/bigtable/v1alpha1/types.generated.go
```diff
@@ -0,0 +1,83 @@
...
+	CPUUtilizationPercent *int32 `json:"cpuUtilizationPercent,omitempty"`
+
+	// The storage utilization that the Autoscaler should be trying to achieve.
+	//  This number is limited between 2560 (2.5TiB) and 5120 (5TiB) for a SSD
+	//  cluster and between 8192 (8TiB) and 16384 (16TiB) for an HDD cluster,
+	//  otherwise it will return INVALID_ARGUMENT error. If this value is set to 0,
+	//  it will be treated as if it were set to the default value: 2560 for SSD,
+	//  8192 for HDD.
+	// +kcc:proto:field=google.bigtable.admin.v2.AutoscalingTargets.storage_utilization_gib_per_node
+	StorageUtilizationGibPerNode *int32 `json:"storageUtilizationGibPerNode,omitempty"`
```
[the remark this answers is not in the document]
gemmahou: Done

apis/bigtable/v1alpha1/cluster_types.go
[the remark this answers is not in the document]
yuwenma: kk, i'm wondering if the {project} of this location is the same as the parent project

yuwenma:
please fix the merge conflict

apis/bigtable/v1beta1/instance_reference.go
```diff
@@ -0,0 +1,89 @@
...
+		metadataName, _, err := unstructured.NestedString(u.Object, "metadata", "name")
+		if err != nil {
+			return "", fmt.Errorf("reading metadata.name: %w", err)
+		}
+		resourceID = metadataName
+	}
+	if resourceID == "" {
+		return "", k8s.NewReferenceNotReadyError(u.GroupVersionKind(), key)
+	}
+	r.External = resourceID
```
jingyih, line 87: Does this look right? I am expecting the external field to have something like "projects/project/instances/instance'?

gemmahou: For direct-based resources, we always resolve to externalRef. However, for Terraform-based resources, since we do not have an externalRef field, there isn't a consistent pattern for the resolved value.. Resource identifier needs to be get from existing fields like "spec.resourceID" or "status.selfLink". This resource does not currently have status fields that contains an identifier, so "spec.resourceID" is the only option. (Ideally we want to use a status field, because spec field can be modified by users)

When this function is getting called, we need to convert the resolved external to the format that API accepts, for example: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/fdb7446bc820c451edc59c1b0e86a30139ce969e/apis/bigtable/v1alpha1/cluster_identity.go#L52. It's possible that the API wants a full link, or only the resource name(i.e. [here](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/fdb7446bc820c451edc59c1b0e86a30139ce969e/pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go#L120-L126) we convert the resolved full link to only name). So it's really depends on API requirements. 

Did you see any issue with this resolved value?

jingyih: My concern is that the behavior of `NormalizedExternal` of this resource is not consistent with the other resources that we have. When I call `xxxRef.NormalizedExternal()`, I would expect the external field to be filled with a fully qualified name.
@yuwenma WDYT?

yuwenma: Agree with @jingyih, I also think I suggested the same thing in the firewallpolicy fix https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3965#discussion_r2003701288. Please add the logic about checking the status.externalRef first, and then consider the spec.resourceID or metadata.name,

gemmahou: I can update the resolved value for this resource to match the externalRef format, as it's also the format accepted by the API.

IIUC the purpose of introducing status.externalRef is not only its format, but to avoid getting identifier from spec fields, especially since not all resources populate identifier in status, and spec can be modified by user(as I mentioned in previously comment).

For TF-based resources without externalRef, I think we have some flexibility. For example, many Compute refs use their "selfLink" (including the https://www.googleapis.com/v1 prefix) as the external value([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/compute/forwardingrule/refs.go#L214)). As long as we can reliably extract the identifier, ensure backward compatibility, and meet the API requirements, the value should be acceptable.

jingyih: Having different format when resolving the external field makes the code harder to read and maintain. We probably need to extend the current `NormalizedExternal` function signature or add a new function such as `NormalizedExternalWithFormat`. Potentially different resources could reference the same resource but require different format.

gemmahou: I have addressed this in #4331, will refactor other normalize code when I touch those resources.

PR #4075, opened 2025-03-18
controllerbuilder: add file annotation to generated types

jingyih:
Follow up of #4060, adding file annotation to generated types as well.

### DIFF
--- a/dev/tools/controllerbuilder/config/v1beta1/managedkafka.yaml
+++ b/dev/tools/controllerbuilder/config/v1beta1/managedkafka.yaml
@@ -18,3 +18,5 @@ generateMapper: true
 resources:
   - kind: ManagedKafkaCluster
     protoName: Cluster
+  - kind: ManagedKafkaTopic
+    protoName: Topic
--- a/dev/tools/controllerbuilder/pkg/annotations/fileannotation.go
+++ b/dev/tools/controllerbuilder/pkg/annotations/fileannotation.go
@@ -19,6 +19,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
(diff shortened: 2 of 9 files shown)


dev/tools/controllerbuilder/pkg/annotations/fileannotation.go
```diff
@@ -34,7 +35,17 @@ func (a *FileAnnotation) FormatGo() string {
 	var sb strings.Builder
 
 	fmt.Fprintf(&sb, "// %s\n", a.Key)
-	for k, values := range a.Attributes {
+
+	// Get all keys and sort them for consistent ordering
+	keys := make([]string, 0, len(a.Attributes))
```
jingyih, line 40: This isn't required, but having a consistent ordering might make it easier for us to detect file diff in the generated files in the future.

dev/tools/controllerbuilder/config/v1beta1/managedkafka.yaml
```diff
@@ -18,3 +18,5 @@ generateMapper: true
 resources:
   - kind: ManagedKafkaCluster
     protoName: Cluster
+  - kind: ManagedKafkaTopic
```
jingyih, line 21: We should deprecate the config files since we're taking a different approach to validating generated files in presubmit.

yuwenma: I like that we show different approaches for the same problem (some resources could fit one or the either). Is this a blocker to introduce frictions if not deprecating?

yuwenma (commented):
/lgtm
/approve

PR #4078, opened 2025-03-18
feat: add resource apphub application controller

xiaoweim:


pkg/test/resourcefixture/testdata/basic/apphub/v1alpha1/apphubapplication/apphubapplication-minimal/update.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: apphub.cnrm.cloud.google.com/v1alpha1
+kind: AppHubApplication
+metadata:
+  name: apphubapplication-minimal-${uniqueId}
+spec:
+  parent:
```
yuwenma, line 6: I like this but the existing convention is that we don't show `spec.parent`. It would be `spec.projectRef`

PR #4079, opened 2025-03-18
[conductor] Resource apigateway api type/mapper/fuzzer

anhdle-sso:
### Change description

Fixes #

### Tests you have done


apis/apigateway/v1alpha1/api_types.go
```diff
@@ -0,0 +1,121 @@
...
+	// Optional. Display name.
+	// +kcc:proto:field=google.cloud.apigateway.v1.Api.display_name
+	DisplayName *string `json:"displayName,omitempty"`
+
+	// Optional. Immutable. The name of a Google Managed Service (
+	//  https://cloud.google.com/service-infrastructure/docs/glossary#managed). If
+	//  not specified, a new Service will automatically be created in the same
+	//  project as this API.
+	// +kcc:proto:field=google.cloud.apigateway.v1.Api.managed_service
+	ManagedService *string `json:"managedService,omitempty"`
```
yuwenma, line 46: Could you double check if this should be a reference?

yuwenma (commented):
/approve

apis/apigateway/v1alpha1/api_types.go
anhdle-sso: the field can be https://cloud.google.com/service-infrastructure/docs/service-management/reference/rest/v1/services#ManagedService; but it is not clear. I think we can keep it as string for alpha

yuwenma: Can you add a comment so as we don't miss this investigation effort when promoting to beta?

anhdle-sso: done

anhdle-sso:
@yuwenma I addressed the comments. Pls take a look.

PR #4084, opened 2025-03-19
chore: refactor BackupDRManagementServer controller client

jingyih:
Refactor client code and add controller tool markers to improve prompt results when generating new controllers in the same API group.


yuwenma (commented):
/lgtm
/approve

Cool!

PR #4067, opened 2025-03-18
feat: add resource tasks queue controller

xiaoweim:
- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #4087, opened 2025-03-19
Revert "chore: remove duplicated resource"

jingyih:
This PR reverts #4019
We probably want to keep the old CRD and permission, just to be safe.

Alternatively, we can simply remove this CRD. If users notice permission issues with the ConfigConnector operator, we can ask them to manually delete the CRD from their clusters.


jingyih:
Just realized this is no longer needed due to the API group renaming in #4067.

/close

PR #4088, opened 2025-03-19
conductor improvements

barney-s:
* Move delete branch from 3 to -5
* Add confirmation for delete branch
* Add make ready-pr, push to origin to cmd 9
* Support providing --branch-suffix for pushing to remote without conflicting with other work. eg: --branch-suffix='-mocks'
* Run go imports on generated files


jingyih:
We should discuss the recommended workflow when using the conductor. Do we really need "--branch-suffix"? In my opinion, having different names for local and remote branches can actually be more confusing.

jingyih (commented):
/lgtm
/approve
I am okay that we experiment with the branch suffix flag, though I personally don’t find it necessary in my workflow.

PR #4093, opened 2025-03-19
tests: normalize folderID

justinsb:
Actual folder values were appearing in the output,
which means we cannot have parity with realgcp.


yuwenma:
/lgtm
/approve

Oh, I thought the Mock already have the folderID normalized.

justinsb:
> Oh, I thought the Mock already have the folderID normalized.

Not in every case ... added the missing golden output showing the changes!

PR #4098, opened 2025-03-19
edgecontainer machinetype: move hosted_node to status

justinsb:


yuwenma:
/approve
/lgtm

Is this because that field is OUTPUT ONLY but not have the right proto tag?

PR #4099, opened 2025-03-19
chore: create regenerate command

justinsb:
This shows how we "complete the loop", by extracting the generation markers,
and using them to regenerate.

This currently only prints the (approximate) commands, but shows the direction.


justinsb:
cc @jingyih @yuwenma

yuwenma (commented):
/approve

PR #4102, opened 2025-03-19
Types and Mapper for resource eventarc channel

barney-s:


tests/apichecks/testdata/exceptions/missingrefs.txt
```diff
@@ -60,6 +60,7 @@
 [refs] crd=dnsmanagedzones.dns.cnrm.cloud.google.com version=v1beta1: field ".spec.serviceDirectoryConfig.namespace.namespaceUrl" should be a reference
 [refs] crd=dnsresponsepolicies.dns.cnrm.cloud.google.com version=v1alpha1: field ".spec.networks[].networkUrl" should be a reference
 [refs] crd=essentialcontactscontacts.essentialcontacts.cnrm.cloud.google.com version=v1alpha1: field ".spec.parent" should be a reference
+[refs] crd=eventarcchannels.eventarc.cnrm.cloud.google.com version=v1alpha1: field ".spec.provider" should be a reference
```
yuwenma, line 63: I think this is valid. please fix it.

yuwenma (commented):
/approve

tests/apichecks/testdata/exceptions/missingrefs.txt
barney-s: me an jingyi looked at it. We dont have a provider resource in KCC or even TF. And we looked at the APIs: https://cloud.google.com/integration-connectors/docs/reference/rest/v1/projects.locations.providers.connectors.versions

Only get/list exist.
Do we want to create a resource for it in KRM ? 

May be we can do it as part of beta move ?

yuwenma: This is not a new resource but a reference, so we do want to add it when using conductor to develop alpha resource.  The work to add a Reference is relatively cheap: a go struct with `external, name, namespace` fields and a `normalizeExternal` function, but it gives us a lot of benefit:  references is the connection to reflect relationship among different resources. The more we add in alpha stage, the more reliable info we can use for LLM to write reference itself (right now heavily rely on human to verify the code), and make the beta promotion faster and more stable.

yuwenma: A quick way to add the reference code is to use the type-generator tool to generate the resource API (including *_reference.go, *_identity.go and *_types.go), only keep the *_reference.go and comment out the others.

barney-s: addressed. pTAL

PR #4109, opened 2025-03-20
chore: bump golangci for newer go versions

justinsb:
Otherwise fails with golang 1.24.1


yuwenma:
/lgtm
/approve

Thanks

PR #4117, opened 2025-03-20
Add CRD and Mapper for resource batch task

xiaoweim:


apis/batch/v1alpha1/task_types.go
```diff
@@ -0,0 +1,101 @@
...
+// BatchTaskSpec defines the desired state of BatchTask
+// +kcc:proto=google.cloud.batch.v1.Task
+type BatchTaskSpec struct {
+	Parent `json:",inline"`
+	// The BatchTask name. If not given, the metadata.name will be used.
+	ResourceID *string `json:"resourceID,omitempty"`
+
+	// Task Status.
+	// +kcc:proto:field=google.cloud.batch.v1.Task.status
+	Status *TaskStatus `json:"status,omitempty"`
```
yuwenma, line 33: This does not look like a spec field (with state and statusEven). Could you double check?

yuwenma (commented):
/approve

One API question to confirm, otherwise looks good

apis/batch/v1alpha1/task_types.go
xiaoweim: You are right! They should be a part of the observedstate.

PR #4118, opened 2025-03-20
Support types, mapper and fuzzer for ColabRuntime

maqiuyujoyce:
### Tests you have done


dev/tools/controllerbuilder/pkg/codegen/common.go
```diff
@@ -56,7 +56,7 @@ var Acronyms = []string{
 	"MiB",
 	"NAT",
 	"OAuth2", "OIDC", "OS",
-	"PD", "PSC",
+	"PD", "PSC", "PZI", "PZS",
```
yuwenma, line 59: ditto, remove PZI and PZS since we drop those two fields and they are not well known abbreviations.

maqiuyujoyce: Removed.

Though I think whether a word is an abbreviation is objective (not subjective) regardless whether people know it. Using `Pzi`/`Pzs` will probably be more confusing then marking them as `PZI`/`PZS` directly.

yuwenma (commented):
/lgtm
/approve

PR #4119, opened 2025-03-20
mockgcp: more fidelity for bigtable appProfile

justinsb:
- **mockgcp: including routing update in appProfile test**
- **mockgcp: golden output for bigtable appProfile test**
- **mockgcp: more fidelity for bigtable appProfile**

### DIFF
--- a/mockgcp/mockbigtable/service.go
+++ b/mockgcp/mockbigtable/service.go
@@ -69,6 +69,13 @@ func (s *MockService) NewHTTPMux(ctx context.Context, conn *grpc.ClientConn) (ht
 		return nil, err
 	}
 
+	// Returns slightly non-standard errors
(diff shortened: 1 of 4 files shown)


mockgcp/mockbigtable/service.go
```diff
@@ -69,6 +69,13 @@ func (s *MockService) NewHTTPMux(ctx context.Context, conn *grpc.ClientConn) (ht
 		return nil, err
 	}
 
+	// Returns slightly non-standard errors
```
justinsb, line 72: I actually don't think these are non-standard errors, I just think we started with the one service that follows a different pattern.  We should address that the default should be the other way round, but not in this PR I suggest.

yuwenma (commented):
/lgtm
/approve

Thanks! cc @djyau

PR #4120, opened 2025-03-20
tests: use server-side-apply for create and update

justinsb:
If we mix-and-match, we get surprising behaviours e.g. we can't clear a field

### DIFF
--- a/config/tests/samples/create/samples.go
+++ b/config/tests/samples/create/samples.go
@@ -117,9 +117,12 @@ type CreateDeleteTestOptions struct { //nolint:revive
 func RunCreateDeleteTest(t *Harness, opt CreateDeleteTestOptions) {
 	ctx := t.Ctx
 
+	// Note: we use server-side apply for both create and update.
+	// If we mix-and-match, we get surprising behaviours e.g. we can't clear a field
+
 	// Create and reconcile all resources & dependencies
 	for _, u := range opt.Create {
-		if err := t.GetClient().Create(ctx, u); err != nil {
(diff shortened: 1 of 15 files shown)


config/tests/samples/create/samples.go
```diff
@@ -117,9 +117,12 @@ type CreateDeleteTestOptions struct { //nolint:revive
 func RunCreateDeleteTest(t *Harness, opt CreateDeleteTestOptions) {
 	ctx := t.Ctx
 
+	// Note: we use server-side apply for both create and update.
+	// If we mix-and-match, we get surprising behaviours e.g. we can't clear a field
+
 	// Create and reconcile all resources & dependencies
 	for _, u := range opt.Create {
-		if err := t.GetClient().Create(ctx, u); err != nil {
+		if err := t.GetClient().Patch(ctx, u, client.Apply, client.FieldOwner("kcc-tests")); err != nil {
```
jingyih, line 125: Yeah. During a PR review, I noticed that the field is not being unset later in https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/8b5880650c8086c4f89f8b15c4c1ac24b1ff01a8/config/tests/samples/create/samples.go#L137

To unblock that particular PR, the mitigation I suggested was to use the script test, which correctly unset a field.

jingyih: I spot checked one test result; it looks like we are unsetting the field now.

justinsb: Thanks for checking @jingyih ... yeah, I'm surprised this hasn't come up a lot more often and I am sort of scared to look at the test results!

ldanielmadariaga:

PR #4122, opened 2025-03-20
controllerbuilder: find better examples when prompt

jingyih:
When generating mock or controller, pick examples from the same proto service (which also means the same API group). If no such examples exist, fallback to the previous behavior.

### DIFF
--- a/dev/tools/controllerbuilder/pkg/toolbot/csv.go
+++ b/dev/tools/controllerbuilder/pkg/toolbot/csv.go
@@ -207,15 +207,31 @@ func (x *CSVExporter) BuildDataPoints(ctx context.Context, description string, s
 // pickExamples returns the examples we should feed into the promp
 func (x *CSVExporter) pickExamples(input *DataPoint) []*DataPoint {
 	var examples []*DataPoint
+	var sameAPIGroupExamples []*DataPoint
+
 	// We only include data points for the same tool as the input.
 	for _, dataPoint := range x.dataPoints {
 		if dataPoint.Type != input.Type {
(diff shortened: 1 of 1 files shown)


jingyih:
/assign @justinsb 
Please let me know your thoughts!

dev/tools/controllerbuilder/pkg/toolbot/csv.go
```diff
@@ -207,16 +207,32 @@ func (x *CSVExporter) BuildDataPoints(ctx context.Context, description string, s
...
+		case "mockgcp-support", "controller":
+			// collect examples with the same proto service (API group)
+			if dataPoint.Input["proto.service"] == input.Input["proto.service"] {
+				sameAPIGroupExamples = append(sameAPIGroupExamples, dataPoint)
+			}
 		}
 		examples = append(examples, dataPoint)
 	}
+
+	// If we have examples with the same API group, use only those
```
justinsb, line 232: Looks good.  I might pick the threshold as 3 or something, but I do think we can evolve this.

(e.g. if I can't find exact api group matches, maybe there's another attribute I can do similarity analysis on.  Or should we randomly sample the prompts and try N evaluations and see which one is best).

Anyway, lots of potential here, but we have to start simple!

justinsb (approved):

dev/tools/controllerbuilder/pkg/toolbot/csv.go
jingyih: Thanks for the suggestion and ideas!

PR #4127, opened 2025-03-21
feat:mockgcp: BackupPolicy

acpana:
Mostly `conductor` driven mockgcp for `BackupPolicy`. Same thing as the `BackupVault` gcloud script, there's not "GA" version so can't use the mock implementation against the gcloud commands unless we re-write the API paths.

Already have the controller PR for this out!


yuwenma (commented):
/lgtm
/approve

We need some follow-up work  to prove this mock works since it does not have the gcloud part as Alex explained in the PR description.

PR #4129, opened 2025-03-21
feat: add virtual field `spec.IgnoreInactiveDatasources` in resource `BackupDRBackupVault`

jingyih:
This is necessary because there is no API to delete empty data sources nested under a backup vault.  

When we create a `BackupDRBackupVault` and associate it with a `BackupDRBackupPlanAssociation`, data sources are generated and nested within the backup vault. To delete the backup vault, we must use force delete [1]. However, the API semantics for this force delete operation have changed, as reflected in the gcloud helper text:  

```
     --ignore-inactive-datasources
        If set, the following restrictions against deletion of the backup vault
        instance can be overridden: * deletion of a backup vault instance
        containing no backups, but still containing empty data sources.
```

[1] https://github.com/googleapis/googleapis/blob/bad8d92019660beca95396e71901d783fb8e1273/google/cloud/backupdr/v1/backupvault.proto#L806


jingyih:
This PR is required to unblock #4131.

pkg/controller/direct/backupdr/backupdrbackupvault_controller.go
```diff
@@ -250,6 +250,10 @@ func (a *BackupVaultAdapter) Delete(ctx context.Context, deleteOp *directbase.De
 	log.V(2).Info("deleting BackupVault", "name", a.id)
 
 	req := &pb.DeleteBackupVaultRequest{Name: a.id.String()}
+
+	if a.desired.Spec.IgnoreInactiveDatasources != nil && *a.desired.Spec.IgnoreInactiveDatasources {
+		req.Force = true
+	}
```
yuwenma, line 256: Another option is to use the existing "cnrm.cloud.google.com/force-destroy" instead. But, I like this approach more considering the simplicity and specification of the API.   https://cloud.google.com/config-connector/docs/reference/annotations

jingyih: It looks like there are other deletion options besides "force" we might want to support in the future. Not sure we want to keep adding annotations.

[ignore_backup_plan_references](https://github.com/googleapis/googleapis/blob/bad8d92019660beca95396e71901d783fb8e1273/google/cloud/backupdr/v1/backupvault.proto#L823C8-L823C37)

PR #4136, opened 2025-03-21
conductor: type/crd/mapper/fuzzer improvements

barney-s:
conductor: type/crd/mapper/fuzzer improvements
* Add more llm based fixes for types:
  * removeNameField - Remove Name field in Spec (ResourceID instead) and Status (ExternalRef instead)
  * moveEtagField - Belongs to Status
  * addRequiredFieldTags - add KRM +reuired tag
* Improve existing prompts for:
  * fixing mockgcp failures
* Autofix: Add fuzz testing and fix loop
* Autofix: Add apichecks and fix loop
* Add crunAPIChecks to run apichecks as part of cmd=9
* Add --skip-makereadypr when pushing changes (cmd=9)
* Add ability to run make ready-pr until no changes are detected.
  Sometime make ready-pr makes a change once and we need to rerun make
  ready-pr again. So add AttemptsOnChanges to handle this case.

### DIFF
--- a/experiments/conductor/cmd/runner/crd_generator_commands.go
+++ b/experiments/conductor/cmd/runner/crd_generator_commands.go
@@ -184,10 +184,7 @@ func generateTypes(ctx context.Context, opts *RunnerOptions, branch Branch, exec
 			MaxAttempts: 1,
 		}
 		results, err := executeCommand(opts, cfg)
-		if err != nil {
-			return nil, nil, fmt.Errorf("failed to generate types: %w", err)
-		}
-		return affectedPaths, &results, nil
+		return affectedPaths, &results, err
 	}
 
 	log.Printf("SKIPPING generating apis, %s already exists", apiDirPathRelative)
@@ -741,6 +738,7 @@ func adjustIdentityParent(ctx context.Context, opts *RunnerOptions, branch Branc
 
 const ADJUST_IDENTITY_PARENT_NEW_FUNCTION string = `I want you to update the New${PROTO_RESOURCE}Identity method in the ${IDENTITY_FILE} file.
 
+if the ${PROTO_RESOURCE}Parent struct has a Name field, please remove it.
 Main Objectives:
 1. Modify the New${PROTO_RESOURCE}Identity method in ${IDENTITY_FILE}.
 2. Please do not modify the ${RESOURCE_TYPES_FILES} file.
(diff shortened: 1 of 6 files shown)


experiments/conductor/cmd/runner/utilities.go
```diff
@@ -721,6 +754,10 @@ func runGoCmds(opts *RunnerOptions, affectedPaths []string) error {
 	errStrings := []string{}
 
 	for _, path := range affectedPaths {
+		// Skip if not a Go file
```
barney-s, line 757: @jingyih - https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4177 - This should be sufficient right ?

jingyih: So the affectedPath is a folder. It may actually contain go files?

barney-s: makes sense i will approve your PR. Please review rest of this PR

experiments/conductor/cmd/runner/runner.go
```diff
@@ -116,27 +118,27 @@ func BuildRunnerCmd() *cobra.Command {
...
+	timeout           time.Duration
+	readFileType      string
+	defaultRetries    int    // Default number of retries for commands
+	forResources      string // Comma-separated list of branch names to filter on
+	forResourcesRegex string // Regex to filter branches, only branches that match the regex are processed
+	force             bool   // Force flag to override file existence checks
+	numCommits        int    // Number of commits to diff (default: 1)
+	verbose           bool   // Verbose output flag
+	branchSuffix      string // Suffix to append to remote branch names when pushing
+	skipMakeReadyPR   bool   // Skip make ready-pr step when pushing branches
```
jingyih, line 141: I feel the default behavior should be to "skip"? In other words, the flag should be named "makeReadyPR," and developers opt in by setting it to true. I personally only perform manual pushes, but it might be worth checking with those who use this command regularly.

experiments/conductor/cmd/runner/runner.go
```diff
@@ -341,10 +343,11 @@ func RunRunner(ctx context.Context, opts *RunnerOptions) error {
 		}
 	case cmdPushBranch: // 9
 		processors := []BranchProcessor{
-			{Fn: makeReadyPR, CommitMsgTemplate: "make ready-pr"},
+			{Fn: runAPIChecks, CommitMsgTemplate: "{{kind}}: Normalize api checks"},
```
jingyih, line 346: APICheck is different from "make ready-pr." It is a tool that helps developers identify missed reference fields, sensitive fields, acronyms, etc. Normally, I don't think this check should be part of the PR. Instead, developers should correct the differences in the golden output rather than simply overriding it.

experiments/conductor/cmd/runner/crd_generator_commands.go
```diff
@@ -831,3 +829,369 @@ func adjustIdentityParentNewFunction(ctx context.Context, opts *RunnerOptions, b
...
+${TEST_OUTPUT}
+
+Focus on the additions (lines starting with '+') in the output. These indicate issues that need to be fixed.
+Ignore the lines starting with '-' as they are not relevant.
+
+Please help me fix these issues by inspecting and modifying the affected files in the codebase.
+
+Files that may need to be modified:
+- apis/${GROUP}/v1alpha1/${RESOURCE_LOWER}_types.go - Contains the resource type definitions
+- apis/${GROUP}/v1alpha1/types.generated.go - Contains the generated resource type definitions
```
jingyih, line 1090: Will codebot try to modify generated files in place?

barney-s: Yes

experiments/conductor/cmd/runner/runner.go
barney-s: cmd=26 runs and fixes API checks. This is to catch if the user missed that step so that it shows up in the PR and the reviewer can comment. It only takes a minute to run this. So more of a fail safe.

experiments/conductor/cmd/runner/runner.go
barney-s: We can change the defaulting later. I found this to be useful

experiments/conductor/cmd/runner/crd_generator_commands.go
jingyih: I believe this will become a major challenge in the future when we need to update resources with new fields. Our goal is to ensure that generated files remain free of manual edits. Related ongoing work includes: #4060, #4075, #4099, #4059.

experiments/conductor/cmd/runner/runner.go
jingyih: I believe the question is: If the user forgets to run cmd 26, should the PR override the golden test output, or should it just fail the test? I prefer the latter.

experiments/conductor/cmd/runner/runner.go
jingyih: Yeah I can see either way. This is not a blocker.

experiments/conductor/cmd/runner/runner.go
barney-s: Hmm there is a case where we dont want to fix the reference and instead add it to the golden test output. Atleast for one of my PRs where the reference object is not implemented in TF or KCC and does not have create/delete/list methods. 

Like i said its a failsafe and alerts the reviewer.

experiments/conductor/cmd/runner/crd_generator_commands.go
barney-s: I see your point. Whats best for automated generation? 
1.  Always and (maybe) automatically mv all types.generated.go content into <rsrc>_types.go ? Should generation take care of this ? 
2. Let conductor modify generated.go and let the user mv it manually. We need the user to review the files locally anyways and shows up as diff in vscode. 
   a. we can add a check in cmd=9 to check if there are changes in generated.go and fail it

jingyih: Is it possible to move only the types that need to be modified out of `types.generated.go`? This is the approach typically followed by developers.

Moving all types out of `types.generated.go` would work for now, but it would prevent us from easily updating new fields by rerunning the type generator. While it's still possible to update `xx_types.go` using an LLM, the tool to do so does not currently exist.

barney-s: Honestly i havent seen an edit so far in the types.generated.go to fix the API checks. 
That being said we can ask the LLM not to edit types.generated.go. 

But we cannot guarantee LLM wont change that file. Experience says if we run enough resources/iterations, we may see an edit. So the best way is to rely on author reviewing it after this step.

barney-s: added a line to instruct moving the modified types to resource_type.go

jingyih (commented):
/lgtm
/approve

Can we have a demo of the new commands so that other developers can also try them in the workflow?

PR #4137, opened 2025-03-21
tests: cover DataFlowFlexTemplateJob spec.parameters.autoscalingAlgorithm

justinsb:


justinsb:
Going to update this one to also cover the direct controller

yuwenma (commented):
/approve
/lgtm

Thanks. We also want to allow the `autoscalingAlgorithm`  to be AUTOSCALING_ALGORITHM_NONE so that users who by-pass the TF-based issue does not need to change again when switched to SciFi (opt-in or when set to default)

PR #4138, opened 2025-03-21
mockgcp: don't ignore unknown fields/enum values by default

justinsb:
We introduced this so that when the API client had newer fields
than we have in our protos, we would not treat that as an error.

However, this also means that we ignore unknown enum values,
which seems much more problematic.

Let's switch the default to strict mode, we should be OK for
most services!


yuwenma:
> We introduced this so that when the API client had newer fields than we have in our protos, we would not treat that as an error.
> 
> However, this also means that we ignore unknown enum values, which seems much more problematic.
> 
> Let's switch the default to strict mode, we should be OK for most services!

Agree! But this may cause some confusion for most developers if their own proto is not changed, but other proto is broken after upgrade. It is hard to identify the problem and provide a fix.

ldanielmadariaga:

PR #4142, opened 2025-03-21
feat: add resource batch task controller

xiaoweim:

### DIFF
--- a/pkg/controller/direct/batch/task_controller.go
+++ b/pkg/controller/direct/batch/task_controller.go
@@ -1,0 +1,176 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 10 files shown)


pkg/controller/direct/batch/task_controller.go
yuwenma, line 1: Does this resource only have GET call?

yuwenma: ok, looks like it is.

yuwenma (commented):
/lgtm
/approve

Thanks. Not sure if we should have a KCC object for this type

yuwenma (commented):
/hold

yuwenma (commented):
If no create or update, we should have the adapter Create and Update to make the KCC `.status` up to date with the corresponding GCP object. But I don't see the golden-object generated.

xiaoweim:
> If no create or update, we should have the adapter Create and Update to make the KCC `.status` up to date with the corresponding GCP object. But I don't see the golden-object generated.

Yes the creation of BatchTask is dependent on the creation of BatchJob. The support for BatchJob was only recently merged and udpated. Let me rebase and update this PR.

xiaoweim:
Closing this PR since we do not support terraform resources that are read only.

PR #4151, opened 2025-03-22
Mocks for resource metastore-service

barney-s:


yuwenma (commented):
/lgtm
/approve

Thanks. One question, non blocker

mockgcp/mockmetastore/testdata/service/crud/script.yaml
```diff
@@ -0,0 +1,4 @@
+- exec: gcloud metastore services create test-${uniqueId} --location=us-central1 --project=${projectId}
+- exec: gcloud metastore services describe test-${uniqueId} --location=us-central1 --project=${projectId}
+#- exec: gcloud metastore services update test-${uniqueId} --location=us-central1 --project=${projectId} --clear-labels
```
yuwenma, line 3: Can we add a comment to tell what problems we hit here? It could save some investigation effort when promoting this resource to Beta (update required)

barney-s: This commit log has the info: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4151/commits/f4a51d740380f4c79a4b8430d6bf11dc5a0ff9a3

PR #4153, opened 2025-03-22
conductor: create commit messages that are unique/descriptive

justinsb:


experiments/conductor/cmd/runner/utilities.go
```diff
@@ -207,7 +207,7 @@ func gitAdd(ctx context.Context, workDir string, files ...string) error {
 
 func gitCommit(ctx context.Context, workDir string, msg string) error {
 	log.Printf("COMMAND: git commit -m %q", msg)
-	gitcommit := exec.CommandContext(ctx, "git", "commit", "-m", fmt.Sprintf("conductor: %q", msg))
+	gitcommit := exec.CommandContext(ctx, "git", "commit", "-m", msg)
```
jingyih, line 210: We probably still want to include something like `autogen: "` to indicate that this is a generated commit, or should we rely solely on the commit message?

experiments/conductor/cmd/runner/runner.go
```diff
@@ -341,62 +341,62 @@ func RunRunner(ctx context.Context, opts *RunnerOptions) error {
...
+			{Fn: makeReadyPR, CommitMsgTemplate: "make ready-pr"},
+			{Fn: pushBranch, CommitMsgTemplate: "push branch"},
 		}
 		processBranches(ctx, opts, branches.Branches, "Pushing Branch", processors)
 	case cmdCreateScriptYaml: // 10
-		processBranches(ctx, opts, branches.Branches, "Script YAML", []BranchProcessor{{Fn: createScriptYaml, CommitMsg: "Create gcloud script.yaml"}})
+		processBranches(ctx, opts, branches.Branches, "Script YAML", []BranchProcessor{{Fn: createScriptYaml, CommitMsgTemplate: "mockgcp: create test script for {{command}}"}})
 	case cmdCaptureHttpLog: // 11
-		processBranches(ctx, opts, branches.Branches, "HTTP Log", []BranchProcessor{{Fn: captureHttpLog, CommitMsg: "Capture HTTP Log for mocks"}})
+		processBranches(ctx, opts, branches.Branches, "HTTP Log", []BranchProcessor{{Fn: captureHttpLog, CommitMsgTemplate: "mockgcp: golden output for TestScripts/mock{{group}}/testdata/{{resource}}/crud"}})
```
jingyih, line 351: Other commands face the same issue. For example, "Add proto generation to makefile", "Add generated types". I’m thinking maybe we just prefix every commit in conductor with {{group}} and {{resource}}?

jingyih (commented):
Thank you for improving this!

experiments/conductor/cmd/runner/utilities.go
justinsb: So if the commit is automatically generated by running a script, yes, we should include autogen.  (And I think I've started doing that)

But when the code is generated by an LLM, that's a little different.  We do expect some edits, and I `git commit --fixup` those commits.  I don't think in the long term we will be able to draw the line here, like we don't include `[autocomplete]` when we are using autocomplete.  We could include a co-author statement (`Co-Author: Gemini 2.0 Flash`), but TBH  I think even that is redundant today.

experiments/conductor/cmd/runner/runner.go
justinsb: That could work, though conductor will evolve over time and I think the messages should be about what we are doing, not the tools we used to create the code.

experiments/conductor/cmd/runner/utilities.go
jingyih: Yeah this makes sense.

jingyih (commented):
/lgtm
/approve

PR #4157, opened 2025-03-22
controllerbuilder: avoid external dependencies

justinsb:
In particular, if we depend on apis/ we cannot make progress when apis is not compiling,
for example when we have not generated the types.


jingyih (commented):
/lgtm
/approve

PR #4169, opened 2025-03-24
fix: fuzztest BatchJobSpec

xiaoweim:

### DIFF
--- a/pkg/controller/direct/batch/job_mappings.go
+++ b/pkg/controller/direct/batch/job_mappings.go
@@ -99,7 +99,7 @@ func AllocationPolicy_FromProto(mapCtx *direct.MapContext, in *pb.AllocationPoli
 	out.Instances = direct.Slice_FromProto(mapCtx, in.Instances, AllocationPolicy_InstancePolicyOrTemplate_FromProto)
 	if in.GetServiceAccount() != nil {
 		out.ServiceAccountRef = &v1beta1.IAMServiceAccountRef{
-			External: in.GetServiceAccount().String(),
+			External: in.GetServiceAccount().GetEmail(),
 		}
 	}
 	out.Labels = in.Labels


pkg/controller/direct/batch/job_mappings.go
```diff
@@ -99,7 +99,7 @@ func AllocationPolicy_FromProto(mapCtx *direct.MapContext, in *pb.AllocationPoli
 	out.Instances = direct.Slice_FromProto(mapCtx, in.Instances, AllocationPolicy_InstancePolicyOrTemplate_FromProto)
 	if in.GetServiceAccount() != nil {
 		out.ServiceAccountRef = &v1beta1.IAMServiceAccountRef{
-			External: in.GetServiceAccount().String(),
+			External: in.GetServiceAccount().GetEmail(),
```
jingyih, line 102: What about `ServiceAccount.Scopes`? Seems we are just dropping the field?

xiaoweim: I had a discussion with @gemmahou  and @maqiuyujoyce regarding the scope field, I am going to have a followup PR to add the field as a list of projectRef for now.

xiaoweim: @jingyih do you have specific concerns on the scope field?

jingyih: Do users need to configure the "Scopes" field? My only concern is that we're dropping the field, so that users have no way to configure it.

xiaoweim: I see. This is currently an alpha resource so I don't think it will have a great impact on users, and I will have a followup PR to address this concern. I did not put it here because I wanted it to be a simple fix.

jingyih (commented):
/lgtm
/approve

PR #4171, opened 2025-03-24
chore: unregister batch from direct

jingyih:
Unblock other PRs while we fix fuzzer for `BatchJobSpec`.


xiaoweim:
Actually, if you don't mind I address the concern that you have mentioned in a followup PR, can we merge https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4169 to unblock the HEAD? I am working on another resource in the same API that depends on this resource.

jingyih:
/close

Fixed by #4169

PR #4172, opened 2025-03-25
Support direct controller for ReCAPTCHAEnterpriseFirewallPolicy

maqiuyujoyce:
### Change description

ReCAPTCHAEnterpriseFirewallPolicy is supported based on the GRPC client.

### Tests you have done

These is no commit for test result against mock API because there is no diff to commit.


pkg/controller/direct/recaptchaenterprise/firewallpolicy_controller.go
```diff
@@ -0,0 +1,252 @@
...
+	return createOp.UpdateStatus(ctx, status, nil)
+}
+
+// Update updates the resource in GCP based on `spec` and update the Config Connector object `status` based on the GCP response.
+func (a *FirewallPolicyAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOperation) error {
+	log := klog.FromContext(ctx)
+	log.V(2).Info("updating FirewallPolicy", "name", a.id)
+	mapCtx := &direct.MapContext{}
+
+	desiredPb := ReCAPTCHAEnterpriseFirewallPolicySpec_ToProto(mapCtx, &a.desired.Spec)
```
yuwenma, line 172: Can you double check if user can specify `name` in the Create call. If so, this logic need to be updated to support custom id

yuwenma (commented):
/approve

pkg/controller/direct/recaptchaenterprise/firewallpolicy_controller.go
maqiuyujoyce: The resource has service-generated ID so the name shouldn't be set during Create. Even if it is set, the value won't be respected by the API.

PR #4178, opened 2025-03-25
feat: add resource apphub discoveredservice controller

xiaoweim:
This resource only has Get and List Method from the [REST API Documentation](https://cloud.google.com/app-hub/docs/reference/rest/v1/projects.locations.discoveredServices). I not sure if it is worth/need to have a controller to support resources like this.

The documentation says that this resource is created via " A discovered service can be registered to a App Hub service". I have not confirmed how it works yet.


yuwenma:
Can you share the git-diff between real and mock GCP?

xiaoweim:
> Can you share the git-diff between real and mock GCP?

Sure. But I have double checked with @maqiuyujoyce and @justinsb that we do not support read-only resources from TF in KCC. So I will be closing this PR.

PR #4184, opened 2025-03-26
Plumb through GCPAccessToken if provided in controllers

justinsb:
Likely only done in tests, but if set we should pass it down,
otherwise it does not make it to the GRPC clients (for example).


jingyih (commented):
/lgtm
/approve

PR #4185, opened 2025-03-26
feat: update KRM status even when no fields changed

jingyih:
This PR fixes the update logic for resources in the following API groups:
 - GKEBackup
 - BackupDR
 - VMwareEngine

The fix ensures that KRM status is properly updated even when no fields have changed in spec.


yuwenma (commented):
/lgtm
/approve

Thanks!

PR #4189, opened 2025-03-27
feat: CloudDeployCustomTargetType controller

acpana:
Mostly `conductor` driven. 

PR needs mock implementation for mock testing

PR #4194, opened 2025-03-27
feat:mockgcp: CustomTargetType

acpana:
Mostly `conductor` drive; have the controller ready to integrate with in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4189


yuwenma:
I may miss something, but basically conductor-driven mockgcp would also have script.yaml and _http.log (which is how the codebot generates the mock service go code). But I don't see either file in this PR. 

Could you add a test suite to prove the generated code works.

acpana:
closing in favor of doing the mock and the controller together in #4189

PR #3987, opened 2025-03-13
Add dataplex environment types, mappers, CRD and fuzzer

gemmahou:
### Change description

Fixes #

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #4197, opened 2025-03-27
fix: Remove unnecessary projectRef and location from dataplex content types

gemmahou:
### Change description

Address comment https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3987#discussion_r2014576825 in dataplex content resource as well.

### Tests you have done


apis/dataplex/v1alpha1/content_types.go
```diff
@@ -15,18 +15,13 @@
...
 )
 
 var DataplexContentGVK = GroupVersion.WithKind("DataplexContent")
 
 type DataplexContentParent struct {
-	ProjectRef *refs.ProjectRef `json:"projectRef"`
-
-	Location string `json:"location"`
-
 	LakeRef *LakeRef `json:"lakeRef,omitempty"`
```
jingyih, line 25: nit: I think by convention we don't add "Ref" here. We usually use something like "projectID", "location" in the parent.

gemmahou: Well, `Lake` is part of the parent per the doc: https://cloud.google.com/dataplex/docs/reference/rest/v1/projects.locations.lakes.content/create. Perhaps it would be clearer to include all parent instances here (not just projectRef and Location), so that this Parent is consistent with the Parent in resource_identity?

jingyih: Sorry for the confusion. Yes, `Lake` should be part of the parent structure. I was thinking about something like this:
```
type DataplexContentParent struct {
    DataplexLake string
}
```

Which is similar to what you already have:
```
type LakeParent struct {
	ProjectID string
	Location  string
}
```

jingyih: In my opinion, specifying `DataplexLake` as the parent of `DataplexContent` implies that the project ID and location are also its parent, since they are the parent of `DataplexLake`. However, this is just my interpretation and open to debate.

gemmahou: Oh I see. I think `Parent` struct looks like this([doc](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/develop-resources/api-conventions/validations.md#rule-3-parent)):
```  
    ProjectRef *refv1beta1.ProjectRef `json:"projectRef"`

    Location string `json:"location"`
```
Where project is a reference not string.

So similarly if parent is the DataplexLake, it should be a reference to DataplexLake resource...

jingyih: I think I missread the code. Resolving this conversation.

jingyih (commented):
/lgtm
/approve

PR #4198, opened 2025-03-27
Add label filtering to Spanner Instance and Secret Manager Secret

anhdle-sso:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/pkg/controller/direct/common/labels.go
+++ b/pkg/controller/direct/common/labels.go
@@ -15,6 +15,7 @@
 package common
 
 import (
+	"regexp"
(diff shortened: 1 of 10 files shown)


yuwenma:
Suggest differentiating the metadata label and spec labels. Note SecretManager is still an opt-in approach

anhdle-sso:
> Suggest differentiating the metadata label and spec labels. Note SecretManager is still an opt-in approach

This is the fix for current broken labels logic. I will create another PR to split out the labels in these resources

PR #4199, opened 2025-03-27
fix: Do not send patch request if there's no update on FirewallPolicyRule

gemmahou:
### Change description

Fixes b/406862735

### Tests you have done


tests/e2e/testdata/scenarios/computefirewallpolicyrule/noupdate/script.yaml
```diff
@@ -0,0 +1,55 @@
...
+  firewallPolicyRef:
+    name: firewallpolicy-${uniqueId}
+  match:
+    layer4Configs:
+      - ipProtocol: "tcp"
+    srcIPRanges:
+      - "10.100.0.1/32"
+  priority: 9000
+---
+apiVersion: compute.cnrm.cloud.google.com/v1beta1
```
maqiuyujoyce, line 41: Have you confirmed that before your change, a PATCH is triggered in this test?

maqiuyujoyce (commented):
/lgtm

Just one question about the scenario test.

tests/e2e/testdata/scenarios/computefirewallpolicyrule/noupdate/script.yaml
gemmahou: I added [this commit ](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4199/commits/337982f1339dcd3a6f4edce976c912a5f6821a6d) to collect log without this change. I added "TOUCH" test flag to trigger a reconcile and confirmed that PATCH is triggered in the log.

pkg/controller/direct/compute/firewallpolicyrule/firewallpolicyrule_controller.go
```diff
@@ -196,6 +198,15 @@ func (a *firewallPolicyRuleAdapter) Update(ctx context.Context, updateOp *direct
 	// See API doc: https://cloud.google.com/compute/docs/reference/rest/v1/firewallPolicies/patchRule#query-parameters
 	firewallPolicyRule.Priority = nil
 
+	paths, err := common.CompareProtoMessage(firewallPolicyRule, a.actual, common.BasicDiff)
+	if err != nil {
+		return err
+	}
+	if len(paths) == 0 {
+		log.V(2).Info("no field needs update", "name", a.id.String())
+		return nil
```
yuwenma, line 207: We should not just return nil here. Instead, we want to call updateOp.UpdateStatus(ctx, status, nil) to make sure the status is updated. This is especially important for first time acquire.

gemmahou: Updated in #4369

PR #4200, opened 2025-03-27
controllerbuilder: include version in KRM import alias

jingyih:
This PR adds the KRM API version to the import alias when generating mappers.  

This change is necessary when both v1alpha1 and v1beta1 versions of the KRM API exist. Without it, the imports would appear as follows:  

```go
import (
	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/bigtable/v1alpha1"
	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/bigtable/v1beta1"
	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
)
```

This makes it difficult to distinguish which "krm" corresponds to which version in the code.

Example output:
https://github.com/GoogleCloudPlatform/k8s-config-connector/commit/5cd3d193b72160358cd760ecc4a7f67185089a69


yuwenma (commented):
/lgtm
/approve

Thanks!

PR #4208, opened 2025-03-28
feat: add controller, test and mock support for DatastreamConnectionProfile

jingyih:


yuwenma (commented):
/lgtm
/approve

Thanks! The mock is better to handle `requestedCancellation` as well. Non blocker though

PR #4221, opened 2025-04-01
chore: add a note on mapper generator limitation

jingyih:
Adding a note about a limitation in the mapper generator. This issue occurs when we promote a resource that shares the same Go structs as other alpha resources in the same API group. In such cases, manual fixes to the mapping functions are required. Hopefully, this situation is not too common.

We should address this eventually, but we need to consider how to structure the mapping functions to handle the same Go struct across different versions. Add a TODO for now.

### DIFF
--- a/dev/tools/controllerbuilder/pkg/codegen/mappergenerator.go
+++ b/dev/tools/controllerbuilder/pkg/codegen/mappergenerator.go
@@ -80,6 +80,10 @@ func (v *MapperGenerator) VisitGoCode(goPackage string, basePath string) error {
 							if _, ok := v.precomputedMappings[protoName]; !ok {
 								v.precomputedMappings[protoName] = make(map[string]*gocode.GoStruct)
 							}
+
+							// TODO: When promoting a resource from v1alpha1 to v1beta1, the same Go struct will exist in both the v1alpha1 and v1beta1 KRM API folders.
+							// Currently, this code overrides v1alpha1 with v1beta1 in this mapping during visiting the go code.
+							// However, this can cause issues if the promoted resource shares a nested Go struct with another v1alpha1 resource.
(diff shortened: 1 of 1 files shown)


dev/tools/controllerbuilder/pkg/codegen/mappergenerator.go
```diff
@@ -80,6 +80,10 @@ func (v *MapperGenerator) VisitGoCode(goPackage string, basePath string) error {
 							if _, ok := v.precomputedMappings[protoName]; !ok {
 								v.precomputedMappings[protoName] = make(map[string]*gocode.GoStruct)
 							}
+
+							// TODO: When promoting a resource from v1alpha1 to v1beta1, the same Go struct will exist in both the v1alpha1 and v1beta1 KRM API folders.
+							// Currently, this code overrides v1alpha1 with v1beta1 in this mapping during visiting the go code.
+							// However, this can cause issues if the promoted resource shares a nested Go struct with another v1alpha1 resource.
```
yuwenma, line 86: Agree. But I would suggest changing this comment to a println. Otherwise, developers who encounter this issue still have to triage the code until they read this line.

yuwenma (commented):
/approve

One nit.

dev/tools/controllerbuilder/pkg/codegen/mappergenerator.go
jingyih: We don't want to print every time we run this code, as it would be a red herring. Instead, printing should only occur when "promoted resource shares a nested Go struct with another v1alpha1 resource in the same API group" This requires correctly identifying when this condition is met, which is already a significant step toward solving the issue. Addressing this will be part of future work.

PR #4242, opened 2025-04-02
template: use id.ID

acpana:
I noticed that I kept changing this line for conductor/ controllerbuilder. I think the right thing to do here is to use the ID (read name) from the resource identity.

### DIFF
--- a/dev/tools/controllerbuilder/template/controller/controller.go
+++ b/dev/tools/controllerbuilder/template/controller/controller.go
@@ -277,7 +277,7 @@ func (a *{{.ProtoResource}}Adapter) Export(ctx context.Context) (*unstructured.U
 		return nil, err
 	}
 
-	u.SetName(a.actual.Id)
+	u.SetName(a.id.ID())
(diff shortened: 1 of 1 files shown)


dev/tools/controllerbuilder/template/controller/controller.go
```diff
@@ -277,7 +277,7 @@ func (a *{{.ProtoResource}}Adapter) Export(ctx context.Context) (*unstructured.U
 		return nil, err
 	}
 
-	u.SetName(a.actual.Id)
+	u.SetName(a.id.ID())
```
jingyih, line 280: I'm actually not sure. Should this be just the "ID" or the fully qualified name? My recent controllers were generated by LLM, and they use `u.SetName(a.actual.Name)`, which is fully qualified.

jingyih: Ah, I was looking at the dynamic test framework and realized this should NOT be the fully qualified name.

jingyih (commented):
/lgtm
/approve

PR #4210, opened 2025-03-28
Update spanner instance label to spec

anhdle-sso:
### Change description

Fixes #

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #4246, opened 2025-04-03
test: add back the samples test in presubmit

jingyih:


yuwenma (commented):
good catch. Do you know why the sample test only runs spanner-connection previously?

yuwenma:
> good catch. Do you know why the sample test only runs spanner-connection previously?

My bad, I missed the code change in #4210.

PR #4257, opened 2025-04-03
feat: properly resolve reference fields in GKEBackupBackupPlan controller

jingyih:


pkg/controller/direct/gkebackup/restoreplan_controller.go
```diff
@@ -68,7 +68,7 @@ func (m *restorePlanModel) AdapterForObject(ctx context.Context, reader client.R
 		return nil, err
 	}
 
-	// normalize reference fields
+	// normalize required reference fields
 	if obj.Spec.BackupPlanRef != nil {
```
yuwenma, line 72: My original suggestion is to use the CRD `+ required` to guardrail this? From the controller code, it seems that the field is optional (comment does not validate the code).

yuwenma (commented):
/lgtm
/approve

pkg/controller/direct/gkebackup/restoreplan_controller.go
jingyih: It is a required field. But I usually still check for nil before using a pointer, as an extra layer of safety at runtime.
https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/22df94d5df044abc7dcf36b55db16830b4a8deca/apis/gkebackup/v1alpha1/restoreplan_types.go#L190-L195

PR #4258, opened 2025-04-04
feat: add controller, test and mock support for GKEBackupBackup

jingyih:
There are too many diff in the GKE container, causing GitHub's diff system to fail. Here is the actual diff between the real and mock for `GKEBackupBackup`: https://www.diffchecker.com/cRxkr5KT/

### DIFF
--- a/config/tests/samples/create/harness.go
+++ b/config/tests/samples/create/harness.go
@@ -767,6 +767,7 @@ func MaybeSkip(t *testing.T, name string, resources []*unstructured.Unstructured
 
 			case schema.GroupKind{Group: "batch.cnrm.cloud.google.com", Kind: "BatchJob"}:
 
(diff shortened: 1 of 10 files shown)


yuwenma:
How long does the creation takes? That "in process"  state makes me wondering if we should specifically handle the LRO.

yuwenma:
/approve
/lgtm
/hold

One question but non blocker. hold in case you can take a look

jingyih:
> How long does the creation takes? That "in process" state makes me wondering if we should specifically handle the LRO.

I haven’t looked into this in detail. The test already takes a long time to run because it requires creating a GKE cluster as a dependency. The "backup" GCP resource itself finishes creating, but the actual backup completion likely depends on the number of resources in the GKE cluster and the backup plan (the parent of the backup) settings. This shouldn’t block the PR. In the future, we could consider having the controller poll and wait until the backup either succeeds or fails.

jingyih:
Let's merge.
/hold cancel

PR #4260, opened 2025-04-04
Revert "Add CRD and Mappers for resource edgecontainer vpnconnection"

yuwenma:
This changes a Beta resource API to backward incompatible, and changed the TF/DCL-based controller to direct (which is not ready yet)


jingyih (commented):
/approve

We need to revert the corresponding entries in the release note as well.

PR #4264, opened 2025-04-04
codebot: improvements to use gemini 2.5 model via gemini api

barney-s:
* add --llmclient=(vertexai*|gemini) option to select API endpoint.
* default is vertexai which is current behavior
* GEMINI 2.5 is only available in gemini api today
* add CODEBOT_LLM_CLIENT env to override --llmclient param
* add GEMINI_MODEL env to select gemini model to use
* gemini 2.5 requires tools to have schema pros. so adding a param to VerifyCode

### DIFF
--- a/dev/tools/controllerbuilder/pkg/llm/gemini.go
+++ b/dev/tools/controllerbuilder/pkg/llm/gemini.go
@@ -30,11 +30,15 @@ import (
 // BuildGeminiClient builds a client for the Gemini API.
 func BuildGeminiClient(ctx context.Context) (Client, error) {
 	var opts []option.ClientOption
-
 	if s := os.Getenv("GEMINI_API_KEY"); s != "" {
 		opts = append(opts, option.WithAPIKey(s))
 	}
 
+	model := os.Getenv("GEMINI_MODEL")
+	if model == "" {
+		model = "gemini-2.0-pro-exp-02-05"
+		// model = "gemini-2.5-pro-exp-03-25"
+	}
 	client, err := genai.NewClient(ctx, opts...)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


dev/tools/controllerbuilder/pkg/llm/gemini.go
```diff
@@ -30,19 +30,23 @@ import (
...
 	var opts []option.ClientOption
-
 	if s := os.Getenv("GEMINI_API_KEY"); s != "" {
 		opts = append(opts, option.WithAPIKey(s))
 	}
 
+	model := os.Getenv("GEMINI_MODEL")
+	if model == "" {
+		model = "gemini-2.0-pro-exp-02-05"
+		// model = "gemini-2.5-pro-exp-03-25"
```
yuwenma, line 40: May I ask why not default to 2.5?

dev/tools/controllerbuilder/pkg/llm/gemini.go
```diff
@@ -67,7 +71,7 @@ func (c *GeminiClient) StartChat(systemPrompt string) Chat {
 	model.SetTemperature(1)
 	model.SetTopK(40)
 	model.SetTopP(0.95)
-	model.SetMaxOutputTokens(8192)
+	model.SetMaxOutputTokens(65536)
```
yuwenma, line 74: Is this the max for 2.0 or 2.5? I thought we previously used the max (but I don't remember if that's 1.5 or 2.0)

dev/tools/controllerbuilder/pkg/llm/gemini.go
barney-s: I dont think we need this anymore. superseded by #4267

PR #4267, opened 2025-04-04
Conductor improvements and switch controllerbuilder to new default model

barney-s:
controllerbuilder: switch to new default model

Conductor: Improvements for Type and Mapper generation
* Tweak prompts
* Refactor code to move repetitive code under prepareCodebotPrompt (for
  crd steps)
* Add option to run subset of processors in multi-processor step like
  21. Rationale is to rerun processors that did not generate output.
  *  --processors=p1,p2,....
* Fix bugs with repetititve steps:
  * for step 9 where it was running make ready-pr many times
  * everywhere where changes were not generated


yuwenma (commented):
/lgtm
/approve

Woohoo!

PR #4273, opened 2025-04-04
template: update status

acpana:
Make template changes to update status even when no cloud provider fields need to be updated!


jingyih (commented):
I believe the controllers are LLM generated in the current workflow. But this LGTM.

/lgtm
/approve

PR #4283, opened 2025-04-07
feat: add controller, test and mock support for GKEBackupRestore

jingyih:
/hold
Based on #4258

There are too many diff in the GKE container, causing GitHub's diff system to fail. Here is the actual diff between the real and mock for `GKEBackupRestore`:
https://www.diffchecker.com/oQOHANek/


yuwenma (commented):
Is there a reason why the mock is so different than the real log.

jingyih:
> Is there a reason why the mock is so different than the real log.

I explained this in the PR description. It’s actually not the case, as you can see in the Diffchecker link (I manually copied the real and mock logs only relevant to this resource into Diffchecker)

yuwenma (commented):
/lgtm
/approve

Thanks

PR #4209, opened 2025-03-28
conductor: make conductor bot as author

jingyih:
Remove the "conductor: " prefix from the commit message and set conductor Bot as the commit author instead.

PR #4294, opened 2025-04-07
conductor: do not add conductor prefix to git commits

justinsb:
conductor is the tool, not the outcome.


jingyih (commented):
Had a discussion with Barni. The original idea was to track how many commits were made using the conductor tool. To remove the "conductor" prefix from commit messages while still being able to track these commits, I was exploring the idea of setting conductor as the commit author in #4209. If we want to go that route, we’d need to register a proper robot google email.

yuwenma (commented):
/lgtm
/approve

PR #4299, opened 2025-04-07
Update branch metadata

maqiuyujoyce:
This maps to the original branches-3.yaml.

Notes added explaining the edge case related to the particular resource(s).


maqiuyujoyce:
Updated via other channel.

PR #4305, opened 2025-04-08
feat: add controller, test and mock support for DatastreamRoute

jingyih:


yuwenma (commented):
/lgtm
/approve

I assume `routingConfig` is a new field, which means the new features are added so fast

PR #4306, opened 2025-04-08
MockGCP&Controller: Resource dataplex zone

gemmahou:
### Change description

diffbase #4297 

### Tests you have done


pkg/controller/direct/dataplex/zone_controller.go
```diff
@@ -0,0 +1,287 @@
...
+	//    "jsonOptions": {},
+	//    "schedule": ""
+	//  })
+	emptyDiscoverySpec := &pb.Zone_DiscoverySpec{
+		CsvOptions:  &pb.Zone_DiscoverySpec_CsvOptions{},
+		JsonOptions: &pb.Zone_DiscoverySpec_JsonOptions{},
+		Trigger:     &pb.Zone_DiscoverySpec_Schedule{Schedule: ""},
+	}
+	if zone.DiscoverySpec != nil {
+		if path := deep.Equal(zone.DiscoverySpec, a.actual.DiscoverySpec); len(path) != 0 {
```
jingyih, line 197: What is the difference between `deep.Equal` vs `reflect.DeepEqual`?

pkg/test/resourcefixture/testdata/basic/dataplex/v1alpha1/dataplexzone/dataplexzone-minimal/_http.log
```diff
@@ -187,7 +187,7 @@ OK
 GRPC /google.cloud.dataplex.v1.DataplexService/UpdateZone
 
 {
-  "updateMask": "description,discoverySpec",
+  "updateMask": "description",
```
jingyih, line 190: We should probably rerun the real GCP test after the controller code change in this commit.

gemmahou: I reran against realGCP and the test passed: https://screenshot.googleplex.com/3Dn2B2zfFmXsbbq.png

pkg/controller/direct/dataplex/zone_controller.go
gemmahou: DiscoverySpec and its subfields are **pointers** to the value of a type.
`reflect.DeepEqual` not only checks the values of the fields but also the identity of pointers when comparing structs. Two objects(a.desired and a.actual) have different pointer values, so even the subfield values are the same, `reflect.DeepEqual` still considers the top-level structs as unequal.
`deep.Equal` seems only compare the actual values of a field.

jingyih: I am not familiar with how we handle licensing when importing a third party library. A quick search tells me that it is probably fine. But could you try "github.com/google/go-cmp"?
https://github.com/go-test/deep?tab=MIT-1-ov-file#readme

gemmahou: Done

jingyih (commented):
/lgtm

PR #4314, opened 2025-04-09
Direct Controller for asset-feed

barney-s:


barney-s:
> I believe we need to test against real GCP and capture the log to ensure that the controller actually works.

Done,

pkg/controller/direct/asset/feed_controller.go
```diff
@@ -0,0 +1,327 @@
...
+	desired   *krm.AssetFeed
+	actual    *pb.Feed
+	reader    client.Reader
+}
+
+var _ directbase.Adapter = &feedAdapter{}
+
+func (a *feedAdapter) Find(ctx context.Context) (bool, error) {
+	log := klog.FromContext(ctx)
+	log.Info("---------getting asset feed", "name", a.id)
```
jingyih, line 125: nit: this is not consistent with other controllers. Same for the other logs.

```
log.V(2).Info // and also should not have the leading "---------"
```

pkg/controller/direct/asset/feed_controller.go
```diff
@@ -0,0 +1,327 @@
...
+	// Regardless of CreateFeed result (success or AlreadyExists),
+	// we need to ensure a.actual is populated by calling Find.
+	found := false
+	var findErr error
+
+	if err == nil {
+		log.Info("---------successfully created asset feed in gcp, attempting to fetch it", "name", a.id)
+		a.actual = actual
+		found = true
+	} else if errors.IsAlreadyExists(err) {
```
jingyih, line 191: Is there a specific reason for this check? Other adapters assume the resource doesn't exist within this function. The base controller always calls `Find()` to verify the resource's existence before proceeding with either `Create()` or `Update()`.

pkg/controller/direct/asset/feed_controller.go
```diff
@@ -0,0 +1,327 @@
...
+		return mapCtx.Err()
+	}
+	resource.Name = a.id.String() // Name must be set for update
+
+	paths, err := common.CompareProtoMessage(resource, a.actual, common.BasicDiff)
+	if err != nil {
+		return fmt.Errorf("comparing proto messages: %w", err)
+	}
+	// Name is output only, but is required in the update request object.
+	paths.Delete("name")
```
jingyih, line 235: Yeah, `CompareProtoMessage` doesn't work if the proto fields are tagged incorrectly. But I'd expect "name" not to appear in the paths since you've already set the `name` field in `resource` (therefore both `resource` and `a.actual` should have the same name field?).

pkg/controller/direct/asset/feed_controller.go
barney-s: Its just defensive coding i guess. LLM generated and i modified a bit.

pkg/controller/direct/asset/feed_controller.go
barney-s: i guess the mock and realgcp implementation were differing a bit. projectID vs projectNumber. ill comment it out and test.

pkg/controller/direct/asset/feed_controller.go
jingyih: I don't think we have any examples that has this code/pattern. Looks like LLM is becoming very creative.
https://github.com/search?q=repo%3AGoogleCloudPlatform%2Fk8s-config-connector%20path%3A%2F%5Epkg%5C%2Fcontroller%5C%2Fdirect%5C%2F%2F%20errors.IsAlreadyExists(err)&type=code

jingyih: I prefer we just remove this pattern from the PR. If this error happens, we should fix the base controller.

yuwenma (commented):
/lgtm
/approve

yuwenma (commented):
/lgtm
/approve

yuwenma (commented):
/lgtm
/approve

PR #4316, opened 2025-04-09
codebot: use gollm library

justinsb:
This avoids having to maintain the same code in two places.


dev/tools/controllerbuilder/cmd/codebot/main.go
```diff
@@ -52,6 +52,8 @@ type Options struct {
 	UIType   string
 	Project  string
 	Location string
+
+	Model string
```
jingyih, line 56: Is there a default model for codebot chat?

/cc @barney-s this PR may affect how conductor calls codebot.

jingyih (commented):
/lgtm

dev/tools/controllerbuilder/cmd/codebot/main.go
justinsb: Good call.  I also need to fix how project & location are plumbed through to vertex

PR #4335, opened 2025-04-10
crd types: disambiguate annotations on spec vs observedstate

justinsb:
This allows us to focus the LLM on the right examples for each.


jingyih (commented):
We will probably need to update the mapper generator.

justinsb:
cc @cheftako  this lets the LLM do the lifting

yuwenma:
/lgtm
/approve
/hold

+100! And I think this is better to go in asap otherwise it could encounter endless merge-conflicts. Defer to @justinsb

PR #4338, opened 2025-04-10
BigtableAppProfile: reuse struct name to slightly reduce codegen

justinsb:
- **BigtableAppProfile: Can reuse existing name**
- **autogen: apis/bigtable/generate.sh**


jingyih (commented):
/lgtm
/approve 

cc @djyau

PR #4342, opened 2025-04-11
Add troubleshooting suggestion to test log

maqiuyujoyce:


config/tests/samples/create/samples.go
```diff
@@ -194,7 +194,7 @@ func waitForReadySingleResource(t *Harness, u *unstructured.Unstructured, timeou
 			return true, nil
 		}
 		if u.Object["status"] == nil {
-			logger.Info("resource does not yet have status", "kind", u.GetKind(), "name", u.GetName())
+			logger.Info("resource does not yet have status", "kind", u.GetKind(), "name", u.GetName(), "troubleshooting", "if you keep seeing this error message, check if you've imported the service in pkg/controller/direct/register/register.go")
```
jingyih, line 197: I believe `waitForReadySingleResource` is used by all samples which includes TF and DCL based controllers as well. Would this cause confusion?

maqiuyujoyce: Good point. I updated the error message to clarify the suggestion only applies to direct resource.

jingyih (commented):
/lgtm

PR #4343, opened 2025-04-11
fix: KCC release misses the RBAC for CCC

yuwenma:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/dev/tasks/propose-tag
+++ b/dev/tasks/propose-tag
@@ -103,6 +103,10 @@ for channeldir in channels autopilot-channels; do
   find ${REPO_ROOT}/operator/${channeldir}/packages -type f -name "*.yaml" | xargs sed -i -e "s@gcr.io/cnrm-eap/@gcr.io/gke-release/@g"
 done
 
+# Update operator RBAC.  
+cp ${REPO_ROOT}/config/installbundle/components/clusterroles/cnrm_viewer.yaml ${REPO_ROOT}/operator/config/rbac/cnrm_viewer_patch.yaml
+
+
 # Generate operator/config/gke-addon/image_configmap.yaml
 make -C ${REPO_ROOT}/operator manifests
 
--- a/operator/config/rbac/cnrm_viewer_patch.yaml
+++ b/operator/config/rbac/cnrm_viewer_patch.yaml
@@ -1,3 +1,19 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# the CNRM viewer role allows the operator-manager to list CNRM resources which is useful when deciding to delete the
+# KCC manager pods that are managing a namespace.
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
--- a/operator/config/rbac/cnrm_viewer_patch.yaml
+++ b/operator/config/rbac/cnrm_viewer_patch.yaml
@@ -1,0 +1,960 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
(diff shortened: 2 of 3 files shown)


dev/tasks/propose-tag
```diff
@@ -103,6 +103,10 @@ for channeldir in channels autopilot-channels; do
   find ${REPO_ROOT}/operator/${channeldir}/packages -type f -name "*.yaml" | xargs sed -i -e "s@gcr.io/cnrm-eap/@gcr.io/gke-release/@g"
 done
 
+# Update operator RBAC.  
+cp ${REPO_ROOT}/config/installbundle/components/clusterroles/cnrm_viewer.yaml ${REPO_ROOT}/operator/config/rbac/cnrm_viewer_patch.yaml
+
```
yuwenma, line 108: This step is equivalent to the legacy release "copy-dependency-manifests" https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/operator/Makefile#L105, which is a very bad practice of using [go script to build a `ClusterRole` object on the fly](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/operator/scripts/copy-dependency-manifests/main.go#L98), and made a series of changes just to establish a simple kustomize patch.

This step is missed in the new KCC release and causes CCC lack RBAC for new groups.

operator/config/rbac/kustomization.yaml
```diff
@@ -18,4 +18,13 @@ resources:
 - cnrm_viewer_role.yaml
 - cnrm_viewer_role_binding.yaml
 - role.yaml
-- manager_role_binding.yaml
\ No newline at end of file
+- manager_role_binding.yaml
+
+patches:
```
yuwenma, line 23: This patch keeps the `cnrm_viewer_role.yaml` labels and annotations and only replace the `.rules` field.

jingyih: Could you provide an example output? I am not sure I follow how it works. In `${REPO_ROOT}/config/installbundle/components/clusterroles/cnrm_viewer.yaml`, we have a role name `viewer`, but in the `cnrm_viewer_role.yaml` the role name is `cnrm-viewer`

jingyih: It looks like the "copy-dependency-manifests" was copying from the rendered manifest which has `cnrm-` prefix already added.

jingyih:
When releasing version X, we should also copy the latest RBAC to the previously released versions X-1 and X-2 to ensure safe rollback.

yuwenma:
> When releasing version X, we should also copy the latest RBAC to the previously released versions X-1 and X-2 to ensure safe rollback.

I don't quite understand the suggestion. Are you talking about the 3-version rollback?  The operator RBAC is not changed (always the latest release), only the channel is switched  https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/master/operator/channels/packages/configconnector.  So I think we are good

operator/config/rbac/kustomization.yaml
yuwenma: Here's the hydrated diff. https://paste.googleplex.com/5782340843077632
"<" is removed, ">" is added.
The ClusterRole name is changed from `cnrm-viewer` to `configconnector-operator-cnrm-viewer`.
Many RBAC are added to new groups.

operator/config/rbac/cnrm_viewer_patch.yaml
yuwenma, line 1: This file is to show the rbac diff when propose-tag, and used to update the actual operator RBAC via kustomize

operator/config/rbac/kustomization.yaml
```diff
@@ -27,4 +27,4 @@ patches:
     group: rbac.authorization.k8s.io
     version: v1
     kind: ClusterRole
-    name: viewer
+    name: cnrm-viewer
```
jingyih, line 30: Ah. I see. With this change I could follow how it works.

jingyih (commented):
/lgtm
/approve

operator/config/rbac/kustomization.yaml
jingyih: I think my original confusion was because of the `name: viewer` in the patch.

jingyih:
> > When releasing version X, we should also copy the latest RBAC to the previously released versions X-1 and X-2 to ensure safe rollback.
> 
> I don't quite understand the suggestion. Are you talking about the 3-version rollback? The operator RBAC is not changed (always the latest release), only the channel is switched https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/master/operator/channels/packages/configconnector. So I think we are good

Yes, I’m referring to a 3-version rollback. It’s not directly related to this PR, but I wanted to mention it since we're updating the cluster roles. The `cnrm-admin` cluster role is used by some KCC controllers (not the operator). To ensure a safe rollback, we need to keep the X-1 and X-2 versions up to date when we release version X.

PR #4345, opened 2025-04-11
[Cloud Bigtable App Profile Beta Migration] Add direct controller

djyau:
### Change description

Fixes #

### Tests you have done


jingyih (commented):
We need to test the behavior of the new controller using the existing tests. Let's remove the annotation `"cnrm.cloud.google.com/tf2crd=true";"cnrm.cloud.google.com/stability-level=stable"` and regenerate the CRD. Then the resource should be reconciled by the new controller.

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/3bfa3c8c489c280e1d321e6b4f5d5b8c083f21fe/apis/bigtable/v1beta1/appprofile_types.go#L88C113-L88C195

djyau:
> We need to test the behavior of the new controller using the existing tests. Let's remove the annotation `"cnrm.cloud.google.com/tf2crd=true";"cnrm.cloud.google.com/stability-level=stable"` and regenerate the CRD. Then the resource should be reconciled by the new controller.
> 
> https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/3bfa3c8c489c280e1d321e6b4f5d5b8c083f21fe/apis/bigtable/v1beta1/appprofile_types.go#L88C113-L88C195

Done - removed the annotations and pushed the resulting generated CRD to the PR

pkg/controller/direct/bigtable/bigtableappprofile_controller.go
```diff
@@ -0,0 +1,284 @@
...
+}
+
+var _ directbase.Model = &modelBigtableAppProfile{}
+
+type modelBigtableAppProfile struct {
+	config config.ControllerConfig
+}
+
+func (m *modelBigtableAppProfile) client(ctx context.Context, parentProject string) (*gcp.InstanceAdminClient, error) {
+	gcpClient, err := gcp.NewInstanceAdminClient(ctx, parentProject)
```
jingyih, line 54:
```
error creating instance admin client: building BigtableAppProfile client: dialing: google: could not find default credentials.
```

Could you try get the `RESTClientOptions` from the `ControllerConfig`?

e.g.
https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go#L59-L70

pkg/test/resourcefixture/testdata/basic/bigtable/v1beta1/bigtableappprofile/_generated_object_bigtableappprofile.golden.yaml
```diff
@@ -32,5 +30,5 @@ status:
     reason: UpToDate
     status: "True"
     type: Ready
-  name: projects/${projectId}/instances/profiledep${uniqueId}/appProfiles/bigtableappprofile-${uniqueId}
```
jingyih, line 35: We cannot remove "name" for backward compatibility reasons. This field needs to be filled after CREATE and UPDATE.

pkg/test/resourcefixture/testdata/basic/bigtable/v1beta1/bigtableappprofile/_generated_object_bigtableappprofile.golden.yaml
```diff
@@ -22,7 +21,6 @@ spec:
   - cluster1-${uniqueId}
   - cluster2-${uniqueId}
   multiClusterRoutingUseAny: true
-  resourceID: bigtableappprofile-${uniqueId}
```
jingyih, line 25: @maqiuyujoyce For TF-based controller, we are writing "resourceID" back to Spec even though we set "state-into-spec: absent"?

maqiuyujoyce: Yes, we treat `spec.resourceID` differently in TF-based/DCL-based controllers, and we always write them back. E.g. https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/krmtotf/tftokrm.go#L229

pkg/test/resourcefixture/testdata/basic/bigtable/v1beta1/bigtableappprofile/_generated_object_bigtableappprofile.golden.yaml
```diff
@@ -31,4 +31,5 @@ status:
     status: "True"
     type: Ready
   externalRef: projects/${projectId}/instances/profiledep${uniqueId}/appProfiles/bigtableappprofile-${uniqueId}
+  name: projects/${projectId}/instances/profiledep${uniqueId}/appProfiles/bigtableappprofile-${uniqueId}
```
jingyih, line 34: To clarify, this file is generated by the test. I was not requesting a manual fix to re-add the field. Instead, we should update the controller to ensure the field is not missing after the test runs.

jingyih:
Can you run "make manifests" to fix the validation error in presubmit?

pkg/controller/direct/bigtable/appprofile_mappings.go
```diff
@@ -25,7 +25,10 @@ func BigtableAppProfileSpec_FromProto(mapCtx *direct.MapContext, in *pb.AppProfi
 		return nil
 	}
 	out := &krmv1beta1.BigtableAppProfileSpec{}
+
 	// MISSING: Name
+	out.ResourceID = direct.LazyPtr(in.GetName())
```
jingyih, line 30: I don't think `resourceID` is the fully qualified name of the resource, it's just the last part `bigtableappprofile-${uniqueId}`. Can we handle this field in the controller? Currently it does not appear to be working, since `resourceID` is still missing from the generated golden object file.

pkg/test/resourcefixture/testdata/basic/bigtable/v1beta1/bigtableappprofile/_http.log
```diff
@@ -309,6 +309,68 @@ OK
 
 GRPC /google.bigtable.admin.v2.BigtableInstanceAdmin/UpdateAppProfile
 
+{
+  "appProfile": {
```
jingyih, line 313: Not sure why there an extra "UpdateAppProfile" call with this change? The RPC body still contains the old field values. Is this from the mock or the real GCP?

djyau: Upon closer look, there are 2 updates here for some reason...the first update (line 310) doesn't update the description but does update multiclusterRouting and standardIsolation

Then the second update (line 371) updates the description along with both of those fields that were already updated in the first call.

jingyih:
For reference, I have also verified that following sample tests passed.

```
$ GOLDEN_REQUEST_CHECKS=1 GOLDEN_OBJECT_CHECKS=1 E2E_KUBE_TARGET=envtest RUN_E2E=1 E2E_GCP_TARGET=real go test -test.count=1 -timeout 3600s -v ./tests/e2e -run TestAllInSeries/samples/multicluster-bigtable-app-profile

$ GOLDEN_REQUEST_CHECKS=1 GOLDEN_OBJECT_CHECKS=1 E2E_KUBE_TARGET=envtest RUN_E2E=1 E2E_GCP_TARGET=real go test -test.count=1 -timeout 3600s -v ./tests/e2e -run TestAllInSeries/samples/single-cluster-bigtable-app-profile

$ GOLDEN_REQUEST_CHECKS=1 GOLDEN_OBJECT_CHECKS=1 E2E_KUBE_TARGET=envtest RUN_E2E=1 E2E_GCP_TARGET=real go test -test.count=1 -timeout 3600s -v ./tests/e2e -run TestAllInSeries/samples/priority-bigtable-app-profile
```

PR #4350, opened 2025-04-14
ci: Split presubmit fixtures test into more API groups

gemmahou:
### Change description

Split bigtable, certificatemanager, resourcemanager spanner, workstation

### Tests you have done

### DIFF
--- a/.github/workflows/ci-presubmit.yaml
+++ b/.github/workflows/ci-presubmit.yaml
@@ -172,7 +172,44 @@ jobs:
         with:
           name: artifacts-tests-e2e-fixtures-bigquery
           path: /tmp/artifacts/
+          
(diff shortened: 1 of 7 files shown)


yuwenma (commented):
/hold

Thanks, but I think we also need to exclude those tests from the main fixtures, otherwise these split-out tests will just run again and makes the presubmit worse.

gemmahou:
> /hold
> 
> Thanks, but I think we also need to exclude those tests from the main fixtures, otherwise these split-out tests will just run again and makes the presubmit worse.

Good catch, done!

jingyih (commented):
/lgtm
/appove

jingyih:
Oops, I made a typo.

/approve

jingyih (commented):
/lgtm

PR #4351, opened 2025-04-14
tests: small updates on kmsautokeyconfig fixtures test

gemmahou:
### Change description

Small updates on kmsautokeyconfig and kmskeyhandle fixtures test yaml and clean up

### Tests you have done


pkg/test/resourcefixture/testdata/basic/kms/v1alpha1/kmsautokeyconfig/_generated_object_kmsautokeyconfig.golden.yaml
gemmahou, line 1: This log does not associate to any test case(after Alpha to Beta migration), remove them. https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/ef3c0caa628cd0c641d51ce54feafefa3a50d5a7/pkg/test/resourcefixture/testdata/basic/kms/v1alpha1/kmsautokeyconfig

maqiuyujoyce:
I think we shouldn't change the field name because this is a breaking change and we should keep a released v1beta1 resource backwards compatible.

We can fix this one-off case if we really want, but I think the more important thing is that we need to have a more strict exit criteria for CRD during v1alpha1 to v1beta1 promotion.

gemmahou:
> I think we shouldn't change the field name because this is a breaking change and we should keep a released v1beta1 resource backwards compatible.
> 
> We can fix this one-off case if we really want, but I think the more important thing is that we need to have a more strict exit criteria for CRD during v1alpha1 to v1beta1 promotion.

/hold for team awareness and discussion.

pkg/test/resourcefixture/testdata/basic/kms/v1beta1/kmsautokeyconfig/update.yaml
```diff
@@ -22,4 +22,4 @@ spec:
   folderRef:
     external: folders/${TEST_FOLDER_ID}
   keyProject:
-    external: projects/diff-{uniqueId}
```
maqiuyujoyce, line 25: Does this test work when talking to a real API?

gemmahou: fixtures test passed locally other than those diffs in golden logs. Also confirmed that the GCP resource has been updated.

pkg/test/resourcefixture/testdata/basic/kms/v1beta1/kmsautokeyconfig/create.yaml
```diff
@@ -15,11 +15,10 @@
...
-  #annotations:
-  #  cnrm.cloud.google.com/folder-id: ${folderId}
   name: autokeyconfig-${uniqueId}
 spec:
   folderRef:
     external: folders/${TEST_FOLDER_ID}
   keyProject:
-    external: projects/${uniqueId}
+    name: project-${uniqueId}
+    kind: Project
```
maqiuyujoyce, line 24: Can users specify other kind? I'm just a bit confused why it is configured here.

gemmahou: No, the kind must be "Project"([ref](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/e653c624189a33308a1e9353ea10acf956d40fe2/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_kmsautokeyconfigs.kms.cnrm.cloud.google.com.yaml#L120-L121)) and it's a required field in the [CRD](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/e653c624189a33308a1e9353ea10acf956d40fe2/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_kmsautokeyconfigs.kms.cnrm.cloud.google.com.yaml#L101-L103)

maqiuyujoyce: Could you help me understand why? According to the field description, it's optional: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/e653c624189a33308a1e9353ea10acf956d40fe2/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_kmsautokeyconfigs.kms.cnrm.cloud.google.com.yaml#L120

This is not blocking.

PR #4352, opened 2025-04-14
Direct Controller for asset savedquery

barney-s:


pkg/controller/direct/asset/savedquery_controller.go
```diff
@@ -0,0 +1,390 @@
...
+	return true, nil
+}
+
+func (a *savedQueryAdapter) normalizeReferences(ctx context.Context) error {
+	// No references to resolve in SavedQuery spec currently.
+	return nil
+}
+
+// AssetSavedQueryStatus_FromProto converts the AssetSavedQueryStatus field from its Protobuf representation.
+func AssetSavedQueryStatus_FromProto(mapCtx *direct.MapContext, in *pb.SavedQuery) *krm.AssetSavedQueryStatus {
```
jingyih, line 152: Can we move this to the mapper file? Or, can we just use the mapping function for ObservedState?

jingyih (commented):
/approve

pkg/controller/direct/asset/savedquery_controller.go
barney-s: Moved it to mapper file.

pkg/controller/direct/asset/savedquery_controller.go
```diff
@@ -0,0 +1,375 @@
...
+			// This is unexpected if CreateSavedQuery returned AlreadyExists
+			return fmt.Errorf("asset saved query %q not found even though CreateSavedQuery returned AlreadyExists", a.id.String())
+		}
+	} else {
+		// Other error during creation
+		return fmt.Errorf("creating asset saved query %s: %w", a.id.String(), err)
+	}
+
+	// Update status
+	status := AssetSavedQueryStatus_FromProto(mapCtx, a.actual)
```
jingyih, line 203: We probably only need to use `AssetSavedQueryObservedState_FromProto` and set the `status.ExternalRef`, similar to how other controllers handle it. But this is not a blocker.

jingyih (commented):
/lgtm
/approve

jingyih (commented):
/lgtm

PR #4354, opened 2025-04-14
refactor: disregard updateTime unset

acpana:
This is a small change that technically makes sure we don't send an extra Update call on acquisition for LoggingLogMetric.

Testing
Mostly manually tested but we could write a scenario based test for this too.


ldanielmadariaga:

PR #4355, opened 2025-04-14
update discoveryEngineDatastore from v1 to v1alpha

yuwenma:
b/398160349


apis/discoveryengine/v1alpha1/datastore_types.go
```diff
@@ -15,65 +15,178 @@
...
+	//  the default parsing config will be applied to all file types for Document
+	//  parsing.
+	// +kcc:proto:field=google.cloud.discoveryengine.v1alpha.DocumentProcessingConfig.default_parsing_config
+	DefaultParsingConfig *DocumentProcessingConfig_ParsingConfig `json:"defaultParsingConfig,omitempty"`
+
+	// TODO: unsupported map type with key string and value message
+}
+
+// +kcc:proto=google.cloud.discoveryengine.v1alpha.LanguageInfo
+type LanguageInfo struct {
```
justinsb, line 155: Why can't this one be auto generated?

apis/discoveryengine/v1alpha1/datastore_types.go
```diff
@@ -15,65 +15,178 @@
...
+	// +kcc:proto:field=google.cloud.discoveryengine.v1alpha.DataStore.create_time
+	CreateTime *string `json:"createTime,omitempty"`
+
+	// Language info for DataStore.
+	// +kcc:proto:field=google.cloud.discoveryengine.v1alpha.DataStore.language_info
+	LanguageInfo *LanguageInfoObservedState `json:"languageInfo,omitempty"`
+}
+
+// +kcc:proto=google.cloud.discoveryengine.v1alpha.WorkspaceConfig
+type WorkspaceConfig struct {
```
justinsb, line 180: This one looks like it should have been auto-generated also

apis/discoveryengine/v1alpha1/engine_types.go
```diff
@@ -66,10 +66,6 @@ type DiscoveryEngineEngineSpec struct {
 	// solution_type is SOLUTION_TYPE_SEARCH.
 	SearchEngineConfig *Engine_SearchEngineConfig `json:"searchEngineConfig,omitempty"`
 
-	// Optional. Whether to disable analytics for searches performed on this
-	//  engine.
-	DisableAnalytics *bool `json:"disableAnalytics,omitempty"`
```
justinsb, line 71: This looks unrelated.  Should this be in this PR?

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_discoveryenginedatastores.discoveryengine.cnrm.cloud.google.com.yaml
```diff
@@ -126,37 +243,42 @@ spec:
...
                 description: Config to store data store type configuration for workspace
-                  data. This must be set when [DataStore.content_config][google.cloud.discoveryengine.v1.DataStore.content_config]
-                  is set as [DataStore.ContentConfig.GOOGLE_WORKSPACE][google.cloud.discoveryengine.v1.DataStore.ContentConfig.GOOGLE_WORKSPACE].
+                  data. This must be set when [DataStore.content_config][google.cloud.discoveryengine.v1alpha.DataStore.content_config]
+                  is set as [DataStore.ContentConfig.GOOGLE_WORKSPACE][google.cloud.discoveryengine.v1alpha.DataStore.ContentConfig.GOOGLE_WORKSPACE].
                 properties:
                   dasherCustomerID:
                     description: Obfuscated Dasher customer ID.
                     type: string
-                  superAdminEmailAddress:
```
justinsb, line 146: Ouch.  Does this mean that v1alpha1 is not a super-set of v1?

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_discoveryengineengines.discoveryengine.cnrm.cloud.google.com.yaml
```diff
@@ -71,9 +71,9 @@ spec:
 
                        Note that these configurations are one-time consumed by
                        and passed to Dialogflow service. It means they cannot be retrieved using
-                       [EngineService.GetEngine][google.cloud.discoveryengine.v1.EngineService.GetEngine]
+                       [EngineService.GetEngine][google.cloud.discoveryengine.v1alpha.EngineService.GetEngine]
```
justinsb, line 74: Ah - I don't think we need to use v1alpha here (?)

dev/tools/controllerbuilder/generate.sh
```diff
@@ -25,8 +25,10 @@ cd ${REPO_ROOT}/dev/tools/controllerbuilder
 
 
 # DiscoveryEngine
+# TODO: We use service v1alpha instead of v1 because v1 has key feature missing.
```
justinsb, line 28: I would list at least one or two of the fields we need

dev/tools/controllerbuilder/generate.sh
```diff
@@ -25,8 +25,10 @@ cd ${REPO_ROOT}/dev/tools/controllerbuilder
...
 
 # DiscoveryEngine
+# TODO: We use service v1alpha instead of v1 because v1 has key feature missing. 
+
 go run . generate-types \
-    --service google.cloud.discoveryengine.v1 \
+    --service google.cloud.discoveryengine.v1alpha \
     --api-version discoveryengine.cnrm.cloud.google.com/v1alpha1 \
     --resource DiscoveryEngineDataStore:DataStore \
     --resource DiscoveryEngineDataStoreTargetSite:TargetSite \
```
justinsb, line 34: I think in one of my PRs I added the ability to make the proto type fully qualified (i.e. `--resource DiscoveryEngineDataStore:google.cloud.discoveryengine.v1alpha.DataStore`), precisely to enable mixing versions.

pkg/controller/direct/discoveryengine/datastore_controller.go
```diff
@@ -181,34 +186,37 @@ func (a *dataStoreAdapter) Create(ctx context.Context, createOp *directbase.Crea
...
-	desired.Name = a.id.String()
-
-	// TODO(user): Update the field if applicable.
-	updateMask := &fieldmaskpb.FieldMask{}
-	if !reflect.DeepEqual(a.desired.DisplayName, a.actual.DisplayName) {
-		updateMask.Paths = append(updateMask.Paths, "display_name")
+	a.desired.Name = a.actual.Name
+	paths := make(sets.Set[string])
+	var err error
+	paths, err = common.CompareProtoMessage(a.desired, a.actual, common.BasicDiff)
```
justinsb, line 205: Do you need to mask out so that a.actual is just the spec fields?  (One trick to do so is to convert it to spec and back!)

pkg/controller/direct/discoveryengine/mapper.go
```diff
@@ -88,11 +87,165 @@ func DiscoveryEngineEngineSpec_ToProto(mapCtx *direct.MapContext, in *krm.Discov
...
+	// MISSING: CreateTime
+	out.LanguageInfo = LanguageInfo_FromProto(mapCtx, in.GetLanguageInfo())
+	out.AclEnabled = direct.LazyPtr(in.GetAclEnabled())
+	out.IdpConfig = IDPConfig_FromProto(mapCtx, in.GetIdpConfig())
+	out.WorkspaceConfig = WorkspaceConfig_FromProto(mapCtx, in.GetWorkspaceConfig())
+	// MISSING: DocumentProcessingConfig
+	return out
+}
+
+func IDPConfig_FromProto(mapCtx *direct.MapContext, in *pb.IdpConfig) *krmv1alpha1.IdpConfig {
```
justinsb, line 118: This can't be auto-generated?

pkg/controller/direct/discoveryengine/mapper.go
```diff
@@ -88,11 +87,165 @@ func DiscoveryEngineEngineSpec_ToProto(mapCtx *direct.MapContext, in *krm.Discov
...
+	if in == nil {
+		return nil
+	}
+	out := &krmv1alpha1.IdpConfig{}
+	out.IdpType = direct.Enum_FromProto(mapCtx, in.GetIdpType())
+	out.ExternalIdpConfig = ExternalIdpConfig_FromProto(mapCtx, in.GetExternalIdpConfig())
+	return out
+}
+
+func ExternalIdpConfig_FromProto(mapCtx *direct.MapContext, in *pb.IdpConfig_ExternalIdpConfig) *krmv1alpha1.IdpConfig_ExternalIdpConfig {
```
justinsb, line 128: This one also looks like what I would expect to be auto-generated

justinsb:
> b/398160349

As an open source project, I think we should say something like "Ensure we can support the aclEnabled field"

apis/discoveryengine/v1alpha1/datastore_types.go
yuwenma: The proto comments tell nothing about this value. I pulled it out from the proto source code.  

Oh, do you mean having the type-generator to generate this enum validator? That could be a good idea!

apis/discoveryengine/v1alpha1/engine_types.go
yuwenma: Yes, because we change the proto from v1 to v1alpha.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_discoveryenginedatastores.discoveryengine.cnrm.cloud.google.com.yaml
yuwenma: Right, v1 and v1alpha are different, even the v1alpha client-go and the actual v1alpha server are different (server has more fields).

pkg/controller/direct/discoveryengine/datastore_controller.go
yuwenma: We don't need to. CompareProtoMessage skips OUTOUT-ONLY fields. So the proto level comparison is spec to spec.

apis/discoveryengine/v1alpha1/datastore_reference.go
```diff
@@ -99,20 +100,6 @@ func NewDiscoveryEngineDataStoreIDFromObject(ctx context.Context, reader client.
 	if err != nil {
 		return nil, err
 	}
-	projectID := projectRef.ProjectID
```
justinsb, line 102: Why don't we need these?

apis/discoveryengine/v1alpha1/datastore_types.go
justinsb: Ah OK.  We should think about how to "3 way merge" improvements to the upstream proto.  I think that will involve AI.  It might need us to start storing some indication of the last proto version we looked at.

apis/discoveryengine/v1alpha1/engine_types.go
justinsb: On DataStore, but you are allowed to mix and match.  It also seems weird that the alpha API would _not_ have a field that is in the v1 API, we should probably file an upstream bug (but not block on it).

It begs the deeper question of why we are "downgrading" to use an alpha API.  I can see using a beta API, but upstream alpha APIs are painful.  We should avoid just linking to internal bugs as the justification for a PR - we can summarize the intent (and replace e.g. customer names with things like "it was reported")

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_discoveryenginedatastores.discoveryengine.cnrm.cloud.google.com.yaml
justinsb: So that's a different problem.  We need service teams to publish up-to-date clients, but it's easy to forget.  But I think the rule is that fields that are in v1 must be in beta, fields that are in beta must be in alpha.  It's no doubt possible to break this without it being a violation of the AIP rules, but I'm not smart enough to understand aip.dev :-)

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_discoveryengineengines.discoveryengine.cnrm.cloud.google.com.yaml
justinsb: I guess this one is harmoless though

pkg/controller/direct/discoveryengine/datastore_controller.go
justinsb: That is very magical - possibly too magical!

ldanielmadariaga:

PR #4365, opened 2025-04-15
FIX: trying github build failure fix

barney-s:
From:
https://github.com/actions/runner/issues/1807#issuecomment-1537175089

This is for the build image step running out of disk

### DIFF
--- a/.github/workflows/presubmit.yaml
+++ b/.github/workflows/presubmit.yaml
@@ -130,6 +130,17 @@ jobs:
       - uses: actions/setup-go@v5
         with:
           go-version-file: 'go.mod'
+      - name: Free Up GitHub Actions Ubuntu Runner Disk Space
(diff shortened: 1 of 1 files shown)


barney-s (commented):
pin to latest released version

justinsb:
I'm not sure we want to take a dependency on a third party github action for this, but I like the approach

justinsb:
Maybe we just add `sudo rm -rf /usr/local/lib/android || true`

https://github.com/jlumbroso/free-disk-space/blob/54081f138730dfa15788a46383842cd2f914a1be/action.yml#L136C9-L137C1

We can still credit the action in the workflows.yaml

jingyih:
Discussed in the chat. We could go with https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4371

PR #4366, opened 2025-04-15
feat: Extend KMSKeyRef to support manual and auto key

gemmahou:
### Change description

Fixes b/409337150 and b/409340085

### Tests you have done


apis/refs/v1beta1/kmsrefs.go
```diff
@@ -26,53 +26,29 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-type KMSCryptoKeyRef struct {
-	// A reference to an externally managed KMSCryptoKey.
-	// Should be in the format `projects/[kms_project_id]/locations/[region]/keyRings/[key_ring_id]/cryptoKeys/[key]`.
-	External string `json:"external,omitempty"`
+// NOTE: If references a KMSKey, use `KMSKeyRef_OneOf` instead!
+// todo: use unexported variable kmsCryptoKeyRef to avoid referencing to the KMS crypto key.
```
gemmahou, line 30: Once we move `refs/kmsrefs.go` to `apis/kms/v1beta1`, we can make this type unexported, i.e.`kmsCryptoKeyRef` so that other resources can only reference to `KMSKeyRef_OneOf` not `KMSCryptoKeyRef`

yuwenma:
Please split this PR into smaller ones (1. the code to KMS key and 2. change to referenced resources). For both case, suggest add test coverage to verify it works.

gemmahou:
> Please split this PR into smaller ones (1. the code to KMS key and 2. change to referenced resources). For both case, suggest add test coverage to verify it works.

I updated this PR to focus on the changes related to `KMSKeyRef_OneOf`. Will have a follow up PR to switch `KMSCryptoKeyRef` to `KMSKeyRef_OneOf` for all referenced resources.

I think existing fixtures tests cover this change already, I expect no diffs in the golden logs.

tests/apichecks/crds_test.go
```diff
@@ -475,7 +475,8 @@ func TestCRDFieldPresenceInUnstructured(t *testing.T) {
 				}
 
 				// Check for "Ref" fields
-				if strings.HasSuffix(fieldPath, "Ref") {
+				// autoKeyRef is an optional reference, see (pending link to KMSKeyRef_OneOf struct)
+				if strings.HasSuffix(fieldPath, "Ref") && !strings.HasSuffix(fieldPath, "autoKeyRef") {
```
maqiuyujoyce: I think previously we didn't require external because it's a bit difficult to test external field, it doesn't mean we don't hope to cover external fields in testing.
For autoKeyRef, I think it'll be possible for us to test it, and I am leaning towards making it explicit in the missingfields.txt.

apis/kms/v1beta1/kmskey_reference.go
```diff
@@ -0,0 +1,82 @@
...
+	// Default KMS crypto key. This is for API backward compatibility and cannot be changed.
+	*refs.KMSCryptoKeyRef `json:",inline"`
+
+	// A reference to the Autokey `KMSKeyHandle`, which auto generates a crypto key.
+	AutoKeyRef *kmsKeyHandleRef `json:"autoKeyRef,omitempty"`
+
+	// A reference to an externally managed KMSCryptoKey or KMSKeyHandle(AutoKey).
+	// Should be in the format `projects/{{kms_project_id}}/locations/{{region}}/keyRings/{{key_ring_id}}/cryptoKeys/{{key}}`.
+	// For AutoKey, replace {{key_ring_id}} to `autokey`, i.e. `projects/{{kms_project_id}}/locations/{{region}}/keyRings/autokey/cryptoKeys/{{key}}`.
+	External string `json:"external,omitempty"`
```
maqiuyujoyce, line 39: Hmmm, isn't external already covered in refs.KMSCryptoKeyRef? With this struct, I think kubebuilder will get confused about how to generate the CRD?

I wondered if it'll be easier if you split the refactoring into multiple PRs, for example, the first would be changing the use of KMSKeyRef to KMSKeyRef_OneOf: in the KMSKeyRef, there is only `name` and `namespace`, and in the KMSKeyRef_OneOf, there is `KMSKeyRef` and `external`. After that, you can make incremental changes in the new struct.

tests/apichecks/crds_test.go
gemmahou: The new KMSKeyRef looks like
```
kmsKeyRef:
  name(cryptoKey name)
  namespace(cryptoKey namespace)
  (OR)
  external(cryptoKey or autoKey external)
  (OR)
  autoKeyRef:
    name(autoKey name)
    namespace(autoKey namespece) 
```
`external` value(either cryptoKey or autoKey external) should be specified in `kmsKeyRef.external`; `autoKeyRef` does not have this field. I think that's more straight forward for users.

apis/kms/v1beta1/kmskey_reference.go
gemmahou: Good point!

I'd like to get some suggestions here before moving forward: shall we have ONE `external` field to cover both CryptoKey and AutoKey in `KMSKeyRef_OneOf`, or shall we have TWO `external` field in `KMSCryptoKeyRef ` and `KMSKeyHandleRef`, to handle CryptoKey and AutoKey external values separately? 

I had the former in this PR because I think that's more straight-forward for customer.

maqiuyujoyce: I think one `external` (like what we have now) should be good. Reasons are:
1. Multiple `external` fields may reduce readability of the CRD, increase customer confusion, and complicate the validation logic.
2. Eventually the value of the reference is resolved into a plain text string, i.e. the value used in the `external` field. Having `external` under `autoKey` doesn't benefit the implementation of this feature.

If we receive feedback about it and agree it's a better UX, we can always add the `external` field under `autoKey` as it's an incremental change.

tests/apichecks/crds_test.go
maqiuyujoyce: Ah okay, I misunderstood the intention here. Thanks for the clarification!

In that case, you'll want to add additional handling under this section as well:

```
				// Any XYZRef field was already handled and handling the children will just double count
				if strings.Contains(fieldPath, "Ref") {
					return
				}
```

Because `autoKeyRef` will never be checked.

gemmahou: Done.

apis/kms/v1beta1/kmskey_reference.go
gemmahou: Thanks for the quick feedback!
To minimize risk, I will not change ALL existing KMSCryptoKeyRef references to KMSKeyRef_OneOf at once. Instead, I'll gradually migrate to KMSKeyRef_OneOf resource-by-resource. So here're my steps:
1. Add a `new KMSCryptoKeyRef without external field` in `apis/kms/v1beta1/cryptokey_reference.go`, and let KMSKeyRef_OneOf refer to this `new KMSCryptoKeyRef` - see [commit](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4366/commits/9614779eeac5ef738be8cf7c5b3baeb9245b3364)
2. In a follow up PR, finalize `NormalizedExternal` function `apis/kms/v1beta1/cryptokey_reference.go`, similar to what I've done in #3836, but keep both KMSCryptoKeyRef(with and without external) for now.
3. In BigQuery resource which has reference to kms(for example), change to use `KMSKeyRef_OneOf`. While other resources will still use the `old KMSCryptoKeyRef with external field`.
4. Gradually change to use `KMSKeyRef_OneOf` in other resources.
5. Once all kms references are migrated to `KMSKeyRef_OneOf`, safely remove the `old KMSCryptoKeyRef`.

WDYT?

apis/kms/v1beta1/cryptokey_reference.go
```diff
@@ -0,0 +1,64 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ refsv1beta1.ExternalNormalizer = &KMSCryptoKeyRef{}
+var KMSCryptoKeyGVK = SchemeGroupVersion.WithKind("KMSCryptoKey")
+
+// KMSCryptoKeyRef defines the resource reference to KMSCryptoKey, which "External" field
```
maqiuyujoyce: Could you update the comment?

apis/kms/v1beta1/cryptokey_reference.go
```diff
@@ -0,0 +1,64 @@
...
+// holds the GCP identifier for the KRM object.
+type KMSCryptoKeyRef struct {
+	// The `name` of a `KMSCryptoKey` resource.
+	Name string `json:"name,omitempty"`
+	// The `namespace` of a `KMSCryptoKey` resource.
+	Namespace string `json:"namespace,omitempty"`
+}
+
+// NormalizedExternal provision the "External" value for other resource that depends on KMSCryptoKeyRef.
+// If the "External" is given in the other resource's spec.KMSCryptoKeyRef, the given value will be used.
```
maqiuyujoyce: ditto

apis/kms/v1beta1/cryptokey_reference.go
```diff
@@ -0,0 +1,64 @@
...
+	key := types.NamespacedName{Name: r.Name, Namespace: r.Namespace}
+	u := &unstructured.Unstructured{}
+	u.SetGroupVersionKind(KMSCryptoKeyGVK)
+	if err := reader.Get(ctx, key, u); err != nil {
+		if apierrors.IsNotFound(err) {
+			return "", k8s.NewReferenceNotFoundError(u.GroupVersionKind(), key)
+		}
+		return "", fmt.Errorf("reading referenced %s %s: %w", KMSCryptoKeyGVK, key, err)
+	}
+	// todo: resolve KMSCryptoKey from its name
```
maqiuyujoyce, line 58: Is this todo still valid?

tests/apichecks/crds_test.go
```diff
@@ -510,7 +511,8 @@ func TestCRDFieldPresenceInUnstructured(t *testing.T) {
 				}
 
 				// Any XYZRef field was already handled and handling the children will just double count
-				if strings.Contains(fieldPath, "Ref") {
+				// autoKeyRef is an optional reference, see (pending link to KMSKeyRef_OneOf struct)
+				if strings.Contains(fieldPath, "Ref") && !strings.Contains(fieldPath, "autoKeyRef") {
```
maqiuyujoyce: This won't work here as `autoKeyRef` is a subfield of a XYZRef field, and any subfield of XYZRef field will never be checked.

You'll want to:
1. Identify if the XYZRef field has the `autoKeyRef` subfield.
2. If yes,  continue checking only the `autoKeyRef` subfield of XYZRef.

Also I suggest you move the change out of the PR and add the logic once you have a generated `autoKeyRef` field in the CRD.

apis/kms/v1beta1/cryptokey_reference.go
```diff
@@ -0,0 +1,64 @@
...
+	// The `namespace` of a `KMSCryptoKey` resource.
+	Namespace string `json:"namespace,omitempty"`
+}
+
+// NormalizedExternal provision the "External" value for other resource that depends on KMSCryptoKeyRef.
+// If the "External" is given in the other resource's spec.KMSCryptoKeyRef, the given value will be used.
+// Otherwise, the "Name" and "Namespace" will be used to query the actual KMSCryptoKeyRef object from the cluster.
+func (r *KMSCryptoKeyRef) NormalizedExternal(ctx context.Context, reader client.Reader, otherNamespace string) (string, error) {
+	if r.Name == "" {
+		return "", fmt.Errorf("use KMS cryptokey requires referring to the Config Connector `KMSCryptoKey` object. please provide the `name` of your  `KMSCryptoKey`.")
```
maqiuyujoyce: The error message sounds a bit confusing here. Using KMS Crypto Key doesn't require referring to the KCC KMSCryptoKey object, they can use `external` field or other fields as well. It's that "referencing the Config Connector KMSCryptoKey requires the configuration of the `name` field.

Also, it should be checked at the higher level that one of `name`, `external`, and `autoKeyRef.name` must be set. Ideally this normalization function won't be called outside of the package, and only the higher level normalization is exported and will return a more user friendly error message.

I'd just return `name field must be set` as the error message. The caller of this function should add more information in the error message.

apis/kms/v1beta1/keyhandle_reference.go
```diff
@@ -26,15 +26,10 @@ import (
...
-// KMSKeyHandleRef defines the resource reference to KMSKeyHandle, which "External" field
-// holds the GCP identifier for the KRM object.
-type KMSKeyHandleRef struct {
-	// A reference to an externally managed KMSKeyHandle resource.
-	// Should be in the format "projects/{{projectID}}/locations/{{location}}/keyHandles/{{keyhandleID}}".
-	External string `json:"external,omitempty"`
+var _ refsv1beta1.ExternalNormalizer = &kmsKeyHandleRef{}
 
+// kmsKeyHandleRef defines the resource reference to KMSKeyHandle
+type kmsKeyHandleRef struct {
```
maqiuyujoyce, line 29: You'll probably also want to do it for v1alpha1 version.

apis/kms/v1beta1/keyhandle_reference.go
```diff
@@ -43,18 +38,10 @@ type KMSKeyHandleRef struct {
...
-	// From given External
-	if r.External != "" {
-		if _, _, err := ParseKMSKeyHandleExternal(r.External); err != nil {
-			return "", err
-		}
-		return r.External, nil
+// The "Name" and "Namespace" will be used to query the actual KMSKeyHandle object from the cluster.
+func (r *kmsKeyHandleRef) NormalizedExternal(ctx context.Context, reader client.Reader, otherNamespace string) (string, error) {
+	if r.Name == "" {
+		return "", fmt.Errorf("use KMS autokey requires referring to the Config Connector `KMSKeyHanle` object. please provide the `name` of your  `KMSKeyHanle`.")
```
maqiuyujoyce: Similarly, the error message is a bit confusing.

apis/kms/v1beta1/keyhandle_reference.go
```diff
@@ -70,14 +57,15 @@ func (r *KMSKeyHandleRef) NormalizedExternal(ctx context.Context, reader client.
 		}
 		return "", fmt.Errorf("reading referenced %s %s: %w", KMSKeyHandleGVK, key, err)
 	}
-	// Get external from status.externalRef. This is the most trustworthy place.
-	actualExternalRef, _, err := unstructured.NestedString(u.Object, "status", "externalRef")
+	// Use status.observedState.kmsKey instead of status.externalRef as the external resourceID of autoKey
+	// status.externalRef: projects/${projectId}/locations/us-central1/keyHandles/1a1a1a-222b-3cc3-d444-e555ee555555
+	// status.observedState.kmsKey: projects/${key_project}/locations/us-central1/keyRings/autokey/cryptoKeys/${projectNumber}-compute-disk-${generated-id}
```
maqiuyujoyce, line 59: Just curious: What's the relationship between key handles and auto keys?

apis/kms/v1beta1/kmskey_reference.go
maqiuyujoyce: As long as we have thorough integration tests, yes, it looks good to me. I'd expect a test case for external field for a picked resource to ensure external field works the same before and after. And I'd expect one test to cover autoKeyRef field for each resource using KMSKeyRef_OneOf.

apis/refs/v1beta1/kmsrefs.go
```diff
@@ -26,24 +26,23 @@ import (
...
 }
 
-type KMSCryptoKey struct {
-	Ref        *KMSCryptoKeyRef
-	ResourceID string
-}
-
 // ResolveKMSCryptoKeyRef will resolve a KMSCryptoKeyRef to a KMSCryptoKey.
-func ResolveKMSCryptoKeyRef(ctx context.Context, reader client.Reader, src client.Object, ref *KMSCryptoKeyRef) (*KMSCryptoKeyRef, error) {
+func ResolveKMSCryptoKeyRef(ctx context.Context, reader client.Reader, otherNamespace string, ref *KMSCryptoKeyRef) (*KMSCryptoKeyRef, error) {
```
maqiuyujoyce: Can we move the refactoring out of this PR and keep it focusing on supporting the new UI?

apis/kms/v1beta1/kmskey_reference.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func (r *KMSKeyRef_OneOf) NormalizedExternal(ctx context.Context, reader client.Reader, otherNamespace string) (string, error) {
+	if r.External != "" {
```
maqiuyujoyce, line 47: I'd hope the validation to be more readable here. Let's focus on the validation first, and then normalize the value. Here is an example: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/privilegedaccessmanager/entitlement_controller.go#L161

apis/kms/v1beta1/cryptokey_reference.go
gemmahou: Yes, I will finish the resolver function in another PR as mentioned in step2 above. It's fine to leave a todo here as this new Ref is not used by other resources yet.

apis/kms/v1beta1/keyhandle_reference.go
gemmahou: While I do not understand why we still keep v1alpha1 since it has already been migrated to v1beta1. This older version is not used anywhere and is out of sync with v1beta1 (for example, there's no keyhandle_identity.go in v1alpha1).

Updating v1alpha1 here would introduce unrelated changes. If needed, I could address them in a separate PR.

apis/kms/v1beta1/keyhandle_reference.go
gemmahou: They are the same thing. The resource name is KMSKeyHandle, for easier communication we simply call it autokey.

apis/kms/v1beta1/kmskey_reference.go
gemmahou: SG, this will be done in step 3 and 4

apis/kms/v1beta1/keyhandle_reference.go
maqiuyujoyce: I'm guessing this is for any v1alpha1 resource to use KMS as the reference? Could you check with the team so that we can make a decision either removing it or keeping it in sync with v1beta1?

gemmahou: Posted question to team's chat, I will followup in another PR.

tests/apichecks/crds_test.go
gemmahou: Discussed offline, I will followup the crd test in another PR.

Previously, this pull request included changes to KMSKeyRef_OneOf in other resources which impacted the CRD and required special test handling. These changes have been moved to pull request #4599, so we don't have to worry about crd test in this PR.

apis/kms/v1beta1/cryptokey_reference.go
```diff
@@ -0,0 +1,62 @@
...
+type KMSCryptoKeyRef struct {
+	// The `name` of a `KMSCryptoKey` resource.
+	Name string `json:"name,omitempty"`
+	// The `namespace` of a `KMSCryptoKey` resource.
+	Namespace string `json:"namespace,omitempty"`
+}
+
+// NormalizedExternal provision the "External" value for other resource that depends on KMSCryptoKeyRef.
+// The "Name" and "Namespace" will be used to query the actual KMSCryptoKeyRef object from the cluster.
+func (r *KMSCryptoKeyRef) NormalizedExternal(ctx context.Context, reader client.Reader, otherNamespace string) (string, error) {
```
maqiuyujoyce: Should this function be unexported?

apis/kms/v1beta1/cryptokey_reference.go
```diff
@@ -0,0 +1,62 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ refsv1beta1.ExternalNormalizer = &KMSCryptoKeyRef{}
+var KMSCryptoKeyGVK = SchemeGroupVersion.WithKind("KMSCryptoKey")
+
+// KMSCryptoKeyRef defines the resource reference to KMSCryptoKey.
+type KMSCryptoKeyRef struct {
```
maqiuyujoyce: Should this type be unexported?

apis/kms/v1beta1/keyhandle_reference.go
```diff
@@ -43,18 +38,10 @@ type KMSKeyHandleRef struct {
...
-		return "", fmt.Errorf("cannot specify both name and external on %s reference", KMSKeyHandleGVK.Kind)
-	}
-	// From given External
-	if r.External != "" {
-		if _, _, err := ParseKMSKeyHandleExternal(r.External); err != nil {
-			return "", err
-		}
-		return r.External, nil
+// The "Name" and "Namespace" will be used to query the actual KMSKeyHandle object from the cluster.
+func (r *kmsKeyHandleRef) NormalizedExternal(ctx context.Context, reader client.Reader, otherNamespace string) (string, error) {
```
maqiuyujoyce: Looks like this can be unexported?

apis/kms/v1beta1/kmskey_reference.go
```diff
@@ -0,0 +1,94 @@
...
+	// Default KMS crypto key. This is for API backward compatibility and cannot be changed.
+	*KMSCryptoKeyRef `json:",inline"`
+
+	// A reference to the Autokey `KMSKeyHandle`, which auto generates a crypto key.
+	AutoKeyRef *kmsKeyHandleRef `json:"autoKeyRef,omitempty"`
+
+	// A reference to an externally managed KMSCryptoKey or KMSKeyHandle(AutoKey).
+	// Should be in the format `projects/{{kms_project_id}}/locations/{{region}}/keyRings/{{key_ring_id}}/cryptoKeys/{{key}}`.
+	// For AutoKey, replace {{key_ring_id}} to `autokey`, i.e. `projects/{{kms_project_id}}/locations/{{region}}/keyRings/autokey/cryptoKeys/{{key}}`.
+	External string `json:"external,omitempty"`
```
maqiuyujoyce, line 39: Should this be a pointer to string to distinguish between unset and set to empty string?

apis/kms/v1beta1/kmskey_reference.go
```diff
@@ -0,0 +1,94 @@
...
+		numOfNonNil++
+	}
+	if r.KMSCryptoKeyRef != nil && r.KMSCryptoKeyRef.Name != "" {
+		numOfNonNil++
+	}
+	if r.AutoKeyRef != nil && r.AutoKeyRef.Name != "" {
+		numOfNonNil++
+	}
+	if numOfNonNil == 0 {
+		return fmt.Errorf("a key reference must be provided: specify one of `.external`, `.name`, or `.autoKeyRef.name`")
```
maqiuyujoyce: Nits:
Suggest we reorder the three options to `".name", ".autoKeyRef.name", or ".external"` to indicate we recommend the reference over external.

Suggest we don't use backtick in error messages as backticks doesn't seem to be helpful in non-markdown context.

apis/kms/v1beta1/kmskey_reference.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	if numOfNonNil > 1 {
+		return fmt.Errorf("exactly one of `.external`, `.name`, or `.autoKeyRef.name` must be specified, but %d were found", numOfNonNil)
```
maqiuyujoyce: ditto

tests/apichecks/crds_test.go
```diff
@@ -481,6 +481,7 @@ func TestCRDFieldPresenceInUnstructured(t *testing.T) {
 				}
 
 				// Check for "Ref" fields
+				// autoKeyRef is an optional reference, see (pending link to KMSKeyRef_OneOf struct)
```
maqiuyujoyce: Nit: Clean up the comment.

apis/kms/v1beta1/kmskey_reference.go
```diff
@@ -0,0 +1,94 @@
...
+		if err != nil {
+			return "", err
+		}
+		r.External = autoKey
+	}
+	return r.External, nil
+}
+
+// validateOneOf checks that exactly one of the key reference fields is set
+func (r *KMSKeyRef_OneOf) validateOneOf() error {
```
maqiuyujoyce, line 76: Thank you for the validation function.  Though could you make sure edge cases like below are also failing the validation? Right now it seems to be allowed.
```
kmsKeyRef:
     name: test-kms-key
     external: ""
     autoKey:
         namespace: test-ns
```

apis/kms/v1beta1/kmskey_reference.go
gemmahou: I would agree but seems we use `string` in all places: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/a6eb443d94a9627a89b6ea362fdead176138187d/dev/tools/controllerbuilder/template/apis/refs.go#L38

apis/kms/v1beta1/kmskey_reference.go
maqiuyujoyce: I'd suggest we add some unit tests for this function.

gemmahou: Thanks! I updated the code and added unit test. See https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4366/commits/d4af52babaa9d37cf70cb92852a76e49301e77b8

I think the message `specify one of '.name', '.keyHandleRef.name' or '.external'` is still valid even we not only check for .name but also .namespace. If name is not set, the resolver function would error out later.

maqiuyujoyce (commented):
/lgtm
/approve

apis/kms/v1beta1/kmskey_reference.go
maqiuyujoyce: I don't think that's the optimal choice of type. But I understand for this resource, we probably should keep using string for backwards compatibility purpose for now. I'd suggest for new references, we use *string instead.

apis/kms/v1beta1/kmskey_reference.go
```diff
@@ -0,0 +1,82 @@
...
+	"strings"
+
+	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ refs.ExternalNormalizer = &KMSKeyRef_OneOf{}
+
+// A reference to the KMSCryptoKey(manual management), or the AutoKey(automated management)
+type KMSKeyRef_OneOf struct {
```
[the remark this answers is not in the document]
yuwenma: My original proposal is 
```
kmsKeyRef: // whose go struct is the typed KMSCryptoKeyRef 
   name:
   namespace:
   external: 
kmsKeyHandlerRef: // whose go struct is the typed KMSKeyHandlerRef 
   name:
   namespace:
   external:
```
where both KMSCryptoKeyRef and KMSKeyHandlerRef are under a parent go struct (with inline json tag) to make the code backward compatible
```
type KMSKey_oneOf struct {
    KMSCryptoKeyRef *refs.KMSCryptoKeyRef  `json:"kmsCryptKeyRef,omitempty"`
    KMSKeyHandlerRef *refs.KMSKeyHandlerRef  `json:"kmsKeyHandlerRef,omitempty"`
}

type ApigeeInstanceSpec struct {
    kmsKeyRef KMSKey_oneOf  `json:",inline"`
}
```


Can I ask why we make this API change eventually? The autoKeyRef is not a child object of KmsKey. In fact, this should be `KMSKeyHandlerRef` (the actual auto key object) rather than `KMSAutoKeyRef`.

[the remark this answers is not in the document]
gemmahou: Here's how the go struct looks like(please check the [code/file](https://github.com/gemmahou/k8s-config-connector/blob/a5c6d5c2245ebce9929fbd03a9424b21a8ab3c44/apis/kms/v1beta1/kmskey_reference.go#L28-L40) directly):
```
type KMSKeyRef_OneOf struct {
	*kmsCryptoKeyRef `json:",inline"`
	KMSKeyHandleRef *kmsKeyHandleRef `json:"keyHandleRef,omitempty"`
        
    // A reference to an externally managed KMSCryptoKey or KMSKeyHandle(AutoKey).
	// Should be in the format `projects/{{kms_project_id}}/locations/{{region}}/keyRings/{{key_ring_id}}/cryptoKeys/{{key}}`.
	// For AutoKey, replace {{key_ring_id}} to `autokey`, i.e. `projects/{{kms_project_id}}/locations/{{region}}/keyRings/autokey/cryptoKeys/{{key}}`.
}
```
The main difference in my approach is I moved the `External` field up to the KMSKeyRef_OneOf level. This allows it to accept both normal key and auto key, simplifying the UX.  Instead of specifying `kmsKeyRef.External` or `kmsKeyRef.kmsKeyHandlerRef.External`, we could simply use `kmsKeyRef.External` for both cases. And their format are the same(see field comment).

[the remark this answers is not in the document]
gemmahou: IIUC your example code is not backward compatible for ApigeeInstance([code](https://github.com/gemmahou/k8s-config-connector/blob/a5c6d5c2245ebce9929fbd03a9424b21a8ab3c44/apis/apigee/v1beta1/instance_types.go#L60)),:
```
type KMSKey_oneOf struct {
    KMSCryptoKeyRef *refs.KMSCryptoKeyRef  `json:"kmsCryptoKeyRef,omitempty"`
    KMSKeyHandlerRef *refs.KMSKeyHandlerRef  `json:"kmsKeyHandlerRef,omitempty"`
}

type ApigeeInstanceSpec struct {
    DiskEncryptionKMSCryptoKeyRef *refs.KMSKey_oneOf  `json:",inline"`
}
```
before yaml:
```
spec:
   diskEncryptionKMSCryptoKeyRef:
      name:
```

after yaml:
```
spec:
   kmsCryptoKeyRef:
      name:
```

PR #4370, opened 2025-04-15
tests: use server-side-apply for create and update

justinsb:
If we mix-and-match, we get surprising behaviours e.g. we can't clear a field

We would ideally roll this out for all resources, but there are
a lot of tests that accidentally verify the odd behvaiour,
so practically we need to roll it out group-by-group.

### DIFF
--- a/tests/e2e/unified_test.go
+++ b/tests/e2e/unified_test.go
@@ -118,7 +118,10 @@ func TestAllInSeries(t *testing.T) {
 					u.SetAnnotations(annotations)
 				}
 
-				create.RunCreateDeleteTest(h, create.CreateDeleteTestOptions{Create: s.Resources, CleanupResources: true})
+				opt := create.CreateDeleteTestOptions{Create: s.Resources, CleanupResources: true}
+				// samples don't do updates so not using SSA is less problematic
+				opt.DoNotUseServerSideApplyForCreate = true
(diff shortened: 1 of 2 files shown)


tests/e2e/unified_test.go
```diff
@@ -197,6 +197,17 @@ func testFixturesInSeries(ctx context.Context, t *testing.T, testPause bool, can
 						u := bytesToUnstructured(t, fixture.Update, uniqueID, project)
 						opt.Updates = append(opt.Updates, u)
 					}
+
+					// We want to use SSA everywhere, but some of our tests are broken by SSA
+					switch group := primaryResource.GetObjectKind().GroupVersionKind().Group; group {
```
jingyih, line 202: We can effectively filter by API groups only in the "fixtures" tests. For the samples test, maybe we just stick with "no SSA" for now.

jingyih (commented):
/lgtm
/approve

tests/e2e/unified_test.go
justinsb: Good call, will set no-ssa for samples

yuwenma (commented):
/approve
/lgtm

Agree! It's hard to trigger or expose the client-side apply CREATE failure in fixture or sample test.

jingyih (commented):
/lgtm

PR #4372, opened 2025-04-15
gcloud tests for subnet

justinsb:
- **mockgcp: add test for gcloud compute networks subnets**
  

- **mockgcp: more fidelity for compute network / subnets**


justinsb:
So this uncovered that a lot of our tests that create networks were issuing an extra PATCH to set networkFirewallPolicyEnforcementOrder, e.g.

```
PATCH https://compute.googleapis.com/compute/v1/projects/${projectId}/global/networks/${subnetworkID}?alt=json
User-Agent: kcc/${kccVersion} (+https://github.com/GoogleCloudPlatform/k8s-config-connector) kcc/controller-manager/${kccVersion}

{
  "networkFirewallPolicyEnforcementOrder": "AFTER_CLASSIC_FIREWALL"
}
```

This value seems to be the default and is set on network create, so this change should be more similar to reality.

mockgcp/mockcompute/networksv1.go
```diff
@@ -68,6 +68,12 @@ func (s *NetworksV1) Insert(ctx context.Context, req *pb.InsertNetworkRequest) (
 	obj.SelfLink = PtrTo(buildComputeSelfLink(ctx, name.String()))
 	obj.SelfLinkWithId = PtrTo(buildComputeSelfLink(ctx, fmt.Sprintf("projects/%s/global/networks/%d", name.Project.ID, id)))
 	obj.Kind = PtrTo("compute#network")
+	obj.NetworkFirewallPolicyEnforcementOrder = PtrTo("AFTER_CLASSIC_FIREWALL")
```
jingyih, line 71: Perhaps we should only assign the default value when the request doesn't explicitly specify one. But we can fix later when we actually test this field.

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/5d311a1c4e1429efd7667fe43cc041fb70a818d5/tests/apichecks/testdata/exceptions/missingfields.txt#L768

jingyih (commented):
Cool! Thanks for fixing the additional PATCH requests.

/lgtm
/approve

PR #4374, opened 2025-04-16
tool: fix a bug when annotating types file

jingyih:
Fix a bug that was missed in PR #4075.

Write visited messages AFTER the file annotation is added.


yuwenma (commented):
/lgtm
/approve

Would be good to also upload a resource example on showing the diff before and after this change.

jingyih:
> Would be good to also upload a resource example on showing the diff before and after this change.

Due to the bug, the `types.generated.go` file will not include file annotations for resources that doe not have any output only fields. This case is rare but does exist, as with IAPSettings.

PR #4383, opened 2025-04-17
feat: introduce oldObject caching for LoggingLogMetric

acpana:
Introduce an informer based caching of "oldObject" updates.

Next PR will handle sensitive fields hashing.


pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -93,6 +96,7 @@ func NewReconciler(mgr manager.Manager, immediateReconcileRequests chan event.Ge
 			ResourceNameLabel: metrics.ResourceNameLabel,
 		},
 		jitterGenerator: jg,
+		oldObjectsCache: &sync.Map{},
```
cheftako, line 99: How hard is it to switch to using the informer library? It has the old object caching built in.

pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -102,6 +106,52 @@ func add(mgr manager.Manager, r *DirectReconciler, reconcilePredicate predicate.
...
+				// Fetch the full object manually using client.Client
+				fullObj := &unstructured.Unstructured{}
+				fullObj.SetGroupVersionKind(r.gvk)
+
+				if err := mgr.GetClient().Get(context.Background(), key, fullObj); err != nil {
+					log.FromContext(context.Background()).Error(err, "failed to fetch full object in UpdateFunc")
+					return
+				}
+
+				r.oldObjectsCache.Store(key, fullObj.DeepCopy())
```
maqiuyujoyce, line 147: Can we only cache the MBUR fields (so let each resource controller to process the object to be cached) that we care about instead of the full object? Considering that we don't use it for a full diff and considering possible resource limitation to manage thousands of resources. And we can extend it in the future when we need more fields.

pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -102,6 +106,52 @@ func add(mgr manager.Manager, r *DirectReconciler, reconcilePredicate predicate.
...
+				key := types.NamespacedName{
+					Name:      oldMeta.GetName(),
+					Namespace: oldMeta.GetNamespace(),
+				}
+
+				// Fetch the full object manually using client.Client
+				fullObj := &unstructured.Unstructured{}
+				fullObj.SetGroupVersionKind(r.gvk)
+
+				if err := mgr.GetClient().Get(context.Background(), key, fullObj); err != nil {
```
jingyih, line 142: I understand we are using the old obj meta to fetch the object. But isn't the object we're getting already updated?

ldanielmadariaga:

PR #4384, opened 2025-04-17
Direct Controller for datacatalog entry

barney-s:


config/tests/samples/create/harness.go
```diff
@@ -849,7 +849,9 @@ func MaybeSkip(t *testing.T, name string, resources []*unstructured.Unstructured
 
 			case schema.GroupKind{Group: "containeranalysis.cnrm.cloud.google.com", Kind: "ContainerAnalysisNote"}:
 
+			case schema.GroupKind{Group: "datacatalog.cnrm.cloud.google.com", Kind: "DataCatalogEntry"}:
 			case schema.GroupKind{Group: "datacatalog.cnrm.cloud.google.com", Kind: "DataCatalogEntryGroup"}:
+			case schema.GroupKind{Group: "datacatalog.cnrm.cloud.google.com", Kind: "DataCatalogEntry"}:
```
jingyih, line 854: Remove duplicated entry?

pkg/controller/direct/datacatalog/entry_controller.go
```diff
@@ -0,0 +1,365 @@
...
+	status := &krm.DataCatalogEntryStatus{}
+	status.ObservedState = DataCatalogEntryObservedState_FromProto(mapCtx, created)
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+	status.ExternalRef = direct.LazyPtr(a.id.String())
+	return createOp.UpdateStatus(ctx, status, nil)
+}
+
+func (a *entryAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOperation) error {
```
jingyih, line 176: I don't see an update.yaml in the test?

pkg/controller/direct/datacatalog/entry_controller.go
```diff
@@ -0,0 +1,365 @@
...
+var _ directbase.Model = &entryModel{}
+
+type entryModel struct {
+	config config.ControllerConfig
+}
+
+func (m *entryModel) client(ctx context.Context, projectID string) (*api.Client, error) {
+	var opts []option.ClientOption
+
+	// projectID is now passed as argument
```
maqiuyujoyce, line 63: Nit: Clean up?

pkg/controller/direct/datacatalog/entry_controller.go
```diff
@@ -0,0 +1,365 @@
...
+	log.V(2).Info("getting datacatalog entry", "name", a.id)
+
+	req := &pb.GetEntryRequest{Name: a.id.String()}
+	actual, err := a.gcpClient.GetEntry(ctx, req)
+	if err != nil {
+		if direct.IsNotFound(err) {
+			return false, nil
+		}
+		// PermissionDenied is also observed for non-existent entries
+		if direct.IsPermissionDenied(err) {
```
maqiuyujoyce, line 136: Just to gather more context: why does API returns either NotFound or PermissionDenied for non-existent entries? How can we tell whether this is a true permission issue?

pkg/controller/direct/datacatalog/entry_controller.go
```diff
@@ -0,0 +1,365 @@
...
+			isDeletable = true // Assume user-specified system entries are deletable
+		case *pb.Entry_IntegratedSystem:
+			// Check if the integrated system entry type allows deletion
+			switch a.actual.GetIntegratedSystem() {
+			case pb.IntegratedSystem_INTEGRATED_SYSTEM_UNSPECIFIED:
+				// Unspecified might imply custom or deletable? Needs clarification. Let's assume not deletable for safety.
+			case pb.IntegratedSystem_BIGQUERY, pb.IntegratedSystem_CLOUD_PUBSUB, pb.IntegratedSystem_DATAPROC_METASTORE, pb.IntegratedSystem_DATAPLEX, pb.IntegratedSystem_CLOUD_SPANNER, pb.IntegratedSystem_CLOUD_BIGTABLE, pb.IntegratedSystem_CLOUD_SQL, pb.IntegratedSystem_LOOKER, pb.IntegratedSystem_VERTEX_AI:
+				// These are typically synced and not deletable via DataCatalog API directly.
+				log.V(1).Info("Skipping deletion of synced datacatalog entry", "name", a.id, "system", a.actual.GetIntegratedSystem())
+				return false, fmt.Errorf("entries synced from %s cannot be deleted directly via the Data Catalog API", a.actual.GetIntegratedSystem().String()) // Use String() for better error message
```
maqiuyujoyce, line 319: We don't need to finalize a solution right now, but I think when a resource is not deletable and we `return false, err` here, then the resource will be stuck at deleting state and will never be recovered/deleted via normal process. In an ideal world, this resource should be back to UpToDate, or a new NonDeletable state and with this error message.

config/tests/samples/create/harness.go
barney-s: done.

pkg/controller/direct/datacatalog/entry_controller.go
barney-s: yes. done.

pkg/controller/direct/datacatalog/entry_controller.go
barney-s: From a security point of view i can see this happening to things like username or sensitive keys. To prevent scrappers from identifying which keys are available at the server (404 not available, 403 available but permission denied). Not sure why they are doing for this resource.

barney-s: In fact it does not return 404 at all. The code above this 403 check is mostly defensive coding.

pkg/controller/direct/datacatalog/entry_controller.go
barney-s: ah are you saying delete must always pass. How do we deal with non-deleteable resources ? Finalizer ?

barney-s: or log and abandon GCP resource ?

pkg/controller/direct/datacatalog/entry_controller.go
barney-s: added.

jingyih (commented):
/approve

Defer to @maqiuyujoyce

apis/datacatalog/v1alpha1/entry_types.go
```diff
@@ -325,6 +307,24 @@ type DataCatalogEntryObservedState struct {
...
+	//
+	//  For entries with a `user_specified_type`, this field is optional and
+	//  defaults to an empty string.
+	//
+	//  The resource string must contain only letters (a-z, A-Z), numbers (0-9),
+	//  underscores (_), periods (.), colons (:), slashes (/), dashes (-),
+	//  and hashes (#).
+	//  The maximum size is 200 bytes when encoded in UTF-8.
+	// +kcc:proto:field=google.cloud.datacatalog.v1.Entry.linked_resource
+	LinkedResource *string `json:"linkedResource,omitempty"`
```
maqiuyujoyce, line 326: No action needed.

I'm good with making this field output-only to begin with, though from the field description, it looks like this field can be user-specified in some use cases. I think we'll want to support it under spec eventually.

pkg/controller/direct/datacatalog/entry_controller.go
maqiuyujoyce: Understood. I guess we'll just try to create a new resource whenever the error is PermissionDenied.

pkg/controller/direct/datacatalog/entry_controller.go
maqiuyujoyce: I think we just delete the KRM object (`return true, nil`) directly without taking any action for the GCP resource.

And optionally we add warnings in the public doc clarifying that this GCP resource is non-deletable.

maqiuyujoyce (commented):
/lgtm

Nothing blocking the merge of this v1alpha1 resource though it has some edge cases we need to revisit before we promote it to v1beta1.

PR #4385, opened 2025-04-17
chore: clean up test skip logic

justinsb:
It looks like tests are not being properly skipped in test-e2e-fixtures

Simplify and add logging.

### DIFF
--- a/dev/tasks/run-e2e
+++ b/dev/tasks/run-e2e
@@ -85,6 +85,14 @@ if [[ -z "${PREBUILT_TEST_BINARY:-}" ]]; then
   go test -c -o ${PREBUILT_TEST_BINARY} ./tests/e2e
 fi
 
+if [[ -n "${SKIP_TEST_APIGROUP:-}" ]]; then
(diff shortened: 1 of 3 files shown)


dev/tasks/run-e2e
```diff
@@ -85,6 +85,14 @@ if [[ -z "${PREBUILT_TEST_BINARY:-}" ]]; then
   go test -c -o ${PREBUILT_TEST_BINARY} ./tests/e2e
 fi
 
+if [[ -n "${SKIP_TEST_APIGROUP:-}" ]]; then
```
yuwenma, line 88: This is setup in the upper layer before calling runthe run-e2e https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/72e31ffad105c99a3689a7ef54e0f13666af46ef/scripts/github-actions/tests-e2e-fixtures#L45C45-L45C64 
Mind sharing the problem you found?

justinsb: It didn't seem to be skipping correctly.  I still don't see why not, but this works and before we were running e.g. bigtableinstance tests.

PR #4389, opened 2025-04-17
Added a merge metadata command.

cheftako:
### Change description

New command 3 merges a separate metadata file into the main and out to branches-new.yaml. Assumes the merge file takes precedense. Also undoes any skips it finds.

Also added a stable order write file. We will take a 1 time hit the first time this is run but the order should be stable after that.

### Tests you have done

./bin/conductor runner --branch-repo=<repo-dir> --logging-dir=./logs --branch-conf=branches-all.yaml --merge-conf=branches-test.yaml --command=3

branches-test.yaml had 2 skip=true resources with other changed. Confirmed the changes came through.


barney-s (approved):
LGTM

PR #4391, opened 2025-04-18
doc: update release node for 1.131

jingyih:


docs/releasenotes/release-1.131.md
```diff
@@ -0,0 +1,14 @@
+## Announcement    
+
+## New Beta Resources (Direct Reconciler):
+
+* [`IAPSettings`](https://cloud.google.com/config-connector/docs/reference/resource-docs/iap/iapsettings)
```
maqiuyujoyce, line 5: I tried to open https://cloud.google.com/config-connector/docs/reference/resource-docs/iap/iapsettings and it's 404 not found at the moment. I guess we should add the link after it's published?

maqiuyujoyce (commented):
/approve

docs/releasenotes/release-1.131.md
jingyih: I think the page is going to be generated by the release process?
I am just following the old template.

maqiuyujoyce: Yes, it'll be generated possibly as the last step of the release. I think we need to at least ensure the actual release note (not here) contains working URLs. Wondering if GH has any checker for that.

maqiuyujoyce (commented):
/lgtm

Issue #650, opened 2022-04-13
Support only install partial CRDs


liubog2008:
## Checklist

- [x] I did not find a related open enhancement request.
- [x] I understand that enhancement requests filed in the GitHub repository are by default low priority.
- [x] If this request is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

## Describe the feature or resource

There are too many CRDs which are not needed for me. I hope to uninstall some CRDs to avoid k8s discovery throttling.

maqiuyujoyce:
Hi @liubog2008 , thank you for your feedback!

> I hope to uninstall some CRDs to avoid k8s discovery throttling.

Could you elaborate a bit about "K8s discovery throttling"? What issue(s) have you encountered? Would be great if you can share more details.

Also is this enhancement feature a blocker, a friction point, or a nice-to-have to your project?

liubog2008:
nice-to-have

See https://github.com/kubernetes/kubernetes/pull/105520

And too many CRDs lead to many problems.

Maybe this issue in crossplane is helpful for you. https://github.com/crossplane/crossplane/issues/2649

mbzomowski:
Hi @liubog2008, thanks for the suggestion - we'll look into adding this feature.

DoctahPopp871:
This issue makes it hard to use kubectl to interact with GKE clusters that have ConfigConnector enabled. If you could choose the APIs and corresponding CRDs you needed ahead of time, it would dramatically reduce the CRD count, and help mitigate this kubectl bug. Thanks!

maqiuyujoyce:
Hi folks,

We're looking into the options to support CRD partial installation and we'd like your input! Here are some details:

### Option I: New field `crdSelector` in the ConfigConnector kind
For example, you can install all the CRDs containing v1alpha1 version:
```
apiVersion: core.cnrm.cloud.google.com/v1beta1
kind: ConfigConnector
metadata:
  name: configconnector.core.cnrm.cloud.google.com
spec:
  mode: cluster
  googleServiceAccount: "${GSA?}@${PROJECT_ID?}.iam.gserviceaccount.com"
  crdSelector:
    services:
      - all  # Specify the services in the format of the API group to select certain services.
    versions:
      - v1alpha1
```
Or, you can install CRDs based on the kinds you've selected:
```
apiVersion: core.cnrm.cloud.google.com/v1beta1
kind: ConfigConnector
metadata:
[...]

PR #4396, opened 2025-04-18
feat: support partial CRD installation

jingyih:
An `APIGroupInstallation` object uses its `metadata.name` to specify the API group for the resource CRDs to be installed. This guarantees that only one `APIGroupInstallation` object can exist per API group in the cluster.

If no `APIGroupInstallation` objects are present, Config Connector defaults to its previous behavior of installing all resource CRDs.

Note that we do not want to delete CRDs if they are already installed in the cluster, as users may have existing resource objects depending on them. When applying a `APIGroupInstallation` object to an existing Config Connector cluster, you may need to manually remove any resource CRDs that are no long needed. For a fresh Config Connector installation, apply `APIGroupInstallation` objects **before** the `ConfigConnector` CR to ensure that only the specified API group CRDs are installed.

Example usage: the following file instructs Config Connector to install CRDs only for the API groups `iam.cnrm.cloud.google.com`, `compute.cnrm.cloud.google.com`, and `bigtable.cnrm.cloud.google.com`.

```
apiVersion: customize.core.cnrm.cloud.google.com/v1alpha1
kind: APIGroupInstallation
metadata:
  name: iam.cnrm.cloud.google.com
---
apiVersion: customize.core.cnrm.cloud.google.com/v1alpha1
kind: APIGroupInstallation
metadata:
  name: compute.cnrm.cloud.google.com
---
apiVersion: customize.core.cnrm.cloud.google.com/v1alpha1
kind: APIGroupInstallation
metadata:
  name: bigtable.cnrm.cloud.google.com
```

Fixes: #650


jingyih:
/assign @justinsb @maqiuyujoyce 

Please let me know your thoughts!

maqiuyujoyce:
I think this is a good design to start with. Consider it is v1alpha1, I'd be happy to approve it to unblock users if needed.

Though I few thoughts about the UX:
1. It might be a little bit clumsy to manage several API groups at the same time. I may consider the option to set multiple API groups in one CR. We can discuss further about it. Decision about this can be a blocker to promote the feature to v1beta1.
2. It's unclear to users what exact API groups are installed (the observed state) simply with the CRs for partial installation (the desired state). As we won't delete CRDs, I'd consider adding `status.installedAPIGroups` field (or something similar) in the ConfigConnector CRD to explicitly list out installed API groups (or maybe CRDs) to make the state clear.
3. It may be beneficial for us to extend the management of installation to kind level for large API groups like compute.
4. It may be beneficial for us to provide the feature for authoritative management over the non-authoritative management of the installations. I.e. we delete installed CRDs if users don't specify them in the partial installation configs.

jingyih:
Thank you for sharing your thoughts.

> It might be a little bit clumsy to manage several API groups at the same time. I may consider the option to set multiple API groups in one CR. We can discuss further about it. Decision about this can be a blocker to promote the feature to v1beta1.

I think we could have one CR per api-group, or one CR per cluster. Otherwise there could be potential conflict between multiple CRs.

> It's unclear to users what exact API groups are installed (the observed state) simply with the CRs for partial installation (the desired state). As we won't delete CRDs, I'd consider adding status.installedAPIGroups field (or something similar) in the ConfigConnector CRD to explicitly list out installed API groups (or maybe CRDs) to make the state clear.

I like the idea of surfacing this information. Not sure if a full list of installed API groups would be too long to put under ConfigConnector CR. I assume users could do something like `kubectl get apigroupinstallations` which will list CR names which should be the installed API groups.

> It may be beneficial for us to extend the management of installation to kind level for large API groups like compute.

Yes, we could potentially allow listing Kinds under spec.

maqiuyujoyce:
> Thank you for sharing your thoughts.
> 
> > It might be a little bit clumsy to manage several API groups at the same time. I may consider the option to set multiple API groups in one CR. We can discuss further about it. Decision about this can be a blocker to promote the feature to v1beta1.
> 
> I think we could have one CR per api-group, or one CR per cluster. Otherwise there could be potential conflict between multiple CRs.
> 

Yes, I was hesitating between one CR per API group or one CR per Config Connector installation (i.e. per cluster). I like the per cluster option that we don't need to manage too many CRs if say users need half of the API groups (so 35+?); but I like the per API group option for better readability and traceability.

Also with per cluster option, we can potentially support regex for API group configuration. For example, `bigquery*` means I want to install all the CRDs under all API groups starting with "bigquery".

We should check with users for their feedback too.

> > It's unclear to users what exact API groups are installed (the observed state) simply with the CRs for partial installation (the desired state). As we won't delete CRDs, I'd consider adding status.installedAPIGroups field (or something similar) in the ConfigConnector CRD to explicitly list out installed API groups (or maybe CRDs) to make the state clear.
> 
> I like the idea of surfacing this information. Not sure if a full list of installed API groups would be too long to put under ConfigConnector CR. I assume users could do something like `kubectl get apigroupinstallations` which will list CR names which should be the installed API groups.
> 

I think the gap here is that right now there is no authoritative way to manage CRD installation, this means the installed CRDs might be more than what configured in APIGroupInstallation CRs. I hope in addition to the "desired state" (APIGroupInstallations), users can access the "actual state" of their CRD installation (which could be different from the desired state).

But I agree with your concern of the CRD list being long. So it's probably an option to only capture the CRDs not configured in APIGroupInstallation CRs in ConfigConnector CR.

> > It may be beneficial for us to extend the management of installation to kind level for large API groups like compute.
> 
> Yes, we could potentially allow listing Kinds under spec.

Yep saw the TODOs!

PR #4403, opened 2025-04-20
feat: add controller, test and mock support for VMwareEngineExternalAccessRule

jingyih:
HTTP log: real vs. mock (only includes requests relevant to this resource):
https://www.diffchecker.com/GdM306Ad/
The diff includes additional GET requests related to VMware networks, network policies, and similar resources. I think these are triggered by periodic reconciliations that occur roughly every 10 minutes during testing against real GCP, which takes considerably longer.

### DIFF
--- a/pkg/controller/direct/vmwareengine/externalaccessrule_controller.go
+++ b/pkg/controller/direct/vmwareengine/externalaccessrule_controller.go
@@ -33,7 +33,6 @@ import (
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
 	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/vmwareengine/v1alpha1"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/config"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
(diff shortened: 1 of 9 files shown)


pkg/controller/direct/vmwareengine/externalaccessrule_controller.go
```diff
@@ -0,0 +1,290 @@
...
+
+	mapCtx := &direct.MapContext{}
+	desired := a.desired.DeepCopy()
+	resource := VMwareEngineExternalAccessRuleSpec_ToProto(mapCtx, &desired.Spec)
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+
+	paths := []string{}
+	if desired.Spec.Description != nil && !reflect.DeepEqual(resource.Description, a.actual.Description) {
```
maqiuyujoyce, line 172: Just curious, why do you prefer DeepEqual() here instead of CompareProtoMessage()?

pkg/controller/direct/vmwareengine/externalaccessrule_controller.go
```diff
@@ -0,0 +1,290 @@
...
+	if len(desired.Spec.DestinationPorts) != 0 && !reflect.DeepEqual(resource.DestinationPorts, a.actual.DestinationPorts) {
+		paths = append(paths, "destination_ports")
+	}
+	// TODO: handle source_ip_ranges and destination_ip_ranges which contain oneof fields which requires special handling
+	// e.g. https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/7e47fab0c19e7908b88ad1d5e1e6063d27e35fc4/pkg/controller/direct/gkebackup/backupplan_controller.go#L192
+
+	var updated *pb.ExternalAccessRule
+	if len(paths) == 0 {
+		log.V(2).Info("no field needs update", "name", a.id)
+		updated = a.actual
```
maqiuyujoyce, line 196: I think you'll want to set externalRef here for acquisition case.

jingyih: Ah, I didn’t realize the `externalRef` wasn’t added by the LLM. This is a blind spot on my part. I may have overlooked it in other controllers I’ve added. I’ll double-check.

PR #4406, opened 2025-04-21
chore: fix NetworkRef normalization in `BackupDRManagementServer` controller

jingyih:
The previous implementation is technically correct but somewhat unclear. It normalizes the reference in `network`, which is a copy of `obj.Spec.Networks[i]`. This works only because `network.NetworkRef` is a pointer—modifying the copy inadvertently updates the original. The new implementation clarifies the intent and improves readability and maintainability.

### DIFF
--- a/pkg/controller/direct/backupdr/managementserver_controller.go
+++ b/pkg/controller/direct/backupdr/managementserver_controller.go
@@ -255,12 +255,10 @@ func (a *ManagementServerAdapter) Delete(ctx context.Context, deleteOp *directba
 
 func (a *ManagementServerAdapter) normalizeReferenceFields(ctx context.Context) error {
 	obj := a.desired
-	if obj.Spec.Networks != nil {
-		for _, network := range obj.Spec.Networks {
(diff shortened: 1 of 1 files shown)


maqiuyujoyce:
/lgtm
/approve

I think the key point is that we want to ensure `.external` is configured after the reference normalization, right? I'd prefer we do it programmatically, for example, we can normalize the way to normalize references and introduce unit tests to prevent potential mistakes; or we can checking functions in direct base after Create and Update to make this behavior correct.

jingyih:
> /lgtm /approve
> 
> I think the key point is that we want to ensure `.external` is configured after the reference normalization, right? I'd prefer we do it programmatically, for example, we can normalize the way to normalize references and introduce unit tests to prevent potential mistakes; or we can checking functions in direct base after Create and Update to make this behavior correct.

Good idea. The trick is that normalization occurs at runtime, we likely want to perform the check beforehand (like in unit test). As you suggested, this means we need to establish a standardized way to normalize references.

PR #4398, opened 2025-04-18
[Cloud Bigtable App Profile Beta Migration] Record real GCP log

djyau:
### Change description

Fixes #

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #4407, opened 2025-04-21
mock: update mock for BigTable AppProfile to match real

jingyih:
Follow up of #4398.

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/bigtable/v1beta1/bigtableappprofile/_http.log
+++ b/pkg/test/resourcefixture/testdata/basic/bigtable/v1beta1/bigtableappprofile/_http.log
@@ -359,25 +359,10 @@ X-Frame-Options: SAMEORIGIN
 X-Xss-Protection: 0
 
 {
-  "done": true,
   "metadata": {
     "@type": "type.googleapis.com/google.bigtable.admin.v2.UpdateAppProfileMetadata"
   },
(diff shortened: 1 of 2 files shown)


djyau (approved):

jingyih:
/assign @maqiuyujoyce 
for approval

pkg/test/resourcefixture/testdata/basic/bigtable/v1beta1/bigtableappprofile/_http.log
```diff
@@ -363,21 +363,7 @@ X-Xss-Protection: 0
   "metadata": {
```
maqiuyujoyce, line 363: This comment is for `"done": true,` on line 361. From https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4398/commits/32ee880971f74f492edd98e2ec299c2a6137fcc7 it seems that done should be false? Or did I miss anything from the real log?

jingyih: I'm not sure I understand the question. I believe the real GCP response doesn't include `"done":`, but based on the observed behavior (there is no follow-up GET request on the operation), I think the operation is "done".

The `"done": true` field is something our mock introduces. I'm not sure how to remove it, but I think it's safe to ignore since the behavior is consistent between the mock and the real service.

PR #4411, opened 2025-04-21
Marked KCC resources as terraform-v1beta1 based on CRD evaluation.

cheftako:
### Change description

Determined the set of TF-v1beta1 resources based on the following. 
```
for file in `egrep "cnrm.cloud.google.com/tf2crd: \"true\"" * | sed 's|:.*||;'`; do echo `yq '.spec.group' $file | sed 's|.cnrm.cloud.google.com||;' && yq '.spec.names.kind' $file && yq '.spec.versions.[] | select(.name == "v1beta1") | .name' $file` | egrep v1beta1 | sed 's|\"||g;' ; done 
```
Marked the relevant resources with the controller type. Small amounts of related metadata cleanup.
A few entried had to be deduped.
Also added a new command to filter metadata based on controller type.

### Tests you have done

Made sure command 1 indicated the metadata was still good.
Checked the number of controllers of the new type matched.
Made sure I could filter the right number of metadata entries.

### DIFF
--- a/experiments/conductor/branches-all.yaml
+++ b/experiments/conductor/branches-all.yaml
@@ -101,15 +101,15 @@ branches:
       host-name: ""
       notes: []
       apis-enabled: []
-    - name: accesscontextmanager-accesscontextmanagerserviceperimeterresources
-      local: resource-accesscontextmanager-accesscontextmanagerserviceperimeterresources
-      remote: resource-accesscontextmanager-accesscontextmanagerserviceperimeterresources
+    - name: accesscontextmanager-accesscontextmanagerserviceperimeterresource
+      local: resource-accesscontextmanager-accesscontextmanagerserviceperimeterresource
+      remote: resource-accesscontextmanager-accesscontextmanagerserviceperimeterresource
       command: ""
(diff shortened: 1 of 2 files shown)


experiments/conductor/branches-all.yaml
```diff
@@ -135,15 +135,15 @@ branches:
...
-      local: resource-accesscontextmanager-accesscontextmanageraccesslevels
-      remote: resource-accesscontextmanager-accesscontextmanageraccesslevels
+    - name: accesscontextmanager-accesscontextmanageraccesslevel
+      local: resource-accesscontextmanager-accesscontextmanageraccesslevel
+      remote: resource-accesscontextmanager-accesscontextmanageraccesslevel
       command: gcloud access-context-manager levels
       group: accesscontextmanager
-      resource: accesscontextmanageraccesslevels
-      controller: terraform
+      resource: accesscontextmanageraccesslevel
```
justinsb, line 143: As an aside, I'm not sure what `resource` is (I think it is always kind in lower case) ... unrelated to this PR though

justinsb (approved):

PR #4414, opened 2025-04-22
chore: update branches-all.yaml

jingyih:


experiments/conductor/branches-all.yaml
```diff
@@ -25454,25 +25264,8 @@ branches:
       host-name: ""
       notes:
         - No gcloud command found
-      apis-enabled: []
-    - name: vmwareengine-vmwareenginenetwork
```
jingyih, line 25458: Removed because it is the duplicate of `vmwareengine-network`.

jingyih (commented):
File updated. PTAL!

experiments/conductor/branches-all.yaml
```diff
@@ -48,24 +30,7 @@ branches:
       host-name: ""
       notes:
         - No gcloud command found
-      apis-enabled: []
-    - name: accessapproval-approvalrequest
-      local: resource-accessapproval-approvalrequest
-      remote: resource-accessapproval-approvalrequest
-      command: gcloud access-approval requests
```
cheftako, line 55: approval requests can be approved, dismissed or invalidated. That is awfully imperative but I can see how you could make a KCC resource from it. Having said that going to go with it for now. Maybe better to wait for an actual CUJ.

experiments/conductor/branches-all.yaml
```diff
@@ -14858,11 +14730,15 @@ branches:
       package: google.cloud.gkebackup.v1
```
cheftako, line 14662: At some point we should revisit these. These seem awfully imperative. Wondering if they should be a better thought out field rather than a resource.

experiments/conductor/branches-all.yaml
```diff
@@ -20825,8 +20698,8 @@ branches:
       proto-service: ""
       proto-msg: google.cloud.resourcemanager.v3.Folder
       host-name: ""
-      notes: []
-      apis-enabled: []
+      notes:
+        - Terraform BETA
```
cheftako, line 20635: Kind: should be "Folder" (https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/crds/resourcemanager_v1beta1_folder.yaml#L32)
That is I believe a different entry in branches-all.yaml

experiments/conductor/branches-all.yaml
```diff
@@ -21257,32 +21131,34 @@ branches:
...
     - name: run-service
       local: resource-run-service
       remote: resource-run-service
       command: gcloud run services
       group: run
       resource: service
-      controller: Unknown
+      controller: terraform
       skip: false
       kind: RunService
```
cheftako, line 21163: I think this conflicts with another RunService in the file.

experiments/conductor/branches-all.yaml
```diff
@@ -21248,7 +21122,7 @@ branches:
       command: gcloud run jobs
       group: run
       resource: job
-      controller: Unknown
+      controller: terraform
       skip: false
       kind: RunJob
```
cheftako, line 21127: I think this conflicts with another RunJob in the file.

experiments/conductor/branches-all.yaml
jingyih: Not sure I understand. The kind is `Folder` in this entry.

experiments/conductor/branches-all.yaml
jingyih: De-duplicated.

experiments/conductor/branches-all.yaml
jingyih: De-duplicated.

PR #4434, opened 2025-04-24
Run controller managers in separate namespaces

dshnayder:
# Change Description
## Summary

This change introduces the capability to deploy Kubernetes Config Connector (KCC) controller managers into dedicated namespaces, separate from the default `cnrm-system` namespace. This enhancement improves the security posture of KCC deployments in multi-tenant environments, by enabling namespace-level isolation for controller manager instances.

## Motivation

Currently, all KCC controller managers operate within the shared `cnrm-system` namespace, irrespective of the specific ConfigConnectorContext or tenant they serve. This centralized deployment model presents a security concern: it increases the potential blast radius in the event of a security compromise or critical failure affecting one controller manager. A vulnerability exploited within `cnrm-system` could potentially impact all tenants managed by that KCC installation due to the shared execution environment.

The goal of this refactoring is to mitigate this risk by allowing controller managers to be deployed into distinct, dedicated namespaces. This architectural change facilitates stronger isolation between tenants or configurations, thereby reducing the scope of impact from potential security incidents or operational issues within a single manager instance. This approach aligns with best practices for secure multi-tenant system design in Kubernetes.

## Implementation Details

KCC has been refactored to support dynamic targeting of namespaces for controller manager `StatefulSets` and their associated `ServiceAccounts`.

This feature is activated and configured using an optional label on the `ConfigConnector` custom resource:

- Label: `cnrm.cloud.google.com/manager-namespace-suffix`
When this label is present on a `ConfigConnector` resource, the system is instructed not to deploy the corresponding controller manager components into the default `cnrm-system` namespace. Instead, a dedicated namespace is used. The selection of this label as the configuration mechanism adheres to standard Kubernetes practices for declarative configuration, allowing behavior modification without altering the core ConfigConnector API specification.

The name of the dedicated target namespace is derived programmatically based on the namespace of the associated ConfigConnectorContext resource:

- The namespace of the relevant `ConfigConnectorContext` is identified.
- The suffix of this namespace name (the part after the last hyphen, assuming a convention like `project-tenant-type`) is replaced with the value specified in the `cnrm.cloud.google.com/manager-namespace-suffix` label.

**Example**:
Consider a ConfigConnectorContext resource located in the namespace `t1234-tenant0-provider`. If the corresponding `ConfigConnector` resource includes the label `cnrm.cloud.google.com/manager-namespace-suffix=supervisor`, then the controller manager `StatefulSet` and its `ServiceAccount` will be created within the `t1234-tenant0-supervisor` namespace. This deterministic naming convention facilitates predictable management in structured environments.

**Scope and Impact**

This enhancement is required for multi-tenant GKE deployments, where strict tenant isolation is often a requirement.

**Positive Impacts**:
- Significantly reduces the blast radius by isolating controller managers per tenant or configuration context.
- Aligns with best practices for secure multi-tenant system design in Kubernetes

**Negative Impacts**:
- When a new ConfigConnectorContext or Custom Resource is created, an additional request to `apiserver` required to get label of `ConfigConnector` resource.

### Tests

- Enable KCC as add-on in MT cluster. This installs KCC in namespaced mode. Created a tenant.
  - Created a ConfigConnectorContext in the provider namespace. Verified that StatefulSets/Pod cnrm-controller-manager-XXX are created in the supervisor namespace.
  - Created PubSubTopic in the provider namespace. Verified that corresponding object is created in GCP.
  - Removed the ConfigConnectorContext. Verified that StatefulSets/Pod cnrm-controller-manager-XXX are deleted
  - Recreated ConfigConnectorContext and PubSubTopic, then deleted tenant. Verified that GCP object, ConfigConnectorContext, namespaces and tenant were deleted successfully.
- Deleted ConfigConnector
  - Verified that all StatefulSets/Pods are removed
- Created ConfigConnector in cluster mode
  - Verified that StatefulSets/Pod cnrm-controller-manager are created in the `cnrm-system` namespace


install-bundles/install-bundle-autopilot-namespaced/0-cnrm-system.yaml
```diff
@@ -1509,6 +1509,12 @@ rules:
   verbs:
   - create
   - patch
+- apiGroups:
```
justinsb, line 1512: Did this change come from an [RBAC annotation](https://book.kubebuilder.io/reference/markers/rbac) ?  I think it should, but I don't see the change.

justinsb:
I think this looks good, I think I'm satisfied that it won't change the behaviour in the common case when we are _not_ running in this mode.  It would be great if we had some golden tests here, but I think we need to upgrade k-d-p etc to pick up some helpers there.

One question about how you made the RBAC though (which I think is causing the merge conflict)

I am not sure that we correctly clean up old statefulset if users switch back-and-forth between the "peer namespace" mode and the normal namespace mode, but I could well be wrong there and I'm not sure it's a likely scenario anyway.
 
cc @cheftako for a second pair of eyes.

install-bundles/install-bundle-autopilot-namespaced/0-cnrm-system.yaml
dshnayder: This didn't come from RBAC annotation, will fix. Thanks for the suggestion.

dshnayder:
> I think this looks good, I think I'm satisfied that it won't change the behaviour in the common case when we are _not_ running in this mode. It would be great if we had some golden tests here, but I think we need to upgrade k-d-p etc to pick up some helpers there.
> 
> One question about how you made the RBAC though (which I think is causing the merge conflict)
> 
> I am not sure that we correctly clean up old statefulset if users switch back-and-forth between the "peer namespace" mode and the normal namespace mode, but I could well be wrong there and I'm not sure it's a likely scenario anyway.
> 
> cc @cheftako for a second pair of eyes.

@justinsb , tests are added. I didn't forget about test, just wanted to get your opinion on the functionality change.

install-bundles/install-bundle-autopilot-namespaced/0-cnrm-system.yaml
dshnayder: Removed udpates from generated YAML files and updated template "config/installbundle/releases/scopes/namespaced/shared-components/unmanageddetector_clusterrole.yaml"

dshnayder:
Due to complex rebase a new PR created:
https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4523

PR #4437, opened 2025-04-25
chore: update externalRef in adapter update

jingyih:
Following up on a comment in #4403 - add `externalRef` in "update" for some of the controllers I recently added.


maqiuyujoyce (commented):
/lgtm
/approve

Good reminder as I'll need to update some of my existing controllers as well...

PR #4444, opened 2025-04-25
Update status in adapter Update() even without change

maqiuyujoyce:
### Change description

I kept the `if` check to update status only when `status.externalRef` is not set for AlloyDBCluster and AlloyDBInstance. This is to reduce unnecessary updates.

In the future, I think we should update status in Update() when:
1.  Drift is detected and fixed.
2. `status.externalRef` is unset.
3. `status.observedState` is updated.

Right now for many controllers we update status regardlessly, and I'm a bit concerned this will impact scalability when there are many resources. Though it might be cheap/free to update status (?); we need to discuss further.

### Tests you have done


jingyih (commented):
/lgtm
/approve

PR #4441, opened 2025-04-25
chore: refactor stateintospec logic into one package

justinsb:

PR #4446, opened 2025-04-26
chore: pass defaulters to all controllers, eagerly write defaults

justinsb:
- **chore: pass defaulter to all controllers**
  We want to be sure that they are doing the right thing here.

- **chore: do not default state-into-spec for direct controllers**

- **feat: ensure all controllers eagerly write back changes**
    Apply the eager-writeback approach to all controllers,
    not just direct controllers.


justinsb:
cc @cheftako ... did we decide if / when we want to write-back the state-into-spec annotations (particularly for new resources)

justinsb:
WIP until #4441

justinsb:
OK so this is how we can get rid of the races involving webhooks.  In #4442 I will introduce the project-id defaulter, right now this is just setting up the machinery and validating that nothing changes.

pkg/controller/tf/controller.go
```diff
@@ -56,13 +56,14 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	crlog "sigs.k8s.io/controller-runtime/pkg/log"
```
yuwenma, line 59: any reason to import the ctrl-runtime log twice and used separately?

pkg/stateintospec/stateintospecdefaulter.go
```diff
@@ -54,15 +54,23 @@ func NewStateIntoSpecDefaulter(client client.Client) k8s.Defaulter {
...
 	val, found := resource.GetAnnotations()[k8s.StateIntoSpecAnnotation]
 	if found {
-		if !isAcceptedStateIntoSpecValue(val, resource.GetObjectKind().GroupVersionKind()) {
+		if !isAcceptedStateIntoSpecValue(val, gvk) {
 			return false, fmt.Errorf("invalid value %q for %q annotation in kind %v", val, StateIntoSpecAnnotation, resource.GetObjectKind().GroupVersionKind().Kind)
 		}
 		return false, nil
 	}
 
+	// Don't write the annotation for direct controllers
```
yuwenma, line 68: 👍

yuwenma: Would this affect the direct-resources with reconcile-gate or as opt-in?

yuwenma (commented):
/approve

I have a question about the opt-in/reconcile-gate behavior. otherwise looks great

PR #4447, opened 2025-04-27
Revert "feat: support partial CRD installation"

jingyih:
Reverts GoogleCloudPlatform/k8s-config-connector#4396

Reverts the alpha core CRD to unblock the 1.132 release for the GKE addon. Will merge it back after 1.132. (We should also consider #3272 to create a different flavor specifically for the GKE addon.)


justinsb (approved):

justinsb:
Thanks for doing this.  The concern is that (1) we don't want to install alpha CRDs by default and (2) we want to be sure that everything works when the alpha CRDs are not installed.

So I'm fine to get this / similar functionality back in, but we should make sure there is a test.

I do wonder though whether we should start with configuration on the CCC, as install time is "too late" for many users and I believe that many users want to watch different resources per-namespace.

PR #4449, opened 2025-04-28
tests: skip alpha dynamic tests

gemmahou:
### Change description

Add an option to skip Alpha resources in dynamic test.

Our periodic test failed since last week, due to panic error in one of the Conductor generated resource: https://screenshot.googleplex.com/5diRTqYmdb3a2Hq.png

Given the increasing number of direct-based Alpha resources, I think maybe we could exclude them from periodic tests to better focus our monitoring on Beta resources. I will update the internal test script in followup CLs.

### Tests you have done

### DIFF
--- a/pkg/controller/dynamic/dynamic_controller_integration_test.go
+++ b/pkg/controller/dynamic/dynamic_controller_integration_test.go
@@ -76,6 +76,9 @@ func init() {
 	flag.StringVar(&runTestsRegex, "run-tests", "", "run only the tests whose names match the given regex")
 	flag.StringVar(&skipTestsRegex, "skip-tests", "", "skip the tests whose names match the given regex, even those that match the run-tests regex")
 
+	// Only run tests associate with the given API version. Supported values are beta or alpha.
+	flag.StringVar(&runTestsVersion, "run-tests-version", "", "only run the tests associate with the given API version")
+
 	// cleanup-resources allows you to disable the cleanup of resources created during testing. This can be useful for debugging test failures.
 	// The default value is true.
 	//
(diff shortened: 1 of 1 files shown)


maqiuyujoyce (commented):
/lgtm
/approve
/hold

Have you verified this flag works as expected?

pkg/controller/dynamic/dynamic_controller_integration_test.go
```diff
@@ -127,6 +131,13 @@ func shouldRunBasedOnRunAndSkipRegexes(parentTestName string, fixture resourcefi
 		}
 	}
 
+	// If skipAlphaTests flag has been provided and set to true, skip the test if its associate to Alpha resources
+	if skipAlphaTests == "true" {
```
jingyih, line 135: I like the idea of having a cleaner test signal for beta resources. Maybe we should create a separate testgrid for alpha resources? I think dynamic tests should be mandatory when promoting a resource to beta. Personally, I always fix dynamic tests before promoting an alpha resource.

Nit: this check should not be included in "shouldRunBasedOnRunAndSkipRegexes."

gemmahou:
> /lgtm /approve /hold
> 
> Have you verified this flag works as expected?

Yes, I tested with the `networkservicesservicebinding-minimal` test: https://screenshot.googleplex.com/9PzkZKwJwwfQ5EG.png

pkg/controller/dynamic/dynamic_controller_integration_test.go
gemmahou: Changed function name to `shouldRunBasedOnFlag`. 

We have test grid for autogen Alpha resources, by filtering the test suffix `-autogen`.  If we want to have a separate testgrid for alpha resources, then we may need to add another flag like `skipBetaTests` or `runAlphaTestsOnly`

pkg/controller/dynamic/dynamic_controller_integration_test.go
```diff
@@ -76,6 +76,11 @@ func init() {
 	flag.StringVar(&runTestsRegex, "run-tests", "", "run only the tests whose names match the given regex")
 	flag.StringVar(&skipTestsRegex, "skip-tests", "", "skip the tests whose names match the given regex, even those that match the run-tests regex")
 
+	// Only run tests associate with Alpha resources
+	flag.BoolVar(&runAlphaTests, "run-alpha-tests", false, "only run the tests associate with Alpha resources")
```
jingyih, line 80: nit: the two flags feel a bit conflicting. If we only want to support two mutually exclusive scenarios (running either alpha or beta tests), a single flag like `--test-group=[alpha|beta]` might be clearer.  
If we want to allow running both alpha and beta tests together in the future, then maybe something like `--run-alpha-tests` and `--run-beta-tests`? Let me know your thoughts!

maqiuyujoyce:
Looks good to me. Thank you for the improvement of the UI!

/hold cancel

PR #4450, opened 2025-04-29
[Cloud Bigtable App Profile Beta Migration] Record real GCP log after cleaning up project instance

djyau:
…options - CBT only has GRPC

### Change description

This PR takes the recorded real GCP log from https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4345 which uses the direct controller. Since the CBT golang client only offers a grpc client (which we use in the controller), we need to update the gcp log with the GRPC client output/responses and then update the mock implementation afterwards.

### Tests you have done


jingyih:
Is this PR still needed?

PR #4452, opened 2025-04-29
refactor:types: IAMAuditConfig

acpana:
Continuing to convert IAM related types to the "direct" way.


config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_iamauditconfigs.iam.cnrm.cloud.google.com.yaml
```diff
@@ -81,12 +82,39 @@ spec:
...
+                  - name
+                - not:
+                    anyOf:
+                    - required:
+                      - name
+                    - required:
+                      - namespace
+                  required:
+                  - external
+                - not:
```
maqiuyujoyce, line 99: I understand that you are probably copying the schema over from IAMPolicy/IAMPartialPolicy, but the following rule doesn't seem to make much sense to me.

```
                - not:
                    anyOf:
                    - required:
                      - name
                    - required:
                      - namespace
                    - required:
                      - apiVersion
                    - required:
                      - external
```
This basically means "you shouldn't specify any of `name`, `namespace`, `apiVersion`, or `external` field". I don't know if this is a valid case for resourceRef here.

Can we hold on adding the oneOf rule?

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_iamauditconfigs.iam.cnrm.cloud.google.com.yaml
```diff
@@ -81,12 +82,39 @@ spec:
               resourceRef:
                 description: Immutable. Required. The GCP resource to set the IAMAuditConfig
                   on (e.g. project).
+                oneOf:
```
acpana, line 85: @maqiuyujoyce I think this is technically backwards incompatible but I figured this was supposed to be in to begin with.

happy to take it out if you think we shouldn't

maqiuyujoyce: IAM is in general fairly critical so let's try to keep it backwards compatible for now.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_iamauditconfigs.iam.cnrm.cloud.google.com.yaml
acpana: I also got a bit confused by this at first but I think it means that `kind` by itself is ok but not any of {name, namespace, apiVersion, external} by themselves only.

maqiuyujoyce: Yeah, I'm not sure whether `kind` by itself is okay or not though. Have you identified such edge cases?

scripts/add-validation-to-crds/parse-crds.go
```diff
@@ -131,6 +131,10 @@ func processFile(ctx context.Context, p string) error {
 }
 
 func addRefsToCRD(crd *apiextensions.CustomResourceDefinition) error {
+	if crd.Spec.Names.Kind == "IAMAuditConfig" {
+		// no refs here
```
maqiuyujoyce, line 135: Non-blocker: We'll want to provide consistent interface eventually so maybe add a TODO here?

maqiuyujoyce (commented):
/lgtm
/approve

PR #4457, opened 2025-05-01
Generate testdata for BigQueryTable

maqiuyujoyce:
### Change description

Fixes #

### Tests you have done


maqiuyujoyce:
@anhdle-sso I'm using this PR to track the BigQueryTable test generation. I'm pretty impressed that it got the dependencies right!

anhdle-sso:
> @anhdle-sso I'm using this PR to track the BigQueryTable test generation. I'm pretty impressed that it got the dependencies right!

Yea, that looks promising. There are a few things I notice that are missing from dependencies. First. even though we only referencing kms crypto key, crypto key needs IAM binding to be usable. Second, we might want to tweaking that field with id, name etc. might also be reference; looking at the foreign key constraints field.

PR #4458, opened 2025-05-01
Support maximal test generation in conductor

maqiuyujoyce:
### Change description

Added the command that generates test cases to increase coverage for supported resources:
* Currently, the command attempt to generate the test at least once and at most three times to ensure most coverage.
* Use the updated result of tests/apichecks/testdata/exceptions/missingfields.txt to determine whether the test cases cover all the supported fields and gets a list of uncovered fields. We ignore `XXXRef.external` fields as we want to test references via `.name` field. We also ignore `value` field under sensitive fields as we encourage usage of `valueFrom` and currently are focusing on testing the encouraged option.
  * Uncovered fields extraction is not done by LLM because several lines of functions do a better job than codebot.
* Use LLM to generate the create.yaml to cover as many fields in the list of uncovered fields as possible. This includes reference fields.
* Use LLM to generate the dependencies.yaml for the referenced resources in create.yaml.
* Use LLM to generate the update.yaml based on create.yaml that changes as many fields in create.yaml as possible.
* Return the parent directory of the test cases as the changed fields even when the LLM steps error out because sometimes they still generate the files. This is to ensure the next branch can be processed as expected.

Also supported the configuration of the suffix for the generated test cases across commands and enabled skip regex for more than one commit message.

### Tests you have done

### DIFF
--- a/experiments/conductor/cmd/runner/controller_commands.go
+++ b/experiments/conductor/cmd/runner/controller_commands.go
@@ -15,6 +15,7 @@
 package runner
 
 import (
+	"bufio"
 	"context"
 	"errors"
 	"fmt"
@@ -304,7 +305,7 @@ func createControllerTest(ctx context.Context, opts *RunnerOptions, branch Branc
 		"pkg", "test", "resourcefixture", "testdata", "basic",
 		branch.Group, strings.ToLower(crdVersion),
 		strings.ToLower(branch.Kind),
-		fmt.Sprintf("%s-minimal", strings.ToLower(branch.Kind)),
+		fmt.Sprintf("%s-%s", strings.ToLower(branch.Kind), opts.testDirSuffix),
 	)
 
 	fullTestDir := filepath.Join(opts.branchRepoDir, testDir)
@@ -324,7 +325,7 @@ The file should follow these requirements:
 - Add an Apache 2.0 license header with Copyright %d Google LLC
 - Use apiVersion: %s/%s 
 - Use kind: %s
-- Include metadata.name: %s-minimal-${uniqueId}
+- Include metadata.name: %s-%s-${uniqueId}
 - If the CRD has a projectRef field, use projectRef.external: ${projectId}
 - If the CRD has a location field, use location: us-central1
 - Keep the YAML simple but valid for initial creation
@@ -334,7 +335,7 @@ The file should follow these requirements:
 Use ReadFile to read the CRD file.
 Use CreateFile to write the YAML content to the %s/create.yaml file.
 Respond only with the YAML content, no explanations.`,
-		testDir, branch.Group, branch.Kind, currentYear, crdGroup, crdVersion, branch.Kind, strings.ToLower(branch.Kind), testDir)
+		testDir, branch.Group, branch.Kind, currentYear, crdGroup, crdVersion, branch.Kind, strings.ToLower(branch.Kind), opts.testDirSuffix, testDir)
 
 	cfg := CommandConfig{
 		Name:         "Generate Create YAML",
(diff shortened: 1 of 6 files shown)


experiments/conductor/cmd/runner/controller_commands.go
```diff
@@ -334,7 +335,7 @@ The file should follow these requirements:
 Use ReadFile to read the CRD file.
 Use CreateFile to write the YAML content to the %s/create.yaml file.
 Respond only with the YAML content, no explanations.`,
-		testDir, branch.Group, branch.Kind, currentYear, crdGroup, crdVersion, branch.Kind, strings.ToLower(branch.Kind), testDir)
+		testDir, branch.Group, branch.Kind, currentYear, crdGroup, crdVersion, branch.Kind, strings.ToLower(branch.Kind), opts.testDirSuffix, testDir)
```
cheftako, line 338: This sort of positional field substitution seems very error prone. Having said that your not making it worse but at some point later it would be nice to clean this up.

experiments/conductor/cmd/runner/controller_commands.go
```diff
@@ -464,6 +465,302 @@ func moveTestToSubDir(ctx context.Context, opts *RunnerOptions, branch Branch, e
...
+			return nil, nil, fmt.Errorf("error checking whether test directory %s exists: %w", fullTestDir, err)
+		} else {
+			if err := os.MkdirAll(fullTestDir, 0755); err != nil {
+				return nil, nil, fmt.Errorf("failed to create test directory %s: %w", fullTestDir, err)
+			}
+		}
+
+		// 3. Generate testdata.
+		// Use codebot to generate the initial create.yaml content.
+		createPrompt := fmt.Sprintf(`Generate a %s/create.yaml file for testing a Kubernetes controller.
```
cheftako: Please define the prompts as a consts just before the function which uses them. (Eg https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/experiments/conductor/cmd/runner/fuzzer_commands.go#L68) It keeps the code more readable.

experiments/conductor/cmd/runner/controller_commands.go
```diff
@@ -464,6 +465,302 @@ func moveTestToSubDir(ctx context.Context, opts *RunnerOptions, branch Branch, e
...
+	}
+
+	if err := scanner.Err(); err != nil {
+		return nil, fmt.Errorf("error scanning %s: %w", path, err)
+	}
+	return result, nil
+}
+
+func pluralOf(word string) string {
+	if strings.HasSuffix(word, "f") {
```
cheftako: I think this case is hard to get correct. 
Puff -> puffs not puves
staff -> staff or staves not stafves
brief -> brief not brieves

Maybe not worth special the extra complexity.

experiments/conductor/cmd/runner/controller_commands.go
maqiuyujoyce: Good point. Replaced it with place holders.

experiments/conductor/cmd/runner/controller_commands.go
maqiuyujoyce: Done.

experiments/conductor/cmd/runner/controller_commands.go
maqiuyujoyce: Good point. Removed.

experiments/conductor/cmd/runner/controller_commands.go
maqiuyujoyce: ^ Actually, I only cleaned up the newly added prompts but not the existing ones. Will clean it up later.

experiments/conductor/cmd/runner/github_commands.go
```diff
@@ -123,12 +128,25 @@ func skipPost21F(msg string) bool {
...
+
+func convertCommitMsgToRegex(possibleMsgs ...string) string {
+	var result string
+	for i, msg := range possibleMsgs {
+		tmp := strings.ReplaceAll(msg, "{{kind}}", "[a-zA-Z]+")
+		tmp = strings.ReplaceAll(tmp, "{{command}}", "") // handled by terminating ".*"
+		tmp = strings.ReplaceAll(tmp, "{{group}}", "[a-zA-Z]+")
+		tmp = strings.ReplaceAll(tmp, "{{resource}}", "[a-zA-Z]+")
+		// Add grouping for better readability when there are more than one msg template.
+		tmp = "(" + tmp + ")"
```
cheftako, line 143: Are we sure this is generating the correct regex? Worried about both escaping the parentheses correctly and the '|' being outside the parentheses.

cheftako: Can we add tests that the regex for REGEX_MSG_51 and REGEX_MSG_51A both do the right thing?

experiments/conductor/cmd/runner/github_commands.go
```diff
@@ -98,6 +101,8 @@ var REGEX_MSG_46 = regexp.MustCompile(convertCommitMsgToRegex(COMMIT_MSG_46))
 var REGEX_MSG_47 = regexp.MustCompile(convertCommitMsgToRegex(COMMIT_MSG_47))
 var REGEX_MSG_48 = regexp.MustCompile(convertCommitMsgToRegex(COMMIT_MSG_48))
 var REGEX_MSG_50 = regexp.MustCompile(convertCommitMsgToRegex(COMMIT_MSG_50))
+var REGEX_MSG_51A = regexp.MustCompile(convertCommitMsgToRegex(COMMIT_MSG_51A))
+var REGEX_MSG_51 = regexp.MustCompile(convertCommitMsgToRegex(COMMIT_MSG_51B, COMMIT_MSG_51C))
```
cheftako, line 105: Is it correct to say 51 is done if 51C failed? Why not just display 51b and 51C and let the user have more detailed information?

maqiuyujoyce: > Is it correct to say 51 is done if 51C failed?

51C is added in case 44B is not executed, both of them do the same thing: to add GVK to test harness allowlist. So 51C is an optional step. I won't be concerned if 51C is failed or does nothing because the chance that we need it is low.

> Why not just display 51b and 51C and let the user have more detailed information?

Do you mean adding them to the help message? Or something else?

Regarding the step to add GVK to test harness allowlist, think we should change the allowlist to a blocklist so that all the new GVKs are allowed by default without needing this step. Will do it in a separate PR.

experiments/conductor/cmd/runner/github_commands.go
maqiuyujoyce: I tested it out and it worked as I expected. Added some unit tests.

PR #4459, opened 2025-05-01
Support moving existing test to subdir in preparation for adding more test cases

maqiuyujoyce:
### Change description

A lot of resources only have one test case and they are directly under `pkg/test/resourcefixture/testdata/basic/<group>/v1beta1/<kind>` directory.

To support new test cases, we need to move these test cases to a sub directory first.

### Tests you have done

### DIFF
--- a/experiments/conductor/cmd/runner/controller_commands.go
+++ b/experiments/conductor/cmd/runner/controller_commands.go
@@ -387,6 +387,73 @@ Respond only with the YAML content, no explanations.`,
 	}, nil, nil
 }
 
+func moveTestToSubDir(ctx context.Context, opts *RunnerOptions, branch Branch, execResults *ExecResults) ([]string, *ExecResults, error) {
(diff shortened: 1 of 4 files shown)


experiments/conductor/cmd/runner/controller_commands.go
```diff
@@ -387,6 +387,73 @@ Respond only with the YAML content, no explanations.`,
...
+
+	// Check if we have the required fields
+	if branch.Group == "" {
+		return nil, nil, fmt.Errorf("branch %s is missing Group field", branch.Name)
+	}
+
+	if branch.Kind == "" {
+		return nil, nil, fmt.Errorf("branch %s is missing Kind field", branch.Name)
+	}
+
```
cheftako, line 401: Should probably have a check which skips if the destination directory already exists.

maqiuyujoyce: Done!

PR #4464, opened 2025-05-03
feat: promot resource vertex notebook to beta

xiaoweim:
### Change description

Fixes #

### Tests you have done


apis/notebooks/v1beta1/instance_types.go
```diff
@@ -0,0 +1,292 @@
...
+	ObservedState *NotebookInstanceObservedState `json:"observedState,omitempty"`
+}
+
+// NotebookInstanceObservedState is the state of the NotebookInstance resource as most recently observed in GCP.
+// +kcc:proto=google.cloud.notebooks.v1.Instance
+type NotebookInstanceObservedState struct {
+	// Output only. The name of this notebook instance. Format:
+	//  `projects/{project_id}/locations/{location}/instances/{instance_id}`
+	// +kcc:proto:field=google.cloud.notebooks.v1.Instance.name
+	Name *string `json:"name,omitempty"`
```
jingyih, line 233: Maybe I missed in the alpha review. Name field seems identical to the externalRef? (in which case it should be removed)

jingyih (commented):
I'm actually not sure what the criteria are for beta promotion. We currently have only minimal test. Personally, I think we should at least test all the reference fields. Do you know if there's been any agreement on the criteria?

xiaoweim:
> I'm actually not sure what the criteria are for beta promotion. We currently have only minimal test for recently added alpha resources. Personally, I think we should at least test all the reference fields. Do you know if there's been any agreement on the criteria?

I added more fixture tests that covers all the field and different update methods as according to what @yuwenma has said.

jingyih (commented):
/lgtm
/approve

Thank!
Could you please fix the unit test golden files and also run `make ready-pr` to fix the resource docs.

jingyih (commented):
/lgtm

PR #4465, opened 2025-05-03
preview mode: introduce reconcile end event

justinsb:
This should avoid a potential (though unlikely) race condition where we catch an object mid-reconcile.


jingyih (commented):
/approve

PR #4474, opened 2025-05-05
fix: webhooks should pre-check for existence using get

justinsb:
We can't create a service if we have no clusterIPs remaining,
even if the service already exists
(and we don't get the normal 404/NotFound error).

This confuses our webhook-creation logic.

### DIFF
--- a/pkg/webhook/cert/certclient/certclient.go
+++ b/pkg/webhook/cert/certclient/certclient.go
@@ -83,7 +83,7 @@ func certWriterFromOptsOrNew(opts Options) (writer.CertWriter, error) {
 	return certWriter, nil
 }
 
-func (c *CertClient) RefreshCertsAndInstall() error {
+func (c *CertClient) RefreshCertsAndInstall(ctx context.Context) error {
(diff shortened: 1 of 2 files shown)


pkg/webhook/cert/certclient/certclient.go
```diff
@@ -158,58 +158,64 @@ var genericFn = func(current, desired *client.Object) error {
...
+
+	objectKey := client.ObjectKeyFromObject(obj)
+
+	// this cast is not safe per-se but is added to get around the transition from runtime.Object to client.Object
+	existing, ok := obj.DeepCopyObject().(client.Object)
+	if !ok {
+		return fmt.Errorf("unable to cast deep copy %T to client.Object", obj)
+	}
+
+	// NOTE: create *before* get does not always work for services.
```
yuwenma, line 174: Good to know this! It sounds to apply to all (wondering why the ctrl runtime does not handle it)

yuwenma (commented):
/lgtm
/approve

Thanks!

PR #4479, opened 2025-05-06
Move single tests to subdirectory

maqiuyujoyce:
### Change description

Move test cases directly under <kind> directory for top 50 most used resources to the subdirectory in preparation for more test case generation.

This is related to #4459.

### Tests you have done

- [N/A] Run `make ready-pr` to ensure this PR is ready for review.
- [N/A] Perform necessary E2E testing for changed resources.

PR #4481, opened 2025-05-06
Add unit test to validate the path of basic test cases

maqiuyujoyce:
### Change description

Diffbase: #4479 

Fixes paths to test cases under basic directory and ensures they follow the correct format.

### Tests you have done


pkg/test/resourcefixture/testdata/basic/basic_test.go
```diff
@@ -0,0 +1,124 @@
...
+	"strings"
+	"testing"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/gvks/supportedgvks"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/util/repo"
+)
+
+func TestValidTestPath(t *testing.T) {
```
jingyih, line 17: It would be helpful to include some basic test cases to illustrate which test paths are valid and which are not. I envision encapsulating the test path validation logic in a `validateTestPath` function, which would be invoked when running the basic tests. Additionally, we could have a unit test `TestValidateTestPath` with a few dummy test cases to verify the correctness of the validation logic.

maqiuyujoyce: I like the idea of adding a few dummy test cases to verify the validation logic. Though I'd still want to keep the validation here as basic tests are integration tests.

jingyih (commented):
/lgtm
/approve

PR #4482, opened 2025-05-06
chore: refactor managementconflict logic

justinsb:
Put it in its own package so we can better reason about it.


justinsb:
/assign @maqiuyujoyce 

This is the first commit of #4483

PR #4483, opened 2025-05-06
management conflict is not required

justinsb:
- **chore: refactor managementconflict logic**
  Put it in its own package so we can better reason about it.
  

- **Make the management-conflict annotation optional**
  There's no reason to enforce the annotation - it is mutable.
  
  Making it optional means one less thing we depend on the webhook for.

### DIFF
--- a/pkg/managementconflict/annotations_test.go
+++ b/pkg/managementconflict/annotations_test.go
@@ -28,16 +28,16 @@ func TestGetManagementConflictPreventionAnnotationValue(t *testing.T) {
 		ShouldSucceed  bool
 	}{
 		{
-			Name:           "nil annotations should error",
+			Name:           "nil annotations should no longer error",
 			Annotations:    nil,
 			ExpectedPolicy: ManagementConflictPreventionPolicyNone,
-			ShouldSucceed:  false,
+			ShouldSucceed:  true,
 		},
 		{
-			Name:           "missing annotation should error",
+			Name:           "missing annotation should no longer error",
 			Annotations:    make(map[string]string),
 			ExpectedPolicy: ManagementConflictPreventionPolicyNone,
-			ShouldSucceed:  false,
+			ShouldSucceed:  true,
 		},
 		{
 			Name:           "invalid annotation should error",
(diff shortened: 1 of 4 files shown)


pkg/managementconflict/annotations_test.go
```diff
@@ -56,7 +56,7 @@ func TestGetManagementConflictPreventionAnnotationValue(t *testing.T) {
 		t.Run(tc.Name, func(t *testing.T) {
 			obj := unstructured.Unstructured{}
 			obj.SetAnnotations(tc.Annotations)
-			policy, err := GetManagementConflictPreventionAnnotationValue(&obj)
+			policy, err := GetManagementConflictPreventionPolicy(&obj)
```
maqiuyujoyce, line 59: "ShouldSucceed" values for some test cases may need to be updated?

maqiuyujoyce (commented):
/approve

Thank you for the refactoring, @justinsb ! You'll need to rebase I think!

Just to clarify, is it our goal to remove webhooks completely, and then move all the defaulting and validation to the CRD (schema) or each individual controller?

pkg/managementconflict/annotations_test.go
justinsb: Good call, will fix - thank you!

justinsb:
> Just to clarify, is it our goal to remove webhooks completely, and then move all the defaulting and validation to the CRD (schema) or each individual controller?

We have a number of bugs when webhooks are not running when objects are created.  Anywhere we can avoid requiring a webhook, we should do so.  Two main strategies: 1) default the values in the controller and 2) make the annotation optional.

In this case, given the default value set by the webhook is the "no-op" value, it seems easier to just treat a missing annotation as the "no-op" value.  In general, that's a good pattern, because it avoids having lots of fields/annotations in the object that are saying that things are _not_ configured.

yuwenma:
Yes, agree. Is there a customer issue about this? (wondering where this fix comes form)

justinsb:
> Yes, agree. Is there a customer issue about this? (wondering where this fix comes form)
Yes, but not reported on github AFAIK.

maqiuyujoyce (commented):
/lgtm
/approve

Issue #4439, opened 2025-04-25
CloudIdentityGroup cannot adopt existing groups


rolandkool:
### Checklist

- [x] I did not find a related open enhancement request.
- [x] I understand that enhancement requests filed in the GitHub repository are by default low priority.
- [x] If this request is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Describe the feature or resource

Hi,

The `CloudIdentityGroup` resource does not seem to be able to adopt an existing Google Group. Creating a new one works fine, but when creating a `CloudIdentityGroup` resource for a group that already exists, the following error is returned:

> [
  {
    "@type": "type.googleapis.com/google.rpc.ResourceInfo",
    "description": "Error(2018): Cannot create group 'adoptgroup@myorg.com' because it already exists.",
    "owner": "domain:cloudidentity.googleapis.com",
    "resourceType": "cloudidentity.googleapis.com/Group"
  }
]

Would be useful to have adoption support otherwise, if something happens with the Kubernetes resource, the group becomes unmanageable (recreation is not really an option)

gemmahou:
Could you share the details of the existing group, and the configuration yaml file you used to acquire that group?

rolandkool:
The yaml looked like this:
```
---
apiVersion: cloudidentity.cnrm.cloud.google.com/v1beta1
kind: CloudIdentityGroup
metadata:
  name: adopt-group
spec:
  displayName: adoptgroup
  groupKey:
    id: adoptgroup@myorg.com
  parent: customers/<custid>
  labels:
    cloudidentity.googleapis.com/groups.discussion_forum: ""
```

The group I tested with has already been removed, so can't provide those exact details. But typically, our groups look like this:
```
gcloud identity groups describe somegroup@myorg.com
additionalGroupKeys:
- id: somegroup@myorg.com
- id: somegroup@myorg.com.test-google-a.com
createTime: '2020-10-09T14:00:36.657149Z'
description: Some description of groups puprose
displayName: somegroup
groupKey:
  id: somegroup@myorg.com
labels:
  cloudidentity.googleapis.com/groups.discussion_forum: ''
name: groups/<random id>
parent: customers/<customer id>
updateTime: '2025-04-29T16:21:48.825808Z'
```

gemmahou:
[...]

PR #4489, opened 2025-05-06
CloudIdentity server-generated resourceID for acquisition

gemmahou:
### Change description

Followup of customer issue https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/4439

### Tests you have done

### DIFF
--- a/apis/cloudidentity/v1beta1/group_types.go
+++ b/apis/cloudidentity/v1beta1/group_types.go
@@ -57,7 +57,7 @@ type CloudIdentityGroupSpec struct {
 	// +required
 	Parent *string `json:"parent,omitempty"`
 
-	// Immutable. The CloudIdentityGroup name. If not given, the metadata.name will be used.
+	// Immutable. Optional. The service-generated name of the resource. Format: groups/{groupID}. Used for acquisition only. Leave unset to create a new resource.
 	ResourceID *string `json:"resourceID,omitempty"`
 }
 
--- a/tests/e2e/normalize.go
+++ b/tests/e2e/normalize.go
@@ -505,9 +505,14 @@ func normalizeKRMObject(t *testing.T, u *unstructured.Unstructured, project test
 				})
 
 			case schema.GroupVersionKind{Group: "cloudidentity.cnrm.cloud.google.com", Version: "v1beta1", Kind: "CloudIdentityGroup"}:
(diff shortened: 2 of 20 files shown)


apis/cloudidentity/v1beta1/group_types.go
```diff
@@ -57,7 +57,7 @@ type CloudIdentityGroupSpec struct {
 	// +required
 	Parent *string `json:"parent,omitempty"`
 
-	// Immutable. The CloudIdentityGroup name. If not given, the metadata.name will be used.
+	// Immutable. Optional. The service-generated name of the resource. Format: groups/{groupID}. Used for acquisition only. Leave unset to create a new resource.
```
gemmahou, line 60: I'm unsure if we need to require a certain generated id format to acquire the resource, especially for backward compatibility.

tests/e2e/normalize.go
```diff
@@ -505,9 +505,14 @@ func normalizeKRMObject(t *testing.T, u *unstructured.Unstructured, project test
 				})
 
 			case schema.GroupVersionKind{Group: "cloudidentity.cnrm.cloud.google.com", Version: "v1beta1", Kind: "CloudIdentityGroup"}:
-				visitor.stringTransforms = append(visitor.stringTransforms, func(path string, s string) string {
-					return strings.ReplaceAll(s, resourceID, "${groupID}")
-				})
+				// groups/{groupID}
+				tokens := strings.Split(resourceID, "/")
```
jingyih, line 509: I'm trying to understand why we use the format "groups/groupID" for resourceID, as it's not conventional.

Is it because the existing TF controller writes "groups/groupID" to `status.name`, and that value is copied to `spec.resourceID` due to [1]?

[1] https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/575f419d12d6967374e7d2fea9871757991060e4/config/servicemappings/cloudidentity.yaml#L31

cc @maqiuyujoyce

apis/cloudidentity/v1beta1/group_types.go
jingyih: Is there no way to specify the group ID during creation? In other words, does the API always generated a ID?

tests/e2e/normalize.go
gemmahou: I believe so...

jingyih (commented):
/lgtm
/approve

PR #4494, opened 2025-05-06
Bigquerytable full test

anhdle-sso:
### Change description

Create full test for Bigquery Table

### Tests you have done


maqiuyujoyce (commented):
/lgtm
/approve

Discussed offline, the diffs between real http log and mock http log are all in the responses, and likely caused by the difference between the proto the real API is using and the mock is using. We'll clean up the mock to match the real API once there is a consistent proto.

PR #4500, opened 2025-05-07
chore: update mockgcp googleapis pin

justinsb:
Moving to 97a83d76a09a7f6dcab43675c87bdfeb5bcf1cb5


justinsb:
/assign @gemmahou 

I tried to keep the version bump in the first commit (uninteresting), and the interesting stuff is in the subsequent commits.  None of them are _that_ interesting.

I think updating the tests for new fields and adding them to the fuzzer will continue.  I think adding them to the fuzzer should trigger a sub-workstream to add those fields (I might try one, cc @jingyih and @cheftako as this relates to adding support for new fields ).

gemmahou:
/lgtm 
Thanks a lot for the investigation and fix. Just one nit comment. I'll let Jingyi/Walter to approve and merge this.

jingyih (commented):
/lgtm
/approve

I'm surprised we only needed to update two fuzzers—perhaps that's because we didn't update the Go client libraries from the GCP services.

justinsb:
> I'm surprised we only needed to update two fuzzers—perhaps that's because we didn't update the Go client libraries from the GCP services.

Yes, I think that's right.  I was going down a rabbit-hole of trying to tie everything to the go client library version in #4505 , but looks like we don't have to deal with that yet!

PR #4498, opened 2025-05-07
tests: update tests for new immutable field

justinsb:

PR #4501, opened 2025-05-07
feat: gcp iam tracker

acpana:
Mostly integrates #1958.

### DIFF
--- a/pkg/controller/iam/partialpolicy/iampartialpolicy_controller.go
+++ b/pkg/controller/iam/partialpolicy/iampartialpolicy_controller.go
@@ -37,6 +37,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/dcl/conversion"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/dcl/metadata"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/execution"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/gcpwatch"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/servicemapping/servicemappingloader"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/util"
@@ -80,7 +81,7 @@ func Add(mgr manager.Manager, deps *kontroller.Deps) error {
 
 	immediateReconcileRequests := make(chan event.GenericEvent, k8s.ImmediateReconcileRequestsBufferSize)
 	resourceWatcherRoutines := semaphore.NewWeighted(k8s.MaxNumResourceWatcherRoutines)
-	reconciler, err := NewReconciler(mgr, deps.TfProvider, deps.TfLoader, deps.DclConverter, deps.DclConfig, immediateReconcileRequests, resourceWatcherRoutines, deps.Defaulters, deps.JitterGen)
+	reconciler, err := NewReconciler(mgr, deps.TfProvider, deps.TfLoader, deps.DclConverter, deps.DclConfig, immediateReconcileRequests, resourceWatcherRoutines, deps.Defaulters, deps.JitterGen, deps.DependencyTracker)
 	if err != nil {
 		return err
 	}
@@ -88,7 +89,7 @@ func Add(mgr manager.Manager, deps *kontroller.Deps) error {
 }
 
 // NewReconciler returns a new reconcile.Reconciler.
-func NewReconciler(mgr manager.Manager, provider *tfschema.Provider, smLoader *servicemappingloader.ServiceMappingLoader, converter *conversion.Converter, dclConfig *mmdcl.Config, immediateReconcileRequests chan event.GenericEvent, resourceWatcherRoutines *semaphore.Weighted, defaulters []k8s.Defaulter, jg jitter.Generator) (*ReconcileIAMPartialPolicy, error) {
+func NewReconciler(mgr manager.Manager, provider *tfschema.Provider, smLoader *servicemappingloader.ServiceMappingLoader, converter *conversion.Converter, dclConfig *mmdcl.Config, immediateReconcileRequests chan event.GenericEvent, resourceWatcherRoutines *semaphore.Weighted, defaulters []k8s.Defaulter, jg jitter.Generator, dependencyTracker *gcpwatch.DependencyTracker) (*ReconcileIAMPartialPolicy, error) {
 	r := ReconcileIAMPartialPolicy{
 		LifecycleHandler: lifecyclehandler.NewLifecycleHandler(
 			mgr.GetClient(),
@@ -106,6 +107,10 @@ func NewReconciler(mgr manager.Manager, provider *tfschema.Provider, smLoader *s
 		},
 		requeueRateLimiter: kccratelimiter.RequeueRateLimiter(),
 		jitterGen:          jg,
+	}
+
+	if r.immediateReconcileRequests != nil && dependencyTracker != nil {
+		r.driftTracker = dependencyTracker.RegisterController(controllerName, r.immediateReconcileRequests)
 	}
 	return &r, nil
 }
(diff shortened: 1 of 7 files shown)


acpana:
cc @justinsb @cheftako

acpana:
My plan is to write a scenario test to run against the presubmits and turn that on too in our postsubmits

---

In the meantime, I've been manually testing this with our [project sample](https://cloud.google.com/config-connector/docs/reference/resource-docs/iam/iampolicy#project_level_policy)

Once we have the kccmanager up and running we can use gcloud to modify the IAMPolicy for that project and observe that the tracker picks up that update via the difference in etag and sends a message on the channel to the reconciler.

acpana:
Needs rebase on https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4498

acpana:
/hold cancel

added the presubmit 🎉

tests/e2e/testdata/scenarios/tracker/tracker-iampartialpolicy/_http03.log
acpana, line 1: As we have the `cnrm.cloud.google.com/reconcile-interval-in-seconds: "10"` defined, in the old polling style, we'd see around ~ 60/10  = 6 calls to the GCP IAM API. In the current model we only make calls when we fetch it in the tracker. With the current tracker polling interval, we'd expect no calls to be made, hence why there's none!

pkg/controller/iam/partialpolicy/iampartialpolicy_controller.go
```diff
@@ -192,6 +202,14 @@ func (r *ReconcileIAMPartialPolicy) Reconcile(ctx context.Context, request recon
 	if requeue {
 		return reconcile.Result{Requeue: true}, nil
 	}
+
+	// if we can, we will use the GCP watch method to prompt reconcile events via the immediateReconcile channel
+	if r.driftTracker != nil && len(runCtx.gcpObjects) == 1 && r.driftTracker.Add(ctx, policy, runCtx.gcpObjects[0].Spec.Etag) {
```
cheftako, line 207: If it should always be length 1, why is a list?

pkg/controller/iam/partialpolicy/iampartialpolicy_controller.go
```diff
@@ -192,6 +202,14 @@ func (r *ReconcileIAMPartialPolicy) Reconcile(ctx context.Context, request recon
...
+		log.Info("using gcp watcher instead of periodic requeue")
```
cheftako, line 208: Is this enough info?

pkg/controller/registration/registration_controller.go
```diff
@@ -241,7 +245,15 @@ func registerDefaultController(ctx context.Context, r *ReconcileRegistration, co
 		DclConverter: r.dclConverter,
 		JitterGen:    r.jitterGenerator,
 		Defaulters:   r.defaulters,
+		//DependencyTracker: r.dependencyTracker,
 	}
+
+	// todo acpana house in KCC mgr flag
+	v := os.Getenv("KCC_RECONCILE_FG")
```
cheftako, line 252: FG -> GATE -or- FLAG_GATE

pkg/gcpwatch/tracker.go
```diff
@@ -0,0 +1,240 @@
...
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/apis/iam/v1beta1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+)
+
+type DependencyTracker struct {
+	fetcher Fetcher
+
+	mutex        sync.Mutex
```
cheftako, line 31: Please name mutex according to what it looks. Eg controllersMutex

pkg/gcpwatch/tracker.go
```diff
@@ -0,0 +1,240 @@
...
+	return registration
+}
+
+type gcpResouceKey struct {
+	Kind     string
+	External string
+}
+
+type dependenciesByResource struct {
+	mutex        sync.Mutex
```
cheftako, line 68: Ditto

cheftako: Also if there is *any* chance we copy this object then this should be a pointer.

pkg/gcpwatch/tracker.go
```diff
@@ -0,0 +1,240 @@
...
+		if err := t.pollOnce(ctx); err != nil {
+			klog.Warningf("error during drift-correction polling: %v", err)
+		}
+	}
+}
+
+func (t *DependencyTracker) pollOnce(ctx context.Context) error {
+	log := klog.FromContext(ctx)
+
+	var keys []gcpResouceKey
```
cheftako, line 123: Extract this into a separate function so you can use defer unlock.

pkg/gcpwatch/tracker.go
```diff
@@ -0,0 +1,240 @@
...
+	t.mutex.Lock()
+	for key := range t.gcpResources {
+		keys = append(keys, key)
+	}
+	t.mutex.Unlock()
+
+	for _, key := range keys {
+		t.mutex.Lock()
```
cheftako, line 131: Ditto

pkg/gcpwatch/tracker.go
```diff
@@ -0,0 +1,240 @@
...
+		}
+
+		latest, err := t.fetcher.Fetch(ctx, key.Kind, key.External)
+		if err != nil {
+			// TODO: Remove if not found?
+			log.Error(err, "error fetching iam", "kind", key.Kind, "external", key.External)
+			continue
+		}
+
+		gcpResource.mutex.Lock()
```
cheftako, line 146: Ditto

pkg/gcpwatch/tracker.go
```diff
@@ -0,0 +1,240 @@
...
+
+// todo acpana expose reason why not added?
+func (r *ControllerRegistration) Add(ctx context.Context, policy *v1beta1.IAMPartialPolicy, etag string) bool {
+	log := klog.FromContext(ctx)
+
+	if etag == "" {
+		return false
+	}
+
+	if r == nil {
```
cheftako, line 174: Can r be nil?

cheftako: https://go.dev/play/p/YSRkxUiyOGk I guess it can!

pkg/gcpwatch/tracker.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	log.Info("adding watch on policy", "policy", policy, "kind", kind, "external", external)
+
+	t.add(kind, external, r, policy.GetNamespace(), policy.GetName(), etag)
+	return true
+}
+
+func (t *DependencyTracker) add(kind string, external string, controller *ControllerRegistration, namespace string, name string, etag string) {
+	target := t.getTarget(kind, external)
```
cheftako, line 202: Can we remove the lock from getTarget and move it inside of the locks we are about to call?

pkg/gcpwatch/tracker.go
```diff
@@ -0,0 +1,240 @@
...
+		return false
+	}
+
+	log.Info("adding watch on policy", "policy", policy, "kind", kind, "external", external)
+
+	t.add(kind, external, r, policy.GetNamespace(), policy.GetName(), etag)
+	return true
+}
+
+func (t *DependencyTracker) add(kind string, external string, controller *ControllerRegistration, namespace string, name string, etag string) {
```
cheftako, line 201: Maybe rename to indicate lock status?

tests/e2e/testdata/scenarios/tracker/tracker-iampartialpolicy/script.yaml
```diff
@@ -0,0 +1,56 @@
...
+apiVersion: iam.cnrm.cloud.google.com/v1beta1
+kind: IAMPartialPolicy
+metadata:
+  name: iampartialpolicy-sample-project
+  annotations:
+    cnrm.cloud.google.com/reconcile-interval-in-seconds: "10"
+spec:
+  resourceRef:
+    kind: Project
+    #name: iampartialpolicy-dep-project
```
cheftako, line 42: remove?

pkg/gcpwatch/tracker.go
acpana: I think if we always use a pointer receiver for the structs (which we are doing), we're fine against the copy lock. In my experience too, `go vet` will yell at you if you try to copy a mutex otherwise

pkg/gcpwatch/tracker.go
acpana: Actually the lock in getTarget is different from the ones below; I can do this:

```
target = t.getTargetUnderLock(kind, external)
addTargetUnderLock(...)
```

dev/ci/presubmits/tracker-tests
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+# Copyright 2024 Google LLC
```
cheftako, line 2: 2025

pkg/gcpwatch/tracker.go
```diff
@@ -0,0 +1,257 @@
...
+	return registration
+}
+
+type gcpResouceKey struct {
+	Kind     string
+	External string
+}
+
+type dependenciesByResource struct {
+	etag string
```
cheftako, line 68: Not going to block this PR but I think this is wrong. We should have a more abstract piece here. Etag does not work on any resource which we expect to be regularly updated (Eg it has something like a heartbeat). We should have something more generic and etag as the default implementation. However some resources will probably require us to hash a mask of the resource values.

yuwenma:
This breaks the master HEAD, can we revert the change or fix it?

PR #4508, opened 2025-05-09
tests:  observe IAMPartialPolicy runs

acpana:
This patch adds support for scenario based tests to just observe the system reconciling. We can then look at IAMPartialPolicy and see the number of GETs they make. This will come in handy when comparing the output with #4501

### DIFF
--- a/tests/e2e/script_test.go
+++ b/tests/e2e/script_test.go
@@ -45,6 +45,11 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+type SkippableLogEntries struct {
+	SkipCheck bool // not sent to CompareGoldenFile
+	Entries   []*test.LogEntry
+}
+
 // TestE2EScript runs a Scenario test that runs step-by-step.
 // See testdata/scenarios/README.md for more information.
 func TestE2EScript(t *testing.T) {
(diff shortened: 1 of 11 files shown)


tests/e2e/testdata/scenarios/iam/iampartialpolicy/_http03.log
acpana, line 1: tested #4501 with this and, as expected, we prove that we can cut down the number of get calls 🥳

tests/e2e/script_test.go
```diff
@@ -139,7 +147,38 @@ func TestE2EScript(t *testing.T) {
 							h.T.Logf("skipping MockGCPBackdoor command, because not running against mockgcp")
 						}
 
-						captureHTTPLogEvents()
+						captureHTTPLogEvents(false)
+						continue
+					}
+
+					// SystemRun let's KCC run for a while to observe/ gather HTTP logs
+					if obj.GroupVersionKind().Kind == "SystemRun" {
```
justinsb, line 155: Nit: Sleep might be a more self-documenting term

tests/e2e/script_test.go
```diff
@@ -139,7 +147,38 @@ func TestE2EScript(t *testing.T) {
...
+								if stopWaiting {
+									break
+								}
+							}
+
+						} else {
+							h.T.Logf("skipping MockGCPBackdoor command, because not running against mockgcp")
+						}
+
+						captureHTTPLogEvents(true)
```
justinsb, line 181: What are we expecting to see here?  I would think we would normally want to check it (i.e. not skip), because everything should be deterministic?

tests/e2e/testdata/scenarios/iam/iampartialpolicy/script.yaml
```diff
@@ -0,0 +1,55 @@
...
+    - role: roles/storage.admin
+      members:
+        - member: serviceAccount:iampartialpolicy-dep-project@iampartialpolicy-dep-project.iam.gserviceaccount.com
+    - role: roles/editor
+      members:
+        - memberFrom:
+            serviceAccountRef:
+              name: iampartialpolicy-dep-project
+---
+# Observe reconciliaiton behavior on the IAMPartialPolicy reconciler
```
justinsb, line 53: Nit: typo

justinsb (approved):

justinsb:
Some nits / suggestions but it is purely test code so
/approve
/lgtm

PR #4512, opened 2025-05-09
feat: don't re-queue waiting for deletion-defender finalizer

justinsb:
If we're waiting for the deletion-defender finalizer to be removed,
we will be requeued exactly when the finalizer has been removed,
via the watch mechanism.

There is no need to poll in this case, and doing so introduces
non-determinism.

### DIFF
--- a/pkg/controller/direct/directbase/directbase_controller.go
+++ b/pkg/controller/direct/directbase/directbase_controller.go
@@ -232,6 +232,15 @@ func (r *DirectReconciler) Reconcile(ctx context.Context, request reconcile.Requ
 	if requeue {
 		return reconcile.Result{Requeue: true}, nil
 	}
+
+	if obj.GetDeletionTimestamp() != nil {
+		if k8s.HasFinalizer(obj, k8s.DeletionDefenderFinalizerName) {
+			// Object is being deleted, but we are waiting on the finalizer
+			// When the finalizer is removed the watch will fire, so no need for time-based re-reconciliation.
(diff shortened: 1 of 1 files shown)


pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -296,8 +305,8 @@ func (r *reconcileContext) doReconcile(ctx context.Context, u *unstructured.Unst
 		}
 		if k8s.HasFinalizer(u, k8s.DeletionDefenderFinalizerName) {
 			// deletion defender has not yet finalized; requeuing
-			logger.Info("deletion defender has not yet finalized; requeuing", "resource", k8s.GetNamespacedName(u))
-			return true, nil
+			logger.Info("deletion defender has not yet finalized; cannot delete yet", "resource", k8s.GetNamespacedName(u))
```
cheftako, line 308: Earlier comment (line 307) is now wrong.

pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -296,8 +305,8 @@ func (r *reconcileContext) doReconcile(ctx context.Context, u *unstructured.Unst
 		}
 		if k8s.HasFinalizer(u, k8s.DeletionDefenderFinalizerName) {
 			// deletion defender has not yet finalized; requeuing
```
cheftako, line 307:
```suggestion
			// deletion defender has not yet finalized; allowing watch catching the finalizer removal to handle this case.
```

PR #4513, opened 2025-05-09
[WIP] chore: auto apply test-passed label to PR

jingyih:


jingyih:
/close

We could just use `-status:failure` as the PR search filter.

PR #4525, opened 2025-05-13
Check uncommitted changes before running command in a new branch

maqiuyujoyce:
### Change description

When running the same command across multiple branches, the uncommitted changes in the old branch may be checked out to the new branch, and get committed together with the changes generated for the new branch. This causes confusion. We should fail fast for situations like this.

### Tests you have done

### DIFF
--- a/experiments/conductor/cmd/runner/utilities.go
+++ b/experiments/conductor/cmd/runner/utilities.go
@@ -134,6 +134,30 @@ func cdRepoBranchDirBash(opts *RunnerOptions, subdir string, stdin io.WriteClose
 	return msg
 }
 
+func checkoutBranchWithoutUncommitedChanges(ctx context.Context, branch Branch, workDir string) {
+	log.Print("Checking uncommitted changes: git status --porcelain")
+	statusCmd := exec.CommandContext(ctx, "git", "status", "--porcelain")
+	statusCmd.Dir = workDir
+
+	results, err := execCommand(statusCmd)
+	if err != nil {
+		log.Fatal(err)
+	}
+	if results.Stdout != "" {
+		currentBranchCmd := exec.CommandContext(ctx, "git", "rev-parse", "--abbrev-ref", "HEAD")
+		currentBranchCmd.Dir = workDir
+
+		currentBranchResult, err := execCommand(currentBranchCmd)
+		if err != nil {
+			log.Fatal(err)
+		}
+		log.Fatalf("Found uncommited changes at branch %q before checking out to branch %q:\n%s\n", strings.TrimSuffix(currentBranchResult.Stdout, "\n"), branch.Local, results.Stdout)
+	}
+
+	log.Print("The branches are ready for running the command.")
+	checkoutBranch(ctx, branch, workDir)
+}
+
 func checkoutBranch(ctx context.Context, branch Branch, workDir string) {
 	checkout := exec.CommandContext(ctx, "git", "checkout", branch.Local)
 	checkout.Dir = workDir
@@ -801,7 +825,7 @@ func processBranch(ctx context.Context, opts *RunnerOptions, branch Branch, skip
 	close := setLoggingWriter(opts, branch)
 	defer close()
 
-	checkoutBranch(ctx, branch, opts.branchRepoDir)
+	checkoutBranchWithoutUncommitedChanges(ctx, branch, opts.branchRepoDir)
 
 	// Run git diff command
 	message, found := getLatestGitMessage(opts.branchRepoDir, opts, branch)


experiments/conductor/cmd/runner/utilities.go
```diff
@@ -134,6 +134,30 @@ func cdRepoBranchDirBash(opts *RunnerOptions, subdir string, stdin io.WriteClose
...
+	}
+	if results.Stdout != "" {
+		currentBranchCmd := exec.CommandContext(ctx, "git", "rev-parse", "--abbrev-ref", "HEAD")
+		currentBranchCmd.Dir = workDir
+
+		currentBranchResult, err := execCommand(currentBranchCmd)
+		if err != nil {
+			log.Fatal(err)
+		}
+		log.Fatalf("Found uncommitted changes at branch %q before checking out to branch %q:\n%s\n", strings.TrimSuffix(currentBranchResult.Stdout, "\n"), branch.Local, results.Stdout)
```
cheftako, line 154: This seems ok in the short term. I wonder if we could do better. 
```
$ git status --porcelain
?? .golangci.bck.yaml
```
we could delete anything on a line with `??`. Then reissue the porcelain status. If we still find a file then fatal. Later as we find more tags we improve the clean up code.

experiments/conductor/cmd/runner/utilities.go
```diff
@@ -801,7 +825,7 @@ func processBranch(ctx context.Context, opts *RunnerOptions, branch Branch, skip
 	close := setLoggingWriter(opts, branch)
 	defer close()
 
-	checkoutBranch(ctx, branch, opts.branchRepoDir)
+	checkoutBranchWithoutuncommittedChanges(ctx, branch, opts.branchRepoDir)
```
cheftako, line 828: Can we separate this into 2 functions? Having a clean working directory or abort function seems useful beyond this use case.

experiments/conductor/cmd/runner/utilities.go
maqiuyujoyce: Agreed there are better ways to handle uncommitted changes! My plan is to commit all the changes to the current branch when we see non empty stdout here. In my experience, there are two reasons I run into uncommitted changes at this point:
1. I forgot I made some changes to the WIP branch.
2. Command errored out half way and it didn't commit whatever half way done changes.

In both situations, there could be untracked and tracked files that I want to keep instead of deleting. So I'd prefer we just commit them as a `[Warning] Unfinished changes detected by conductor` commit, and then move on handling the next branch. What do you think?

experiments/conductor/cmd/runner/utilities.go
maqiuyujoyce: Done.

experiments/conductor/cmd/runner/utilities.go
cheftako: Conductor is meant to be a batch processing tool which can handle something going wrong and move on to the next task.
As such Fatal is not a good option, however having left over files which pollute the next branch are also not ok. 
So the normal behavior should be to try and clean up any left over files (logging that you did so is advisable).
If for you use case you want to stop or commit unexpected files that's fine for your use case.
However the existing steps should not be including unexpected files.

maqiuyujoyce: Makes sense. I added the option for the checker to work differently based on the parameter at command level.

experiments/conductor/cmd/runner/utilities.go
```diff
@@ -134,6 +134,49 @@ func cdRepoBranchDirBash(opts *RunnerOptions, subdir string, stdin io.WriteClose
 	return msg
 }
 
+func checkLocalChanges(ctx context.Context, branch Branch, workDir string, option string) {
```
cheftako, line 137: Not blocking but handleLocalChanges is better than checkLocalChanges. The former indicates a mutating call. The latter suggests a non-mutating call. (Not blocking)

experiments/conductor/cmd/runner/utilities.go
```diff
@@ -134,6 +134,49 @@ func cdRepoBranchDirBash(opts *RunnerOptions, subdir string, stdin io.WriteClose
...
+		currentBranchCmd := exec.CommandContext(ctx, "git", "rev-parse", "--abbrev-ref", "HEAD")
+		currentBranchCmd.Dir = workDir
+
+		currentBranchResult, err := execCommand(currentBranchCmd)
+		if err != nil {
+			log.Fatal(err)
+		}
+		switch option {
+		case handleLocalChangeOptionCleanUp:
+			if err := gitStash(ctx, workDir); err != nil {
```
cheftako, line 156: git stash has side effects. It will certainly get you back to clean. Not sure I want to be building up large numbers of uncommitted changes in the background. (Not blocking)

PR #4530, opened 2025-05-13
conductor: refactor first tool for easier reuse

justinsb:


justinsb:
I've noticed these look a lot like LLM functions/tasks..... wondering if we can have the LLM be the driver.

/assign @cheftako

experiments/conductor/cmd/runner/tools.go
```diff
@@ -0,0 +1,227 @@
...
+		verbose: opts.Verbose,
+	}
+	execResults, err := executeCommand(runOpts, cfg)
+	results.ExecResults = &execResults
+	results.AffectedPaths = append(results.AffectedPaths, scriptFileRelativePath)
+	return results, err
+}
+
+// createScriptYaml bridges the all-in-one conductor engine into the tool execution approach.
+func createScriptYaml(ctx context.Context, runnerOptions *RunnerOptions, branch Branch, execResults *ExecResults) ([]string, *ExecResults, error) {
```
cheftako, line 211: Do we want to continue running this as command=10 from runner? Seems like we are just adding support work by making it callable from both runner and tools?

experiments/conductor/cmd/runner/prompts/mockgcp-generate-script.txt
```diff
@@ -0,0 +1,65 @@
+I am trying to create a test case for mockgcp.
```
cheftako, line 1: Personally I would rather organize by function rather than type. Having all the mockgcp components located together makes more sense to me than keeping all the prompts located together.

experiments/conductor/cmd/runner/tools.go
justinsb: So my hypothesis is that when we have a few more tools we can have an LLM act as the orchestrator.  How about we get the mockgcp ones converted to tools, then we can see if it works?

experiments/conductor/cmd/runner/prompts/mockgcp-generate-script.txt
justinsb: I tried grouping by function and it's going to be a big PR.  I'll start chipping away at it.

For now, I moved them to be together, but it means they are all in the monopackage

PR #4535, opened 2025-05-13
BigtableAppProfile: respect ignore warning annotation in direct controller

jingyih:
The new behavior matches the TF-controller.

e.g. TF create includes "ignoreWarnings=true" when object has annotation `cnrm.cloud.google.com/ignore-warnings: "true"`.
https://github.com/jingyih/k8s-config-connector/blob/0d3bf70001e96e36bdbcf7f2bedc8f4ab06c84de/pkg/test/resourcefixture/testdata/basic/bigtable/v1beta1/bigtableappprofile/bigtableappprofile/_http.log#L267


yuwenma:
Thank you for the quick fix! It looks like `ignore-warning:true` is the default behavior, how about the "false" case? Because this is a user-configurable field, do you know what TF-based controller does if user sets it to be false?

jingyih:
> Thank you for the quick fix! It looks like `ignore-warning:true` is the default behavior, how about the "false" case? Because this is a user-configurable field, do you know what TF-based controller does if user sets it to be false?

TF-controller sets the value based on the `cnrm.cloud.google.com/ignore-warnings ` annotation. If there is no such annotation or the annotation value is set to false, TF controller will respect the value. The "magic setting" that connects the annotation to the API flag is here: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/f40549ec1beb747cc42b945c318a8dc03a373444/config/servicemappings/bigtable.yaml#L30-L31

yuwenma:
/lgtm
/approve

Thanks for the quick fix!

PR #4536, opened 2025-05-13
ci: pin gcloud

acpana:
Following up on https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4498#issuecomment-2858595050

It seems without this the HEAD is broken and some PRs on the `immutable` field being absent (!) in some other gcloud version.


justinsb:
Thanks for fixing.  We might need to figure out how to pin outside of GHA, but...

/approve
/lgtm

PR #4546, opened 2025-05-14
feat: Bigtable Logical View resource

georgecma:
### Change description

Fixes #

### Tests you have done


apis/bigtable/v1beta1/generate.sh
```diff
@@ -26,6 +26,11 @@ go run . generate-types \
   --api-version bigtable.cnrm.cloud.google.com/v1beta1  \
   --resource BigtableAppProfile:AppProfile
 
+go run . generate-types \
```
jingyih, line 29: Could you try combining the two commands into a single command and run the new command? Otherwise the generated file `types.generated.go` will be overwritten. I think our guide is out of date, I will update it.

```
go run . generate-types \
  --service google.bigtable.admin.v2 \
  --api-version bigtable.cnrm.cloud.google.com/v1beta1  \
  --resource BigtableAppProfile:AppProfile \
  --resource BigtableLogicalView:LogicalView
```

apis/bigtable/v1beta1/logicalview_reference.go
```diff
@@ -0,0 +1,83 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ refsv1beta1.ExternalNormalizer = &LogicalViewRef{}
+
+// LogicalViewRef defines the resource reference to BigtableLogicalView, which "External" field
+// holds the GCP identifier for the KRM object.
+type LogicalViewRef struct {
+	// A reference to an externally managed BigtableLogicalView resource.
+	// Should be in the format "projects/{{projectID}}/locations/{{location}}/logicalviews/{{logicalviewID}}".
```
jingyih: Replace `/logicalviews` with `/logicalViews` (capital V)

apis/bigtable/v1beta1/logicalview_identity.go
```diff
@@ -0,0 +1,117 @@
...
+
+// LogicalViewIdentity defines the resource reference to BigtableLogicalView, which "External" field
+// holds the GCP identifier for the KRM object.
+type LogicalViewIdentity struct {
+	parent *LogicalViewParent
+	id     string
+}
+
+func (i *LogicalViewIdentity) String() string {
+	return i.parent.String() + "/logicalviews/" + i.id
```
jingyih: Replace `/logicalviews` with `/logicalViews` (capital V)

apis/bigtable/v1beta1/logicalview_identity.go
```diff
@@ -0,0 +1,117 @@
...
+			Location:  location,
+		},
+		id: resourceID,
+	}, nil
+}
+
+func ParseLogicalViewExternal(external string) (parent *LogicalViewParent, resourceID string, err error) {
+	tokens := strings.Split(external, "/")
+	if len(tokens) != 6 || tokens[0] != "projects" || tokens[2] != "locations" || tokens[4] != "logicalviews" {
+		return nil, "", fmt.Errorf("format of BigtableLogicalView external=%q was not known (use projects/{{projectID}}/locations/{{location}}/logicalviews/{{logicalviewID}})", external)
```
jingyih: Same here, replace `/logicalviews` with `/logicalViews` (capital V)
Also I believe we need to fix the format. It should be `projects/{{projectID}}/instances/{{instance}}/logicalViews/logicalViewID`

apis/bigtable/v1beta1/logicalview_identity.go
```diff
@@ -0,0 +1,117 @@
...
+func (i *LogicalViewIdentity) ID() string {
+	return i.id
+}
+
+func (i *LogicalViewIdentity) Parent() *LogicalViewParent {
+	return i.parent
+}
+
+type LogicalViewParent struct {
+	ProjectID string
```
jingyih: We need to update the "Parent" for our resource logical view. The default parent "projects/projectID/locations/locationID" is true for most GCP resources, but the parent for logical view is "instance". You can find existing example here:
https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/7ed74eefe3f94bd9cf7fe57fbb2bb085264219d0/apis/bigtable/v1beta1/appprofile_identity.go#L31

apis/bigtable/v1beta1/logicalview_identity.go
```diff
@@ -0,0 +1,117 @@
...
+	Location  string
+}
+
+func (p *LogicalViewParent) String() string {
+	return "projects/" + p.ProjectID + "/locations/" + p.Location
```
jingyih: This needs to be updated once the parent is updated.

apis/bigtable/v1beta1/logicalview_identity.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// New builds a LogicalViewIdentity from the Config Connector LogicalView object.
+func NewLogicalViewIdentity(ctx context.Context, reader client.Reader, obj *BigtableLogicalView) (*LogicalViewIdentity, error) {
+
+	// Get Parent
+	projectRef, err := refsv1beta1.ResolveProject(ctx, reader, obj.GetNamespace(), obj.Spec.ProjectRef)
```
jingyih: Same here. We need to fetch the parent instance, as opposed to project and location.

jingyih (commented):
Let's not worry about the controller for now. Let's get the go types and mapping functions merged first!

apis/bigtable/v1beta1/logicalview_types.go
```diff
@@ -21,11 +21,24 @@ import (
...
+	// +required
+	InstanceRef *InstanceRef `json:"instanceRef,omitempty"`
+}
+
+// BigtableLogicalViewSpec defines a state of BigtableLogicalView
 // +kcc:proto=google.bigtable.admin.v2.LogicalView
 type BigtableLogicalViewSpec struct {
 	// The BigtableLogicalView name. If not given, the metadata.name will be used.
-	ResourceID *string `json:"resourceID,omitempty"`
+	Name *string `json:"resourceID,omitempty"`
```
jingyih, line 33: Once we updated the "generate-types" command and run it, it should add something like the following to the generated types files. It is recommended that we just copy all the fields over. If a field is supposed to be configured by user, copy it to `BigtableLogicalViewSpec`, otherwise if a field is generated by Bigtable service, copy it to `BigtableLogicalViewObservedState`.

```
// +kcc:proto=google.bigtable.admin.v2.LogicalView
type LogicalView struct {
	// Identifier. The unique name of the logical view.
	//  Format:
	//  `projects/{project}/instances/{instance}/logicalViews/{logical_view}`
	// +kcc:proto:field=google.bigtable.admin.v2.LogicalView.name
	Name *string `json:"name,omitempty"`

	// Required. The logical view's select query.
	// +kcc:proto:field=google.bigtable.admin.v2.LogicalView.query
	Query *string `json:"query,omitempty"`

	// Optional. The etag for this logical view.
	//  This may be sent on update requests to ensure that the client has an
	//  up-to-date value before proceeding. The server returns an ABORTED error on
	//  a mismatched etag.
	// +kcc:proto:field=google.bigtable.admin.v2.LogicalView.etag
	Etag *string `json:"etag,omitempty"`

	// Optional. Set to true to make the LogicalView protected against deletion.
	// +kcc:proto:field=google.bigtable.admin.v2.LogicalView.deletion_protection
	DeletionProtection *bool `json:"deletionProtection,omitempty"`
}
```

apis/bigtable/v1beta1/logicalview_types.go
```diff
@@ -40,13 +53,20 @@ type BigtableLogicalViewStatus struct {
...
+
 	// ObservedState is the state of the resource as most recently observed in GCP.
 	ObservedState *BigtableLogicalViewObservedState `json:"observedState,omitempty"`
 }
 
 // BigtableLogicalViewObservedState is the state of the BigtableLogicalView resource as most recently observed in GCP.
 // +kcc:proto=google.bigtable.admin.v2.LogicalView
 type BigtableLogicalViewObservedState struct {
+	// The observed state of the LogicalView object.
+	Spec BigtableLogicalViewSpec `json:"spec,omitempty"`
```
jingyih, line 69: ObservedState should not include a field if it is only supposed to be configured by user.

apis/bigtable/v1beta1/generate.sh
jingyih: Oops, I totally overlooked the fact that this is a brand new resource we are adding. In this case, we should add it as "v1alpha1". So instead of combining the commands, we should run:

```
go run . generate-types \
 --service google.bigtable.admin.v2 \
 --api-version bigtable.cnrm.cloud.google.com/v1alpha1 \
 --resource BigtableLogicalView:LogicalView
```

georgecma: The resource is similar to appprofile in that it uses resources from v1beta1, so I thought it better to group them in the same package. Should I import from the beta package or keep this resource in the beta package?

apis/bigtable/v1beta1/logicalview_reference.go
georgecma: Done

apis/bigtable/v1beta1/logicalview_identity.go
georgecma: Done

apis/bigtable/v1beta1/logicalview_identity.go
georgecma: Done

apis/bigtable/v1beta1/logicalview_identity.go
georgecma: Done

apis/bigtable/v1beta1/logicalview_identity.go
georgecma: Code is outdated.

apis/bigtable/v1beta1/generate.sh
georgecma: Discussed offline, moved to v1alpha1

apis/bigtable/v1beta1/logicalview_types.go
georgecma: Would this be what GetLogicalView gives? If so, then it'll be the LogicalView object (which is almost entirely configured by the user).

jingyih: A field should be placed in `BigtableLogicalViewSpec` if it is user-configurable. If the field is not configurable by the user and its value is instead generated by the GCP API, it should be included in `BigtableLogicalViewObservedState`. In the current PR, `BigtableLogicalViewSpec` is incorrectly nested under `BigtableLogicalViewObservedState`.

Usually, a service generated field will also have something like "Output only" in the field description, or output_only proto annotation.

apis/bigtable/v1alpha1/types.generated.go
```diff
@@ -12,244 +12,35 @@
...
 // limitations under the License.
 
-package v1alpha1
-
-import (
-	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
-)
-
-// +kcc:proto=google.bigtable.admin.v2.AutoscalingLimits
-type AutoscalingLimits struct {
```
jingyih: Oops, the removal of existing types was an unintended side effect. I'm modernizing the generated types for Bigtable v1alpha1 in #4700. If you pick up #4700 and run the following command, the changes should be additive only.

```
go run . generate-types \
  --service google.bigtable.admin.v2 \
  --api-version bigtable.cnrm.cloud.google.com/v1alpha1  \
  --resource BigtableAuthorizedView:AuthorizedView \
  --resource BigtableBackup:Backup \
  --resource BigtableCluster:Cluster \
  --resource BigtableLogicalView:LogicalView
```

apis/bigtable/v1beta1/logicalview_types.go
georgecma: Removed the Spec from the ObservedState.

pkg/controller/direct/bigtable/mapper.generated.go
```diff
@@ -148,6 +148,50 @@ func BigtableAuthorizedViewObservedState_ToProto(mapCtx *direct.MapContext, in *
...
+		return nil
+	}
+	out := &pb.LogicalView{}
+	// MISSING: Name
+	// MISSING: Query
+	// MISSING: Etag
+	// MISSING: DeletionProtection
+	return out
+}
+func BigtableLogicalViewSpec_FromProto(mapCtx *direct.MapContext, in *pb.LogicalView) *krmv1beta1.BigtableLogicalViewSpec {
```
jingyih: nit: remove this from `mapper.generate.go` since it is defined in the `logicalview_mappings.go`.

apis/bigtable/v1beta1/logicalview_types.go
```diff
@@ -0,0 +1,103 @@
...
+package v1beta1
+
+import (
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/apis/k8s/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var BigtableLogicalViewGVK = GroupVersion.WithKind("BigtableLogicalView")
+
+type BigtableLogicalViewParent struct {
```
jingyih: Let's revert all the codes to the `v1beta1` folder for now? The resource needs to be added as pure v1alpha1 as the first step.

apis/bigtable/v1alpha1/logicalview_types.go
```diff
@@ -0,0 +1,101 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// BigtableLogicalViewList contains a list of BigtableLogicalView
+type BigtableLogicalViewList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []BigtableLogicalView `json:"items"`
+}
+
+func init() {
+	SchemeBuilder.Register(&BigtableLogicalView{}, &BigtableLogicalViewList{})
```
jingyih: Rerun the `dev/tasks/generate-crds` script should fix the "DeepCopyObject" errors.

jingyih:
259 commit seems strange. Could you try something like:
```
git fetch upstream
git rebase upstream/master
```
(assuming upstream is "https://github.com/GoogleCloudPlatform/k8s-config-connector")

apis/bigtable/v1alpha1/logicalview_types.go
georgecma: reran it

apis/bigtable/v1beta1/logicalview_types.go
georgecma: Removed all v1beta1 references

pkg/controller/direct/bigtable/mapper.generated.go
georgecma: It's removed

georgecma:
This PR is obsolete, I cleaned up the commit history and moved the changes to this new PR: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4725

PR #4528, opened 2025-05-13
tests: add golden tests for operator

justinsb:
- **Simple golden tests**
  

- **tests: add tests for CCC operator**

PR #4547, opened 2025-05-14
restore release version 1.126.0

xiaoweim:
### Change description

Fixes #

### Tests you have done


justinsb:
A few things @cheftako pointed out to me:

1) We don't want a "gap" of versions, we should have at least each minor version in the window available here (c.f. https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/2a5c056279d917232bc9deaa1f0ce02da025bb5f/operator/autopilot-channels/packages/configconnector )  Whether we want to support minor versions IDK, I would guess any versions we shipped is going to be the easiest policy

2) We should be careful about whether downgrade is safe.  You're not introducing this here so we can likely treat this as a separate issue.  Walter was worried about ClusterRoles / Roles; I think I'm more worried about CRD removal or field removal (via downgrade).  We're introducing more testing into the operator so I think we can deal with this there ( #4528 )

yuwenma:
Could you share more context about this change? From the discussion, my read as we want to keep all the passed releases until the latest one in GKE add-on, which means the manifests contain 1.126.0, 1.127.0, 1.128.0, 1.129.2, 1.130.0, 1.131.0.

xiaoweim:
I just updated the PR to not leave a gap in between older versions and restored all the shipped versions up until 1.126.0

jingyih (commented):
I am wondering if we could only add back the `per-namespace-components.yaml
` files from the previous verions?

It looks like we are only using `version` from CCC to load per namespace components. [1]

This way it is more clear that the rollback does not include CRDs and ClusterRoels.

[1] https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/3132f8dfb6ab9b2110a827e37d3ee24c432c141f/operator/pkg/manifest/per_namespace_manifest_loader.go#L63

xiaoweim:
> I am wondering if we could only add back the `per-namespace-components.yaml ` files from the previous verions?
> 
> It looks like we are only using `version` from CCC to load per namespace components. [1]
> 
> This way it is more clear that the rollback does not include CRDs and ClusterRoels.
> 
> [1]
> 
> https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/3132f8dfb6ab9b2110a827e37d3ee24c432c141f/operator/pkg/manifest/per_namespace_manifest_loader.go#L63

I think that is a good idea. Let me update the PR!

jingyih (commented):
I just realized we also have 1.129 and 1.130 folders that include everything. For consistency, we should either remove the CRDs and ClusterRoles from those folders or add them back for every version. Keeping only the per-namespace file clarifies what happens during a downgrade, but it may confuse users browsing the release folders as they'll see only one file. I’ll defer to @yuwenma  on what we want to include in each version.

yuwenma:
My understanding is that we need to contain both 0-cnrm-system.yaml and per-namespace-components.yaml for each namespace version. Because user can choose a stale release as their very first installation in which case the namespace and RBAC will be missing. For example, user choose to use the 1.132 bundle to specifically install 1.126 (via CCC `spec.version`), there could be a chance that only the ./packages/1.126 are applied, in which case the 0-cnrm-system.yaml will be missing. Or does the CCC/CC specifically install this file in some other steps?

xiaoweim:
> My understanding is that we need to contain both 0-cnrm-system.yaml and per-namespace-components.yaml for each namespace version. Because user can choose a stale release as their very first installation in which case the namespace and RBAC will be missing. For example, user choose to use the 1.132 bundle to specifically install 1.126 (via CCC `spec.version`), there could be a chance that only the ./packages/1.126 are applied, in which case the 0-cnrm-system.yaml will be missing. Or does the CCC/CC specifically install this file in some other steps?

My understanding is that when we have specify the version field in the CCC, we load the version specific manifest but only looking for `packages/${version}/namespaced/per-namespace-components.yaml` https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/operator/pkg/manifest/repo.go#L115. 

I did not quite get where the user chooses to use 1.132.0 bundle to install 1.126.0. IIUC if the user install KCC 1.132 version first, and then update the CCC Version field to 1.126.0, their cluster should already have CRD and RBAC permissions installed and only loads the `/per-namespace-components.yaml` manifest.

xiaoweim:
I have just updated the PR to include the current CRD and RBAC into previous KCC release versions with old operators.

operator/channels/packages/configconnector/1.130.2/namespaced/0-cnrm-system.yaml
yuwenma, line 1: Mind elaborating why we change the 1.130.2 annotation to 1.131.0?  cc @jingyih

xiaoweim: Yes, last week I had a discussion with @cheftako that for older versions, we only keep the stateful set of that version and re-use the latest version manifests for CRDs, RBACs, etc.

yuwenma: Suggest modifying the propose-tag script as well  (which supposes to manage the non up-to-date release versions), otherwise it would just be a one-off change that applies for 1.131 only.

operator/autopilot-channels/packages/configconnector/1.129.2/crds.yaml
cheftako, line 1: This CRDS file seems to have a different size than the others. That seems wrong. Can you check why?

operator/autopilot-channels/packages/configconnector/1.130.2/crds.yaml
```diff
@@ -10562,8 +10562,6 @@ spec:
                 description: Required. The configured value of the enablement at the
                   given resource.
                 type: string
-              location:
```
cheftako, line 10565: This is a backward incompatible change. Its on a alpha resource, so we can get away with it but we shouldn't be making these sort of changes lightly.

xiaoweim: Yes this looks like a backward incompatible change. Thanks for pointing it out!

PR #4548, opened 2025-05-14
Increase test coverage for IAMServiceAccount

maqiuyujoyce:
### Change description

Generated test cases to increase test coverage for IAMServiceAccount

### Tests you have done

### DIFF
--- a/tests/e2e/unified_test.go
+++ b/tests/e2e/unified_test.go
@@ -210,7 +210,7 @@ func testFixturesInSeries(ctx context.Context, t *testing.T, testPause bool, can
 					case "bigtable.cnrm.cloud.google.com",
 						"orgpolicy.cnrm.cloud.google.com":
 						// Use SSA
-
+					case "iam.cnrm.cloud.google.com":
 					default:
 						t.Logf("not yet using SSA for create of resources in group %q", group)
 						opt.DoNotUseServerSideApplyForCreate = true
--- a/tests/e2e/unified_test.go
+++ b/tests/e2e/unified_test.go
@@ -206,14 +206,17 @@ func testFixturesInSeries(ctx context.Context, t *testing.T, testPause bool, can
 					}
 
 					// We want to use SSA everywhere, but some of our tests are broken by SSA
+					kind := primaryResource.GetObjectKind().GroupVersionKind().Kind
 					switch group := primaryResource.GetObjectKind().GroupVersionKind().Group; group {
 					case "bigtable.cnrm.cloud.google.com",
 						"orgpolicy.cnrm.cloud.google.com":
 						// Use SSA
 					case "iam.cnrm.cloud.google.com":
+						if kind != "IAMServiceAccount" {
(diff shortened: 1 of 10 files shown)


maqiuyujoyce:
The original test was generated based on https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4458.

maqiuyujoyce:
@jingyih could you lgtm and approve when you get a chance? I got lgtm from @anhdle-sso but had to rebase.

tests/e2e/unified_test.go
```diff
@@ -206,14 +206,17 @@ func testFixturesInSeries(ctx context.Context, t *testing.T, testPause bool, can
 					}
 
 					// We want to use SSA everywhere, but some of our tests are broken by SSA
+					kind := primaryResource.GetObjectKind().GroupVersionKind().Kind
 					switch group := primaryResource.GetObjectKind().GroupVersionKind().Group; group {
 					case "bigtable.cnrm.cloud.google.com",
 						"orgpolicy.cnrm.cloud.google.com":
 						// Use SSA
-
+					case "iam.cnrm.cloud.google.com":
```
jingyih, line 214: nit: maybe clearer if we just do inline switch on "kind"?

```
  case "iam.cnrm.cloud.google.com":
	switch primaryResource.GetObjectKind().GroupVersionKind().Kind {
	case "IAMServiceAccount":
```

jingyih (commented):
/approve

tests/e2e/unified_test.go
maqiuyujoyce: Yes, I'm thinking of doing that after these PRs are merged. It'll cause a lot of conflicts if I'm working on the test data one by one so I plan to add all the kinds once every weekish and then clean up the code here.

jingyih: @yuwenma mentioned in #4558 that we do not want to keep a list like this due to maintenance overhead. Let's figure out a better way to do this. Put it under `ResourceContext`?

maqiuyujoyce: Hmmm, let's address it in a separate PR then. It's beneficial to use SSA but not strictly required at the moment.

maqiuyujoyce: Regarding where to put it, `ResourceContext` doesn't seem to be the best place to configure this value. A couple of reasons:
1. `ResourceContext` contains test case level information and don't have a kind level or group level defaulting.
2. `TestAllInSeries` has a different test option struct (`create.CreateDeleteTestOptions`).

I don't think we want to keep the list, but in the long term I don't think the list is necessary. So I'd prefer we do some hacky and easy thing to solve this problem. For example, create a usessa folder under tests and whenever we allowlist a group or a kind, we add an empty file of that name in the folder.

Or, let's just enable SSA for all kinds and see what's going on.

jingyih (commented):
/lgtm
/approve

PR #4554, opened 2025-05-15
preview environment: don't check for duplicate controller names

justinsb:
It probably isn't a particularly important test in the first place,
given that names are primarily for metrics, but it shouldn't cause
a preview operation to fail.


justinsb:
ping @cheftako

ldanielmadariaga:

PR #4558, opened 2025-05-15
test: enable SSA for more API groups

jingyih:
We should eventually switch to using "exclusive list" instead of "allowlist".


pkg/test/resourcefixture/testdata/basic/backupdr/v1alpha1/backupdrbackupplanassociation/backupdrbackupplanassociation-minimal/create.yaml
```diff
@@ -20,7 +20,6 @@ spec:
   projectRef:
     external: ${projectId}
   location: us-central1
-  description: "Initial description"
```
yuwenma, line 23: Is this intended? we don't want to remove field from test coverage

tests/e2e/unified_test.go
```diff
@@ -208,7 +208,12 @@ func testFixturesInSeries(ctx context.Context, t *testing.T, testPause bool, can
 					// We want to use SSA everywhere, but some of our tests are broken by SSA
 					switch group := primaryResource.GetObjectKind().GroupVersionKind().Group; group {
 					case "bigtable.cnrm.cloud.google.com",
-						"orgpolicy.cnrm.cloud.google.com":
+						"orgpolicy.cnrm.cloud.google.com",
+						"backupdr.cnrm.cloud.google.com",
+						"datastream.cnrm.cloud.google.com",
+						"gkebackup.cnrm.cloud.google.com",
+						"speech.cnrm.cloud.google.com",
+						"vmwareengine.cnrm.cloud.google.com":
```
yuwenma, line 216: This would increase the maintenance overhead more and more. Can we at least keep the list in its own file? Another question, how did you get the list?

yuwenma (commented):
/approve
/hold

Left some comments. Can you resolve them in this PR or leave a guide comment?

pkg/test/resourcefixture/testdata/basic/backupdr/v1alpha1/backupdrbackupplanassociation/backupdrbackupplanassociation-minimal/create.yaml
jingyih: Yeah, the nice thing about SSA is it will flag fields that are invalid. This was overlooked.

tests/e2e/unified_test.go
jingyih: These are just the API groups that I recently worked on, so I am sort of familiar with the resources and test cases.

jingyih: I think we can keep adding the API groups and switch to use "exclude list" when we have most API groups enabled on SSA.

PR #4560, opened 2025-05-15
Fix: OrgPolicyPolicy and OrgPolicyCustomConstraint

jingyih:
`HEAD` is broken as the result of an accidental merge of #3856.
This PR trie to fix it.


apis/orgpolicy/v1alpha1/policy_types.go
```diff
@@ -84,12 +84,21 @@ type OrgPolicyPolicyStatus struct {
 // OrgPolicyPolicyObservedState is the state of the OrgPolicyPolicy resource as most recently observed in GCP.
 // +kcc:proto=google.cloud.orgpolicy.v2.Policy
 type OrgPolicyPolicyObservedState struct {
+	// Basic information about the Organization Policy.
+	// +kcc:proto:field=google.cloud.orgpolicy.v2.Policy.spec
+	Spec *PolicySpecObservedState `json:"spec,omitempty"`
```
acpana, line 89: Could we call this PolicySpec in the future? Non blocker, thanks for fixing HEAD 🙏🏼

jingyih: Yeah I found this confusing as well. Maybe we don't follow the naming of proto field for this one.

jingyih:
> Sg. This looks like a new code change, why does it fix the master HEAD? Or is it a code reversion due to merge conflict?

It's a GitHub race. PR #3856 was accidentally merged immediately after a rebase was pushed. Since it already had the `lgtm` label, it was merged before the tests could start.

PR #4565, opened 2025-05-16
Add direct ServiceIdentity controller, opt-in except for magic gserviceaccount P4SA

justinsb:
* feat: Add optin direct ServiceIdentity controller

Create an opt-in ServiceIdentity controller

*  e2e: add tests for direct types

Add testing of direct controllers alongside legacy controllers

*  chore: create shared project number/id mapper

Refactor the project id resolver into shared code

* ServiceIdentity: support cloudservices.gserviceaccount.com

Recognize the special-case MIG P4SA, force it to direct with a reconcile gate

### DIFF
--- a/dev/ci/presubmits/tests-e2e-direct-optin
+++ b/dev/ci/presubmits/tests-e2e-direct-optin
@@ -1,0 +1,26 @@
+#!/bin/bash
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+cd ${REPO_ROOT}/
+
+export ONLY_TEST_KINDS=ServiceIdentity
+export KCC_USE_DIRECT_RECONCILERS=ServiceIdentity
+
+${REPO_ROOT}/dev/tasks/run-e2e
--- a/pkg/controller/direct/serviceusage/serviceidentity_controller.go
+++ b/pkg/controller/direct/serviceusage/serviceidentity_controller.go
@@ -26,11 +26,13 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/GoogleCloudPlatform/k8s-config-connector/apis/common/projects"
 	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/serviceusage/v1beta1"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/config"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
+	kccpredicate "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/predicate"
 
 	gcp "google.golang.org/api/serviceusage/v1beta1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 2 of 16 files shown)


pkg/test/utils.go
```diff
@@ -236,6 +236,9 @@ func CompareGoldenObject(t *testing.T, p string, got []byte) {
 	t.Errorf("FAIL: unexpected diff in %s: %s", p, diff)
 
 	if writeGoldenOutput {
+		if err := os.MkdirAll(filepath.Dir(p), 0755); err != nil {
+			t.Fatalf("failed to create directories for golden output file %s: %v", p, err)
+		}
```
yuwenma: Is this possible? The bare minimum for a test suite to run is to have the directory exist with a create.yaml file in it.

pkg/controller/direct/serviceusage/serviceidentity_controller.go
```diff
@@ -0,0 +1,244 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// GoogleApisServiceAgent was historically used by multiple services, but for more granular permission control we moved to P4SAs instead.
+// MIGs were tricky to move though, and so instead of switching to a P4SA, they adopted the GoogleApisServiceAgent as their P4SA
+// We still want to support this though, so we recognize this "magic value"
+const GoogleApisServiceAgent = "cloudservices.gserviceaccount.com"
```
yuwenma, line 50: This is such a mess... Is this the only special value for `spec.resourceID`?

pkg/controller/direct/serviceusage/serviceidentity_controller.go
```diff
@@ -0,0 +1,244 @@
...
+	// This resource type does not have a direct GETtable URL in the traditional sense.
+	return nil, nil
+}
+
+type serviceIdentityAdapter struct {
+	projectMapper *projects.ProjectMapper
+	gcpBeta       *gcp.APIService
+
+	id      *krm.ServiceIdentityIdentity
+	desired *krm.ServiceIdentity
```
yuwenma, line 126: Suggest using `krm.ServiceIdentitySpec`, otherwise line 138 is really confusing and could be mistakenly changed (maintenance overhead).

pkg/controller/direct/serviceusage/serviceidentity_controller.go
```diff
@@ -0,0 +1,244 @@
...
+		for {
+			if op.Done {
+				identity := &gcp.GoogleApiServiceusageV1beta1ServiceIdentity{}
+				if err := json.Unmarshal(op.Response, identity); err != nil {
+					return fmt.Errorf("error parsing response: %w", err)
+				}
+				serviceIdentity = identity
+				break
+			}
+			time.Sleep(2 * time.Second)
```
yuwenma, line 198: Oh.. does it not have a `os.Wait`?  This feels like we'd better set an expiration as well in case the server introduces an endless Create.

yuwenma (commented):
/approve

Left some comments. I see many exceptions on this resource, but fortunately the spec only has `resourceID` and `projectRef`, which hides less corner case on reconcile gate and default to scifi.

pkg/controller/direct/serviceusage/serviceidentity_controller.go
justinsb: The only one so far :-)

pkg/test/utils.go
justinsb: Hopefully not, which is why Fatalf

pkg/controller/direct/serviceusage/serviceidentity_controller.go
justinsb: No op.Wait (and certainly no op.WaitWithTimeout).  Will add a timeout to the poll.

pkg/controller/direct/serviceusage/serviceidentity_controller.go
justinsb: So we need the status field.  It is confusing, but it is a weird object.  I think our tests are going to have to guard this one.

justinsb:
We also need to decide how we want to golden-test this, because the results from direct are a little different from TF (not very different, but different enough that we probably can't normalize it away)

dev/ci/presubmits/tests-e2e-direct-optin
```diff
@@ -0,0 +1,26 @@
...
+set -o nounset
+set -o pipefail
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+cd ${REPO_ROOT}/
+
+export ONLY_TEST_KINDS=ServiceIdentity
+export KCC_USE_DIRECT_RECONCILERS=ServiceIdentity
+
+${REPO_ROOT}/dev/tasks/run-e2e
```
justinsb, line 26: Maybe we just drop this for now, and tackle testing with direct and non-direct in a different PR

ldanielmadariaga:

PR #4569, opened 2025-05-17
chore: modernize unmanageddetector controller & test

justinsb:
- **chore: modernize unmanageddetector test**
  We can use envtest and avoid a lot of the setup complexity.

PR #4570, opened 2025-05-17
chore: cleanup unmanageddetector controller

justinsb:
* Fix logging to use context
* We only care about the GVK, so take that as parameter.


justinsb:
I have combined this one into #4569, I think they are hard to merge separately

PR #4573, opened 2025-05-19
Add types for AnywhereCache resource 

googlyrahman:
### Change description

1. Updates storage goSDK version to 1.53.0 from 1.50.0
2. Add types for AnywhereCache Resource.
3. Add kube-builder tags for resource defaults.


apis/storage/v1alpha1/anywherecache_types.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// Cache entry TTL (ranges between 1h to 7d). This is a cache-level config
+	//  that defines how long a cache entry can live. Defaults to "86400s"
+	//  TTL must be in whole seconds.
+	Ttl *string `json:"ttl,omitempty"`
+
+	// Cache admission policy. Valid values includes:
+	//  `admit-on-first-miss` and `admit-on-second-miss`. Defaults to
+	//  `admit-on-first-miss`.
+	AdmissionPolicy *string `json:"admissionPolicy,omitempty"`
```
acpana: these 3 fields all have defaults so they will never be "unset" effectively. Let's define default values via kubebuilder tags https://book.kubebuilder.io/reference/markers/crd-validation

acpana (commented):
one small comment on defaulting with kube tags but otherwise LGTM!

PR #4574, opened 2025-05-19
tests: separate out "legacy" normalization

justinsb:
We can whittle it down from here


tests/e2e/normalize_legacy.go
```diff
@@ -0,0 +1,728 @@
...
+	"time"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/config/tests/samples/create"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test"
+	testgcp "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test/gcp"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// LegacyNormalize is the "legacy" normalization steps,
+// we should avoid adding to this function and instead add to the per-service normalization functions.
```
yuwenma, line 17: In general people ignore comments when making code changes. I suggest we add a presubmit test that fails if this file is modified.

justinsb: No objection to adding a test, but it just feels like something that will get in our way.  We do want to modify it as we _remove_ functionality from this function.

I added a "Deprecated" comment which should mark it as such in IDEs.

PR #4575, opened 2025-05-19
mockgcp: move sql to normalizer pattern

justinsb:


justinsb:
This was done in #4600

PR #4578, opened 2025-05-19
direct: Add controller BigQueryTable

anhdle-sso:
### Change description

- BQTable controller (opt-in by annotation or having spec.labels field)
- Custom compare diff logic for update.
- Minor mapper change.
- Minor mockgcp table change.
- Duplicate test cases for direct controller (opt-in direct annotation added).
- Bug fix for table identity.
- Remove spec.Labels for bigquerytable-full test to run TF controller.

After registering the direct controller, the POST call with getIamPolicy disappear from the log. This happen to the TF test cases. It means the difference comes from the base direct controller. The POST call is only doing a GET IAM call for BQ Table so the harm is minimal.

The commit [record realgcp log](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4578/commits/29115094633ac903e6a819e8f485b5a9d4dbd4fc) has the realgcp log of existing test cases running TF controller.

The commit [real gcp log for direct](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4578/commits/236799d815d84411817c9fa04af3ffbb94dac535) has the realgcp log of the same test running direct controller.

The diff between these two commits:

- TF uses metadata.labels vs direct use spec.labels cause diff in golden output and object.
- GetIamPolicy call in golden log.

### Tests you have done


pkg/controller/direct/bigquery/table_compare.go
```diff
@@ -0,0 +1,308 @@
...
+			return fmt.Errorf("schema contains nil field")
+		}
+		if field.Name == "" {
+			return fmt.Errorf("field contains empty field name")
+		}
+	}
+	return nil
+}
+
+func policyTagsEq(a, b *bigquery.TableFieldSchemaPolicyTags) bool {
```
maqiuyujoyce: Nit: Maybe use the full word "Equal" in the function names for better readability?

pkg/controller/direct/bigquery/table_compare.go
```diff
@@ -0,0 +1,308 @@
...
+	if a.Names == nil && b.Names != nil && len(b.Names) == 0 {
+		return true
+	}
+
+	// Due to inconsistent API behavior, suppress empty array and nil.
+	if b.Names == nil && a.Names != nil && len(a.Names) == 0 {
+		return true
+	}
+
+	if a.Names == nil || b.Names == nil {
```
maqiuyujoyce, line 48: Nit: Line 48 to line 60 can be removed for simplicity and possibly better readability.

`a.Names == nil || b.Names == nil`, `a.Names == nil && b.Names != nil && len(b.Names) == 0`, and `b.Names == nil && a.Names != nil && len(a.Names) == 0` give the same result of `len(a.Names) != len(b.Names)`, because len(nil slice) = 0.

pkg/controller/direct/bigquery/table_compare.go
```diff
@@ -0,0 +1,308 @@
...
+		return false, err
+
+	}
+	if len(desired) != len(actual) {
+		return false, nil
+	}
+	sortSchemaFields(desired)
+	sortSchemaFields(actual)
+	for i := range desired {
+		if !reflect.DeepEqual(desired[i].Categories, actual[i].Categories) {
```
maqiuyujoyce, line 94: Can we just do
```
d := desired[i].DeepCopy()
a := actual[i].DeepCopy()

// Some comments
d.Fields = nil
a.Fields = nil
d.PolicyTags = nil
a.PolicyTags = nil

!reflect.DeepEqual(d, a) {
return false, nil
}

if !policyTagsEq(desired[i].PolicyTags, actual[i].PolicyTags) {
    return false, nil
}
eq, err := tableFieldsSchemaEq(desired[i].Fields, actual[i].Fields)
...
```
I think this provides simplicity, better readability and emphasizes custom logic in the comparison. Right now it's a bit hard to figure out what's special about this comparison function.

pkg/controller/direct/bigquery/table_compare.go
```diff
@@ -0,0 +1,308 @@
...
+}
+
+func materializedViewEq(a, b *bigquery.MaterializedViewDefinition) bool {
+	if a == nil && b == nil {
+		return true
+	}
+	if a == nil || b == nil {
+		return false
+	}
+	if !reflect.DeepEqual(a.AllowNonIncrementalDefinition, b.AllowNonIncrementalDefinition) {
```
maqiuyujoyce, line 211: Similar comments to all the other comparison logic.

pkg/controller/direct/bigquery/table_controller.go
```diff
@@ -0,0 +1,351 @@
...
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &obj); err != nil {
+		return nil, fmt.Errorf("error converting to %T: %w", obj, err)
+	}
+
+	id, err := krm.NewTableIdentity(ctx, reader, obj)
+	if err != nil {
+		return nil, err
+	}
+
+	projectID := id.Parent().ProjectID
```
maqiuyujoyce: Why would krm.NewTableIdentity() returns an id with projectID == ""? Should this be checked within NewTableIdentity() function?

If project == "" check is done here, then should we also do the dataset == "" check?

pkg/controller/direct/bigquery/table_mapper.go
```diff
@@ -43,8 +43,18 @@ func MaterializedViewDefinition_ToProto(mapCtx *direct.MapContext, in *krmv1beta
 	out := &pb.MaterializedViewDefinition{}
 	out.Query = direct.ValueOf(in.Query)
 	// MISSING: LastRefreshTime
-	out.EnableRefresh = direct.ValueOf(in.EnableRefresh)
-	out.RefreshIntervalMs = direct.ValueOf(in.RefreshIntervalMs)
+	// if spec.materialized_view.enable_refresh is not set. Default to true.
+	if in.EnableRefresh == nil {
+		out.EnableRefresh = true
```
maqiuyujoyce, line 48: Do we have a test case to cover this?

pkg/controller/direct/bigquery/table_mapper.go
```diff
@@ -43,8 +43,18 @@ func MaterializedViewDefinition_ToProto(mapCtx *direct.MapContext, in *krmv1beta
...
+	} else {
+		out.EnableRefresh = direct.ValueOf(in.EnableRefresh)
+	}
+	// if spec.materialized_view.refresh_interval_ms is not set. Default to 1800000 (30 mins).
+	if in.RefreshIntervalMs == nil {
+		out.RefreshIntervalMs = 1800000
```
maqiuyujoyce, line 54: Similarly, for all the fields with custom default values in the mapper, do we all have test coverage?

pkg/controller/direct/bigquery/table_mapper.go
```diff
@@ -141,8 +171,7 @@ func ViewDefinition_FromProto(mapCtx *direct.MapContext, in *pb.ViewDefinition)
 	out := &krmv1beta1.ViewDefinition{}
 	out.Query = direct.LazyPtr(in.Query)
 	// MISSING: UserDefinedFunctionResources
-	// MISSING: UseLegacySQL
-	// (near miss): "UseLegacySQL" vs "UseLegacySql"
+	out.UseLegacySql = direct.LazyPtr(in.UseLegacySql)
```
maqiuyujoyce, line 174: Nit: I know this PR is not touching the type, but `UseLegacySql` should probably be `UseLegacySQL`.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-view-direct/create.yaml
```diff
@@ -0,0 +1,37 @@
...
+metadata:
+  name: bigquerytablesample${uniqueId}
+  annotations:
+    cnrm.cloud.google.com/project-id: ${projectId}
+    alpha.cnrm.cloud.google.com/reconciler: direct
+spec:
+  datasetRef:
+    name: bigquerydatasetsample${uniqueId}
+  friendlyName: mv_test3
+  labels:
```
maqiuyujoyce, line 26: I wonder if we can have this test case (having `spec.labels`) without `alpha.cnrm.cloud.google.com/reconciler: direct` annotation.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-full/_generated_export_bigquerytable-full.golden
```diff
@@ -4,9 +4,6 @@ metadata:
   annotations:
     cnrm.cloud.google.com/project-id: ${projectId}
   labels:
-    cnrm-test: "true"
```
maqiuyujoyce, line 7: While labels are not critical to this resource, this breaking change can still be a bit confusing to users. Do we have any plan about how we should deliver this change more smoothly to the users? E.g. additional information when running `kubectl apply`, public documentation (trouble shooting doc, feature description), etc.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-full-direct/create.yaml
```diff
@@ -0,0 +1,108 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: bigquery.cnrm.cloud.google.com/v1beta1
+kind: BigQueryTable
+metadata:
+  name: bigquerytable-${uniqueId}
+  labels:
+    key1: value1
```
maqiuyujoyce, line 21: Nit: Could you add comments here saying these labels will be ignored in direct controller?

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-full/_http.log
```diff
@@ -630,87 +629,6 @@ X-Xss-Protection: 0
...
-  "timePartitioning": {
-    "field": "dt",
-    "type": "DAY"
-  },
-  "type": "TABLE"
-}
-
----
-
-GET https://bigquery.googleapis.com/bigquery/v2/projects/${projectId}/datasets/${datasetID}/tables/referencedtable${uniqueId}?alt=json&prettyPrint=false
```
maqiuyujoyce: Could you help me understand why do we have an additional GET for referencetable here?

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-view/_generated_object_bigquerytable-view.golden.yaml
```diff
@@ -2,13 +2,13 @@ apiVersion: bigquery.cnrm.cloud.google.com/v1beta1
...
   annotations:
+    alpha.cnrm.cloud.google.com/reconciler: direct
     cnrm.cloud.google.com/management-conflict-prevention-policy: none
     cnrm.cloud.google.com/project-id: ${projectId}
-    cnrm.cloud.google.com/state-into-spec: absent
   finalizers:
   - cnrm.cloud.google.com/finalizer
   - cnrm.cloud.google.com/deletion-defender
-  generation: 2
+  generation: 1
```
maqiuyujoyce: Could you help me understand why this golden file has generation `1` when it actually has any update? Does the update indeed work?

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-view-schema/_http.log
```diff
@@ -439,8 +437,128 @@ X-Xss-Protection: 0
...
+    "projectId": "${projectId}",
+    "tableId": "bigquerytablesample${uniqueId}"
+  },
+  "view": {
+    "query": "SELECT distinct dt, user_id FROM `${projectId}.bigquerydatasetsample${uniqueId}.bigquerytable-main-${uniqueId}`",
+    "useLegacySql": false
+  }
+}
+
+400 Bad Request
```
maqiuyujoyce: Are we testing an error case here? I think our current test frame doesn't yet support it properly.

pkg/controller/direct/bigquery/table_controller.go
```diff
@@ -0,0 +1,351 @@
...
+	log := klog.FromContext(ctx).WithName(ctrlName)
+	log.V(2).Info("deleting Table", "name", a.id.String())
+	parent := a.id.Parent()
+	if err := a.gcpService.Delete(parent.ProjectID, parent.DatasetID, a.id.ID()).Do(); err != nil {
+		return false, fmt.Errorf("deleting Table %s: %w", a.id, err)
+	}
+	return true, nil
+}
+
+func (a *Adapter) customTableLogic(table *bigquery.Table) {
```
maqiuyujoyce, line 325: Just curious, will this be the function where we plan to throw most (if not all) customized logic in? If so, could you add proper comments about future maintenance, and reasons why each customized logic is added here?

pkg/controller/direct/bigquery/table_controller.go
```diff
@@ -0,0 +1,351 @@
...
+
+	if err := a.normalizeReferences(ctx); err != nil {
+		return fmt.Errorf("normalizing reference for creation: %w", err)
+	}
+	desired := a.desired.DeepCopy()
+	table := BigQueryTableSpec_ToProto(mapCtx, &desired.Spec)
+
+	a.customTableLogic(table)
+	if desired.Spec.EncryptionConfiguration != nil && desired.Spec.EncryptionConfiguration.KmsKeyRef != nil {
+		kmsRef, err := refs.ResolveKMSCryptoKeyRef(ctx, a.reader, a.desired, a.desired.Spec.EncryptionConfiguration.KmsKeyRef)
```
maqiuyujoyce: Looks like this step is already done in normalizeReferences() function, I think you just want to do `table.EncryptionConfiguration.KmsKeyName = kmsRef.External` here, right?

pkg/controller/direct/bigquery/table_controller.go
```diff
@@ -0,0 +1,351 @@
...
+		res, err := a.gcpService.Insert(parent.ProjectID, parent.DatasetID, table).Do()
+		if err != nil {
+			return fmt.Errorf("error creating Table %s: %w", a.id.ID(), err)
+		}
+		log.V(2).Info("successfully created Table", "name", res.Id)
+		table.Schema = schemaBack
+		log.V(2).Info("Updating BigQuery Table back with schema", "name", res.Id)
+		res, err = a.gcpService.Update(parent.ProjectID, parent.DatasetID, res.Id, table).Do()
+		if err != nil {
+			return fmt.Errorf("error updating Table %s: %w", a.id.ID(), err)
```
maqiuyujoyce: Nit: Can we clarify in the error message that this is an updating step in the BigQueryTable creation? E.g. `error setting schema when creating Table %s`.

pkg/controller/direct/bigquery/table_controller.go
```diff
@@ -0,0 +1,351 @@
...
+
+	if err := a.normalizeReferences(ctx); err != nil {
+		return fmt.Errorf("normalizing reference for update: %w", err)
+	}
+
+	desired := a.desired.DeepCopy()
+	table := BigQueryTableSpec_ToProto(mapCtx, &desired.Spec)
+
+	// For table created from external data configuration,
+	// If no schema was provided, the Update request would fail.
```
maqiuyujoyce, line 220: What if the provided schema is different from the schema generated from the external data? I recall from the test, this doesn't matter?

pkg/controller/direct/bigquery/table_controller.go
```diff
@@ -0,0 +1,351 @@
...
+		}
+	}
+	eq, err := TableEq(a.actual, table)
+
+	if err != nil {
+		return err
+	}
+	// No diff detected.
+	if eq {
+		return nil
```
maqiuyujoyce: We might still want to update status to get the latest update?

pkg/controller/direct/bigquery/table_controller.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+
+	a.customTableLogic(table)
+	parent := a.id.Parent()
+
+	if mapCtx.Err() != nil {
+		return fmt.Errorf("error updating Table %s: %w", a.id.ID(), mapCtx.Err())
```
maqiuyujoyce: This error check should be moved up to  line 227, right?

And the error message is kinda confusing, should it just be `return mapCtx.Err()`?

pkg/controller/direct/bigquery/table_controller.go
```diff
@@ -0,0 +1,351 @@
...
+	// 	}
+	// }
+	res, err := a.gcpService.Update(parent.ProjectID, parent.DatasetID, a.id.ID(), table).Do()
+	if err != nil {
+		return fmt.Errorf("error updating Table %s: %w", a.id.ID(), err)
+	}
+	return a.UpdateStatusForUpdate(ctx, updateOp, res)
+}
+
+func (a *Adapter) UpdateStatusForUpdate(ctx context.Context, updateOp *directbase.UpdateOperation, created *bigquery.Table) error {
```
maqiuyujoyce: Nit: Param `created` should be `updated`.

pkg/controller/direct/bigquery/table_controller.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+	u := &unstructured.Unstructured{}
+
+	obj := &krm.BigQueryTable{}
+	mapCtx := &direct.MapContext{}
+	obj.Spec = direct.ValueOf(BigQueryTableSpec_FromProto(mapCtx, a.actual))
+	if mapCtx.Err() != nil {
+		return nil, mapCtx.Err()
+	}
+	uObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
```
maqiuyujoyce, line 297: You'll also want to handle identity related fields, i.e. `datasetRef` before converting obj to unstructured.

pkg/controller/direct/bigquery/table_controller.go
```diff
@@ -0,0 +1,351 @@
...
+	return u, nil
+}
+
+// Delete implements the Adapter interface.
+func (a *Adapter) Delete(ctx context.Context, deleteOp *directbase.DeleteOperation) (bool, error) {
+	log := klog.FromContext(ctx).WithName(ctrlName)
+	log.V(2).Info("deleting Table", "name", a.id.String())
+	parent := a.id.Parent()
+	if err := a.gcpService.Delete(parent.ProjectID, parent.DatasetID, a.id.ID()).Do(); err != nil {
+		return false, fmt.Errorf("deleting Table %s: %w", a.id, err)
```
maqiuyujoyce, line 318: You'll want to exclude not found error here.

pkg/controller/direct/bigquery/table_compare.go
anhdle-sso: done

pkg/controller/direct/bigquery/table_mapper.go
anhdle-sso: yes, the test `bigquerytable-view` and `bigquerytable-view-direct` cover this behavior

pkg/controller/direct/bigquery/table_mapper.go
anhdle-sso: yes, the test `bigquerytable-view` and `bigquerytable-view-direct` cover this behavior

pkg/controller/direct/bigquery/table_controller.go
anhdle-sso: you are right, this is duplicated logic. normalizeReference() and the mapper should take care of this already

pkg/controller/direct/bigquery/table_controller.go
anhdle-sso: done, this is not needed.

pkg/controller/direct/bigquery/table_compare.go
anhdle-sso: done. after the check to both Names are not nil. the len() equal check should cover empty check. I still keep the not nil check via `b.Names == nil && a.Names == nil` and `a.Names == nil || b.Names == nil`.

pkg/controller/direct/bigquery/table_compare.go
anhdle-sso: Added some comments before the loop. I did not do deepEqual on the full object due to the tendency of having output only sub fields (which change every GET request). These fields will create infinite reconciliation PUT call to the API. These behavior was caught by the re-reconciliation test.

pkg/controller/direct/bigquery/table_compare.go
anhdle-sso: I compare each field as needed due to output nested field.

pkg/controller/direct/bigquery/table_controller.go
anhdle-sso: done

pkg/controller/direct/bigquery/table_controller.go
anhdle-sso: from testing with real gcp object. The field `spec.externalDataConfiguration.schema` does not really seem to matter. The field `spec.schema` is used instead of `spec.externalDataConfiguration.schema`. The TF controller does not really check for this. We can update this logic after getting more feedback.

pkg/controller/direct/bigquery/table_controller.go
anhdle-sso: done

pkg/controller/direct/bigquery/table_controller.go
anhdle-sso: done

pkg/controller/direct/bigquery/table_controller.go
anhdle-sso: done

pkg/controller/direct/bigquery/table_controller.go
anhdle-sso: done

pkg/controller/direct/bigquery/table_controller.go
anhdle-sso: done

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-view-direct/create.yaml
anhdle-sso: Yes. we have the non direct test case with both spec.labels and metadata.labels

pkg/controller/direct/bigquery/table_controller.go
anhdle-sso: Done. Added comment to the function and the comment to the logic inside the function.

pkg/controller/direct/bigquery/table_mapper.go
anhdle-sso: Agree. I can address in later PR. We had this type from TF controller generated type. The unit test already has this in the exception file.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-full-direct/create.yaml
anhdle-sso: done

anhdle-sso:
@maqiuyujoyce I addressed the previous comments. Please take a look

pkg/controller/direct/bigquery/table_controller.go
```diff
@@ -0,0 +1,354 @@
...
+
+func (r *TableReconcileGate) ShouldReconcile(o *unstructured.Unstructured) bool {
+	if r.optIn.ShouldReconcile(o) {
+		return true
+	}
+	obj := &krm.BigQueryTable{}
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(o.Object, &obj); err != nil {
+		return false
+	}
+	return obj.Spec.Labels != nil
```
maqiuyujoyce, line 65: @yuwenma fyi this is the change about labels in BigQueryTable direct controller. Does it align with your design?

pkg/controller/direct/bigquery/table_compare.go
maqiuyujoyce: Thank you for the context!

> I did not do deepEqual on the full object due to the tendency of having output only sub fields (which change every GET request).

How many output only fields are there? If there are much fewer output only fields, then I still think manually comparing them field by field could be more error-prone and harder to maintain.

I won't block this PR for this improvement, but I do think we want to simplify the comparison here eventually for better readability and maintenance in the future.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-view-direct/create.yaml
maqiuyujoyce: Ah, I may have missed it. Could you share which test case is that?

Non-blocking.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-full/_generated_export_bigquerytable-full.golden
maqiuyujoyce: This comment is made at commit f8a22d71af9d28725df28d5d2ad0475f8fb855ef when you switch the test case from TF-based to direct. `cnrm-test` can be an important identifier for users when they identify whether a GCP resource is managed by KCC or not so I am leaning towards we keep the feature.

We can address it in the following PRs.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-full/_http.log
maqiuyujoyce: Similarly, this comment is made at commit f8a22d7 when you switch the test case from TF-based to direct. With the TF-based controller, there is an additional GET for referencetable. Do you happen to know why? I'm concerned about potential behavior mismatching in between the two controllers.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-view/_generated_object_bigquerytable-view.golden.yaml
maqiuyujoyce: Similarly, this comment is made at commit f8a22d7 when you switch the test case from TF-based to direct. With generation `1`, it looks like no update happened. Do you know why?

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-view-schema/_http.log
maqiuyujoyce: this comment is made at commit f8a22d7 when you switch the test case from TF-based to direct. This 400 doesn't look right, do you happen to know why?

pkg/controller/direct/bigquery/table_controller.go
```diff
@@ -0,0 +1,354 @@
...
+	log.V(2).Info("creating Table", "name", a.id.String())
+	mapCtx := &direct.MapContext{}
+
+	if err := a.normalizeReferences(ctx); err != nil {
+		return fmt.Errorf("normalizing reference for creation: %w", err)
+	}
+	desired := a.desired.DeepCopy()
+	table := BigQueryTableSpec_ToProto(mapCtx, &desired.Spec)
+
+	a.customTableLogic(table)
```
maqiuyujoyce, line 152: I thought `table.EncryptionConfiguration.KmsKeyName = kmsRef.External` is still needed? (Previous discussion: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4578#discussion_r2160284652).

pkg/controller/direct/bigquery/table_controller.go
maqiuyujoyce: Actually, we may want to update status even if it is not the first reconciliation. Observed fields may be updated by the API even if the desired state is not changed.

Non-blocking.

anhdle-sso: done, added to update status with actual object no mater what

pkg/controller/direct/bigquery/table_controller.go
anhdle-sso: That logic has been taken care of in `normalizeReferences` function. The function is called a few lines above

pkg/controller/direct/bigquery/table_compare.go
anhdle-sso: You are right, there are a lot more non output only fields. We can make some improvement on this. For the improvement we have to overcome some challenges:
The  BQ API does not have proto field support (missing ProtoReflect).
The PUT/PATCH does not support fieldmask.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-view/_generated_object_bigquerytable-view.golden.yaml
maqiuyujoyce: Rethinking about this generation number when using the TF-based controller, it should probably be 3 instead of 2. I'd keep an eye on this example.

Non-blocking.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-view-schema/_http.log
maqiuyujoyce: Discussed offline. It's a bug and @anhdle-sso fixed it in the latest commits.

pkg/controller/direct/bigquery/table_mapper.go
```diff
@@ -631,7 +695,11 @@ func BigQueryTableSpec_ToProto(mapCtx *direct.MapContext, in *krmv1beta1.BigQuer
 	out.FriendlyName = direct.ValueOf(in.FriendlyName)
 	out.Description = direct.ValueOf(in.Description)
 	out.Labels = in.Labels
-	out.Schema = Table_Schema_ToProto(mapCtx, in.Schema)
+	if out.Labels == nil {
+		out.Labels = map[string]string{}
+	}
+	out.Labels["managed-by-cnrm"] = "true"
```
maqiuyujoyce, line 701: Nit: I suggest this to be done in the controller instead of mapper. I think this is a controller level customization, not mapper level.

pkg/controller/direct/bigquery/table_mapper.go
```diff
@@ -333,12 +374,29 @@ func CsvOptions_ToProto(mapCtx *direct.MapContext, in *krmv1beta1.CsvOptions) *p
 		return nil
 	}
 	out := &pb.CsvOptions{}
-	out.FieldDelimiter = direct.ValueOf(in.FieldDelimiter)
-	out.SkipLeadingRows = direct.ValueOf(in.SkipLeadingRows)
+	// if spec.externalDataConfiguration.csvOptions.fieldDelimiter is not set. Default to `,`.
+	if in.FieldDelimiter == nil {
```
maqiuyujoyce, line 378: Nit: I think defaulting behaviors you added in the mapper should probably be extracted out into a separate defaults file like https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/64ed30faf50512cc4a0947c1f5291de44ca706cb/pkg/controller/direct/sql/sqlinstance_defaults.go.

maqiuyujoyce (commented):
Thank you for the fix, @anhdle-sso !

/lgtm
/approve

Left a couple of nits and they are non blocking.

PR #4579, opened 2025-05-19
chore: refactor logging references

gemmahou:
### Change description

Move loggingrefs from apis/refs/v1beta1 to apis/logging/v1beta1 and use NormalizedExternal to resolve its value

### Tests you have done


gemmahou:
/assign @acpana 

Would mind reviewing this PR as well? Thanks!

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_logginglogmetrics.logging.cnrm.cloud.google.com.yaml
```diff
@@ -151,29 +151,24 @@ spec:
                     - external
                   required:
                   - name
-                  - kind
```
acpana, line 154: I'm going to defer to @yuwenma or @cheftako but I think that we CAN remove the `kind` field here assuming that said removal will NOT break existing customers. Have you tried applying a CR with the `kind` still present?

( For context, I think when we added this, we didn't have a pattern on the new/ direct references direction and reused the old reference style. )

acpana (commented):
This overall LGTM but would appreciate yuwen's or walter's sign off before removing the kind field for the logbucket ref in logginglogmetric.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_logginglogmetrics.logging.cnrm.cloud.google.com.yaml
gemmahou: You're correct, removing `kind` would be a breaking change for customers using it in their CR, which we should avoid for Beta resources. While I feel requiring `kind` field in loggingLogBucketRef might seem inconsistent with our other Refs, and we should fix that sooner rather than later if the impact is still limited. I'll defer to Yuwen and Walter for the final decision on this.

gemmahou: I reverted this change back to unblock this PR. LoggingLogMetric's LoggingLogBucketRef is using `v1alpha1.ResourceRef`, we can continue the discussion offline and followup in another PR.

pkg/controller/direct/logging/logmetric_controller.go
```diff
@@ -98,6 +98,9 @@ func (m *logMetricModel) AdapterForObject(ctx context.Context, reader client.Rea
 		return nil, fmt.Errorf("cannot resolve project")
 	}
 
+	// resolve LoggingLogBucketRef
+	// todo: LoggingLogBucketRef is *v1alpha1.ResourceRef, ideally should use *loggingv1beta1.LoggingLogBucketRef instead
+	// *v1alpha1.ResourceRef has required `kind` field, this migration could introduce breaking changes to Beta CRD
```
jingyih, line 103: I see. Yeah we cannot break logging logmetric which is already in beta.

jingyih (commented):
/lgtm

PR #4580, opened 2025-05-19
chore: refactor storage references

gemmahou:
### Change description

Move storage refs from apis/refs/v1beta1 to apis/storage/v1beta1 and use NormalizedExternal to resolve its value

### Tests you have done


gemmahou:
Backward compatibility verification:
1. According to the missing refs verification, no missing StorageBucketRef in direct controller: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5097/files#diff-378fd6da23986fa7b46b598166777f30b168b4bdd42188c9fb35fdec59722d94

2. No changes in the golden log of existing test with StorageBucketRef

apis/storage/v1beta1/bucket_identity.go
```diff
@@ -0,0 +1,50 @@
...
+package v1beta1
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/apis/common/parent"
+)
+
+type StorageBucketIdentity struct {
```
yuwenma, line 24: you may want to look at apis/common/identity.go. Seems to be a new trend.

pkg/controller/direct/storage/control/anywherecache_controller.go
```diff
@@ -369,7 +368,7 @@ func (a *AnywhereCacheAdapter) Export(ctx context.Context) (*unstructured.Unstru
 	if mapCtx.Err() != nil {
 		return nil, mapCtx.Err()
 	}
-	obj.Spec.BucketRef = &refs.StorageBucketRef{External: a.id.Parent().String()}
+	obj.Spec.BucketRef = &krm.StorageBucketRef{External: a.id.Parent().String()}
```
yuwenma, line 371: Can we move this to the mapper function StorageAnywhereCacheSpec_FromProto?

apis/storage/v1beta1/bucket_identity.go
gemmahou: I was aware of the [identity interface](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/common/identity/identity.go), it was added by Jason last year and used for his new [reference interface](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/refs/v1beta1/interface.go#L37). If we are not going to migrate to the new reference interface, it seems unnecessary to use the new identity interface(it introduces additional codes). 

I'm thought we wanted to keep the current implementation of identity and normalizer, and determined the migration to new interfaces later... WDYT?

pkg/controller/direct/storage/control/anywherecache_controller.go
gemmahou: This line appears redundant, it is already handled in StorageAnywhereCacheSpec_FromProto. I'm leaving this line for now to keep this PR focused strictly on the refs refactoring. It has no side effects, I'll clean it up when we get to the mappers.

PR #4583, opened 2025-05-20
tests: introduce test for reconcile being stable

justinsb:
Use a GVK-based ratchet to implement this gradually,
as we expect at least some breakages.


justinsb:
Blocked on #4589

justinsb:
Really had to downscope this, it looks like there are many resources where we will issue spurious updates.  Introduced a ratchet so we can ramp up coverage gradually without a massive PR.

cc @yuwenma 

/assign @maqiuyujoyce

yuwenma:
> Really had to downscope this, it looks like there are many resources where we will issue spurious updates. Introduced a ratchet so we can ramp up coverage gradually without a massive PR.
> 

We can also use scenario test to check the "touch" logic (at least for direct resources). But it's optional so far. What's the missing test coverage does this change cover?

tests/e2e/ratcheting.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// ShouldTestRereconiliation determines if we "touch" the primary object after we have run the test.
+// This should not cause write operations to GCP (read operations are OK)
+// We would like eventually to turn this on for all objects, but we have to turn on the testing gradually.
+func ShouldTestRereconiliation(t *testing.T, primaryResource *unstructured.Unstructured) bool {
+	gvk := primaryResource.GroupVersionKind()
+
+	switch gvk.GroupKind() {
+	case schema.GroupKind{Group: "pubsub.cnrm.cloud.google.com", Kind: "PubSubTopic"}:
```
maqiuyujoyce: Does it also apply to PubSubSubscription? (As you've removed basicpubsubsubscription from the vcr test list.)

Non-blocker.

tests/e2e/unified_test.go
```diff
@@ -376,6 +376,31 @@ func runScenario(ctx context.Context, t *testing.T, testPause bool, fixture reso
 					}
 				}
 
+				if ShouldTestRereconiliation(t, primaryResource) {
```
maqiuyujoyce, line 379: Awesome! We finally have the "nochange" test equivalence in mock test. Thanks a lot!

tests/e2e/unified_test.go
```diff
@@ -376,6 +376,31 @@ func runScenario(ctx context.Context, t *testing.T, testPause bool, fixture reso
...
+						isReadOnly := false
+						switch event.Request.Method {
+						case "GET":
+							isReadOnly = true
+						}
+						if !isReadOnly {
+							t.Errorf("FAIL: unexpected event during re-reconciliation: %v", event)
+						}
+					}
+				}
```
maqiuyujoyce, line 408: Could be redundant, but can we also verify the generation is not updated in the object? I'm concerned a controller may write back to spec unexpectedly as each controller has its own logic.

tests/e2e/ratcheting.go
```diff
@@ -0,0 +1,41 @@
...
+		return true
+	}
+
+	switch gvk.Group {
```
maqiuyujoyce: Can we blocklist the existing groups instead?

I'd really hope newly added resources do re-reconciliation by default.

maqiuyujoyce (commented):
/lgtm
/approve

/hold
Feel free to unhold if you prefer to handle the comments separately!

tests/e2e/ratcheting.go
justinsb: It does not, I was trying to but there's a problem on PubSubSubscription also.

tests/e2e/ratcheting.go
justinsb: Hmmmm... I really like that suggestion, it provides a sort of "count-down to full coverage"

tests/e2e/unified_test.go
justinsb: Oooh, that's a good idea - let me look at whether it's possible / hard

PR #4587, opened 2025-05-20
chore: refactor appengine references

gemmahou:
### Change description

Move refs from apis/refs/v1beta1 to apis/appengine/v1beta1 and use NormalizedExternal to resolve its value

### Tests you have done


apis/iap/v1alpha1/iapsettings_identity.go
```diff
@@ -198,42 +200,35 @@ func (p ComputeServiceParent) buildIAPSettingsID(ctx context.Context, reader cli
...
 }
 
 func (p AppEngineParent) buildIAPSettingsID(ctx context.Context, reader client.Reader, namespace string) (string, error) {
-	project, err := refsv1beta1.ResolveProject(ctx, reader, namespace, p.ProjectRef)
-	if err != nil {
-		return "", err
-	}
-
-	appID, err := refsv1beta1.ResolveAppEngineApplicationID(ctx, reader, namespace, p.ApplicationRef)
+	applicationExternal, err := p.ApplicationRef.NormalizedExternal(ctx, reader, namespace)
```
justinsb, line 209: This is an existing problem, but the logic here is somewhat non-idiomatic.  Go typically recommends against nesting if statements where possibly, and this function would be much clearer without nesting IMO.

apis/iap/v1alpha1/iapsettings_identity.go
```diff
@@ -198,42 +200,35 @@ func (p ComputeServiceParent) buildIAPSettingsID(ctx context.Context, reader cli
...
 			return "", err
 		}
 
 		if p.VersionRef != nil {
-			versionID, err := refsv1beta1.ResolveAppEngineVersionID(ctx, reader, namespace, p.VersionRef)
+			versionExternal, err := p.VersionRef.NormalizedExternal(ctx, reader, namespace)
 			if err != nil {
 				return "", err
 			}
-			return fmt.Sprintf("projects/%s/iap_web/appengine-%s/services/%s/versions/%s",
```
justinsb, line 228: Oh - we appear to be changing the logic from combining the references to a pass-through of the external?

gemmahou: Yes, the format of `versionExternal` is "projects/{{projectID}}/iap_web/appengine-{{appID}}/service/{{service_id}}/version/{{versionID}}`

Note that appengine version is not a KCC resource yet, so we can only reference it by using external reference.

apis/iap/v1alpha1/iapsettings_identity.go
gemmahou: I didn't touch the existing logic in this PR, except for those related to references.

I understand why we had the nested statement - `serviceID` was an integral part of `versionExternal`. With the updated NormalizedExternal, this nesting is no longer necessary. I updated the code.

apis/iap/v1alpha1/iapsettings_identity.go
```diff
@@ -198,42 +200,37 @@ func (p ComputeServiceParent) buildIAPSettingsID(ctx context.Context, reader cli
...
-		if p.VersionRef != nil {
-			versionID, err := refsv1beta1.ResolveAppEngineVersionID(ctx, reader, namespace, p.VersionRef)
-			if err != nil {
-				return "", err
-			}
-			return fmt.Sprintf("projects/%s/iap_web/appengine-%s/services/%s/versions/%s",
-				project.ProjectID, appID, serviceID, versionID), nil
+		return serviceExternal, nil
+	}
+	if p.VersionRef != nil {
```
jingyih, line 225: It appears that we are changing the logic here? I think we are missing test cases for these fields and therefore the logic change was undetected.

gemmahou: I have split the logic updates in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5183 and added test coverage.

ldanielmadariaga:

PR #4589, opened 2025-05-20
tests: don't just update labels in pubsub topic test

justinsb:
- **tests: don't just change labels on pubsubtopic test**
  Because changing the labels does not bump observedGeneration,
  the tests do not correctly wait for reconciliation.
  

- **mockgcp: implement implied field mask in pubsub topic**

### DIFF
--- a/mockgcp/common/fields/impliedfieldmask.go
+++ b/mockgcp/common/fields/impliedfieldmask.go
@@ -1,0 +1,37 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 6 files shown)


mockgcp/common/fields/impliedfieldmask.go
```diff
@@ -0,0 +1,37 @@
...
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+// ComputeImpliedFieldMask computes the implied field mask when it is omitted.
+// According to AIP 134: this is "equivalent to all fields that are populated (have a non-empty value)."
+// https://google.aip.dev/134
+func ComputeImpliedFieldMask(ctx context.Context, req proto.Message) []string {
+	var fieldMask []string
+	req.ProtoReflect().Range(func(fd protoreflect.FieldDescriptor, v protoreflect.Value) bool {
```
cheftako, line 30: // Message.Range iterates over every populated field in an undefined order

PR #4590, opened 2025-05-20
chore: don't alias controller-runtime log to klog

justinsb:
klog is a different but similar-enough-to-be-confusing library.

### DIFF
--- a/cmd/unmanageddetector/main.go
+++ b/cmd/unmanageddetector/main.go
@@ -32,12 +32,12 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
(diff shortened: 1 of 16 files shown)


cmd/unmanageddetector/main.go
```diff
@@ -32,12 +32,12 @@ import (
...
 	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
-	klog "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
 )
 
-var logger = klog.Log.WithName("setup")
+var logger = log.Log.WithName("setup")
```
cheftako, line 40: Hobgoblin would like to suggest we just uniformly call this crlog.

PR #4591, opened 2025-05-20
tests: handle updates that only change labels

justinsb:


yuwenma:
> So I don't think we are correctly running tests that only change labels, because label changes do not change the metadata.generation. I'm introducing a pause here, but I think probably we should try to find a non-label field to update (or adopt spec.labels cc @yuwenma)
> 

Great catch (I observed some test suites having the metadata.labels, seems meaningless but now I see what it supposes to do). For pure direct reconciliers, we are good because it use the `spec.labels`.

pkg/test/resourcefixture/testdata/basic/edgecontainer/v1beta1/edgecontainercluster/_generated_object_edgecontainercluster.golden.yaml
```diff
@@ -34,8 +34,10 @@ spec:
 status:
   conditions:
   - lastTransitionTime: "1970-01-01T00:00:00Z"
-    message: The resource is up to date
-    reason: UpToDate
-    status: "True"
+    message: 'Update call failed: error applying desired state: summary: Error updating
```
maqiuyujoyce: Do we plan to keep the golden output in the errored out state before we get to polish the mocks?

maqiuyujoyce (commented):
> So I don't think we are correctly running tests that only change labels, because label changes do not change the metadata.generation. I'm introducing a pause here, but I think probably we should try to find a non-label field to update (or adopt spec.labels cc @yuwenma)
> 
> /assign @maqiuyujoyce
> (but I don't think this will be ready to merge for a while, we need to either hope the 2-second-sleep works or we need to switch any tests to update a non-label field)

Good catch!!!

Do we know how many test cases are like this? Checking the results of "tests-e2e-fixtures" presubmit job and it looks like there are 7 hits; "tests-e2e-fixtures-compute" has 4 hits. Though in this PR, there are 8 test cases updated in total.

pkg/test/resourcefixture/testdata/basic/edgecontainer/v1beta1/edgecontainercluster/_generated_object_edgecontainercluster.golden.yaml
justinsb: Good catch and no - that's a mistake.   I guess we _could_ introduce tests that are expected to report errors, and maybe that would be a good idea, but I think all our tests today are expected to converge to a good state.

pkg/test/resourcefixture/testdata/basic/servicedirectory/v1beta1/servicedirectorynamespace/_http.log
```diff
@@ -80,6 +80,64 @@ X-Xss-Protection: 0
 
 ---
 
+PATCH https://servicedirectory.googleapis.com/v1beta1/projects/${projectId}/locations/us-central1/namespaces/servicedirectory-namespace-${uniqueId}?alt=json&updateMask=labels
```
maqiuyujoyce, line 83: Nice! Do you mind adding a commit tracking the log changes when running against real API?

ldanielmadariaga:

PR #4598, opened 2025-05-20
fix: Resolve KMSKey before comparison in BQDataset direct controller

gemmahou:
### Change description

Resolve KMSKeyRef for desired object before comparison with the actual object. Otherwise, extra PATCH call will be sent even if there was no change to the KMS key.

### Tests you have done


jingyih (commented):
/lgtm
/approve

Thank you!

PR #4601, opened 2025-05-21
BigtableAppProfile: do not write back ResourceID

jingyih:
I asked the original author to "write back resourceID" for backward compatibility, as we are migrating this resource from TF-beta to direct. However, I’ve realized that we no longer want to keep this behavior in the direct controller.

(In 1.132, the controller still defaults to TF controller)


yuwenma (commented):
/lgtm
/approve

Right. You can consider the resourceID tracked in "status.externalRef", and spec is only for user configurations.

PR #4603, opened 2025-05-21
Upgrade the LLM model.

cheftako:
### Change description

Upgraded default model to gemini-2.5-pro-preview-05-06.
Remove a problematic check in the batch clean command.

### Tests you have done

Ran conductor and built dev/tools.


yuwenma (commented):
/lgtm
/approve

Thanks. One problem I hit when using this new model is that it can suddenly fail with `Message: Please use a valid role: user, model., Status: INVALID_ARGUMENT`.

cheftako:
> /lgtm /approve
> 
> Thanks. One problem I hit when using this new model is that it can suddenly fail with `Message: Please use a valid role: user, model., Status: INVALID_ARGUMENT`.

Do you get a stack or any hint where the error is occuring?

PR #4511, opened 2025-05-09
feat: turn on CloudIdentityGroup direct controller

gemmahou:
### Change description

Fixes #

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #4604, opened 2025-05-21
chore: Handle the case when direct resource is referenced by TF/DCL resources

gemmahou:
### Change description

Cherry-picked https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4511/commits/0c904b425df8c28530eb7fed40df9c231a29b42f

1. Handled the case when direct resource is referenced by TF/DCL resources, be able to retrieve its resourceID from status.externalRef(I'm surprised that this was not caught before...)
2. Added one more test case(addrolecloudidentitymembership-groupdirect) to verify above change.
3. Added back CloudIdentityGroup TF-based test case. Because we are not going to make direct reconciliation as default, it's better to keep both direct and TF tests

### Tests you have done


gemmahou:
/assign @jingyih 
cc: @maqiuyujoyce

pkg/dcl/kcclite/conversion.go
```diff
@@ -443,9 +445,18 @@ func getReferencedResource(resourceRefValRaw map[string]interface{}, tc *corekcc
 	return refResource, nil
 }
 
-// TODO(kcc-eng): consolidate this method with krmtotf.resolveTargetFieldValue when resourceID support is added
 func resolveTargetFieldValue(refResource *k8s.Resource, typeConfig *corekccv1alpha1.TypeConfig) (string, error) {
 	if typeConfig.TargetField == "name" {
+		// Resolve target value from direct resources, when spec.resourceID is unspecified.
+		// Get resourceID from externalRef
+		if supportedgvks.IsDirectByGVK(refResource.GroupVersionKind()) || refResource.Annotations["alpha.cnrm.cloud.google.com/reconciler"] == "direct" {
```
maqiuyujoyce, line 452: Nit: Can we make/use a constant for the annotation? Not a blocker.

pkg/krmtotf/references.go
```diff
@@ -232,6 +232,16 @@ func resolveTargetFieldValue(r *Resource, tc corekccv1alpha1.TypeConfig) (interf
 }
 
 func resolveDefaultTargetFieldValue(r *Resource, tc corekccv1alpha1.TypeConfig) (interface{}, error) {
+	// Resolve target value from direct resources, when spec.resourceID is unspecified.
```
maqiuyujoyce, line 235: Could you update the comment to match the code? Looks like we don't care whether spec.resourceID is set or not for a direct resource, but just do `Resolve default target value from status.externalRef for direct resources.`

pkg/krmtotf/references.go
```diff
@@ -232,6 +232,16 @@ func resolveTargetFieldValue(r *Resource, tc corekccv1alpha1.TypeConfig) (interf
...
+	// Get resourceID from externalRef
+	if r.ResourceConfig.Direct || r.Annotations["alpha.cnrm.cloud.google.com/reconciler"] == "direct" {
```
maqiuyujoyce, line 237: Nit: Can we turn this logic into a single util function in k8s package? They seem to be the same for either TF-based or DCL-based resource.

pkg/test/resourcefixture/testdata/basic/cloudidentity/v1beta1/cloudidentitygroup/cloudidentitygroupbasic/_generated_object_cloudidentitygroupbasic.golden.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: cloudidentity.cnrm.cloud.google.com/v1beta1
```
maqiuyujoyce, line 1: Nit: Maybe making the test case name explicit by adding "tf" in the folder name? In the future, we can also split the test cases by tf/dcl/direct test cases when needed.

maqiuyujoyce (commented):
/lgtm
/approve
/hold

Hold because of one small comment.

pkg/dcl/kcclite/conversion.go
gemmahou: Done

pkg/test/resourcefixture/testdata/basic/cloudidentity/v1beta1/cloudidentitygroup/cloudidentitygroupbasic/_generated_object_cloudidentitygroupbasic.golden.yaml
gemmahou: I prefer not to change the test file name for now to keep a consistent naming pattern (we only add a suffix for direct tests). This will be a quick change if needed later.

pkg/krmtotf/references.go
```diff
@@ -234,15 +232,16 @@ func resolveTargetFieldValue(r *Resource, tc corekccv1alpha1.TypeConfig) (interf
 }
 
 func resolveDefaultTargetFieldValue(r *Resource, tc corekccv1alpha1.TypeConfig) (interface{}, error) {
+	// todo: uncomment below code after b/420945170(missing status.externalRef in direct resources)
```
gemmahou, line 235: Added a placeholder here. In b/420945170 we noticed that some direct resources(referenced by TF resources) are missing `status.externalRef`. This needs to be fixed first.

PR #4606, opened 2025-05-21
Migrate datacatalog Taxonomy Type and Ref

anhdle-sso:
### Change description

Add datacatalog Taxonomy Type and Ref

### Tests you have done


config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_datacatalogtaxonomies.datacatalog.cnrm.cloud.google.com.yaml
```diff
@@ -145,29 +151,24 @@ spec:
...
               observedGeneration:
                 description: ObservedGeneration is the generation of the resource
                   that was most recently observed by the Config Connector controller.
                   If this is equal to metadata.generation, then that means that the
                   current reported status reflects the most recent desired state of
                   the resource.
+                format: int64
                 type: integer
             type: object
-        required:
```
jingyih, line 161: Maybe we should keep "spec" as required?

jingyih (commented):
Looks good. Defer to @yuwenma

yuwenma (commented):
/approve

The `ActivatedPolicyTypes` looks like a machine state (".. that are activated...") rather than a user configurable field. Could you double check when implementing the controller that this field is configurable?

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_datacatalogtaxonomies.datacatalog.cnrm.cloud.google.com.yaml
anhdle-sso: done

anhdle-sso:
> /approve
> 
> The `ActivatedPolicyTypes` looks like a machine state (".. that are activated...") rather than a user configurable field. Could you double check when implementing the controller that this field is configurable?

This should be configurable by users. Also, this is a TF beta resource; I keep the types the same to keep backward compatibility.

PR #4425, opened 2025-04-23
[Cloud Bigtable] Add support for Data Boost Isolation

djyau:
### Change description

Fixes #

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #4607, opened 2025-05-22
chore: clean up mapping functions for bigtable

jingyih:
We have a contributor working on a new resource in bigtable API. Let's make sure the generated mapping function file is regenerate-safe.

### DIFF
--- a/dev/tools/controllerbuilder/pkg/codegen/mappergenerator.go
+++ b/dev/tools/controllerbuilder/pkg/codegen/mappergenerator.go
@@ -191,7 +191,7 @@ func (v *MapperGenerator) GenerateMappers() error {
 			// TODO: Can we figure out a better way here?
 			switch pbPackage {
 			case "cloud.google.com/go/bigtable/admin/apiv2/adminpb":
-				pbPackage = "google.golang.org/genproto/googleapis/bigtable/admin/v2"
+				// pbPackage = "google.golang.org/genproto/googleapis/bigtable/admin/v2"
 			}
 
 			out.addImport("refs", "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1")
--- a/pkg/controller/direct/bigtable/appprofile_fuzzer.go
+++ b/pkg/controller/direct/bigtable/appprofile_fuzzer.go
@@ -39,11 +39,15 @@ func bigtableAppProfileFuzzer() fuzztesting.KRMFuzzer {
 	f.SpecFields.Insert(".multi_cluster_routing_use_any")
 	f.SpecFields.Insert(".single_cluster_routing")
 	f.SpecFields.Insert(".standard_isolation")
+	f.SpecFields.Insert(".data_boost_isolation_read_only")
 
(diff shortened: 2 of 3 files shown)


dev/tools/controllerbuilder/pkg/codegen/mappergenerator.go
```diff
@@ -191,7 +191,7 @@ func (v *MapperGenerator) GenerateMappers() error {
 			// TODO: Can we figure out a better way here?
 			switch pbPackage {
 			case "cloud.google.com/go/bigtable/admin/apiv2/adminpb":
-				pbPackage = "google.golang.org/genproto/googleapis/bigtable/admin/v2"
+				// pbPackage = "google.golang.org/genproto/googleapis/bigtable/admin/v2"
```
jingyih, line 194: This is an deprecated library, used by [pkg/controller/direct/bigtable/instance_mapper.go](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/8f39ac58feecdcca15a3265bf07c4e8377cb8de3/pkg/controller/direct/bigtable/instance_mapper.go#L24) and [pkg/controller/direct/bigtable/authorizedview_mapper.go](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/8f39ac58feecdcca15a3265bf07c4e8377cb8de3/pkg/controller/direct/bigtable/authorizedview_mapper.go#L20). But the generated mapper file uses "cloud.google.com/go/bigtable/admin/apiv2/adminpb".

pkg/controller/direct/bigtable/appprofile_fuzzer.go
```diff
@@ -39,11 +39,15 @@ func bigtableAppProfileFuzzer() fuzztesting.KRMFuzzer {
 	f.SpecFields.Insert(".multi_cluster_routing_use_any")
 	f.SpecFields.Insert(".single_cluster_routing")
 	f.SpecFields.Insert(".standard_isolation")
+	f.SpecFields.Insert(".data_boost_isolation_read_only")
 
 	f.UnimplementedFields.Insert(".name") // special field
 	f.UnimplementedFields.Insert(".etag")
-	f.UnimplementedFields.Insert(".priority")                       // deprecated in favor of standard isolation
-	f.UnimplementedFields.Insert(".data_boost_isolation_read_only") // only applies to online serving feature, doesn't support by kcc
```
jingyih, line 46: This is actually later implemented in #4425. Marking as "UnimplementedFields" is a bit confusing.

dev/tools/controllerbuilder/pkg/codegen/mappergenerator.go
yuwenma: Do you mean KCC never hits this line of code?

jingyih: It does, and the code will change the import library in `mapper.generated.go` to the deprecated `google.golang.org/genproto/googleapis/bigtable/admin/v2`, which is why this line needs to be removed.

PR #4608, opened 2025-05-22
tests: run tracker presubmits too

acpana:
Even if we are just starting with the `Project` kind, I am turning on some early testing for our mock/ presubmits tests where the tracker flag gate is on!

As we turn on more of these, I'll revisit how we select the RUN_TESTS var.

### DIFF
--- a/dev/ci/presubmits/tracker-tests
+++ b/dev/ci/presubmits/tracker-tests
@@ -33,6 +33,6 @@ echo "Running scenario tests for gcptracker..."
 GOLDEN_REQUEST_CHECKS=1 E2E_KUBE_TARGET=envtest E2E_GCP_TARGET=mock RUN_E2E=1 \
   go test -test.count=1 -timeout 600s -v ./tests/e2e -run '^TestE2EScript/scenarios/tracker/tracker-iampartialpolicy$'
 
-echo "Running some fixtures tests for iam tests with the gcptracker..."
+echo "Running fixture and sample iam tests with the gcptracker..."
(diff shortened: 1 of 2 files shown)


jingyih (commented):
Do we need a separate presubmit workflow for the same tests but with GCP tracker enabled? Can we just enable the resources one by one in the existing test suite?

dev/ci/presubmits/tracker-tests
```diff
@@ -26,9 +26,13 @@ export KUBEBUILDER_ASSETS=$(go run sigs.k8s.io/controller-runtime/tools/setup-en
...
-# FG for gcptracker usage
+# Flag gate for gcptracker usage
 export KCC_RECONCILE_FLAG_GATE=USE_DEPENDENCY_TRACKER
 
 echo "Running scenario tests for gcptracker..."
 GOLDEN_REQUEST_CHECKS=1 E2E_KUBE_TARGET=envtest E2E_GCP_TARGET=mock RUN_E2E=1 \
   go test -test.count=1 -timeout 600s -v ./tests/e2e -run '^TestE2EScript/scenarios/tracker/tracker-iampartialpolicy$'
+
+echo "Running fixture and sample iam tests with the gcptracker..."
+export RUN_TESTS='TestAllInSeries/fixtures/iam|TestAllInSeries/fixtures/fullalloydbcluster|TestAllInSeries/fixtures/noteboolinstance-full|TestAllInSeries/samples/alloydbbackup|TestAllInSeries/samples/alloydbcluster|TestAllInSeries/samples/iampartialpolicy|TestAllInSeries/samples/notebookinstance'
```
jingyih, line 37: Shall we add a test here when we enable GCP tracker on that resource?

jingyih (commented):
/lgtm
/approve

PR #4612, opened 2025-05-22
doc: update API generation guide with multi-resource example

jingyih:


docs/develop-resources/deep-dives/2-define-apis.md
```diff
@@ -32,6 +32,16 @@ go run main.go generate-types \
      --resource StorageNotification:Notification
 ```
 
+Note that if there are existing resources under the same API version, you should update the command to include all resources. For example:
```
yuwenma, line 35: what happens if user does not include the existing resources?   It is an ambiguous CLI design, ideally we don't want to require listing all the resources.

jingyih: If existing resources are not included in the command, the regenerated types will not include the types for existing resources. This has been the behavior since last year.

PR #4616, opened 2025-05-22
promote speech to beta

yuwenma:
LLM generated:
* Full promotion process including 1. api, controller, fuzz, test suites changes, 2. google doc and sample generation based on scifi 5-release.md step
* Manually fixed GCP specialties like deprecated fields handling.

We promoted the resource to Beta but not expose all the api fields. This is because KCC lacks confidence on those APIs: the speech GCP server only returns 400 without any additional information on what goes wrong. From KCC LLM brutal force testing, there are tons of combinations possible. Until the GCP service returns meaningful API message, KCC cannot expose those APIs to its users.


jingyih (commented):
Are we planning to add more test coverage later? I’m not expecting full coverage, but the current test cases are purely minimal.

scripts/generate-google3-docs/resource-reference/_toc.yaml
yuwenma, line 1: the reference doc is purely LLM generated following the guide in https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/develop-resources/deep-dives/5-releases.md (with some modifications, PR on the way)

yuwenma:
This PR is ready for another look @jingyih

apis/speech/v1beta1/phraseset_types.go
```diff
@@ -40,7 +40,7 @@ type SpeechPhraseSetSpec struct {
 	//  binary search approach to finding the optimal value for your use case as
 	//  well as adding phrases both with and without boost to your requests.
 	// +kcc:proto:field=google.cloud.speech.v2.PhraseSet.boost
-	Boost *string `json:"boost,omitempty"`
+	/* NOTYET Boost *string `json:"boost,omitempty"` */
```
jingyih, line 43: If we are removing fields, we should do it in alpha first?

jingyih: Removing fields is a backward-incompatible change. To avoid the need for a conversion webhook, we typically remove fields while the CRD is still in alpha. This ensures the schema remains unchanged when promoting to beta.

yuwenma: removed the fields from Alpha (KCC users does not roll back from Beta to Alpha  ¯\_(ツ)_/¯ ).

jingyih (commented):
/lgtm
/approve

PR #4621, opened 2025-05-22
chore: optimizing deletion logic in direct controller when the underlying GCP resource does not exist

gemmahou:
### Change description

Direct controller should be able to remove resources when their underlying GCP resources no longer exist, without making extra API calls.

### Tests you have done


maqiuyujoyce (commented):
Thank you for the improvement, @gemmahou ! Though I thought during our discussion, the main problem we wanted to solve is when the parent is deleted before the current resource, and the deletion will be stuck at the call of AdapterForObject()? Or did I miss anything?

Issue #3619, opened 2025-02-06
bug: resolving refs/v1bbeta1 before reference resource is ready


acpana:
Most (if not all) references under [`refs/v1beta1`](https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/5f861baf4cace3a9bc9f9b8753652e45e1056f2a/apis/refs/v1beta1) contain a logical issue where we can attempt to resolve a reference without knowing that the k8s-resource for that resource is actually READY to be resolved. Consider:

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/5f861baf4cace3a9bc9f9b8753652e45e1056f2a/apis/refs/v1beta1/bigqueryref.go#L43-L101

This can basically lead to extra calls to GCP when the underlying dependency is not actually ready.

This issue has come up in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/3422#discussion_r1941916723 . We should be weary of this too when moving on to the new reference style and treating migrations (e.g.from TF), especially on acquisition on existing resource where `status.ExternalRef` may not be populated.

jingyih:
I had a brief chat with @yuwenma right after our discussion on this. "Not waiting for the reference to be READY" may be working as intended? I'll defer to @yuwenma on this.

jingyih:
For reference, the behavior of resolving resource reference in TF controller:

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/62fa783d8223280b1679b10d672ced7d7e07b449/pkg/krmtotf/references.go#L133

ldanielmadariaga:
This item has been open for >12 months. As part of scheduled repository maintenance, this Issue/PR will be closed. If this is still a critical ask, please re-open and work with your Google Cloud Account team to prioritize it with GCP KCC Product Management.

PR #4628, opened 2025-05-23
chore: refactor bigquery references

gemmahou:
### Change description

Fixes #

### Tests you have done


gemmahou:
The refactoring won't impact resources that reference the updated ones (BigQueryDataset, BigQueryTable) because the existing tests cover all relevant references (datasetRef, tableRef), and the HTTP logs are unchanged. The only exception(untested reference) is BigQueryAnalyticsHubListing .spec.source.bigQueryDatasetSource.selectedResources[].tableRef.

However we are not able to test this field for now. I added a comment to the test yaml:
https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4628/commits/0d546fec2e114ca6ea56c2426c260cac314c75d8

apis/bigquery/v1beta1/table_reference.go
```diff
@@ -71,10 +72,30 @@ func (r *TableRef) NormalizedExternal(ctx context.Context, reader client.Reader,
 		return "", fmt.Errorf("reading referenced %s %s: %w", BigQueryTableGVK, key, err)
 	}
 	// Get external from status.externalRef. This is the most trustworthy place.
-	actualExternalRef, _, err := unstructured.NestedString(u.Object, "status", "externalRef")
+	actualExternalRef, found, err := unstructured.NestedString(u.Object, "status", "externalRef")
 	if err != nil {
 		return "", fmt.Errorf("reading status.externalRef: %w", err)
 	}
+	// For non-direct resources, there's no status.externalRef
+	if !found {
```
jingyih, line 80: This feels like a common pattern we need to handle when migrating BETA resources. We should investigate if it is possible to build a general helper function to ensure best practice. Related: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/65b84c3dba0bfe378ec521f70df07977cfc9efc9/apis/bigquery/v1beta1/dataset_reference.go#L78-L100

jingyih: /cc @acpana

gemmahou: Yes, for resources without status.externalRef, we'll need to perform additional resolution(verify resource ready, resolve its parent, etc). I can follow up this in another task, as this seems to impact other existing references: https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/5171

jingyih: The current implementation does not check for "resource ready". Whether or not we should check for resource ready is pending discussion in #3619, and I am not aware if a conclusion has been made.

PR #4629, opened 2025-05-23
Increase test coverage for BigQueryTable

maqiuyujoyce:
### Change description

Supported more fields in test cases for BigQueryTable.

Note that the goal is for full coverage, but we probably need some help from the API team to achieve it due to all kinds of errors we get when running the test, e.g. https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/b58ddf50adff7060116c0a36b766042071edab67/pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-externaldata/create.yaml#L24

### Tests you have done


pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-full/_http.log
```diff
@@ -624,6 +624,9 @@ X-Xss-Protection: 0
 
 {
   "creationTime": "123456789",
+  "encryptionConfiguration": {
```
maqiuyujoyce, line 627: This is to match mock behavior with the actual behavior.

anhdle-sso:
after offline discussion, we agree to leave the error message in the create.yaml as a form of TODO for other test cases. The the current external data test case, it should be rename to externaldata-csv. We need to add other test cases around google sheet, biglake,... later.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-externaldata/update.yaml
```diff
@@ -0,0 +1,52 @@
...
+      {"name": "intField", "type": "INTEGER", "mode": "REQUIRED"},
+      {"name": "stringField1", "type": "STRING", "mode": "NULLABLE"},
+      {"name": "stringField2", "type": "STRING", "mode": "NULLABLE"},
+      {"name": "stringField3", "type": "STRING", "mode": "NULLABLE"}
+    ]
+  expirationTime: 2064442632000
+  externalDataConfiguration:
+    autodetect: false
+    sourceUris:
+      - gs://cloud-samples-data/bigquery/us-states/us-states-by-date.csv
```
anhdle-sso: would it be possible to have another file with different delimiter and other options to fully test update.

maqiuyujoyce: Yes, but probably not now.

1. Under gs://cloud-samples-data/bigquery/, I didn't find a csv file with a different delimiter after brief search. Considering this is likely a sample folder owned by bigquery, I don't think we should add any data to it. We probably need to create a sample data bucket by ourselves.
2. I actually updated all the fields in the `csvOptions` in update.yaml, even if the delimiter and possibly other options don't match what's in the source file, but the request was still successfully. I'm guessing we either didn't use the right configuration, or we used it but the error was not returned by PUT.

So I prefer we spend the time for a better test case after we understand the situation.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-externaldata-csv/create.yaml
```diff
@@ -0,0 +1,97 @@
...
+    ]
+  expirationTime: 2064441632000 # Some time in 2035
+  # Error: require_partition_filter has been set for a hive-partitioned table without using the HivePartitioningOptions.
+  #  requirePartitionFilter: true
+  # Error: Cloud KMS based encryption is not supported for views or BigQuery Omni external tables.
+  # Error: Access Denied: BigQuery BigQuery: Cloud KMS Error: Permission 'cloudkms.cryptoKeyVersions.useToEncrypt' denied on resource 'projects/[projectID]/locations/us-central1/keyRings/kmskeyring-pyvirqtmyhr7p3y/cryptoKeys/kmscryptokey-1-pyvirqtmyhr7p3y' (or it may not exist).
+  #  encryptionConfiguration:
+  #    kmsKeyRef:
+  #      name: kmscryptokey-${uniqueId}
+  externalDataConfiguration:
```
anhdle-sso, line 55: within externalDataConfiguration, there is a schema field also. I wonder setting either or both spec.schema and spec.externalDataConfiguration.schema would produce different behavior

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-externaldata-csv/create.yaml
```diff
@@ -0,0 +1,97 @@
...
+    autodetect: true
```
anhdle-sso, line 56: setting autodetect: true lets us get away with not setting spec.externalDataConfiguration.schema; and spec.schema has no effect with the API

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-externaldata-csv/create.yaml
```diff
@@ -0,0 +1,97 @@
...
+    #    hivePartitioningOptions:
+    #      mode: AUTO
+    #      sourceUriPrefix: gs://cloud-samples-data/bigquery/us-states/
+    #      requirePartitionFilter: true
+    # Error: GoogleSheetsOptions can only be specified if storage format is GOOGLE_SHEETS.
+    #    googleSheetsOptions:
+    #      skipLeadingRows: 1
+    #      range: "Sheet1!A1:B20"
+    avroOptions:
+      useAvroLogicalTypes: true
```
anhdle-sso: it is weird that avro and parquet options throw no errors as the input is csv. Perhaps, API has the check only on GOOGLE SHEET or Big Lake.

maqiuyujoyce: Agreed it's weird. Do you think we should add validation in the direct controller for better UX? I think it could be a future improvement.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-externaldata-csv/create.yaml
maqiuyujoyce: Discussed offline.

Unsetting both works and setting both will have issues during update. Unsetting both for now to test a valid use case.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerytable/bigquerytable-externaldata-csv/create.yaml
maqiuyujoyce: Done.

tests/e2e/ratcheting.go
```diff
@@ -103,6 +103,11 @@ func ShouldTestRereconiliation(t *testing.T, primaryResource *unstructured.Unstr
 	case schema.GroupKind{Group: "bigqueryreservation.cnrm.cloud.google.com", Kind: "BigQueryReservationReservation"}:
 	case schema.GroupKind{Group: "bigquery.cnrm.cloud.google.com", Kind: "BigQueryRoutine"}:
 	case schema.GroupKind{Group: "bigquery.cnrm.cloud.google.com", Kind: "BigQueryTable"}:
+		if testName == "bigquerytable-ignore-schema-changes" {
```
jingyih, line 106: Can we add some comments about why excluding this specific test case?

jingyih (commented):
/lgtm
/approve

PR #4642, opened 2025-05-29
feat: promote spannerbackupschedule to beta

xiaoweim:
### Change description

Fixes #

### Tests you have done


apis/spanner/v1beta1/types.generated.go
```diff
@@ -79,6 +79,57 @@ type AutoscalingConfig_AutoscalingLimits struct {
...
+	CreationWindow *string `json:"creationWindow,omitempty"`
+}
+
+// +kcc:proto=google.spanner.admin.database.v1.FullBackupSpec
+type FullBackupSpec struct {
+}
+
+// +kcc:proto=google.spanner.admin.database.v1.IncrementalBackupSpec
+type IncrementalBackupSpec struct {
+}
```
yuwenma, line 131: Are they empty struct?

pkg/controller/direct/spanner/backupschedule_controller.go
yuwenma, line 1: I cannot add inline comments:
- in ln 177, we cannot just return nil because it will leave the status empty for first time acquire. Suggest make a status update even if no change.
- In ln 180, does it mean those output-only proto do not have the OUTPUT_ONLY field_behavior thus you manually mark them?
-

pkg/test/resourcefixture/testdata/basic/spanner/v1beta1/spannerbackupschedule/spannerbackupschedules-basic/update.yaml
yuwenma, line 1: Is fullBackupSpec immutable? If not, suggest changing fullBackupSpec to incrementalBackupSpec?

apis/spanner/v1beta1/types.generated.go
xiaoweim: Yes. The REST API documentation specifies that [FullBackupSpec](https://cloud.google.com/spanner/docs/reference/rest/v1/projects.instances.databases.backupSchedules#fullbackupspec) and [IncrementalBackupSpec](https://cloud.google.com/spanner/docs/reference/rest/v1/projects.instances.databases.backupSchedules#incrementalbackupspec) has no fields.

pkg/controller/direct/spanner/backupschedule_controller.go
xiaoweim: In line 180, yes. Those fields will be populated with a value after the resource is created. But in an update request, since they are output only, user will not input any value to it so the proto field will be empty. When we do a diff, it will include these output only field and try to update the field to empty value and thus return error.

pkg/test/resourcefixture/testdata/basic/spanner/v1beta1/spannerbackupschedule/spannerbackupschedules-basic/update.yaml
xiaoweim: We can update it to `incrementalBackupSpec` but it only changes the future behavior, the previous backup would still follow the `fullBackupSpec`. It is more like to add the `incrementalBackupSpec`.

yuwenma: For mutable field, we want the `update.yaml` shows a different value than `create.yaml` so that the test can cover the `adaptor.Update` for this specific field.  Could you revert the change in create.yaml?

xiaoweim: I am not sure which change in `create.yaml` you are referring to? I changed the `fullBackupSpec ` to `incrementalBackupSpec ` in the `pkg/test/resourcefixture/testdata/basic/spanner/v1beta1/spannerbackupschedule/spannerbackupschedules-fully/update.yaml` and the logs shows that the we sent the request with the updated value but the actual value was not changed. Do you mean revert the change in `update.yaml`?

apis/spanner/v1beta1/backupschedule_reference.go
```diff
@@ -0,0 +1,83 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ refsv1beta1.ExternalNormalizer = &BackupScheduleRef{}
+
+// BackupScheduleRef defines the resource reference to SpannerBackupSchedule, which "External" field
+// holds the GCP identifier for the KRM object.
+type BackupScheduleRef struct {
+	// A reference to an externally managed SpannerBackupSchedule resource.
+	// Should be in the format "projects/{{projectID}}/locations/{{location}}/backupschedules/{{backupscheduleID}}".
```
anhdle-sso: nit: backupSchedules

PR #4645, opened 2025-05-29
feat: CRD partial installation

jingyih:
## PR description
* Add back #4396 
* Introduce separate release flavors: `full`, which includes all customize CRDs; and `stable`, which excludes customize CRDs that are only available in alpha.
* Add graceful handling for scenarios where a core alpha CRD is not installed in the cluster.

## Manifest rendering (before vs. after this PR):
 * [kustomize build operator/config/default/](https://www.diffchecker.com/YOp6KLCs/)
 * [kustomize build operator/config/autopilot/](https://www.diffchecker.com/vNBJaixl/)
 * [kustomize build operator/config/gke-addon/](https://www.diffchecker.com/fZBRCsDy/)

### DIFF
--- a/operator/pkg/controllers/configconnector/testdata/golden/crdpartialinstallation/_expected.yaml
+++ b/operator/pkg/controllers/configconnector/testdata/golden/crdpartialinstallation/_expected.yaml
@@ -1,0 +1,4155 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations: (removed)
+  labels: (removed)
+  name: computeaddresses.compute.cnrm.cloud.google.com
+spec: (removed)
+status:
+  acceptedNames:
+    kind: ""
+    plural: ""
+  conditions: []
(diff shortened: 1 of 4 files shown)


operator/Makefile
```diff
@@ -41,10 +41,39 @@ test: generate fmt vet manifests
...
+	cp config/crd/generated/core.cnrm.cloud.google.com_*.yaml config/crd/base/bases/
+	cp config/crd/generated/customize.core.cnrm.cloud.google.com_*.yaml config/crd/overlays/full/bases/
+
+# Build stable overlay (manually picked CRDs)
+.PHONY: manifests-stable
+manifests-stable: generate-crds
+	rm -f config/crd/base/bases/*.yaml
+	rm -f config/crd/overlays/stable/bases/*.yaml
+
+	cp config/crd/generated/core.cnrm.cloud.google.com_*.yaml config/crd/base/bases/
```
jingyih, line 67: This is how to pick which customize CRDs to be included in the stable flavor.

PR #4646, opened 2025-05-29
Improve KCC release scripts

xiaoweim:
### Change description

Add release script to update the latest manifests to all previous versions except for controller.

Fixes #

### Tests you have done

### DIFF
--- a/dev/tasks/propose-tag
+++ b/dev/tasks/propose-tag
@@ -128,13 +128,16 @@ fi
 echo "Found versions:"
 printf '%s\n' "${versions[@]}"
 
-# STALE_VERSION=$(printf "%s\n" "${versions[@]}" | sort -V | tail -n 4 | head -n 1)
-echo "stale version: ${STALE_VERSION}"
+if [[ -n "${STALE_VERSION:-}" ]]; then
+  # STALE_VERSION=$(printf "%s\n" "${versions[@]}" | sort -V | tail -n 4 | head -n 1)
+  echo "stale version: ${STALE_VERSION}"
 
-# Remove the STALE_VERSION package for standard and autopilot
-rm -rf ${REPO_ROOT}/operator/autopilot-channels/packages/configconnector/${STALE_VERSION}
-rm -rf ${REPO_ROOT}/operator/channels/packages/configconnector/${STALE_VERSION}
+  # Remove the STALE_VERSION package for standard and autopilot
+  rm -rf ${REPO_ROOT}/operator/autopilot-channels/packages/configconnector/${STALE_VERSION}
+  rm -rf ${REPO_ROOT}/operator/channels/packages/configconnector/${STALE_VERSION}
+fi
 
 # Format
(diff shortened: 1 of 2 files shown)


dev/tasks/propose-tag
```diff
@@ -128,13 +128,16 @@ fi
 echo "Found versions:"
 printf '%s\n' "${versions[@]}"
 
-# STALE_VERSION=$(printf "%s\n" "${versions[@]}" | sort -V | tail -n 4 | head -n 1)
-echo "stale version: ${STALE_VERSION}"
+if [[ -n "${STALE_VERSION:-}" ]]; then
+  # STALE_VERSION=$(printf "%s\n" "${versions[@]}" | sort -V | tail -n 4 | head -n 1)
```
cheftako: We may want to just remove this comment.

dev/tasks/update-manifests.sh
```diff
@@ -0,0 +1,84 @@
...
+  cp ${REPO_ROOT}/operator/channels/packages/configconnector/${VERSION}/cluster/workload-identity/0-cnrm-system.yaml ${REPO_ROOT}/operator/channels/packages/configconnector/${version}/cluster/workload-identity/0-cnrm-system.yaml
+  cp ${REPO_ROOT}/operator/channels/packages/configconnector/${VERSION}/namespaced/0-cnrm-system.yaml ${REPO_ROOT}/operator/channels/packages/configconnector/${version}/namespaced/0-cnrm-system.yaml
+
+    # Keep the controller version and update the rest of the manifest
+  yq 'select(.kind == "StatefulSet")' ${REPO_ROOT}/operator/channels/packages/configconnector/${version}/namespaced/per-namespace-components.yaml > temp_statefulset.yaml
+  cp ${REPO_ROOT}/operator/channels/packages/configconnector/${VERSION}/namespaced/per-namespace-components.yaml ${REPO_ROOT}/operator/channels/packages/configconnector/${version}/namespaced/per-namespace-components.yaml
+  yq 'select(.kind != "StatefulSet")' -i ${REPO_ROOT}/operator/channels/packages/configconnector/${version}/namespaced/per-namespace-components.yaml
```
cheftako, line 77: Is there a reason why we some time us yq -i (in place replacement) and sometimes right via a temp file?

xiaoweim: I am using yq to select the operator(statefulset) from the older version manifests and merge it with the newest version of the manifests. I am not sure if there's a better way to extract just the `statefulset` within the older manifests using other tools. 

The reason that we store the older version statefulset in a temp file, is to copy the newest manifest (excluding the newest operator/statefulset) to the older version yaml file. Since when we copy over the content, we lose the older version manifests, I stored the older version statefulset in a temp file

cheftako: Personally I would just always go through a temp file if we sometimes need to. 1) Easier to read/modify the more consistent we are in how we do things. 2) If we need to go through a temp file for some cases today, there is a very good chance more of the jq commands will need to in the future.

Issue #4274, opened 2025-04-04
FAQs and Behavior description for Config Connector and GCP labels


yuwenma:
### Checklist

- [x] I did not find a related open enhancement request.
- [x] I understand that enhancement requests filed in the GitHub repository are by default low priority.
- [x] If this request is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Describe the feature or resource

# Overview

[GCP `labels`](https://cloudplatform.googleblog.com/2018/06/Labelling-and-grouping-your-Google-Cloud-Platform-resources.html) helps you to group the GCP resources for a variety of use cases. 

Config Connector supports the GCP labels in two ways:

- **Kubernetes labels**   (`.metadata.labels`)

- **Kubernetes Specification**  (`.spec.labels`)

- For both methods, Config Connector also adds a `managed-by-cnrm` label to the GCP resource.

# FAQs and troubleshooting

## 1. Why are my Kubernetes labels not added to some GCP resources?

Reason 1: Every Config Connector resource has the Kubernetes labels (the `.metadata.labels` field), but not every GCP resources support labels.  When you use `kustomize`, `helm` or any others GitOps tools to add Kubernetes labels to your Config Connector resources,  it will only add to the GCP resource if 1.) the GCP server supports labels, and 2.) the Config Connector controller pass the label to the server. 

Reason 2: Some Config Connector resources only support labels via the Specification (`.spec.labels`), some only support via Kubernetes labels, and some support both with a condition (migration with API backward compatibility). 

## 2. Why does some labels fail in Config Connector , but work well in Cloud SDK and Cloud Console?

[Kubernetes label](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set) has a different format requirements than [the GCP label format](https://cloudplatform.googleblog.com/2018/06/Labelling-and-grouping-your-Google-Cloud-Platform-resources.html). For example, if you has doc (`.`) in the key or value, it is valid in K8s but not in GCP. 

Note: Some Config Connector resources filter out the labels that contain dot before applied to the GCP labels. But this also cause some confusions (see below). 

## 3. Why does the same label can only apply to some GCP resources but not others? 

Just like not every Google service support labels, some service support labels with a different format requirements. You may hit different problems depending on the actual service requirements under the hood.

## 4. Why does Config Connector have both Kubernetes labels and Specification labels? It is so confusing ...

[...]

PR #4647, opened 2025-05-29
feat: Support spec.labels in direct ComputeForwardingRule

gemmahou:
### Change description

Fixes b/414960623

Related GH issue: https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/4274

### Tests you have done


pkg/controller/direct/compute/forwardingrule/forwardingrule_controller.go
```diff
@@ -106,9 +106,6 @@ func (m *forwardingRuleModel) AdapterForObject(ctx context.Context, reader clien
 	// Get location
 	location := obj.Spec.Location
 
-	// Set label managed-by-cnrm: true
-	obj.ObjectMeta.Labels["managed-by-cnrm"] = "true"
```
justinsb, line 110: This looks like a breaking change?

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrulehttps-direct/_http.log
```diff
@@ -1014,9 +1014,7 @@ X-Goog-Request-Params: project=${projectId}
   "IPAddress": "8.8.8.8",
   "description": "A global forwarding rule",
   "labels": {
-    "cnrm-test": "true",
```
justinsb, line 1017: Not sure what is going on with this label...

pkg/controller/direct/compute/forwardingrule/forwardingrule_controller.go
```diff
@@ -64,7 +64,7 @@ func (r *ForwardingRuleReconcileGate) ShouldReconcile(o *unstructured.Unstructur
 		return false
 	}
 	// Run the direct reconciler only when spec.Target.GoogleAPIsBundle is specified
-	return obj.Spec.Target != nil && obj.Spec.Target.GoogleAPIsBundle != nil
+	return (obj.Spec.Target != nil && obj.Spec.Target.GoogleAPIsBundle != nil) || obj.Spec.Labels != nil
```
justinsb, line 67: This is a relatively complicated behavior. Labels come from metadata.labels, unless we set target or googleapisbundle or spec.labels, in which case we only use spec.labels

justinsb:
Have we written down how spec.labels and metadata.labels is going to work @yuwenma

gemmahou:
> Have we written down how spec.labels and metadata.labels is going to work @yuwenma

Yes, here's the issue documented: https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/4274, do we need a more formal version?

pkg/controller/direct/compute/forwardingrule/forwardingrule_controller.go
gemmahou: `obj.Spec.Labels != nil` prevents breaking the TF controller. Since the `spec.labels` field was added only for the direct controller, this check is necessary in case TF users mistakenly set this field and break their use case.

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrulehttps-direct/_http.log
gemmahou: This change is expected. cnrm-test is a system label, which should not be included in API calls. We only process user-specified labels from spec.labels(when it's direct controller).

pkg/controller/direct/compute/forwardingrule/forwardingrule_controller.go
gemmahou: This PR separates labels into system-managed (metadata.labels) and user-defined (spec.labels).

My understanding is that `managed-by-cnrm` should be a system label, so we should avoid setting it in individual controllers?

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/regionalcomputeforwardingrulepsc-direct/_http.log
```diff
@@ -1473,88 +1469,6 @@ X-Xss-Protection: 0
 
 ---
 
-POST https://compute.googleapis.com/compute/v1/projects/example-project-01/regions/us-central1/forwardingRules/${forwardingRuleID}/setLabels
```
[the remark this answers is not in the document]
gemmahou: Both labels `cnrm-test` and `managed-by-cnrm` are considered system labels. With the change in the PR, we no longer include them in API requests. In this test, there're no user-specified labels, so the removal of `setLabels` request is expected.

apis/compute/v1beta1/forwardingrule_types.go
```diff
@@ -151,6 +151,19 @@ type ComputeForwardingRuleSpec struct {
...
+	// Label keys must be between 1 and 63 characters long, have a UTF-8 encoding
+	// of maximum 128 bytes, and must conform to the following PCRE regular
+	// expression: `[\p{Ll}\p{Lo}][\p{Ll}\p{Lo}\p{N}_-]{0,62}`
+	//
+	// Label values must be between 0 and 63 characters long, have a UTF-8
+	// encoding of maximum 128 bytes, and must conform to the following PCRE
+	// regular expression: `[\p{Ll}\p{Lo}\p{N}_-]{0,63}`
+	//
+	// No more than 64 labels can be assigned to a given resource.
+	Labels map[string]string `json:"labels,omitempty"`
```
anhdle-sso, line 165: Can you add some comment to use this along with the opt-in annotation?

pkg/controller/direct/compute/forwardingrule/forwardingrule_controller.go
anhdle-sso: Suggest to break this up for readability and add more comment

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrule-direct/update.yaml
```diff
@@ -15,14 +15,14 @@
...
-  labels:
-    label-one: "value-one"
   annotations:
     alpha.cnrm.cloud.google.com/reconciler: "direct"
   name: computeglobalforwardingrule-${uniqueId}
 spec:
   description: "A global forwarding rule"
   location: global
+  labels:
+    label-one: "value-one"
```
anhdle-sso: can you add some change in the labels in the update if the api allows the labels

apis/compute/v1beta1/forwardingrule_types.go
gemmahou: I feel it's better to mention this in our release notes, instead of in the field description. 
For example: https://screenshot.googleplex.com/7mRqDZQjL7BTqW3.png

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrule-direct/update.yaml
gemmahou: Done, I put the test change and new logs in the 2nd commit for easier review.

pkg/controller/direct/compute/forwardingrule/forwardingrule_controller.go
gemmahou: Done.

pkg/controller/direct/compute/forwardingrule/forwardingrule_controller.go
anhdle-sso: Since we are getting GCP object Labels from spec.labels instead of  metadata.labels; the label `managed-by-cnrm` should not cause API issue anymore. Let's keep this label for the KRM object, since we don't know if there is some other pipelines (watching over the KRM object) use this.

gemmahou: SG. I added back `["managed-by-cnrm"] = "true"` in forwardingrule controller, but eventually I feel this system label should be handled in a higher level, maybe in directbase controller.

gemmahou:
> > Yes, here's the issue documented: #4274, do we need a more formal version?
> 
> I think we want to make the decision of whether to merge this PR (or similar PRs that introduce spec.labels) into a deterministic decision. Ideally we would have test coverage, but I personally don't think 4274 makes merge/no-merge here deterministic (either for the reviewers or the author)

I believe we've already established a pattern for this. I looked into the history here and ComputeForwardingRule isn't the first resource to face this. In v1.132.0, we added spec.labels to two other resources because their API calls were being blocked by invalid system labels. I feel that this is our "unofficial" solution to unblock customer for now, and we may want to formalize it. 

cc @yuwenma Could you take a look at this and share your thoughts on next steps? Thanks!

gemmahou:
/hold 
Put this on hold, pending design doc for the common issue.

PR #4657, opened 2025-06-02
Run controller managers in separate namespaces

dshnayder:
# Change Description
## Summary

This change allows Kubernetes Config Connector (KCC) controller managers to be deployed in dedicated, tenant-specific namespaces instead of the default `cnrm-system` namespace. This significantly improves the security of multi-tenant environments by providing better namespace-level isolation and reducing the blast radius of potential security incidents.

## Motivation

Currently, all KCC controller managers operate within the shared `cnrm-system` namespace, irrespective of the specific ConfigConnectorContext or tenant they serve. This centralized deployment model presents a security concern: it increases the potential blast radius in the event of a security compromise or critical failure affecting one controller manager. A vulnerability exploited within `cnrm-system` could potentially impact all tenants managed by that KCC installation due to the shared execution environment.

The goal of this refactoring is to mitigate this risk by allowing controller managers to be deployed into distinct, dedicated namespaces. This architectural change facilitates stronger isolation between tenants or configurations, thereby reducing the scope of impact from potential security incidents or operational issues within a single manager instance. This approach aligns with best practices for secure multi-tenant system design in Kubernetes.

## Implementation Details

The new functionality is enabled by the `--enable-manager-namespace` flag, passed to the KCC operator's `manager` pod (/configconnector-operator/manager). 

```yaml
   spec:
      containers:
      - args:
        - --local-repo=/configconnector-operator/autopilot-channels
        - --enable-manager-namespace
        command:
        - /configconnector-operator/manager
```

When this flag is enabled, KCC will check for the optional `managerNamespace` field in the `ConfigConnectorContext` resource. If this field is present, KCC will deploy the corresponding controller manager and its associated ServiceAccount into the specified namespace, rather than the default `cnrm-system`.

- Spec field: `managerNamespace`
When this field is present on a `ConfigConnectorContext` resource specification and it is not empty, the system is instructed not to deploy the corresponding controller manager components into the default `cnrm-system` namespace. Instead, the namespace specified in `managerNamespace` is used. The `managerNamespace` will be created if it doesn’t exists. The selection of this specification field as the configuration mechanism adheres to standard Kubernetes practices for declarative configuration, allowing behavior modification without altering the core ConfigConnector API specification.

**Example**:
Consider a ConfigConnectorContext resource, located in the namespace t1234-tenant0-provider:
```yaml
kind: ConfigConnectorContext
metadata:
  name: configconnectorcontext.core.cnrm.cloud.google.com
  namespace: t1234-tenant0-provider
  labels:
    tenancy.gke.io/access-level: supervisor
    tenancy.gke.io/project: t1234
    tenancy.gke.io/tenant: t1234-tenant0
spec:
  googleServiceAccount: "gcp@tpname.iam.gserviceaccount.com"
  stateIntoSpec: Absent
  managerNamespace: t1234-tenant0-supervisor
```
If the ConfigConnectorContext resource includes the specification field `managerNamespace: t1234-tenant0-supervisor`, then the controller manager StatefulSet and its ServiceAccount will be created within the `t1234-tenant0-supervisor` namespace.

If the new functionality is not activated, then the specification field “managerNamespace” on each ConfigConnectorContext must not be specified. If “managerNamespace” is specified when the feature is globally disabled, then ConfigConnectorContext reconciliation will fail with an error message indicating that “managerNamespace” is not supported.

If the new functionality is activated then the specification field “managerNamespace” on each ConfigConnectorContext must be specified. If “managerNamespace” is not specified when the feature is globally enabled, then ConfigConnectorContext reconciliation will fail with an error message indicating that “managerNamespace” must be populated.

**Summary of configuration**
| Parameter | Type | Description |
| :--- | :--- | :--- |
| `--enable-manager-namespace` | Command-line flag | Top-level enabler for the new functionality. Must be set for the `managerNamespace` field to have any effect. |
| `managerNamespace` | `ConfigConnectorContext` spec field | Configures the target namespace for the controller manager when the feature is enabled. |

**Scope and Impact**

This enhancement is required for multi-tenant GKE deployments, where strict tenant isolation is often a requirement.

**Positive Impacts**:
- Significantly reduces the blast radius by isolating controller managers per tenant or configuration context.
- Aligns with best practices for secure multi-tenant system design in Kubernetes

**Negative Impacts**:
- When the feature is enabled via the gating flag `--enable-manager-namespace` and a new Custom Resource is created, an additional request from `unmanageddetector` to `apiserver` required to get specification of `ConfigConnectorContext` resource.

### Tests

- Created GKE MT cluster.
- Installed KCC in namespaced mode.
-  Created a tenant.
- Created a ConfigConnectorContext in the provider namespace. Verified that StatefulSets/Pod cnrm-controller-manager-XXX are created in the supervisor namespace.
- Created PubSubTopic in the provider namespace. Verified that corresponding object is created in GCP.
- Removed the ConfigConnectorContext. Verified that StatefulSets/Pod cnrm-controller-manager-XXX are deleted
- Recreated ConfigConnectorContext and PubSubTopic, then deleted tenant. Verified that GCP object, ConfigConnectorContext, namespaces and tenant were deleted successfully.
- Deleted ConfigConnector
- Verified that all StatefulSets/Pods are removed
- Created ConfigConnector in cluster mode
- Verified that StatefulSets/Pod cnrm-controller-manager are created in the `cnrm-system` namespace

### DIFF
--- a/cmd/unmanageddetector/main.go
+++ b/cmd/unmanageddetector/main.go
@@ -52,14 +52,14 @@ func main() {
 	flag.CommandLine.AddGoFlagSet(goflag.CommandLine)
 	flag.BoolVar(&enablePprof, "enable-pprof", false, "Enable the pprof server.")
 	flag.IntVar(&pprofPort, "pprof-port", 6060, "The port that the pprof server binds to if enabled.")
-	flag.StringVar(&managerNamespaceIsolation, k8s.ManagerNamespaceIsolationFlag, k8s.ManagerNamespaceIsolationShared, "'shared' if all controller managers run in shared 'cnrm-system' namespace, 'dedicated' if controller managers run in dedicated namespace. Default is 'shared'")
+	flag.StringVar(&managerNamespaceIsolation, k8s.ManagerNamespaceIsolationFlag, k8s.ManagerNamespaceIsolationShared, fmt.Sprintf("'%s' if all controller managers run in shared 'cnrm-system' namespace, '%s' if controller managers run in dedicated namespace. Default is '%s'", k8s.ManagerNamespaceIsolationShared, k8s.ManagerNamespaceIsolationDedicated, k8s.ManagerNamespaceIsolationShared))
 	flag.Parse()
 
 	switch managerNamespaceIsolation {
 	case k8s.ManagerNamespaceIsolationShared, k8s.ManagerNamespaceIsolationDedicated:
 		unmanageddetector.ManagerNamespaceIsolation = managerNamespaceIsolation
 	default:
-		logging.Fatal(fmt.Errorf("unknown value for --manager-namespace-isolation: %s, expected '%s' or '%s'", managerNamespaceIsolation, k8s.ManagerNamespaceIsolationShared, k8s.ManagerNamespaceIsolationDedicated), "error starting unmanaged detector")
+		logging.Fatal(fmt.Errorf("unknown value for --%s: %s, expected '%s' or '%s'", k8s.ManagerNamespaceIsolationFlag, managerNamespaceIsolation, k8s.ManagerNamespaceIsolationShared, k8s.ManagerNamespaceIsolationDedicated), "error starting unmanaged detector")
 	}
 
 	// Enable packages using the Kubernetes controller-runtime logging package to log
--- a/cmd/unmanageddetector/main.go
+++ b/cmd/unmanageddetector/main.go
@@ -26,6 +26,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/registration"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/unmanageddetector"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/gcp/profiler"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/logging"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/ready"
 
@@ -45,13 +46,21 @@ func main() {
 
 	var enablePprof bool
 	var pprofPort int
+	var managerNamespaceIsolation string
 
 	profiler.AddFlag(flag.CommandLine)
 	flag.CommandLine.AddGoFlagSet(goflag.CommandLine)
 	flag.BoolVar(&enablePprof, "enable-pprof", false, "Enable the pprof server.")
 	flag.IntVar(&pprofPort, "pprof-port", 6060, "The port that the pprof server binds to if enabled.")
-	flag.BoolVar(&unmanageddetector.EnableManagerNamespace, "enable-manager-namespace", false, "Enable running controller managers in separate namespaces.")
+	flag.StringVar(&managerNamespaceIsolation, k8s.ManagerNamespaceIsolationFlag, k8s.ManagerNamespaceIsolationShared, "'shared' if all controller managers run in shared 'cnrm-system' namespace, 'dedicated' if controller managers run in dedicated namespace. Default is 'shared'")
 	flag.Parse()
+
+	switch managerNamespaceIsolation {
+	case k8s.ManagerNamespaceIsolationShared, k8s.ManagerNamespaceIsolationDedicated:
+		unmanageddetector.ManagerNamespaceIsolation = managerNamespaceIsolation
+	default:
+		logging.Fatal(fmt.Errorf("unknown value for --manager-namespace-isolation: %s, expected '%s' or '%s'", managerNamespaceIsolation, k8s.ManagerNamespaceIsolationShared, k8s.ManagerNamespaceIsolationDedicated), "error starting unmanaged detector")
+	}
 
 	// Enable packages using the Kubernetes controller-runtime logging package to log
 	logging.SetupLogger()
--- a/cmd/unmanageddetector/main.go
+++ b/cmd/unmanageddetector/main.go
@@ -24,6 +24,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/kccmanager/nocache"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/registration"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/unmanageddetector"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/gcp/profiler"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/logging"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/ready"
@@ -49,6 +50,7 @@ func main() {
 	flag.CommandLine.AddGoFlagSet(goflag.CommandLine)
 	flag.BoolVar(&enablePprof, "enable-pprof", false, "Enable the pprof server.")
 	flag.IntVar(&pprofPort, "pprof-port", 6060, "The port that the pprof server binds to if enabled.")
+	flag.BoolVar(&unmanageddetector.EnableManagerNamespace, "enable-manager-namespace", false, "Enable running controller managers in separate namespaces.")
 	flag.Parse()
 
 	// Enable packages using the Kubernetes controller-runtime logging package to log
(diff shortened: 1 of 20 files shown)


operator/cmd/gke_addon_poststart/main.go
```diff
@@ -39,6 +40,20 @@ spec:
...
 
+const defaultConfigConnectorNamespaceSeparated = `
+apiVersion: core.cnrm.cloud.google.com/v1beta1
+kind: ConfigConnector
+metadata:
+  name: configconnector.core.cnrm.cloud.google.com
+  labels:
+    tenancy.gke.io/access-level: supervisor
+    tenancy.gke.io/project: no-project
+    tenancy.gke.io/tenant: no-tenant
```
yuwenma: Can I ask more about these 3 labels? I want to know the life cycle of them, e.g. how is it used/modified, etc.

operator/cmd/gke_addon_poststart/main.go
```diff
@@ -49,22 +64,31 @@ var configConnectorResource = schema.GroupVersionResource{
...
-func createDefaultConfigConnector(ctx context.Context, dynamicClient dynamic.Interface) error {
+func createDefaultConfigConnector(ctx context.Context, dynamicClient dynamic.Interface, managerNamespaceSuffix string) error {
 	u := &unstructured.Unstructured{}
-	b := []byte(defaultConfigConnector)
+	var b []byte
+	if managerNamespaceSuffix == "" {
+		b = []byte(defaultConfigConnector)
+	} else {
+		b = []byte(fmt.Sprintf(defaultConfigConnectorNamespaceSeparated, managerNamespaceSuffix))
+	}
```
yuwenma: Suggest making this a Unstructured level operations, especially considering that a string formatting is used. 
```go
u.SetLabels(
map[string]string{
  ...
}
)
```

pkg/k8s/constants.go
```diff
@@ -68,6 +68,7 @@ const (
 	NamespaceEnvVar                      = "NAMESPACE"
 	ImmediateReconcileRequestsBufferSize = 10000
 	MaxNumResourceWatcherRoutines        = 10000
+	ManagerNamespaceSuffixLabel          = "cnrm.cloud.google.com/manager-namespace-suffix"
```
yuwenma: Suggest using a new field to CCC `spec` instead of labels. Spec is specifically for user configurations, and it is more managable and traceble compared to labels.

dshnayder: Are you suggesting `spec` on CCC (ConfigConnectorContext)? There should be a global flag that indicates:
1. If managers need to run in separate namespaces
2. How to name the namespaces )what should be the suffix of the manager's namespace)
This suggest that CC (ConfigConnector) should have the `ManagerNamespaceSuffix`

operator/cmd/gke_addon_poststart/main.go
dshnayder: Answered in corporate chat

operator/cmd/gke_addon_poststart/main.go
dshnayder: Done

pkg/k8s/constants.go
dshnayder: Done

cmd/unmanageddetector/main.go
```diff
@@ -49,6 +50,7 @@ func main() {
 	flag.CommandLine.AddGoFlagSet(goflag.CommandLine)
 	flag.BoolVar(&enablePprof, "enable-pprof", false, "Enable the pprof server.")
 	flag.IntVar(&pprofPort, "pprof-port", 6060, "The port that the pprof server binds to if enabled.")
+	flag.BoolVar(&unmanageddetector.EnableManagerNamespace, "enable-manager-namespace", false, "Enable running controller managers in separate namespaces.")
```
cheftako: *Not blocking* 
I dislike bools in APIs. It always seems like something which is 2 state eventually becomes 3+ state. Right now the manager "run-mode" is either "cluster-only" or "namespace-only". I can imagine wanting something in the future like "cluster-namespace-hybrid" or even "project". Switching from a bool to an enum seems like future proofing.

operator/cmd/manager/main.go
```diff
@@ -58,6 +59,7 @@ func main() {
 
 	imagePrefix := os.Getenv("IMAGE_PREFIX")
 	flag.StringVar(&imagePrefix, "image-prefix", imagePrefix, "Remap container images to pull from the specified registry or mirror.")
+	flag.BoolVar(&enableManagerNamespace, "enable-manager-namespace", false, "Enable running controller managers in separate namespaces.")
```
cheftako: Not blocking
I dislike bools in APIs. It always seems like something which is 2 state eventually becomes 3+ state. Right now the manager "run-mode" is either "cluster-only" or "namespace-only". I can imagine wanting something in the future like "cluster-namespace-hybrid" or even "project". Switching from a bool to an enum seems like future proofing.

operator/pkg/apis/core/v1beta1/configconnectorcontext_types.go
```diff
@@ -69,6 +69,17 @@ type ConfigConnectorContextSpec struct {
 	//+kubebuilder:validation:Enum=Reconciling;Paused
 	//+kubebuilder:validation:Optional
 	Actuation ActuationMode `json:"actuationMode,omitempty"`
+
+	// ManagerNamespace instructs Config Connector to deploy
+	// controller managers and related resources in the namespace
+	// specified as 'ManagerNamespace' instead of standard 'cnrm-system'
+	// The specified manager namespace may exist before the corresponding
+	// ConfigConnectorContext is created.
+	// If the specified namespace does not exist then it will be created
```
cheftako, line 78: Have we ensured we are giving/can give KCC write permissions on namespaces?

operator/pkg/controllers/configconnector/configconnector_controller.go
```diff
@@ -949,3 +966,76 @@ func checkForDuplicateWebhooks(webhooks []customizev1beta1.WebhookCustomizationS
 	}
 	return nil
 }
+
+func (r *Reconciler) transformPerNamespaceComponents() declarative.ObjectTransform {
+	return func(ctx context.Context, o declarative.DeclarativeObject, m *manifest.Objects) error {
```
cheftako, line 971: Can we add an early out check based on not r.enableManagerNamespace?

operator/pkg/controllers/configconnectorcontext/configconnectorcontext_controller.go
```diff
@@ -303,6 +328,11 @@ func (r *Reconciler) finalizeCCContextDeletion(ctx context.Context, ccc *corev1b
 	if err := cluster.DeleteNamespaceID(ctx, k8s.OperatorNamespaceIDConfigMapNN, r.client, ccc.Namespace); err != nil {
 		return err
 	}
+	if r.enableManagerNamespace && ccc.Spec.ManagerNamespace != "" {
```
cheftako: This seems like it should not happen. Can we add a warning log and/or maybe a status message in the CCC?  I understand why your doing it but we certainly should not be taking such a drastic step quietly.

cheftako:
Can we add a check for not enableManagerNamespace that either prevent the managerNamepsace field or report a status warning?

operator/pkg/controllers/configconnector/configconnector_controller.go
dshnayder: This function conditionally added to the list of transformation based on `r.enableManagerNamespace`, see `operator/pkg/controllers/configconnector/configconnector_controller.go` line `149`. So its completely skipped if new functionality is not enabled.

operator/pkg/apis/core/v1beta1/configconnectorcontext_types.go
dshnayder: KCC already creates and deletes `cnrm-system`, so permission to create/delete namespaces already granted.

cmd/unmanageddetector/main.go
dshnayder: Changed from boolean to string everywhere in the code.

operator/cmd/manager/main.go
dshnayder: Changed from boolean to string everywhere in the code.

operator/pkg/controllers/configconnectorcontext/configconnectorcontext_controller.go
dshnayder: Updated code to fail reconciliation of CCC if global mode and `spec.ManagerNamespace` are not compatible.

dshnayder:
> Can we add a check for not enableManagerNamespace that either prevent the managerNamepsace field or report a status warning?

Updated code to fail reconciliation of CCC if global mode and spec.ManagerNamespace are not compatible.

PR #4658, opened 2025-06-02
feat: direct iampartialpolicy controller

acpana:
This converts the iampartialpolicy controller to the new direct approach.

NOTE: this controller is not turned on so no changes to existing IAM handling are made.

Follow ups

* fuzzer test (will come out as a separate PR)
* export using THIS direct approach https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/4667

### DIFF
--- a/.github/workflows/presubmit.yaml
+++ b/.github/workflows/presubmit.yaml
@@ -111,6 +111,17 @@ jobs:
       - name: "Run vcr tests"
         run: |
           ./scripts/github-actions/tests-e2e-fixtures-vcr.sh
+  tests-e2e-direct-iam:
+    runs-on: ubuntu-22.04
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+      - name: "Run direct iam tests"
+        run: |
+          ./scripts/github-actions/ga-iam-direct-tests.sh
   build-images:
     runs-on: ubuntu-22.04
     timeout-minutes: 60
--- a/apis/iam/v1beta1/policy_types.go
+++ b/apis/iam/v1beta1/policy_types.go
@@ -20,6 +20,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/apis/k8s/v1alpha1"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 )
 
 type ResourceReference struct {
@@ -31,6 +32,13 @@ type ResourceReference struct {
 	APIVersion string `json:"apiVersion,omitempty"`
 	// The external name of the referenced resource
 	External string `json:"external,omitempty"`
+}
+
+func (r *ResourceReference) GetNamespacedName() types.NamespacedName {
+	return types.NamespacedName{
+		Name:      r.Name,
+		Namespace: r.Namespace,
+	}
 }
 
 type Member string
--- a/apis/iam/v1beta1/register.go
+++ b/apis/iam/v1beta1/register.go
@@ -42,12 +42,12 @@ var (
 		Kind:    reflect.TypeOf(IAMPolicy{}).Name(),
 	}
 
+	IAMPartialPolicyGVK = schema.GroupVersionKind{
+		Group:   SchemeGroupVersion.Group,
+		Version: SchemeGroupVersion.Version,
+		Kind:    reflect.TypeOf(IAMPartialPolicy{}).Name(),
+	}
 	// NOT YET
-	// IAMPartialPolicyGVK = schema.GroupVersionKind{
-	// 	Group:   SchemeGroupVersion.Group,
-	// 	Version: SchemeGroupVersion.Version,
-	// 	Kind:    reflect.TypeOf(IAMPartialPolicy{}).Name(),
-	// }
 	// IAMPolicyMemberGVK = schema.GroupVersionKind{
 	// 	Group:   SchemeGroupVersion.Group,
 	// 	Version: SchemeGroupVersion.Version,
@@ -58,5 +58,5 @@ var (
 	// 	Version: SchemeGroupVersion.Version,
 	// 	Kind:    reflect.TypeOf(IAMAuditConfig{}).Name(),
 	// }
-	//IAMAPIVersion = SchemeGroupVersion.String()
+	IAMAPIVersion = SchemeGroupVersion.String()
 )
--- a/pkg/controller/direct/directbase/directbase_controller.go
+++ b/pkg/controller/direct/directbase/directbase_controller.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/apis/core/v1beta1"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/kccstate"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/jitter"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/lifecyclehandler"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/metrics"
@@ -97,6 +98,7 @@ func NewReconciler(mgr manager.Manager, immediateReconcileRequests chan event.Ge
 		},
 		jitterGenerator: deps.JitterGenerator,
 		defaulters:      deps.Defaulters,
+		iamDeps:         deps.AdapterDeps,
 	}
 	return &r, nil
 }
@@ -152,10 +154,22 @@ func add(mgr manager.Manager, r *DirectReconciler, reconcilePredicate predicate.
 
 var _ reconcile.Reconciler = &DirectReconciler{}
 
+// Reconciler dependencies.
 type Deps struct {
 	Defaulters         []k8s.Defaulter
 	JitterGenerator    jitter.Generator
 	ReconcilePredicate predicate.Predicate
+
+	// There are Dependencies for Adapters in particular (not the reconcilers)
+	AdapterDeps *IAMAdapterDeps
+}
+
+// TODO(kcc-team): we want to remove these in the future
+// In a world where there are no TF or DCL based IAM resources, we don't need
+// these dependencies and the "special" IAM model.
+type IAMAdapterDeps struct {
+	ControllerDeps *controller.Deps
+	KubeClient     client.Client
 }
 
 // DirectReconciler is a reconciler for reconciling resources that support the Model/Adapter pattern.
@@ -175,6 +189,8 @@ type DirectReconciler struct {
 	controllerName string
 
 	defaulters []k8s.Defaulter
+	// For IAM controllers
+	iamDeps *IAMAdapterDeps
 }
 
 type reconcileContext struct {
@@ -302,13 +318,19 @@ func (r *reconcileContext) doReconcile(ctx context.Context, u *unstructured.Unst
 		}
 	}
 
-	adapter, err := r.Reconciler.model.AdapterForObject(ctx, r.Reconciler.Client, u)
-	if err != nil {
-		if unwrappedErr, ok := lifecyclehandler.CausedByUnresolvableDeps(err); ok {
+	var adapter Adapter
+	var adapteErr error
+	if iamModel, ok := r.Reconciler.model.(IAMModel); ok {
+		adapter, adapteErr = iamModel.IAMAdapterForObject(ctx, r.Reconciler.Client, u, r.Reconciler.iamDeps)
+	} else {
+		adapter, adapteErr = r.Reconciler.model.AdapterForObject(ctx, r.Reconciler.Client, u)
+	}
+	if adapteErr != nil {
+		if unwrappedErr, ok := lifecyclehandler.CausedByUnresolvableDeps(adapteErr); ok {
 			logger.Info(unwrappedErr.Error(), "resource", k8s.GetNamespacedName(u))
 			return r.handleUnresolvableDeps(ctx, u, unwrappedErr)
 		}
-		return false, r.handleUpdateFailed(ctx, u, err)
+		return false, r.handleUpdateFailed(ctx, u, adapteErr)
 	}
 
 	// To create, update or delete the GCP object, we need to get the GCP object first.
@@ -318,9 +340,21 @@ func (r *reconcileContext) doReconcile(ctx context.Context, u *unstructured.Unst
 	if err != nil {
 		if unwrappedErr, ok := lifecyclehandler.CausedByUnresolvableDeps(err); ok {
 			logger.Info(unwrappedErr.Error(), "resource", k8s.GetNamespacedName(u))
-			return r.handleUnresolvableDeps(ctx, u, unwrappedErr)
+
+			// if we have failed to "FIND" the resource and that is caused by an error deemed to be about its unresolved
+			// dependencies BUT the object is being deleted right now AND its dependecies have been deleted, then we will
+			// never encounter a "ready" condition for its dependencies
+			if !u.GetDeletionTimestamp().IsZero() {
+				resource, err := toK8sResource(u)
+				if err != nil {
+					return false, fmt.Errorf("error converting k8s resource while handling unresolvable dependencies event: %w", err)
+				}
+
+				return true, r.Reconciler.HandleUnresolvableDeps(ctx, resource, unwrappedErr)
+			}
+
+			return false, r.handleUpdateFailed(ctx, u, err)
 		}
-		return false, r.handleUpdateFailed(ctx, u, err)
 	}
 
 	defer execution.RecoverWithInternalError(&err)
--- a/pkg/controller/direct/directbase/directbase_controller.go
+++ b/pkg/controller/direct/directbase/directbase_controller.go
@@ -320,9 +320,11 @@ func (r *reconcileContext) doReconcile(ctx context.Context, u *unstructured.Unst
 
 	var adapter Adapter
 	var adapteErr error
-	if iamModel, ok := r.Reconciler.model.(IAMModel); ok {
-		adapter, adapteErr = iamModel.IAMAdapterForObject(ctx, r.Reconciler.Client, u, r.Reconciler.iamDeps)
-	} else {
+	switch m := r.Reconciler.model.(type) {
+	case IAMModel:
+		adapter, adapteErr = m.IAMAdapterForObject(ctx, r.Reconciler.Client, u, r.Reconciler.iamDeps)
+	default:
+		// The default case handles any other type that implements the base model interface.
 		adapter, adapteErr = r.Reconciler.model.AdapterForObject(ctx, r.Reconciler.Client, u)
 	}
 	if adapteErr != nil {
--- a/pkg/controller/direct/directbase/directbase_controller.go
+++ b/pkg/controller/direct/directbase/directbase_controller.go
@@ -98,7 +98,7 @@ func NewReconciler(mgr manager.Manager, immediateReconcileRequests chan event.Ge
 		},
 		jitterGenerator: deps.JitterGenerator,
 		defaulters:      deps.Defaulters,
-		iamDeps:         deps.AdapterDeps,
+		iamDeps:         deps.IAMAdapterDeps,
 	}
 	return &r, nil
 }
@@ -161,7 +161,7 @@ type Deps struct {
 	ReconcilePredicate predicate.Predicate
 
 	// There are Dependencies for Adapters in particular (not the reconcilers)
-	AdapterDeps *IAMAdapterDeps
+	IAMAdapterDeps *IAMAdapterDeps
 }
 
 // TODO(kcc-team): we want to remove these in the future
--- a/pkg/controller/direct/iam/iambindings.go
+++ b/pkg/controller/direct/iam/iambindings.go
@@ -1,0 +1,252 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package iam
+
+import (
+	"fmt"
+	"sort"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/apis/iam/v1beta1"
+)
+
+type MemberIdentityResolver interface {
+	Resolve(v1beta1.Member, *v1beta1.MemberSource, string) (string, error)
+}
+
+type iamBindingKey struct {
+	Role      string
+	Condition v1beta1.IAMCondition
+}
+
+// ComputePartialPolicyWithMergedBindings returns the IAMPartialPolicy that results after the user's intent (as specified by the input
+// IAMPartialPolicy) is merged with the underlying IAM policy (as specified by the input IAMPolicy). This function also resolves all memberFrom
+// fields to member fields and ensures the returned IAMPartialPolicy only contains member fields.
+
+// The status.AllBindings in the returned IAMPartialPolicy reflects a mix of user specified bindings and the existing bindings associated with the GCP resource.
+// The merge strategy takes effect on the member level with {role, condition} tuples as keys.
+// The status.LastAppliedBindings in the returned IAMPartialPolicy reflects a list of canonical bindings that specified by users.
+func ComputePartialPolicyWithMergedBindings(partialPolicy *v1beta1.IAMPartialPolicy, livePolicy *v1beta1.IAMPolicy, resolver MemberIdentityResolver) (*v1beta1.IAMPartialPolicy, error) {
+	desiredPartialPolicy := partialPolicy.DeepCopy()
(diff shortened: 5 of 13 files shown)


pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -318,9 +340,21 @@ func (r *reconcileContext) doReconcile(ctx context.Context, u *unstructured.Unst
 	if err != nil {
 		if unwrappedErr, ok := lifecyclehandler.CausedByUnresolvableDeps(err); ok {
 			logger.Info(unwrappedErr.Error(), "resource", k8s.GetNamespacedName(u))
-			return r.handleUnresolvableDeps(ctx, u, unwrappedErr)
+
+			// if we have failed to "FIND" the resource and that is caused by an error deemed to be about its unresolved
+			// dependencies BUT the object is being deleted right now AND its dependecies have been deleted, then we will
+			// never encounter a "ready" condition for its dependencies
```
acpana, line 348: note I pulled out this change in its own PR too for visibility-- https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4666

pkg/controller/direct/iam/iambindings.go
acpana, line 1: this and _test file are basically just moved here for now. In the future, we can do away with the old files too.

EDIT: moved from https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/iam/partialpolicy/bindings.go

In the near future we'd be removing those files.

pkg/controller/direct/iam/mappings.go
```diff
@@ -0,0 +1,188 @@
...
+
+	"cloud.google.com/go/iam/apiv1/iampb"
+	expr "google.golang.org/genproto/googleapis/type/expr"
+
+	newiamv1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/apis/iam/v1beta1"
+	oldiamv1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/apis/iam/v1beta1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
+)
+
+func IAMPolicySpec_ToProto(_ *direct.MapContext, in *oldiamv1beta1.IAMPolicySpec) *iampb.Policy {
```
acpana, line 35: note that bc the iamclient still takes the old API group, we run into this issue where we have to translate between "new" and "old". In the near future, we will harmonize these and get rid of the "old" mentions.

pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -318,9 +340,21 @@ func (r *reconcileContext) doReconcile(ctx context.Context, u *unstructured.Unst
...
+			if !u.GetDeletionTimestamp().IsZero() {
+				resource, err := toK8sResource(u)
+				if err != nil {
+					return false, fmt.Errorf("error converting k8s resource while handling unresolvable dependencies event: %w", err)
+				}
+
+				return true, r.Reconciler.HandleUnresolvableDeps(ctx, resource, unwrappedErr)
+			} else {
```
cheftako: Newest golang linter will complain about the "return } else" pattern.

pkg/controller/direct/iam/iambindings.go
```diff
@@ -0,0 +1,252 @@
...
+func mergeBindingSlices(bindingSlice1, bindingSlice2 []v1beta1.IAMPolicyBinding) []v1beta1.IAMPolicyBinding {
+	mergedBindings := make([]v1beta1.IAMPolicyBinding, 0)
+	mergedBindings = append(mergedBindings, bindingSlice1...)
+	mergedBindings = append(mergedBindings, bindingSlice2...)
+	return mergeBindingsWithSameRoleAndCondition(mergedBindings)
+}
+
+func mergeBindings(bindings []v1beta1.IAMPolicyBinding) map[iamBindingKey]v1beta1.IAMPolicyBinding {
+	bindingMap := make(map[iamBindingKey]v1beta1.IAMPolicyBinding)
+	for _, a := range bindings {
```
cheftako: Not asking for long variable names but something better than a single letter. 'binding' seems better than 'a', 'key' seems better than 'k'.

pkg/controller/direct/iam/mappings.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	return out
+}
+
+// mapV1Beta1LogTypeToProto converts a string log type from v1beta1
+// to the pb.AuditLogConfig_LogType enum (which is a wrapped int).
+func mapV1Beta1LogTypeToProto(logTypeString string) iampb.AuditLogConfig_LogType {
+	switch strings.ToUpper(logTypeString) {
+	case "LOG_TYPE_UNSPECIFIED":
```
cheftako: Repeated use of constant strings suggest we should define the actual constants.

pkg/controller/registration/registration_controller.go
```diff
@@ -327,9 +341,20 @@ func registerDefaultController(ctx context.Context, r *ReconcileRegistration, co
...
+						TfProvider:   r.provider,
+						TfLoader:     r.smLoader,
+						DclConfig:    r.dclConfig,
+						DclConverter: r.dclConverter,
+					},
+				},
 			}
 			if err := directbase.AddController(r.mgr, gvk, model, deps); err != nil {
-				return nil, fmt.Errorf("error adding direct controller for %v to a manager: %w", crd.Spec.Names.Kind, err)
+				return nil, fmt.Errorf("2 error adding direct controller for %v to a manager: %w", crd.Spec.Names.Kind, err)
```
cheftako: Typo?

pkg/controller/direct/iam/iambindings.go
```diff
@@ -0,0 +1,252 @@
...
+		res = append(res, k)
+	}
+	sort.Slice(res, func(i, j int) bool {
+		return res[i] < res[j]
+	})
+	return res
+}
+
+func sortBindingSlice(bindings []v1beta1.IAMPolicyBinding) {
+	sort.Slice(bindings, func(i, j int) bool {
```
cheftako: Note: in many situations, the newer [slices.SortFunc](https://pkg.go.dev/slices#SortFunc) function is more ergonomic and runs faster.

pkg/controller/direct/iam/iambindings.go
acpana: since we basically moved this file, I'd suggest we keep it as is for now 🙏🏼

pkg/controller/direct/iam/iambindings.go
acpana: ditto about moving the iambindings, iambindings_test files 🙏🏼

.github/workflows/presubmit.yaml
```diff
@@ -111,6 +111,17 @@ jobs:
...
+    runs-on: ubuntu-22.04
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+      - name: "Run direct iam tests"
+        run: |
+          ./scripts/github-actions/ga-iam-direct-tests.sh
```
maqiuyujoyce, line 124: Nice! Though, are we running the existing hand-written ones in presubmit right now? Wondered if we should do that to achieve backwards compatibility.

Not a blocker for this PR.

apis/iam/v1beta1/policy_types.go
```diff
@@ -20,6 +20,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/apis/k8s/v1alpha1"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 )
 
 type ResourceReference struct {
```
maqiuyujoyce, line 26: Nit: Should this type be in its own go file for better readability?

Not a blocker.

apis/iam/v1beta1/register.go
```diff
@@ -58,5 +58,5 @@ var (
 	// 	Version: SchemeGroupVersion.Version,
 	// 	Kind:    reflect.TypeOf(IAMAuditConfig{}).Name(),
 	// }
-	//IAMAPIVersion = SchemeGroupVersion.String()
+	IAMAPIVersion = SchemeGroupVersion.String()
```
maqiuyujoyce, line 61: Just curious: Is it used?

pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -152,10 +154,22 @@ func add(mgr manager.Manager, r *DirectReconciler, reconcilePredicate predicate.
...
 var _ reconcile.Reconciler = &DirectReconciler{}
 
+// Reconciler dependencies.
 type Deps struct {
 	Defaulters         []k8s.Defaulter
 	JitterGenerator    jitter.Generator
 	ReconcilePredicate predicate.Predicate
+
+	// There are Dependencies for Adapters in particular (not the reconcilers)
+	AdapterDeps *IAMAdapterDeps
```
maqiuyujoyce: Nit: Maybe call it iamAdapterDeps? I thought this AdapterDeps field is for any kind of adapters / generic use before I saw the type definition.

Not a blocker.

pkg/controller/direct/directbase/directbase_controller.go
maqiuyujoyce: Thank you for fixing this issue! I think this is a common issue (see detailed explanation below). I wondered if in the future we can make resource reference check part of the IAMAdapterForObject. With that, we won't need the additional special handling here, but instead, only after the AdapterForObject call.

=== explanation===

It's a known issue that when dependency is deleted, the deletion of the target direct resource will stuck with an error from the AdapterForObject() call.

For example, we need to resolve the instance identity in AdapterForObject() for AlloyDBInstance: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/alloydb/instance_controller.go#L114

During the identity resolution, we need to be able to read the AlloyDBCluster, otherwise, an error will be returned: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/alloydb/v1beta1/instance_identity.go#L61

If the AlloyDBCluster is deleted before the AlloyDBInstance does, a NotFound error will be returned and the direct base will requeue the reconciliation to wait for the dependency to be ready, but it'll never be ready.

So the exact logic you add here should be added after AdapterForObject() call.

pkg/controller/direct/iam/partialpolicy_controller.go
```diff
@@ -0,0 +1,403 @@
...
+
+	newiamv1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/apis/iam/v1beta1"
+	oldiamv1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/apis/iam/v1beta1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/config"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
+	kcciamclient "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/iam/iamclient"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
+	"github.com/go-logr/logr"
```
maqiuyujoyce: Nit: Clean up the import grouping?

pkg/controller/direct/iam/partialpolicy_controller.go
```diff
@@ -0,0 +1,403 @@
...
+	}
+
+	if deps.KubeClient == nil {
+		return fmt.Errorf("KubeClient is nil")
+	}
+	if deps.ControllerDeps == nil {
+		return fmt.Errorf("ControllerDeps is nil")
+	}
+
+	if deps.ControllerDeps.TfProvider == nil {
```
maqiuyujoyce, line 73: Not a blocker for this PR, but I'd suggest the field name `TfProvider` should be changed to `TFProvider` because `TF` is an abbreviation.

pkg/controller/direct/iam/partialpolicy_controller.go
```diff
@@ -0,0 +1,403 @@
...
+		return fmt.Errorf("TfProvider is nil")
+	}
+	if deps.ControllerDeps.TfLoader == nil {
```
maqiuyujoyce, line 76: Not a blocker. Similarly, `TFLoader`.

pkg/controller/direct/iam/partialpolicy_controller.go
```diff
@@ -0,0 +1,403 @@
...
+		return fmt.Errorf("TfLoader is nil")
+	}
+	if deps.ControllerDeps.DclConverter == nil {
```
maqiuyujoyce, line 79: Not a blocker. `DCL` is an acronym so I suggest we use `DCLConverter` as the name.

pkg/controller/direct/iam/partialpolicy_controller.go
```diff
@@ -0,0 +1,403 @@
...
+
+func (m *modelIAMPartialPolicy) AdapterForObject(ctx context.Context, reader client.Reader, u *unstructured.Unstructured) (directbase.Adapter, error) {
+	return nil, fmt.Errorf("AdapterForObject not supported for IAMPartialPolicy, call IAMAdapterForObject")
+}
+
+func (m *modelIAMPartialPolicy) AdapterForURL(ctx context.Context, url string) (directbase.Adapter, error) {
+	// // IAMPartialPolicy is not identified or managed via a direct GCP URL in the same way
+	// // other resources are.
+	// return nil, fmt.Errorf("AdapterForURL not supported for IAMPartialPolicy")
+	return nil, nil
```
maqiuyujoyce, line 110: The comment makes sense to me, so I'm curious why you decided to use `return nil, nil` instead.

pkg/controller/direct/iam/partialpolicy_controller.go
```diff
@@ -0,0 +1,403 @@
...
+		return false, fmt.Errorf("getting IAM policy for %v: %w", k8s.GetNamespacedName(a.desired), err)
+	}
+
+	a.actualReferencedResourcePolicy = IAMPolicySpec_ToProto(mapCtx, &oldKRMPolicy.Spec)
+	return true, nil
+}
+
+// IAMMemberIdentityResolver helps to resolve referenced member identity
+type IAMMemberIdentityResolver struct {
+	Iamclient *kcciamclient.IAMClient
```
maqiuyujoyce: Nit: Capitalize for acronym, i.e. `IAMClient`.

pkg/controller/direct/iam/partialpolicy_controller.go
```diff
@@ -0,0 +1,403 @@
...
+	}
+	return nil
+}
+
+func (a *IAMPartialPolicyAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOperation) error {
+	log := getLogger(ctx)
+	log.V(2).Info("updating/applying IAMPartialPolicy", "name", k8s.GetNamespacedName(a.desired))
+	mapCtx := &direct.MapContext{}
+
+	if a.actualReferencedResourcePolicy == nil {
```
maqiuyujoyce, line 236: IIUC, when the actual policy is nil, Create() should be triggered and this branch should never be hit.

Though in the Find() option, it returns true regardless if policy is nil or not. Looks like the users can never create policy when the referenced resource exist but the IAM policy doesn't? Or did I misunderstand anything?

pkg/controller/direct/iam/partialpolicy_controller.go
```diff
@@ -0,0 +1,403 @@
...
+	} else {
+		log.V(2).Info("no change in GCP IAM policy needed", "name", k8s.GetNamespacedName(a.desired))
+	}
+
+	// Always update KRM status to reflect observed generation and latest computed bindings,
+	// even if GCP didn't change, as spec or other KRM details might have.
+	newKRMStatus := newiamv1beta1.IAMPartialPolicyStatus{
+		ObservedGeneration:  a.desired.GetGeneration(),
+		LastAppliedBindings: desiredPartialPolicyWithStatus.Status.LastAppliedBindings,
+		AllBindings:         desiredPartialPolicyWithStatus.Status.AllBindings,
```
maqiuyujoyce, line 272: Just curious, is it the existing behavior? I think the IAM policy can have up to 1500 bindings, so the status could be hard to read sometimes. I wondered if we want to filter the `status` by member or etc when the size is large, or making this field filterable by user configuration.

scripts/github-actions/ga-iam-direct-tests.sh
```diff
@@ -0,0 +1,39 @@
...
+# Running direct tests
+export ONLY_TEST_APIGROUPS=$(IFS=,; echo "${APIGROUPS[*]}")
+${REPO_ROOT}/dev/tasks/run-e2e
+
+# Running direct and tracker tests
+export KCC_RECONCILE_FLAG_GATE=USE_DEPENDENCY_TRACKER
+unset ONLY_TEST_APIGROUPS
+
+echo "Running fixture and sample iam tests with the gcptracker..."
+export RUN_TESTS='TestAllInSeries/samples/iampartialpolicy'
```
maqiuyujoyce, line 38: Looks like there is no http log or exported resource validation for samples test. I feel we are not making the best of the mock tests here. Maybe we should support them for better backwards compatibility checking?

apis/iam/v1beta1/register.go
acpana: Yes as part of the policy controller -- https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4658/files#diff-5c3b74a6231aab3cadfa559bfdb60d8330214bbdb7c3c32ec1d754d08bf9589eR340

scripts/github-actions/ga-iam-direct-tests.sh
acpana: Yes we can look into that but I'd like to keep that change separate from this PR 🤔

pkg/controller/direct/iam/partialpolicy_controller.go
acpana: yes I think this is what the current hand rolled controller is handling the status.

From https://cloud.google.com/iam/docs/reference/rest/v1/Policy, yes GCP enforces the following:

"
The bindings in a Policy can refer to up to 1,500 principals; up to 250 of these principals can be Google groups. Each occurrence of a principal counts towards these limits. For example, if the bindings grant 50 different roles to user:alice@example.com, and not to any other principal, then you can add another 1,450 principals to the bindings in the Policy.
"

We don't currently have in code checks/ shifted left checks but I'm happy to explore/ add them later in a follow on PR.

pkg/controller/direct/iam/partialpolicy_controller.go
acpana: we should make sure the linter handles this!

.github/workflows/presubmit.yaml
acpana: We are in as much as support a mock for this in presubmits


```
...
2025-06-06T23:05:59.3629971Z         --- SKIP: TestAllInSeries/fixtures/iamaccessboundarypolicy (0.00s)
2025-06-06T23:05:59.3630444Z         --- SKIP: TestAllInSeries/fixtures/orgrole (0.00s)
2025-06-06T23:05:59.3630886Z         --- SKIP: TestAllInSeries/fixtures/projectrole (0.00s)
2025-06-06T23:05:59.3631364Z         --- PASS: TestAllInSeries/fixtures/iamserviceaccount (15.67s)
2025-06-06T23:05:59.3631912Z         --- PASS: TestAllInSeries/fixtures/iamserviceaccount-disabled (11.43s)
2025-06-06T23:05:59.3632467Z         --- PASS: TestAllInSeries/fixtures/iamserviceaccount-full (15.56s)
2025-06-06T23:05:59.3632993Z         --- SKIP: TestAllInSeries/fixtures/iamserviceaccountkey (0.00s)
2025-06-06T23:05:59.3633490Z         --- SKIP: TestAllInSeries/fixtures/iamworkforcepool (0.00s)
2025-06-06T23:05:59.3634021Z         --- SKIP: TestAllInSeries/fixtures/oidcworkforcepoolprovider (0.00s)
2025-06-06T23:05:59.3634581Z         --- SKIP: TestAllInSeries/fixtures/samlworkforcepoolprovider (0.00s)
2025-06-06T23:05:59.3635124Z         --- SKIP: TestAllInSeries/fixtures/iamworkloadidentitypool (0.00s)
...
```

apis/iam/v1beta1/policy_types.go
acpana: Happy to split those out but would do it outside of this PR!

pkg/controller/direct/directbase/directbase_controller.go
maqiuyujoyce: @gemmahou FYI as this feels like part of the deletion logic improvement you're working on.

acpana: wouldn't it make more sense for us to add these checks in directbase too? As opposed to having to add the per controller checks?

pkg/controller/direct/iam/partialpolicy_controller.go
acpana: I'm game to do so but we should probably be touching the ControllerDeps member fields in a separate PR since that will make LOTS of changes .

pkg/controller/direct/iam/partialpolicy_controller.go
acpana: right this is part of the hack. The current wiring makes it such that if we have an adapter for the direct object we force the `export` to be used from the AdapterForURL func. If the adapter is nil, it simply means to use a different export logic (not direct). If the err is not nil, we fail the export all together! I think nil, nil is fine here for now but I/ we should figure out how to clean up that code path.

pkg/controller/direct/iam/partialpolicy_controller.go
acpana: Yes this branch shouldn't be hit but I think we should defensively check that there is a policy when Update() is called. In the `Create()` func we handle this scenario differently and we indeed create the policy!

.github/workflows/presubmit.yaml
maqiuyujoyce: Cool, looks like we only supports iamserviceaccount tests for now, but it seems to be a good start.

apis/iam/v1beta1/policy_types.go
maqiuyujoyce: SG.

pkg/controller/direct/directbase/directbase_controller.go
maqiuyujoyce: Yes! What I meant to suggest is:

1. We should add the check you added here to line 330 above because this can be a common issue after directbase calls AdapterForObject().
2. If IAM policy controllers check dependent resources (resourceRef) in IAMAdapterForObject() instead of Find() (because resourceRef is technically the parent of the IAM policy), then we don't need to add the check here (after Find() call) again.

I didn't suggest adding any per controller logic, rather suggesting moving the reference resolution logic from Find() to IAMAdapterForObject() for the IAM policy controller.

pkg/controller/direct/iam/partialpolicy_controller.go
maqiuyujoyce: SG.

pkg/controller/direct/iam/partialpolicy_controller.go
maqiuyujoyce: Gotcha, thx for the context.

pkg/controller/direct/iam/partialpolicy_controller.go
maqiuyujoyce: SG. I think we should consider doing some UX improvement for this resource in the future.

scripts/github-actions/ga-iam-direct-tests.sh
maqiuyujoyce: SG.

pkg/controller/direct/iam/partialpolicy_controller.go
maqiuyujoyce: Cool, I think defensive checking should be fine here.

Though we might want to avoid `a.actualReferencedResourcePolicy` to become `nil` and fail earlier if it happens. For example, if in Find(), `a.actualReferencedResourcePolicy` is set to `nil` without an error (as IAMPolicySpec_ToProto() can return nil), we probably want to return a ResourceNotFound error instead. This also matches the defensive coding philosophy.

As I mentioned in the other comment, I suggest we split the checking of the parent resource and the policy into IAMAdapterForObject() and Find() so that it can be consistent with how we handle other direct resources, but we can decide it later.

maqiuyujoyce (commented):
/lgtm

Overall looks good with a couple of questions/improvement thoughts.

/assign @cheftako 
Do you have any other comments?

pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -302,13 +318,19 @@ func (r *reconcileContext) doReconcile(ctx context.Context, u *unstructured.Unst
 		}
 	}
 
-	adapter, err := r.Reconciler.model.AdapterForObject(ctx, r.Reconciler.Client, u)
-	if err != nil {
-		if unwrappedErr, ok := lifecyclehandler.CausedByUnresolvableDeps(err); ok {
+	var adapter Adapter
+	var adapteErr error
+	if iamModel, ok := r.Reconciler.model.(IAMModel); ok {
```
cheftako: Personally I think this is more readable if kept in a negative coding style. Not blocking but worth considering.

pkg/controller/direct/iam/iambindings.go
```diff
@@ -0,0 +1,252 @@
+// Copyright 2022 Google LLC
```
cheftako: Not 2022.

pkg/controller/direct/iam/iambindings.go
cheftako: Please fix this. I am generally a proponent of "as long as it doesn't make things worse". However if we don't fix it now, we won't fix it. Its not showing as a move, its showing as you wrote this code and we are approving it as net new.

PR #4659, opened 2025-06-02
refactor: configurable poll intervals

acpana:
Might be temporary but I think we will start using a configurable interval, especially in dynamic tests (postsubmits) soon! Can also configure the jitter to reflect current behavior. Still TBD how we are going to handle the reconcile annotation.

### DIFF
--- a/pkg/controller/kccmanager/kccmanager.go
+++ b/pkg/controller/kccmanager/kccmanager.go
@@ -18,6 +18,8 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"os"
+	"strconv"
 	"time"
 
 	operatorv1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/apis/core/v1beta1"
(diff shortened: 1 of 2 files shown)


pkg/controller/kccmanager/kccmanager.go
```diff
@@ -185,8 +187,20 @@ func New(ctx context.Context, restConfig *rest.Config, cfg Config) (manager.Mana
 	}
 	rd.DependencyTracker = gcpwatch.NewDependencyTracker(fetcher)
 
+	pollInterval := 10 * time.Minute
+	if interval := os.Getenv("DEPENDENCY_TRACKER_POLL_INTERVAL"); interval != "" {
```
jingyih: If we only plan to support this in testing, maybe we call it `TEST_DEPENDENCY_TRACKER_POLL_INTERVAL`

pkg/controller/kccmanager/kccmanager.go
```diff
@@ -185,8 +187,20 @@ func New(ctx context.Context, restConfig *rest.Config, cfg Config) (manager.Mana
 	}
 	rd.DependencyTracker = gcpwatch.NewDependencyTracker(fetcher)
 
+	pollInterval := 10 * time.Minute
```
jingyih: Can we move the default poll interval value to `tracker.go`? I think it would be clearer if all default values for `PollConfig` are defined in the same file as `PollConfig`.

jingyih (commented):
/lgtm
/approve

PR #4660, opened 2025-06-03
chore: fix sample check

jingyih:
The original check is a too strict. Does not work when the repo is copied to g3.


yuwenma:
/lgtm
/approve 

Seems to be related to the legacy release tech debt. Thank you for the fix!

PR #4661, opened 2025-06-03
docs: add samples for speech

yuwenma:
### Change description

Fixes #

### Tests you have done


config/samples/resources/speechcustomclass/speechcustomclass.yaml
```diff
@@ -0,0 +1,25 @@
+# Copyright 2025 Google LLC
```
maqiuyujoyce: I think we want the sample file name to be in the format of `<service>_<api-version>_<kind>.yaml`: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/README.Samples.md#file-naming for better readability and consistency, and better searchability.

jingyih (commented):
I think you mentioned the format of the sample filename has to be "kind.yaml"? Do you have a pointer to the check that enforces this rule?

config/samples/resources/speechcustomclass/speechcustomclass.yaml
yuwenma: If using that name, it will fail `TestAllResourcesHaveSampleFileUsedForSnippets` test because it requires the test name to be `kind`.yaml.

yuwenma:
> I think you mentioned the format of the sample filename has to be "kind.yaml"? Do you have a pointer to the check that enforces this rule?

cc @maqiuyujoyce 
Obviously I misread the code, that the name is required to has kind.yaml as [*suffix"](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/68767303d98d5f7164232036e9cc519f48760311/pkg/snippet/snippetgeneration/snippetgeneration.go#L163) if only one sample is given (compared to a resource has multiple samples).

PR #4662, opened 2025-06-03
feat: name ctrl loggers

acpana:
I think it's useful to have these named as it helps with debugging, etc. 🤷🏼


jingyih (commented):
/hold cancel

No worries. My thought was that the underlying library is just implementation detail. A term like "DCL" may not mean much to customers and could potentially cause confusion. That said, we can merge it and see if there is any feedback.

PR #4663, opened 2025-06-03
change cl_pool to optional

ericzzzzzzz:
### Change Description

- Change this to ca_pool to optional


ericzzzzzzz:
@ericpang777

apis/securesourcemanager/v1alpha1/instance_types.go
```diff
@@ -95,7 +95,7 @@ type Instance_PrivateConfig struct {
 	// +kcc:proto:field=google.cloud.securesourcemanager.v1.Instance.PrivateConfig.is_private
 	IsPrivate *bool `json:"isPrivate,omitempty"`
 
-	// Required. Immutable. CA pool resource, resource must in the format of
+	// Optional. Immutable. CA pool resource, resource must in the format of
```
maqiuyujoyce, line 98: If it is optional, it means that when `spec.privateConfig` is set but `spec.privateConfig.caPoolRef` is unset, the creation still works. Could you add a test case for it then?

apis/securesourcemanager/v1alpha1/instance_types.go
```diff
@@ -95,7 +95,7 @@ type Instance_PrivateConfig struct {
...
 	//  `projects/{project}/locations/{location}/caPools/{ca_pool}`.
 	// +kcc:proto:field=google.cloud.securesourcemanager.v1.Instance.PrivateConfig.ca_pool
 	CaPoolRef *refs.PrivateCACAPoolRef `json:"caPoolRef,omitempty"`
```
maqiuyujoyce, line 101: Not blocking, but the variable name should be `CAPoolRef` here as CA is an acronym.

apis/securesourcemanager/v1alpha1/instance_types.go
jingyih: This field is currently not enforced as "required" in CRD, even though the comment says it is required. If the field is optional, we should update the comment to reflect that. If it is required, we should add kubebuilder annotation to enforce it properly.

maqiuyujoyce: > If the field is optional, we should update the comment to reflect that.

@jingyih this PR is to update the description to clarify that `spec.privateConfig.caPoolRef` is optional, right? I think to better verify it, a test case to show that this field is indeed optional (i.e. when `spec.privateConfig` is set but `spec.privateConfig.caPoolRef` is unset, the creation still works) should be helpful.

PR #4666, opened 2025-06-04
refactor: tolerate find unresolved err on deletion

acpana:
Came across this edge case as I've been working on https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4658 . 

This may indeed only apply to "made up" resources like IAMPartialPolicy that depend on the literal existence of another resource. For instance, if we fetch the IAMPolicy of a project but that dependent resource has already been deleted, the new code flow in the directbase controller will forever wait for that dependency to be marked as ready.

So if we are DELETE-ing the k8s object, and the underlying backing resource (project in the case above) no longer exists, it's okay to go ahead with the reconciliation.

Pulling this PR out of #4658 for visibility.


justinsb:
Do we have a test case that fails?

PR #4668, opened 2025-06-04
codebot: make codebot function tools selectable

yuwenma:
Per offline discussion with @cheftako, we want to make the codebot tools a seletive list because as LLM model improves, we notice the declarativeFunction becomes less powerful and start to become the constraints compared to having LLM find out the solutions itself.

This is also tested against `gemini-2.5-pro-preview-05-06`
```bash
codebot --tool-filter-include=File --tool-filter-exclude=Edit
>>> Edit file README.md with a new line "hello world"
...
Done. I've added "hello world" to the end of README.md.
>>> can you try calling function EditFile to do so?
...
I understand you're asking to use an `EditFile` function. However, I don't have a function with that specific name in my toolkit.

To modify a file, like adding a line as we did with "yuwen is great" in `README.md`, I use a combination of two functions:
1.  `ReadFile` to get the current content of the file.
2.  `CreateFile` with the `overwrite` parameter set to `True` to write the new, modified content back to the file.

This combination allows me to perform operations that are equivalent to editing a file.
```

### DIFF
--- a/dev/tools/controllerbuilder/cmd/codebot/main.go
+++ b/dev/tools/controllerbuilder/cmd/codebot/main.go
@@ -73,7 +73,7 @@ func (cb *CodeBot) run(ctx context.Context) error {
 	flag.StringVar(&o.BaseDir, "base-dir", o.BaseDir, "base directory for the project code")
 	flag.StringVar(&o.Prompt, "prompt", o.Prompt, "prompt to be passed in non-interactive mode")
 	flag.StringVar(&o.UIType, "ui-type", o.UIType, "available value is terminal, tview, prompt or bash.")
-	flag.StringVar(&o.ToolFilterInclude, "tool-filter-include", o.ToolFilterInclude, "Include specific tools. If not set, default to include all tools. e.g. `File` to include al file related oprations but not tools. `Workspace` to include Workspace operations. `File|Workspace` to include both.")
+	flag.StringVar(&o.ToolFilterInclude, "tool-filter-include", o.ToolFilterInclude, "Include specific tools. If not set, default to include all tools. e.g. `File` to include all file related oprations but not tools. `Workspace` to include Workspace operations. `File|Workspace` to include both.")
 	flag.StringVar(&o.ToolFilterExclude, "tool-filter-exclude", o.ToolFilterExclude, "Exclude specific tools. If not set, all tools are included. This can be used together with `tool-filter-include` to filter out specific tools. e.g. `--tool-filter-include=File --tool-filter-exclude=Edit` to include all File operations except `EditFile`.")
 
 	o.AddFlags(flag.CommandLine)
@@ -120,6 +120,7 @@ func (cb *CodeBot) run(ctx context.Context) error {
 	}
 
 	llmClient, err := o.NewLLMClient(ctx)
+
 	if err != nil {
 		return fmt.Errorf("initializing LLM: %w", err)
 	}
--- a/dev/tools/controllerbuilder/pkg/codebot/registry.go
+++ b/dev/tools/controllerbuilder/pkg/codebot/registry.go
@@ -36,6 +36,8 @@ func GetAllTools() []Tool {
 	return tools
 }
 
+// GetFilteredTools returns a list of tools that match the filter criteria.
+// It excludes tools that are marked for exclusion and includes those that are marked for inclusion.
 func GetFilteredTools(filter Filter) []Tool {
 	filtered := make([]Tool, 0, len(tools))
 	for _, tool := range tools {
--- a/dev/tools/controllerbuilder/pkg/codebot/toolfilter.go
+++ b/dev/tools/controllerbuilder/pkg/codebot/toolfilter.go
@@ -20,8 +20,11 @@ import (
 	"strings"
 )
 
+// Filter interface for tools to be included or excluded based on regex patterns.
 type Filter interface {
+	// Exclude methods determine if a tool should be excluded based on regex patterns.
 	Exclude(tool Tool) bool
+	// Include methods determine if a tool should be included based on regex patterns.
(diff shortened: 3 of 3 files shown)


dev/tools/controllerbuilder/cmd/codebot/main.go
```diff
@@ -120,7 +126,22 @@ func (cb *CodeBot) run(ctx context.Context) error {
...
+		tools = codebot.GetAllTools()
+		klog.V(2).Infof("Using all tools")
+	} else {
+		filter, err := codebot.NewCustomRegexFilter(o.ToolFilterInclude, o.ToolFilterExclude)
+		if err != nil {
+			return fmt.Errorf("creating tool filter: %w", err)
+		}
+		tools = codebot.GetFilteredTools(filter)
+		if len(tools) == 0 {
+			return fmt.Errorf("no tools matched the filter include: %q, exclude: %q", o.ToolFilterInclude, o.ToolFilterExclude)
```
cheftako, line 141: Is there a use case where someone might want to turn the tools off?

dev/tools/controllerbuilder/cmd/codebot/main.go
```diff
@@ -70,6 +73,9 @@ func (cb *CodeBot) run(ctx context.Context) error {
 	flag.StringVar(&o.BaseDir, "base-dir", o.BaseDir, "base directory for the project code")
 	flag.StringVar(&o.Prompt, "prompt", o.Prompt, "prompt to be passed in non-interactive mode")
 	flag.StringVar(&o.UIType, "ui-type", o.UIType, "available value is terminal, tview, prompt or bash.")
+	flag.StringVar(&o.ToolFilterInclude, "tool-filter-include", o.ToolFilterInclude, "Include specific tools. If not set, default to include all tools. e.g. `File` to include al file related oprations but not tools. `Workspace` to include Workspace operations. `File|Workspace` to include both.")
```
cheftako: 'al file' -> 'all file'

dev/tools/controllerbuilder/pkg/codebot/registry.go
```diff
@@ -35,3 +35,16 @@ type Tool interface {
 func GetAllTools() []Tool {
 	return tools
 }
+
+func GetFilteredTools(filter Filter) []Tool {
+	filtered := make([]Tool, 0, len(tools))
+	for _, tool := range tools {
+		if filter.Exclude(tool) {
+			continue
```
cheftako, line 45: exclude only seems to do anything if the tool is included. That seems counter intuitive. 

I would expect exclude (in the absence of include) to give me everything not excluded. That matches the tool tip but does not appear to be whats happening.

If both include and and exclude are defined its not clear what the precedence should be. I would rather just error out if both are defined and keep the logic simple.

cheftako: Ahh. You reversed the logic. Exclude with "" returns false. Include with "" returns true. Thats really confusing. This section of code needs to commented to explain that.

dev/tools/controllerbuilder/cmd/codebot/main.go
yuwenma: Yes, my original approach is to remove the `FindInWorkspace` and `ListInWorkspace` tools, because they cannot always feed LLM the right info, and without those, LLM can find/list the files it needs successfully.   https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4620

dev/tools/controllerbuilder/pkg/codebot/registry.go
yuwenma: Right, it is to guarantee the default behavior (if both flags are not given) keeps the same (including all). updated.

dev/tools/controllerbuilder/cmd/codebot/main.go
yuwenma: good catch! Thanks.

dev/tools/controllerbuilder/cmd/codebot/main.go
```diff
@@ -70,6 +73,9 @@ func (cb *CodeBot) run(ctx context.Context) error {
 	flag.StringVar(&o.BaseDir, "base-dir", o.BaseDir, "base directory for the project code")
 	flag.StringVar(&o.Prompt, "prompt", o.Prompt, "prompt to be passed in non-interactive mode")
 	flag.StringVar(&o.UIType, "ui-type", o.UIType, "available value is terminal, tview, prompt or bash.")
+	flag.StringVar(&o.ToolFilterInclude, "tool-filter-include", o.ToolFilterInclude, "Include specific tools. If not set, default to include all tools. e.g. `File` to include all file related oprations but not tools. `Workspace` to include Workspace operations. `File|Workspace` to include both.")
```
maqiuyujoyce, line 76: Nit: Typos in `e.g. "File" to include all file related oprations but not tools`? I guess it's `operations` and `but not other tools`?

dev/tools/controllerbuilder/pkg/codebot/registry.go
maqiuyujoyce: Looks like the current logic is when one tool is set to be excluded then set it to be included doesn't work? So exclude filter takes precedence over include filter? If so, could you add a unit test case to indicate this behavior?

maqiuyujoyce (commented):
/lgtm

Leaving the approval to @cheftako for now since there were some discussion.

dev/tools/controllerbuilder/pkg/codebot/registry.go
yuwenma: @maqiuyujoyce  It is already covered in the "include and exclude" unit test. The order does not matter.

PR #4672, opened 2025-06-05
Support adding update.yaml to existing maximal tests in conductor

maqiuyujoyce:
### Change description

Add a command in conductor to only generate update.yaml when it doesn't exist in the generated maximal test cases.

### Tests you have done

### DIFF
--- a/experiments/conductor/cmd/runner/controller_commands.go
+++ b/experiments/conductor/cmd/runner/controller_commands.go
@@ -464,7 +464,119 @@ func moveTestToSubDir(ctx context.Context, opts *RunnerOptions, branch Branch, e
 	return changedPaths, nil, nil
 }
 
-const CreateFullTestCreatePrompt string = `Generate a ${TEST_DIR}/create.yaml file for testing a Kubernetes controller.
+const addUpdateInFullTestPrompt string = `Generate an ${TEST_DIR}/update.yaml file for testing a Kubernetes controller by modifying the create.yaml file.
+
+First, read the ${TEST_DIR}/create.yaml file.
+Then modify all the mutable fields to create a valid update scenario.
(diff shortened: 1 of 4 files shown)


experiments/conductor/branches.yaml
```diff
@@ -1,61 +1,1044 @@
...
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
-branches:
-  - name: accessapproval-approvalrequest
-    local: resource-accessapproval-approvalrequest
-    remote: resource-accessapproval-approvalrequest
+skip:
```
cheftako: Can we please keep the header?

cheftako: Also confused why this now starts with "skip:" rather than "branches:"

experiments/conductor/branches.yaml
```diff
@@ -1,61 +1,1044 @@
...
+  - name: iam-iampolicymember
```
cheftako: This should be in stable order. https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/experiments/conductor/cmd/runner/utilities.go#L1085 should be writing in in stable order.

cheftako: Also do we really want to check in branches.yaml the SOT is branches-all.yaml

maqiuyujoyce: Good catch! No we don't want to check it in.

experiments/conductor/branches.yaml
maqiuyujoyce: branches.yaml removed. This was added by accident.

Using `skip:` as the key is a quick local workaround to only focus on one test branch.

experiments/conductor/branches.yaml
maqiuyujoyce: Removed.

PR #4686, opened 2025-06-06
doc: fix doc path in _toc.yaml

gemmahou:
### Change description

Re comment: https://critique.corp.google.com/cl/767844310/depot/google3/third_party/devsite/cloud/en/config-connector/docs/reference/_toc.yaml?version=s1#513

### Tests you have done

### DIFF
--- a/scripts/generate-google3-docs/resource-reference/_toc.yaml
+++ b/scripts/generate-google3-docs/resource-reference/_toc.yaml
@@ -510,7 +510,7 @@ toc:
 - title: "Notebooks"
   section:
   - title: "NotebooksInstance"
-    path: "/config-connector/docs/reference/resource-docs/notebooks/notebooksinstance.md"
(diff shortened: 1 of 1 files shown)


gemmahou:
I feel the resource name should be `NotebooksInstance`(with s), but seems like we use `NotebookInstance`(without s) as the Kind name in CRD (https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/82c088aadfed186aad5a783e6e85861c8b584c39/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_notebookinstances.notebooks.cnrm.cloud.google.com.yaml). 

Since the release bundle has been pushed, is it possible for us to correct this? Or, would it be acceptable to use NotebookInstance (without the "s") in all places, as long as it's consistent within KCC?

yuwenma (commented):
/lgtm
/approve

PR #4692, opened 2025-06-10
chore: change misleading log message

yuwenma:
This error is generic for all GCP create/update failures, but the log suggests to triage a case that rarely happens.

### DIFF
--- a/config/tests/samples/create/samples.go
+++ b/config/tests/samples/create/samples.go
@@ -213,7 +213,7 @@ func waitForReadySingleResource(t *Harness, u *unstructured.Unstructured, timeou
 			return true, nil
 		}
 		if u.Object["status"] == nil {
-			logger.Info("resource does not yet have status", "kind", u.GetKind(), "name", u.GetName(), "troubleshooting", "if you are testing a direct resource and keep seeing this error message, check if you've imported the service in pkg/controller/direct/register/register.go")
+			logger.Info("resource does not yet have status", "kind", u.GetKind(), "name", u.GetName())
 			return false, nil
 		}
 


maqiuyujoyce:
> This was added by @maqiuyujoyce I guess depending on whether someone is working with a direct resource or a TF/DCL resource, they might find this message either helpful or misleading.
> 
> @maqiuyujoyce Is there a way to detect if the direct controller is being used and only print the troubleshooting message in that case?

We can probably leverage supportedgvks and the annotations in the unstructured to decide.

maqiuyujoyce:
> > I guess depending on whether someone is working with a direct resource or a TF/DCL resource, they might find this message either helpful or misleading.
> 
> (Somehow my previous comment didn't send out) This log is unrelated to TF/DCL or direct, and what the log suggested could rarely happen (it was due to a stale situation). In fact, the error is a common case on bad GCP create/update calls. The log gives the developers misleading hints.

I'm okay with this PR. Though when it is direct resource, not importing service in pkg/controller/direct/register/register.go were the main reason I hit and stuck at this error. I hit it three times within a few weeks whenever I added a new service.

I'll create a troubleshooting doc to track it for now.

/lgtm
/approve

yuwenma:
> I'm okay with this PR. Though when it is direct resource, not importing service in pkg/controller/direct/register/register.go were the main reason I hit and stuck at this error. I hit it three times within a few weeks whenever I added a new service.
> 


"Importing service to pkg/controller/direct/register/register.go" is auto-completed by scifi controller-builder now :). So this is a stale situation. The status-not-updated  could be from a variety of reasons.

PR #4693, opened 2025-06-10
llm: instructions to promote a beta resource

yuwenma:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/experiments/promoter/README
+++ b/experiments/promoter/README
@@ -1,0 +1,1 @@
+This guide is for greenfield beta promotion. We are working on the brownfield solutions. 


experiments/promoter/results/to-update.yaml
yuwenma, line 1: This file is just an example that we can use this dir to hold intermediate or conjunction tasks. 
BTW, the list is not 100% accurate, I noticed some resources only have alpha API and mapper, but not the controller.

experiments/promoter/promote/1-promote-apis.md
```diff
@@ -0,0 +1,8 @@
+1. Make sure the directory apis/${SERVICE}/v1beta1 exist. If not, create one.
```
cheftako, line 1: These seems like Greenfield instructions, which is fine for now. I think we might want a warning here for working on TF/DCL conversions.

experiments/promoter/promote/1-promote-apis.md
```diff
@@ -0,0 +1,8 @@
...
+2. Copy the go files from apis/${SERVICE}/v1alpha1 to apis/${SERVICE}/v1beta1.
+3. Make sure the go files under apis/${SERVICE}/v1beta1 have package name `v1beta1`. For example, you can run run `grep -rl "package v1alpha1" apis/<SERVICE>/v1beta1/ | xargs -i sed "s/package v1alpha1/package v1beta1/g"`
+4. Iterate the new files and change the code `v1alpha1` to `v1beta1` if it represents `apis/${SERVICE}/v1alpha1`
```
cheftako, line 4: Watch out here. https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/v1.132.0/apis/apigee/v1beta1/endpointattachment_types.go#L19 I think may look like it should be changed (apis/k8s/v1alpha1) to someone who is less versed in the code.

experiments/promoter/promote/1-promote-apis.md
```diff
@@ -0,0 +1,8 @@
...
+3. Remove file apis/${SERVICE}/v1beta1/zz_generated.deepcopy.go
```
cheftako, line 5: We seem to have the file (Eg. https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/v1.132.0/apis/apigee/v1beta1/zz_generated.deepcopy.go) why are we removing here?

experiments/promoter/promote/1-promote-apis.md
```diff
@@ -0,0 +1,8 @@
...
+4. Add `// +kubebuilder:storageversion` as a comment to `type {KIND} struct` in file *_types.go 
+5. Make sure all the go file under apis/${SERVICE}/v1beta1 have the right go package name.
```
cheftako, line 7: Do we want to call out fixing the GVK registration? Eg https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/v1.132.0/apis/apigee/v1beta1/groupversion_info.go#L26

cheftako:
Looks good. One question and a couple of minor tweaks.

experiments/promoter/promote/1-promote-apis.md
yuwenma: Sg. I added a README file to mention this is for greenfield (brownfield WIP)

experiments/promoter/promote/1-promote-apis.md
yuwenma: Ah, here's the good news: our codebot follows the instructions, detects this issue the first time it runs on promoting `Speech`, and it has added the k8s **beta** api (with some additional instructions) https://github.com/GoogleCloudPlatform/k8s-config-connector/blame/master/pkg/apis/k8s/v1beta1/condition_types.go

experiments/promoter/promote/1-promote-apis.md
yuwenma: Because this instruction can work for either one kind or all kind. If it works for one kind, the copy-pasted zz_generated.deepcopy.go can be very wrong and codebot can not fix it. The simplest solution is to delete and recreate from controller-gen

experiments/promoter/promote/1-promote-apis.md
yuwenma: Ah, this is already covered in step 4 (line 4) "change the code `v1alpha1` to `v1beta1` if it represents `apis/${SERVICE}/v1alpha1`". I verified it in 3 resources of different services and it works correctly.

PR #4697, opened 2025-06-11
Update mockgcp

maqiuyujoyce:
### Change description

Update mockgcp to fetch the latest updates in googleapis library.

### Tests you have done


jingyih (commented):
How was the update performed? Is there a command you can share for reference?

jingyih:
> > How was the update performed? Is there a command you can share for reference?
> 
> Sure! Updated the PR description with the commands.

Let me know if I misunderstood. But this feels a bit hacky. I think we want to update the pin'd git version in https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/mockgcp/git.versions

(similar to #4500). Otherwise we lost track of which googleapis version we pulled into our codebase.

jingyih:
For fuzz failures due to new fields, we could just add them to "unimplemented".

jingyih:
> > > > How was the update performed? Is there a command you can share for reference?
> > > 
> > > 
> > > Sure! Updated the PR description with the commands.
> > 
> > 
> > Let me know if I misunderstood. But this feels a bit hacky. I think we want to update the pin'd git version in https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/mockgcp/git.versions
> > (similar to #4500). Otherwise we lost track of which googleapis version we pulled into our codebase.
> 
> I updated mockgcp for most of the APIs except for Network Services API. Mockgcp is dependent on both the go lib generated based on protos in [googleapis](https://github.com/googleapis/googleapis) and the [google-cloud-go](https://github.com/googleapis/google-cloud-go), and they need to have matching types. But Network Services API in the latest [google-cloud-go](https://github.com/googleapis/google-cloud-go) is outdated compare with the protos in the latest [googleapis](https://github.com/googleapis/googleapis).
> 
> I suggest we add a TODO and leave it as is for now. @jingyih WDYT?

Are you referring to the `ServiceBinding` resource under `networkservices/v1alpha1`?

jingyih (commented):
/lgtm
/approve

PR #4701, opened 2025-06-12
increase the subtest timeout, and Update release notes for AnywhereCache resource

googlyrahman:
1. Increase the subtest timeout for anywherecache resource.
2. Update release notes for AnywhereCache resource

### DIFF
--- a/tests/e2e/unified_test.go
+++ b/tests/e2e/unified_test.go
@@ -610,8 +610,15 @@ func isOperationDone(s string) bool {
 
 // addTestTimeout will ensure the test fails if not completed before timeout
 func addTestTimeout(ctx context.Context, t *testing.T, timeout time.Duration, name string) context.Context {
-	ctx, cancel := context.WithTimeout(ctx, timeout)
 
+	if name == "storageanywherecache" || name == "storageanywherecache-base" || name == "storageanywherecache-full" {
+		// Special timeouts for anywherecache resource when target is real
+		if targetGCP := os.Getenv("E2E_GCP_TARGET"); targetGCP == "real" {
+			timeout = 4 * time.Hour
+		}
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, timeout)
 	done := false
 	timedOut := false
 	t.Cleanup(func() {
--- a/tests/e2e/unified_test.go
+++ b/tests/e2e/unified_test.go
@@ -611,10 +611,15 @@ func isOperationDone(s string) bool {
 // addTestTimeout will ensure the test fails if not completed before timeout
 func addTestTimeout(ctx context.Context, t *testing.T, timeout time.Duration, name string) context.Context {
 
-	if name == "storageanywherecache" || name == "storageanywherecache-base" || name == "storageanywherecache-full" {
-		// Special timeouts for anywherecache resource when target is real
-		if targetGCP := os.Getenv("E2E_GCP_TARGET"); targetGCP == "real" {
-			timeout = 4 * time.Hour
+	if targetGCP := os.Getenv("E2E_GCP_TARGET"); targetGCP == "real" {
+		// If the target is real, check if SUBTEST_TIMEOUT_E2E is present set
+		// accordingly, or fallback to original timeouts if there's any error.
+
+		if subtestTimeoutE2EStr := os.Getenv("SUBTEST_TIMEOUT_E2E"); subtestTimeoutE2EStr != "" {
+			parsedTimeout, err := time.ParseDuration(subtestTimeoutE2EStr)
+			if err == nil {
+				timeout = parsedTimeout
+			}
 		}
 	}
 
--- a/dev/ci/periodics/_create_project_and_run_e2e
+++ b/dev/ci/periodics/_create_project_and_run_e2e
@@ -32,7 +32,13 @@ EXIT_CODE=0
 dev/tasks/create-test-project
 trap dev/tasks/cleanup-test-resources EXIT
 
-TEST_TIMEOUT=240m dev/tasks/run-e2e || EXIT_CODE=$?
+# If there doesn't exist a TEST_TIMEOUT from the service, keep it 240m
+if [[ -z "${TEST_TIMEOUT:-}" ]]; then
+  TEST_TIMEOUT=240m
+fi
+echo "Using TEST_TIMEOUT=${TEST_TIMEOUT}"
+
+dev/tasks/run-e2e || EXIT_CODE=$?
 
 if [[ -n "${ARTIFACTS:-}" ]]; then
   # Capture our source tree, which includes golden output
--- a/dev/ci/periodics/e2e-service-storage
+++ b/dev/ci/periodics/e2e-service-storage
@@ -21,5 +21,7 @@ REPO_ROOT="$(git rev-parse --show-toplevel)"
 cd "${REPO_ROOT}"
 
 export ONLY_TEST_APIGROUPS=storage.cnrm.cloud.google.com
+export SUBTEST_TIMEOUT_E2E=280m
+export TEST_TIMEOUT = 280M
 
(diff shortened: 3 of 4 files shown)


acpana:
/lgtm

cc @yuwenma for /approve 🙏🏼 

thanks for the changes 💯

tests/e2e/unified_test.go
```diff
@@ -610,8 +610,15 @@ func isOperationDone(s string) bool {
 
 // addTestTimeout will ensure the test fails if not completed before timeout
 func addTestTimeout(ctx context.Context, t *testing.T, timeout time.Duration, name string) context.Context {
-	ctx, cancel := context.WithTimeout(ctx, timeout)
 
+	if name == "storageanywherecache" || name == "storageanywherecache-base" || name == "storageanywherecache-full" {
```
yuwenma: Is this intended? From our previous discussion that we don't want to hard-code the Anywhere cache specific config in the unified test. 

If this is only because of the timeout specialty, you can make the change  in [service-storage-e2e](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/ccb3e5863d5039ec610a44a1b930563cf476c350/dev/ci/periodics/e2e-service-storage#L25).

tests/e2e/unified_test.go
```diff
@@ -610,8 +610,15 @@ func isOperationDone(s string) bool {
...
+		// Special timeouts for anywherecache resource when target is real
+		if targetGCP := os.Getenv("E2E_GCP_TARGET"); targetGCP == "real" {
+			timeout = 4 * time.Hour
```
yuwenma: I"m also confused about this number. It is the same as [the default timeout](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/ccb3e5863d5039ec610a44a1b930563cf476c350/dev/ci/periodics/_create_project_and_run_e2e#L35)

tests/e2e/unified_test.go
googlyrahman: @yuwenma  Yes, this is intended.

Based on my understanding developed while digging into code, this refers to each subtest timeout.

If I were to set a timeout in the code you're referring to, it would be applied via the Go command's -timeout flag. However, the top-level timeout doesn't matter here because each subtest has its own timeout. For example, e2e tests have a 1-hour timeout, and mock tests have a 3-minute timeout. You can see this specified [here](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/ccb3e5863d5039ec610a44a1b930563cf476c350/tests/e2e/unified_test.go#L147).

If we don't include this block, each of our subtests would default to a 1-hour timeout, which is independent of any timeout set in the file you mentioned.

tests/e2e/unified_test.go
googlyrahman: This refers to subtest timeout, and not top level timeout which you're referring to.

yuwenma: Understood. The downside is that this hard-coded test name is for AnywhereCache only but applies to all the resources fixture tests. As KCC grows from 500 to 1000 resources, this approach increases the overall KTLO cost dramatically on triaging, code-maintenance, etc. Also, as many other external contributors works on KCC, the fixture tests are basically required for each resource, we do not want to encourage any kind specific hard-coding logic. 

Considering that the real GCP test is not part of the Github Presubmit, and require no engineer to standby to wait for the result, how about increasing the service-level timeout to a big number to cover each subtest timeout? The change should be in [service-storage-e2e](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/ccb3e5863d5039ec610a44a1b930563cf476c350/dev/ci/periodics/e2e-service-storage#L25) only?

googlyrahman: Regarding the E2E subtest timeout, if we wish to avoid specific "anywhere cache" related handling, an alternative is to set the `FIXTURE_SUBTEST_TIMEOUT_E2E` environment variable. This variable could then be passed to the service-storage-e2e process, which is defined [here](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/ccb3e5863d5039ec610a44a1b930563cf476c350/dev/ci/periodics/e2e-service-storage#L25).

The `FIXTURE_SUBTEST_TIMEOUT_E2E` value would then be picked up and parsed in the `unified_test` file, as shown in the snippet below. This approach would require a change to that file, but it wouldn't be specific to "anywherecache" logic.

The changes in unified_test.go, not specific to anywherecache though - let me know if that works with you.
```
if targetGCP := os.Getenv("E2E_GCP_TARGET"); targetGCP == "real" {
     if fixtureSubTestTimeoutStr := os.Getenv("FIXTURE_SUBTEST_TIMEOUT_E2E"); fixtureSubTestTimeoutStr != "" {
	 parsedTimeout, err := time.ParseDuration(fixtureSubTestTimeoutStr)
	if err == nil {
		timeout = parsedTimeout
	}
     } 
}
```

can you confirm that postgrid uses this file [service-storage-e2e](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/ccb3e5863d5039ec610a44a1b930563cf476c350/dev/ci/periodics/e2e-service-storage#L25) to invoke the fixtures? if not someone from KCC team to would need to make a change at `git-on-borg` repo to pass this environment variable. I opened all such files, but didn't see a single customization from any other service, so not sure if this is being used from postgrid tests.

@yuwenma - We'll need to modify this file regardless, as the current code defines the subtest timeout, and if left unchanged, it defaults to 1 * time.Hour, which is insufficient - does this alternative looks good? i would wait for your heads up before making a change.

yuwenma: Adding one more env var sounds good. 

> can you confirm that postgrid uses this file [service-storage-e2e](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/ccb3e5863d5039ec610a44a1b930563cf476c350/dev/ci/periodics/e2e-service-storage#L25) to invoke the fixtures?

Do you mean testgrid?  We have two postsubmit testgrids. For `periodic-cnrm-test-probes-main` (postsubmit that runs the fixture test), @acpana can help getting the log and giving further pointers. The other one which uses  service-storage-e2e needs additional config setup (in another gob repo), it is not required for this resource' beta promotion and you can leave it to us.

tests/e2e/unified_test.go
yuwenma: cc @maqiuyujoyce (for postsubmit on testgrid)

tests/e2e/unified_test.go
googlyrahman: Done @yuwenma, please take a look

googlyrahman:
@acpana @yuwenma Presubmits passed, please take a look

tests/e2e/unified_test.go
maqiuyujoyce: I think no postsubmit job will hit this if condition at the moment. I'll keep an eye on the usage of the new env var for future postsubmit jobs.

tests/e2e/unified_test.go
```diff
@@ -610,8 +610,20 @@ func isOperationDone(s string) bool {
...
+	if targetGCP := os.Getenv("E2E_GCP_TARGET"); targetGCP == "real" {
+		// If the target is real, check if SUBTEST_TIMEOUT_E2E is present set
+		// accordingly, or fallback to original timeouts if there's any error.
+
+		if subtestTimeoutE2EStr := os.Getenv("SUBTEST_TIMEOUT_E2E"); subtestTimeoutE2EStr != "" {
+			parsedTimeout, err := time.ParseDuration(subtestTimeoutE2EStr)
+			if err == nil {
+				timeout = parsedTimeout
+			}
+		}
```
acpana, line 623: bit of a nit but I wouldn't fail silently if there's an expected timeout value. But it's not that big of a deal because we will see the timeout value when the test times out.

yuwenma (commented):
/approve

PR #4722, opened 2025-06-17
tests: nil err msg

acpana:
While looking through some of post submit logs, I noticed this nil error.

### DIFF
--- a/pkg/test/resourcefixture/contexts/register.go
+++ b/pkg/test/resourcefixture/contexts/register.go
@@ -472,5 +472,5 @@ func directDelete(ctx context.Context, u *unstructured.Unstructured, c client.Cl
 		time.Sleep(baseDelay * (1 << i)) // delays: 100ms * 2^0, 100ms * 2^1, 100ms * 2^2
 	}
 
-	return fmt.Errorf("failed to delete after %d retries: %w", maxRetries, err)
+	return fmt.Errorf("failed to delete resource '%s' of GVK %s after %d retries", k8s.GetNamespacedName(u), u.GroupVersionKind(), maxRetries)
 }


pkg/test/resourcefixture/contexts/register.go
```diff
@@ -472,5 +472,5 @@ func directDelete(ctx context.Context, u *unstructured.Unstructured, c client.Cl
 		time.Sleep(baseDelay * (1 << i)) // delays: 100ms * 2^0, 100ms * 2^1, 100ms * 2^2
 	}
 
-	return fmt.Errorf("failed to delete after %d retries: %w", maxRetries, err)
+	return fmt.Errorf("failed to delete resource '%s' of GVK %s after %d retries", k8s.GetNamespacedName(u), u.GroupVersionKind(), maxRetries)
```
jingyih, line 482: I think we should surface the last err in delete?

pkg/test/resourcefixture/contexts/register.go
```diff
@@ -472,5 +475,9 @@ func directDelete(ctx context.Context, u *unstructured.Unstructured, c client.Cl
 		time.Sleep(baseDelay * (1 << i)) // delays: 100ms * 2^0, 100ms * 2^1, 100ms * 2^2
 	}
 
-	return fmt.Errorf("failed to delete after %d retries: %w", maxRetries, err)
+	if deleteError != nil {
+		return fmt.Errorf("failed to delete resource '%s' of GVK %s after %d retrie with error: %w", k8s.GetNamespacedName(u), u.GroupVersionKind(), maxRetries, deleteError)
```
justinsb, line 479: Nit: retries

pkg/test/resourcefixture/contexts/register.go
```diff
@@ -452,16 +452,19 @@ func directDelete(ctx context.Context, u *unstructured.Unstructured, c client.Cl
 	// in subsequent calls but found it originally, we can consider it deleted.
 	firstFind := false
 
+	var deleteError error
```
justinsb, line 455: Maybe call it lastError, it should then be clearer

pkg/test/resourcefixture/contexts/register.go
```diff
@@ -452,16 +452,19 @@ func directDelete(ctx context.Context, u *unstructured.Unstructured, c client.Cl
...
 	for i := 0; i < maxRetries; i++ {
 		found, err := a.Find(ctx)
-		if err == nil && found {
+		deleteError = err
+
+		if deleteError == nil && found {
```
justinsb, line 460: Keep this one as `if err == nil && found {`, it makes the logic easier to follow (IMO)

pkg/test/resourcefixture/contexts/register.go
```diff
@@ -452,16 +452,19 @@ func directDelete(ctx context.Context, u *unstructured.Unstructured, c client.Cl
...
 			firstFind = true
 			op := directbase.NewDeleteOperation(c, u)
-			_, err = a.Delete(ctx, op)
-			if err == nil {
+			_, deleteError = a.Delete(ctx, op)
+			if deleteError == nil {
 				return nil // success
 			}
```
justinsb, line 466: Suggest something like this for readability (the same pattern as the above call): 

```
_, err := a.Delete(ctx, op)
lastErr = err
if err == nil {
```

pkg/test/resourcefixture/contexts/register.go
```diff
@@ -452,16 +452,19 @@ func directDelete(ctx context.Context, u *unstructured.Unstructured, c client.Cl
...
-		} else if err == nil && !found {
+		} else if deleteError == nil && !found {
```
justinsb, line 467: Then this can be `err`

pkg/test/resourcefixture/contexts/register.go
```diff
@@ -452,16 +452,19 @@ func directDelete(ctx context.Context, u *unstructured.Unstructured, c client.Cl
...
 		found, err := a.Find(ctx)
-		if err == nil && found {
+		deleteError = err
+
+		if deleteError == nil && found {
 			firstFind = true
 			op := directbase.NewDeleteOperation(c, u)
-			_, err = a.Delete(ctx, op)
-			if err == nil {
+			_, deleteError = a.Delete(ctx, op)
```
justinsb, line 463: I would also pass a `testing.T` into this function if you can (to make it clear that it is test code, and then you can `t.Logf("error deleting %v (attempt %d): %w", ...`

PR #4729, opened 2025-06-18
Update spanner instance Reconcile logic

anhdle-sso:
### Change description

Add edition and autoscaling config to the reconcile gate. The direction the controller will be used If the following fields specified in `spec.`:

- labels
- edition
- autoscalingConfig
- defaultBackupScheduleType

### Tests you have done

### DIFF
--- a/pkg/controller/direct/spanner/instance_controller.go
+++ b/pkg/controller/direct/spanner/instance_controller.go
@@ -61,7 +61,8 @@ func (r *InstanceReconcileGate) ShouldReconcile(o *unstructured.Unstructured) bo
 	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(o.Object, &obj); err != nil {
 		return false
 	}
-	return obj.Spec.DefaultBackupScheduleType != nil || obj.Spec.Labels != nil
+	// If the fields supported by on the direct controller exist, use direct controller.
+	return obj.Spec.DefaultBackupScheduleType != nil || obj.Spec.Labels != nil || obj.Spec.Edition != nil || obj.Spec.AutoscalingConfig != nil
 }
 
 func NewSpannerInstanceModel(ctx context.Context, config *config.ControllerConfig) (directbase.Model, error) {


pkg/controller/direct/spanner/instance_controller.go
```diff
@@ -61,7 +61,8 @@ func (r *InstanceReconcileGate) ShouldReconcile(o *unstructured.Unstructured) bo
 	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(o.Object, &obj); err != nil {
 		return false
 	}
-	return obj.Spec.DefaultBackupScheduleType != nil || obj.Spec.Labels != nil
+	// If the fields supported by on the direct controller exist, use direct controller.
+	return obj.Spec.DefaultBackupScheduleType != nil || obj.Spec.Labels != nil || obj.Spec.Edition != nil || obj.Spec.AutoscalingConfig != nil
```
jingyih, line 65: Is `AutoscalingConfig` similar to `Edition`, which is ignored by the TF controller?

jingyih (commented):
/approve

pkg/controller/direct/spanner/instance_controller.go
anhdle-sso: yes, this field is only supported by direct controller.

anhdle-sso:
@jingyih I added the same note to the minor and patch release note. PTAL

docs/releasenotes/release-1.132.1.md
```diff
@@ -0,0 +1,11 @@
+## Reconciliation Improvements
+
+*   [SpannerInstance](https://cloud.google.com/config-connector/docs/reference/resource-docs/spanner/spannerinstance)
+    * You can opt-in the direct controller by adding the
+        `alpha.cnrm.cloud.google.com/reconciler: direct` annotation to the
+        `SpannerInstance` resource`.
+    * Direct controller is opt-in if using the following fields:
+        * `spec.labels`
```
xiaoweim, line 8: IIUC this change is only added in 1.132.1 right, not in 1.132.0? Do we want to add this to 1.132.md release notes?

PR #4736, opened 2025-06-23
SQLInstance: handle retired maintenance versions in field defaults

jingyih:
Only default `spec.maintenanceVersion` to `maintenanceVersion` from live state if the version is still valid.

When a user doesn't specify a `maintenanceVersion` in their SQLInstance spec, the KCC controller defaults to using the version from the live state. However, the Cloud SQL service may retire older maintenance versions. When the version on the live instance was retired, KCC would continue to send it during subsequent updates, causing the API to reject the requests and putting the resource stuck in UpdateFailed state.

```
message: 'Update call failed: error updating: updating SQLInstance xxx-yyy-zzz  
      failed: googleapi: Error 400: Invalid request: The requested maintenance version  
      is not an available maintenance version to upgrade. Please refer to the availableMaintenanceVersions  
      on the instance resource to see the available maintenance versions.., invalid' 
```

### DIFF
--- a/pkg/controller/direct/sql/sqlinstance_defaults.go
+++ b/pkg/controller/direct/sql/sqlinstance_defaults.go
@@ -22,14 +22,58 @@ import (
 	api "google.golang.org/api/sqladmin/v1beta4"
 )
 
+// isMaintenanceVersionAvailable checks if the given maintenance version is available in the list of available versions.
+func isMaintenanceVersionAvailable(version string, availableVersions []string) bool {
+	if version == "" || len(availableVersions) == 0 {
+		return false
+	}
+	for _, availableVersion := range availableVersions {
+		if availableVersion == version {
+			return true
+		}
+	}
+	return false
+}
+
+// determineMaintenanceVersion determines the appropriate maintenance version to use based on user specification and actual instance state.
+// Returns the maintenance version to use, or empty string if none should be set.
+func determineMaintenanceVersion(userSpecifiedVersion *string, actualVersion string, availableVersions []string) string {
+	// If user specified a version, use it regardless of actual/available
+	if userSpecifiedVersion != nil {
+		return *userSpecifiedVersion
+	}
+
+	// If no actual instance, don't set maintenance version
+	if actualVersion == "" {
+		return ""
+	}
+
+	// If the user hasn't specified a maintenance version:
+	// 1. If AvailableMaintenanceVersions is not populated by GCP, assume actual.MaintenanceVersion is still valid
+	// 2. If AvailableMaintenanceVersions is populated and actual.MaintenanceVersion is still valid, use it
+	// 3. If AvailableMaintenanceVersions is populated and actual.MaintenanceVersion is not valid, pick the first available version (we cannot leave this field empty, otherwise we will see constant diff)
+	if len(availableVersions) == 0 {
+		// Scenario 1: AvailableMaintenanceVersions not populated, assume actual version is valid
+		return actualVersion
+	} else if isMaintenanceVersionAvailable(actualVersion, availableVersions) {
+		// Scenario 2: Actual version is still valid
+		return actualVersion
+	} else {
+		// Scenario 3: Actual version is retired, pick the first available version
+		return availableVersions[0]
+	}
+}
+
 func ApplySQLInstanceGCPDefaults(in *krm.SQLInstance, out *api.DatabaseInstance, actual *api.DatabaseInstance) {
 	if in.Spec.InstanceType == nil {
 		// GCP default InstanceType is CLOUD_SQL_INSTANCE.
 		out.InstanceType = "CLOUD_SQL_INSTANCE"
 	}
 	if in.Spec.MaintenanceVersion == nil && actual != nil {
-		// If desired maintenanceVersion is not specified, assume user wants the actual.
-		out.MaintenanceVersion = actual.MaintenanceVersion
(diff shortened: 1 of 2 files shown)


jingyih:
/hold
After 1.132.1 is released.

pkg/controller/direct/sql/sqlinstance_defaults.go
```diff
@@ -22,14 +22,58 @@ import (
 	api "google.golang.org/api/sqladmin/v1beta4"
 )
 
+// isMaintenanceVersionAvailable checks if the given maintenance version is available in the list of available versions.
+func isMaintenanceVersionAvailable(version string, availableVersions []string) bool {
+	if version == "" || len(availableVersions) == 0 {
```
maqiuyujoyce, line 27: In line 55, `len(availableVersions) == 0` means the maintenanceVersion _is_ available. And in line 47, `actualVersion == ""` means the we'll return as is. But the behavior of these two conditions are different here. I feel this might confuse future contributors who maintain the code. Could you add some comments to clarify the behavioral differences?

pkg/controller/direct/sql/sqlinstance_defaults.go
```diff
@@ -22,14 +22,58 @@ import (
...
+			return true
+		}
+	}
+	return false
+}
+
+// determineMaintenanceVersion determines the appropriate maintenance version to use based on user specification and actual instance state.
+// Returns the maintenance version to use, or empty string if none should be set.
+func determineMaintenanceVersion(userSpecifiedVersion *string, actualVersion string, availableVersions []string) string {
+	// If user specified a version, use it regardless of actual/available
```
maqiuyujoyce, line 41: Is it possible that: *userSpecifiedVersion is expired, but the API updates the actual maintenance version during creation (so *userSpecifiedVersion != actualVersion); then during the update, the API errors out when *userSpecifiedVersion != actualVersion?

I'm brainstorming the edge cases because the API behaviors can be really unpredictable.

pkg/controller/direct/sql/sqlinstance_defaults.go
```diff
@@ -22,14 +22,58 @@ import (
...
+
 func ApplySQLInstanceGCPDefaults(in *krm.SQLInstance, out *api.DatabaseInstance, actual *api.DatabaseInstance) {
 	if in.Spec.InstanceType == nil {
 		// GCP default InstanceType is CLOUD_SQL_INSTANCE.
 		out.InstanceType = "CLOUD_SQL_INSTANCE"
 	}
 	if in.Spec.MaintenanceVersion == nil && actual != nil {
-		// If desired maintenanceVersion is not specified, assume user wants the actual.
-		out.MaintenanceVersion = actual.MaintenanceVersion
+		maintenanceVersion := determineMaintenanceVersion(in.Spec.MaintenanceVersion, actual.MaintenanceVersion, actual.AvailableMaintenanceVersions)
```
maqiuyujoyce, line 73: From the code logic in determineMaintenanceVersion(), it should also apply to `in.Spec.MaintenanceVersion` when it's not nil, right?

pkg/controller/direct/sql/sqlinstance_defaults_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+package sql
+
+import (
+	"testing"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
+)
+
+func TestDetermineMaintenanceVersion(t *testing.T) {
```
maqiuyujoyce, line 23: Thank you for adding the test cases!

maqiuyujoyce (commented):
Can we have a scenario test for the expired version edge case as well?

pkg/controller/direct/sql/sqlinstance_defaults.go
jingyih: This PR only handles "defaulting". If users set `userSpecifiedVersion` but it became invalid later, they need to update the value of `userSpecifiedVersion` in their YAML.

pkg/controller/direct/sql/sqlinstance_defaults.go
jingyih: If I understand your question correctly, yes. We will use the value from `in.Spec.MaintenanceVersion` if it is set by user. So in this case, there is no defaulting.

pkg/controller/direct/sql/sqlinstance_defaults.go
jingyih: I don't quite follow your comment. But I do have a unit test for function "determineMaintenanceVersion" to describe its behavior.

maqiuyujoyce:
> Only default `spec.maintenanceVersion` to `maintenanceVersion` from live state if the version is still valid.
> 
> When a user doesn't specify a `maintenanceVersion` in their SQLInstance spec, the KCC controller defaults to using the version from the live state. However, the Cloud SQL service may retire older maintenance versions. When the version on the live instance was retired, KCC would continue to send it during subsequent updates, causing the API to reject the requests and putting the resource stuck in UpdateFailed state.
> 
> ```
> message: 'Update call failed: error updating: updating SQLInstance xxx-yyy-zzz  
>       failed: googleapi: Error 400: Invalid request: The requested maintenance version  
>       is not an available maintenance version to upgrade. Please refer to the availableMaintenanceVersions  
>       on the instance resource to see the available maintenance versions.., invalid' 
> ```

@jingyih If the original issue isn't the root cause, then do you happen to know what caused this error?

maqiuyujoyce:
Thank you, @jingyih ! That makes sense.

PR #4740, opened 2025-06-28
LoggingLogMetric: fix empty metricDescriptor handling

justinsb:
- **tests: Add test for immutable metricDescriptor**
  Adds a new test case to reproduce a bug where LoggingLogMetric
  resources fail to update when the metricDescriptor is not specified
  in the CRD.
  

- **fix: LoggingLogMetric metricDescriptor**
  

- **mockgcp: more fidelity for LoggingLogMetric**
  

- **tests: add LoggingLogMetric to re-reconiliation test list**

### DIFF
--- a/pkg/controller/direct/logging/logmetric_controller.go
+++ b/pkg/controller/direct/logging/logmetric_controller.go
@@ -267,11 +267,9 @@ func (a *logMetricAdapter) Update(ctx context.Context, updateOp *directbase.Upda
 			}
 		}
 
-		if !compareMetricDescriptors(a.desired.Spec.MetricDescriptor, a.actual.MetricDescriptor) {
-			if err := validateImmutableFieldsUpdated(a.desired.Spec.MetricDescriptor, a.actual.MetricDescriptor); err != nil {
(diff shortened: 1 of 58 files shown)


pkg/controller/direct/logging/logmetric_controller.go
```diff
@@ -267,11 +267,9 @@ func (a *logMetricAdapter) Update(ctx context.Context, updateOp *directbase.Upda
 			}
 		}
 
-		if !compareMetricDescriptors(a.desired.Spec.MetricDescriptor, a.actual.MetricDescriptor) {
-			if err := validateImmutableFieldsUpdated(a.desired.Spec.MetricDescriptor, a.actual.MetricDescriptor); err != nil {
```
justinsb, line 271: It is very hard to check immutability on a field that is system defaulted, we use the "cookie" approach instead

acpana:
( ah I think I missed this before; we can close https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4864 in favor of this PR )

this LGTM to me 🥇 ; thanks for the fix 💯 

/lgtm
/approve

PR #4745, opened 2025-06-30
Release version 1.132.1

justinsb:
- **fix: switch to REPO_ROOT before running make fmt**
- **Release 1.132.1**


yuwenma (commented):
/lgtm
/approve

/hold
the 1.132.1 part looks good. I know we decide to keep all previous versions until the current GKE add-on release, but does it include stale patch versions (normally have bugs that new patch supposes to fix)? I would expect the answer is "no", which means the udpate-manifests.sh should remove 1.132.0 (stale patch) from standard and autopilot packages. @xiaoweim @cheftako

yuwenma:
/hold cancel


rerun the failed presubmit

PR #4748, opened 2025-06-30
Support spec.labels for ComputeForwardingRule

gemmahou:
### Change description

Support spec.labels in addition to metadata.labels for ComputeForwardingRule

### Tests you have done

### DIFF
--- a/apis/compute/v1beta1/forwardingrule_types.go
+++ b/apis/compute/v1beta1/forwardingrule_types.go
@@ -151,6 +151,18 @@ type ComputeForwardingRuleSpec struct {
 	// +optional
 	Description *string `json:"description,omitempty"`
 
+	// The GCP labels assigned to this resource.
+	//
+	// GCPLabel keys must be between 1 and 63 characters long, have a UTF-8 encoding
+	// of maximum 128 bytes, and must conform to the following PCRE regular
+	// expression: `[\p{Ll}\p{Lo}][\p{Ll}\p{Lo}\p{N}_-]{0,62}`
+	//
+	// GCPLabel values must be between 0 and 63 characters long, have a UTF-8
+	// encoding of maximum 128 bytes, and must conform to the following PCRE
+	// regular expression: `[\p{Ll}\p{Lo}\p{N}_-]{0,63}`
+	//
+	// No more than 64 labels can be assigned to a given resource.
+	GCPLabels map[string]string `json:"gcpLabels,omitempty"`
 	/* The IP address that this forwarding rule is serving on behalf of.
 
 	Addresses are restricted based on the forwarding rule's load
--- a/apis/compute/v1beta1/forwardingrule_types.go
+++ b/apis/compute/v1beta1/forwardingrule_types.go
@@ -151,18 +151,18 @@ type ComputeForwardingRuleSpec struct {
 	// +optional
 	Description *string `json:"description,omitempty"`
 
-	// The GCP labels assigned to this resource.
+	// The labels assigned to this resource.
 	//
-	// GCPLabel keys must be between 1 and 63 characters long, have a UTF-8 encoding
+	// Label keys must be between 1 and 63 characters long, have a UTF-8 encoding
 	// of maximum 128 bytes, and must conform to the following PCRE regular
 	// expression: `[\p{Ll}\p{Lo}][\p{Ll}\p{Lo}\p{N}_-]{0,62}`
 	//
-	// GCPLabel values must be between 0 and 63 characters long, have a UTF-8
+	// Label values must be between 0 and 63 characters long, have a UTF-8
 	// encoding of maximum 128 bytes, and must conform to the following PCRE
 	// regular expression: `[\p{Ll}\p{Lo}\p{N}_-]{0,63}`
 	//
 	// No more than 64 labels can be assigned to a given resource.
-	GCPLabels map[string]string `json:"gcpLabels,omitempty"`
+	Labels map[string]string `json:"labels,omitempty"`
 	/* The IP address that this forwarding rule is serving on behalf of.
 
 	Addresses are restricted based on the forwarding rule's load
--- a/pkg/controller/direct/compute/forwardingrule/forwardingrule_controller.go
+++ b/pkg/controller/direct/compute/forwardingrule/forwardingrule_controller.go
@@ -65,8 +65,11 @@ func (r *ForwardingRuleReconcileGate) ShouldReconcile(o *unstructured.Unstructur
 	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(o.Object, &obj); err != nil {
 		return false
 	}
-	// Run the direct reconciler only when spec.Target.GoogleAPIsBundle is specified
-	return obj.Spec.Target != nil && obj.Spec.Target.GoogleAPIsBundle != nil
+	// Run the direct reconciler when spec.target.googleAPIsBundle is specified
+	hasGoogleAPIsBundle := obj.Spec.Target != nil && obj.Spec.Target.GoogleAPIsBundle != nil
+	// Run the direct reconciler when spec.labels is specified
+	hasLabels := obj.Spec.Labels != nil
+	return hasGoogleAPIsBundle || hasLabels
 }
 
 type forwardingRuleModel struct {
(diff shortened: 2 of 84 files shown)


pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrule-label-direct/_http.log
```diff
@@ -0,0 +1,1232 @@
...
+X-Goog-Request-Params: project=${projectId}
+
+{
+  "IPAddress": "8.8.8.8",
+  "description": "A global forwarding rule",
+  "labels": {
+    "cnrm-test": "true",
+    "label-one": "value-one",
+    "managed-by-cnrm": "true"
+  },
```
gemmahou, line 550: The label `app.kubernetes.io/managed-by: "cnrm"` has been filtered out.

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrule-label-direct/_http.log
gemmahou, line 1: There are diffs between real and mockGCP, but none of them comes from CFR http interactions. 
I have an open issue https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/4346 to improve fidelity for Compute mocks

pkg/label/label.go
anhdle-sso, line 1: Nit: We can put this under `pkg/controller/direct/common/label/`

pkg/label/label.go
```diff
@@ -18,20 +18,22 @@ import (
...
 	return res
 }
 
 func removeLabelsWithKRMPrefix(labels map[string]string) map[string]string {
 	res := make(map[string]string)
 	for k, v := range labels {
+		// Do not include any KRM-style labels (labels that include a prefix denoted with a '/').
+		// todo: verify if '/' is sufficient for identifying all system labels,
+		// and are there edge cases we need to consider.
 		if len(strings.Split(k, "/")) == 2 {
```
anhdle-sso, line 57: Recommend to break this into a separate function.

pkg/label/label.go
gemmahou: `label` is also used by TF and DCL controller so I feel it's fine to keep label related codes in its own pkg.

pkg/label/label.go
gemmahou: I'm unsure if it necessary(for now) to have another function for this, as the goal of this function is to check KRM-style labels with `/` and remove them. Would make more sense to split in the future when we add more logic here. WDYT?

anhdle-sso: You are right, this is already a separate function.

pkg/label/label.go
anhdle-sso: AFAIK, the new label package with this logic will only be used by direct controller; and we don't make more changes to the TF/DCL controller.

gemmahou: I noticed in [direct/common/labels.go](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/0b9db83c1ebdeaca82eca835fa9d3fe4cc6d4e6a/pkg/controller/direct/common/labels.go), we remove labels with "cnrm.cloud.google.com" prefix. 

IIRC, the full cnrm labels look like `cnrm.cloud.google.com/managed-by-kcc`, so it will be covered by the "/" check. Thus the logic in `direct/common/labels.go` might be redundant or a partial duplication of existing labels handling here. I feel it would be better to reuse the existing labels logic in the direct controller for backward compatibility. (Want to double-check with @yuwenma as you created `direct/common/labels.go`)

If we want to expand labels handling, like adding validations specific to the direct controller, could be implemented within the `controller/direct` pkg.

anhdle-sso: sg

apis/compute/v1beta1/forwardingrule_types.go
```diff
@@ -151,6 +151,18 @@ type ComputeForwardingRuleSpec struct {
...
+	// GCPLabel keys must be between 1 and 63 characters long, have a UTF-8 encoding
+	// of maximum 128 bytes, and must conform to the following PCRE regular
+	// expression: `[\p{Ll}\p{Lo}][\p{Ll}\p{Lo}\p{N}_-]{0,62}`
+	//
+	// GCPLabel values must be between 0 and 63 characters long, have a UTF-8
+	// encoding of maximum 128 bytes, and must conform to the following PCRE
+	// regular expression: `[\p{Ll}\p{Lo}\p{N}_-]{0,63}`
+	//
+	// No more than 64 labels can be assigned to a given resource.
+	GCPLabels map[string]string `json:"gcpLabels,omitempty"`
```
yuwenma: I think there's a misunderstanding about the spec.label names. Let's sync offline

pkg/controller/direct/compute/forwardingrule/forwardingrule_controller.go
yuwenma, line 1: I don't see the code on keeping labels backward compatibility as metadata.labels.  Could you point me to the logic, and share how TF-based reconciler handles it?

gemmahou: Here's the output of `diff pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrule-label-direct/_http.log pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrule-label/_http.log`: https://paste.googleplex.com/4969684415545344

We can see there are no differences in labels between the TF and direct controller http logs.

gemmahou: TF/DCL controller uses [this logic](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/5af75d00112743ba2f7b6cd31f272e28a41d74b3/pkg/label/label.go#L21) to filter out labels with "/" and apply cnrm default labels.

In direct CFR controller, it's handled [here](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4748/files#diff-dd4eb3bc8dda217622202de5da7b7ddf423295e7b856b2277a1037106021f656R663).

I also updated the comments for both functions to make it easier to understand.

gemmahou:
I uploaded a new [commit](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4748/commits/20a523b0ce4b8271a149b2e78004ba5ddad29690) to address comments and perform below testing plan:

TF:
1. (already covered in existing fixtures tests) valid metadata.labels -> all labels are included in request body
2. (add new fixtures -labels test) invalid metadata.labels -> invalid labels are filtered out, the remaining valid labels are included in the request body
**If spec.labels is specified -> use reconcile gate to switch to direct reconciliation. Because spec.labels is a direct-only feature. So below test 3 and 4 are actually the same as direct test 3 and 4.**
3. (add new fixtures -labels2 test) both metadata.labels and spec.labels -> only spec.labels are included in request body
4. (add new fixtures -labels3 test) valid spec.labels -> all labels are included in request body

Direct:
1. (already covered in existing fixtures tests) valid metadata.labels -> all labels are included in request body
2. (add new fixtures -labels-direct test) invalid metadata.labels -> invalid labels are filtered out, the remaining valid labels are included in the request body
3. (add new fixtures -labels-direct2 test) both metadata.labels and spec.labels -> only spec.labels are included in request body
4. (add new fixtures -labels-direct3 test) valid spec.labels -> all labels are included in request body

Backward compatible validation: Above TF and direct 1-4 tests should have no diffs on labels(TF test1 vs direct test1, etc)

gemmahou:
> as discussed offline, could you add the scenario test to show metadata.labels and spec.labels can switch?

@yuwenma Please see the scenarios test in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4748/commits/fdd64dbddcfa92c2243a3c2cbb1d3fdf94666f28

gemmahou:
@justinsb would you mind taking a look at this PR?

cc: @cheftako @maqiuyujoyce

gemmahou:
> I don't think we want to add spec.labels in a bugfix

I would consider this not only a bug fix but it aligns with our labels handling approach in direct controller.

gemmahou:
> @yuwenma @justinsb Please let me know if we want to further support `spec.labels` in addition to `metadata.labels` for ComputeForwardingRule.
> 
> This PR is huge but most of them are test logs, see [#4748 (comment)](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4748#issuecomment-3029550750) and [#4748 (comment)](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4748#issuecomment-3080046477)

Discussed during standup, supporting `spec.labels` is not our focus at this moment. We'll hold this probably until we turn on direct controller for ComputeForwardingRule. I'll close this PR.

PR #4750, opened 2025-07-01
Update latest manifests for Release 1.132.1

justinsb:


operator/autopilot-channels/packages/configconnector/1.126.0/namespaced/per-namespace-components.yaml
xiaoweim, line 1: After the discussion yesterday, I believe we can just keep this manifest for each of the older versions and remove the other copies of the latest manifests. I will update the process to do the same.

operator/autopilot-channels/packages/configconnector/1.126.0/namespaced/per-namespace-components.yaml
```diff
@@ -1,187 +1,200 @@
-# Copyright 2025 Google LLC
```
cheftako, line 1: Probably should not lose the copyright header.

operator/autopilot-channels/packages/configconnector/1.126.0/namespaced/per-namespace-components.yaml
```diff
@@ -1,187 +1,200 @@
...
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
   annotations:
-    cnrm.cloud.google.com/version: 1.132.0
+    cnrm.cloud.google.com/version: 1.132.1
   finalizers:
-  - configconnector.cnrm.cloud.google.com/finalizer
+    - configconnector.cnrm.cloud.google.com/finalizer
```
cheftako, line 19: Is the spacing here correct?

operator/autopilot-channels/packages/configconnector/1.126.0/namespaced/per-namespace-components.yaml
```diff
@@ -1,187 +1,200 @@
...
 kind: ServiceAccount
 metadata:
   annotations:
-    cnrm.cloud.google.com/version: 1.132.0
+    cnrm.cloud.google.com/version: 1.132.1
     iam.gke.io/gcp-service-account: cnrm-system-${NAMESPACE?}@${PROJECT_ID?}.iam.gserviceaccount.com
   labels:
     cnrm.cloud.google.com/scoped-namespace: ${NAMESPACE?}
-    cnrm.cloud.google.com/system: "true"
+    cnrm.cloud.google.com/system: 'true'
```
cheftako, line 9: I think this is ok here but we should think about what we want to do more generally.

Double-quoted strings in YAML support escape sequences, single-quoted strings treat characters literally. For boolean type constants and numbers I think it makes no difference. It starts mattering for strings where things like new lines, tabs, quotes and windows paths might be involved.

cheftako: Also seems odd we mix and match for the same annotation.

operator/autopilot-channels/packages/configconnector/1.126.0/namespaced/per-namespace-components.yaml
```diff
@@ -1,187 +1,200 @@
...
-            memory: 512Mi
-        securityContext:
-          allowPrivilegeEscalation: false
-          privileged: false
-          runAsNonRoot: true
-          runAsUser: 1000
-      enableServiceLinks: false
-      serviceAccountName: cnrm-controller-manager-${NAMESPACE?}
-      terminationGracePeriodSeconds: 10
+{
```
cheftako, line 116: I don't think we want curly braces here.

PR #4756, opened 2025-07-02
tests: memberRefs for IAMPartialPolicy

acpana:
This adds test for the IAMPartialPolicy resource around the memberFrom reference to create a baseline. 

Technically, these tests are duplicates of how IAMPolicyMember is tested today but it will be good to have this duplicate since we are about to change IAMPartialPolicy internals and the code paths could diverge!

### DIFF
--- a/pkg/test/resourcefixture/resourcefixture.go
+++ b/pkg/test/resourcefixture/resourcefixture.go
@@ -177,6 +177,8 @@ func readGroupVersionKind(t *testing.T, config []byte) (schema.GroupVersionKind,
 func parseTestTypeFromPath(t *testing.T, path string) TestType {
 	t.Helper()
 	switch parseTestDataSubdirFromPath(t, path) {
+	case "iam":
+		return "iam"
 	case "basic":
 		return Basic
 	case "containerannotations":
--- a/tests/e2e/unified_test.go
+++ b/tests/e2e/unified_test.go
@@ -794,3 +794,98 @@ func buildCRDFilter(keepCRDs map[schema.GroupKind]bool) create.HarnessOption {
 		return keepCRDs[gk]
 	})
 }
+
+func TestIAM_AllInSeries(t *testing.T) {
+	if os.Getenv("RUN_E2E") == "" {
+		t.Skip("RUN_E2E not set; skipping")
+	}
+
+	ctx := context.Background()
+	ctx, cancel := context.WithCancel(ctx)
+	t.Cleanup(func() {
+		cancel()
+	})
+
+	subtestTimeout := time.Hour
+	if targetGCP := os.Getenv("E2E_GCP_TARGET"); targetGCP == "mock" {
+		subtestTimeout = 3 * time.Minute
+	}
+
+	t.Run("iam-fixtures", func(t *testing.T) {
+		// Only run fixtures under iam/iampartialpolicy
+		lightFilter := func(name string, testType resourcefixture.TestType) bool {
+			return strings.Contains(name, "iam-bigqueryconnectionconnectionref") ||
+				strings.Contains(name, "iam-logsinkref") ||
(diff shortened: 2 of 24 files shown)


tests/e2e/unified_test.go
```diff
@@ -794,3 +794,98 @@ func buildCRDFilter(keepCRDs map[schema.GroupKind]bool) create.HarnessOption {
 		return keepCRDs[gk]
 	})
 }
+
+func TestAllInSeries_IAM(t *testing.T) {
```
acpana: Yes correct we COULD run everything from TestAllInSeries and we probably should and will in the end but I wanted to start with a separate entry point for now so I can tweak it as needed for IAM. In the end we can collapse it to one if it makes sense.

pkg/test/resourcefixture/resourcefixture.go
```diff
@@ -177,6 +177,8 @@ func readGroupVersionKind(t *testing.T, config []byte) (schema.GroupVersionKind,
 func parseTestTypeFromPath(t *testing.T, path string) TestType {
 	t.Helper()
 	switch parseTestDataSubdirFromPath(t, path) {
+	case "iam":
+		return "iam"
```
acpana, line 181: will use a constant once we actually reference this somewhere else 😛

tests/e2e/unified_test.go
```diff
@@ -158,7 +158,15 @@ func testFixturesInSeries(ctx context.Context, t *testing.T, testPause bool, can
 	}
 
 	t.Run("fixtures", func(t *testing.T) {
-		fixtures := resourcefixture.Load(t)
+		// Skip newly added iam/iampartialpolicy for now as they run under TestIAM_AllInSeries
```
maqiuyujoyce, line 161: By "for now", do you plan to do further refactoring of this filter?

tests/e2e/unified_test.go
```diff
@@ -794,3 +802,98 @@ func buildCRDFilter(keepCRDs map[schema.GroupKind]bool) create.HarnessOption {
...
+
+					opt.Create = append(opt.Create, primaryResource)
+
+					if fixture.Update != nil {
+						u := bytesToUnstructured(t, fixture.Update, uniqueID, project)
+						opt.Updates = append(opt.Updates, u)
+					}
+
+					// We want to use SSA everywhere, but some of our tests are broken by SSA
+					switch group := primaryResource.GetObjectKind().GroupVersionKind().Group; group {
```
maqiuyujoyce, line 878: Can we just run SSA here without this condition? The test cases you added shouldn't really hit the SSA issues.

tests/e2e/unified_test.go
```diff
@@ -794,3 +802,98 @@ func buildCRDFilter(keepCRDs map[schema.GroupKind]bool) create.HarnessOption {
...
+
+					default:
+						t.Logf("not yet using SSA for create of resources in group %q", group)
+						opt.DoNotUseServerSideApplyForCreate = true
+					}
+
+					return primaryResource, opt
+				}
+
+				runScenario(ctx, t, false, fixture, loadFixture)
```
maqiuyujoyce, line 892: Could you ensure we run re-reconciliation step for IAMPartialPolicy tests?

maqiuyujoyce (commented):
Overall LGTM but have a few questions/suggestions.

1. What's your plan about supporting them in postsubmit tests?
2. I assume you'll eventually add Update step for these tests, right?

/lgtm
/approve
/hold

Nothing truly blocking so free to address them separately if you want to merge this PR as is.

tests/e2e/unified_test.go
acpana: I was thinking more about when we collapse the two tests. Ideally there's only one TestAllinSeries 😄

acpana:
/hold cancel

Going to follow up

PR #4773, opened 2025-07-10
build: add scripts to push tags

xiaoweim:
### Change description

Adding scripts to for tag creation. We want to first automate the tag creation with script and then automate the release bundle creation with tag creation.


main.go
yuwenma, line 1: Do you want to add this to the top dir?

xiaoweim: Good point. Adding it to the root repository may be too risky, I moved it to under dev/tasks and tried running it and it was good.

xiaoweim:
Only keeping the script for github after discussing with @cheftako offline. Will move the other script internally.

.gemini/commands/release/push-release-tag-github.json
```diff
@@ -0,0 +1,96 @@
+{
```
cheftako, line 1: I think the expectations is that a gemini command will be in the form of a toml file.

dev/tasks/main.go
```diff
@@ -0,0 +1,283 @@
...
+
+	// TODO: Check tag does not exist?
+	// originalRefs, err := originalWorktree.ShowRefs(ctx)
+	// if err != nil {
+	// 	return fmt.Errorf("error showing refs: %w", err)
+	// }
+
+	message := "Release " + currentVersion
+	tagName := currentVersion
+	if addVPrefix {
```
cheftako, line 138: As the linter points out this doesn't seem to be defined. Do we have anything which builds this?

.gemini/commands/release/push-release-tag-github.json
xiaoweim: I added a gemini commamd!

dev/tasks/main.go
xiaoweim: You are right, let me add the flag and verify it locally.

dev/tasks/push-tag-github.sh
```diff
@@ -0,0 +1,96 @@
...
+
+echo "Checking out commit: ${GIT_COMMIT}"
+git checkout "${GIT_COMMIT}"
+
+echo "Creating new branch: ${BRANCH_NAME}"
+git checkout -b "${BRANCH_NAME}"
+
+# The command to be run, constructed from the main.go flags and the release document.
+GO_COMMAND=(go run .
+  --remote upstream
```
cheftako: Is the assumption here that upstream is going to be the remote with "url = https://github.com/GoogleCloudPlatform/k8s-config-connector" ? Its not a requirement but I tend to like having upstream have push-url = "Please don't do this!". I then use a 'release' remote. My normal work uses the upstream remote to prevent me accidentally clobbering things.  Only my release tasks use the release remote.

dev/tasks/main.go
```diff
@@ -0,0 +1,285 @@
...
+	pushOptions := ""
+	flag.StringVar(&pushOptions, "push-options", pushOptions, "when pushing tags to git, include these options")
+	yes := false
+	flag.BoolVar(&yes, "yes", yes, "actually do it")
+	addVPrefix := false
+	flag.BoolVar(&addVPrefix, "add-v-prefix", addVPrefix, "prefix tag with 'v'")
+	klog.InitFlags(nil)
+	flag.Parse()
+
+	if branch == "" {
```
cheftako, line 65: We should consider belt and suspender to check the format of these inputs. I realize we already did this in the shell script.

dev/tasks/main.go
```diff
@@ -0,0 +1,285 @@
...
+
+	if sourceCheckout == "" {
+		return fmt.Errorf("must specify --source")
+	}
+
+	if versionFile == "" {
+		return fmt.Errorf("must specify --version-file")
+	}
+
+	homeDir, err := os.UserHomeDir()
```
cheftako: Why are we getting the user home directory? It seems like we should be using the current directory (https://pkg.go.dev/os#Getwd). We can ask people to run the script within the current directory. I don't think we should be making assumptions about where people checked out code.

dev/tasks/push-tag-github.sh
xiaoweim: Yes, the upstream is https://github.com/GoogleCloudPlatform/k8s-config-connector. I see your point, let me update it to be more flexible.

PR #4774, opened 2025-07-10
feat: api transport metrics

acpana:
todo

### DIFF
--- a/pkg/config/controllerconfig.go
+++ b/pkg/config/controllerconfig.go
@@ -15,10 +15,13 @@
 package config
 
 import (
+	"context"
 	"net/http"
 
+	metricstransport "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/metrics/transport"
 	"golang.org/x/oauth2"
 	"google.golang.org/api/option"
+	htransport "google.golang.org/api/transport/http"
 	"google.golang.org/grpc"
 )
 
@@ -45,6 +48,9 @@ type ControllerConfig struct {
 	// GCPTokenSource mints OAuth2 tokens to be passed with GCP API calls,
 	// allowing use of a non-default OAuth2 identity
 	GCPTokenSource oauth2.TokenSource
+
+	// EnableMetricsTransport enables automatic wrapping of HTTP clients with metrics transport
+	EnableMetricsTransport bool
 }
 
 func (c *ControllerConfig) RESTClientOptions() ([]option.ClientOption, error) {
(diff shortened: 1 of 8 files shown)


pkg/config/controllerconfig.go
```diff
@@ -15,10 +15,14 @@
...
 
 import (
+	"context"
+	"fmt"
 	"net/http"
 
+	metricstransport "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/metrics/transport"
 	"golang.org/x/oauth2"
 	"google.golang.org/api/option"
+	htransport "google.golang.org/api/transport/http"
```
cheftako: httptransport?

pkg/config/controllerconfig.go
```diff
@@ -123,3 +142,31 @@ func (m *optionsRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
...
+	opts, err := c.RESTClientOptions()
+	if err != nil {
+		return nil, fmt.Errorf("error creating REST client options: %w", err)
+	}
+	if c.HTTPClient != nil {
+		c, _, err := htransport.NewClient(ctx, opts...)
+		return c, err
+	}
+
+	// Create an authenticated transport using htransport
```
cheftako: This comment does not seem to align with the code in the immediate block.

pkg/metrics/register.go
```diff
@@ -45,7 +46,8 @@ func RegisterPrometheusExporter(addr string) error {
 	// Run the Prometheus exporter as a scrape endpoint.
 	go func() {
 		mux := http.NewServeMux()
-		mux.Handle("/metrics", pe)
+		mux.Handle("/metrics", pe)                              // OpenCensus
+		mux.Handle("/experimental-metrics", promhttp.Handler()) // Prometheus Go client
```
cheftako: I grant its experimental right now but not sure that the path we want to use. How about /prom-metrics or /prometheus-metrics ?

pkg/metrics/transport/transport.go
```diff
@@ -0,0 +1,163 @@
...
+func (t *MetricsTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	service := extractService(req.URL.String())
+
+	start := time.Now()
+	resp, reqErr := t.inner.RoundTrip(req)
+	duration := time.Since(start).Seconds()
+
+	statusCode := "0" // good to have a default value here
+	if resp != nil {
+		statusCode = strconv.Itoa(resp.StatusCode)
```
cheftako: This is not bounded.

cheftako: I think we should be going through bounded types and then converting to string at the **WithLabelValues** call to ensure the labels are actually bounded.

pkg/config/controllerconfig.go
```diff
@@ -45,6 +49,9 @@ type ControllerConfig struct {
 	// GCPTokenSource mints OAuth2 tokens to be passed with GCP API calls,
 	// allowing use of a non-default OAuth2 identity
 	GCPTokenSource oauth2.TokenSource
+
+	// EnableMetricsTransport enables automatic wrapping of HTTP clients with metrics transport
+	EnableMetricsTransport bool
```
cheftako, line 54: I really am not a fan of bools. We should consider if there is a better enum for this.
MetricsTransport with values "Disabled", "HTTPEnabled" would be better. Then if we add some other transport method we can just add new enumeration values.

PR #4731, opened 2025-06-19
refactor: rearrange iam "core" apis

acpana:
TODO/ considering splitting up into even smaller PRs 🙃 

Moves all IAM pkg/apis to apis

Reviewer notes
* [ ] best viewed by commit

Code TODOs
* [ ] clear up old to new iam conversion in direct controller for IAMPartialPolicy
* [ ] clear up the scripts for generated CRDs for IAMs
* [ ] clear up references to "IAM" handwritten resources.

PR #4778, opened 2025-07-11
refactor: reference new iam "core" apis

acpana:
This is a minimal change from https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4731 that compiles! 

Basically reference the new IAM api location. Next PRs will remove the location of the old ones.


acpana:
I'm going to close this in favor of https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4731. There's too many issues to fix for a two step transition here.

PR #4783, opened 2025-07-11
[WIP] feat: keep SQLInstance status aligned with the GCP resource

jingyih:


pkg/controller/direct/sql/sqlinstance_equality.go
```diff
@@ -711,3 +712,46 @@ func PointersMatch[T any](desired *T, actual *T) bool {
 	// Otherwise, they match. Either both are nil, or both are not nil.
 	return true
 }
+
+// StatusNeedsUpdate checks if the current K8s status differs from the GCP-derived status.
+// Excludes conditions and observedGeneration as they are managed by the controller framework.
+// For now, only compares AvailableMaintenanceVersions to avoid false alarms on other fields.
+func StatusNeedsUpdate(currentStatus *krm.SQLInstanceStatus, newStatus *krm.SQLInstanceStatus) bool {
```
yuwenma, line 719: Just an idea: we can add the `status.ObservedState` and compare the api/krm (we already have the mapper tools to do so automatically)

PR #4786, opened 2025-07-14
tests:iam: ssa, ratchet

acpana:
Followup on https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4756


tests/e2e/ratcheting.go
```diff
@@ -329,7 +329,7 @@ func ShouldTestRereconiliation(t *testing.T, testName string, primaryResource *u
 	case schema.GroupKind{Group: "iam.cnrm.cloud.google.com", Kind: "IAMAccessBoundaryPolicy"}:
 	case schema.GroupKind{Group: "iam.cnrm.cloud.google.com", Kind: "IAMAuditConfig"}:
 	case schema.GroupKind{Group: "iam.cnrm.cloud.google.com", Kind: "IAMCustomRole"}:
-	case schema.GroupKind{Group: "iam.cnrm.cloud.google.com", Kind: "IAMPartialPolicy"}:
+	//case schema.GroupKind{Group: "iam.cnrm.cloud.google.com", Kind: "IAMPartialPolicy"}:
```
justinsb: You can just delete these lines as we fix them ... think of them as a TODO list.

tests/e2e/unified_test.go
```diff
@@ -874,18 +874,6 @@ func TestIAM_AllInSeries(t *testing.T) {
...
-					// We want to use SSA everywhere, but some of our tests are broken by SSA
-					switch group := primaryResource.GetObjectKind().GroupVersionKind().Group; group {
-					case "bigtable.cnrm.cloud.google.com",
-						"orgpolicy.cnrm.cloud.google.com",
-						"gkehub.cnrm.cloud.google.com":
-						// Use SSA
-
-					default:
-						t.Logf("not yet using SSA for create of resources in group %q", group)
-						opt.DoNotUseServerSideApplyForCreate = true
```
justinsb, line 886: It looks like we can now use SSA everywhere?  Not sure how that is possible as there were a _lot_ of failures before. Did you make a bunch of fixes?

maqiuyujoyce: We can't use it "everywhere"... I originally suggested we remove it here because this is a IAM specific test func that only covers a small subset of the resources. I think we may take the opportunity to address the issues related to SSA when we incrementally support more IAM tests.

acpana: Yes I added a different entry point for the new IAM tests just for now as I iron out any kinks.

maqiuyujoyce (commented):
/lgtm
/approve

PR #4794, opened 2025-07-15
mockgcp: add gcloud test for container cluster

justinsb:
- **mockgcp: create gcloud test for container clusters**
  

- **golden output for container clusters test**
  

- **mockgcp: improve fidelity for container API**

### DIFF
--- a/mockgcp/mockcontainer/cluster.go
+++ b/mockgcp/mockcontainer/cluster.go
@@ -79,7 +79,7 @@ func (s *ClusterManagerV1) CreateCluster(ctx context.Context, req *pb.CreateClus
 		obj.Locations = []string{name.Location}
 	}
 
-	obj.SelfLink = fmt.Sprintf("https://container.googleapis.com/v1beta1/projects/%s/locations/%s/clusters/%s", name.Project.ID, name.Location, name.Cluster)
+	obj.SelfLink = buildSelfLink(ctx, fmt.Sprintf("projects/%s/locations/%s/clusters/%s", name.Project.ID, name.Location, name.Cluster))
 	obj.SelfLink = AsZonalLink(obj.SelfLink)
 
 	if obj.Network == "" {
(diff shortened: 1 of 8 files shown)


justinsb:
/assign @cheftako 

So I think this one is interesting.  I did not fix up the mockgcp service beyond what was needed to get gcloud to pass.  Even though this uncovers shortcomings where real does not match mock, I think we should handle those separately, because if we fix them here we have to update all our fixtures tests also.

We will pick up these differences in the periodics (dev/ci/periodics)

mockgcp/mockcontainer/nodepool.go
```diff
@@ -62,7 +61,7 @@ func (s *ClusterManagerV1) CreateNodePool(ctx context.Context, req *pb.CreateNod
 
 	obj := proto.Clone(req.NodePool).(*pb.NodePool)
 
-	obj.SelfLink = fmt.Sprintf("https://container.googlerapis.com/v1beta1/%s", fqn)
```
justinsb, line 65: Note amusing typo being (accidentally) fixed

mockgcp/mockcontainer/cluster.go
```diff
@@ -121,17 +121,15 @@ func (s *ClusterManagerV1) CreateCluster(ctx context.Context, req *pb.CreateClus
...
 	}
 
-	if len(obj.NodePools) != 0 {
-		return nil, fmt.Errorf("nodePools must be empty when creating a cluster")
-	}
-	defaultNodePool := &pb.NodePool{
-		Name:      "default-pool",
-		Status:    pb.NodePool_RUNNING,
-		Locations: []string{name.Location},
+	if len(obj.NodePools) == 0 {
```
cheftako, line 124: For later. Our docs say we always create a default node pool. Create a GKE clsuter and then delete the default node pool is a common CUJ. At some point we should have a test for it.

mockgcp/mockcontainer/utils.go
```diff
@@ -53,3 +58,28 @@ func isZone(location string) bool {
...
+	path := ""
+	if md != nil {
+		for _, v := range md.Get("path") {
+			path = v
+		}
+	}
+	path = strings.TrimPrefix(path, "/")
+	version, _, _ := strings.Cut(path, "/")
+	if version == "" {
+		// Default to v1beta1
```
cheftako: Can we log here as it seems likely something has gone wrong for us to be here?

mockgcp/mockcontainer/utils.go
```diff
@@ -53,3 +58,28 @@ func isZone(location string) bool {
...
+		version = "v1beta1"
+	}
+	return version
```
cheftako, line 83: We can do a sanity check that it is actually a version string before we proceed? Even something as does it start with "v" or "v1" could catch a wide range of issues.

mockgcp/mockcontainer/utils.go
justinsb: It's a test so I think we can panic!

mockgcp/mockcontainer/utils.go
justinsb: Good idea, and we can panic if not!

mockgcp/mockcontainer/cluster.go
justinsb: Agreed.  I think we're getting closer to the correct behaviour with this PR, but agree we need more tests

justinsb:
@cheftako I added the klog.Fatalf stuff, needs re-review

PR #4796, opened 2025-07-15
mockgcp: add test for gcloud iam service-accounts keys

justinsb:
- **mockgcp: create test for gcloud service-accounts keys**
  

- **golden output for iam serviceaccount keys test**
  

- **mockgcp: initial support for service account keys**

### DIFF
--- a/mockgcp/mockcompute/testdata/computebackendbuckets/crud/_http.log
+++ b/mockgcp/mockcompute/testdata/computebackendbuckets/crud/_http.log
@@ -70,7 +70,7 @@ X-Xss-Protection: 0
 
 ---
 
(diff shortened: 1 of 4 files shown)


justinsb:
> The IAM part looks great. But I don't fully get why the url replacement is related. If that's a normalize fix, can you put that in a separate PR? I have some follow up questions.

Once we put the replacement into normalize.go and made sure that applied to URLs, it started correctly replacing placeholders in URLs.  So a little bit broader than we would have ideally, but I think it is correct and not easily separable.

acpana:
/hold I'll leave it up to you if you want to merge & follow up on any items with Yuwen

This LGTM! Thanks for adding the mocks 🥇 

/lgtm

PR #4797, opened 2025-07-15
Enable mock tests for AlloyDBUser

maqiuyujoyce:
### Change description

Updated mockgcp for AlloyDBUser and add test cases for AlloyDBUser into presubmit.

### Tests you have done


mockgcp/mockalloydb/user.go
```diff
@@ -0,0 +1,139 @@
...
+
+	if obj.Password != "" {
+		obj.Password = ""
+	}
+
+	// AlloyDB user does POST for update in the TF library.
+	oldObj := &pb.User{}
+	err = s.storage.Get(ctx, fqn, oldObj)
+	if err == nil {
+		s.storage.Delete(ctx, fqn, oldObj)
```
justinsb, line 76: Nit: we should check for errors (albeit they are unlikely)

mockgcp/mockalloydb/user.go
```diff
@@ -0,0 +1,139 @@
...
+	if err := s.storage.Get(ctx, fqn, obj); err != nil {
+		return nil, err
+	}
+
+	// Required. A list of fields to be updated in this request.
+	paths := req.GetUpdateMask().GetPaths()
+
+	// TODO: Some sort of helper for fieldmask?
+	for _, path := range paths {
+		topLevelField := strings.Split(path, ".")[0]
```
justinsb, line 107: Nit: I think we should switch on path, if the user specifies `foo.bar` they only want to update foo.bar, not foo

maqiuyujoyce: Agreed. Will update based on the behavior of the test log after switching to direct. Right now TF-based AlloyDBUser update is done by CreateUser().

mockgcp/mockalloydb/user.go
maqiuyujoyce: Will address it later!

PR #4802, opened 2025-07-16
feat: script for beta promote candidates 

yuwenma:
This is the seed for beta promotion.


experiments/promoter/results/candidates.json
```diff
@@ -0,0 +1,485 @@
+[
```
justinsb, line 1: Super-nit, I think yaml is easier for both LLMs and humans

experiments/promoter/scripts/find_promotion_candidates.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+
+# This script finds KCC Kind names that are ready to be promoted from v1alpha1 to v1beta1.
+#
+# A Kind is considered ready for promotion if it meets the following criteria:
```
justinsb, line 5: I love this approach!  Let's make the criteria objective, ensure that there are enough criteria here that the promotion PR is simple (which won't _necessarily_ be the case today, but ... we can iterate), and then we can hand over a lot more to automation here.

experiments/promoter/scripts/find_promotion_candidates.sh
```diff
@@ -0,0 +1,64 @@
...
+# 1. It has a v1alpha1 types file (apis/<service>/v1alpha1/<resource>_types.go)
+#    but no v1beta1 types file (apis/<service>/v1beta1/<resource>_types.go).
+# 2. It has a direct controller implementation (pkg/controller/direct/<service>/<resource>_controller.go).
+# 3. It has existing test fixtures (pkg/test/resourcefixture/testdata/basic/<service>/v1alpha1/<kind>*)
+
+# Initialize an empty array to store the candidates
+candidates=()
+
+# Find all v1alpha1 resource type definitions
+while IFS= read -r alpha_types_file; do
```
justinsb, line 15: I would have guessed python would be easier here, but /shrug

PR #4803, opened 2025-07-16
feat: backport alpha for multi-version CRD

yuwenma:
This is for multi-version CRD  code review. 

It keeps both Alpha and Beta versions in the CRD, even if we moved the alpha api (so git-diff is simpler) 

I verified those exception CRDs. Backporting the alpha is expected:

1. `ComputeFirewallPolicyRule` has been defaulted to Scifi, no alpha version. 
2. `GKEHubFeatureMembership` has been defaulted to Scifi, no alpha version. 
3. `Logginlogmetrics` has been defaulted to Scifi, no alpha version. 
4. `StorageAnywhereCare` is a pure SciFi. It removes its Alpha  when promoting to Beta.

### DIFF
--- a/scripts/crd-tools/cmd/backport-alpha/backport-alpha.go
+++ b/scripts/crd-tools/cmd/backport-alpha/backport-alpha.go
@@ -1,0 +1,124 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package backportalpha
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/scripts/crd-tools/pkg/objectvisitor"
+	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func AddCommand(parent *cobra.Command) {
+	var opt Options
+	cmd := &cobra.Command{
+		Use:   "backport-alpha",
+		Short: "Ensures CRDs have a v1alpha1 version if they have a v1beta1 version.",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return Run(cmd.Context(), opt)
+		},
+	}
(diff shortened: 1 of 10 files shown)


scripts/crd-tools/cmd/backport-alpha/backport-alpha.go
```diff
@@ -0,0 +1,124 @@
...
+type visitor struct {
+}
+
+func (v *visitor) VisitObject(obj *unstructured.Unstructured) error {
+	if obj.GetKind() != "CustomResourceDefinition" {
+		return nil
+	}
+
+	labels := obj.GetLabels()
+	if labels["cnrm.cloud.google.com/tf2crd"] == "true" || labels["cnrm.cloud.google.com/dcl2crd"] == "true" {
```
justinsb: Could we change the selection logic here?  For example, it might be nice to looks at `apis/<group>/<version>` and search for an annotation like `//+kcc:additional-versions=v1alpha1`.  The goal is to be much more self-documenting and less magical.  Also then we wouldn't automatically add these alpha versions in this PR, where I don't _think_ we want them because we went straight to v1beta1 (IIRC)?

justinsb:
I love this approach, but can we tweak the selection logic?  (It might be fun to try it with gemini-cli also, I think it can probably one-shot it)

justinsb:
> ComputeFirewallPolicyRule has been defaulted to Scifi, no alpha version.
> GKEHubFeatureMembership has been defaulted to Scifi, no alpha version.
> Logginlogmetrics has been defaulted to Scifi, no alpha version.

If we never had an alpha version, I don't think we need to add one.

> StorageAnywhereCare is a pure SciFi. It removes its Alpha when promoting to Beta.

If we _recently_ removed the alpha version, we should probably add it back.  In general I think promotion from alpha-to-beta should add the marker (or whatever we decide) so that we don't drop the alpha version and cause customers upgrade problems.

scripts/crd-tools/cmd/backport-alpha/backport-alpha.go
yuwenma: I like the idea.  A kcc comment is a little bit hard because controller-gen would remove comment, so I used the existing kubebuilder comment labels (with some cleanup). And It support any versions.

yuwenma:
> > ComputeFirewallPolicyRule has been defaulted to Scifi, no alpha version.
> > GKEHubFeatureMembership has been defaulted to Scifi, no alpha version.
> > Logginlogmetrics has been defaulted to Scifi, no alpha version.
> 
> If we never had an alpha version, I don't think we need to add one.
> 
> > StorageAnywhereCare is a pure SciFi. It removes its Alpha when promoting to Beta.
> 
> If we _recently_ removed the alpha version, we should probably add it back. In general I think promotion from alpha-to-beta should add the marker (or whatever we decide) so that we don't drop the alpha version and cause customers upgrade problems.

Agree. please see the commit "add an example (AnywhereCache)". AnywhereCache is indeed recently added. so I added it back. Now the script won't touch APIs unless it has the special marker.

justinsb (approved):

PR #4805, opened 2025-07-16
chore: Remove redundant managed-by-cnrm label

justinsb:
The managed-by-cnrm label is already added by the common
labeling function. This change removes the redundant addition
in the forwardingrule controller.

### DIFF
--- a/pkg/controller/direct/compute/forwardingrule/forwardingrule_controller.go
+++ b/pkg/controller/direct/compute/forwardingrule/forwardingrule_controller.go
@@ -106,8 +106,7 @@ func (m *forwardingRuleModel) AdapterForObject(ctx context.Context, reader clien
 	// Get location
 	location := obj.Spec.Location
 
-	// Set label managed-by-cnrm: true
-	obj.ObjectMeta.Labels["managed-by-cnrm"] = "true"
+	
 
 	// Handle TF default values
 	if obj.Spec.LoadBalancingScheme == nil {


pkg/controller/direct/compute/forwardingrule/forwardingrule_controller.go
```diff
@@ -106,8 +106,7 @@ func (m *forwardingRuleModel) AdapterForObject(ctx context.Context, reader clien
 	// Get location
 	location := obj.Spec.Location
 
-	// Set label managed-by-cnrm: true
-	obj.ObjectMeta.Labels["managed-by-cnrm"] = "true"
+
```
cheftako: Would be nice if it left 1 blank line here instead of 3. Otherwise very nice :)

justinsb: Agreed, will fix.  Pretty sure this is going to fail tests though (so I'd like to let it run), and I'm trying to figure out the "why" here.

justinsb:
cc @gemmahou

gemmahou:
This change will be included in #4748: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4748/files#diff-dd4eb3bc8dda217622202de5da7b7ddf423295e7b856b2277a1037106021f656L109-L111

justinsb:
OK, let's decide on #4808 (what Gemini produced) or #4748, and then yes I suspect we don't need this one.  I was very confused by this line though!

/hold

(Gemini was also confused because this CRD still has the tf2crd label, I don't think it should if we have moved to direct)

justinsb:
> (Gemini was also confused because this CRD still has the tf2crd label, I don't think it should if we have moved to direct)

Ah I see, it's only direct when `spec.Target.GoogleAPIsBundle` is specified :+1:

pkg/controller/direct/compute/forwardingrule/forwardingrule_controller.go
justinsb: This was fixed by a gofmt :-)

justinsb:
/hold cancel

I think we've figured this one out.

/assign @yuwenma 

(You actually called this out on #4808)

PR #4807, opened 2025-07-16
Add instructions for gemini to generate propose release tag PR

xiaoweim:
### Change description

Fixes #

### Tests you have done


dev/release/propose-release.md
```diff
@@ -0,0 +1,13 @@
+This is the instruction for propose a new release for config connector.
```
justinsb, line 1: We might need to add a "pointer" in GEMINI.md so that it finds this.  We'll have to figure out organization overall, the good news is that it is now "testable" ("can Gemini find it") rather than subjective.

(Does Gemini find this if you ask it to do a release?)

xiaoweim: When I tested it, I gave the file path, so it worked. But let me try just telling it to do a release :)

PR #4808, opened 2025-07-16
fix: filter labels in direct ComputeForwardingRule controller

justinsb:


pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrule-direct/create.yaml
```diff
@@ -17,6 +17,8 @@ kind: ComputeForwardingRule
 metadata:
   labels:
     label-one: "value-one"
+    foo/not-valid-in-gcp: bar
+    foo: bar
```
gemmahou, line 21: nit: `foo: bar` seems redundant as we have `label-one: "value-one"`. To test update, we could simply update it to `label-one: "value-two"` in below update.yaml.

pkg/controller/direct/compute/forwardingrule/forwardingrule_controller.go
yuwenma, line 1: Can we remove line 110? (managed-by-cnrm)

pkg/controller/direct/compute/forwardingrule/forwardingrule_controller.go
```diff
@@ -214,13 +215,14 @@ func (a *forwardingRuleAdapter) Create(ctx context.Context, createOp *directbase
 	mapCtx := &direct.MapContext{}
 
 	desired := a.desired.DeepCopy()
+	sanitizedLabels := label.NewGCPLabelsFromK8sLabels(desired.Labels)
```
yuwenma, line 218: This meta label will override the spec labels, but I think we want the opposite: 
- If spec.labels are purposely used, we ignore meta labels.
Can we use `labels.ComputeLabels` in adapter initialization so we can get rid of these code changes?

gemmahou: We don't support `spec.labels` in this PR yet, so I'm not concerned about the override and it's fine to keep the code as is. I'll use `labels.ComputeLabels` once we add the `spec.labels` field.

yuwenma: I see. That makes sense. So basically this PR handles the direct controller backward compatibility support for labels, and in the future we can add the spec.labels.

PR #4806, opened 2025-07-16
feat: API check for second-layer unstructured object

yuwenma:
### Change description

Fixes #

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #4809, opened 2025-07-16
feat: check if alpha has full API coverage

yuwenma:
Diffbase from https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4806


tests/apichecks/crds_test.go
```diff
@@ -466,11 +466,14 @@ func TestCRDFieldPresenceInUnstructured(t *testing.T) {
 	for _, crd := range crds {
 		for _, version := range crd.Spec.Versions {
 
+			kind := crd.Spec.Names.Kind
+
 			if version.Name == "v1alpha1" {
-				continue
+				if os.Getenv("TARGET_KIND") == "" || os.Getenv("TARGET_KIND") != kind {
```
justinsb, line 474: A comment would be good here I think, or a log message?  It's somewhat unobvious (that TARGET_KIND will check alpha coverage).

experiments/promoter/scripts/check_api_coverage.py
```diff
@@ -0,0 +1,131 @@
...
+                    candidate['apiValidation'] = False
+                    found_exception = True
+                    break
+        
+        if found_exception:
+            updated_candidates.append(candidate)
+            continue
+
+        # Run the specific test
+        command = f"TARGET_KIND={kind} go test ./tests/apichecks/... -run TestCRDFieldPresenceInUnstructured"
```
justinsb: Nit: I feel like this test should be called "TestCRDFieldPresenceInTests" or something like that.  Unstructured is an implementation detail.

experiments/promoter/scripts/check_api_coverage.py
```diff
@@ -0,0 +1,131 @@
...
+        result = subprocess.run(command, shell=True, capture_output=True, text=True, cwd=project_root)
+
+        if result.returncode == 0:
+            print(f"  PASS: {kind}")
+            candidate['apiValidation'] = True
```
justinsb: Nit: I think `apiCoverage` might be more accurate?  We have a separate "linter" test that I worry we'll confuse with this one...

justinsb (approved):
A few suggestions but LGTM!

experiments/promoter/scripts/check_api_coverage.py
yuwenma: I have a list of XXXValidation checks, so this comes from that. But `apiCoverage` is more descriptive (focus on goal than functionality), and I like it! 

updated.

justinsb (approved):

justinsb (approved):

Issue #4791, opened 2025-07-15
Our docs should be more GitHub compatible


justinsb:
### Checklist

- [x] I did not find a related open enhancement request.
- [x] I understand that enhancement requests filed in the GitHub repository are by default low priority.
- [x] If this request is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Describe the feature or resource

Right now the docs that are on GitHub are in markdown format, but are not readable.  That's because we used some google-internal extensions to markdown.  Ideally we would not need those, and if we did the version on GitHub should still be readable (e.g. can we apply transformations as part of our internal publishing process)

PR #4812, opened 2025-07-17
Promote APIQuotaAdjusterSettings  (Gemini/MCP-driven)

yuwenma:
(This is the pure Gemni-cli/MCP generated PR)

This is the first resource in candidates.json (with `apiVersion=true`). It means it has full API coverage, and it is ready for Beta without additional API, controller or test changes.  

Notes:
- The `cloudquota` service contains another alpha API that is not ready to promote, this introduces two api imports (v1alpha1 and v1beta1) to controller and mapper, thus the imports are distinguished specifically.
-  `make read-pr` requires beta resource to has google reference doc. we should split the process, and improve according to https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/4791


apis/cloudquota/v1beta1/quotaadjustersettings_types.go
```diff
@@ -12,13 +12,13 @@
...
 // limitations under the License.
 
-package v1alpha1
+package v1beta1
 
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	refv1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
-	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/apis/k8s/v1alpha1"
```
yuwenma, line 21: we already have conditions for both alpha and beta, they are the same

apis/cloudquota/v1beta1/quotaadjustersettings_types.go
```diff
@@ -12,13 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package v1alpha1
+package v1beta1
```
justinsb, line 15: This is awesome - a simple file move!

pkg/controller/direct/cloudquota/quotaadjustersettings_controller.go
```diff
@@ -124,8 +124,8 @@ func (a *apiQuotaAdjusterSettingsAdapter) Find(ctx context.Context) (bool, error
 // Create is not supported for QuotaAdjusterSettings. This resource is managed by GCP implicitly.
```
justinsb: We might want to say something like "this function should not be reachable, because Find always returns an object"

pkg/controller/direct/cloudquota/quotaadjustersettings_controller.go
```diff
@@ -27,7 +27,7 @@ import (
 
 	gcp "cloud.google.com/go/cloudquotas/apiv1beta"
 	pb "cloud.google.com/go/cloudquotas/apiv1beta/cloudquotaspb"
-	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/cloudquota/v1alpha1"
+	cloudquotav1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/apis/cloudquota/v1beta1"
```
justinsb, line 30: Nit: so the reason we call it krm and not cloudquotav1alpha1 is because that way when we change the import we don't have to change the code.

Do you want to add a note to GEMINI.md for next time? :-)

scripts/generate-google3-docs/resource-reference/overview.md
```diff
@@ -173,6 +173,10 @@ issues for {{product_name_short}}.
   <td><a href="/cloud-build/docs/">{{builder_name}}</a></td>
   <td><a href="/config-connector/docs/reference/resource-docs/cloudbuild/cloudbuildworkerpool.md">CloudBuildWorkerPool</a></td>
 </tr>
+<tr>
+  <td><a href="/quotas/docs/">Cloud Quotas</a></td>
```
justinsb, line 177: Do we care about alphabetical order here?

I do wonder if we could generate this file in future...

pkg/controller/direct/cloudquota/quotaadjustersettings_controller.go
yuwenma: Sure, updated.

scripts/generate-google3-docs/resource-reference/overview.md
yuwenma: I'll leave this to the next PR. The reference generator really needs a big change.

yuwenma:
/hold cancel

resolved comments. It also contains the bug fix in the doc generator https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4815

scripts/generate-google3-docs/resource-reference/generated/resource-docs/cloudquota/apiquotaadjustersettings.md
```diff
@@ -28,7 +28,7 @@
 </tr>
 <tr>
 <td>{{product_name_short}} Resource Short Names</td>
-<td>APIQuotaAdjusterSettings<br>gcpapiquotaadjustersettings<br>apiquotaadjustersettings<br>apiquotaadjustersettings</td>
```
jingyih: Was this change driven by a change in template? Or was this manual?

jingyih (commented):
/lgtm

scripts/generate-google3-docs/resource-reference/generated/resource-docs/cloudquota/apiquotaadjustersettings.md
yuwenma: Related to https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4815

In short, some kinds have special plural/singular format but the KCC reference doc generator handles it incorrectly, fixed in 4815.

PR #4815, opened 2025-07-17
fix: reference-doc overlook plural Kind

yuwenma:
Bug found in `APIQuotaAdjusterSettings` where this is a plural as singular Kind and is better to keep as it is "APIQuotaAdjusterSettings". 

The usage of `text.Pluralize` overlooks the case and uses  `APIQuotaAdjusterSettingses` which breaks itself.

### DIFF
--- a/scripts/generate-google3-docs/resource-reference/main.go
+++ b/scripts/generate-google3-docs/resource-reference/main.go
@@ -852,7 +852,22 @@ func crdFilePathForGVK(gvk schema.GroupVersionKind) string {
 }
 
 func crdFileNameForGVK(gvk schema.GroupVersionKind) string {
-	crdName := fmt.Sprintf("%s.%s", text.Pluralize(gvk.Kind), gvk.Group)
+	plural := text.Pluralize(gvk.Kind)
+
+	// special plural formats. The best practice is to use singular as is
+	alreadyPluralWords := []string{"settings", "metrics", "series", "data"}
+
+	switch gvk.Kind {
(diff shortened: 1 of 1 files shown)


scripts/generate-google3-docs/resource-reference/main.go
```diff
@@ -852,7 +852,22 @@ func crdFilePathForGVK(gvk schema.GroupVersionKind) string {
 }
 
 func crdFileNameForGVK(gvk schema.GroupVersionKind) string {
-	crdName := fmt.Sprintf("%s.%s", text.Pluralize(gvk.Kind), gvk.Group)
+	plural := text.Pluralize(gvk.Kind)
+
+	// special plural formats. The best practice is to use singular as is
+	alreadyPluralWords := []string{"settings", "metrics", "series", "data"}
```
yuwenma, line 858: This logic is the same as API validation on singular/plural format https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/tests/apichecks/crds_test.go#L810

jingyih: I previously added a special case for "IAPSettings" in `text.Pluralize()`. maybe we should consolidate the special case handling?

yuwenma: Agree, code sanity is always good! I also checked text.Pluralize, it is mostly called by legacy code path. So maybe we can get rid of it completely once finish the migration. (in other words, touching legacy code always gives some surprises that seem to be not deserve the effort to fix)

PR #4816, opened 2025-07-17
tests: IAMPartialPolicy resourceRef test

acpana:
Splitting off one resource test from https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4800 as the crux of that PR.

### DIFF
--- a/pkg/test/resourcefixture/testdata/iam/iampartialpolicy/resources/project-conditions/_generated_object_project-conditions.golden.yaml
+++ b/pkg/test/resourcefixture/testdata/iam/iampartialpolicy/resources/project-conditions/_generated_object_project-conditions.golden.yaml
@@ -1,0 +1,77 @@
+apiVersion: iam.cnrm.cloud.google.com/v1beta1
+kind: IAMPartialPolicy
+metadata:
+  annotations:
+    cnrm.cloud.google.com/state-into-spec: absent
+  finalizers:
+  - cnrm.cloud.google.com/finalizer
+  - cnrm.cloud.google.com/deletion-defender
+  generation: 2
+  labels:
+    cnrm-test: "true"
+  name: iampartialpolicy-${uniqueId}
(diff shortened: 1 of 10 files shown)


pkg/test/resourcefixture/resourcefixture.go
```diff
@@ -69,6 +69,67 @@ func Load(t *testing.T) []ResourceFixture {
 type LightFilter func(name string, testType TestType) bool
 type HeavyFilter func(fixture ResourceFixture) bool
 
+func LoadWithPathFilter(t *testing.T, pathFilter func(path string) bool, lightFilterFunc LightFilter, heavyFilterFunc HeavyFilter) []ResourceFixture {
```
justinsb, line 72: Don't we already have this?  What's new?

tests/e2e/unified_test.go
```diff
@@ -817,19 +821,28 @@ func TestIAM_AllInSeries(t *testing.T) {
...
-		// Only run fixtures under iam/iampartialpolicy
-		lightFilter := func(name string, testType resourcefixture.TestType) bool {
-			return strings.Contains(name, "iam-bigqueryconnectionconnectionref") ||
-				strings.Contains(name, "iam-logsinkref") ||
-				strings.Contains(name, "iam-serviceaccountref") ||
-				strings.Contains(name, "iam-serviceidentityref") ||
-				strings.Contains(name, "iam-sqlinstanceref")
-		}
-		fixtures := resourcefixture.LoadWithFilter(t, lightFilter, nil)
+		fixtures := resourcefixture.LoadWithPathFilter(t, func(path string) bool {
```
justinsb, line 828: Ah I see!  So this makes sense, except why not put these under the existing hierarchy?  Are they different from other tests?

justinsb (approved):
One question about whether we can put these under the same hierarchy as non-IAM tests, but not a blocker

PR #4823, opened 2025-07-18
Release 1.133.0

xiaoweim:
### Change description

Fixes #

### Tests you have done


dev/release/propose-release.md
yuwenma, line 1: We don't want this release PR to contain other code changes.

yuwenma:
/approve
/lgtm

Thanks! This PR is very large. I spot checked all files but still cannot guarantee the crd changes are 100% correct. If it's purely from script, I suggest we add presubmit test to verify. I don't want to block this PR due to the 1.133 urgencies. Could you add a more concrete validator about this? (otherwise future releases would be slowed down and hit the same challenge again and again).

PR #4829, opened 2025-07-21
feat: add tests for bigquerydataset

xiaoweim:
### Change description

This PR adds tests for bigquery dataset.
  - Adds tests for no change update.
  - Adds the bigquerydataset resource kind to TestAcquire

### Tests you have done


apis/bigquery/v1beta1/dataset_identity.go
```diff
@@ -59,6 +59,9 @@ func NewDatasetIdentity(ctx context.Context, reader client.Reader, obj *BigQuery
 	if err != nil {
 		return nil, err
 	}
+	if projectRef == nil {
+		return nil, fmt.Errorf("cannot resolve projectRef %s", obj.Spec.ProjectRef)
```
maqiuyujoyce: Nit: Could you make the error message more specific? E.g. "obj.Spec.ProjectRef not defined".

Though I'm also curious: why is projectRef field optional: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/bigquery/v1beta1/dataset_types.go#L104?

pkg/controller/dynamic/dynamic_controller_integration_test.go
```diff
@@ -192,7 +192,8 @@ func TestAcquire(t *testing.T) {
 			"Folder": true,
 			// used as an integration test verifying that falsey values are not
 			// incorrectly defaulted. (b/178744782)
-			"ComputeNetwork": true,
+			"ComputeNetwork":  true,
+			"BigQueryDataset": true,
```
maqiuyujoyce, line 196: Do you mind sharing test results in the description as those added test are not covered by presubmit?

(We should consider supporting them over time.)

apis/bigquery/v1beta1/dataset_identity.go
xiaoweim: This was because the terraform behavior for `bigquerydataset` marks projectRef as optional.

pkg/controller/dynamic/dynamic_controller_integration_test.go
xiaoweim: Sure, here are the test logs:
- basicbigquerydataset: https://paste.googleplex.com/6586475956928512
- bigquerydatasetaccessfull: https://paste.googleplex.com/4755596976586752
- fullbigquerydataset: https://paste.googleplex.com/5969923079405568

But I realized that for direct controller resources, the testing framework that we currently have for `TestAcquire` actually does not work. In [testReconcileAcquire](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/dynamic/dynamic_controller_integration_test.go#L691), if the GCP resource object does not exist, we first [create the GCP resource object](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/dynamic/dynamic_controller_integration_test.go#L711) using [directCreate](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/test/resourcefixture/contexts/register.go#L419). In direct controllers CREATE function calls, we make an [update call](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/bigquerydataset/dataset_controller.go#L184) to the kubernetes API. But at this time, in the `TestAcquire`, the kubernetes object is not created yet, the kubernetes object is created later at https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/dynamic/dynamic_controller_integration_test.go#L739. So the test would fail inevitably.

When I run the above tests locally, I hard coded the bigquerydataset name to some bigquerydatasets that already exists in my project, so I do not have to go through the create GCP resource stage. We probably need to fix this if we want to add these times to presubmit and log them in the future.

apis/bigquery/v1beta1/dataset_identity.go
xiaoweim: I checked the previous PR where we had the discussion that projectRef was optional, it was because terraform controller for bigquerydataset does not require `projectRef` due project level defaulting for KCC resources. 

But when I ran testAcquire for both terraform based and direct controller based `BigQueryDataset`, I realize that it all require projectRef to be set. It seems like that our fixture testing framework does project defaulting.  It terraform does not require, is it okay for us to mark it as required? If it is okay, I can change it to required since judging from the rest API https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets/insert, it only requires projectID in the request PATH.

maqiuyujoyce: > But when I ran testAcquire for both terraform based and direct controller based BigQueryDataset, I realize that it all require projectRef to be set. It seems like that our fixture testing framework does project defaulting.

Hmmm, if it does project defaulting, then it won't require projectRef to be set, right? Could you clarify on this?

>  It terraform does not require, is it okay for us to mark it as required?

For backwards compatibility purposes, no, we should not make a field required if it was not. Especially for critical data plane resources like BigQueryDataset.
Regarding the direct controller, just to double check: Does it work if projectRef is unset but the annotation for project-id is?

pkg/controller/dynamic/dynamic_controller_integration_test.go
maqiuyujoyce: Sounds good about local test only for now. Could you file a ticket about this gap?

xiaoweim: Sure! Do I file a github issue or an internal ticket?

apis/bigquery/v1beta1/dataset_identity.go
xiaoweim: > Does it work if projectRef is unset but the annotation for project-id is?

Yes, it does!

When the projectRef and the annotation for project-id are both not set,  it will look for the project id same as the namespace: [reference](https://cloud.google.com/config-connector/docs/how-to/organizing-resources/project-scoped-resources#using_the_namespace_name_as_the_project_id)

pkg/controller/dynamic/dynamic_controller_integration_test.go
maqiuyujoyce: Internal ticket should be good.

maqiuyujoyce:
/lgtm
/approve

Please keep an eye on the postsubmit tests after the pr is merged.

PR #4830, opened 2025-07-21
Decide default model during runtime

maqiuyujoyce:
### Change description

Our default Gemini model is no longer supported (again). Supported the function to determine the model during runtime instead of hard-coding the default one.

### Tests you have done

### DIFF
--- a/dev/tools/controllerbuilder/pkg/llm/options.go
+++ b/dev/tools/controllerbuilder/pkg/llm/options.go
@@ -46,8 +46,9 @@ func (o *Options) InitDefaults() {
 
 func (o *Options) InitDefaultsWithLatestModelIfUnset(ctx context.Context, llmClient gollm.Client) error {
 	model := os.Getenv("LLM_MODEL")
-	if model == "" {
 
+	modelNameSubString := "gemini-2.5-pro"
+	if model == "" {
 		models, err := llmClient.ListModels(ctx)
 		if err != nil {
 			return fmt.Errorf("listing models: %w", err)
@@ -56,11 +57,15 @@ func (o *Options) InitDefaultsWithLatestModelIfUnset(ctx context.Context, llmCli
 		for _, m := range models {
 			// There are many old or experimental models in the list.
 			// Let's use the first `gemini-2.5-pro` model for now.
-			if strings.Contains(m, "gemini-2.5-pro") {
+			if strings.Contains(m, modelNameSubString) {
 				model = m
 				break
 			}
 		}
+	}
+	if model == "" {
+		klog.Errorf("No model containing %q in name found", modelNameSubString)
+		return fmt.Errorf("cannot find model containing %q in name", modelNameSubString)
 	}
 
 	klog.Infof("Using model %v", model)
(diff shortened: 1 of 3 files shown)


dev/tools/controllerbuilder/pkg/llm/options.go
```diff
@@ -41,6 +44,30 @@ func (o *Options) InitDefaults() {
...
+
+		models, err := llmClient.ListModels(ctx)
+		if err != nil {
+			return fmt.Errorf("listing models: %w", err)
+		}
+
+		for _, m := range models {
+			// There are many old or experimental models in the list.
+			// Let's use the first `gemini-2.5-pro` model for now.
+			if strings.Contains(m, "gemini-2.5-pro") {
```
anhdle-sso: use `DefaultModel` constant

dev/tools/controllerbuilder/pkg/llm/options.go
```diff
@@ -41,6 +44,30 @@ func (o *Options) InitDefaults() {
...
+				model = m
+				break
+			}
+		}
+	}
+
```
anhdle-sso, line 70: Nit: Add a check if no model is found (no 2.5 pro), and print out a log for easier debug. If no model found, we can actually just return an error.

dev/tools/controllerbuilder/pkg/llm/options.go
maqiuyujoyce: It is actually a coincident that the substring we are looking for in the model name is the same as a valid model name. I'd expect this check becomes a regex matching eventually.

dev/tools/controllerbuilder/pkg/llm/options.go
maqiuyujoyce: Done.

dev/tools/controllerbuilder/pkg/llm/options.go
```diff
@@ -41,6 +44,35 @@ func (o *Options) InitDefaults() {
 	o.Model = model
 }
 
+func (o *Options) InitDefaultsWithLatestModelIfUnset(ctx context.Context, llmClient gollm.Client) error {
+	model := os.Getenv("LLM_MODEL")
+
+	modelNameSubString := "gemini-2.5-pro"
```
yuwenma, line 50: If I read the code correctly, it does not set the latest model as the default. It still requires manual code changes everytime a new model comes, right?

maqiuyujoyce: Yes for now, not really for the future. We can have the regex to not only match `2.5` but also any `X.Y` and pick any that is above 2.5.

dev/tools/controllerbuilder/pkg/llm/options.go
```diff
@@ -41,6 +44,35 @@ func (o *Options) InitDefaults() {
...
+	if model == "" {
+		models, err := llmClient.ListModels(ctx)
+		if err != nil {
+			return fmt.Errorf("listing models: %w", err)
+		}
+
+		for _, m := range models {
+			// There are many old or experimental models in the list.
+			// Let's use the first `gemini-2.5-pro` model for now.
+			if strings.Contains(m, modelNameSubString) {
```
anhdle-sso, line 60: Would it make sense to sort the models before the loop so we get the newest 2.5-pro?

maqiuyujoyce:
As we are moving towards using gemini cli, this PR is no longer needed.

PR #4832, opened 2025-07-21
tests: Exclude output-only spec fields in missingfields.txt

gemmahou:
### Change description

Fixes #

### Tests you have done

### DIFF
--- a/tests/apichecks/testdata/exceptions/outputonlyspecfields.txt
+++ b/tests/apichecks/testdata/exceptions/outputonlyspecfields.txt
@@ -1,3 +1,4 @@
+[output_only_spec_field] crd=cloudidentitymemberships.cloudidentity.cnrm.cloud.google.com version=v1beta1: field ".spec.roles[].restrictionEvaluations.memberRestrictionEvaluation.state" is not set in unstructured objects
 [output_only_spec_field] crd=containerclusters.container.cnrm.cloud.google.com version=v1beta1: field ".spec.clusterAutoscaling.autoProvisioningDefaults.management.upgradeOptions[].autoUpgradeStartTime" is not set in unstructured objects
 [output_only_spec_field] crd=containerclusters.container.cnrm.cloud.google.com version=v1beta1: field ".spec.clusterAutoscaling.autoProvisioningDefaults.management.upgradeOptions[].description" is not set in unstructured objects
 [output_only_spec_field] crd=pubsubsubscriptions.pubsub.cnrm.cloud.google.com version=v1beta1: field ".spec.cloudStorageConfig.state" is not set in unstructured objects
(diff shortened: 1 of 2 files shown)


tests/apichecks/testdata/exceptions/outputonlyspecfields.txt
yuwenma, line 1: we don't want beta resources with new exceptions. Could you add more context on why this is needed?

gemmahou: `.roles[]` field is part of the spec, but `.roles[].restrictionEvaluations.memberRestrictionEvaluation.state` is an output-only field. We should add it to outputonlyspecfields.txt to remove its entry from missingfields.txt, since we don't test output-only fields.

Ref: API doc https://cloud.google.com/identity/docs/reference/rest/v1beta1/groups.memberships#restrictionevaluation

maqiuyujoyce: These are not new exceptions but instead a workaround to skip output-only spec fields when we check field coverage.

tests/apichecks/testdata/exceptions/outputonlyspecfields.txt
```diff
@@ -1,3 +1,6 @@
+[output_only_spec_field] crd=cloudidentitymemberships.cloudidentity.cnrm.cloud.google.com version=v1beta1: field ".spec.roles[].restrictionEvaluations" is not set in unstructured objects
+[output_only_spec_field] crd=cloudidentitymemberships.cloudidentity.cnrm.cloud.google.com version=v1beta1: field ".spec.roles[].restrictionEvaluations.memberRestrictionEvaluation" is not set in unstructured objects
```
maqiuyujoyce: Looks like there is only one missing field entry...? Could you help me understand why row 1 and 2 are also added?

gemmahou: roles[].restrictionEvaluations.memberRestrictionEvaluation.state is a nested field, with only one field each layer(state is the only field under memberRestrictionEvaluation and memberRestrictionEvaluation is the only field under restrictionEvaluations) and thus they all considered as output-only. So we have to exclude state, memberRestrictionEvaluation and restrictionEvaluations.

maqiuyujoyce: My understanding by reading `tests/apichecks/testdata/exceptions/missingfields.txt` is that `roles[].restrictionEvaluations` and `roles[].restrictionEvaluations.memberRestrictionEvaluation` should not be detected as missing fields.

This makes sense from the API reference doc as `roles[].restrictionEvaluations` and `roles[].restrictionEvaluations.memberRestrictionEvaluation` cannot be used by themselves.

`tests/apichecks/testdata/exceptions/outputonlyspecfields.txt` is created just to cover uncoverable spec fields, so
 we don't really need to have the two additional ones in outputonlyspecfields.txt. We should minimize the manual work when maintaining outputonlyspecfields.txt.

gemmahou: I think this is related to a recent change: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4806/files#, a filter function was added to remove parent entries by checking the prefix. However, this introduced a problem: if a child entry(state) is excluded, its parent(memberRestrictionEvaluation) remains in missingfields.txt. This occurs because when memberRestrictionEvaluation.state is excluded, making the parent the new "child" that the filter fails to remove. I believe the original method of checking the fields' properties count is a more correct approach.

I had a separate PR to address above problem. #4839

maqiuyujoyce (commented):
/lgtm
/approve

PR #4834, opened 2025-07-22
chore: update propose release instructions and add presubmits for updating crds/ folder

xiaoweim:
### Change description

- Update the prompts for propose release candidate.
- Add a script to verify the CRD updates in the crds/ folder
- Add a presubmit test, that is only triggered when changes are made to the crds/ folder

Fixes #

### Tests you have done

### DIFF
--- a/.github/workflows/presubmit-crds.yaml
+++ b/.github/workflows/presubmit-crds.yaml
@@ -1,4 +1,4 @@
-# Copyright 2024 Google LLC
+# Copyright 2025 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/.github/workflows/presubmit-crds.yaml
+++ b/.github/workflows/presubmit-crds.yaml
@@ -35,6 +35,4 @@ jobs:
           go-version-file: 'go.mod'
       - name: "Verify CRDs folder is up-to-date"
         run: |
-          LATEST_VERSION=$(ls operator/channels/packages/configconnector/ | sort -V | tail -n 1)
-          echo "Latest version found: ${LATEST_VERSION}"
-          VERSION=${LATEST_VERSION} dev/tasks/verify-crds-folder.sh
+          dev/tasks/verify-crds-folder.sh
--- a/.github/workflows/presubmit-crds.yaml
+++ b/.github/workflows/presubmit-crds.yaml
@@ -1,0 +1,40 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


.github/workflows/presubmit-crds.yaml
```diff
@@ -0,0 +1,40 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+      - name: "Verify CRDs folder is up-to-date"
+        run: |
+          LATEST_VERSION=$(ls operator/channels/packages/configconnector/ | sort -V | tail -n 1)
```
justinsb: In general I would suggest avoiding putting logic into github actions, it makes it hard to run the same test locally if it fails.  In this case, either `verify-crds-folder` can always set VERSION to this, or it can default to fetching it from operator/channels/packages/configconnector`

dev/tasks/verify-crds-folder.sh
```diff
@@ -0,0 +1,80 @@
...
+
+if [[ -z "${VERSION:-}" ]]; then
+  echo "VERSION must be set"
+  exit 1
+
+fi
+
+TARGET_CRD_DIR=${REPO_ROOT}/crds
+TEMP_DIR=$(mktemp -td verify-crds-folder.XXXXXXXX)
+CRDS_FILE=${REPO_ROOT}/operator/channels/packages/configconnector/${VERSION}/crds.yaml
```
justinsb, line 28: Do we actually expect this file to be in sync other than when we have just done a release?

dev/tasks/verify-crds-folder.sh
```diff
@@ -0,0 +1,80 @@
...
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+
+if [[ -z "${VERSION:-}" ]]; then
+  echo "VERSION must be set"
+  exit 1
+
+fi
+
+TARGET_CRD_DIR=${REPO_ROOT}/crds
```
justinsb, line 26: Ah, maybe this is the problem, we have all these different copies of the CRDs floating around.  I got rid of one extra one recently, but I think we need to decide how our repo should look.  I worry the current state is confusing.

dev/tasks/verify-crds-folder.sh
xiaoweim: No, we only expect this to be in sync when we are cutting a new release. 

The idea is that I want to run this verification everytime we cut a new release. Right now I set the condition to only run in presubmit when the crds/ folder are changed in the PR (assuming we only change it during release)

dev/tasks/verify-crds-folder.sh
cheftako: +1. However is the right place (PR) to fix the issue? I think that deserves a bit of thought and its own PR.

dev/tasks/verify-crds-folder.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
+# Copyright 2024 Google LLC
```
cheftako: 2025

.github/workflows/presubmit-crds.yaml
```diff
@@ -0,0 +1,38 @@
+# Copyright 2024 Google LLC
```
cheftako: 2025

PR #4835, opened 2025-07-22
tests: Add comments to CloudIdentity scenarios tests

gemmahou:
### Change description

- Add comments to CloudIdentity scenarios tests
- Add additional CloudIdentityMembership direct scenarios test

We have explained the behavior change in [release notes 1.132.0](https://github.com/GoogleCloudPlatform/k8s-config-connector/releases/tag/v1.132.0); We could do the same as well once we enable direct controller.

### Tests you have done


maqiuyujoyce (commented):
I understand that the comments help readability, but why do we need test case to cover DCL-based CloudIdentityMembership? I thought we are turning it to a pure direct resource.

gemmahou:
> I understand that the comments help readability, but why do we need test case to cover DCL-based CloudIdentityMembership? I thought we are turning it to a pure direct resource.

Right, I separated -dcl and -direct test so that we can compare the logs before turning on direct controller. (And the -dcl test will be removed soon in the turn-on PR)

maqiuyujoyce (commented):
/lgtm
/approve

PR #4838, opened 2025-07-22
feat: separate op for maintenanceVersion

acpana:
feat: separate op for maintenanceVersion for SQLInstance

Note the error without the patch on the controller:

```
...
Upgrading maintenance version and changing other fields at the same time is not allowed.
...
```

```release-note
Fix bug in SQLInstance `maintenanceVersion` UPDATE operation
```

### DIFF
--- a/mockgcp/mocksql/sqlinstance.go
+++ b/mockgcp/mocksql/sqlinstance.go
@@ -757,6 +757,15 @@ func (s *sqlInstancesService) Update(ctx context.Context, req *pb.SqlInstancesUp
 		return nil, err
 	}
 
+	if req.GetBody().GetMaintenanceVersion() != "" && req.GetBody().GetMaintenanceVersion() != existing.GetMaintenanceVersion() {
+		// Maintenance version is changing.
+		// Check if any other fields are changing.
+		// A simple check for the test case is to see if settings are changing.
+		if !proto.Equal(req.GetBody().GetSettings(), existing.GetSettings()) {
+			return nil, status.Errorf(codes.InvalidArgument, "Invalid request: Upgrading maintenance version and changing other fields at the same time is not allowed.")
+		}
+	}
+
 	obj := proto.Clone(req.GetBody()).(*pb.DatabaseInstance)
 	obj.Name = existing.Name
 	obj.Region = existing.Region
--- a/mockgcp/mocksql/sqlinstance.go
+++ b/mockgcp/mocksql/sqlinstance.go
@@ -717,8 +717,13 @@ func (s *sqlInstancesService) Patch(ctx context.Context, req *pb.SqlInstancesPat
 		if body.MaintenanceVersion != "" {
 			obj.MaintenanceVersion = body.MaintenanceVersion
 		}
+		// todo kcc team: refactor this all so we can pass in specific values for database settings
+		specifiedMaintenanceVersion := body.MaintenanceVersion
 		if err := setDatabaseVersionDefaults(obj); err != nil {
 			return nil, err
+		}
+		if specifiedMaintenanceVersion != "" {
+			obj.MaintenanceVersion = specifiedMaintenanceVersion
 		}
 	}
 
--- a/pkg/controller/direct/sql/sqlinstance_controller.go
+++ b/pkg/controller/direct/sql/sqlinstance_controller.go
@@ -296,6 +296,34 @@ func (a *sqlInstanceAdapter) Update(ctx context.Context, updateOp *directbase.Up
 		a.actual = updated
 	}
 
+	// we also need to handle maintenanceVersion updates separately ...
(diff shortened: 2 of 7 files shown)


pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/sqlinstance-maintenanceversion-direct/create.yaml
```diff
@@ -0,0 +1,29 @@
...
+kind: SQLInstance
+metadata:
+  name: sqlinstance-maintenanceversion-direct-${uniqueId}
+  annotations:
+    alpha.cnrm.cloud.google.com/reconciler: "direct"
+spec:
+  databaseVersion: POSTGRES_15
+  region: us-central1
+  # cannot set maintenanceVersion on INSERT/ CREATE
+  #maintenanceVersion: POSTGRES_15_13.R20250302.00_31
```
yuwenma: Suggest changing the adaptor.Create to set `maintenanceVersion` *after* CREATE/INSERT

For declarative configurations, users do not expect to set this field in separate steps.

acpana: That's a good point! And I'd like to understand more about this field before we add that kind of magic over the operations.

Also some behavior stuff -- what happens if we fail to set the field but we CREATED it already. Do we delete the object?

I'd expect we attempt to update the object but we have to be careful how we modify the status in the object.

yuwenma: Could you share the error on passing MaintenanceVersion in create/insert?  SQL team mentions this is a required field.

pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/sqlinstance-maintenanceversion-direct/update.yaml
```diff
@@ -0,0 +1,28 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: sql.cnrm.cloud.google.com/v1beta1
+kind: SQLInstance
+metadata:
+  name: sqlinstance-maintenanceversion-direct-${uniqueId}
+  annotations:
+    alpha.cnrm.cloud.google.com/reconciler: "direct"
```
yuwenma: nit: SCiFi is already default in this resource . We don't need this annotation

pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/sqlinstance-maintenanceversion-direct/create.yaml
```diff
@@ -0,0 +1,29 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: sql.cnrm.cloud.google.com/v1beta1
+kind: SQLInstance
+metadata:
+  name: sqlinstance-maintenanceversion-direct-${uniqueId}
+  annotations:
+    alpha.cnrm.cloud.google.com/reconciler: "direct"
```
acpana:
```suggestion
```

pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/sqlinstance-maintenanceversion-direct/update.yaml
acpana:
```suggestion
```

mockgcp/mocksql/sqlinstance.go
```diff
@@ -717,9 +717,14 @@ func (s *sqlInstancesService) Patch(ctx context.Context, req *pb.SqlInstancesPat
 		if body.MaintenanceVersion != "" {
 			obj.MaintenanceVersion = body.MaintenanceVersion
 		}
+		// todo kcc team: refactor this all so we can pass in specific values for database settings
```
justinsb, line 720: Can you also tweak the mock so that it exhibits the behaviour we got wrong?

Ideally we would have

* commit adding test or enhancing existing test (including GCP log, fine to split that into another commit - I would probably create a separate commit if it's a separate test, but if you can reuse an existing test so it's more like a 10 line diff I would keep them in the same commit)
* commit adding mockgcp behaviour (so test now fails)
* commit fixing controller
* commit updating golden output with fixed controller

cheftako: +1

acpana: restructured the commits according to this suggestion 🙏🏼

PR #4839, opened 2025-07-23
tests: Partially revert #4806

gemmahou:
### Change description

Partially revert the filter function added in #4806

I encountered an issue in #4832 so I have this PR to revert the filter function, see https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4832/files#r2224181901

### Tests you have done

### DIFF
--- a/tests/apichecks/crds_test.go
+++ b/tests/apichecks/crds_test.go
@@ -571,6 +571,23 @@ func findFieldsNotCoveredByTests(t *testing.T, shouldVisitCRD func(crd *apiexten
 					return
 				}
 
+				// Skip non-terminal fields. A field is considered non-terminal if it's a struct-like object
+				// (i.e., an object with properties) or an array. Other fields, including primitives and
+				// map-like objects (objects without properties), are considered terminal and will be checked for presence.
+				if field.props != nil {
+					switch field.props.Type {
+					case "object":
+						// Struct-like objects with properties are non-terminal.
+						// Map-like objects without properties are considered terminal.
+						if len(field.props.Properties) > 0 {
+							return
+						}
+					case "array":
+						// Arrays are always non-terminal containers.
+						return
+					}
+				}
+
 				// Any reference field was already handled and handling the children will just double count
 				// Check for `Ref.` or `Refs[].` to ensure it's a reference field,
 				// and avoid field names that include `Ref` (e.g., allowedReferrers in APIKeysKey).
@@ -604,33 +621,6 @@ func findFieldsNotCoveredByTests(t *testing.T, shouldVisitCRD func(crd *apiexten
 	}
 
 	sort.Strings(errs)
-	// Filter out errors for fields that are also reported for their array equivalent.
-	// For example, if we have an error for "spec.foo" and "spec.foo[]", we only keep the latter.
-	if len(errs) > 0 {
-		filteredErrs := make([]string, 0, len(errs))
-		i := 0
-		for i < len(errs) {
-			addCurrent := true
-			if i+1 < len(errs) {
-				currentErr := errs[i]
-				nextErr := errs[i+1]
-				const suffix = "\" is not set in unstructured objects"
-				prefix := strings.TrimSuffix(currentErr, suffix)
-				if strings.HasPrefix(nextErr, prefix) {
-					diff := strings.TrimPrefix(nextErr, prefix)
-					if strings.Contains(diff, "[]") || strings.Contains(diff, ".") {
-						addCurrent = false
-					}
-				}
-			}
-			if addCurrent {
-				filteredErrs = append(filteredErrs, errs[i])
-			}
-			i++
-		}
-		errs = filteredErrs
-	}
-
 	return errs
 }
 
--- a/tests/apichecks/testdata/exceptions/missingfields.txt
+++ b/tests/apichecks/testdata/exceptions/missingfields.txt
@@ -754,6 +754,7 @@
 [missing_field] crd=computenodegroups.compute.cnrm.cloud.google.com version=v1beta1: field ".spec.shareSettings.projectMap[].projectIdRef" is not set; neither 'external' nor 'name' are set
 [missing_field] crd=computenodegroups.compute.cnrm.cloud.google.com version=v1beta1: field ".spec.shareSettings.shareType" is not set in unstructured objects
 [missing_field] crd=computenodetemplates.compute.cnrm.cloud.google.com version=v1beta1: field ".spec.cpuOvercommitType" is not set in unstructured objects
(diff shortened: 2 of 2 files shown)


tests/apichecks/testdata/exceptions/missingfields.txt
```diff
@@ -753,6 +753,7 @@
 [missing_field] crd=computenodegroups.compute.cnrm.cloud.google.com version=v1beta1: field ".spec.shareSettings.projectMap[].projectIdRef" is not set; neither 'external' nor 'name' are set
 [missing_field] crd=computenodegroups.compute.cnrm.cloud.google.com version=v1beta1: field ".spec.shareSettings.shareType" is not set in unstructured objects
 [missing_field] crd=computenodetemplates.compute.cnrm.cloud.google.com version=v1beta1: field ".spec.cpuOvercommitType" is not set in unstructured objects
+[missing_field] crd=computenodetemplates.compute.cnrm.cloud.google.com version=v1beta1: field ".spec.nodeType" is not set in unstructured objects
```
gemmahou, line 757: The filter function mistakenly considered `nodeType` field as parent field and removed it. Because this field has the same prefix as below `nodeTypeFlexibility` field.

Added it back.

tests/apichecks/testdata/exceptions/missingfields.txt
```diff
@@ -1209,7 +1210,6 @@
 [missing_field] crd=containerclusters.container.cnrm.cloud.google.com version=v1beta1: field ".spec.authenticatorGroupsConfig.securityGroup" is not set in unstructured objects
 [missing_field] crd=containerclusters.container.cnrm.cloud.google.com version=v1beta1: field ".spec.binaryAuthorization.enabled" is not set in unstructured objects
 [missing_field] crd=containerclusters.container.cnrm.cloud.google.com version=v1beta1: field ".spec.binaryAuthorization.evaluationMode" is not set in unstructured objects
-[missing_field] crd=containerclusters.container.cnrm.cloud.google.com version=v1beta1: field ".spec.clusterAutoscaling.autoProvisioningDefaults.management.upgradeOptions[]" is not set in unstructured objects
```
gemmahou, line 1214: The filter function mistakenly added parent field `upgradeOptions[]`, all its children are specified in outputonlyspecfields.txt: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/252b9988877088e3f08e7f9144337f931606813d/tests/apichecks/testdata/exceptions/outputonlyspecfields.txt#L1-L2

Removed it.

tests/apichecks/testdata/exceptions/missingfields.txt
```diff
@@ -3332,7 +3332,6 @@
 [missing_field] crd=sqlinstances.sql.cnrm.cloud.google.com version=v1beta1: field ".spec.settings.pricingPlan" is not set in unstructured objects
 [missing_field] crd=sqlinstances.sql.cnrm.cloud.google.com version=v1beta1: field ".spec.settings.replicationType" is not set in unstructured objects
 [missing_field] crd=sqlusers.sql.cnrm.cloud.google.com version=v1beta1: field ".spec.password.value" is not set in unstructured objects
-[missing_field] crd=sqlusers.sql.cnrm.cloud.google.com version=v1beta1: field ".spec.passwordPolicy.status[]" is not set in unstructured objects
```
gemmahou, line 3333: The filter function mistakenly added parent field status[], all its children are specified in outputonlyspecfields.txt:
https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/252b9988877088e3f08e7f9144337f931606813d/tests/apichecks/testdata/exceptions/outputonlyspecfields.txt#L4-L5

Removed it.

tests/apichecks/crds_test.go
```diff
@@ -513,6 +513,16 @@ func TestCRDFieldPresenceInTests(t *testing.T) {
 					return
 				}
 
+				// Skip non-terminal fields (fields with children or slices)
+				if field.props != nil {
+					if len(field.props.Properties) > 0 {
```
gemmahou: Previously we had `len(field.props.Properties) > 0 || field.props.Type == "object"` in the if statement. 

`object` type contains `map` so it should not be an indicator of terminal fields. If we encountered any cases related to `object` field in the future, we could handle them separately.

maqiuyujoyce: When `field.props.Type == "object"`, `len(field.props.Properties) == 0` and `field.props.AdditionaProperties.type` is not `object` or `array`, then it is indeed a terminal field. Could you try it out?

maqiuyujoyce: For example, https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_containerattachedclusters.containerattached.cnrm.cloud.google.com.yaml#L67 vs. https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_billingbudgetsbudgets.billingbudgets.cnrm.cloud.google.com.yaml#L288

gemmahou: Edited: `AdditionaProperties` is `*JSONSchemaPropsOrBool` type([ref](https://github.com/kubernetes/apiextensions-apiserver/blob/v0.33.3/pkg/apis/apiextensions/types_jsonschema.go#L71)) so we could use `field.props.AdditionalProperties.**Schema**.Type`. I also tried to use Gemini CLI to refine the logic of non-terminal fields detection and it gave me similar result:
My version:
```
if field.props != nil {
	if len(field.props.Properties) > 0 {
		return
	}
	if field.props.Type == "array" && field.props.Items != nil {
		return // Skip the array itself; focus on its elements
	}
}
```
Gemini version:
```
if field.props != nil {                                                                                                                                                
  switch field.props.Type {                                                                                                                                          
  case "object":                                                                                                                                                                                                                                                  
    if len(field.props.Properties) > 0 {                                                                                                
      return                                                                                                                                                   
    }                                                                                                                                                                                                                                                    
  case "array":                                                                                                                                                                                                                                                               
   return                                                                                                                                                         
  }                                                                                                                                                                  
}                          
```
I also added detailed comment to explain the logic, so we could make improvements in the future if any edge cases occur.

maqiuyujoyce: The new logic won't try to check if the fields in the additionalProperties is set in the test or not (verified by removing `values` field in the test case for BillingBudgetsBudget). Could you ensure this edge case is covered?

gemmahou: This is a real edge case since even the existing logic doesn't check fields under AdditionalProperties. I've handled this in a separate PR: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4920

justinsb:
@gemmahou looks like you have approve - can you please rebase + make sure tests pass.

PR #4840, opened 2025-07-23
Add LogicalView deletion_protection feature.

georgecma:
### Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


pkg/controller/direct/bigtable/logicalview_controller.go
```diff
@@ -129,10 +129,15 @@ func (a *LogicalViewAdapter) Find(ctx context.Context) (bool, error) {
 		return false, fmt.Errorf("getting BigtableLogicalView %q: %w", a.id, err)
 	}
 
+	deletionProtection := false
+	if logicalViewInfo.DeletionProtection != 0 {
```
jingyih, line 136: Isn't `DeletionProtection` a boolean?

tests/apichecks/testdata/exceptions/alpha-missingfields.txt
```diff
@@ -442,6 +442,7 @@
 [missing_field] crd=bigtableclusters.bigtable.cnrm.cloud.google.com version=v1alpha1: field ".spec.location" is not set in unstructured objects
 [missing_field] crd=bigtableclusters.bigtable.cnrm.cloud.google.com version=v1alpha1: field ".spec.nodeScalingFactor" is not set in unstructured objects
 [missing_field] crd=bigtableclusters.bigtable.cnrm.cloud.google.com version=v1alpha1: field ".spec.serveNodes" is not set in unstructured objects
+[missing_field] crd=bigtablelogicalviews.bigtable.cnrm.cloud.google.com version=v1alpha1: field ".spec.deletionProtection" is not set in unstructured objects
```
jingyih: Can we set and update this field in the test?
https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/master/pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview

pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview/_http.log
```diff
@@ -301,7 +301,7 @@ GRPC /google.bigtable.admin.v2.BigtableInstanceAdmin/UpdateLogicalView
     "name": "projects/${projectId}/instances/lvdep${uniqueId}/logicalViews/lv-${uniqueId}",
     "query": "SELECT 1 AS value"
   },
-  "updateMask": "query"
+  "updateMask": "query,deletionProtection"
```
jingyih, line 278: I am a bit surprised that `deletionProtection` showed up in the update mask. I don't see the field is being set in the test YAML files.

apis/bigtable/v1alpha1/logicalview_types.go
yuwenma, line 1: Please include the generate.sh file so that we can verify the code.

pkg/controller/direct/bigtable/logicalview_controller.go
georgecma: https://pkg.go.dev/cloud.google.com/go/bigtable#DeletionProtection

The field is actually an int

pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview/_http.log
georgecma: Can you advise which command to use? https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/660fb03b863bed84b6ddf481eca04125fdfedfad/mockgcp/docs/ai/create-mockgcp.md

record-gcp doesn't exist

tests/apichecks/testdata/exceptions/alpha-missingfields.txt
georgecma: What command can I use to update this field?

pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview/_http.log
jingyih: Did you try this script?
hack/record-gcp

Please update the [create.yaml](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/fa5e8d57ebec7ab3876df3fd7f266b7211c1c766/pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview/create.yaml) and [update.yaml](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/fa5e8d57ebec7ab3876df3fd7f266b7211c1c766/pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview/update.yaml) to include the `deletionProtection` field before running the script.

pkg/controller/direct/bigtable/logicalview_controller.go
jingyih: Oh interesting. Then we should probably use `*int` for the `DeletionProtection` field in apis/bigtable/v1alpha1/logicalview_types.go? Currently the field is a boolean.

georgecma: I'm not sure why DeletionProtection in the go bigtable package is int but deletion protection in other types, e.g. authorizedview_types.go, is also Bool (along with our API).

tests/apichecks/testdata/exceptions/alpha-missingfields.txt
georgecma: This has been removed after updating it in the test.

pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview/_http.log
georgecma: Updated, please take a look thanks

pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview/_http.log
```diff
@@ -343,19 +375,26 @@ GRPC /google.bigtable.admin.v2.BigtableInstanceAdmin/UpdateLogicalView
     "name": "projects/${projectId}/instances/lvdep${uniqueId}/logicalViews/lv-${uniqueId}",
     "query": "SELECT 2 AS new_value"
   },
-  "updateMask": "query"
+  "updateMask": "query,deletionProtection"
```
jingyih, line 322: I feel we need to debug this. From what I can tell, we set `deletionProtection` to false in both yaml files, which means we should not expect `deletionProtection` to be included in the update mask.

pkg/controller/direct/bigtable/logicalview_controller.go
```diff
@@ -195,16 +204,24 @@ func (a *LogicalViewAdapter) Update(ctx context.Context, updateOp *directbase.Up
 	if !reflect.DeepEqual(a.desired.Spec.Query, a.actual.Query) {
 		updateMask.Paths = append(updateMask.Paths, "query")
 	}
-	// TODO: Add deletion protection field.
+	if !reflect.DeepEqual(a.desired.Spec.DeletionProtection, a.actual.DeletionProtection) {
```
jingyih: Maybe instead of using `reflect.DeepEqual`, we just directly compare the value of `DeletionProtection` from spec and actual. So that we can more explicitly control if we want to add `DeletionProtection` to the update mask.

pkg/controller/direct/bigtable/logicalview_controller.go
georgecma: Actually, I tried moving to GCP.DeletionProtection and got this error:
{"severity":"info","timestamp":"2025-09-11T16:30:24.054Z","msg":"resource is not ready","kind":"BigtableLogicalView","name":"lv-jxegqbhm52f4eny","conditions":[{"lastTransitionTime":"2025-09-11T16:30:21Z","message":"Update call failed: error converting to *v1alpha1.BigtableLogicalView: unrecognized type: int","reason":"UpdateFailed","status":"False","type":"Ready"}]}

After a bit of digging, my (very superficial) understanding is that converting from Unstructured to LogicalView via DefaultUnstructuredConverter is not trivial when it's a custom GCP type. This may be why authorized Views used a bool in the end. If you have guidence here I'm happy to follow it, otherwise I'll move back a bool

pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview/_http.log
georgecma: Ack, after my latest upload, it looks like creation will set deletionProtection correctly but not update:
{"severity":"error","timestamp":"2025-09-11T16:48:38.016Z","msg":"Reconciler error","controller":"bigtablelogicalview-parent-controller","controllerGroup":"bigtable.cnrm.cloud.google.com","controllerKind":"BigtableLogicalView","BigtableLogicalView":{"name":"lv-t23haglfqtld46q","namespace":"t23haglfqtld46q"},"namespace":"t23haglfqtld46q","name":"lv-t23haglfqtld46q","reconcileID":"d3b43134-5aad-4b9d-bad4-38b223dd0ba5","error":"Delete call failed: deleting LogicalView projects/georgecma-sandbox-664649/instances/lvdept23haglfqtld46q/logicalViews/lv-t23haglfqtld46q: rpc error: code = FailedPrecondition desc = Logical View projects/georgecma-sandbox-664649/instances/lvdept23haglfqtld46q/logicalViews/lv-t23haglfqtld46q has deletion protection enabled."}

Debugging, but it's not apparent to me why this is the case..

pkg/controller/direct/bigtable/logicalview_controller.go
georgecma: Done

pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview/_http.log
georgecma: Wait, is the KCC controller reconciling to logicalView->DeletionProtection=True, and then attempting deletion directly? 
Maybe Update is not used at all?

If so, then this is a WAI, because the later Update action also doesn't include the deletion_protection update mask anymore

georgecma: If so, mockgcp/mockbigtable/testdata/logicalview/crud/script.yaml should be enough to test deletionProtection behavior? How do I assert that it works?

jingyih: Not sure if the behavior is specifically because we are setting the field to "false", which is the default value. Could you try setting the field to a different value and see what happens?

georgecma: If I set the value to deletionProtection: true in create.yaml, running ` hack/record-gcp fixtures/bigtablelogicalview` 

results in a failed test because the logical view cannot be deleted (due to deletionProtection: true)

no _http log is generated in that case

jingyih: but can you update it to "false" in the update.yaml? The test framework only deletes the resource after applying update.yaml.

georgecma: Thanks for the catch, I think the complex logic in updateLogicalView caused a bug where update was not updating deletion protection properly. I fixed it and now the logs look accurate. 

Note that there's no deletion_protection: false in the updateLogicalView call because DeletionProtection is a bool with omit empty in bigtable adminpb. I'll raise an issue with the client team to fix this because it's conflating empty value and false bool: https://stackoverflow.com/questions/37756236/json-golang-boolean-omitempty

// A SQL logical view object that can be referenced in SQL queries.
type LogicalView struct {
	state         protoimpl.MessageState
	sizeCache     protoimpl.SizeCache
	unknownFields protoimpl.UnknownFields

	// Identifier. The unique name of the logical view.
	// Format:
	// `projects/{project}/instances/{instance}/logicalViews/{logical_view}`
	Name string `protobuf:"bytes,1,opt,name=name,proto3" json:"name,omitempty"`
	// Required. The logical view's select query.
	Query string `protobuf:"bytes,2,opt,name=query,proto3" json:"query,omitempty"`
	// Optional. The etag for this logical view.
	// This may be sent on update requests to ensure that the client has an
	// up-to-date value before proceeding. The server returns an ABORTED error on
	// a mismatched etag.
	Etag string `protobuf:"bytes,3,opt,name=etag,proto3" json:"etag,omitempty"`
	// Optional. Set to true to make the LogicalView protected against deletion.
	DeletionProtection bool `protobuf:"varint,6,opt,name=deletion_protection,json=deletionProtection,proto3" json:"deletion_protection,omitempty"`
}

pkg/controller/direct/bigtable/logicalview_controller.go
```diff
@@ -226,10 +215,10 @@ func (a *LogicalViewAdapter) Update(ctx context.Context, updateOp *directbase.Up
 
 		desiredlogicalview := gcp.LogicalViewInfo{
 			LogicalViewID: a.id.ID(),
-			Query:         desiredPb.Query,
+			Query:         *spec.Query,
 		}
 		if slices.Contains(updateMask.Paths, "deletion_protection") {
-			log.V(1).Info("Adding deletion protection bit in the object itself")
+			// Convert bool deletion protection to gcp deletion protection enum.
```
jingyih: Can this conversion logic live inside the mapping functions `BigtableLogicalViewSpec_ToProto` and `BigtableLogicalViewSpec_ToProto`?

pkg/controller/direct/bigtable/logicalview_controller.go
```diff
@@ -202,21 +196,16 @@ func (a *LogicalViewAdapter) Create(ctx context.Context, createOp *directbase.Cr
...
 
-	desiredPb := BigtableLogicalViewSpec_ToProto(mapCtx, &a.desired.DeepCopy().Spec)
-	if mapCtx.Err() != nil {
-		return mapCtx.Err()
-	}
+	spec := a.desired.Spec
 
 	updateMask := &fieldmaskpb.FieldMask{}
-	if !reflect.DeepEqual(a.desired.Spec.Query, a.actual.Query) {
+	if (spec.Query != nil) && (*spec.Query != a.actual.Query) {
```
jingyih, line 192: Same comment about dropping the ` BigtableLogicalViewSpec_ToProto` function. Have you tried something like:

```
if (spec.DeletionProtection != nil) && (desiredPb.DeletionProtection != a.actual.DeletionProtection) {
```

The current code is a bit confusing because technically `spec.DeletionProtection` is a KRM field, whereas `a.actual.DeletionProtection` is a GCP proto field. They could in theory have different type.

pkg/controller/direct/bigtable/logicalview_controller.go
georgecma: Sorry I'm confused - since we're using the go bigtable client method, it takes a LogicalViewInfo object and not pb.LogicalView, I think I'm missing some context as to why we want the pb.LogicalView object. 

I can add a util function to convert deletion_protection fields. Or should I add something like BigtableLogicalViewSpec_ToLogicalViewInfo?

jingyih: Oh I see. I thought we were calling RPCs in "cloud.google.com/go/bigtable/admin/apiv2/adminpb", which works with [LogicalView](https://github.com/googleapis/google-cloud-go/blob/bigtable/v1.39.0/bigtable/admin/apiv2/adminpb/instance.pb.go#L1088). The existing mapping functions are between "KRM go struct" and "LogicView".
Since we are using "cloud.google.com/go/bigtable", we should probably update the mapping functions to use "LogicalViewInfo" instead of "LogicalView". So yes you are right, we want something like `BigtableLogicalViewSpec_ToLogicalViewInfo`.

georgecma: Gotcha, I will make that change.
One more question - so the current LogicalViewSpec_ToLogicalViewProto is used in the fuzzer. I added it there because I assume the fuzzer checks for CreateLogicalView return type (which is a pb.LogicalVIew object).

How does the fuzzer fit into this? Should I support both KRM-to-pb.LogicalView for the fuzzer, and KRM-to-LogicalVIewInfo for the controller?

jingyih: Can you check if we can update the fuzzer to only check for KRM-to-LogicalVIewInfo?

georgecma: I'm not sure if that's possible:" https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/fuzztesting/fuzzkrm.go#L51



I did update the controller to use a _ToLogicalViewInfo proto function, though

jingyih: Oh I see. It is not a proto. Can you try to use this library instead? 
pkg/fuzztesting/fuzzkrm_noproto.go

georgecma: Is there prior art for this? I don't see any fuzzkrm_noproto usage and I'm not familiar with this framework. Otherwise, no worries - I can dedicate some time next week to figuring it out.

pkg/controller/direct/bigtable/logicalview_controller.go
georgecma: Changed

pkg/controller/direct/bigtable/logicalview_controller.go
jingyih: Yes. https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/72ea5e83fb6879e766614be0cd0c925c75f6bab7/pkg/controller/direct/servicenetworking/peereddnsdomain_fuzzer.go#L31

pkg/controller/direct/bigtable/logicalview_controller.go
```diff
@@ -142,19 +149,8 @@ func (a *LogicalViewAdapter) Create(ctx context.Context, createOp *directbase.Cr
 	log := klog.FromContext(ctx)
 	log.V(2).Info("creating LogicalView", "name", a.id)
 	mapCtx := &direct.MapContext{}
+	logicalViewInfo := BigtableLogicalViewSpec_ToLogicalViewInfo(mapCtx, &a.desired.Spec, a.id)
```
maqiuyujoyce, line 153: Could you add the error check for mapCtx.Err()?

Also, do we still need to assign logicalViewInfo.LogicalViewID with a.id.ID()?

pkg/controller/direct/bigtable/logicalview_controller.go
```diff
@@ -54,6 +53,9 @@ type modelLogicalView struct {
 func (m *modelLogicalView) client(ctx context.Context, parentProject string) (*gcp.InstanceAdminClient, error) {
 	var opts []option.ClientOption
 	opts, err := m.config.GRPCClientOptions()
+	if err != nil {
```
maqiuyujoyce, line 57: Good catch!

pkg/controller/direct/bigtable/logicalview_controller.go
maqiuyujoyce: Looks like the API proto of this field uses bool: https://github.com/googleapis/googleapis/blob/master/google/bigtable/admin/v2/instance.proto#L502 but somehow the go client uses the int. I think this is almost a bug in the go client implementation. @georgecma do you happen to know the reason of this discrepancy?

pkg/controller/direct/bigtable/logicalview_controller.go
```diff
@@ -184,31 +180,31 @@ func (a *LogicalViewAdapter) Create(ctx context.Context, createOp *directbase.Cr
...
 
-	desiredPb := BigtableLogicalViewSpec_ToProto(mapCtx, &a.desired.DeepCopy().Spec)
-	if mapCtx.Err() != nil {
-		return mapCtx.Err()
-	}
+	spec := a.desired.Spec
 
 	updateMask := &fieldmaskpb.FieldMask{}
-	if !reflect.DeepEqual(a.desired.Spec.Query, a.actual.Query) {
+	if (spec.Query != nil) && (*spec.Query != a.actual.Query) {
```
maqiuyujoyce, line 192: Hmmm, I think we will unset the field during update if the corresponding KRM field is unset. So if spec.Query is unset, it should mean the update request will unset the query field.

pkg/controller/direct/bigtable/logicalview_controller.go
```diff
@@ -184,31 +180,31 @@ func (a *LogicalViewAdapter) Create(ctx context.Context, createOp *directbase.Cr
...
 		updateMask.Paths = append(updateMask.Paths, "query")
 	}
-	// TODO: Add deletion protection field.
+
+	if (spec.DeletionProtection != nil) && (*spec.DeletionProtection != a.actual.DeletionProtection) {
```
maqiuyujoyce: Similarly, we should unset this field unless it is required.

pkg/controller/direct/bigtable/logicalview_controller.go
```diff
@@ -184,31 +180,31 @@ func (a *LogicalViewAdapter) Create(ctx context.Context, createOp *directbase.Cr
...
 	} else {
 		log.V(2).Info("fields need update", "name", a.id, "paths", updateMask.Paths)
 
-		desiredlogicalview := gcp.LogicalViewInfo{
-			LogicalViewID: a.id.ID(),
-			Query:         desiredPb.Query,
-		}
-		err := a.gcpClient.UpdateLogicalView(ctx, a.id.ParentInstanceIdString(), desiredlogicalview)
+		mapCtx := &direct.MapContext{}
+		desiredlogicalview := BigtableLogicalViewSpec_ToLogicalViewInfo(mapCtx, &a.desired.Spec, a.id)
```
maqiuyujoyce, line 206: You'll want to check mapCtx.Err() after the conversion func.

pkg/controller/direct/bigtable/logicalview_controller.go
```diff
@@ -184,31 +180,31 @@ func (a *LogicalViewAdapter) Create(ctx context.Context, createOp *directbase.Cr
...
+
+		log.V(1).Info("Updating logical view with desired logical view", desiredlogicalview)
+
+		err := a.gcpClient.UpdateLogicalView(ctx, a.id.ParentInstanceIdString(), *desiredlogicalview)
 		if err != nil {
-			return fmt.Errorf("updating LogicalView %s: %w", a.id, err)
+			return fmt.Errorf("updating L`ogicalView %s: %w", a.id, err)
```
maqiuyujoyce: Nit: Remove the backtick.

pkg/controller/direct/bigtable/logicalview_controller.go
```diff
@@ -184,31 +180,31 @@ func (a *LogicalViewAdapter) Create(ctx context.Context, createOp *directbase.Cr
...
 
-		desiredlogicalview := gcp.LogicalViewInfo{
-			LogicalViewID: a.id.ID(),
-			Query:         desiredPb.Query,
-		}
-		err := a.gcpClient.UpdateLogicalView(ctx, a.id.ParentInstanceIdString(), desiredlogicalview)
+		mapCtx := &direct.MapContext{}
+		desiredlogicalview := BigtableLogicalViewSpec_ToLogicalViewInfo(mapCtx, &a.desired.Spec, a.id)
+
+		log.V(1).Info("Updating logical view with desired logical view", desiredlogicalview)
```
maqiuyujoyce: Nit: Use level 2. Level 2-4 are for debugging purposes.

pkg/controller/direct/bigtable/logicalview_mappings.go
```diff
@@ -20,6 +20,7 @@
 package bigtable
 
 import (
+	gcp "cloud.google.com/go/bigtable"
```
maqiuyujoyce, line 23: Nit: Make this lib a separate import group.

pkg/controller/direct/bigtable/logicalview_mappings.go
```diff
@@ -33,20 +34,44 @@ func BigtableLogicalViewSpec_FromProto(mapCtx *direct.MapContext, in *pb.Logical
...
+	if in.DeletionProtection != nil {
+		if *in.DeletionProtection {
+			gcpDeletionProtection = gcp.Protected
+		} else {
+			gcpDeletionProtection = gcp.Unprotected
+		}
+	}
+	out := &gcp.LogicalViewInfo{
+		LogicalViewID: identity.ID(),
+		Query: *in.Query,
```
maqiuyujoyce: A nil check would be needed for Query as it is a pointer.

pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview/dependencies.yaml
```diff
@@ -21,7 +21,4 @@ spec:
   cluster:
   - clusterId: cluster1-${uniqueId}
     zone: us-central1-a
-    numNodes: 3
```
maqiuyujoyce, line 24: Just curious: Why this is removed?

maqiuyujoyce (commented):
Could you make sure there is no `Merge branch` commits in the commit history? This makes the diffs in between commits in this PR not really readable.

Also do you mind filling out the description of this PR with the context?

pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview/dependencies.yaml
georgecma: We don't need a second cluster for Logical View tests. It also makes the _http.log flaky because view creation routing is not deterministic (it could go to either cluster).

maqiuyujoyce: Just curious: is it because the clusters in the list are not sorted at server side?

pkg/controller/direct/bigtable/logicalview_controller.go
georgecma: Resolving since reviewer changed. Feel free to reopen @maqiuyujoyce

pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview/_http.log
georgecma: Resolving since reviewer changed. Feel free to reopen @maqiuyujoyce

tests/apichecks/testdata/exceptions/alpha-missingfields.txt
georgecma: Resolving since reviewer changed. Feel free to reopen @maqiuyujoyce

pkg/controller/direct/bigtable/logicalview_controller.go
georgecma: Probably historical purposes given that there's 3 states for deletion protection? cc @bhshkh as our go client SME

pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview/_http.log
georgecma: Resolving since reviewer changed, feel free to reopen @maqiuyujoyce

pkg/controller/direct/bigtable/logicalview_controller.go
georgecma: This is not set unless deletion protection spec is not nil

pkg/controller/direct/bigtable/logicalview_mappings.go
georgecma: Done

pkg/controller/direct/bigtable/logicalview_controller.go
georgecma: Added error check. a.Id should be present in the LogicalViewInfo object

pkg/controller/direct/bigtable/logicalview_controller.go
georgecma: I think it is auto unset unless we explicitly set it?

maqiuyujoyce: This is doing a PATCH, right? IIUC, PATCH won't touch the field that is not set in the request. If query field was set during creation, then updated to empty/nil, then the query field in the GCP resource will be kept as the value set during creation time even after KCC updates it, right? This is something we want to avoid.

pkg/controller/direct/bigtable/logicalview_controller.go
maqiuyujoyce: Similarly, IIUC, if DeletionProtection is unset in update and set to true during creation, then it probably means user wants to unset this field in the GCP resource during the update (this might mean the field is set back to false, or something else). Then if we don't explicitly mention it in the PATCH request, the PATCH won't touch this field, and it should probably be kept as true in the GCP resource.

pkg/controller/direct/bigtable/logicalview_controller.go
maqiuyujoyce: Error check here.

pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview/dependencies.yaml
maqiuyujoyce: Discussed offline. Filed  #5277 to cover this use case.

maqiuyujoyce (commented):
Again, could you make sure there is no `Merge branch` commits in the commit history?

pkg/controller/direct/bigtable/logicalview_controller.go
georgecma: I'm not sure I understand - so:
spec -> unset -> don't touch the field in update
spec -> set false -> set field to false if field was not false
spec -> set true -> set field to true if field was not true 

This is the code logic

pkg/controller/direct/bigtable/logicalview_mappings.go
georgecma: Question - why gcp as a separate group?

maqiuyujoyce:
Hi @georgecma , thank you so much for cleaning up the commit history. Do you mind splitting the second commit into two commits: The first one with the real GCP log, and the second one with the mock GCP log? This will be used by the reviewer and also by future contributor for posterity purposes.

pkg/controller/direct/bigtable/logicalview_mappings.go
maqiuyujoyce: Oh gcp is no a separate group, I think for this one it should be gcp and pb as a group. Usually we keep the packages from our own repo a separate import group. Overall it's a nit.

pkg/controller/direct/bigtable/logicalview_controller.go
maqiuyujoyce: Thank you for listing the use cases.

> spec -> unset -> don't touch the field in update

This is not how KCC wants to design the resource behavior. When a field is unset, it means the field should work as unset or be set to nil in the GCP resource. So it is not "don't touch the field".

pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview/_http.log
```diff
@@ -216,7 +216,7 @@ GRPC /google.bigtable.admin.v2.BigtableInstanceAdmin/GetLogicalView
   "name": "projects/${projectId}/instances/lvdep${uniqueId}/logicalViews/lv-${uniqueId}"
 }
 
-error: rpc error: code = NotFound desc = Failed to read: projects/{${projectNumber}}/instances/lvdep${uniqueId}/logicalViews/lv-${uniqueId}
+error: rpc error: code = NotFound desc = projects/${projectId}/instances/lvdep${uniqueId}/logicalViews/lv-${uniqueId} not found
```
maqiuyujoyce, line 219: Nit: The error message should match.

pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview/_http.log
```diff
@@ -235,38 +235,15 @@ GRPC /google.bigtable.admin.v2.BigtableInstanceAdmin/CreateLogicalView
...
-
 {
   "done": true,
   "metadata": {
-    "@type": "type.googleapis.com/google.bigtable.admin.v2.CreateLogicalViewMetadata",
-    "endTime": "2024-04-01T12:34:56.123456Z",
-    "originalRequest": {},
-    "startTime": "2024-04-01T12:34:56.123456Z"
+    "@type": "type.googleapis.com/google.bigtable.admin.v2.CreateLogicalViewMetadata"
   },
```
maqiuyujoyce, line 242: It's probably a bit out of scope for this particular feature, but in general we want to ensure the the LRO is also mocked.

pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview/_http.log
```diff
@@ -323,18 +298,11 @@ OK
 {
   "done": true,
   "metadata": {
-    "@type": "type.googleapis.com/google.bigtable.admin.v2.UpdateLogicalViewMetadata",
-    "endTime": "2024-04-01T12:34:56.123456Z",
-    "originalRequest": {
-      "logicalView": {},
-      "updateMask": ""
-    },
-    "startTime": "2024-04-01T12:34:56.123456Z"
```
maqiuyujoyce: Nit: Match update response.

georgecma: This is the file that was generated by 

```
$ hack/compare-mock fixtures/bigtablelogicalview

...
    harness.go:715: controller-runtime manager is shutdown
=== NAME  TestAllInSeries
    unified_test.go:150: shutting down manager
--- PASS: TestAllInSeries (17.63s)
    --- PASS: TestAllInSeries/fixtures (17.63s)
        --- PASS: TestAllInSeries/fixtures/bigtablelogicalview (17.29s)
```
I didn't change anything and it passes

pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview/_http.log
georgecma: Noted, I'll resolve this for now since there's no action items

pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview/_http.log
georgecma: This is the file that was generated by

$ hack/compare-mock fixtures/bigtablelogicalview

...
    harness.go:715: controller-runtime manager is shutdown
=== NAME  TestAllInSeries
    unified_test.go:150: shutting down manager
--- PASS: TestAllInSeries (17.63s)
    --- PASS: TestAllInSeries/fixtures (17.63s)
        --- PASS: TestAllInSeries/fixtures/bigtablelogicalview (17.29s)
I didn't change anything and it passes

pkg/controller/direct/bigtable/logicalview_controller.go
georgecma: @maqiuyujoyce Hi Joyce, thanks for reviewing, re making the update mask declarative - can I change it back to reflect.DeepEqual then or is there some prior art I should follow

pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview/_http.log
georgecma: So I'm not sure I understand, but following: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/5be31460e2cc27221729dacabd078e8317fd58d4/docs/develop-resources/deep-dives/1-add-mockgcp-tests.md

But running these 2 give me different _http logs:
```
10238  [2025-10-09 18:22:13 +0000] hack/record-gcp fixtures/bigtablelogicalview
10239  [2025-10-09 18:26:00 +0000] hack/compare-mockgcp fixtures/bigtablelogicalview
```

pkg/controller/direct/bigtable/logicalview_controller.go
```diff
@@ -187,11 +188,11 @@ func (a *LogicalViewAdapter) Update(ctx context.Context, updateOp *directbase.Up
 	spec := a.desired.Spec
 
 	updateMask := &fieldmaskpb.FieldMask{}
-	if (spec.Query != nil) && (*spec.Query != a.actual.Query) {
+	if !reflect.DeepEqual(spec.Query, a.actual.Query) {
 		updateMask.Paths = append(updateMask.Paths, "query")
 	}
 
-	if (spec.DeletionProtection != nil) && (*spec.DeletionProtection != a.actual.DeletionProtection) {
+	if !reflect.DeepEqual(spec.DeletionProtection, a.actual.DeletionProtection) {
```
maqiuyujoyce, line 196: `spec.DeletionProtection` is a pointer that can be nil while `actual.DeletionProtection` is a bool literal that cannot be nil. You'll want to handle the use case when  `spec.DeletionProtection` is nil here.

pkg/controller/direct/bigtable/logicalview_controller.go
```diff
@@ -187,11 +188,11 @@ func (a *LogicalViewAdapter) Update(ctx context.Context, updateOp *directbase.Up
 	spec := a.desired.Spec
 
 	updateMask := &fieldmaskpb.FieldMask{}
-	if (spec.Query != nil) && (*spec.Query != a.actual.Query) {
+	if !reflect.DeepEqual(spec.Query, a.actual.Query) {
```
maqiuyujoyce: `spec.Query` is a pointer that can be nil while `actual.Query` is a string literal that cannot be nil. You'll want to handle the use case when `spec.Query` is nil here.

pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview/_http.log
maqiuyujoyce: This is not an important discrepancy but is just distractive for current and future code review and mockgcp correctness verification. It is not a blocker for this PR.

It'll need some updates in the mock logicalview.

pkg/test/resourcefixture/testdata/basic/bigtable/v1alpha1/bigtablelogicalview/_http.log
maqiuyujoyce: Not a blocker. The ideal situation is that _http logs are the same running against real and mock.

PR #4858, opened 2025-07-25
[direct] Add spec/metadata label logic to SecretManagerSecret controller

anhdle-sso:
### Change description

Add logic to spec.labels vs metadata.labels logic to SecretManagerSecret direct controller

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


yuwenma:
Could you fix the presubmit failure?

pkg/controller/direct/secretmanager/secret_controller.go
```diff
@@ -196,9 +203,24 @@ func MergeMap(a, b map[string]string) map[string]string {
 	return copy
 }
 
+// Annotation keys must be between 1 and 63 characters long
+// have a UTF-8 encoding of maximum 128 bytes
+// begin and end with an alphanumeric character ([a-z0-9A-Z]),
+// may have dashes (-), underscores (_), dots (.), and alphanumerics in between these symbols.
+func IsValidAnnotation(s string) bool {
```
gemmahou, line 211: nit: can we make this a shared function in case it's needed by other resources? Not a blocker

anhdle-sso: The regex I have here is strictly for this resource annotation field. We can add a function to filter by regex (not matching). I can add it in another PR.

pkg/test/resourcefixture/testdata/basic/secretmanager/v1beta1/secretmanagersecret/fullsecretmanagersecret-manual-direct/create.yaml
```diff
@@ -18,6 +18,9 @@ metadata:
   annotations:
     cnrm.cloud.google.com/project-id: ${projectId}
     alpha.cnrm.cloud.google.com/reconciler: "direct"
+    cnrm.cloud.google.com/remove-this-annotation: "true"
+    alpha.cnrm.cloud.google.com/remove-this-annotation-2: "true"
+    kubectl.kubernetes.io/last-applied-configuration: "{}"
   labels:
     label-one: value-one
```
cheftako: Please add a label here ```configsync.gke.io: "true"```. Then confirm that this label does not get sent to GCP.

cheftako: Also can we please ensure that metadata.labels != spec.labels for testing? As it stands we cannot be sure where the key:value pairs came from.

anhdle-sso: done

pkg/controller/direct/secretmanager/secret_controller.go
```diff
@@ -196,9 +203,24 @@ func MergeMap(a, b map[string]string) map[string]string {
...
+	// A string is valid if it matches the pattern AND its length is between 1 and 63.
+	return validPattern.MatchString(s) && len(s) >= 1 && len(s) <= 63
+}
+
 func ComputeAnnotations(secret *krm.SecretManagerSecret) map[string]string {
 	annotations := MergeMap(secret.GetAnnotations(), secret.Spec.Annotations)
-	common.RemoveByPrefixes(annotations, "cnrm.cloud.google.com", "alpha.cnrm.cloud.google.com")
+	for key := range annotations {
+		if !IsValidAnnotation(key) {
+			delete(annotations, key)
```
cheftako, line 223: Flagged by the LLM. An an absolute *minimum* we should log that the key is being deleted because it does not adhere to the secret manager annotations requirements (https://cloud.google.com/secret-manager/docs/creating-and-managing-annotations#add-annotations-to-a-secret). Even better would be adding a status message but that might be challenging.

anhdle-sso: done

PR #4859, opened 2025-07-25
mappers: support multiversion mappers

justinsb:
- **generators: support multi version and better package imports**
  

- **backupdr: generate multiversion mapper**


justinsb:
/assign @yuwenma 

Gemini was happy to obey my commands, but the mappers in the controller got confused by the split versions.  Introducing "multi-version" should let us proceed (I hope!)

PR #4862, opened 2025-07-26
tests: write list of missing field coverage for alpha resources

justinsb:
We want to have a list so we can work through it.


tests/apichecks/crds_test.go
```diff
@@ -451,6 +451,49 @@ func TestCRDShortNames(t *testing.T) {
...
+	}
+
+	missing := findFieldsNotCoveredByTests(t, shouldVisitCRD)
+
+	want := strings.Join(missing, "\n")
+	test.CompareGoldenFile(t, "testdata/exceptions/missingfields.txt", want)
+}
+
+// Run this test with WRITE_GOLDEN_OUTPUT set to update the exceptions list.
+func TestCRDFieldPresenceInTestsForAlpha(t *testing.T) {
```
yuwenma, line 479: Nit! Was thinking rerunning each alpha kind is not that efficient.

PR #4863, opened 2025-07-28
Add a preliminary version of a Gemini github command.

cheftako:
### Change description

/review:pull looks for possible problem.
Uses a {{args}} setup to pass command line arg.
Only currently meant to be an aid.
Fixes # N/A

```release-note
NONE
```

```docs
NONE
```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/.gemini/commands/review/main.toml
+++ b/.gemini/commands/review/main.toml
@@ -1,0 +1,72 @@
+# In: ~/.gemini/commands/review/main.toml
+# Invoked via: /review:main "Pull/4860"
+# Gemini custom command docs can be found at https://github.com/google-gemini/gemini-cli/blob/main/docs/cli/commands.md#toml-file-format-v1
+
+description = "Flag any review concerns that Gemini can discern."
+prompt = """
+Please analyze the git pull request which can be found here: https://github.com/GoogleCloudPlatform/k8s-config-connector/{{args}}.
+
+## AI Pull Request Review Assistant
+
+This section defines your role when I ask you to review a Pull Request.
+
+### 1. Persona and Goal
+
+You are an expert AI assistant specializing in reviewing Config Connector (kcc or cnrm) pull requests. Your
+primary goal is to accelerate my understanding of the PR and to surface potential issues for my
+review.
+
+**DO NOT** recommend whether to merge the PR. Instead, focus on providing a clear, objective
+analysis and a list of specific concerns.
+
+**DO NOT** attempt to post any comments to the PR on my behalf.
+
+### 2. Information Gathering
+
+Gather the full context of the PR linked above. You must analyze the PR description,
+comments from all contributors, and the code changes (`diff`). Cross-reference the changes with
+surrounding code in the repository to validate assumptions about system behavior.
+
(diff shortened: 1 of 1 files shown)


.gemini/commands/review/main.toml
```diff
@@ -0,0 +1,72 @@
+# In: ~/.gemini/commands/review/main.toml
+# Invoked via: /review:main "Pull/4860"
+# Gemini custom command docs can be found at https://github.com/google-gemini/gemini-cli/blob/main/docs/cli/commands.md#toml-file-format-v1
+
+description = "Flag any review concerns that Gemini can discern."
+prompt = """
+Please analyze the git pull request which can be found here: https://github.com/GoogleCloudPlatform/k8s-config-connector/{{args}}.
```
justinsb, line 7: Does this actually get substituted?  Or is gemini smart enough to figure out that this is a placeholder and just does it...

.gemini/commands/review/main.toml
```diff
@@ -0,0 +1,72 @@
...
+
+## AI Pull Request Review Assistant
+
+This section defines your role when I ask you to review a Pull Request.
+
+### 1. Persona and Goal
+
+You are an expert AI assistant specializing in reviewing Config Connector (kcc or cnrm) pull requests. Your
+primary goal is to accelerate my understanding of the PR and to surface potential issues for my
```
justinsb, line 16: Specifically for KCC, in future I would expect many PRs would fall into one of a set of well-known categories (e.g. add gcloud support to mockgcp, add CRD/mapper/fuzzer, add missing field, add controller, add test case etc).  It might be interesting (in a future tool) to try to identify the type of PR and see whether we have specific criteria for each.  I think for _some_ of them we might have objective criteria and might want to actually post draft comments on the PR (hopefully `/lgtm`!)

.gemini/commands/review/main.toml
cheftako: {{args}} is from https://github.com/google-gemini/gemini-cli/blob/main/docs/cli/commands.md#toml-file-format-v1. There is some sort of odd caching going on. So I know it worked with " {{args}}" but you had to pass the whole url. I think this should work but I need to work out how to clear the cache to test the substitution as part of the string.

.gemini/commands/review/main.toml
cheftako: +1. I started a little bit down this path in the "backward compatibility" section specifying differing requirements for different directories.

.gemini/commands/review/main.toml
```diff
@@ -0,0 +1,72 @@
...
+review.
+
+**DO NOT** recommend whether to merge the PR. Instead, focus on providing a clear, objective
+analysis and a list of specific concerns.
+
+**DO NOT** attempt to post any comments to the PR on my behalf.
+
+### 2. Information Gathering
+
+Gather the full context of the PR linked above. You must analyze the PR description,
```
maqiuyujoyce, line 26: And we can inject mock http log and real http log comparison analysis here later!

cheftako: +1.

Issue #4714, opened 2025-06-17
LoggingLogMetric failing reconciliation of unspecified `metricDescriptor` field


m3adow:
### Checklist

- [x] I did not find a related open issue.
- [x] I did not find a solution in the troubleshooting guide: (https://cloud.google.com/config-connector/docs/troubleshooting)
- [x] If this issue is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Bug Description

Creating a `LoggingLogMetric` with unspecified `metricDescriptor` field results in the CRD permanently staying in the `UpdateFailed` status with the error message `Update call failed: error updating: logMetric update failed: cannot make changes to immutable field: metricDescriptor`. While the CRD never changes to the `UpToDate` status, the log based metric is created correctly (see Additional Diagnostic Information below).

Workaround: Deleting the initially created log based metric in GCP results in a recreation and a proper reconciliation of the `LoggingLogMetric` CRD.

### Additional Diagnostic Information

Details of a log based metric created with a YAML snippet similarly to the one posted below:

```
$ gcloud logging metrics describe --project=<OURCLUSTERPROJECT> logginglogmetric-test
createTime: '2025-06-13T13:09:20.459383319Z'
description: This is a test for the creation of LoggingLogMetric CRDs
filter: |-
  severity="ERROR"
  resource.type="k8s_container"
  resource.labels.cluster_name="<OURCLUSTERPROJECT>"
  resource.labels.namespace_name="logginglogmetric-test"
metricDescriptor:
  description: This is a test for the creation of LoggingLogMetric CRDs
  metricKind: DELTA
  name: projects/<OURCLUSTERPROJECT>/metricDescriptors/logging.googleapis.com/user/till-logginglogmetric-test
  type: logging.googleapis.com/user/till-logginglogmetric-test
  unit: '1'
  valueType: INT64
name: till-logginglogmetric-test
resourceName: projects/<OURCLUSTERPROJECT>/metrics/till-logginglogmetric-test
```

### Kubernetes Cluster Version

v1.33.1-gke.1386000

### Config Connector Version

1.126.0

### Config Connector Mode

cluster mode

### Log Output

```
[...]

PR #4864, opened 2025-07-28
fix: no metricDescriptor

acpana:
Fixes https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/4714

Thoughts with Alex:

* we can probably one shot refactor the logmetric_controller code to look more like the standard direct controllers 
* ditto for the tests

Good follow up would also be turning on the re-reconciliaiton test on LLM.

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/logging/v1beta1/logginglogmetric/nometricdescriptor/_generated_export_nometricdescriptor.golden
+++ b/pkg/test/resourcefixture/testdata/basic/logging/v1beta1/logginglogmetric/nometricdescriptor/_generated_export_nometricdescriptor.golden
@@ -6,12 +6,6 @@ spec:
   description: A simple log metric with no metric descriptor.
   disabled: false
   filter: resource.type=global AND severity=INFO
-  metricDescriptor:
-    displayName: ""
-    launchStage: ""
-    metricKind: DELTA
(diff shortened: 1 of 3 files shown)


pkg/controller/direct/logging/logmetric_controller.go
```diff
@@ -267,11 +267,13 @@ func (a *logMetricAdapter) Update(ctx context.Context, updateOp *directbase.Upda
 			}
 		}
 
-		if !compareMetricDescriptors(a.desired.Spec.MetricDescriptor, a.actual.MetricDescriptor) {
-			if err := validateImmutableFieldsUpdated(a.desired.Spec.MetricDescriptor, a.actual.MetricDescriptor); err != nil {
-				return fmt.Errorf("logMetric update failed: %w", err)
+		if a.desired.Spec.MetricDescriptor != nil {
```
yuwenma, line 270: Does the GCP metricDescriptor have a defaulting value? If so, we want to consider the `a.desired.Spec.MetricDescriptor == nil` case. Otherwise, we are switching back from "unset" to "unmanage" :)

yuwenma: oh, for #4714 another option is to make this field as required (not in the CRD validation, but in create ONLY because it is already Beta)

PR #4879, opened 2025-07-29
feat: dcl api metrics

acpana:
### Change description

Adds support for dcl level metrics too!

```
$ curl http://localhost:8888/prom-metrics | grep "gcp_api_requests_total"
...
# HELP configconnector_gcp_api_requests_total Total number of GCP API requests
# TYPE configconnector_gcp_api_requests_total counter
configconnector_gcp_api_requests_total{controller_name="dataproccluster-controller",method="GET",status_code="4xx"} 3
configconnector_gcp_api_requests_total{controller_name="dataproccluster-controller",method="POST",status_code="4xx"} 1
...
```

#### Does this PR add something which needs to be 'release noted'?

```release-note

```

- [ ] Reviewer reviewed release note.

#### Additional documentation e.g., references, usage docs, etc.:

```docs

```

#### Intended Milestone

Please indicate the intended milestone. 
- [ ] Reviewer tagged PR with the actual milestone.

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #4880, opened 2025-07-30
feat: tf api metrics

acpana:
### Change description

Adds support for tf level metrics too!

```
$ curl http://localhost:8888/prom-metrics | grep "gcp_api_requests_total"

# HELP configconnector_gcp_api_requests_total Total number of GCP API requests
# TYPE configconnector_gcp_api_requests_total counter
...
configconnector_gcp_api_requests_total{controller_name="tagstagkey-controller",method="GET",status_code="2xx"} 4
configconnector_gcp_api_requests_total{controller_name="tagstagkey-controller",method="POST",status_code="2xx"} 3
```

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/transport/config.go
+++ b/third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/transport/config.go
@@ -322,30 +322,6 @@ type Config struct {
 	RequestBatcherIam          *RequestBatcher
 }
 
-func (c *Config) Clone() *Config {
-	clone := *c
-
-	// To prevent subtle bugs, create new copies of any slice fields.
-	// This ensures that appending to a slice in the clone doesn't
-	// accidentally modify the original config's slice.
-	if c.ImpersonateServiceAccountDelegates != nil {
-		clone.ImpersonateServiceAccountDelegates = make([]string, len(c.ImpersonateServiceAccountDelegates))
-		copy(clone.ImpersonateServiceAccountDelegates, c.ImpersonateServiceAccountDelegates)
-	}
-
-	if c.Scopes != nil {
-		clone.Scopes = make([]string, len(c.Scopes))
-		copy(clone.Scopes, c.Scopes)
-	}
-
-	if c.gRPCLoggingOptions != nil {
-		clone.gRPCLoggingOptions = make([]option.ClientOption, len(c.gRPCLoggingOptions))
-		copy(clone.gRPCLoggingOptions, c.gRPCLoggingOptions)
-	}
-
-	return &clone
-}
-
 const AccessApprovalBasePathKey = "AccessApproval"
 const AccessContextManagerBasePathKey = "AccessContextManager"
 const ActiveDirectoryBasePathKey = "ActiveDirectory"


pkg/controller/tf/controller.go
```diff
@@ -247,7 +256,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (res
 	if err := resourceoverrides.Handler.PreActuationTransform(&resource.Resource); err != nil {
 		return reconcile.Result{}, r.HandlePreActuationTransformFailed(ctx, &resource.Resource, fmt.Errorf("error applying pre-actuation transformation to resource '%v': %w", req.NamespacedName.String(), err))
 	}
-	requeue, err := r.sync(ctx, resource)
+
+	meta := r.provider.Meta()
+	if baseConfig, ok := meta.(*tpgconfig.Config); ok {
+		metaC := baseConfig.Clone()
+		metaC.Context = metricstransport.WithControllerName(ctx, r.controllerName) // for metrics transport
+		meta = metaC
```
acpana: in my experience, we needed to clone this Config as the Context that it uses for HTTP requests is the one that we first configure the provider with AND NOT the ctx that we have in hand here.

pkg/metrics/transport/transport.go
```diff
@@ -16,6 +16,7 @@ package transport
 
 import (
 	"context"
+	//"fmt"
```
cheftako: Please delete

pkg/controller/tf/controller.go
```diff
@@ -247,7 +254,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (res
...
+	}
+
+	requeue, err := r.sync(ctx, resource, meta)
```
cheftako, line 265: Seems strange you are passing 1 context directly to sync and a different context to sync through the meta object.

acpana: Yes that's what I wanted to highlight here: 

https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4880#discussion_r2241335580

Note "we" were always doing this before in that the TF Apply command takes in the `ctx` for some operations but for actual HTTP calls and transport calls it uses the Context in the meta

cheftako:
> /hold want to merge #4879 first

4879 is merged

pkg/controller/tf/controller.go
cheftako: Still don't like this. To quote from https://go.dev/blog/context-and-structs "Contexts should not be stored inside a struct type, but instead passed to each function that needs it."

To quote from https://pkg.go.dev/context "Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. This is discussed further in https://go.dev/blog/context-and-structs. The Context should be the first parameter, typically named ctx:"

cheftako: I realize this is not an anti-pattern your putting in, but it is very much an anti-pattern.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/transport/config.go
```diff
@@ -322,6 +322,30 @@ type Config struct {
 	RequestBatcherIam          *RequestBatcher
 }
 
+func (c *Config) Clone() *Config {
```
justinsb: Does this really need to be in the third_party directory?  Could we maybe move it into the controller, and have a function like `CloneAndChangeContext`, and in the comment for that we say what it is doing, how this is sort of a hack, but it's a reasonable thing to do given what we have to work with.

PR #4882, opened 2025-07-30
tests: don't skip the webhooks in our tests

justinsb:
We were skipping the deny-immutable-field-updates webhook,
which meant that errors here are harder to detect.


yuwenma (commented):
/approve
/lgtm 
Awesome!

cheftako:
Justin can you rebase and see if that gets the tests to pass?

PR #4884, opened 2025-07-30
linters: add tests for labels

justinsb:


justinsb:
/hold

We need to fix the lack of test coverage before we merge this one

yuwenma:
> We need to fix the lack of test coverage before we merge this one

I'm okay to add the label coverage in a follow-up PR, and the exception should be removed. 

I like the controller-type exception list! But we have no quick fixes right now:
- the "is not direct but has no TF or DCL labels" is due to missing direct controllers. The `generate-types` works well, but the direct controller has real issues, so it's not implemented yet
- the "is direct but has TF label" is due to beta TF/DCL resource has the direct as opt-in.

justinsb:
> another option is to add it to the kustomization 
We should be committing correct CRDs in the repo (and not committing intermediates): for tests, for customers, for AI

PR #4887, opened 2025-07-30
doc: Add comment to CloudIdentityMembership memberkey field

gemmahou:
### Change description

- Add comment to CloudIdentityMembership memberkey field
- Add this field to outputonlyspecfields.txt

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/apis/cloudidentity/v1beta1/membership_types.go
+++ b/apis/cloudidentity/v1beta1/membership_types.go
@@ -32,6 +32,7 @@ type CloudIdentityMembershipSpec struct {
 	GroupRef GroupRef `json:"groupRef"`
 
 	// Immutable. The `EntityKey` of the member. Either `member_key` or `preferred_member_key` must be set when calling MembershipsService.CreateMembership but not both; both shall be set when returned.
+	// Do not set. This is a legacy OUTPUT-ONLY field.
 	// +kcc:proto:field=google.apps.cloudidentity.groups.v1beta1.Membership.member_key
 	MemberKey *EntityKey `json:"memberKey,omitempty"`
 
--- a/tests/apichecks/testdata/exceptions/missingfields.txt
+++ b/tests/apichecks/testdata/exceptions/missingfields.txt
@@ -310,8 +310,6 @@
 [missing_field] crd=cloudbuildtriggers.cloudbuild.cnrm.cloud.google.com version=v1beta1: field ".spec.webhookConfig.secretRef" is not set; neither 'external' nor 'name' are set
 [missing_field] crd=cloudbuildtriggers.cloudbuild.cnrm.cloud.google.com version=v1beta1: field ".spec.webhookConfig.state" is not set in unstructured objects
 [missing_field] crd=cloudidentitygroups.cloudidentity.cnrm.cloud.google.com version=v1beta1: field ".spec.groupKey.namespace" is not set in unstructured objects
-[missing_field] crd=cloudidentitymemberships.cloudidentity.cnrm.cloud.google.com version=v1beta1: field ".spec.memberKey.id" is not set in unstructured objects
-[missing_field] crd=cloudidentitymemberships.cloudidentity.cnrm.cloud.google.com version=v1beta1: field ".spec.memberKey.namespace" is not set in unstructured objects
 [missing_field] crd=cloudidentitymemberships.cloudidentity.cnrm.cloud.google.com version=v1beta1: field ".spec.preferredMemberKey.namespace" is not set in unstructured objects
 [missing_field] crd=cloudidsendpoints.cloudids.cnrm.cloud.google.com version=v1beta1: field ".spec.description" is not set in unstructured objects
 [missing_field] crd=cloudschedulerjobs.cloudscheduler.cnrm.cloud.google.com version=v1beta1: field ".spec.appEngineHttpTarget.appEngineRouting.instance" is not set in unstructured objects
--- a/tests/apichecks/testdata/exceptions/outputonlyspecfields.txt
+++ b/tests/apichecks/testdata/exceptions/outputonlyspecfields.txt
@@ -1,3 +1,5 @@
+[output_only_spec_field] crd=cloudidentitymemberships.cloudidentity.cnrm.cloud.google.com version=v1beta1: field ".spec.memberKey.id" is not set in unstructured objects
+[output_only_spec_field] crd=cloudidentitymemberships.cloudidentity.cnrm.cloud.google.com version=v1beta1: field ".spec.memberKey.namespace" is not set in unstructured objects
 [output_only_spec_field] crd=cloudidentitymemberships.cloudidentity.cnrm.cloud.google.com version=v1beta1: field ".spec.roles[].restrictionEvaluations.memberRestrictionEvaluation.state" is not set in unstructured objects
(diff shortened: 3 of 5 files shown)


tests/apichecks/testdata/exceptions/missingfields.txt
```diff
@@ -309,8 +309,6 @@
 [missing_field] crd=cloudbuildtriggers.cloudbuild.cnrm.cloud.google.com version=v1beta1: field ".spec.webhookConfig.secretRef" is not set; neither 'external' nor 'name' are set
 [missing_field] crd=cloudbuildtriggers.cloudbuild.cnrm.cloud.google.com version=v1beta1: field ".spec.webhookConfig.state" is not set in unstructured objects
 [missing_field] crd=cloudidentitygroups.cloudidentity.cnrm.cloud.google.com version=v1beta1: field ".spec.groupKey.namespace" is not set in unstructured objects
-[missing_field] crd=cloudidentitymemberships.cloudidentity.cnrm.cloud.google.com version=v1beta1: field ".spec.memberKey.id" is not set in unstructured objects
```
gemmahou, line 313: The test is currently failing as it's blocked by https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4839

apis/cloudidentity/v1beta1/membership_types.go
```diff
@@ -32,6 +32,7 @@ type CloudIdentityMembershipSpec struct {
 	GroupRef GroupRef `json:"groupRef"`
 
 	// Immutable. The `EntityKey` of the member. Either `member_key` or `preferred_member_key` must be set when calling MembershipsService.CreateMembership but not both; both shall be set when returned.
+	// This field is considered output only and should not be set as PreferredMemberKey is required.
```
yuwenma: nit: "is considered output only" (and the whole comment) -->  Do not set. This is a legacy OUTPUT-ONLY field. 

Let's give straightforward, actionable suggestions. Otherwise we should explain what is "output only", what is `PreferredMemberKey`, and what's the relationship between this and `PreferredMemberKey` and the rationale to "consider it as output-only", ...

justinsb: I find this API confusing.  Is memberKey deprecated or output only?  (Or both?)

gemmahou: The API removed `MemberKey` field in v1 but for backward compatibility with legacy DCL controller we are still using v1beta1.

In v1beta1, `either member_key or preferred_member_key must be set when calling, but not both` and `preferred_member_key is required`. That means `member_key` should not be set in the request, and its value will be returned in the response. In v1beta1, this field isn't marked as deprecated or output-only, but it behaves as an output-only field.

apis/cloudidentity/v1beta1/membership_types.go
```diff
@@ -32,6 +32,7 @@ type CloudIdentityMembershipSpec struct {
 	GroupRef GroupRef `json:"groupRef"`
 
 	// Immutable. The `EntityKey` of the member. Either `member_key` or `preferred_member_key` must be set when calling MembershipsService.CreateMembership but not both; both shall be set when returned.
+	// Do not set. This is a legacy OUTPUT-ONLY field.
```
maqiuyujoyce, line 35: Could you clarify the consequence of using this field? Without the clarification, users might be concerned that it will error out if it is set. Though it won't and shouldn't for backwards compatibility purposes.

E.g. `This field and its subfields are output-only and should not be set. Values set under this field will be ignored during the reconciliation.`

apis/cloudidentity/v1beta1/membership_types.go
```diff
@@ -32,6 +32,7 @@ type CloudIdentityMembershipSpec struct {
 	GroupRef GroupRef `json:"groupRef"`
 
 	// Immutable. The `EntityKey` of the member. Either `member_key` or `preferred_member_key` must be set when calling MembershipsService.CreateMembership but not both; both shall be set when returned.
```
maqiuyujoyce, line 34: The comment literally said that this field should be set... isn't this contradicting with the output-only clarification?

tests/apichecks/testdata/exceptions/outputonlyspecfields.txt
```diff
@@ -1,3 +1,5 @@
+[output_only_spec_field] crd=cloudidentitymemberships.cloudidentity.cnrm.cloud.google.com version=v1beta1: field ".spec.memberKey.id" is not set in unstructured objects
```
maqiuyujoyce, line 1: If both the nested fields are output-only. Could you also add the similar comment to the nested fields?

apis/cloudidentity/v1beta1/membership_types.go
gemmahou: It will error out - IIRC the error message looks like `cannot set both member_key and preferred_member_key`. So I think mentioning "Do not set" is fine? Adding to outputonlyspec.txt is for test coverage purposes, as we could skip this field.

I think this is an existing bahavior in TF controller as well as both TF and direct are using v1beta1 API.

apis/cloudidentity/v1beta1/membership_types.go
gemmahou: The API description saying either member_key or preferred_member_key should be set bur not both, and preferred_member_key is also a required field(see its description). This is a "wrong" API design, and they removed member_key field in v1. 

Documented in https://docs.google.com/document/d/1zaR2XNbG9Fw81TSuOpzmtowoyHj0YZfvc06ezgHWXFk/edit?resourcekey=0-h8xZEMPhxDBK73Fc4PB1KQ&tab=t.0#heading=h.yn4apy6jsuor

tests/apichecks/testdata/exceptions/outputonlyspecfields.txt
gemmahou: We cannot comment on the nested field, the nested fields(*EntityKey, contains id and namespace) are also used in required preferred_member_key.

apis/cloudidentity/v1beta1/membership_types.go
maqiuyujoyce: Basically this field description is incorrect and we should make it more clear. E.g.

```
	// Do not set. This is a legacy OUTPUT-ONLY field.
	// +kcc:proto:field=google.apps.cloudidentity.groups.v1beta1.Membership.member_key
	MemberKey *EntityKey `json:"memberKey,omitempty"`

	// Required. Immutable. The `EntityKey` of the member. Must be set when calling MembershipsService.CreateMembership.
	// +kcc:proto:field=google.apps.cloudidentity.groups.v1beta1.Membership.preferred_member_key
	// +required
	PreferredMemberKey *EntityKey `json:"preferredMemberKey,omitempty"`
```

apis/cloudidentity/v1beta1/membership_types.go
maqiuyujoyce: If it'll error out then sure the warning makes sense.

apis/cloudidentity/v1beta1/membership_types.go
maqiuyujoyce: In that case we should modify the field description to match its true behavior.

tests/apichecks/testdata/exceptions/outputonlyspecfields.txt
maqiuyujoyce: Ack.

PR #4890, opened 2025-07-30
chore: remove old install-bundles

justinsb:
These are not used by this repo, and are no longer updated,
so they should not be used.  Delete to avoid problems.


yuwenma:
/lgtm
/approve

happy to see this!

PR #4891, opened 2025-07-30
fix: add missing label to CloudBuildWorkerPool

justinsb:


justinsb:
/assign @yuwenma 

Also note that we can split labels over multiple lines - seems much more readable!

yuwenma (commented):
/lgtm
/approve

Thanks

PR #4893, opened 2025-07-30
fix: resourcewatcher should terminate on delete

justinsb:
It isn't going to become ready from deletion,
and this is causing our tests to hang.


pkg/controller/resourcewatcher/resourcewatcher_test.go
yuwenma, line 1: Can we add a "DELETED" test case in this file ?

justinsb:
Added a deletion test case cc @yuwenma

PR #4904, opened 2025-07-31
feat(testing): add test to check for missing reference docs

yuwenma:
documentation. The test is located in `tests/releases/reference_docs_test.go`.

The test iterates through all supported resources and verifies that a corresponding `.tmpl` file exists for each. Any resource missing its documentation template is recorded in the `tests/releases/testdata/missing_reference.txt` golden file, causing the test to fail if the golden file is not updated.

To support this new testing strategy, the presubmit process has been refactored:
- The `resource-docs` make target, which performs a full documentation generation, has been removed from the `ready-pr` presubmit check.
- A new `release-check` make target has been created to house the `resource-docs` target, intended for use during release preparation.

This change shifts the presubmit validation from a slow doc generation step to a fast unit test. This improves PR turnaround time while still enforcing that all resources are documented before a release.

### Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


Makefile
```diff
@@ -258,7 +258,11 @@ ensure:
 
 # Should run all needed commands before any PR is sent out.
 .PHONY: ready-pr
-ready-pr: lint manifests resource-docs ensure fmt
+ready-pr: lint manifests ensure fmt
+
+# Should run all needed commands to prepare a release.
+.PHONY: release-check
+release-check: resource-docs
```
yuwenma, line 265: We will extend this list to 1. add release note. 2. verify release tags. ...

tests/releases/reference_docs_test.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		t.Fatalf("error getting all GVKs: %v", err)
+	}
+
+	var missing []string
+	for _, gvk := range allGVKs {
+		if strings.HasPrefix(gvk.Version, "v1alpha1") {
+			continue
+		}
+		// IAMServiceAccount is a special case, it is not a real resource.
```
justinsb: Don't tell the IAM team that ;-)

tests/releases/reference_docs_test.go
```diff
@@ -0,0 +1,110 @@
...
+		if gvk.Kind == "IAMServiceAccount" {
+			continue
+		}
+		if hasReferenceDoc(t, gvk) {
+			continue
+		}
+		missing = append(missing, gvk.String())
+	}
+
+	if len(missing) == 0 {
```
justinsb: This looks like a tests/apichecks test.  Can we follow the pattern there?

yuwenma: I was thinking of that at the beginning. But it's for "api" checks, and I want to add more release checks in the new dir.

Makefile
justinsb: You can also add it to https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/dev/ci/presubmits/validate-generated-files

justinsb: Actually I guess not, because we only want to run this pre-release!

It might be easier to just auto-generate a basic resource-doc as part of promotion, but .. not a blocker.

tests/releases/reference_docs_test.go
yuwenma: done.

justinsb (approved):

justinsb (approved):

tests/releases/reference_docs_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	var missing []string
+	for _, crd := range crds {
+		for _, version := range crd.Spec.Versions {
+			gvk := schema.GroupVersionKind{
+				Group:   crd.Spec.Group,
+				Version: version.Name,
+				Kind:    crd.Spec.Names.Kind,
+			}
+			if gvk.Group == "core.cnrm.cloud.google.com" {
```
cheftako, line 44: Should this just be a suffix check for "core.cnrm.cloud.google.com"? Seems like any <X>.core.cnrm.cloud.google.com should be a continue.

tests/releases/reference_docs_test.go
```diff
@@ -0,0 +1,89 @@
...
+				continue
+			}
+			if gvk.Group == "customize.core.cnrm.cloud.google.com" {
+				continue
+			}
+			if strings.HasPrefix(gvk.Version, "v1alpha1") {
```
cheftako, line 50: OK for now. I think if we want customer's to use alpha resources provisionally we need to document them but with some sort of warning.

tests/releases/testdata/missing_reference.txt
```diff
@@ -0,0 +1,7 @@
+bigquery.cnrm.cloud.google.com/v1beta1, Kind=BigQueryRoutine
```
cheftako, line 1: Is the goal for this empty to be empty? Do we want to try to hold a line of not adding to this file? If so it would be good to have a comment which explained that.

tests/releases/reference_docs_test.go
yuwenma: Sg

tests/releases/testdata/missing_reference.txt
yuwenma: We want to add a check to have a clean "missing_reference" before each release. It would be a presubmit check for release tag PRs (or some other auto-check)

PR #4910, opened 2025-08-01
mockgcp: add dev/tasks

justinsb:
I think these will map to MCP tasks, so let's normalize them.

Also add a task to record GCP output from a gcloud test.


mockgcp/dev/tasks/record-gcp
yuwenma, line 1: nit: suggest renaming to `record-gcp.py` so it's more natural to run cmd `python3`. Maybe better to put it in a subdir so we can have more python scripts.

PR #4913, opened 2025-08-01
tests: verify that CRDs round-trip

justinsb:


justinsb:
/assign @cheftako 

Note that this is really a TODO list of problems!

yuwenma (commented):
But we already have `TestMultiVersionCRDNoDiff` for the same purpose. Mind sharing how this is different?

PR #4917, opened 2025-08-01
Fix: Correctly determine sample API group for test filtering

yuwenma:
The logic for SKIP_TEST_APIGROUP and ONLY_TEST_APIGROUPS was not working reliably because the API group for a sample was
  determined by the first resource in the sample file, which could be a dependency.

  This change fixes the issue by:
   - Introducing an APIGroup field to the Sample and SampleKey structs.
   - Updating ListMatchingSamples to parse the sample files and identify the main resource by matching the resource kind with the sample's directory name.
   - Extracting the apiGroup from the apiVersion of the main resource.
   - Using the new APIGroup field in the E2E test filtering logic.

  This ensures that the test filtering based on API groups works correctly.

### Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/config/tests/samples/create/samples.go
+++ b/config/tests/samples/create/samples.go
@@ -49,6 +49,7 @@ const DefaultWaitForReadyTimeout = 35 * time.Minute
 
 type Sample struct {
 	Name      string
+	APIGroup  string
 	Resources []*unstructured.Unstructured
 }
 
@@ -358,6 +359,7 @@ func LoadSample(t *testing.T, sampleKey SampleKey, project testgcp.GCPProject) S
 type SampleKey struct {
 	Name      string
 	SourceDir string
+	APIGroup  string
 	files     []string
 }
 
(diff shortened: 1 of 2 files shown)


yuwenma:
The flaky tests "regular-cluster" and "secondary-cluster" are tested in "tests-e2e-samples-alloydb" instead of unclassified shows that this code fixes the problem.

config/tests/samples/create/samples.go
```diff
@@ -358,6 +359,7 @@ func LoadSample(t *testing.T, sampleKey SampleKey, project testgcp.GCPProject) S
 type SampleKey struct {
 	Name      string
 	SourceDir string
+	APIGroup  string
```
justinsb, line 362: Nit: Would be good to add a comment about what this is, and maybe just export the GVK (because 1. it is strongly-typed and 2. you have it anyway).  But not a blocker.

justinsb (approved):

PR #4921, opened 2025-08-01
WIP: ServiceNetworkingPeeredDNSDomain: controller

justinsb:
- **ServiceNetworkingPeeredDnsDomain: Add CRD/Mapper/Fuzzer**
  

- **autogen: apis/servicenetworking/v1alpha1/generate.sh**
  

- **tests: add fuzzer for types without protos**
  

- **ServiceNetworkingPeeredDNSDomain: create controller**
  

- **autogen: ServiceNetworkingPeeredDNSDomain controller**
  

- **ServiceNetworkingPeeredDNSDomain: simple test**
  

- **servicenetworking: required project number**
  Fix the mock and introduce the shared project mapper.
  

- **ServiceNetworkingPeeredDnsDomain: dnsSuffix must end with dot**
  

- **tests: golden output for ServiceNetworkingPeeredDnsDomain**


justinsb:
Issue #5037

pkg/controller/direct/servicenetworking/peereddnsdomain_controller.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+	status.ExternalRef = direct.PtrTo(a.id.String())
+	return createOp.UpdateStatus(ctx, status, nil)
+}
+
+func (a *dataStoreAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOperation) error {
+	log := klog.FromContext(ctx)
+	log.V(2).Info("updating servicenetworking peeredDnsDomain", "name", a.id)
+
+	desired := ReflectClone(a.desired)
```
yuwenma, line 211: I don't see the desired being used. in this function. Is it intended?

pkg/controller/direct/servicenetworking/utils.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"encoding/json"
+
+	"k8s.io/klog/v2"
+)
+
+// ReflectClone creates a deep copy of the given object.
+// It is implemented by serializing and unserializing the object to/from json.
+func ReflectClone[T any](obj T) T {
```
yuwenma, line 25: One general question about ReflectClone: how can we distinguish field unset vs default? I thought we use krm object mostly for that purpose.

apis/refs/v1beta1/computerefs.go
```diff
@@ -57,6 +57,17 @@ func (c *ComputeNetworkID) String() string {
...
 
+func (p *ComputeNetworkID) FromExternal(ref string) error {
+	id, err := ParseComputeNetworkID(ref)
+	if err != nil {
+		return fmt.Errorf("error parsing ComputeNetworkID from %q: %w", ref, err)
+	}
+	p.Project = id.Project
+	p.Network = id.Network
+	return nil
+}
```
yuwenma, line 68: Could you elaborate a little bit more about this function? My question is mostly that why making this a struct/class function, rather than pass in a string, and return a ComputeNetworkID. (like what ParseComputeNetworkID does). Because the ID override behavior will introduce mis-use in caller functions and make the code hard to maintain.

apis/refs/v1beta1/computerefs.go
```diff
@@ -57,6 +57,17 @@ func (c *ComputeNetworkID) String() string {
...
+
+// Deprecated: use ComputeNetworkID.FromExternal instead.
```
yuwenma, line 70: Can I ask what problem you hit? Our plan is to move this dir under apis/compute/v1beta1 and not deprecate `ParseComputeNetworkID`.

pkg/controller/direct/servicenetworking/utils.go
justinsb: Looks like we need a design doc for cookies.  Do you want to write it, or should I?

apis/refs/v1beta1/computerefs.go
justinsb: We need to implement identity.Identity

apis/refs/v1beta1/computerefs.go
justinsb: It duplicates functionality in identity.Identity

pkg/controller/direct/servicenetworking/peereddnsdomain_controller.go
justinsb: Ah good call.  This is a weird function because the service doesn't support update.  I'll clean it up!

apis/refs/v1beta1/computerefs.go
yuwenma: ParseComputeNetworkID covers more corner cases especially the `fixStaleExternalFormat` which comes from a real customer report. If we prefer the Identity code style, could you apply the ParseComputeNetworkID fixes?

justinsb: ComputeNetworkID::FromExternal calls ParseComputeNetworkID, so it includes those fixes.  We just want to encourage people to use identity and references homogeneously (through interfaces), so we deprecate the "special case method".  But we will then move the logic into ComputeNetworkID::FromExternal

PR #4923, opened 2025-08-02
feat: turn on direct ctrl

acpana:
### Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/controller/direct/iam/partialpolicy_controller.go
+++ b/pkg/controller/direct/iam/partialpolicy_controller.go
@@ -37,6 +37,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
 	kcciamclient "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/iam/iamclient"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/predicate"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
 )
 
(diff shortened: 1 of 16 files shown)


acpana:
I think we want to plug in the flow from here -- https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4930

pkg/controller/direct/iam/partialpolicy_controller.go
```diff
@@ -44,7 +45,42 @@ const (
...
+
+var _ predicate.ReconcileGate = &IAMPartialPolicyReconcileGate{}
+
+func (r *IAMPartialPolicyReconcileGate) ShouldReconcile(o *unstructured.Unstructured) bool {
+	// Check for alpha reconciler annotation to opt out
+	if v := o.GetAnnotations()["alpha.cnrm.cloud.google.com/reconciler"]; v == "legacy" {
+		return false
+	}
+
+	// todo acpana integrate with experiments controls for CCC/ CC
```
maqiuyujoyce, line 74: Thank you for clarifying this. If this integration hasn't be supported, then I suggest we don't merge #4930, as the feature won't work.

maqiuyujoyce (commented):
Just curious, can users opt-in using direct controller for IAMPartialPolicy right now?

pkg/controller/direct/iam/partialpolicy_controller.go
acpana: > Just curious, can users opt-in using direct controller for IAMPartialPolicy right now?

Yes it's on by default with this change

pkg/config/controllerconfig.go
```diff
@@ -52,6 +52,9 @@ type ControllerConfig struct {
 
 	// EnableMetricsTransport enables automatic wrapping of HTTP clients with metrics transport
 	EnableMetricsTransport bool
+
+	// Use the legacy/ handwritten IAM controllers.
+	UseLegacyIAM bool
```
yuwenma, line 57: The default value is false which means we use the new IAM controller directly. Another option is to use `UseK8sIAM` (like opt-in), , the legacy would be the default and then we can flip the bit to use k8s in the next 1-2 versions.

acpana: I think per my chatter with @cheftako we might want to turn this on by default and not require practitioners/ CXs to opt in.

PR #4486, opened 2025-05-06
operator: add support for experiments

justinsb:
- **WIP: update kdp**
  

- **Simple golden tests**
  

- **operator: add experiments field to ConfigConnector CRD**
  

- **operator: implement first experiment**
  

- **test for no-monitoring-service experiment**
  

- **tests: add tests for CCC operator**
  

- **chore: refactor operator constants into operator API**

PR #4930, opened 2025-08-04
feat: operator experiments spec field CC/CCC

acpana:
Takes over https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4486 to add extra args to the kccmgr based on the CC/CCC experiments list.


operator/pkg/controllers/configconnector/experiments.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+}
+
+func (r *Reconciler) applyExperiments(ctx context.Context, cc *corev1beta1.ConfigConnector, m *manifest.Objects) error {
+	log := log.FromContext(ctx)
+
+	for _, experiment := range cc.Spec.Experiments {
+		key := strings.ToLower(experiment)
+		switch key {
+		case "legacy-iam-reconciler":
```
maqiuyujoyce: Nit: This experiment name is mentioned at multiple places, can we use a constant?

operator/pkg/controllers/configconnector/experiments.go
```diff
@@ -0,0 +1,87 @@
...
+		default:
+			log.Info("ignoring unknown experiment", "key", key)
+			// TODO: add to status?
+		}
+	}
+
+	return nil
+}
+
+func (r *Reconciler) applyExperimentLegacyIAMReconciler(ctx context.Context, cc *corev1beta1.ConfigConnector, m *manifest.Objects) error {
```
maqiuyujoyce: cc Doesn't seem to be a used parameter here, can we remove it?

operator/pkg/controllers/configconnectorcontext/experiments.go
```diff
@@ -0,0 +1,96 @@
...
+		default:
+			log.Info("ignoring unknown experiment", "key", key)
+			// TODO: add to status?
+		}
+	}
+
+	return nil
+}
+
+func (r *Reconciler) applyExperimentLegacyIAMReconciler(ctx context.Context, cc *corev1beta1.ConfigConnector, ccc *corev1beta1.ConfigConnectorContext, m *manifest.Objects) error {
```
maqiuyujoyce: Neither cc nor ccc is used in the function, do we still need them in the function signature?

maqiuyujoyce (commented):
Have you tested it locally to make sure the experiment flag works E2E?

acpana:
> Have you tested it locally to make sure the experiment flag works E2E?

yes we also have a smoke test in the golden files; I'm going to work that flag into the mgr in an upcoming PR that also glues in the direct ctrl turning on

operator/pkg/controllers/configconnector/testdata/golden/legacy-iam-reconciler/input.yaml
jingyih, line 1: nit: maybe prefix `experiment-` to the folder.
`operator/pkg/controllers/configconnector/testdata/golden/experiment-legacy-iam-reconciler/input.yaml`

operator/pkg/apis/core/v1beta1/configconnector_types.go
```diff
@@ -66,6 +66,12 @@ type ConfigConnectorSpec struct {
 	//+kubebuilder:validation:Enum=Absent;Merge
 	//+kubebuilder:validation:Optional
 	StateIntoSpec *StateIntoSpecValue `json:"stateIntoSpec,omitempty"`
+
+	// Experiments are a list of strings that allow opting-in to previews/experiments.
+	// This functionality is generally not supported and not expected to work going forwards,
+	// the intention is to allow new functionality to be delivered to non-production environments quickly,
+	// proving the functionality before it is baked into the supported API.
+	Experiments []string `json:"experiments,omitempty"`
```
jingyih, line 74: Is it possible to validate the experiments using Kubebuilder validation (`+kubebuilder:validation`) or CEL (`+kubebuilder:validation:XValidation`)?

operator/pkg/controllers/configconnectorcontext/experiments.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (r *Reconciler) applyExperimentLegacyIAMReconciler(ctx context.Context, m *manifest.Objects) error {
+	log := log.FromContext(ctx)
+	for _, obj := range m.Items {
+		if obj.Kind != "StatefulSet" || !strings.HasPrefix(obj.GetName(), "cnrm-controller-manager") {
```
jingyih, line 77: Can we document the behavior of the experiment you are adding? I am not sure I fully understand the logic here. Do we want to be able to scope the experiment to namespace scope? Or is the experiment always cluster scoped?

jingyih: Oh actually I might have misread the code earlier. I think the idea is that if KCC is running in cluster mode, use CC's spec.experiments field; If KCC is running in namespaced mode, use CCC's spec.experiments field?

maqiuyujoyce: I think we can start with having experiments as the cluster-level configuration, and make this field a status field in the CCC. If we get feedback then we can make it overridable at namespace level.

Though I feel it's up to our definition of "experiments". If it is a baking stage of a new feature, then yes, we can let users gradually turn certain experiment on across namespaces to do a safer "rollout". If an experiment is something that may not be landed as an official feature, and may even have unexpected consequences, I don't know if I want non-admin roles to have the flexibility to turn on an experiment.

operator/pkg/apis/core/v1beta1/configconnector_types.go
acpana: sure what kind of validation would you like to see? we can also add it on as a follow up 😛 

I think as we are prototypying this we can keep the validations a bit loose at first

jingyih: I was thinking we could validate the experiment names early in CRD (as opposed to in controller). But agreed we could do a follow up.

operator/pkg/controllers/configconnector/experiments.go
```diff
@@ -0,0 +1,88 @@
...
+	"sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/declarative/pkg/manifest"
+
+	corev1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/apis/core/v1beta1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/experiments"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func (r *Reconciler) transformForExperiments() declarative.ObjectTransform {
+	return func(ctx context.Context, o declarative.DeclarativeObject, m *manifest.Objects) error {
+		cc, ok := o.(*corev1beta1.ConfigConnector)
```
maqiuyujoyce, line 33: If you want to support the experiments field without supporting the flag yet in #4923, I suggest you just return an error here to make the "feature not yet available" explicit.

acpana:
We're changing direction to have the IAM cutover be opt in and not defaulted.

PR #4935, opened 2025-08-05
codebot: update go to 1.24.5

justinsb:


scripts/environment-setup/golang-setup.sh
```diff
@@ -17,7 +17,7 @@
 set -o errexit
 
 # Automatically updated by dev/codebots/update-golang-version
-GO_VERSION="1.23.4"
```
cheftako, line 20: A little worried this was stale.

PR #4936, opened 2025-08-05
feat: turn on direct controller for bigquerydataset

xiaoweim:
### Change description

Test Coverage:

* CRD Field coverage: 100%
* KCC scenario coverage:
    * create, no change, update, delete in fixture tests
    * acquisition (test log in #4829)
          * Tested acquire using basic yaml with pre-created dataset that has data stored within, the data did not get overriden.
    * Drift detection: verified manually
    * TF to direct migration: 
        * verified manually (created a bigquerydataset resource managed by TF, then update the resource to use direct controller). 
        * commit 8926a197932f6fcde2cceb8ef16de4f4c4d1998c shows the diff between TF-based controller and direct controller running against real GCP
* Controller logic branch coverage: 78%

Fixes #

```release-note
Added support for direct reconciliation for BigqueryDataset resource.
```

```docs

```

#### Intended Milestone

1.135.0

### Tests you have done

### DIFF
--- a/pkg/controller/direct/bigquerydataset/bigquerydataset_controller.go
+++ b/pkg/controller/direct/bigquerydataset/bigquerydataset_controller.go
@@ -260,15 +260,18 @@ func (a *Adapter) Update(ctx context.Context, updateOp *directbase.UpdateOperati
 		resource.Access = desired.Access
 		updateMask.Paths = append(updateMask.Paths, "access")
 	}
+	if desired.Labels != nil && !reflect.DeepEqual(desired.Labels, resource.Labels) {
+		// We always send the full map of labels to GCP, so we do not need to specify each label key in the update mask.
+		resource.Labels = desired.Labels
+		updateMask.Paths = append(updateMask.Paths, "labels")
+	}
 	if len(updateMask.Paths) == 0 {
 		return nil
 	}
 
 	// Compute the dataset metadate for update request
 	datasetMetadataToUpdate := BigQueryDataset_ToMetadataToUpdate(mapCtx, resource, updateMask.Paths)
-	for k, v := range label.NewGCPLabelsFromK8sLabels(a.desired.GetObjectMeta().GetLabels()) {
-		datasetMetadataToUpdate.SetLabel(k, v)
-	}
+
 	// Call update
 	dsHandler := a.gcpService.DatasetInProject(a.id.Parent().ProjectID, a.id.ID())
 	updated, err := dsHandler.Update(ctx, *datasetMetadataToUpdate, "")
--- a/pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/basicbigquerydataset-direct/_http.log
+++ b/pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/basicbigquerydataset-direct/_http.log
@@ -152,11 +152,7 @@ Content-Type: application/json
 User-Agent: kcc/${kccVersion} (+https://github.com/GoogleCloudPlatform/k8s-config-connector) kcc/controller-manager/${kccVersion}
 
 {
-  "friendlyName": "bigquerydataset-sample-updated",
-  "labels": {
-    "cnrm-test": "true",
-    "managed-by-cnrm": "true"
-  }
+  "friendlyName": "bigquerydataset-sample-updated"
 }
 
 200 OK
--- a/pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/bigquerydatasetaccessblock-direct/_generated_export_bigquerydatasetaccessblock-direct.golden
+++ b/pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/bigquerydatasetaccessblock-direct/_generated_export_bigquerydatasetaccessblock-direct.golden
@@ -9,14 +9,12 @@ metadata:
   name: bigquerydataset${uniqueId}
 spec:
   access:
+  - domain: google.com
+    role: READER
+  - iamMember: allUsers
+    role: READER
   - role: OWNER
     specialGroup: projectOwners
-  - role: OWNER
-    userByEmail: user@google.com
-  - role: READER
-    specialGroup: projectReaders
-  - role: WRITER
-    specialGroup: projectWriters
   defaultTableExpirationMs: 7200000
   description: BigQuery Dataset With Access Block v2 Updated
   friendlyName: bigquerydataset-accessblock
--- a/pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/bigquerydatasetaccessfull-direct/_http.log
+++ b/pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/bigquerydatasetaccessfull-direct/_http.log
@@ -963,11 +963,7 @@ User-Agent: kcc/${kccVersion} (+https://github.com/GoogleCloudPlatform/k8s-confi
         "tableId": "bigquerytablesample${uniqueId}"
       }
     }
-  ],
-  "labels": {
-    "cnrm-test": "true",
-    "managed-by-cnrm": "true"
-  }
+  ]
 }
 
 200 OK
@@ -1164,11 +1160,7 @@ User-Agent: kcc/${kccVersion} (+https://github.com/GoogleCloudPlatform/k8s-confi
       }
     }
   ],
-  "description": "BigQuery Dataset With Full Access Block Updated",
-  "labels": {
-    "cnrm-test": "true",
-    "managed-by-cnrm": "true"
-  }
+  "description": "BigQuery Dataset With Full Access Block Updated"
 }
 
 200 OK
--- a/pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/fullybigquerydataset-direct/_http.log
+++ b/pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/fullybigquerydataset-direct/_http.log
@@ -565,10 +565,6 @@ User-Agent: kcc/${kccVersion} (+https://github.com/GoogleCloudPlatform/k8s-confi
   "description": "Fully Configured BigQuery Dataset updated",
   "friendlyName": "bigquerydataset-fullyconfigured-updated",
   "isCaseInsensitive": false,
-  "labels": {
-    "cnrm-test": "true",
-    "managed-by-cnrm": "true"
-  },
   "maxTimeTravelHours": "96"
 }
 
--- a/pkg/test/resourcefixture/testdata/stateabsentinspec/bigquerydataset/_http.log
+++ b/pkg/test/resourcefixture/testdata/stateabsentinspec/bigquerydataset/_http.log
@@ -153,11 +153,7 @@ Content-Type: application/json
 User-Agent: kcc/${kccVersion} (+https://github.com/GoogleCloudPlatform/k8s-config-connector) kcc/controller-manager/${kccVersion}
 
 {
-  "friendlyName": "bigquerydataset-sample-updated",
-  "labels": {
-    "cnrm-test": "true",
-    "managed-by-cnrm": "true"
-  }
+  "friendlyName": "bigquerydataset-sample-updated"
 }
 
 200 OK
--- a/mockgcp/mockbigquery/datasets.go
+++ b/mockgcp/mockbigquery/datasets.go
@@ -256,30 +256,16 @@ func (s *datasetsServer) PatchDataset(ctx context.Context, req *pb.PatchDatasetR
 	if updated.StorageBillingModel == nil {
 		updated.StorageBillingModel = existing.StorageBillingModel
 	}
+	if updated.Labels == nil {
+		updated.Labels = existing.Labels
+	}
 	if updated.Type == nil {
 		updated.Type = existing.Type
 	}
 	updated.SelfLink = PtrTo("https://bigquery.googleapis.com/bigquery/v2/" + name.String())
 
 	if len(updated.Access) == 0 {
-		updated.Access = []*pb.DatasetAccess{
-			{
-				Role:         PtrTo("WRITER"),
-				SpecialGroup: PtrTo("projectWriters"),
-			},
-			{
-				Role:         PtrTo("OWNER"),
-				SpecialGroup: PtrTo("projectOwners"),
-			},
-			{
-				Role:        PtrTo("OWNER"),
-				UserByEmail: PtrTo("me@example.com"),
-			},
-			{
-				Role:         PtrTo("READER"),
-				SpecialGroup: PtrTo("projectReaders"),
-			},
-		}
+		updated.Access = existing.Access
 	}
 
 	sortAccess(updated)
--- a/pkg/controller/direct/bigquerydataset/bigquerydataset_mappings.go
+++ b/pkg/controller/direct/bigquerydataset/bigquerydataset_mappings.go
@@ -224,6 +224,11 @@ func BigQueryDataset_ToMetadataToUpdate(mapCtx *direct.MapContext, in *pb.Datase
 	if slices.Contains(updatePaths, "storage_billing_model") {
 		out.StorageBillingModel = in.StorageBillingModel
 	}
+	if slices.Contains(updatePaths, "labels") {
+		for k, v := range in.Labels {
+			out.SetLabel(k, v)
+		}
+	}
 	return out
 }
 func DatasetAccessEntry_FromProto(mapCtx *direct.MapContext, in *pb.DatasetAccessEntry) *krm.DatasetAccessEntry {
--- a/pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/bigquerydatasetaccessblock-direct/_http.log
+++ b/pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/bigquerydatasetaccessblock-direct/_http.log
@@ -166,11 +166,7 @@ User-Agent: kcc/${kccVersion} (+https://github.com/GoogleCloudPlatform/k8s-confi
 
 {
   "defaultTableExpirationMs": "7200000",
-  "description": "BigQuery Dataset With Access Block v2 Updated",
-  "labels": {
-    "cnrm-test": "true",
-    "managed-by-cnrm": "true"
-  }
+  "description": "BigQuery Dataset With Access Block v2 Updated"
 }
 
 200 OK
@@ -186,20 +182,16 @@ X-Xss-Protection: 0
 {
   "access": [
     {
+      "domain": "google.com",
+      "role": "READER"
+    },
+    {
+      "iamMember": "allUsers",
+      "role": "READER"
+    },
+    {
       "role": "OWNER",
       "specialGroup": "projectOwners"
-    },
-    {
-      "role": "OWNER",
-      "userByEmail": "user@google.com"
-    },
-    {
-      "role": "READER",
-      "specialGroup": "projectReaders"
-    },
-    {
-      "role": "WRITER",
-      "specialGroup": "projectWriters"
     }
   ],
   "creationTime": "123456789",
@@ -242,20 +234,16 @@ X-Xss-Protection: 0
 {
   "access": [
     {
+      "domain": "google.com",
+      "role": "READER"
+    },
+    {
+      "iamMember": "allUsers",
+      "role": "READER"
+    },
+    {
       "role": "OWNER",
       "specialGroup": "projectOwners"
-    },
-    {
-      "role": "OWNER",
-      "userByEmail": "user@google.com"
-    },
-    {
-      "role": "READER",
-      "specialGroup": "projectReaders"
-    },
-    {
-      "role": "WRITER",
-      "specialGroup": "projectWriters"
     }
   ],
   "creationTime": "123456789",
@@ -298,20 +286,16 @@ X-Xss-Protection: 0
 {
   "access": [
     {
+      "domain": "google.com",
+      "role": "READER"
+    },
+    {
+      "iamMember": "allUsers",
+      "role": "READER"
+    },
+    {
       "role": "OWNER",
       "specialGroup": "projectOwners"
-    },
-    {
-      "role": "OWNER",
-      "userByEmail": "user@google.com"
-    },
-    {
-      "role": "READER",
-      "specialGroup": "projectReaders"
-    },
-    {
-      "role": "WRITER",
-      "specialGroup": "projectWriters"
     }
   ],
   "creationTime": "123456789",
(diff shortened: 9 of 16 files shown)


pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/fullybigquerydataset-direct/_generated_export_fullybigquerydataset-direct.golden
```diff
@@ -23,7 +23,6 @@ spec:
   defaultTableExpirationMs: 3800000
   description: Fully Configured BigQuery Dataset updated
   friendlyName: bigquerydataset-fullyconfigured-updated
-  isCaseInsensitive: true
```
maqiuyujoyce: Do you happen to know why this diff happens?

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/basicbigquerydataset-direct/_http.log
```diff
@@ -89,16 +140,28 @@ X-Xss-Protection: 0
...
   "location": "US",
+  "maxTimeTravelHours": "168",
   "selfLink": "https://bigquery.googleapis.com/bigquery/v2/projects/${projectId}/datasets/bigquerydatasetsample${uniqueId}",
   "type": "DEFAULT"
 }
 
 ---
 
-GET https://bigquery.googleapis.com/bigquery/v2/projects/${projectId}/datasets/${datasetID}?alt=json
+PATCH https://bigquery.googleapis.com/bigquery/v2/projects/${projectId}/datasets/${datasetID}?alt=json&prettyPrint=false
```
maqiuyujoyce: Looks like we turn one PUT into more than one PATCH requests and at least additional GET request. Have we considered the potential scalability restriction (e.g. quota limit) for this change?

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/basicbigquerydataset-direct/_http.log
```diff
@@ -38,13 +37,65 @@ User-Agent: kcc/${kccVersion} (+https://github.com/GoogleCloudPlatform/k8s-confi
...
+  },
+  "etag": "abcdef0123A=",
+  "friendlyName": "bigquerydataset-sample",
+  "id": "000000000000000000000",
+  "kind": "bigquery#dataset",
   "labels": {
     "cnrm-test": "true",
     "managed-by-cnrm": "true"
   },
-  "location": "US"
```
maqiuyujoyce: Looks like we no longer default this value?

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/bigquerydatasetaccessfull-direct/_http.log
```diff
@@ -93,10 +92,73 @@ X-Xss-Protection: 0
...
+  "lastModifiedTime": "123456789",
+  "location": "us-central1",
+  "maxTimeTravelHours": "168",
+  "selfLink": "https://bigquery.googleapis.com/bigquery/v2/projects/${projectId}/datasets/bigquerydatasetsample${uniqueId}",
+  "type": "DEFAULT"
+}
+
+---
+
+PATCH https://bigquery.googleapis.com/bigquery/v2/projects/${projectId}/datasets/${datasetID}?alt=json&prettyPrint=false
```
maqiuyujoyce: Could you help me understand why does it need an additional PATCH here before the next POST while the old log doesn't?

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/fullybigquerydataset-direct/_http.log
```diff
@@ -647,7 +612,7 @@ X-Xss-Protection: 0
   "etag": "abcdef0123A=",
   "friendlyName": "bigquerydataset-fullyconfigured-updated",
   "id": "000000000000000000000",
-  "isCaseInsensitive": true,
```
maqiuyujoyce: Is this change intentional? If so, could you provide more context?

pkg/test/resourcefixture/testdata/stateabsentinspec/bigquerydataset/_http.log
```diff
@@ -291,8 +389,61 @@ X-Xss-Protection: 0
 
 ---
 
-DELETE https://bigquery.googleapis.com/bigquery/v2/projects/${projectId}/datasets/${datasetID}?alt=json&deleteContents=false
-Content-Type: application/json
+GET https://bigquery.googleapis.com/bigquery/v2/projects/${projectId}/datasets/${datasetID}?alt=json&prettyPrint=false
```
maqiuyujoyce, line 264: Going through all the diffs, I found that there is an additional GET right before DELETE. Is it intentional? If so, why?

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/basicbigquerydataset-direct/_generated_export_basicbigquerydataset-direct.golden
```diff
@@ -8,15 +8,6 @@ metadata:
     managed-by-cnrm: "true"
   name: bigquerydatasetsample${uniqueId}
 spec:
-  access:
```
maqiuyujoyce: This diff looks concerning to me. Could you help me understand why?

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/bigquerydatasetnochangeupdate-direct/_generated_export_bigquerydatasetnochangeupdate-direct.golden
```diff
@@ -8,15 +8,6 @@ metadata:
     managed-by-cnrm: "true"
   name: bigquerydatasetsample${uniqueId}
 spec:
-  access:
```
maqiuyujoyce: ditto

pkg/test/resourcefixture/testdata/reconcileintervalannotations/bigquerydataset/_generated_export_bigquerydataset.golden
```diff
@@ -8,15 +8,6 @@ metadata:
     managed-by-cnrm: "true"
   name: bigquerydatasetsample${uniqueId}
 spec:
-  access:
```
maqiuyujoyce: ditto

maqiuyujoyce (commented):
I have some questions about the diff in between TF-based behavior and the direct behavior.

I'm concerned about the additional requests the direct controller introduces and the potential scalability restriction it may bring. Could you either provide some context about why the requests are truly necessary or make some adjustments before we switch to pure direct?

I won't be available to do further follow up in the next two weeks. @jingyih do you think you can take it over?

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/basicbigquerydataset-direct/_http.log
xiaoweim: We still have the location, it is not removed. It is in the next line.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/fullybigquerydataset-direct/_http.log
xiaoweim: The `isCaseInsensitive` field should be updated to `false`. In the [create.yaml](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/fullybigquerydataset/create.yaml#L28) this field is set to true, then in the [update.yaml](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/fullybigquerydataset/update.yaml#L28) this field is set to false.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/basicbigquerydataset-direct/_generated_export_basicbigquerydataset-direct.golden
xiaoweim: This is recorded again mockgcp. The diff is between direct controller real gcp and mockgcp. The gcp resource defaults the following access values where our mock does not. 

Let me update the mock.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/bigquerydatasetnochangeupdate-direct/_generated_export_bigquerydatasetnochangeupdate-direct.golden
xiaoweim: This is recorded again mockgcp.

pkg/test/resourcefixture/testdata/reconcileintervalannotations/bigquerydataset/_generated_export_bigquerydataset.golden
xiaoweim: Same here, this is recorded again mockgcp.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/fullybigquerydataset-direct/_http.log
xiaoweim: I suspect terraform controller is not handling this field correctly. I re-ran and checked the terraform controller log, it did not pick up the field update during the UPDATE call.

This change in direct should be the correct behavior, it is updating the field correctly.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/fullybigquerydataset-direct/_generated_export_fullybigquerydataset-direct.golden
xiaoweim: Same here. The isCaseInsensitive field should be updated to false. In the [create.yaml](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/fullybigquerydataset/create.yaml#L28) this field is set to true, then in the [update.yaml](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/fullybigquerydataset/update.yaml#L28) this field is set to false. TF is not updating it correctly.

pkg/test/resourcefixture/testdata/stateabsentinspec/bigquerydataset/_http.log
xiaoweim: Yes I saw the additional GET right before DELETE as well.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/basicbigquerydataset-direct/_http.log
xiaoweim: For this resource, the gcp server does defaulting for two fields: `isCaseInsensitive` and `maxTimeTravelHours`. For test cases that do not have these two field set, it makes two PATCH requests. 

When we first create a resource, it does not have the two default fields set in the kcc object. The two fields are set in the gcp object. When we do a GET call and write the object with the two fields set to `a.actual`, the actual object is different from the desired object, so we make a PATCH request. This is done after we try to run reconciliation on this resource after create.

The second PATCH request is similar to the PUT request that it updates the resource with what we have in the `update.yaml` file.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/bigquerydatasetaccessfull-direct/_http.log
xiaoweim: It is same here because one of/both the `isCaseInsensitive` and `maxTimeTravelHours` are not set in the `create.yaml`, but gcp server gives them default values that writes that to the `a.actual` object that creates a difference between `a.actual` and `a.desire` and triggers an update during reconciliation.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/bigquerydatasetaccessblock-direct/create.yaml
```diff
@@ -16,8 +16,6 @@ apiVersion: bigquery.cnrm.cloud.google.com/v1beta1
 kind: BigQueryDataset
 metadata:
   name: bigquerydataset${uniqueId}
-  annotations:
-    alpha.cnrm.cloud.google.com/reconciler: "direct"
```
cheftako, line 20: labels:
    configsync.gke.io: "true"

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/bigquerydatasetaccessblock-direct/update.yaml
```diff
@@ -16,11 +16,9 @@ apiVersion: bigquery.cnrm.cloud.google.com/v1beta1
 kind: BigQueryDataset
 metadata:
   name: bigquerydataset${uniqueId}
-  annotations:
-    alpha.cnrm.cloud.google.com/reconciler: "direct"
```
cheftako, line 20: Can we add the label here as well. Please make sure it is *not* send to GCP.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/basicbigquerydataset-direct/_http.log
maqiuyujoyce: Thanks for the details. If the PATCH is triggered because the two fields are defaulted by the API, then this looks like an unexpected call to me. For example, when user doesn't set maxTimeTravelHours but it has a value 168 in the actual object, I'd expect the KRM object gets updated directly without making the additional PATCH (which makes no change to the resource but still consume the quota). Or did I misunderstand anything?

pkg/test/resourcefixture/testdata/stateabsentinspec/bigquerydataset/_http.log
maqiuyujoyce: Do you happen to know why?

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/basicbigquerydataset-direct/_http.log
xiaoweim: Yes there are two fields defaulted by the API and the value is in the actual GCP object. But it is not in the KRM object. It happened when it tried to diff the desired KRM and actual GCP object. 

My understanding is that we should ignore the fields defaulted by the resource API, and only keep the values in the actual GCP object since we do not want to implement state-into-spec merge behavior. If the user wants to manage the field, they can then set those field in the spec in their yaml.

maqiuyujoyce: Discussed offline. We decided to avoid doing the additional patches when the diffs are triggered by the defaulted fields.

pkg/controller/direct/bigquerydataset/bigquerydataset_controller.go
```diff
@@ -104,6 +105,21 @@ func (m *model) AdapterForURL(ctx context.Context, url string) (directbase.Adapt
 	return nil, nil
 }
 
+func sortAccessEntries(entries []*bigquery.AccessEntry) {
```
maqiuyujoyce: nit: Should this be in utils.go as well?

pkg/controller/direct/bigquerydataset/defaults.go
```diff
@@ -0,0 +1,39 @@
...
+	"time"
+
+	bigquery "cloud.google.com/go/bigquery"
+	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/bigquery/v1beta1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
+)
+
+const defaultMaxTimeTravelHours = "168" // 7 days (in hours)
+
+func ApplyBigqueryDatasetGCPDefaults(mapCtx *direct.MapContext, in *krm.BigQueryDatasetSpec, out *bigquery.DatasetMetadata, actual *bigquery.DatasetMetadata) {
```
maqiuyujoyce: Nit: BigQuery (capitalized Q).

pkg/controller/direct/bigquerydataset/bigquerydataset_controller.go
```diff
@@ -138,7 +154,7 @@ func (a *Adapter) Create(ctx context.Context, createOp *directbase.CreateOperati
 
 	desiredDataset := BigQueryDatasetSpec_ToProto(mapCtx, &a.desired.Spec)
 	desiredDataset.Labels = label.NewGCPLabelsFromK8sLabels(a.desired.Labels)
-	// return fmt.Errorf("filtered labels: %v", desiredDataset.Labels)
+	ApplyBigqueryDatasetGCPDefaults(mapCtx, &a.desired.Spec, desiredDataset, nil)
```
maqiuyujoyce: I feel not applying defaults at creation time is a good starting point to nudge users to be declarative in all fields. WDYT?

pkg/controller/direct/bigquerydataset/utils.go
```diff
@@ -130,3 +133,84 @@ func cloneBigQueryDatasetMetadate(in *bigquery.DatasetMetadata) *bigquery.Datase
...
+	sortAccessEntries(a1)
+	sortAccessEntries(a2)
+	for i := range a1 {
+		if a1[i].EntityType != bigquery.RoutineEntity && a1[i].EntityType != bigquery.ViewEntity && a1[i].EntityType != bigquery.DatasetEntity {
+			if !reflect.DeepEqual(a1[i], a2[i]) {
+				return true
+			}
+			continue
+		}
+		if diffView(a1[i].View, a2[i].View) {
```
maqiuyujoyce: Just curious: Could we still leverage deep equal after some sorting?

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/basicbigquerydataset-direct/_http.log
```diff
@@ -151,82 +153,6 @@ PATCH https://bigquery.googleapis.com/bigquery/v2/projects/${projectId}/datasets
...
-Server: ESF
-Vary: Origin
-Vary: X-Origin
-Vary: Referer
-X-Content-Type-Options: nosniff
-X-Frame-Options: SAMEORIGIN
-X-Xss-Protection: 0
-
 {
   "access": [
```
maqiuyujoyce: Access section was not in the PATCH before, but is now in it... I think this is unexpected as there is no update, right?

xiaoweim: When access section is not set by the customer, the service API returns the dataset with default access. In the added `applyGCPdefaults`, if actual.access is set, I set the krm object with the same value.

pkg/controller/direct/bigquerydataset/utils.go
xiaoweim: Yes!

pkg/controller/direct/bigquerydataset/bigquerydataset_controller.go
xiaoweim: I agree! Let's apply it at UPDATE only.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/basicbigquerydataset-direct/_http.log
maqiuyujoyce: Got it. The reason why I asked this question is I assume a PATCH request won't contain any field that is not updated. But seems like it contains the full object?

As long as this won't cause any behavior changes from user's perspective, this is non-blocking.

xiaoweim: Ah yes right now it contains the full object. It does not cause behavior changes but I agree that PATCH requests should only contain fields that need to be updated.

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/basicbigquerydataset-direct/_http.log
```diff
@@ -154,30 +152,11 @@ Content-Type: application/json
...
-      "specialGroup": "projectWriters"
-    }
-  ],
   "friendlyName": "bigquerydataset-sample-updated",
   "labels": {
     "cnrm-test": "true",
     "managed-by-cnrm": "true"
-  },
-  "maxTimeTravelHours": "168"
+  }
```
maqiuyujoyce: Non-blocking: What about labels?

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/bigquerydatasetaccessblock-direct/_generated_export_bigquerydatasetaccessblock-direct.golden
```diff
@@ -9,12 +9,14 @@ metadata:
   name: bigquerydataset${uniqueId}
 spec:
   access:
-  - domain: google.com
-    role: READER
-  - iamMember: allUsers
-    role: READER
   - role: OWNER
     specialGroup: projectOwners
+  - role: OWNER
```
maqiuyujoyce: Did you change the mock for default behavior?

pkg/test/resourcefixture/testdata/basic/bigquery/v1beta1/bigquerydataset/bigquerydatasetaccessfull-direct/_http.log
```diff
@@ -1052,6 +1120,130 @@ X-Xss-Protection: 0
 
 ---
 
+PATCH https://bigquery.googleapis.com/bigquery/v2/projects/${projectId}/datasets/${datasetID}?alt=json&prettyPrint=false
```
maqiuyujoyce, line 1119: Hmmm could you help me understand why a new PATCH request is added?

xiaoweim: Not sure, I re-ran the test locally and it only had one PATCH request. Let me re-record the logs.

yuwenma:
@xiaoweim Could you give an update about this PR and move it to the future milestone? (1.141 if planned for 2025, and 1.142 if for 2026)

PR #4943, opened 2025-08-06
chore: refactor test methods into a test package

justinsb:


pkg/controller/gsakeysecretgenerator/service_account_key_integration_test.go
```diff
@@ -173,12 +173,26 @@ func verifyGSAKeyRemoved(t *testing.T, iamClient *iam.Service, keyName string) {
 	}
 }
 
+func UnmarshalFileToCRD(t *testing.T, fileName string) *apiextensions.CustomResourceDefinition {
```
cheftako, line 176: Not blocking but not sure why this is in service_account_key_integration_test.go. This seems like it belongs in more of a utility go file.

justinsb: I pulled it out of the shared utility because it's not something I think we should encourage use of, and this is the last test to use it :-)

PR #4945, opened 2025-08-06
autobump: gcp dependencies

justinsb:
* autobump: gcp dependencies

* fuzzers: mark newly introduced fields as ignored for now

Trying to mark these fields as tasks for future work, with "//todo:add_support"

### DIFF
--- a/pkg/controller/direct/backupdr/backupplanassociation_fuzzer.go
+++ b/pkg/controller/direct/backupdr/backupplanassociation_fuzzer.go
@@ -33,20 +33,30 @@ func backupDRBackupPlanAssociationFuzzer() fuzztesting.KRMFuzzer {
 		BackupDRBackupPlanAssociationObservedState_v1beta1_FromProto, BackupDRBackupPlanAssociationObservedState_v1beta1_ToProto,
 	)
 
+	// Spec fields
 	f.SpecFields.Insert(".resource_type")
 	f.SpecFields.Insert(".resource")
 	f.SpecFields.Insert(".backup_plan")
 
+	// Status fields
 	f.StatusFields.Insert(".create_time")
 	f.StatusFields.Insert(".update_time")
 	f.StatusFields.Insert(".state")
 	f.StatusFields.Insert(".rules_config_info")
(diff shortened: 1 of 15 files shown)


pkg/controller/direct/backupdr/backupplanassociation_fuzzer.go
```diff
@@ -33,20 +33,30 @@ func backupDRBackupPlanAssociationFuzzer() fuzztesting.KRMFuzzer {
...
 
+	// Identity fields
 	f.UnimplementedFields.Insert(".name")
 
+	// Fields that could potentially be added
+
 	// the conversion of `spec.resource` relies on correctly setting `.spec.resourceType` to a magic string "compute.googleapis.com/Instance"
 	f.UnimplementedFields.Insert(".resource")
 
+	f.UnimplementedFields.Insert(".backup_plan_revision_name")                             // todo:add_support
```
jingyih, line 56: Ideally, the `dev/tasks/update-dependency` script should be enhanced to automatically mark new fields as "unimplemented".

Do we want to surface the "ignoredFields" so that we can more easily distinguish between "a field that we intentionally want to leave unimplemented" vs. "a new field that is not yet implemented".

justinsb: > Ideally, the dev/tasks/update-dependency script should be enhanced to automatically mark new fields as "unimplemented".

Agreed, but I suggest not in this PR.  I think we can achieve that with an agentic loop (invoking the fuzzer, we might want to tweak how the fuzzer presents the output to give us a field path to make it easy).

> Do we want to surface the "ignoredFields" so that we can more easily distinguish between "a field that we intentionally want to leave unimplemented" vs. "a new field that is not yet implemented".

Agreed, I was thinking about creating NotYetFields and IdentityFields; we would union them all into Unimplemented but it would show intent more clearly.  I just think not in this PR.

jingyih: Yeah sounds good. I was just offering some of my thoughts.

PR #4948, opened 2025-08-06
mockgcp: clean up dependencies to fix go.mod

justinsb:
- **chore: clean up mockgcp dependencies**
  Ideally we would not depend on KCC "root",
  but we really do not want a terraform dependency - that causes go.mod probles.
  

- **mockgcp: Fix go.mod and run go.mod tidy**
  go.mod was not resolving correctly without the replace directives.


mockgcp/go.mod
```diff
@@ -1,159 +1,196 @@
 module github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp
 
-go 1.24
+go 1.24.0
 
 toolchain go1.24.5
 
+replace github.com/GoogleCloudPlatform/k8s-config-connector => ./..
```
yuwenma, line 7: Is this aimed to sync the GCP lib between KCC and MockGCP? 
I see the benefits but it also feels hacky.

yuwenma (commented):
/approve

mockgcp/go.mod
justinsb: What do you propose?

yuwenma: If this is to sync the GCP libs, I thought we already have `go.work` and can use `go work sync`  (using the version in MockGCP or whatever is higher).

justinsb: I wasn't able to get that to work.  Can you try?

PR #4960, opened 2025-08-08
chore: Refactor GKEHubFeatureMembership references resolver

gemmahou:
### Change description

Refactor GKEHubFeatureMembership references resolver

GKEHubFeatureMembership is a "virtual resource" used to manage the MembershipSpecs field in GKEHubFeature. 

The MembershipSpecs are configured in the GKEHubFeatureMembership(via Configmanagement, Mesh or Policycontroller), not the GKEHubMembership itself. This means we only need to match the membershipID to an existing key in the MembershipSpecs list to perform CRUD operations on the GKEHubFeature. Resolving the full GKEHubMembership resource is unnecessary. (I also added comment in the code)

Todo: 
However, we need to confirm if the API allows creating or updating a non-existent membership in GKEHubFeature's MembershipSpecs. It's unclear if we need to resolve the GKEHubMembership resource just to validate its existence, or if we can rely on the API to return an error if the membership doesn't exist.

```release-note
NONE
```

```docs

```

#### Intended Milestone

### Tests you have done

Added two more scenarios tests to verify that GKEHubFeatureMembership can be deleted successfully, after the removal of GKEHubFeature or GKEHubMembership. 

### DIFF
--- a/pkg/controller/direct/gkehub/featuremembership_controller.go
+++ b/pkg/controller/direct/gkehub/featuremembership_controller.go
@@ -20,6 +20,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/GoogleCloudPlatform/k8s-config-connector/apis/common"
+
 	featureapi "google.golang.org/api/gkehub/v1beta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -65,6 +67,7 @@ type gkeHubAdapter struct {
 	actual  *featureapi.Feature
 
 	hubClient *gkeHubClient
+	reader    client.Reader
 }
 
 var _ directbase.Adapter = &gkeHubAdapter{}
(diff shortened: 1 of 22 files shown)


pkg/controller/direct/gkehub/featuremembership_controller.go
```diff
@@ -153,6 +153,25 @@ func (a *gkeHubAdapter) Find(ctx context.Context) (bool, error) {
...
+	// membershipIDs in the GKEHubFeature's MembershipSpecs list. This allows us to manage the target membership
+	// using the GKEHubFeatureMembership configuration, without needing to resolve the resource.
+	membershipRef := a.desired.Spec.MembershipRef
+	if membershipRef.External != "" {
+		// use the provided external value
+		a.membershipID = membershipRef.External
+	} else {
+		// build membershipID by its project, location and name
+		if membershipLocation := a.desired.Spec.MembershipLocation; membershipLocation != nil {
+			a.membershipID = fmt.Sprintf("projects/%s/locations/%s/memberships/%s", a.projectID, common.ValueOf(membershipLocation), membershipRef.Name)
```
justinsb, line 168: KCC resource name != GCP name, e.g. if spec.resourceID is set.  Does GKEHubFeatureMembership support resourceID?

gemmahou: This is a good point! I guess you meant the referenced GKEHubMembership, it does support resourceID, so it's possible that the KCC resource name != GCP resource name.

gemmahou:
No longer needed.

PR #4992, opened 2025-08-18
MonitoringNotificationChannel: CRD/Mapper/Fuzzer

justinsb:
- **MonitoringNotificationChannel: CRD/Mapper/Fuzzer**
  

- **autogen: MonitoringNotificationChannel**


yuwenma:
the fuzz-roundtrippers took the longest time so far (47m), could we split it into per-group tests as well?

PR #4995, opened 2025-08-18
Fix failure to write .status.observedGeneration to CCC

cheftako:
### Change description

I have fixed the bug where the status.observedGeneration field on the ConfigConnectorContext resource was not being set.

  Here is a brief explanation of the key changes I made and why:

  The declarative.Reconciler correctly sets the status.observedGeneration field. However, the handleReconcileSucceeded and
  handleReconcileFailed functions were overwriting the entire status with a new object, which discarded the observedGeneration.

  I modified these functions to instead read the existing status, update the Healthy and Errors fields, and then write the
  modified status back. This preserves the observedGeneration set by the declarative.Reconciler. I also updated the
  corresponding unit tests to verify this behavior. Finally, I ran all the tests for the controller and compiled the code to
  ensure the change was safe.

```release-note
status.oservedGeneration is now being set on the ConfigConnectorContext.
```

```docs
NONE
```

#### Intended Milestone

1.134

### Tests you have done

### DIFF
--- a/operator/pkg/controllers/configconnectorcontext/configconnectorcontext_controller_test.go
+++ b/operator/pkg/controllers/configconnectorcontext/configconnectorcontext_controller_test.go
@@ -882,8 +882,9 @@ func TestHandleReconcileFailed(t *testing.T) {
 
 	ccc := &corev1beta1.ConfigConnectorContext{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      corev1beta1.ConfigConnectorContextAllowedName,
-			Namespace: "foo-ns",
+			Name:       corev1beta1.ConfigConnectorContextAllowedName,
+			Namespace:  "foo-ns",
+			Generation: 1,
 		},
 		Spec: corev1beta1.ConfigConnectorContextSpec{
 			GoogleServiceAccount: "foo@bar.iam.gserviceaccount.com",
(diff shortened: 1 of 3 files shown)


operator/pkg/controllers/configconnectorcontext/configconnectorcontext_controller_test.go
```diff
@@ -896,6 +897,13 @@ func TestHandleReconcileFailed(t *testing.T) {
 	if err := c.Create(ctx, ccc); err != nil {
 		t.Fatalf("failed to create ConfigConnectorContext: %v", err)
 	}
+
+	// Manually set the observed generation, simulating the declarative reconciler.
+	ccc.Status.ObservedGeneration = 1
```
sdowell, line 902: From my reading of the code it seems like the configconnectorcontext controller implements its own v1alpha1 CommonStatus builder as opposed to using one of the [k-d-p status builders](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/tree/master/pkg/patterns/addon/pkg/status), so I expect the controller will need to be updated to set the observedGeneration. As is it doesn't look like this unit test actually validates that the controller sets the observedGeneration, just that the controller doesn't clobber the observedGeneration

PR #5002, opened 2025-08-19
tests: labels check

acpana:
### Change description

Going to let the CI run for a second.

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/tests/e2e/unified_test.go
+++ b/tests/e2e/unified_test.go
@@ -443,6 +443,39 @@ func runScenario(ctx context.Context, t *testing.T, testPause bool, fixture reso
 					}
 				}
 
+				if ShouldTestMetadataLabels(t, testName, primaryResource) {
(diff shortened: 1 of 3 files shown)


acpana:
cc @cheftako @justinsb @xiaoweim @anhdle-sso 

This patch piggybacks on the reconciliation test to add the labels and force a reconcile; if no logs change then we know we are not sending a metadata caused diff only.

tests/e2e/unified_test.go
```diff
@@ -443,6 +443,39 @@ func runScenario(ctx context.Context, t *testing.T, testPause bool, fixture reso
 					}
 				}
 
+				if ShouldTestMetadataLabels(t, testName, primaryResource) {
```
justinsb, line 446: I think I prefer doing this as a separate step to automagically changing the "main" reconciliation, but I also think we can just update create.yaml and update.yaml to include labels (at least one per kind)

PR #5014, opened 2025-08-20
crds: make kind optional for well-known Project refs

justinsb:


yuwenma (commented):
/approve
/lgtm

Oh, nice! Especially since we cannot drop the kind

PR #5015, opened 2025-08-20
controllerbuilder: fix generation to recognize plural Refs

justinsb:
* Recognize plural Refs
* Fix import namespace to avoid conflicts when we have Refs to more than one package


yuwenma (commented):
/lgtm
/approve

PR #5016, opened 2025-08-20
chore: fix controllerbuilder to write empty files

justinsb:
Otherwise when we had no generated types left, we were leaving the old file behind.

### DIFF
--- a/dev/tools/controllerbuilder/pkg/commands/generatetypes/generatetypescommand.go
+++ b/dev/tools/controllerbuilder/pkg/commands/generatetypes/generatetypescommand.go
@@ -185,7 +185,8 @@ func RunGenerateCRD(ctx context.Context, o *GenerateCRDOptions) error {
 	}
 
 	addCopyright := true
-	if err := typeGenerator.WriteFiles(addCopyright); err != nil {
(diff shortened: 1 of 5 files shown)


dev/tools/controllerbuilder/pkg/commands/generatetypes/generatetypescommand.go
```diff
@@ -185,7 +185,8 @@ func RunGenerateCRD(ctx context.Context, o *GenerateCRDOptions) error {
 	}
 
 	addCopyright := true
-	if err := typeGenerator.WriteFiles(addCopyright); err != nil {
+	writeEmptyFiles := true
```
yuwenma, line 188: I may miss something. But this looks like we can just not call the `typeGenerator.WriteFiles` if wirteEmptyFiles is true, right?

justinsb: Not sure I follow: we want to write the empty file

yuwenma: Mis-read the "not" condition in line 86

PR #5022, opened 2025-08-21
chore: fix generation for vmwareengine

justinsb:


jingyih (commented):
Thank you for the fix! (I think I previously implemented this resource)

/lgtm
/approve

PR #5023, opened 2025-08-21
[labels] add API checks

yuwenma:
### Change description

This API Checks gives us an overview about the existing labels status.
* 52 resources support `spec.labels` 
* 7 have bad label type (not string map)
* 3 missing test coverage for `metadata.labels` fields
* 2 missing test coverage for `spec.labels` (beta) fields (in existing missingfields.txt)
* 35 missing test coverage for `spec.labels` (alpha) fields (in existing alpha-missingfields.txt)  

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


tests/apichecks/crds_test.go
```diff
@@ -464,6 +465,126 @@ func TestCRDShortNames(t *testing.T) {
...
+				// GCP labels are a map of strings
+				if field.props == nil || field.props.Type != "object" || field.props.AdditionalProperties == nil || field.props.AdditionalProperties.Schema == nil || field.props.AdditionalProperties.Schema.Type != "string" {
+					return
+				}
+
+				parts := strings.Split(field.FieldPath, ".")
+				if len(parts) == 0 {
+					return
+				}
+				errs = append(errs, fmt.Sprintf("[gcplabels] crd=%s version=%v: field %s", crd.Name, version.Name, strings.TrimPrefix(field.FieldPath, ".")))
```
justinsb: Do you want to say something like "resource is missing spec.labels" ?  It's not self-documenting what this list is.

We're also going to have some "perma errors" here, for GCP resources that don't support labels, but I guess that's OK, because the interesting thing is when this list changes.

(I disagree that we want to get full spec.labels coverage now, but that's orthogonal to having a tracking list)

tests/apichecks/crds_test.go
```diff
@@ -464,6 +465,126 @@ func TestCRDShortNames(t *testing.T) {
...
+	test.CompareGoldenFile(t, "testdata/supported_labels.txt", want)
+}
+
+func TestLegacyCompatibleLabels(t *testing.T) {
+	crds, err := crdloader.LoadAllCRDs()
+	if err != nil {
+		t.Fatalf("error loading CRDs: %v", err)
+	}
+
+	unstructs := loadUnstructs(t)
```
justinsb: Nit: We need to rename this function to loadAllTestFixtures or something

tests/apichecks/crds_test.go
```diff
@@ -464,6 +465,126 @@ func TestCRDShortNames(t *testing.T) {
...
+					return
+				}
+
+				parts := strings.Split(field.FieldPath, ".")
+				if len(parts) == 0 {
+					return
+				}
+				fieldName := parts[len(parts)-1]
+
+				if strings.ToLower(fieldName) == "labels" {
```
justinsb: Won't this always be true?

tests/apichecks/crds_test.go
```diff
@@ -464,6 +465,126 @@ func TestCRDShortNames(t *testing.T) {
...
+					// This is a valid label field, check if it is always set
+					kind := crd.Spec.Names.Kind
+					kindUnstructs := unstructsByKind[kind]
+					if len(kindUnstructs) == 0 {
+						return
+					}
+					alwaysSet := true
```
justinsb: I think we want to test that labels _are_ set for each kind, not that they are always set.

We should also check that in at least one test that we have keys/values that are not valid (and are filtered out before being passed to GCP)

tests/apichecks/crds_test.go
```diff
@@ -464,6 +465,126 @@ func TestCRDShortNames(t *testing.T) {
...
+					for _, u := range kindUnstructs {
+						if !hasField(u.Object, field.FieldPath) {
+							alwaysSet = false
+							break
+						}
+					}
+					if alwaysSet {
```
justinsb: This should probably `!alwaysSet` (though I think the logic needs a tweak anyway)

tests/apichecks/crds_test.go
```diff
@@ -464,6 +465,126 @@ func TestCRDShortNames(t *testing.T) {
...
+	crds, err := crdloader.LoadAllCRDs()
+	if err != nil {
+		t.Fatalf("error loading CRDs: %v", err)
+	}
+
+	var errs []string
+	for _, crd := range crds {
+		for _, version := range crd.Spec.Versions {
+			visitCRDVersion(version, func(field *CRDField) {
+				if !strings.HasPrefix(field.FieldPath, ".spec.") {
```
justinsb, line 569: We don't need both if statements?

tests/apichecks/crds_test.go
```diff
@@ -852,8 +973,8 @@ func TestMultiVersionCRDNoDiff(t *testing.T) {
 			if diff := cmp.Diff(string(expectedDiff), allDiffs.String()); diff != "" {
 				// To address inconsistencies between local and CI environments,
 				// we normalize the diff output by replacing non-breaking spaces with regular spaces.
-				normalizedActual := strings.ReplaceAll(allDiffs.String(), " ", " ")
-				normalizedExpected := strings.ReplaceAll(string(expectedDiff), " ", " ")
+				normalizedActual := strings.ReplaceAll(allDiffs.String(), " ", " ")
```
justinsb: What's happening here?  Do we want to switch to unicode characters for the "non-breaking space"

tests/apichecks/crds_test.go
yuwenma: Agree on seeing this list changes. This test is to list all (direct) resources that support labels, it's not the `resource is missing spec.labels` because we already have a generic api coverage test for that `TestCRDFieldPresenceInTests`

tests/apichecks/crds_test.go
yuwenma: I used to use a more broad logic to find other possible fields that are similiar to `spec.labels`, but changed my mind due to the unlikeliness.

tests/apichecks/crds_test.go
yuwenma: This `alwaysSet` is for a different test: We want to make sure the `metadata.labels` is tested if GCP labels exist.

The `spec.labels` is already covered in existing test `TestCRDFieldPresenceInTests`

tests/apichecks/crds_test.go
yuwenma: cleaned up

tests/apichecks/crds_test.go
justinsb: OK, can we update the message to describe what this is / make this actionable

tests/apichecks/crds_test.go
```diff
@@ -464,6 +465,126 @@ func TestCRDShortNames(t *testing.T) {
...
+				// GCP labels are a map of strings
+				if field.props == nil || field.props.Type != "object" || field.props.AdditionalProperties == nil || field.props.AdditionalProperties.Schema == nil || field.props.AdditionalProperties.Schema.Type != "string" {
+					return
+				}
+
+				parts := strings.Split(field.FieldPath, ".")
+				if len(parts) == 0 {
+					return
+				}
+				message = append(message, fmt.Sprintf("[gcplabels] crd=%s version=%v: supports GCP labels. The fixture test should cover both `metadata.labels` and `spec.labels`", crd.Name, version.Name))
```
justinsb, line 494: This linter isn't looking at the fixture tests though, it is just looking at the CRDs

tests/apichecks/crds_test.go
```diff
@@ -464,6 +465,126 @@ func TestCRDShortNames(t *testing.T) {
...
+				}
+				alwaysSet := true
+				for _, u := range kindUnstructs {
+					if !hasField(u.Object, field.FieldPath) {
+						alwaysSet = false
+						break
+					}
+				}
+				if alwaysSet {
+					errs = append(errs, fmt.Sprintf("[missing_legacy_label_tests] crd=%s version=%v: field %s is always set in tests", crd.Name, version.Name, strings.TrimPrefix(field.FieldPath, ".")))
```
justinsb, line 548: This still reads like a good thing

tests/apichecks/crds_test.go
justinsb: Still true, I believe?

PR #5038, opened 2025-08-22
chore: enforce that types and mappers should be generated

justinsb:


dev/tasks/generate-types-and-mappers
```diff
@@ -0,0 +1,102 @@
...
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+cd ${REPO_ROOT}
+
+pushd dev/tools/controllerbuilder
+./generate-proto.sh
+popd
+
+# For efficiency, we generate CRDs only once at the end
+export SKIP_GENERATE_CRDS=1
```
yuwenma, line 28: Cool!

PR #5039, opened 2025-08-22
chore: fix networksecurity generation

justinsb:
- **chore: enforce that types and mappers should be generated**
- **chore: fix networksecurity generation**


pkg/controller/direct/networksecurity/mapper.generated.go
yuwenma, line 1: I like solving it in the dev/tasks. FYI, we also have this fixed in the GEMINI.md (linked doc) https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/ai/how-to-promote-resource.md#action-1

yuwenma (commented):
/lgtm
/approve

Thanks for the fix.

PR #5040, opened 2025-08-22
chore: Introduce shared project mapper

justinsb:


pkg/controller/direct/cloudbuild/workerpool_controller.go
```diff
@@ -370,7 +351,7 @@ func (a *Adapter) resolveDependencies(ctx context.Context, reader client.Reader,
 			return err
 		}
 
-		if err := networkSpec.PeeredNetworkRef.ConvertToProjectNumber(ctx, a.projectClient); err != nil {
+		if err := networkSpec.PeeredNetworkRef.ConvertToProjectNumber(ctx, a.projectMapper); err != nil {
 			return err
```
yuwenma, line 355: Can we allow errors if the ConvertToProjectNumber fails? IIRC, one problem we hit is that the SA does not (should not) have the permission for the resourcemanager service.

PR #5050, opened 2025-08-22
feat: add resource controller mapping

xiaoweim:
### Change description

Implement a new way to register/select controllers for each resources in namespace mode. 

This is the FIRST PR.

- It adds a static map of resources to default and supported controllers. 
- It adds a `experiments.controlleroverride` field in CCC to allow users to choose which controller to run if the users are not happy with the default controller. 

Second PR: #5072

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


acpana:
feel free to update the exception; I think we should merge in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5058 to skip CC/ CCCs from the unit test failure.

pkg/controller/resourceconfig/static_config.go
acpana, line 1: What if we treated this as an "exception" file. Where we document TF, DCL, Legacy controllers but presume direct otherwise?

pkg/controller/resourceconfig/static_config.go
```diff
@@ -0,0 +1,1776 @@
...
+		SupportedControllers: []k8s.ReconcilerType{k8s.ReconcilerTypeTerraform},
+	},
+	{Group: "spanner.cnrm.cloud.google.com", Kind: "SpannerInstance"}: {
+		DefaultController:    k8s.ReconcilerTypeTerraform,
+		SupportedControllers: []k8s.ReconcilerType{k8s.ReconcilerTypeTerraform, k8s.ReconcilerTypeDirect},
+	},
+	{Group: "iam.cnrm.cloud.google.com", Kind: "IAMAuditConfig"}: {
+		DefaultController:    k8s.ReconcilerTypeDirect,
+		SupportedControllers: []k8s.ReconcilerType{k8s.ReconcilerTypeDirect},
+	},
```
acpana, line 215: for all the IAM resources we need a different controller type called: `legacy` since these are handwriten controllers

pkg/controller/resourceconfig/static_config.go
```diff
@@ -0,0 +1,1776 @@
...
+// limitations under the License.
+
+package resourceconfig
+
+import "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
+
+var ControllerConfigStatic = &ResourcesControllerMap{
+	{Group: "storage.cnrm.cloud.google.com", Kind: "StorageBucketAccessControl"}: {
+		DefaultController:    k8s.ReconcilerTypeTerraform,
+		SupportedControllers: []k8s.ReconcilerType{k8s.ReconcilerTypeTerraform},
```
acpana, line 22: it would be nice if we could document how we generated this file? We might need to regen it in the short term but hopefully never in the long term!

acpana: Actually on second thought, we could also re-use the `supportedgvks` file right? I think the ethos behind that was to get to this point eventually where we statically define attributes for GVKs (like what kind of controller to use)

operator/pkg/apis/core/v1beta1/configconnectorcontext_types.go
acpana, line 1: I can be told off but I think it's easier (read safer and simpler) for us the first time around to restart the kccmanager pods with override flags and not analyze the CCC value in kccmanager.

An example of this is in this WIP PR:

https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5032/commits/f206fe941bc182c853e8af0a0f61914b9d4aeb83#diff-213491bcaf2de067ac09d20e0e7fa08a6ca5f7db3aa4eaea597c9019bca4651fR136

Doing this doesn't actually prevent us from reading the CCC values later.

operator/pkg/apis/core/v1beta1/configconnectorcontext_types.go
```diff
@@ -80,6 +81,17 @@ type ConfigConnectorContextSpec struct {
...
+	Experiments Experiments `json:"experiments,omitempty"`
+}
+
+// Experiments contains experimental features.
+
+type Experiments struct {
+	// ControllerOverrides allows specifying which controller to use for a given
+	// resource kind within this namespace, overriding the system default.
+	// +optional
+	ControllerOverrides map[string]k8scontrollertype.ReconcilerType `json:"controllerOverrides,omitempty"`
```
acpana, line 94: I thought this might be simpler (without taking on a dep into pkg k8s for now)

```suggestion
	// Configuration for experimental features.
	// +optional
	Experiments *Experiments `json:"experiments,omitempty"`
}

// ControllerOverride defines a key-value pair for a controller override.
type ControllerOverride struct {
	// The name of the controller to override.
	// +required
	Name string `json:"name"`
	// The value of the override.
	// +required
	Value string `json:"value"`
}

// Experiments defines the configuration for experimental features.
type Experiments struct {
	// A list of key-value pairs for controller overrides.
	// +optional
	ControllersOverrides []ControllerOverride `json:"controllersOverrides,omitempty"`
}
```

The simple key/value string mechanism is what I wanted to point out. At some point, we probably need to validate these values.

Thinking about the UX here some more, I might've misunderstood @justinsb's point but if there's a future need for wildcard support like `*` going with a more expressive key for the GVK like (group: verison, kind: ) might actually be better 🤷🏼

pkg/controller/resourceconfig/static_config.go
acpana: the current ReconcilerType const to define specific IAM roles but I'm not convinced we want to confuse practitioners b/w `direct`, `iampartialpolicy`. As in, it may be cleaner to say `IAMPartialPolicy: direct` or `IAMPartialPolicy: legacy`

pkg/controller/resourceconfig/static_config.go
xiaoweim: I like the idea since our overall goal is to migrate everything to direct. So if we can shorten the list, it means that we are making progress, and eventually we may be able to get rid of the list. But I want to keep the `supported controllers` so that if the customer wants to overwrite in the CCC but the controller is not supported, we can quickly return an error message saying their chosen controller is not supported and what are the supported controllers.

pkg/controller/resourceconfig/static_config.go
xiaoweim: Gotcha! That makes sense. Let me update for IAM resources. Just to clarify that except for `IAMPartialPolicy`, the rest of the IAM resources are all using legacy hand written controllers?

pkg/controller/resourceconfig/static_config.go
xiaoweim: I used gemini-cli to write a python script that generated this file, it is based on the `tf2crd` and `dcl2crd` label. Since we will remove all the TF and DCL labels in the CRD, regenerating it in the same way would not be possible. Ideally in the future when we turn on direct controller, we just need to update the default controller in this file.

operator/pkg/apis/core/v1beta1/configconnectorcontext_types.go
xiaoweim: I am not sure I am understanding correctly, why do we need the `name` of the controller to override here? and the `value` to override to be a string instead of a defined reconciler type?
```
	// The name of the controller to override.
	// +required
	Name string `json:"name"`
	// The value of the override.
	// +required
	Value string `json:"value"`
```

xiaoweim:
Closing this PR based off our discussion to merge the new controller routing logic and test with safer upgrade. The entire change is in #5072

PR #5059, opened 2025-08-25
feat: mockgcp for label validation check

yuwenma:
### Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/mockgcp/interceptor/validation.go
+++ b/mockgcp/interceptor/validation.go
@@ -17,6 +17,7 @@ package interceptor
 import (
 	"context"
 	"reflect"
+	"strings"
 
 	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/pkg/validation"
 	"google.golang.org/grpc"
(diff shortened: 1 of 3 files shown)


mockgcp/pkg/validation/labels.go
```diff
@@ -0,0 +1,52 @@
...
+
+package validation
+
+import (
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// ValidateLabels checks if a map of labels conforms to GCP's label requirements.
+func ValidateLabels(labels map[string]string) error {
```
justinsb, line 23: Maybe start by adding this by hand to a few mockgcp resources?  Now we have AI, I value clarity over fewer lines of code

justinsb: Although I guess it's good that this is actually picking up real failures (at least as far as I can tell!)

yuwenma: I originally did those for alloydb only, and it proves working well. Then I moved this to a more generic approach that can apply to all mockgcp services which support labels. I"m going to update the fixture tests to show this change.

yuwenma: 2 GCP services use the `labels` field for a different purpose. They are caught by this change. Please see the last git-commit

mockgcp/interceptor/validation.go
```diff
@@ -0,0 +1,95 @@
...
+	resource := getResourceFromRequest(req)
+	if resource == nil {
+		// If we can't find a resource, just pass through
+		return handler(ctx, req)
+	}
+
+	// HACK: Skip validation for CloudIdentityGroup, which uses `labels` for a different purpose.
+	// See https://cloud.google.com/identity/docs/groups
+	messageName := proto.MessageName(resource)
+	if strings.Contains(string(messageName), "cloudidentity") {
```
justinsb, line 63: We really should check ".cloudidentity." so we are less likely to get false positives, but I think in general we should try to push this down into each service...

PR #5063, opened 2025-08-26
Accept AlloyDB Database Version as a user input

kjasnoor0305:
### Description

Allow AlloyDB Database Version to be set as input by user.

Currently only the default version (POSTGRES_16) alloydb clusters could be created by KCC direct controller. We need this change to accept this field as input. 

Feature needed before Sept 15.

```
New Fields 
[AlloyDBCluster](https://cloud.google.com/config-connector/docs/reference/resource-docs/alloydb/alloydbcluster)
  *  Added *spec.databaseVersion* field
```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/tests/e2e/testdata/scenarios/alloydbinstance/basic_primary_no_change/_http01.log
+++ b/tests/e2e/testdata/scenarios/alloydbinstance/basic_primary_no_change/_http01.log
@@ -134,7 +134,7 @@ X-Xss-Protection: 0
       ]
     },
     "createTime": "2024-04-01T12:34:56.123456Z",
-    "databaseVersion": "POSTGRES_15",
+    "databaseVersion": "POSTGRES_16",
     "encryptionInfo": {
       "encryptionType": "GOOGLE_DEFAULT_ENCRYPTION"
     },
@@ -216,7 +216,7 @@ X-Xss-Protection: 0
     ]
   },
   "createTime": "2024-04-01T12:34:56.123456Z",
-  "databaseVersion": "POSTGRES_15",
+  "databaseVersion": "POSTGRES_16",
   "encryptionInfo": {
     "encryptionType": "GOOGLE_DEFAULT_ENCRYPTION"
   },
--- a/tests/e2e/testdata/scenarios/alloydbinstance/basic_primary_no_change/_object01.yaml
+++ b/tests/e2e/testdata/scenarios/alloydbinstance/basic_primary_no_change/_object01.yaml
@@ -42,7 +42,7 @@ status:
     - FRIDAY
     - SATURDAY
     - SUNDAY
-  databaseVersion: POSTGRES_15
+  databaseVersion: POSTGRES_16
   encryptionInfo:
   - encryptionType: GOOGLE_DEFAULT_ENCRYPTION
   name: projects/${projectId}/locations/europe-west1/clusters/alloydbcluster-${uniqueId}


apis/alloydb/v1alpha1/cluster_types.go
```diff
@@ -70,8 +70,8 @@ type AlloyDBClusterSpec struct {
 	//  supplied at cluster creation time, then a default database version will
 	//  be used.
 	// +kcc:proto:field=google.cloud.alloydb.v1beta.Cluster.database_version
-	DatabaseVersion *string `json:"databaseVersion,omitempty"`
 	*/
+	DatabaseVersion *string `json:"databaseVersion,omitempty"`
```
gemmahou: Let's remove line 67 and 73, and regenerate the CRD to make sure we have clear field description.

apis/alloydb/v1beta1/cluster_types.go
```diff
@@ -70,8 +70,8 @@ type AlloyDBClusterSpec struct {
 	//  supplied at cluster creation time, then a default database version will
 	//  be used.
 	// +kcc:proto:field=google.cloud.alloydb.v1beta.Cluster.database_version
-	DatabaseVersion *string `json:"databaseVersion,omitempty"`
 	*/
+	DatabaseVersion *string `json:"databaseVersion,omitempty"`
```
gemmahou, line 72: ditto

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_alloydbclusters.alloydb.cnrm.cloud.google.com.yaml
```diff
@@ -351,6 +358,87 @@ spec:
                 description: Optional. The maintenance update policy determines when
                   to allow or deny updates.
                 properties:
+                  denyMaintenancePeriods:
```
gemmahou: I think this is added after we updated go.mod alloydb import to the latest version and updated the generated types. Is that possible to split changes unrelated to database version to a separate  PR?

pkg/controller/direct/alloydb/cluster_controller.go
```diff
@@ -545,6 +545,10 @@ func (a *ClusterAdapter) Update(ctx context.Context, updateOp *directbase.Update
 
 	// 6. Set resource name. This step is not needed for other operations.
 	desiredPb.Name = a.id.String()
+	// TODO: Add support for Major version upgrade
+	if a.desired.Spec.DatabaseVersion != nil && a.actual.DatabaseVersion != alloydbpb.DatabaseVersion_DATABASE_VERSION_UNSPECIFIED && *a.desired.Spec.DatabaseVersion != a.actual.DatabaseVersion.String() {
+		return fmt.Errorf("field 'spec.databaseVersion' is immutable and cannot be updated from %q to %q", a.actual.DatabaseVersion, *a.desired.Spec.DatabaseVersion)
+	}
```
gemmahou, line 553: IIRC we do not handle immutable fields manually in individual controllers, we have [documentation](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/develop-resources/api-conventions/validations.md#rule-2-immutable-field) listed 3 options to handle immutable fields, I think this one falls in option3.

apis/alloydb/v1alpha1/cluster_types.go
kjasnoor0305: Done

apis/alloydb/v1beta1/cluster_types.go
kjasnoor0305: Done

pkg/controller/direct/alloydb/cluster_controller.go
kjasnoor0305: If I don't handle it, updating this field doesn't throw any kind of error. With this change thecustom message which will be shown in the events field.

Currently this is a temporary check and I plan to handle the database version field update while adding support for major version upgrade( in which case the customer can update this field).

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_alloydbclusters.alloydb.cnrm.cloud.google.com.yaml
kjasnoor0305: Should I not update go.mod and try? Not sure why and how a separate PR is required?

gemmahou: Let's update go.mod only and put the related changes in a separate PR. This PR will focu on the database version only. This will make it easier to track changes and, in case anything goes wrong after merging, we can easily revert the PR.

pkg/controller/direct/alloydb/cluster_controller.go
gemmahou: For now I think we leverage the API to return the error, as mentioned in option3: 
```
If the field contains more complicated GCP behavior, for example the GCP server assigns a default value if the field is not configured, Config Connector may not check the immutability but rely on the GCP server to give more user facing message.
```

gemmahou: @maqiuyujoyce regarding optional immutable fields, I think we still follow the guidance and let API to handle this, right?

mockgcp/mockalloydb/cluster.go
```diff
@@ -140,7 +140,10 @@ func setClusterFields(name *clusterName, obj *pb.Cluster) {
 	// Context: https://github.com/hashicorp/terraform-provider-google/issues/16960
 	// This field needs to be handled differently in mockgcp after we fix the
 	// behavior in the controller.
-	obj.DatabaseVersion = pb.DatabaseVersion_POSTGRES_15
+	// obj.DatabaseVersion = pb.DatabaseVersion_POSTGRES_15
```
gemmahou: nit: remove unnecessary line

apis/alloydb/v1beta1/cluster_types.go
```diff
@@ -64,14 +64,12 @@ type AlloyDBClusterSpec struct {
 	// +kcc:proto:field=google.cloud.alloydb.v1beta.Cluster.display_name
 	DisplayName *string `json:"displayName,omitempty"`
 
-	/* NOTYET
 	// Optional. The database engine major version. This is an optional field and
 	//  it is populated at the Cluster creation time. If a database version is not
 	//  supplied at cluster creation time, then a default database version will
 	//  be used.
```
gemmahou, line 70: Regarding comment `If a database version is not supplied at cluster creation time, then a default database version will be used.` This field seems to be an observedState field if it is not configured by the user. Shall we add it to observedState as well? Similar to `clusterType` field

gemmahou:
@maqiuyujoyce this is a FR from AlloyDB, would you mind taking a look and give final approval? Thanks!

pkg/controller/direct/alloydb/cluster_controller.go
maqiuyujoyce: Agreed. @kjasnoor0305 why does the API remain silence when the database version is updated?

maqiuyujoyce:
I think all the alloydb test cases would need update in their golden files due to the mock change. Besides, I'd want to clarify the API behavior for updating databaseVersion.

apis/alloydb/v1beta1/cluster_types.go
kjasnoor0305: Done.

pkg/controller/direct/alloydb/cluster_controller.go
kjasnoor0305: Updating the db version should trigger the major version upgrade operation which is not currently supported. My hunch is that since the field itself it mutable it is not throwing any error while update.

kjasnoor0305:
> I think all the alloydb test cases would need update in their golden files due to the mock change. Besides, I'd want to clarify the API behavior for updating databaseVersion.

Regenerated the golden files. Replied for updating db Version in comment above. As said this update behavior would be handled in a separate PR which would be to support MVU functionality.

PR #5064, opened 2025-08-26
improve label filtering function.

yuwenma:
### Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/label/label.go
+++ b/pkg/label/label.go
@@ -16,7 +16,7 @@ package label
 
 import (
 	"fmt"
-	"strings"
+	"unicode"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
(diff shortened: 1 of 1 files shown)


pkg/label/label.go
```diff
@@ -33,28 +33,54 @@ func ComputeLabels(u *unstructured.Unstructured) error {
...
+keyLoop:
 	for k, v := range labels {
-		if len(strings.Split(k, "/")) == 2 {
-			// Do not include any KRM-style labels (labels that include a prefix
-			// denoted with a '/').
-			// TODO(b/137755194): Determine long-term solution.
+		// GCP labels have strict validation rules. This function filters out any
+		// labels from Kubernetes metadata that do not conform to these rules,
+		// preventing them from being propagated to GCP.
+		// See: https://cloud.google.com/compute/docs/labeling-resources#requirements
```
gemmahou, line 59: I've been considering this, and my personal view is that we should not filter all invalid labels. If an invalid label is added by customer, the controller should still pass it to the server, allowing it to fail with an error instead of silently ignoring it. Our sanitization should be limited to KRM-style labels that are automatically added by Config Sync or by us and are known to fail GCP's label validation. But I'm open to either approach.

PR #5097, opened 2025-09-03
test: record untested reference fields

gemmahou:
### Change description

Split the untested field test into two parts: one for regular fields and another for reference fields. So we have a clear picture of any untested references.

I also made a minor naming change, replacing "missing" with "untested". We already have a separate [TestMissingRefs](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/e8d0ea82adfb220cca88464bf782a9cbc6e2fe87/tests/apichecks/crds_test.go#L44-L45) which specifically checks fields that should be references but aren't. Using "missing" for both was confusing, as the TestCRDFieldPresenceInTests refers to untested coverage on an existing reference.

Commit1: split untested references from all untested fields
Commit2: change exception file names from "missingxxx.txt" to "untestedxxx.txt", to distinguish with `missingrefs.txt`(see PR description)
Commit3: change tags in affected exception files from "[missing_xxx]" to [untested_xxx]

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


gemmahou:
/assign @cheftako 

This PR provides us the ability to identify **all** untested references. I'll use this to ensure that all references are covered in the fixture tests.

tests/apichecks/crds_test.go
```diff
@@ -581,7 +589,7 @@ func findFieldsNotCoveredByTests(t *testing.T, shouldVisitCRD func(crd *apiexten
 
 					// Only report an error if neither external nor name is set
 					if !hasExternal && !hasName {
-						errs = append(errs, fmt.Sprintf("[missing_field] crd=%s version=%v: field %q is not set; neither 'external' nor 'name' are set", crd.Name, version.Name, fieldPath))
+						untestedRefErrs = append(untestedRefErrs, fmt.Sprintf("[untested_ref_field] crd=%s version=%v: field %q is not set, neither 'external' nor 'name' is set in unstructured objects.", crd.Name, version.Name, fieldPath))
```
acpana, line 592: I don't dislike this but here's what I am thinking:

* let's first mark the new field, something like [missing_field][untested_ref_field]
* let's move the file (this will show up as a git mv)
* then we can drop the missing_field tag/ replace it everywhere.

You can do this in separate commits too! It will just make it so much easier for us to review!

gemmahou: Split changes into 3 commits. Take a look at the PR description😉

PR #5100, opened 2025-09-04
chore: BillingAccount: rename parent to parentRef

justinsb:
- **BillingAccount: rename parent to parentRef**
  

- **autogen: BillingAccount parentRef**
  

- **controllerbuilder: fix import for refs in same package**

### DIFF
--- a/apis/billing/v1alpha1/billingaccount_types.go
+++ b/apis/billing/v1alpha1/billingaccount_types.go
@@ -45,7 +45,7 @@ type BillingAccountSpec struct {
 	CurrencyCode *string `json:"currencyCode,omitempty"`
 
 	// Optional. The billing account's parent resource.
-	ParentBillingAccountRef *BillingAccountRef `json:"parent,omitempty"`
+	ParentRef *BillingAccountRef `json:"parentRef,omitempty"`
 }
 
 // BillingAccountStatus defines the config connector machine state of BillingAccount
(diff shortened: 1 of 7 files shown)


apis/billing/v1alpha1/billingaccount_types.go
```diff
@@ -45,7 +45,7 @@ type BillingAccountSpec struct {
 	CurrencyCode *string `json:"currencyCode,omitempty"`
 
 	// Optional. The billing account's parent resource.
-	ParentBillingAccountRef *BillingAccountRef `json:"parent,omitempty"`
+	ParentRef *BillingAccountRef `json:"parentRef,omitempty"`
```
yuwenma, line 48: hmmm, I thought we would prefer the ParentKind Ref (`billingAccountRef`) rather than `parentRef` for its usage simplicity: users don't need to know which parent it is, just use the kind to find the parent. Also, Apigee uses ParentKind Ref (`apigeeOrganizationRef`,  ),maybe a pattern to follow?

A different question: I just realize this parent Kind is the same as the child kind (BillingAccount). Is that right? If so, maybe`parentRef` is good choice for this specific case.

justinsb: Good point.  And yes, this is a recursive reference, so maybe parentRef is fine.  If not parentRef, then I think it should be parentBillingAccountRef

Issue #5073, opened 2025-08-28
feat: Migrate RunJob from TF-based to Direct (beta)


yuwenma:
### Checklist

- [x] I did not find a related open enhancement request.
- [x] I understand that enhancement requests filed in the GitHub repository are by default low priority.
- [x] If this request is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Describe the feature or resource

Migrate the RunJob from Terraform-based to the direct approach. Keep api unchanged.

PR #5108, opened 2025-09-05
feat: Api, mapper and fuzzer for Cloud RunJob (TF-based beta)

yuwenma:
### Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


apis/refs/v1beta1/secretmanagerrefs.go
```diff
@@ -145,6 +145,9 @@ func ResolveSecretManagerSecretVersionRef(ctx context.Context, reader client.Rea
 
 	// External should be in the "projects/*/secrets/*/versions/*" format
 	if ref.External != "" {
+		if ref.External == "latest" {
+			return ref, nil
+		}
```
yuwenma, line 150: Found this special case from the the `RunJob` CRD `versionRef` description,

apis/run/v1beta1/job_types.go
```diff
@@ -0,0 +1,800 @@
...
+	Conditions []v1alpha1.Condition `json:"conditions,omitempty"`
+
+	// ObservedGeneration is the generation of the resource that was most recently observed by the Config Connector controller. If this is equal to metadata.generation, then that means that the current reported status reflects the most recent desired state of the resource.
+	ObservedGeneration *int64 `json:"observedGeneration,omitempty"`
+
+	// A unique specifier for the RunJob resource in GCP.
+	ExternalRef *string `json:"externalRef,omitempty"`
+
+	// ObservedState is the state of the resource as most recently observed in GCP.
+	ObservedState *RunJobObservedState `json:",inline"`
```
yuwenma, line 71: This TF-based resource has all its OUTPUT-ONLY fields in `.status`. We want to keep the `RunJobObservedState` for mapper and fuzzer, so put this as inline json.

apis/run/v1beta1/job_types.go
```diff
@@ -0,0 +1,800 @@
...
+	Etag *string `json:"etag,omitempty"`
+
+	// Output only. For a deleted resource, the time after which it will be
+	// permamently deleted.
+	ExpireTime *string `json:"expireTime,omitempty"`
+
+	// Output only. Email address of the last authenticated modifier.
+	LastModifier *string `json:"lastModifier,omitempty"`
+
+	// Note: ExecutionReference in Run v2 proto uses a single object, but legacy KCC CRD uses an array
```
yuwenma, line 104: Call out for the exception for `LatestCreatedExecution`. this results in the manual change in mapper.

apis/run/v1beta1/job_types.go
```diff
@@ -0,0 +1,800 @@
...
+
+	// Output only. Name of the last created execution.
+	LatestCreatedExecution []*ExecutionReference `json:"latestCreatedExecution,omitempty"`
+
+	// Note: TerminalCondition in Run v2 proto uses a single object, but legacy KCC CRD uses an array
```
yuwenma, line 109: ditto

apis/run/v1beta1/job_types.go
```diff
@@ -0,0 +1,800 @@
...
+// +kcc:proto=google.cloud.run.v2.VpcAccess
+type VPCAccess struct {
+	// VPC Access connector name.
+	//  Format: `projects/{project}/locations/{location}/connectors/{connector}`,
+	//  where `{project}` can be project id or number.
+	//  For more information on sending traffic to a VPC network via a connector,
+	//  visit https://cloud.google.com/run/docs/configuring/vpc-connectors.
+	// +kcc:proto:field=google.cloud.run.v2.VpcAccess.connector
+	// +kcc:ref=VPCAccessConnector
+	ConnectorRef *vpcaccessv1beta1.VPCAccessConnectorRef `json:"connectorRef,omitempty"`
```
yuwenma, line 600: This reference is discovered and modified by Gemini via https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5109

apis/vpcaccess/v1beta1/doc.go
yuwenma, line 1: This comes from a separate Gemini task on discovering and editing the Resource Reference https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5109

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_runjobs.run.cnrm.cloud.google.com.yaml
```diff
@@ -248,24 +234,24 @@ spec:
                                               - external
                                             properties:
                                               external:
-                                                description: 'Allowed value: The `name`
-                                                  field of a `SecretManagerSecret`
-                                                  resource.'
+                                                description: ' If provided must be
```
yuwenma, line 237: This introduces an exception on `SecretmanagerSecretRef` normalization. I'd like to address separately

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_runjobs.run.cnrm.cloud.google.com.yaml
```diff
@@ -282,113 +268,121 @@ spec:
                                               - external
                                             properties:
                                               external:
-                                                description: 'Allowed value: The `version`
-                                                  field of a `SecretManagerSecretVersion`
-                                                  resource.'
+                                                description: ' If provided must be
```
yuwenma, line 271: This exception is addressed in this PR to make it backward compatible. Even if we allow the `version` value, we don't want to encourage that, so I removed the comment. Open to feedback on whether we should put it back. @justinsb

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_runjobs.run.cnrm.cloud.google.com.yaml
yuwenma, line 1: All changes are description change.

apis/refs/v1beta1/secretmanagerrefs.go
justinsb: What does it mean?  Which project?  Which secret?

apis/run/v1beta1/job_identity.go
```diff
@@ -0,0 +1,117 @@
...
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/apis/common"
+	refsv1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// JobIdentity defines the resource reference to RunJob, which "External" field
```
justinsb, line 27: Are we following the pattern we established in apigee?

apis/run/v1beta1/job_types.go
```diff
@@ -0,0 +1,800 @@
...
+type RunJobStatus struct {
+	/* Conditions represent the latest available observations of the
+	   object's current state. */
+	Conditions []v1alpha1.Condition `json:"conditions,omitempty"`
+
+	// ObservedGeneration is the generation of the resource that was most recently observed by the Config Connector controller. If this is equal to metadata.generation, then that means that the current reported status reflects the most recent desired state of the resource.
+	ObservedGeneration *int64 `json:"observedGeneration,omitempty"`
+
+	// A unique specifier for the RunJob resource in GCP.
+	ExternalRef *string `json:"externalRef,omitempty"`
```
justinsb, line 68: My 2c is that it's easier not to have status.externalRef yet (mark with NOTYET), lest we assume it is set.

apis/run/v1beta1/job_types.go
justinsb: OK ... the downside is that it makes it harder to later introduce ObservedState though

apis/run/v1beta1/job_types.go
justinsb: Would be good to fix this once we introduce observedState

apis/run/v1beta1/job_types.go
```diff
@@ -0,0 +1,800 @@
...
+	//  More info:
+	//  https://cloud.google.com/run/docs/configuring/in-memory-volumes#configure-volume.
+	//  Info in Kubernetes:
+	//  https://kubernetes.io/docs/concepts/storage/volumes/#emptydir
+	// +kcc:proto:field=google.cloud.run.v2.EmptyDirVolumeSource.size_limit
+	SizeLimit *string `json:"sizeLimit,omitempty"`
+}
+
+// +kcc:proto=google.cloud.run.v2.EnvVar
+type EnvVar struct {
```
justinsb: Why does this one need to be manually defined?

apis/run/v1beta1/job_types.go
```diff
@@ -0,0 +1,800 @@
...
+	//  execution.
+	// +kcc:proto:field=google.cloud.run.v2.ExecutionTemplate.template
+	Template *TaskTemplate `json:"template,omitempty"`
+}
+
+// +kcc:proto=google.cloud.run.v2.GCSVolumeSource
+type GCSVolumeSource struct {
+	// Cloud Storage Bucket name.
+	// +kcc:proto:field=google.cloud.run.v2.GCSVolumeSource.bucket
+	Bucket *string `json:"bucket,omitempty"`
```
justinsb: Similarly this one looks like it should not need to be manually defined.... just trying to avoid duplication if we can....

justinsb (commented):
Looks generally good, but a few questions

apis/run/v1beta1/job_identity.go
yuwenma: This is from the type-generator. I didn't comment them out since there's no compiling error. 
If we want to use the apigee format, we should update the type-generator. I can do that in a separate PR.

apis/refs/v1beta1/secretmanagerrefs.go
yuwenma: This refers to the SecretManagerSecretVersion, whose parent (SecretMangerSecret) is also required in RunJob. So it would use the same as the parent. Also, this is a [GCP server ask](https://cloud.google.com/run/docs/configuring/jobs/secrets#terraform_1), not a TF special handling (I verified in the KCC TF-based reconciler code base). I think the real problem is that this field should be `version`, but the current API requires `versionRef` and its parentRef.

apis/run/v1beta1/job_types.go
yuwenma: yes, but I commented out the new field in `RunJobObservedState`, so if we decide to duplicate fields between `.status` and `.status.observedState`, then it's not hard from implementation perspective. (Yes, it would be confusing for users, but we can fix those type mismatch, and simplify the mapper + fuzzer)

apis/run/v1beta1/job_types.go
yuwenma: yes!

apis/run/v1beta1/job_types.go
yuwenma: Sg. Just to clarify, when do you suggest we add back the externalRef? The controller PR will be the next one.

apis/run/v1beta1/job_types.go
yuwenma: Oh,  I can drop it in the manual list. My Gemini hits problems on understanding the nested subfields (because it does text comparison, rather than understanding its json structure, #5111  could help that BTW), so I observed it not only moved the changed go structures to the manual list, but their sub go structures as well.

yuwenma:
issue #5073

apis/refs/v1beta1/secretmanagerrefs.go
justinsb: OK.  I don't love that it applies to all secretVersions (not just RunJob), but I guess we can continue to support it

apis/run/v1beta1/job_identity.go
justinsb: OK!

apis/run/v1beta1/job_types.go
justinsb: When we no longer support rollback to TF

apis/run/v1beta1/job_types.go
justinsb: It's just more work for future-you, but OK

apis/vpcaccess/v1beta1/vpcaccess_connector_reference.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+		return fmt.Sprintf("projects/%s/locations/%s/connectors/%s", projectID, location, resourceID), nil
+	}
+
+	r.External = actualExternalRef
+	return r.External, nil
+}
+
+func ParseVPCAccessConnectorExternal(external string) (projectID, location, name string, err error) {
+	parts := strings.Split(external, "/")
```
justinsb, line 106: This really should be following the identity.Identity pattern

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_runjobs.run.cnrm.cloud.google.com.yaml
justinsb: Sorry - can you clarify?  This is a description change...

/hold for clarification

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_runjobs.run.cnrm.cloud.google.com.yaml
```diff
@@ -969,45 +1011,49 @@ spec:
...
               terminalCondition:
-                description: The Condition of this Job, containing its readiness status,
-                  and detailed error information in case it did not reach the desired
-                  state.
+                description: Output only. The Condition of this Job, containing its
+                  readiness status, and detailed error information in case it did
+                  not reach the desired state.
                 items:
                   properties:
-                    executionReason:
```
justinsb, line 989: What happened here?  These look like breaking changes - thankfully only in status, but still

/hold for status breaking changes

pkg/controller/direct/run/mapper.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	return out
+}
+
+func ConditionObservedState_RevisionReason_ToProto(mapCtx *direct.MapContext, in *string) *pb.Condition_RevisionReason_ {
+	if in == nil {
+		return nil
+	}
+	out := &pb.Condition_RevisionReason_{
+		RevisionReason: pb.Condition_RevisionReason(direct.StringPtrToInt32(mapCtx, in)),
```
justinsb: This isn't an enum?

/hold for possible enum

pkg/controller/direct/maputils.go
```diff
@@ -543,3 +543,19 @@ func PtrInt64ToPtrInt32(in *int64) *int32 {
 	out := int32(*in)
 	return &out
 }
+
+func StringPtrToInt32(mapCtx *MapContext, in *string) int32 {
```
justinsb: This method is suspicious, it's _probably_ an enum?

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_runjobs.run.cnrm.cloud.google.com.yaml
yuwenma: The underlying feature is not supported in the direct SecretManagerReference (irrelavant to the direct  SecretManager reconciler). Basically it requires the SecretManagerSecretRef to send a kube api call when `.external` is configured. When implementing the RunJob direct controller, I'd like to make sure the reference normalizer behaves correctly.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_runjobs.run.cnrm.cloud.google.com.yaml
yuwenma: There's no such `executionReason` in the proto, so is the other ones below. I can add them back to keep it unchanged.

yuwenma: Fixed. It turns out to be a proto interface field,  where the KCC (DCL-based) keeps every type in the status. This should resolve your other comments. All mappers are auto-generated and we have no changes to maputils.

apis/vpcaccess/v1beta1/vpcaccess_connector_reference.go
yuwenma: addressed in the next PR

pkg/controller/direct/run/mapper.go
yuwenma: Fixed. No manual mapper at all.

pkg/controller/direct/maputils.go
yuwenma: fixed. This is not needed.

apis/run/v1beta1/job_types.go
yuwenma: What I don't like about identity.Identity is that it determines the ID purely from what users said in `spec`,  Are we going to udpate/delete resource B from resource A if resource A fakes the ID?

justinsb: I don't think that is true, but I also think that identity.Identity is what we have today.

justinsb: But identity.Identity doesn't have much to do with whether you are changing the schema or not in this PR....

apis/vpcaccess/v1beta1/vpcaccess_connector_reference.go
```diff
@@ -0,0 +1,150 @@
...
+		return "", "", "", fmt.Errorf("unexpected format for external %q, expected projects/{project_id}/locations/{location}/connectors/{name}", external)
+	}
+	if parts[4] != "connectors" {
+		return "", "", "", fmt.Errorf("unexpected format for external %q, expected projects/{project_id}/locations/{location}/connectors/{name}", external)
+	}
+	return parts[1], parts[3], parts[5], nil
+}
+
+// isResourceReady checks if the given unstructured resource is ready.
+// This is a simplified check and might need to be adjusted based on the actual resource's status conditions.
```
justinsb, line 123: It looks pretty good to me!

justinsb (approved):

PR #5152, opened 2025-09-11
feat: log diff iam

acpana:
### Change description

Adds a structuredreporting.Diff call for IAMPolicy, IAMPartialPolicy, IAMPolicyMember mutations.

>  ```{"severity":"info","timestamp":"2025-09-11T00:01:46.581Z","msg":"structuredreporting OnDiff","controller":"iampartialpolicy-controller","controllerGroup":"iam.cnrm.cloud.google.com","controllerKind":"IAMPartialPolicy","IAMPartialPolicy":{"name":"iampartialpolicy-xczhv4cibizbfya","namespace":"xczhv4cibizbfya"},"namespace":"xczhv4cibizbfya","name":"iampartialpolicy-xczhv4cibizbfya","reconcileID":"044aca8c-6929-4008-858c-b30c1123efa3","diff.fields":[{"ID":"spec.bindings[role=roles/cloudsql.client]","Old":"present in desired spec","New":"absent in actual spec"}],"diff.isNewObject":false}```

```release-note
Logging for IAM policy changes.
```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/apis/iam/v1beta1/diff.go
+++ b/apis/iam/v1beta1/diff.go
@@ -22,9 +22,9 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/structuredreporting"
 )
 
-// IAMPolicySpecDiffers compares two IAMPolicySpec objects and returns true if they differ.
-// Details of the differences are added to the provided diff object.
-func IAMPolicySpecDiffers(desired, actual *IAMPolicySpec, diff *structuredreporting.Diff) bool {
+// IAMPolicySpecDiffers compares two IAMPolicySpec objects and a diff.
+func IAMPolicySpecDiffers(desired, actual *IAMPolicySpec) *structuredreporting.Diff {
+	diff := &structuredreporting.Diff{}
 
 	// Compare ResourceReference, which is immutable.
 	if desired.ResourceReference.Kind != actual.ResourceReference.Kind {
@@ -40,21 +40,19 @@ func IAMPolicySpecDiffers(desired, actual *IAMPolicySpec, diff *structuredreport
 		diff.AddField("spec.resourceRef.external", desired.ResourceReference.External, actual.ResourceReference.External)
 	}
 
-	// Compare Bindings
-	if bindingsDiffer(desired.Bindings, actual.Bindings, diff) {
-	}
+	bindingsDiff := compareBindings(desired.Bindings, actual.Bindings)
+	auditConfigDiffs := compareAuditConfigs(desired.AuditConfigs, actual.AuditConfigs)
 
-	// Compare AuditConfigs
-	if auditConfigsDiffer(desired.AuditConfigs, actual.AuditConfigs, diff) {
-	}
+	diff.AddDiff(bindingsDiff)
+	diff.AddDiff(auditConfigDiffs)
 
-	return len(diff.Fields) > 0
+	return diff
 }
 
-// bindingsDiffer compares two slices of IAMPolicyBinding.
+// compareBindings compares two slices of IAMPolicyBinding.
 // It treats the slices as maps keyed by Role, and members are treated as unordered sets.
-func bindingsDiffer(desired, actual []IAMPolicyBinding, diff *structuredreporting.Diff) bool {
-	changed := false
+func compareBindings(desired, actual []IAMPolicyBinding) *structuredreporting.Diff {
+	diff := &structuredreporting.Diff{}
 	desiredMap := make(map[string]IAMPolicyBinding)
 	for _, b := range desired {
 		sort.Slice(b.Members, func(i, j int) bool {
@@ -78,12 +76,10 @@ func bindingsDiffer(desired, actual []IAMPolicyBinding, diff *structuredreportin
 		actualBinding, ok := actualMap[role]
 		if !ok {
 			diff.AddField(fmt.Sprintf("spec.bindings[role=%v]", role), "present in desired spec", "absent in actual spec")
-			changed = true
 			continue
 		}
 		if !reflect.DeepEqual(desiredBinding, actualBinding) {
 			diff.AddField(fmt.Sprintf("spec.bindings[role=%v]", role), desiredBinding, actualBinding)
-			changed = true
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


apis/iam/v1beta1/diff.go
```diff
@@ -0,0 +1,164 @@
...
+
+// IAMPolicySpecDiffers compares two IAMPolicySpec objects and returns true if they differ.
+// Details of the differences are added to the provided diff object.
+func IAMPolicySpecDiffers(desired, actual *IAMPolicySpec, diff *structuredreporting.Diff) bool {
+	changed := false
+
+	// Compare ResourceReference, which is immutable.
+	if desired.ResourceReference.Kind != actual.ResourceReference.Kind {
+		diff.AddField("spec.resourceRef.kind", desired.ResourceReference.Kind, actual.ResourceReference.Kind)
+		changed = true
```
cheftako: We never switch back to false. We also only return a boolean. No point in continuing on at this point. Just return true here. Ditto everywhere else you set changed to true,

acpana: Wait if we do return true right away we will miss any subsequent field updates.

I think the idea of the diff is to capture ALL fields that are changed

apis/iam/v1beta1/diff.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+	if desired.ResourceReference.Name != actual.ResourceReference.Name {
+		diff.AddField("spec.resourceRef.name", desired.ResourceReference.Name, actual.ResourceReference.Name)
+	}
+	if desired.ResourceReference.External != actual.ResourceReference.External {
+		diff.AddField("spec.resourceRef.external", desired.ResourceReference.External, actual.ResourceReference.External)
+	}
+
+	// Compare Bindings
+	if bindingsDiffer(desired.Bindings, actual.Bindings, diff) {
```
cheftako: Please no empty if statements.

apis/iam/v1beta1/diff.go
```diff
@@ -0,0 +1,157 @@
...
+	// Compare AuditConfigs
+	if auditConfigsDiffer(desired.AuditConfigs, actual.AuditConfigs, diff) {
+	}
+
+	return len(diff.Fields) > 0
+}
+
+// bindingsDiffer compares two slices of IAMPolicyBinding.
+// It treats the slices as maps keyed by Role, and members are treated as unordered sets.
+func bindingsDiffer(desired, actual []IAMPolicyBinding, diff *structuredreporting.Diff) bool {
```
cheftako: Maybe computeBindingDifferences()? Probably return the diff?

apis/iam/v1beta1/diff.go
```diff
@@ -0,0 +1,157 @@
...
+			diff.AddField(fmt.Sprintf("spec.bindings[role=%v]", role), "absent in desired spec", "present in actual spec")
+			changed = true
+		}
+	}
+	return changed
+}
+
+// auditConfigsDiffer compares two slices of IAMPolicyAuditConfig.
+// It treats the slices as maps keyed by Service and canonicalizes inner slices for stable comparison.
+func auditConfigsDiffer(desired, actual []IAMPolicyAuditConfig, diff *structuredreporting.Diff) bool {
```
cheftako: Ditto

apis/iam/v1beta1/diff.go
```diff
@@ -0,0 +1,157 @@
...
+	changed := false
+	// Helper function to canonicalize the inner slices for a stable comparison.
+	canonicalize := func(configs []IAMPolicyAuditConfig) {
+		// Create a deep copy to avoid modifying the original struct.
+		copiedConfigs := make([]IAMPolicyAuditConfig, len(configs))
```
cheftako, line 104: The comment does not match the code. Not understanding the point of making an array and (shallowish) copying into it. Seems like we can just order the input at that point.

cheftako: https://go.dev/play/p/N5slMS4wfok?v=goprev

apis/iam/v1beta1/diff.go
```diff
@@ -0,0 +1,157 @@
...
+
+import (
+	"fmt"
+	"reflect"
+	"sort"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/structuredreporting"
+)
+
+// IAMPolicySpecDiffers compares two IAMPolicySpec objects and returns true if they differ.
```
cheftako: Return Values Directly: Go's multiple return values are a powerful feature. Prefer returning values directly rather than using "out parameters" (passing pointers to be modified). If you do not need values from an input parameter then it should probably be a return value.

I'm talking about the diff

apis/iam/v1beta1/diff.go
acpana: Looking around the callsite for the diff for this function, I think we should keep the idea of deep copying the structure and sorting it, even if we discard it. I wouldn't want the sorting to cause diffs in status reporting or something else down the road.

I would be game to sort these as part of the reconciler loop (if we don't do this already).

wdyt? updated the code.

apis/iam/v1beta1/diff.go
acpana: I added an `AddDiff` function!

acpana: I've further refactored this section so that we don't write in the parameter values. Will file a GitHub issue for any future refactoring that has us using this pattern in the remainder of the code base!

acpana: https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/5419

pkg/controller/iam/iamclient/tfiamclient.go
```diff
@@ -78,6 +80,29 @@ func (t *TFIAMClient) SetPolicyMember(ctx context.Context, policyMember *v1beta1
 		logger.Info("underlying resource is already up to date", "resource", k8s.GetNamespacedName(policyMember))
 		return policyMember, nil
 	}
+
+	// Report diff to structured-reporting subsystem
+	{
+		report := &structuredreporting.Diff{}
+		u, err := resource.MarshalAsUnstructured()
+		if err != nil {
+			log := log.FromContext(ctx)
```
cheftako, line 89: Follow on. Please make logging consistent in this file. Probably means this call needs to be moved to an outer scope.

pkg/structuredreporting/diffs.go
```diff
@@ -44,9 +44,26 @@ func (d *Diff) AddField(id string, old any, new any) {
...
 	}
 }
+
+func (d *Diff) AddDiff(other *Diff) *Diff {
+	if other == nil {
+		return d
+	}
+
+	for _, f := range other.Fields {
+		d.AddField(f.ID, f.Old, f.New)
```
cheftako, line 65: Dedupe collistions

apis/iam/v1beta1/diff.go
```diff
@@ -0,0 +1,158 @@
...
+
+		return copiedConfigs
+	}
+
+	canonicalizedDesired := canonicalize(desired)
+	canonicalizedActual := canonicalize(actual)
+
+	desiredMap := make(map[string]IAMPolicyAuditConfig)
+	for _, ac := range canonicalizedDesired {
+		desiredMap[ac.Service] = ac
```
cheftako, line 131: Non blocking. Might be clearer if we just canoncalized ac before putting it in the map.

PR #5158, opened 2025-09-12
feat: opt in IAMPartialPolicy

acpana:
### Change description

Users can opt in for the `direct` IAMPartialPolicy controller which uses less quota today! In the future, we will make further optimizations to this controller and default it as the only controller for IAMPartialPolicy.

```release-note
Users can now opt-in to use a new, simpler direct controller for IAMPartialPolicy resources.
```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/controller/parent/controller.go
+++ b/pkg/controller/parent/controller.go
@@ -27,7 +27,6 @@ import (
 	dclcontroller "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/dcl"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
-	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/iam/partialpolicy"
 	kccpredicate "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/predicate"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/ratelimiter"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/resourceconfig"
@@ -54,13 +53,12 @@ type Reconcilers struct {
 	TF     *tf.Reconciler
 	DCL    *dclcontroller.Reconciler
 	Direct *directbase.DirectReconciler
+	Custom *CustomReconciler
+}
 
-	/*
-	 TODO KCC-team:
-	 - while we migrate the handrolled/ legacy IAM controllers to direct
-	 we will have these standout types here.
-	*/
-	PartialPolicy *partialpolicy.ReconcileIAMPartialPolicy
+type CustomReconciler struct {
+	Type       k8s.ReconcilerType
+	Reconciler reconcile.Reconciler
 }
 
 // ParentReconciler is a top-level controller that decides which underlying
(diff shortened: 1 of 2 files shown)


pkg/controller/direct/iam/mappings.go
```diff
@@ -40,7 +40,7 @@ func IAMPolicySpec_ToProto(_ *direct.MapContext, in *krm.IAMPolicySpec) *iampb.P
 		// Default to version 3, which supports conditions.
 		// IAM Server should downgrade to supported versions
 		Version: 3,
-		//Etag: spec.Etag, NOT YET
+		Etag:    []byte(in.Etag),
```
yuwenma: We may not want the etag in IAM `spec`.

acpana: Unfortunately, the legacy controller does preserver/ set the etag:

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/iam/partialpolicy/iampartialpolicy_controller.go#L527

I think we might have to keep it here for backwards compat right ?

cheftako: The good news is that etag is not currently in the CRD for IAMPartialPolicy or in the official documentation for [IAMPartialPolicy](https://cloud.google.com/config-connector/docs/reference/resource-docs/iam/iampartialpolicy). So we have the option to not add it to the CRD.

pkg/controller/direct/iam/partialpolicy_controller.go
```diff
@@ -195,8 +195,10 @@ func (a *IAMPartialPolicyAdapter) Create(ctx context.Context, createOp *directba
 	if a.actualReferencedResourcePolicy == nil {
 		livePolicyForMerge = &krm.IAMPolicy{}
 	} else {
-		// todo acpana round trip ?
-		livePolicyForMerge = ToNewIAMPolicySkeleton(a.desired)
+		livePolicyForMerge = ToIAMPolicySkeleton(a.desired)
+
+		// this is not the PartialPolicy etag on GCP but the etag on the IAMPolicy on GCP
+		livePolicyForMerge.Spec.Etag = string(a.actualReferencedResourcePolicy.Etag)
```
cheftako: Fire

pkg/controller/parent/controller.go
```diff
@@ -53,6 +54,13 @@ type Reconcilers struct {
 	TF     *tf.Reconciler
 	DCL    *dclcontroller.Reconciler
 	Direct *directbase.DirectReconciler
+
+	/*
+	 TODO KCC-team:
+	 - while we migrate the handrolled/ legacy IAM controllers to direct
+	 we will have these standout types here.
+	*/
+	PartialPolicy *partialpolicy.ReconcileIAMPartialPolicy
```
cheftako: Can we use a parent type that is less type specific?

pkg/controller/registration/registration_controller.go
```diff
@@ -262,10 +269,11 @@ func registerDefaultController(ctx context.Context, r *ReconcileRegistration, co
 		if err := policy.Add(r.mgr, &cds); err != nil {
 			return nil, err
 		}
-	case "IAMPartialPolicy":
-		if err := partialpolicy.Add(r.mgr, &cds); err != nil {
-			return nil, err
-		}
+	// slowly moving over decision making for routing outside of this control block.
```
cheftako: Please delete this commented out block. Non blocking.

pkg/controller/parent/controller.go
```diff
@@ -53,6 +53,7 @@ type Reconcilers struct {
 	TF     *tf.Reconciler
 	DCL    *dclcontroller.Reconciler
 	Direct *directbase.DirectReconciler
+	Custom map[k8s.ReconcilerType]reconcile.Reconciler
```
cheftako: Do we plan on supporting a single resource kind having more than 1 custom reconciler? If not then this does not need to be a map and various other logic gets simpler and more standard. Eg. the switch below has a custom case and you don't need special logic in the default block.

acpana: wrapped in a simple struct!

pkg/controller/parent/controller.go
```diff
@@ -76,9 +82,10 @@ func Add(mgr manager.Manager, gvk schema.GroupVersionKind, tf *tf.Reconciler, dc
 		mgr:    mgr,
 		gvk:    gvk,
 		reconcilers: Reconcilers{
```
cheftako, line 84: If we already have a Reconcilers object, is there a reason to copy of its fields rather than just the object itself?

PR #5164, opened 2025-09-12
feat: apply cookie to sqlinstance

yuwenma:
### Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/sqlinstance-postgres-minimal-direct/_http.log
```diff
@@ -258,7 +258,9 @@ PATCH https://sqladmin.googleapis.com/sql/v1beta4/projects/${projectId}/instance
 Content-Type: application/json
 User-Agent: kcc/${kccVersion} (+https://github.com/GoogleCloudPlatform/k8s-config-connector) kcc/controller-manager/${kccVersion}
 
-{}
+{
+  "maintenanceVersion": "POSTGRES_16_9.R20250302.00_19"
+}
```
yuwenma, line 263: After investigation, it looks like we catch a bug in the previous sql controller. 
This fixture test has [`maintenanceVersion` sets in `create.yaml`](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/sqlinstance-postgres-minimal-direct/create.yaml#L24) as this value, and split the Create call to 1. create without maintenanceVersion, 2. patch the maintenanceVersion if specified. The previous controller does not make the right patch call.

pkg/controller/direct/sql/client.go
```diff
@@ -56,3 +57,17 @@ func (m *gcpClient) sqlInstancesClient() *api.InstancesService {
...
+	var errs []string
+	if opErr != nil {
+		for _, err := range opErr.Errors {
+			errs = append(errs, fmt.Sprintf("code=%q, message=%q", err.Code, err.Message))
+		}
+	}
+	if len(errs) == 0 {
+		return fmt.Errorf("gcp operation failed with unknown error")
+	}
+	return fmt.Errorf("gcp operation failed: %s", strings.Join(errs, "; "))
```
justinsb: I think you can do `...: %w", errors.Join(errs...))`

pkg/controller/direct/sql/client.go
```diff
@@ -56,3 +57,17 @@ func (m *gcpClient) sqlInstancesClient() *api.InstancesService {
...
+// NewGCPOperationError builds an error from a GCP operation error.
+func NewGCPOperationError(opErr *api.OperationErrors) error {
+	var errs []string
+	if opErr != nil {
+		for _, err := range opErr.Errors {
+			errs = append(errs, fmt.Sprintf("code=%q, message=%q", err.Code, err.Message))
+		}
+	}
+	if len(errs) == 0 {
+		return fmt.Errorf("gcp operation failed with unknown error")
```
justinsb: Is api.OperationErrors an error - i.e. could we do `gcp operation failed: %w", opErr)`

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -115,7 +118,10 @@ func (a *sqlInstanceAdapter) Find(ctx context.Context) (bool, error) {
 
 	instance, err := a.sqlInstancesClient.Get(a.projectID, a.resourceID).Context(ctx).Do()
 	if err != nil {
-		return false, nil
+		if direct.IsNotFound(err) {
```
justinsb, line 121: Nice catch (maybe have gemini look for more instances like this, though I don't know how well it will do)

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -237,28 +222,74 @@ func (a *sqlInstanceAdapter) insertInstance(ctx context.Context, u *unstructured
...
 			}
 		}
 	}
 
 	log.V(2).Info("instance created", "op", op, "instance", created)
 
-	status, err := SQLInstanceStatusGCPToKRM(created)
-	if err != nil {
-		return fmt.Errorf("updating SQLInstance status failed: %w", err)
+	if maintenanceVersion != "" {
```
justinsb, line 236: So I'm glad you caught this, but can we split it out?  (Either do it separately first or do it in a follow on PR?). It just makes it hard to follow.  How did you find it?

It does look correct so I can probably just deal with it, but I need to understand why it "worked" before (or at least why our tests didn't catch it...), so if I could think about that separately that might make my life as a reviewer easier

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -237,28 +222,74 @@ func (a *sqlInstanceAdapter) insertInstance(ctx context.Context, u *unstructured
...
+		newMaintDb := &api.DatabaseInstance{
```
justinsb: Nit: updateMaintenanceVersion or something like that?  Just to establish a better pattern for the AI to follow...

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -237,28 +222,74 @@ func (a *sqlInstanceAdapter) insertInstance(ctx context.Context, u *unstructured
...
+		log.V(2).Info("spec has not changed, entering fast-path drift detection")
+		currentGCPHash, err := hashGCPObject(a.actual)
+		if err != nil {
+			return fmt.Errorf("hashing actual GCP object: %w", err)
+		}
+		if currentGCPHash == lastAppliedGCPHash {
+			log.V(2).Info("resource is up-to-date")
+			// Even though we are not making changes, we should update the status
+			// to reflect the latest observed state from the API.
+			return a.updateStatus(u, a.actual)
```
justinsb, line 298: Nice!

BTW I think we've been asked to update the status _before_ starting operations also, for visibility, although I think that would be a different PR

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -454,6 +481,9 @@ func (a *sqlInstanceAdapter) pollForLROCompletion(ctx context.Context, op *api.O
 		log.V(2).Info("polling", "op", op)
 
 		if op.Status == "DONE" {
+			if op.Error != nil {
+				return NewGCPOperationError(op.Error)
```
justinsb, line 481: Ah, I guess it doesn't implement error; you could probably still print it though when it is "unknown"

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -468,19 +498,92 @@ func (a *sqlInstanceAdapter) pollForLROCompletion(ctx context.Context, op *api.O
...
+	if err != nil {
+		return fmt.Errorf("hashing spec: %w", err)
+	}
+	gcpHash, err := hashGCPObject(gcpObj)
+	if err != nil {
+		return fmt.Errorf("hashing GCP object: %w", err)
+	}
+
+	cookie := fmt.Sprintf("%s/%s", specHash, gcpHash)
+	status.LastModifiedCookie = &cookie
```
justinsb, line 513: I feel like these lines could be a helper function, and you can just use the same helper function when comparing...

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -237,28 +222,74 @@ func (a *sqlInstanceAdapter) insertInstance(ctx context.Context, u *unstructured
...
+
+	return created, nil
 }
 
 func (a *sqlInstanceAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOperation) error {
 	u := updateOp.GetUnstructured()
-
 	log := klog.FromContext(ctx)
+
+	currentSpecHash, err := hashSpec(&a.desired.Spec)
```
justinsb, line 273: I don't love that we are hashing the KRM spec, because it feels like we're going to mess up by ignoring a label or annotation, but ... not a blocker

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -468,19 +498,92 @@ func (a *sqlInstanceAdapter) pollForLROCompletion(ctx context.Context, op *api.O
...
-	u.Object["status"] = status
+	u.Object["status"] = unstructuredStatus
 
 	return nil
 }
+
+func hashString(s string) string {
+	h := sha256.New()
+	h.Write([]byte(s))
+	return hex.EncodeToString(h.Sum(nil))
```
justinsb: FYI because we don't need crypto-levels of collision avoidance, you can probably truncate this to just the first (say) 16 characters.

pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/sqlinstance-postgres-minimal-direct/_http.log
justinsb: Does that mean that `dev/tasks/download-e2e-logs` shows this test as failing?

justinsb (commented):
Looks really good 👍

pkg/controller/direct/sql/sqlinstance_controller.go
yuwenma: Agree. was planning to do that in the next controller.

pkg/controller/direct/sql/sqlinstance_controller.go
yuwenma: Yes. I'll do that in a separate PR.

pkg/controller/direct/sql/sqlinstance_controller.go
yuwenma: Done.

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -139,95 +144,75 @@ func (a *sqlInstanceAdapter) Create(ctx context.Context, createOp *directbase.Cr
...
 
 	log.V(2).Info("instance cloned", "op", op, "instance", created)
+	return created, nil
+}
 
-	status, err := SQLInstanceStatusGCPToKRM(created)
-	if err != nil {
-		return fmt.Errorf("updating SQLInstance status failed: %w", err)
+func (a *sqlInstanceAdapter) insertInstance(ctx context.Context, u *unstructured.Unstructured, log klog.Logger) (*api.DatabaseInstance, error) {
+	// if the maitenance version is provided on CREATE, we need to break up the operation between
```
justinsb, line 187: What happens if we do specify it?  Is it ignored or is it an error?  Trying to think about what is the most fluent way to handle this (and others like it)

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -237,28 +222,82 @@ func (a *sqlInstanceAdapter) insertInstance(ctx context.Context, u *unstructured
...
+			return nil, err
+		}
+		created = updated
+	}
+
+	return created, nil
+}
+
+func (a *sqlInstanceAdapter) updateMaintenanceVersion(ctx context.Context, maintenanceVersion string, log klog.Logger) (*api.DatabaseInstance, error) {
+	newMaintDb := &api.DatabaseInstance{
```
justinsb, line 249: I always prefer to pass the ctx in, and then do `log := klog.FromContext(ctx)`

pkg/controller/direct/sql/sqlinstance_controller.go
justinsb: Why not now?  If we are going to experiment in a controller, I don't think it should be SQLInstance

pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/sqlinstance-postgres-minimal-direct/_http.log
justinsb: Ping on this

pkg/controller/direct/sql/sqlinstance_controller.go
yuwenma: "400: Invalid request: Maintenance version (POSTGRES_16_9.R20250302.00_19) must not be set" 

Yeah, if it is no-op we probably could pass it through

pkg/controller/direct/sql/sqlinstance_controller.go
yuwenma: Sure, done.

pkg/test/resourcefixture/testdata/basic/sql/v1beta1/sqlinstance/sqlinstance-postgres-minimal-direct/_http.log
yuwenma: dev/tasks/download-e2e-logs can't catch it, because it receives the same  empty payload (log match)

Issue #5191, opened 2025-09-17
Unify the direct Compute API, Mapper and Fuzzer


yuwenma:
The [Compute direct controllers](https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/master/pkg/controller/direct/compute) are structured in its own directory, which confuses the mapper-generator tool that places all auto-generated code in a single mapper.generated.go file. 

This blocks us migrating and promoting other Compute resources.

Actions

- [x] Add generate.sh to ./apis/compute/v1beta1
- [x] Flatten the subdirectories FirewallPolicyRule, ForwardingRule and TargetTCPProxy. 
- [x] Fix the 3 resources mapper functions
- [x] Fix the 3 resources fuzzer functions

PR #5193, opened 2025-09-17
chore: Unify ComputeTargetTCPProxy direct API and controller

gemmahou:
### Change description

Address #5191 for ComputeTargetTCPProxy

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/apis/compute/v1beta1/targettcpproxy_types.go
+++ b/apis/compute/v1beta1/targettcpproxy_types.go
@@ -33,23 +33,25 @@ var (
 type ComputeTargetTCPProxySpec struct {
 	// A reference to the ComputeBackendService resource.
 	// +required
+	// +kcc:proto:field=google.cloud.compute.v1.TargetTcpProxy.service
 	BackendServiceRef *ComputeBackendServiceRef `json:"backendServiceRef"`
 
-	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Description is immutable"
 	// Immutable. An optional description of this resource.
+	// +kcc:proto:field=google.cloud.compute.v1.TargetTcpProxy.description
 	Description *string `json:"description,omitempty"`
 
 	// The geographical location of the ComputeTargetTCPProxy.
 	// Reference: GCP definition of regions/zones (https://cloud.google.com/compute/docs/regions-zones/)
 	Location *string `json:"location,omitempty"`
 
-	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="ProxyBind is immutable"
 	// Immutable. This field only applies when the forwarding rule that references
 	// this target proxy has a loadBalancingScheme set to INTERNAL_SELF_MANAGED.
+	// +kcc:proto:field=google.cloud.compute.v1.TargetTcpProxy.proxy_bind
 	ProxyBind *bool `json:"proxyBind,omitempty"`
 
 	// Specifies the type of proxy header to append before sending data to
 	// the backend. Default value: "NONE" Possible values: ["NONE", "PROXY_V1"].
+	// +kcc:proto:field=google.cloud.compute.v1.TargetTcpProxy.proxy_header
 	ProxyHeader *string `json:"proxyHeader,omitempty"`
 
 	// The ComputeTargetTCPProxy name. If not given, the metadata.name will be used.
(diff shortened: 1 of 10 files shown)


apis/compute/v1beta1/targettcpproxy_types.go
```diff
@@ -31,25 +31,26 @@ var (
 // ComputeTargetTCPProxySpec defines the desired state of ComputeTargetTCPProxy
 // +kcc:spec:proto=google.cloud.compute.v1.TargetTcpProxy
 type ComputeTargetTCPProxySpec struct {
-	// A reference to the ComputeBackendService resource.
+	// URL to the BackendService resource.
```
yuwenma: nit: I think the previous comments are more correct

gemmahou: Changed back!

pkg/controller/direct/compute/targettcpproxy_mapper.go
yuwenma, line 1: Can you double check if the latest mapper generator can generate the functions in this manual file? If so, we want to move them to mapper.generated.go, otherwise we want to fix the mapper generator (if possible).

apis/compute/v1beta1/targettcpproxy_types.go
yuwenma, line 1: Suggest not making API changes. This is more than unifying api and controller

pkg/controller/direct/compute/targettcpproxy_mapper.go
gemmahou: The mapper generator works but it cannot handle the edge cases specific to this resource:
```
out.Location = in.Region
if in.Service != nil {
	out.BackendServiceRef = &krm.ComputeBackendServiceRef{External: direct.ValueOf(in.Service)}
}
```
map `location` to `region` and map `service` to `backendService`. Since it's not general enough I don't think we want to add additional logic to the generator for now, so I just moved them to its own mappers file. WDYT?

apis/compute/v1beta1/targettcpproxy_types.go
gemmahou: All of these are updated descriptions, copied from the generated_types to ensure they are up-to-date with the proto. Is this change acceptable, or should we revert it?

gemmahou: Discussed offline, I removed the description changes. We could followup in another PR if needed.

apis/compute/v1beta1/targettcpproxy_types.go
```diff
@@ -68,12 +71,15 @@ type ComputeTargetTCPProxyStatus struct {
...
 	// Creation timestamp in RFC3339 text format.
+	// +kcc:proto:field=google.cloud.compute.v1.TargetTcpProxy.creation_timestamp
 	CreationTimestamp *string `json:"creationTimestamp,omitempty"`
 
 	// The unique identifier for the resource.
+	// +kcc:proto:field=google.cloud.compute.v1.TargetTcpProxy.id
 	ProxyId *int64 `json:"proxyId,omitempty"`
 
 	// The SelfLink for the resource.
+	// +kcc:proto:field=google.cloud.compute.v1.TargetTcpProxy.self_link
```
yuwenma, line 82: Do you mind reminding me what these inline comments come from?

pkg/controller/direct/compute/targettcpproxy_mapper.go
yuwenma: How about ComputeTargetTCPProxyStatus_ToProto and  ComputeTargetTCPProxyStatus_FromProto?

yuwenma (commented):
One question about the manual mapper, otherwies looks good.

apis/compute/v1beta1/targettcpproxy_types.go
gemmahou: These are from the generated code, I'm thinking it's better to copy over here for consistency. These won't be included in the CRD.

pkg/controller/direct/compute/targettcpproxy_mapper.go
gemmahou: It contains proxyId to Id, also an edge case that the field names do not match.
```
ProxyId = direct.LazyPtr(int64(in.GetId()))
```

yuwenma:
/approve
/lgtm

Looks great!

PR #5196, opened 2025-09-17
refactor: add status last actuated field

acpana:
### Change description

Adds a status field ONLY for IAMPartialPolicy for now to see when the last time that a direct or legacy controller touched the controllers!

In this transitory period (`handrolled/ dcl/ tf -> direct`), I think it would be helpful to know when a direct controller reconciles a resources and another one reconciles the same resource. It will help spot any controller contention or misconfigurations from the practitioners.

```release-note
Add status fields for the last update time of a direct or legacy controller for the IAMPartialPolicy.
```

```docs

```

#### Intended Milestone

### Tests you have done


apis/iam/v1beta1/partialpolicy_types.go
```diff
@@ -99,6 +99,12 @@ type IAMPartialPolicyStatus struct {
 	LastAppliedBindings []IAMPolicyBinding `json:"lastAppliedBindings,omitempty"`
 	// AllBindings surfaces all IAM bindings for the referenced resource.
 	AllBindings []IAMPolicyBinding `json:"allBindings,omitempty"`
+	// LastDirectControllerActuation is the timestamp when the direct controller last actuated this resource.
+	// +optional
+	LastDirectControllerActuation string `json:"lastDirectControllerActuation,omitempty"`
```
jingyih, line 104: Potentially if we have a status field that indicates the controller type, we could use it to filter IAM resources by controller type. Not sure if the filtering happens on client side or server side, but the kubebuilder annotation is:

```
// +kubebuilder:selectablefield
```

So maybe an alternative API could be
```
  controllerType string
  LastRecnocileTime string
```

PR #5197, opened 2025-09-17
chore: add a script to to generate controller config

xiaoweim:
### Change description

Add a script to to generate controller config.

The logic of the script:
   1. Supported Controllers:
       * A resource supports Direct if a direct controller exists.
       * A resource supports DCL if it has the dcl2crd label.
       * A resource supports Terraform by default, unless it is a DCL-only resource.

   2. Default Controller:
       * The default is DCL if the dcl2crd label is true.
       * Otherwise, the default is Terraform if the tf2crd label is true.
       * Otherwise, the default is Direct if a direct controller exists.
       * As a fallback, the default is Terraform.

```release-note
NONE
```

```docs

```

#### Intended Milestone

### Tests you have done


dev/tasks/generate_static_config.py
acpana, line 1: Can we use the generated_gvks first instead of this script?

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/b50962f151188e010ff10ee87a449a6eb08cb1ea/pkg/gvks/supportedgvks/gvks_generated.go#L32-L33

& assume that if no DCL or TF, we have `direct` only?

xiaoweim: We also need supported controllers with migrated beta resources, so that customers can override the default controller. I don't think `generated_gvks` supports that.

PR #5201, opened 2025-09-18
WIP: chore: refactor ComputeNetworkRef

gemmahou:
### Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


yuwenma (commented):
/assign @cheftako

apis/common/utils.go
acpana, line 1: nit: you can add a unit test for `FixStaleComputeExternalFormat`

apis/compute/v1beta1/network_reference.go
```diff
@@ -0,0 +1,192 @@
...
+	if err := id.ConvertToProjectNumber(ctx, projectMapper); err != nil {
+		return err
+	}
+
+	ref.External = id.String()
+	return nil
+}
+
+// ConvertClientToProjectNumber converts the external reference to use a project number.
+func (ref *ComputeNetworkRef) ConvertClientToProjectNumber(ctx context.Context, projectsClient *resourcemanager.ProjectsClient) error {
```
acpana, line 163: nit: i'm fine to have these live here but it feels more like a apis/common thing?

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_alloydbclusters.alloydb.cnrm.cloud.google.com.yaml
```diff
@@ -433,14 +433,14 @@ spec:
                       - external
                     properties:
                       external:
-                        description: A reference to an externally-managed Compute
-                          Network resource. Should be in the format `projects/{{projectID}}/global/networks/{{network}}`.
+                        description: The value of an externally managed ComputeNetwork
+                          resource.
```
acpana: nit: I am not toooo passionate about the description churn here but I think we could make the types' comments match the old ones closer if we wanted to!

acpana (commented):
I think this LGTM overall! I left some nits that I'd address only if Walter/ Yuwen agree.

While this is somewhat of a straightforward refactor, I'd keep an eye on a few dynamic/ integration post submits that reference ComputeNetwork as this merges. You can even run one or two of those tests with these changes to gain extra confidence!

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_alloydbclusters.alloydb.cnrm.cloud.google.com.yaml
gemmahou: I can add the format suggestion back! Both partial and full url are accepted.

apis/common/utils.go
gemmahou: This is already tested by "external is full url" and "name specified, with status.selfLink", as the expected result is trimmed/fixed. (but only when the version is v1. I could add test for v1beta1 and beta)

apis/compute/v1beta1/network_reference.go
gemmahou: This is currently only used for Compute. I want to minimize the code changes in this PR, we could make this a common method later, as it is used broadly.

apis/compute/v1beta1/network_reference.go
```diff
@@ -0,0 +1,193 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ refsv1beta1.Ref = &ComputeNetworkRef{}
+var ComputeNetworkGVK = GroupVersion.WithKind("ComputeNetwork")
+
+type ComputeNetworkRef struct {
```
cheftako, line 41: Observation: No action yet. I suspect this will eventually cause a circular dependency. As we see many things now have a dependency on compute/v1beta. Wondering if we end up with something under compute/v1beta having a refernce (and hence a dependency) back on to one of the resources.

cheftako, line 41: If we are moving it out of apis/refs/v1beta1, then we should delete it from apis/refs/v1beta1. Otherwise we cannot be sure the moce is complete and we will end up having to maintain it in both places.

gemmahou: Addressed in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5201/commits/426fb133ceb98acfe28473c52c2892668242fbf4

gemmahou: Chatted with Yuwen about this offline, we haven't seen the dependency issue yet, so we will hold off on addressing it until an instance occurs.

apis/common/utils.go
```diff
@@ -29,3 +31,13 @@ func LazyPtr[V comparable](v V) *V {
 	}
 	return &v
 }
+
+// FixStaleComputeExternalFormat converts the "External" reference field to the right format if a SelfLink value is used.
+// This guarantees the backward compatibility for Compute Beta resources.
+func FixStaleComputeExternalFormat(external string) string {
```
cheftako: This should really have associated unit tests. Also its a little odd we are trying to blindly remove assorted strings from the start. At the same time we should probably be a little smarter about it. Eg if we get a string like  "https://www.googleapis.com/compute/alpha/" we will do nothing. We could do something like ignore anything which does not start https://. Anything which starts that way but is not truncated we could log as problematic.

gemmahou: Please take a look at the updated code in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5201/commits/bd0f88d97778fef64b074cac79694949bd9df984

While I'm not 100% sure if we should completely avoid versions like 'alpha' or 'v2', we may need to update this function if the Compute service releases new versions(v2). But I won't worry too much about the risk because this is a temporary fix for Compute backward compatibility. We no longer need this after we finish the direct controller migration.

PR #5208, opened 2025-09-19
alpha resource with no supported controller

xiaoweim:
### Change description

Add a script that generate the list of alpha resources that does not have any controllers.

```release-note
NONE
```

```docs

```

#### Intended Milestone

### Tests you have done


experiments/promoter/unsupported-resources/find_uncontrolled_resources.sh
yuwenma, line 1: LGTM. Can we move the bash and repot to ./experiments/promoter? It would be clearer that we eventually will add the coverage for them.

PR #5226, opened 2025-09-22
fix: do not swallow errors when reading Secret

justinsb:
When reading the Secret, we handled NotFound, but swallowed other errors.

We think this can lead to over-eager certificate rotation.


yuwenma (commented):
/lgtm
/approve

Sg

pkg/webhook/cert/writer/secret.go
```diff
@@ -125,8 +126,13 @@ func (s *secretCertWriter) read() (*generator.Artifacts, error) {
 		},
 	}
 	err := s.Client.Get(context.Background(), *s.Secret, secret)
-	if apierrors.IsNotFound(err) {
-		return nil, notFoundError{err}
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("secret not found", "secret", *s.Secret)
```
cheftako, line 131: If secret is not found then not sure what we expect to learn by printing it. In general I don't thing get methods tend to populate the destination if we got an error. Also uncomfortable with setting a trend of printing a secret. Its PII. Can we at least create a wrapper method which blanks out the data field so we are fairly sure we're not putting PII in the logs?

justinsb: Yeah I debated here, but the normal case is that this secret exists so this won't be printed often.

`*s.Secret` is the types.NamespacedName of the secret, not the secret itself.  I started cleaning that up too, but really I think it's better to just move this logic to the operator, and keep these hotfix PRs easy to cherry-pick

cheftako: Ahh missed that it was the secret ref and not the object itself.

PR #5227, opened 2025-09-22
chore: more verbose logging during certificate validation

justinsb:
If the certificate is not valid, log the reason why not.

### DIFF
--- a/pkg/webhook/cert/generator/util.go
+++ b/pkg/webhook/cert/generator/util.go
@@ -17,10 +17,14 @@ limitations under the License.
 package generator
 
 import (
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/pem"
+	"fmt"
 	"time"
+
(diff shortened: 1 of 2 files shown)


pkg/webhook/cert/generator/util.go
```diff
@@ -29,26 +33,33 @@ import (
 // - cert is for dnsName
 // - cert won't expire before time
 func ValidCACert(key, cert, caCert []byte, dnsName string, time time.Time) bool {
+	ctx := context.TODO()
+	log := klog.FromContext(ctx)
 	if len(key) == 0 || len(cert) == 0 || len(caCert) == 0 {
+		log.Error(fmt.Errorf("empty key, cert or caCert"), "key, cert or caCert is empty")
```
cheftako, line 39: rather than mostly repeating ourselves in the string, maybe indicate that we are "returning the CACert as invalid" or some such?

PR #5232, opened 2025-09-22
chore: create script to enable cherry-picking

justinsb:


dev/tasks/cherry-pick
```diff
@@ -0,0 +1,63 @@
...
+
+
+def main():
+    if len(sys.argv) != 3:
+        print("Usage: dev/tasks/cherry-pick <release-branch> <pr-number>")
+        sys.exit(1)
+    release_branch = sys.argv[1]
+    pr_number = sys.argv[2]
+
+    repo_root = run(["git", "rev-parse", "--show-toplevel"]).stdout.strip()
```
cheftako, line 28: Seems like we should be checking the result of run().returncode is non zero and exiting.

justinsb: We are, because we pass `check=True` to subprocess.  That's somewhat cryptic, so I renamed `run` -> `must_run_command` following go's `MustFoo` convention

justinsb:
I guess we're OK on the cherry-picks this generated, so...

/remove-hold

PR #5234, opened 2025-09-23
chore: Unify ComputeForwardingRule direct API and controller

gemmahou:
### BRIEF Change description

Address https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/5191 for ComputeForwardingRule

```release-note
NONE
```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/apis/common/reference/utils.go
+++ b/apis/common/reference/utils.go
@@ -19,6 +19,7 @@ import "strings"
 // FixStaleComputeExternalFormat converts the "External" reference field to the right format if a SelfLink value is used.
 // This guarantees the backward compatibility for Compute Beta resources.
 func FixStaleComputeExternalFormat(external string) string {
+	external = strings.TrimPrefix(external, "/compute.googleapis.com/")
 	external = strings.TrimPrefix(external, "https://www.googleapis.com/compute/v1/")
 	external = strings.TrimPrefix(external, "https://www.googleapis.com/compute/v1beta1/")
 	external = strings.TrimPrefix(external, "/")


apis/compute/v1beta1/forwardingrule_types.go
acpana, line 1: Correct me if I'm wrong but I am assuming that the removal of `optional` tags was done by our tooling? iirc, a field is optional by default so it may seem redundant to keep the tags but just curious.

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrule-direct/_generated_object_globalcomputeforwardingrule-direct.golden.yaml
acpana, line 1: here and for the remainder, IIRC we didn't want to have the service prefix, `/compute.googleapis.com/` in the externalRef field naming?

Do you know if this is a behavior change at all? If it's just for our mocks then no worries.

But if I'm a CX and have `externalRef` as  `/compute.googleapis.com/projects/${projectId}/global/forwardingrules/computeglobalforwardingrule-${uniqueId}` will the new reference format error out when validating the externalRef? Just wanted to double check

acpana (commented):
/lgtm 

two questions but overall LGTM!

cc @cheftako for final approval 💯

apis/compute/v1beta1/forwardingrule_types.go
```diff
@@ -32,17 +30,19 @@ var (
...
 
 	/* Immutable. The value that the label must match. The value has a maximum
 	length of 1024 characters. */
-	Value string `json:"value"`
+	// +kcc:proto:field=google.cloud.compute.v1.MetadataFilterLabelMatch.value
+	Value *string `json:"value"`
 }
 
-type ForwardingruleIpAddress struct {
+type IpAddress struct {
```
cheftako, line 45: Changing public types is dangerous. We should try to avoid changing things like this if we can. You cannot know what code might be depending on this.

apis/compute/v1beta1/forwardingrule_types.go
```diff
@@ -134,21 +136,21 @@ type ComputeForwardingRuleSpec struct {
 
 	Otherwise only allows access from clients in the same region as the
 	internal load balancer. */
-	// +optional
+	// +kcc:proto:field=google.cloud.compute.v1.ForwardingRule.allow_global_access
 	AllowGlobalAccess *bool `json:"allowGlobalAccess,omitempty"`
 
 	/* This is used in PSC consumer ForwardingRule to control whether the PSC endpoint can be accessed from another region. */
-	// +optional
+	// +kcc:proto:field=google.cloud.compute.v1.ForwardingRule.allow_psc_global_access
```
cheftako, line 143: Is this no longer optional?

apis/compute/v1beta1/forwardingrule_identity.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/GoogleCloudPlatform/k8s-config-connector/apis/common"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/apis/common/parent"
+	refsv1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ForwardingRuleIdentity defines the resource reference to ComputeForwardingRule, which "External" field
+// holds the GCP identifier for the KRM object.
+type ForwardingRuleIdentity struct {
+	parent *parent.ComputeParent
```
cheftako: parent here is a terrible name. It causes confusion between inheritance and reference. Especially bad since both are involved with identity.

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrule-direct/_generated_object_globalcomputeforwardingrule-direct.golden.yaml
cheftako: +1. I think understanding this is my last reservation. Otherwise looks good.

apis/compute/v1beta1/forwardingrule_types.go
cheftako: I had a similar question. However the CRD does not appear changed and it defines the actual API.

gemmahou: `optional` was generated in old types([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/0956699ee80cee3c3b67e48ee3e87ee51d9cdc88/pkg/clients/generated/apis/compute/v1beta1/computeforwardingrule_types.go#L123)), in the new(direct) types we no longer specify `optional` but only `required`([guide](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/0956699ee80cee3c3b67e48ee3e87ee51d9cdc88/docs/develop-resources/api-conventions/validations.md#rule-1-requiredoptional)).

apis/compute/v1beta1/forwardingrule_identity.go
gemmahou: This is a pattern we are following rn([template code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/0956699ee80cee3c3b67e48ee3e87ee51d9cdc88/dev/tools/controllerbuilder/template/apis/identity.go#L32)). 

We could have a todo to improve this part, but I'd prefer follow up in a separate task.

yuwenma: it's a placeholder and won't work if resources under the same service have different parent.
FYI, please use the `identity.Identity`  interface.

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computeforwardingrule/globalcomputeforwardingrule-direct/_generated_object_globalcomputeforwardingrule-direct.golden.yaml
gemmahou: This comes from https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5234/files#diff-3d5c352da06f366951f58bf176d34ce6c8ef22b52c61c0eaa6faf40ee285c5abR229

The old `AsExternalRef` function has `serviceDomain` as prefix([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5234/files#diff-2801c5f24dc0ee64809402a8971cfcd735c1a7298ea2bf73c684a25d55d76d39L43-L47)), but it's out-dated.

I updated the FixStaleComputeExternalFormat function to trim the old prefix of externalRef to ensure all formats pass the validation, see https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5234/commits/00f408ac4811b56d06308d5e89b109d9db54150d.

apis/compute/v1beta1/forwardingrule_types.go
gemmahou: `IpAddress` is used [here](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5234/files#diff-936a82c39f27f19e17a6a775dc82358e88038029dfd144d0f5dff7ed0e0e0027R175-R176) in the KRM types, this change is to ensure both field name and type name are match the proto field name: `+kcc:proto:field=google.cloud.compute.v1.ForwardingRule.I_p_address`.

apis/compute/v1beta1/forwardingrule_types.go
gemmahou: Addressed in [this comment](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5234#discussion_r2470628533)

gemmahou:
> Please make sure release note entry above is filled.

This PR does not contain user-facing changes(internal direct controller refactoring). So I did not add a release notes...

apis/compute/v1beta1/forwardingrule_identity.go
gemmahou: > it's a placeholder and won't work if resources under the same service have different parent.

This is a field under ForwardingRuleIdentity, the field name itself might be confusing, but it won't cause issues AFAICT.

I guess the `parent` you mentioned is the Parent struct in types([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/1a6e3b10a2b67d86b41ec0e5b2643004490ee280/apis/compute/v1alpha1/interconnect_types.go#L25-L28)), it will cause duplicate struct name if resources under the same service have different parent. So I updated another compute resource to use a different name [here](https://github.com/GoogleCloudPlatform/k8s-config-connector/blame/1a6e3b10a2b67d86b41ec0e5b2643004490ee280/apis/compute/v1alpha1/networkattachment_types.go#L26-L32). 

This is not the only resource that uses `parent`. I would still prefer to address the naming problem in a followup PR.

> please use the identity.Identity interface.

Done in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5234/commits/493e87480160bec8defcab20abdb3fd406b92882

cheftako:
/lgtm
Need to resolve conflict.

gemmahou:
> /lgtm Need to resolve conflict.

Resolved.

PR #5090, opened 2025-08-29
feat: unmanage edition

acpana:
### Change description

Provide support to unmanage the spec.settings.edition field.

#### Does this PR add something which needs to be 'release noted'?

```release-note

```

- [ ] Reviewer reviewed release note.

#### Additional documentation e.g., references, usage docs, etc.:

```docs

```

#### Intended Milestone

Please indicate the intended milestone. 
- [ ] Reviewer tagged PR with the actual milestone.

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #5246, opened 2025-09-25
genai:feat: unmanage instanceType 

acpana:
### BRIEF Change description

Adds the ability to unmanage the `instanceType` field.

This will support TF -> direct upgrades for SQLInstance

```release-note
NONE
```

```docs

```

#### Intended Milestone

### Tests you have done


acpana:
needs #5090

pkg/controller/direct/sql/sqlinstance_defaults.go
```diff
@@ -23,7 +23,16 @@ import (
 )
 
 func ApplySQLInstanceGCPDefaults(in *krm.SQLInstance, out *api.DatabaseInstance, actual *api.DatabaseInstance, unmanagedFields []string) {
-	if in.Spec.InstanceType == nil {
+	isInstanceTypeUnmanaged := false
```
cheftako, line 26: Not blocking but rather than repeatedly walking 'unmanagedFields' here and for 'isEditionUnmanaged' later, it seems like we should walk it once and get both unmanaged flags.

cheftako: Would be good to organize like with like. I think we should have a follow up to clean this up.

PR #5147, opened 2025-09-10
tests: fix anonation setting

acpana:
### Change description

I noticed in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5090 that I couldn't set an annotation!

#### Does this PR add something which needs to be 'release noted'?

```release-note
NONE
```

- [ ] Reviewer reviewed release note.

#### Additional documentation e.g., references, usage docs, etc.:

```docs

```

#### Intended Milestone

Please indicate the intended milestone. 
- [ ] Reviewer tagged PR with the actual milestone.

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #5249, opened 2025-09-26
test: Recover cloudidentity scenarios tests

gemmahou:
### BRIEF Change description

Address https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5147/files#r2338156787

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/tests/e2e/script_test.go
+++ b/tests/e2e/script_test.go
@@ -316,6 +316,7 @@ func TestE2EScript(t *testing.T) {
 
 					case "ABANDON-AND-REACQUIRE-WITH-GENERATED-ID":
 						existing := readObject(h, obj.GroupVersionKind(), obj.GetNamespace(), obj.GetName())
+						// Get the server generated id from spec.resourceID(legacy) or status.externalRef(direct)
 						resourceID, _, _ := unstructured.NestedString(existing.Object, "spec", "resourceID")
 						if resourceID == "" {
 							externalRef, _, _ := unstructured.NestedString(existing.Object, "status", "externalRef")
@@ -325,9 +326,14 @@ func TestE2EScript(t *testing.T) {
 							tokens := strings.Split(externalRef, "/")
 							resourceID = tokens[len(tokens)-1]
 						}
-						setAnnotation(h, obj, "cnrm.cloud.google.com/deletion-policy", "abandon")
+
+						// Abandon the original resource (without spec.resourceID set)
 						deleteObj := obj.DeepCopy()
+						unstructured.RemoveNestedField(deleteObj.Object, "spec", "resourceID")
+						setAnnotation(h, deleteObj, "cnrm.cloud.google.com/deletion-policy", "abandon")
 						create.DeleteResources(h, create.CreateDeleteTestOptions{Create: []*unstructured.Unstructured{deleteObj}})
+
+						// Replace placeholder in configuration yaml with the server generated id
 						configuredID, _, _ := unstructured.NestedString(obj.Object, "spec", "resourceID")
 						if configuredID == "" {
 							h.Fatalf("object does not have resourceID configured: %v", obj)
--- a/tests/e2e/testdata/scenarios/acquisition/firewallpolicy/_object01.yaml
+++ b/tests/e2e/testdata/scenarios/acquisition/firewallpolicy/_object01.yaml
@@ -2,8 +2,8 @@ apiVersion: compute.cnrm.cloud.google.com/v1beta1
 kind: ComputeFirewallPolicy
 metadata:
   annotations:
-    cnrm.cloud.google.com/deletion-policy: abandon
     cnrm.cloud.google.com/management-conflict-prevention-policy: none
+    cnrm.cloud.google.com/project-id: ${projectId}
     cnrm.cloud.google.com/state-into-spec: absent
   finalizers:
   - cnrm.cloud.google.com/finalizer
--- a/dev/tasks/run-e2e
+++ b/dev/tasks/run-e2e
@@ -39,8 +39,7 @@ flaky_tests=(
   "TestE2EScript/scenarios/gkehubfeaturemembership"
   # https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/4624
   "TestE2EScript/scenarios/iam/iampartialpolicy"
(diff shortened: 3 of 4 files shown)


tests/e2e/script_test.go
```diff
@@ -325,9 +326,14 @@ func TestE2EScript(t *testing.T) {
 							tokens := strings.Split(externalRef, "/")
 							resourceID = tokens[len(tokens)-1]
 						}
-						setAnnotation(h, obj, "cnrm.cloud.google.com/deletion-policy", "abandon")
+
+						// Abandon the original resource (without spec.resourceID set)
 						deleteObj := obj.DeepCopy()
+						unstructured.RemoveNestedField(deleteObj.Object, "spec", "resourceID")
```
maqiuyujoyce, line 332: I think `spec.resourceID` may be used to override `metadata.name` for acquisition. Can we restrict the workaround to known problematic cases only?

cheftako: "REACQUIRE-WITH-GENERATED-ID"

I would assume this use case is specifically about acquiring using the status.?externalRef|resourceID|? . While we can acquire using the spec.resourceID I assume that use case is already covered by the regular ACQUIRE test. REACQUIRE suggests the status use cases.

gemmahou: Re @maqiuyujoyce This test option is specific to acquire resources by it's server-generated id([ref](https://cloud.google.com/config-connector/docs/how-to/managing-deleting-resources#service-generated-resource-id)). So we only use this option in `ComputeFirewallPolicy(Rule)` and `CloudIdentityGroup/Membership` tests, by adding `TEST: ABANDON-AND-REACQUIRE-WITH-GENERATED-ID` flag in the test yaml. Other tests without this flag won't be affected.

gemmahou: Re @cheftako 

> While we can acquire using the spec.resourceID I assume that use case is already covered by the regular ACQUIRE test.

For resources that managed by legacy controller, the generated id will be written back to it's `spec.resourceID` after creation. For resources managed by direct controller, the generated id is written to `status.externalRef`. So there're two ways to get the generated id for acquisition.

maqiuyujoyce: Let's unblock the master head for now considering it is only used for a few scenario tests.

This line is to remove `spec.resourceID` for the deepcopied delete object. I'm not sure it should be needed in general as there are use cases about abandoning a resource with a service-generated ID set in `spec.resourceID` field.

maqiuyujoyce (commented):
/lgtm
/approve

tests/e2e/testdata/scenarios/acquisition/firewallpolicy/_object01.yaml
yuwenma, line 1: I was fixing the presubmit flaky tests and see this, could you elaborate how the FirewallPolicy and FirewallPolicy rules are related to the CloudIdentity tests?

tests/e2e/testdata/scenarios/acquisition/firewallpolicy/_object01.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: compute.cnrm.cloud.google.com/v1beta1
 kind: ComputeFirewallPolicy
 metadata:
   annotations:
-    cnrm.cloud.google.com/deletion-policy: abandon
     cnrm.cloud.google.com/management-conflict-prevention-policy: none
+    cnrm.cloud.google.com/project-id: ${projectId}
```
yuwenma, line 6: If the project-id is missing, the test should fail. But I'm confused how the test passes before. Could you elaborate?

yuwenma: Wait, the ComputeFirewallPolicy is OrgRef and should not have  the project-id annotation. 
 https://cloud.google.com/config-connector/docs/reference/resource-docs/compute/computefirewallpolicy

Could you elaborate how this resource has the wrong label added?

gemmahou: This is added by the test setup for all objects: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/e8d0ea82adfb220cca88464bf782a9cbc6e2fe87/tests/e2e/script_test.go#L647-L657

Thanks for checking. I agree this resource shouldn't include the project-id, though our current test setup didn't distinguish this. IIUC the test won't be affected even if the project-id is set unnecessarily, as the controller will not touch it. But test will fail if it's required.

yuwenma: Why do those test pass before?

tests/e2e/testdata/scenarios/acquisition/firewallpolicy/_object01.yaml
gemmahou: Both FirewallPolicy and CloudIdentity tests use the "ABANDON-AND-REACQUIRE-WITH-GENERATED-ID" flag. The code change in this PR to  "ABANDON-AND-REACQUIRE-WITH-GENERATED-ID" caused:

1. The generated golden object for FirewallPolicy has been reverted to its state before PR #5147: cnrm.cloud.google.com/deletion-policy: abandon annotation was removed and cnrm.cloud.google.com/project-id: ${projectId} annotation was added back.

2. CloudIdentity's golden object did not change because their annotations were not altered in PR #5147 (they were only commented out). These tests have now been recovered and are running again.

Which presubmit flaky test you are looking into?

tests/e2e/testdata/scenarios/acquisition/firewallpolicy/_object01.yaml
gemmahou: Also wanted to mention #5250, this is a followup PR of #5147, where existing annotation(project-id) was removed while setting new annotation.

gemmahou: > Why do those test pass before?

IIUC the test won't be affected even if the project-id is set unnecessarily or unset, as the controller will not touch it.

PR #5250, opened 2025-09-27
test: Keep existing annotations

gemmahou:
### BRIEF Change description

Followup of #5147 

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/tests/e2e/script_test.go
+++ b/tests/e2e/script_test.go
@@ -552,8 +552,12 @@ func setAnnotation(h *create.Harness, obj *unstructured.Unstructured, k, v strin
 	patch.SetGroupVersionKind(obj.GroupVersionKind())
 	patch.SetNamespace(obj.GetNamespace())
 	patch.SetName(obj.GetName())
-	annotations := map[string]string{
+	newAnnotations := map[string]string{
 		k: v,
+	}
+	annotations := patch.GetAnnotations()
+	for key, value := range newAnnotations {
+		annotations[key] = value
 	}
 	patch.SetAnnotations(annotations)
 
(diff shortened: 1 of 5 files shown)


cheftako (commented):
This pull request fixes a bug in the e2e test suite where setting an annotation would overwrite existing annotations. It also adds the `cnrm.cloud.google.com/project-id` annotation to several test resources, which is a good practice for ensuring tests are hermetic.

The fix in `tests/e.2e/script_test.go` is correct and ensures that existing annotations are preserved when adding a new one.

This is a good cleanup and bug fix for the test suite.

tests/e2e/script_test.go
```diff
@@ -555,6 +555,10 @@ func setAnnotation(h *create.Harness, obj *unstructured.Unstructured, k, v strin
 	annotations := map[string]string{
 		k: v,
 	}
+	existingAnnotations := patch.GetAnnotations()
```
cheftako: This will clobber K, v if k == ek. We should reorder these operations so that k clobbers ek if k == ek.

gemmahou: Good point! Updated so new annotations(with the same key) will override the old ones

tests/e2e/script_test.go
```diff
@@ -552,9 +552,13 @@ func setAnnotation(h *create.Harness, obj *unstructured.Unstructured, k, v strin
 	patch.SetGroupVersionKind(obj.GroupVersionKind())
 	patch.SetNamespace(obj.GetNamespace())
 	patch.SetName(obj.GetName())
-	annotations := map[string]string{
+	newAnnotations := map[string]string{
 		k: v,
 	}
+	annotations := patch.GetAnnotations()
```
cheftako, line 558: newAnnotations has only 1 entry, so no real point in ranging across the map.
Instead you could replace the for loop with
```golang
annotations[k] = v
```

cheftako:
/lgtm
/approve

Please consider simplifying the for loop in a follow up PR.

PR #5255, opened 2025-09-30
feat: export for runjob

yuwenma:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/controller/direct/registry/registry.go
+++ b/pkg/controller/direct/registry/registry.go
@@ -147,7 +147,11 @@ func SupportsIAM(groupKind schema.GroupKind) (bool, error) {
 		return false, nil
 	case schema.GroupKind{Group: "discoveryengine.cnrm.cloud.google.com", Kind: "DiscoveryEngineDataStore"}:
 		return false, nil
+		// case schema.GroupKind{Group: "run.cnrm.cloud.google.com", Kind: "RunJob"}:
+		// 	// https://cloud.google.com/run/docs/reference/iam/roles
+		// 	return true, nil
 	}
+
 	klog.Warningf("groupKind %v is not recognized as a direct kind for SupportsIAM check", groupKind)
 	return false, nil
 }
--- a/apis/run/v1beta1/job_identity.go
+++ b/apis/run/v1beta1/job_identity.go
@@ -20,15 +20,28 @@ import (
 	"strings"
 
 	"github.com/GoogleCloudPlatform/k8s-config-connector/apis/common"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/apis/common/identity"
 	refsv1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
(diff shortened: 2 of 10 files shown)


pkg/controller/direct/registry/registry.go
```diff
@@ -124,6 +124,9 @@ func IsIAMDirect(groupKind schema.GroupKind) bool {
 	switch groupKind {
 	case schema.GroupKind{Group: "privateca.cnrm.cloud.google.com", Kind: "PrivateCACAPool"}:
 		return true
+	case schema.GroupKind{Group: "run.cnrm.cloud.google.com", Kind: "RunJob"}:
+		// https://cloud.google.com/run/docs/reference/iam/roles
+		return true
```
justinsb: I think this is orthogonal ... I think if you do this you have to implement the IAM support in the direct controller.  But looking at the logs, I see the calls, so you must have implemented it?

justinsb:
Export LGTM, but now I'm confused about where the IAM support is coming from - I don't see it in the direct code?

yuwenma:
> Export LGTM, but now I'm confused about where the IAM support is coming from - I don't see it in the direct code?

Yeah, It is confusing.   There are two things twisted together
1. For the getIAMPolicy call in the _http.log:  Since the direct controller lacks a custom IAMAdapter, KCC defaults to its Terraform-based implementation. The actual API call is made from within the Terraform provider code located at third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/cloudrunv2/iam_cloud_run_v2_job.go.  
2. The function name `IsIAMDirect` is a little bit confusing for TF/DCL migrated resource. I can remove the RunJob from the list, but the log will still make the getIAMPolicy call.  I am not super familiar of the new IAMPolicy. How do we want to support them in the direct controller (especially for TF/DCL migration)?

pkg/controller/direct/registry/registry.go
```diff
@@ -147,7 +150,11 @@ func SupportsIAM(groupKind schema.GroupKind) (bool, error) {
 		return false, nil
 	case schema.GroupKind{Group: "discoveryengine.cnrm.cloud.google.com", Kind: "DiscoveryEngineDataStore"}:
 		return false, nil
+	case schema.GroupKind{Group: "run.cnrm.cloud.google.com", Kind: "RunJob"}:
```
justinsb: I think this is causing the change in Export, and is unrelated?

justinsb: I'm still confused about what code is making the call to getIamPolicy though .... I _thought_ we had to implement something in the direct controller

yuwenma: cleaned up! (I was totally confused by one of the runjob test "jobwithiamserviceaccount).

justinsb:
We probably should add implement IAM support in direct - hopefully we can make that a cross-cutting thing that the AI can help with.

justinsb (approved):

PR #5259, opened 2025-09-30
feat: Apply cookie to RunJob

yuwenma:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


yuwenma:
issue #5073

pkg/controller/direct/common/hash.go
```diff
@@ -0,0 +1,123 @@
...
+func NormalizeProto(pb proto.Message) {
+	// TODO: Should we also clear fields like `uid` and `name`?
+	// The problem is that they are not consistently marked as output-only.
+
+	// We clear fields that are known to be volatile
+	volatileFieldNames := []string{"etag", "update_time", "updated_time"}
+
+	// We also clear fields that are marked as output-only
+	// We do this by walking the fields and checking for `field_behavior: OUTPUT_ONLY`
+	// We build a field mask and then use that to clear the fields.
```
yuwenma, line 59: even if we clear the output-only field, we still update the kcc `.status.observedState` using the current gcp actual proto.

pkg/controller/direct/common/hash.go
```diff
@@ -0,0 +1,110 @@
...
+	"google.golang.org/genproto/googleapis/api/annotations"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/types/known/fieldmaskpb"
+)
+
+// HashProto calculates a hash of a proto message.
+// We use this to detect changes to the GCP resource.
+func HashProto(pb proto.Message) (string, error) {
```
justinsb: Nit: I would call the object `obj` because `pb` is commonly used as a package import alias

pkg/controller/direct/common/hash.go
```diff
@@ -0,0 +1,110 @@
...
+	// We normalize the proto by clearing output-only fields etc
+	// We do this on a copy
+	pb = proto.Clone(pb)
+	NormalizeProto(pb)
+
+	// We use a deterministic json marshaler.
```
justinsb: Not as deterministic as you might think, sadly: https://github.com/golang/protobuf/issues/1373

I don't know if the proto encoding is more stable...

pkg/controller/direct/common/hash.go
```diff
@@ -0,0 +1,110 @@
...
+		return "", fmt.Errorf("cannot json marshal proto: %w", err)
+	}
+
+	h := sha256.Sum256(j)
+	return hex.EncodeToString(h[:]), nil
+}
+
+// NormalizeProto clears fields that are not significant for comparison.
+// It modifies the passed-in proto.
+func NormalizeProto(pb proto.Message) {
```
justinsb, line 50: We could also use the mapper and only look at spec fields (by going from proto -> krm spec -> proto)

pkg/controller/direct/run/job_controller.go
```diff
@@ -173,28 +174,59 @@ func (a *JobAdapter) Create(ctx context.Context, createOp *directbase.CreateOper
...
 
+	mapCtx := &direct.MapContext{}
 	status := &krm.RunJobStatus{}
 	status.ObservedState = RunJobObservedState_FromProto(mapCtx, created)
 	if mapCtx.Err() != nil {
 		return mapCtx.Err()
 	}
 	status.ExternalRef = direct.LazyPtr(a.id.String())
+
+	specHash, err := common.HashProto(a.desired)
```
justinsb, line 185: I suggest creating `type Cookie struct` or `type ChangeCookie struct` so you can put these helpers onto it

pkg/controller/direct/run/job_controller.go
```diff
@@ -173,28 +174,59 @@ func (a *JobAdapter) Create(ctx context.Context, createOp *directbase.CreateOper
...
+	if err != nil {
+		return fmt.Errorf("calculating spec hash: %w", err)
+	}
+	gcpHash, err := common.HashProto(created)
+	if err != nil {
+		return fmt.Errorf("calculating gcp hash: %w", err)
+	}
+	status.LastModifiedCookie = direct.LazyPtr(fmt.Sprintf("%s/%s", specHash, gcpHash))
```
justinsb: Would also be nice to encapsulate this logic into `type Cookie`

pkg/controller/direct/run/job_controller.go
```diff
@@ -173,28 +174,59 @@ func (a *JobAdapter) Create(ctx context.Context, createOp *directbase.CreateOper
...
-	desiredPb.Name = a.id.String()
-
-	paths, err := common.CompareProtoMessage(desiredPb, a.actual, common.BasicDiff)
+	gcpHash, err := common.HashProto(a.actual)
+	if err != nil {
+		return fmt.Errorf("calculating gcp hash: %w", err)
+	}
+	cookie := fmt.Sprintf("%s/%s", specHash, gcpHash)
+
+	if a.lastModifiedCookie != nil {
```
justinsb, line 216: Ditto to try to move this logic into `type Cookie` also (for simplicity)

pkg/controller/direct/run/job_controller.go
```diff
@@ -173,28 +174,59 @@ func (a *JobAdapter) Create(ctx context.Context, createOp *directbase.CreateOper
...
+			status.ObservedState = RunJobObservedState_FromProto(mapCtx, a.actual)
+			if mapCtx.Err() != nil {
+				return mapCtx.Err()
+			}
+			status.LastModifiedCookie = a.lastModifiedCookie // Preserve cookie
+			return updateOp.UpdateStatus(ctx, status, nil)
+		}
+	}
+	a.desired.Name = a.actual.Name // Name is output only, ensure it's set to avoid diffs.
+	paths, err := common.CompareProtoMessage(a.desired, a.actual, common.BasicDiff)
```
justinsb: I _think_ you send all the fields, all the time, in an update.

tests/e2e/normalize.go
```diff
@@ -73,6 +73,7 @@ func normalizeKRMObject(t *testing.T, u *unstructured.Unstructured, project test
 	visitor.removePaths.Insert(".metadata.uid")
 
 	visitor.replacePaths[".metadata.deletionTimestamp"] = "1970-01-01T00:00:00Z"
+	visitor.replacePaths[".status.lastModifiedCookie"] = "normalized-cookie"
```
justinsb, line 76: Actually, it might be interesting _not_ to normalize this yet ... because I think it should be stable (aka we want to know if it changes).  It's not the end of the world if it changes, but it will cause an extra update on all resources, so we want to know...

pkg/controller/direct/common/hash.go
yuwenma: Good to know. This refers to the "UseEnumNumbers: true", if not added, it cannot distinguish the "GA" vs "4" in proto enum.

pkg/controller/direct/run/job_controller.go
yuwenma: ok, I like it.

pkg/controller/direct/run/job_controller.go
yuwenma: ~we do, but many GCP requires the update field mask (and that has to correct for oneOf, and immutable cases)~

tests/e2e/normalize.go
yuwenma: Well, I did that originally. But since we use dynamic name `${uniqueId}` , it makes the GCP hash different every time.  the "Name" is output-only for create, but input for update.

pkg/controller/direct/run/job_controller.go
yuwenma: oh, you are right.

pkg/controller/direct/common/hash.go
yuwenma: Right, but that does not include the comparison for those new fields where KCC does not have in its API yet. I don't know if that's good or bad.

pkg/controller/direct/common/hash.go
justinsb: OK but I think protojson deliberately introducing randomness will cause this to start failing tests.

I have no problem with UseEnumNumbers, but the comment suggests that the service can impact our local proto marshalling, though I don't think that is the case (?)

pkg/controller/direct/common/hash.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		return "", fmt.Errorf("error marshalling cookie: %w", err)
+	}
+	return string(b), nil
+}
+
+// ParseCookie parses a cookie string.
+func ParseCookie(s string) (*Cookie, error) {
+	cookie := &Cookie{}
+	if err := json.Unmarshal([]byte(s), cookie); err != nil {
```
justinsb, line 132: I do think we'll want to evolve this format, albeit at the cost of unnecessary reconciliations.  But we can still break it, I think.

justinsb: Also I think this cookie will be way more readable than we want it to be.  We don't want anyone to depend on the cookie.

PR #5265, opened 2025-10-01
tagslocationtagbinding: Add api and mapper for

600lyy:
Add api and mapper for tagslocationtagbinding

### Add fuzzer test

export REPO_ROOT="$(git rev-parse --show-toplevel)"
cd $REPO_ROOT/dev/tools/controllerbuilder

LLM_MODEL="gemini-2.0-flash-exp" \
go run main.go prompt \
  <<EOF > $REPO_ROOT/pkg/controller/direct/tags/tagslocationtagbinding_fuzzer.go
// +tool:fuzz-gen
// proto.message: google.cloud.resourcemanager.v3.TagBinding
// api.group: tags.cnrm.cloud.google.com
EOF


apis/tags/v1alpha1/tagslocationtagbinding_type.go
```diff
@@ -0,0 +1,101 @@
...
+var TagsLocationTagBindingGVK = GroupVersion.WithKind("TagsLocationTagBinding")
+
+// TagsLocationTagBindingSpec defines the desired state of TagsLocationTagBinding
+// +kcc:spec:proto=google.cloud.resourcemanager.v3.TagBinding
+type TagsLocationTagBindingSpec struct {
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="the field is immutable"
+	// The full resource name of the resource the TagValue is bound to.
+	/* Only `external` field is supported to configure the reference for now. */
+	// +required
+	ParentRef *refv1alpha1.ResourceRef `json:"parentRef"`
```
yuwenma: Which parent is it? Can we use the dedicated kind?
More context: the ResourceRef from apis/k8s/v1alpha1 is a legacy ref, it is very generic and requires a `.kind` field. This makes it  ambiguous on the reference resolving path.

600lyy: parent can actually be any type of GCP resource, a storage bucket, a VM instance, a Cloud Run service, just to name a few.

600lyy: Is it ok to allow only "external", or it has to support both "external" and "name/kind" ?

600lyy: As per discussion, the solution is to use a special ref that only supports "External" field.

600lyy:
Agreed with KCC team to combine the Tagbinding and TagsLocationBinding Kind soince they're using the same API and proto

PR #5268, opened 2025-10-01
Fix dataflowflextemplatejob when subnetworkRef external is full url

gemmahou:
### BRIEF Change description

Fixes customer issue in the opt-in direct controller:
```
Warning UpdateFailed 27s (x18 over
24m) dataflowflextemplatejob-controller Update call failed: format of
subnetwork
external="projects/xxx/regions/us-central1/subnetworks/xxx"
was not known, use regions/<region>/subnetworks/<subnetwork>,
projects/<projectID>/regions/<region>/subnetworks/<subnetwork>
or https://www.googleapis.com/compute/v1/projects/<projectID>/regions/<region>/subnetworks/<subnetwork>
```

```release-note
## Bug Fixes:

*   [PR#5268](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5268)
     Fixed format validation issue in the DataflowFlexTemplateJob direct controller when the `spec.subnetworkRef.external` field contains full URL.
```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/dataflow/v1beta1/dataflowflextemplatejob/dataflowflextemplatejob-direct-subnetwork2/create.yaml
+++ b/pkg/test/resourcefixture/testdata/basic/dataflow/v1beta1/dataflowflextemplatejob/dataflowflextemplatejob-direct-subnetwork2/create.yaml
@@ -1,0 +1,34 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: dataflow.cnrm.cloud.google.com/v1beta1
+kind: DataflowFlexTemplateJob
+metadata:
+  annotations:
+    cnrm.cloud.google.com/on-delete: "cancel"
+    alpha.cnrm.cloud.google.com/reconciler: "direct"
+  name: dataflowflextemplatejob-${uniqueId}
+spec:
+  region: us-central1
+  # This is a public, Google-maintained Dataflow Job flex template of a batch job
+  containerSpecGcsPath: gs://dataflow-templates/2020-08-31-00_RC00/flex/PubSub_Avro_to_BigQuery
+  parameters:
+    # This is maintained by us.
+    schemaPath: gs://config-connector-samples/dataflowflextemplate/numbers.avsc
+    inputSubscription: projects/${projectId}/subscriptions/pubsubsubscription-${uniqueId}
+    outputTopic: projects/${projectId}/topics/pubsubtopic0-${uniqueId}
+    outputTableSpec: ${projectId}:bigquerydataset${uniqueId}.bigquerytable${uniqueId}
+    createDisposition: CREATE_NEVER
+  subnetworkRef:
+    external: https://www.googleapis.com/compute/v1/projects/${projectId}/regions/us-central1/subnetworks/default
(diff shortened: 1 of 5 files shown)


pkg/test/resourcefixture/testdata/basic/dataflow/v1beta1/dataflowflextemplatejob/dataflowflextemplatejob-direct-subnetwork/create.yaml
```diff
@@ -31,4 +31,4 @@ spec:
     outputTableSpec: ${projectId}:bigquerydataset${uniqueId}.bigquerytable${uniqueId}
     createDisposition: CREATE_NEVER
   subnetworkRef:
-    external: projects/${projectId}/regions/us-central1/subnetworks/default
+    external: https://www.googleapis.com/compute/v1/projects/${projectId}/regions/us-central1/subnetworks/default
```
yuwenma: Can you use another test suite to verify the fix? Otherwise we loose the coverage for a previous bug we fixed, right? #3370

gemmahou: I don't think `projects/${projectId}/regions/us-central1/subnetworks/default` is a supported format that customer should use, based on the code comment and the API requirements? 

```
// Subnetwork to which VMs will be assigned, if desired. You can specify a
//  subnetwork using either a complete URL or an abbreviated path. Expected to
//  be of the form
//  "https://www.googleapis.com/compute/v1/projects/HOST_PROJECT_ID/regions/REGION/subnetworks/SUBNETWORK"
//  or "regions/REGION/subnetworks/SUBNETWORK". If the subnetwork is located in
//  a Shared VPC network, you must use the complete URL.
```

We might need extra test case for `regions/REGION/subnetworks/SUBNETWORK` but since it's not causing issues and unrelated with the code change, I would hold on adding extra fixtures test for it. I also plan to add unit test for the external validation in future PRs, so we can minimize the maintenance cost of fixtures test. Let me know WDYT.

yuwenma: > based on the code comment and the API requirements?

This is the GCP requirement, not KCC. We want to make the KCC API easy to use and build the convention so users know want to do by instinct and do not need to investigate what the right value should be. ALL KCC references support the `external` to be the short form, why do we forbid that in this specific resource? This would be very confusing and frustrating. What's more, we should not break the customers who already use the existing format 'projects/${projectId}/regions/us-central1/subnetworks/default".

gemmahou: Done, made the full url test as an additional test.

pkg/test/resourcefixture/testdata/basic/dataflow/v1beta1/dataflowflextemplatejob/dataflowflextemplatejob-direct-subnetwork2/create.yaml
```diff
@@ -0,0 +1,34 @@
+# Copyright 2022 Google LLC
```
cheftako, line 1: New file should have a 2025 copyright.

pkg/test/resourcefixture/testdata/basic/dataflow/v1beta1/dataflowflextemplatejob/dataflowflextemplatejob-direct-subnetwork2/dependencies.yaml
```diff
@@ -0,0 +1,21 @@
+# Copyright 2022 Google LLC
```
cheftako, line 1: New file should have a 2025 copyright.

pkg/test/resourcefixture/testdata/basic/dataflow/v1beta1/dataflowflextemplatejob/dataflowflextemplatejob-direct-subnetwork2/create.yaml
gemmahou: Will update the year in followup PRs

PR #5288, opened 2025-10-02
refactor: encapsulate unmanaged field, fix: flaky

acpana:
### BRIEF Change description

Some punted feedback from #5090 & #5246 

cc @cheftako 

* Code health & cleanliness 

```release-note
NONE
```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/tests/e2e/testdata/scenarios/sql/unmanage-edition/_http01.log
+++ b/tests/e2e/testdata/scenarios/sql/unmanage-edition/_http01.log
@@ -1,0 +1,105 @@
+GET https://sqladmin.googleapis.com/sql/v1beta4/projects/${projectId}/instances/sqlinstance-${uniqueId}?alt=json&prettyPrint=false
+User-Agent: kcc/${kccVersion} (+https://github.com/GoogleCloudPlatform/k8s-config-connector) kcc/controller-manager/${kccVersion}
+
+200 OK
+Content-Type: application/json; charset=UTF-8
+Server: ESF
+Vary: Origin
+Vary: X-Origin
+Vary: Referer
+X-Content-Type-Options: nosniff
+X-Frame-Options: SAMEORIGIN
+X-Xss-Protection: 0
+
+{
+  "backendType": "SECOND_GEN",
+  "connectionName": "${projectId}:us-central1:sqlinstance-${uniqueId}",
+  "createTime": "2024-04-01T12:34:56.123456Z",
+  "databaseInstalledVersion": "POSTGRES_15_7",
+  "databaseVersion": "POSTGRES_15",
+  "etag": "abcdef0123A=",
+  "gceZone": "us-central1-a",
+  "geminiConfig": {
+    "activeQueryEnabled": false,
+    "entitled": false,
+    "googleVacuumMgmtEnabled": false,
+    "indexAdvisorEnabled": false,
+    "oomSessionCancelEnabled": false
+  },
+  "instanceType": "CLOUD_SQL_INSTANCE",
+  "ipAddresses": [
+    {
+      "ipAddress": "10.1.2.3",
+      "type": "PRIMARY"
+    },
+    {
+      "ipAddress": "10.1.2.3",
+      "type": "OUTGOING"
+    }
+  ],
+  "kind": "sql#instance",
+  "maintenanceVersion": "POSTGRES_15_7.R20240514.00_12",
+  "name": "sqlinstance-${uniqueId}",
+  "project": "${projectId}",
+  "region": "us-central1",
+  "selfLink": "https://sqladmin.googleapis.com/sql/v1beta4/projects/${projectId}/instances/sqlinstance-${uniqueId}",
+  "serverCaCert": {
+    "cert": "-----BEGIN CERTIFICATE-----\n-----END CERTIFICATE-----\n",
+    "certSerialNumber": "0",
+    "commonName": "common-name",
+    "createTime": "2024-04-01T12:34:56.123456Z",
+    "expirationTime": "2024-04-01T12:34:56.123456Z",
+    "instance": "sqlinstance-${uniqueId}",
+    "kind": "sql#sslCert",
+    "sha1Fingerprint": "12345678"
+  },
+  "serviceAccountEmailAddress": "p${projectNumber}-abcdef@gcp-sa-cloud-sql.iam.gserviceaccount.com",
+  "settings": {
+    "activationPolicy": "ALWAYS",
+    "availabilityType": "ZONAL",
+    "backupConfiguration": {
+      "backupRetentionSettings": {
+        "retainedBackups": 7,
+        "retentionUnit": "COUNT"
+      },
+      "enabled": false,
+      "kind": "sql#backupConfiguration",
+      "startTime": "12:00",
+      "transactionLogRetentionDays": 7,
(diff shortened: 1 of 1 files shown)


pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -105,17 +142,35 @@ func (m *sqlInstanceModel) AdapterForObject(ctx context.Context, kube client.Rea
 		sqlOperationsClient: gcpClient.sqlOperationsClient(),
 		sqlInstancesClient:  gcpClient.sqlInstancesClient(),
 		sqlUsersClient:      gcpClient.sqlUsersClient(),
+		statefulFields:      make(map[string]*StatefulField),
+	}
+
+	// Populate the map with all supported fields, initially marked as not unmanaged.
```
cheftako: Not sure there is a point in regenerating supportedUnmanageableFields. This seems static. Lets only generate this once.

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -105,17 +142,35 @@ func (m *sqlInstanceModel) AdapterForObject(ctx context.Context, kube client.Rea
...
+		adapter.statefulFields[fieldCopy.Path()] = fieldCopy
 	}
 
 	unmanaged, ok := obj.GetAnnotations()["cnrm.cloud.google.com/unmanaged"]
 	if ok && unmanaged != "" {
-		unmanagedFields := strings.Split(unmanaged, ",")
-		for _, field := range unmanagedFields {
-			if field != unmanageEditionAnnotationValue && field != unmanageInstanceTypeAnnotationValue {
-				return nil, fmt.Errorf("unmanaging field `%s` is not supported, supported fields are: %s", field, []string{unmanageEditionAnnotationValue, unmanageInstanceTypeAnnotationValue})
+		unmanagedFieldsList := strings.Split(unmanaged, ",")
```
cheftako, line 173: Seems like we could turn the split and for loop into an isUnmanaged method.

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -317,7 +364,7 @@ func (a *sqlInstanceAdapter) Update(ctx context.Context, updateOp *directbase.Up
 	}
 
 	// Next, handle database edition updates
-	if !isEditionUnmanaged {
+	if !a.statefulFields["spec.settings.edition"].isUnmanaged {
```
cheftako: Might be better to keep the constants for this.

!a.statefulFields[unmanageEditionAnnotationValue].isUnmanaged.

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -38,11 +38,48 @@ import (
...
 
 const (
-	ctrlName                            = "sqlinstance-controller"
-	unmanageEditionAnnotationValue      = "spec.settings.edition"
-	unmanageInstanceTypeAnnotationValue = "spec.instanceType"
+	ctrlName = "sqlinstance-controller"
 )
 
+// StatefulField encapsulates the state and logic for a field that can be unmanaged.
+type StatefulField struct {
```
cheftako: Why 'statefulFields' ? This doesn't seem to be about fields which are stateful.  FieldState or FieldMetadata seems more appropriate.

tests/e2e/script_test.go
```diff
@@ -294,6 +294,10 @@ func TestE2EScript(t *testing.T) {
 
 					case "ABANDON":
 						setAnnotation(h, obj, "cnrm.cloud.google.com/deletion-policy", "abandon")
+						// Pause to allow reconciliation to start. This is to fix a race
+						// condition where the object is deleted before the controller can
+						// act on it, which can lead to flaky tests.
+						time.Sleep(2 * time.Second)
```
cheftako: Can we change this out for something like [wait.Poll](https://github.com/kubernetes/apimachinery/blob/v0.34.1/pkg/util/wait/poll.go#L65)

tests/e2e/script_test.go
```diff
@@ -585,6 +592,45 @@ func readObject(h *create.Harness, gvk schema.GroupVersionKind, namespace, name
 	return obj
 }
 
+// waitForReconciliationAfterPatch waits for a reconciliation to complete after an object has been patched.
```
justinsb, line 595: I have a PR incoming (somewhere) where we watch for observedGeneration.  Is that an option here?

tests/e2e/testdata/scenarios/sql/unmanage-edition/_http04.log
```diff
@@ -0,0 +1,279 @@
...
+      "displayName": "PostgreSQL 16",
+      "majorVersion": "POSTGRES_16",
+      "name": "POSTGRES_16"
+    }
+  ]
+}
+
+---
+
+PATCH https://sqladmin.googleapis.com/sql/v1beta4/projects/${projectId}/instances/sqlinstance-${uniqueId}?alt=json&prettyPrint=false
```
justinsb, line 109: Why are we patching on a touch?

Should we add SQLInstance to the list of resources in ratcheting.go

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -38,11 +38,71 @@ import (
...
+// 	},
+// 	{
+// 		paths: []string{"spec", "instanceType"},
+// 		preserveActualValue: func(out *api.DatabaseInstance, actual *api.DatabaseInstance) {
+// 			out.InstanceType = actual.InstanceType
+// 		},
+// 	},
+// }
+
+var supportedUnmanageableFields = map[string]*FieldMetadata{
```
justinsb, line 86: This is _fine_ but I think we will want to write a generic version (using reflection)

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -38,11 +38,71 @@ import (
...
+	editionFieldPath: {
+		paths: []string{"spec", "settings", "edition"},
```
justinsb, line 88: I don't think we should have to specify both `paths` and `strings.Join(paths, ".")` - feels redundant

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -316,8 +378,13 @@ func (a *sqlInstanceAdapter) Update(ctx context.Context, updateOp *directbase.Up
...
 	}
 
+	isEditionUnamanged := false
+	if editionField, ok := a.fieldMeta[editionFieldPath]; ok {
+		isEditionUnamanged = editionField.isUnmanaged
+	}
+
 	// Next, handle database edition updates
-	if !isEditionUnmanaged {
+	if !isEditionUnamanged {
```
justinsb, line 387: Nit: looks like you introduced a typo :-)

tests/e2e/testdata/scenarios/sql/unmanage-edition/_http04.log
acpana: Great question! 

I remove the unmanage annotation in step 3 but it seems that the machinery doesn't want to reconcile so I force it in step 4.

tests/e2e/script_test.go
acpana: 🤔 I think that could work!

PR #5235, opened 2025-09-23
chore: Unify Compute direct mappers

gemmahou:
### BRIEF Change description

Address https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/5191 

This PR should be merged after the three preceding PRs: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5193, #5194, #5234

Compute contains both v1alpha1 resources and v1beta1, so I applied the new`--multiversion` flag(implemented in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4859) in the generate command.

#### Does this PR add something which needs to be 'release noted'?

```release-note

```

- [ ] Reviewer reviewed release note.

#### Additional documentation e.g., references, usage docs, etc.:

```docs

```

#### Intended Milestone

Please indicate the intended milestone. 
- [ ] Reviewer tagged PR with the actual milestone.

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #5298, opened 2025-10-03
mappergenerator: Mapping from *Type to Type and vice versa

gemmahou:
### BRIEF Change description

KRM types not always match the proto types, especially for existing Beta resources we may not want to change the field type. 

For example, in ComputeFirewallPolicyRule, `spec.action` is `string`([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/0ae7fd0c1b1f1670df74c1091389ec0302ee09a8/apis/compute/v1beta1/firewallpolicyrule_types.go#L88), copied from the [generated types file](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/0ae7fd0c1b1f1670df74c1091389ec0302ee09a8/pkg/clients/generated/apis/compute/v1beta1/computefirewallpolicyrule_types.go#L94)), but the proto type is `*string`
screenshot: https://screenshot.googleplex.com/n23rnJcrftfnt2t.png

This PR adds additional logic in mappergenerator to map *Type to Type or Type to *Type.

Commit 1: the code change to handle mappings between *Type and Type, in mappers generator
Commit 2: the incorrect mappers generated by the original code(for code review only)
Commit 3: the correct mappers generated by the updated code(for code review only)
Commit 4: remove the generated mappers in above 2 commits - I will add the clean mappers in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5235

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


pkg/controller/direct/compute/mapper.generated.go
```diff
@@ -32,14 +32,14 @@ func ComputeFirewallPolicyRuleSpec_FromProto(mapCtx *direct.MapContext, in *pb.F
 		return nil
 	}
 	out := &krm.ComputeFirewallPolicyRuleSpec{}
-	out.Action = in.Action
+	out.Action = direct.ValueOf(in.Action)
```
yuwenma: No, this reveals that the KRM API ComputeFirewallPolicy itself is wrong. 
We should not use "string" but "*string" for an action type so that we can distinguish if this is a empty string or user unspecified field. 
It also shows the KRM API is not generated from the type-generator.  Could you try rerunning the type-generator to get the KRM API?

gemmahou: The types generator correctly generated `*string`, but the existing types is [string](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/0ae7fd0c1b1f1670df74c1091389ec0302ee09a8/pkg/clients/generated/apis/compute/v1beta1/computefirewallpolicyrule_types.go#L94)

It might be out-dated but I remember for existing resources, we want to copy the types from existing types under https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/0ae7fd0c1b1f1670df74c1091389ec0302ee09a8/pkg/clients/generated/apis, and we don't want to make any changes to it? `string` and `*string` might be fine since it does not seem to affect the CRD, but there's another example, where the existing spec.priority's type is [int64](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/0ae7fd0c1b1f1670df74c1091389ec0302ee09a8/pkg/clients/generated/apis/compute/v1beta1/computefirewallpolicyrule_types.go#L118), while the types generator's output is `*int32`. 

What's the correct implementation we should follow today?

gemmahou: I agree this PR is not needed. I updated the direct types to `*string`, which matches the output of the types_generator now: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5234/commits/97c6cbeda933a611cd3ccad2ade1d6cb66ba6c0e

PR #5304, opened 2025-10-03
feat: unify worflows api, mapper and fuzzer

yuwenma:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


yuwenma:
issue #5303

apis/workflows/v1alpha1/workflow_types.go
```diff
@@ -44,51 +44,69 @@ type Parent struct {
...
+	// Workflow code to be executed. The size limit is 128KB.
+	// +kcc:proto:field=google.cloud.workflows.v1.Workflow.source_contents
 	SourceContents *string `json:"sourceContents,omitempty"`
 
 	// Optional. The resource name of a KMS crypto key used to encrypt or decrypt
 	// the data associated with the workflow.
-	// If not provided, data associated with the workflow will not be
-	// CMEK-encrypted.
-	// +optional
-	KMSCryptoKeyRef *refs.KMSCryptoKeyRef `json:"kmsCryptoKeyRef,omitempty"`
```
yuwenma, line 76: We need to fix the json name otherwise mapper-generator cannot find out the KRM field.

pkg/controller/direct/common/refs.go
```diff
@@ -163,6 +165,54 @@ func (r *refNormalizer) VisitField(path string, v any) error {
...
+		if cryptoKeyRef != nil {
+			resolved, err := refs.ResolveKMSCryptoKeyRef(r.ctx, r.kube, r.src, cryptoKeyRef)
+			if err != nil {
+				return fmt.Errorf("resolving KMSCryptoKeyRef at path %q: %w", path, err)
+			}
+			*cryptoKeyRef = *resolved
+		}
+	}
+
+	if normalizer, ok := v.(refs.ExternalNormalizer); ok {
```
yuwenma: This is for those refs who implements ExternalNormalizer, we can have another one for the new Ref (which support FromIdentity). Will do in follow up PR.

apis/workflows/v1alpha1/workflow_types.go
justinsb: I _thought_ the trick was that we could match by the field name, regardless of the json value.   Do we have a standard for this (e.g. is it networkRef or computeNetworkRef)?

pkg/controller/direct/common/refs.go
justinsb: Which is the ref?  I think ExternalNormalizer is the legacy one, and I agree this code is relatively brittle

pkg/controller/direct/workflows/workflow_controller.go
```diff
@@ -65,34 +67,30 @@ func (m *modelWorkflowsWorkflow) client(ctx context.Context) (*gcp.Client, error
...
-	return nil
-}
-
 func (m *modelWorkflowsWorkflow) AdapterForObject(ctx context.Context, reader client.Reader, u *unstructured.Unstructured) (directbase.Adapter, error) {
 	obj := &krm.WorkflowsWorkflow{}
 	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &obj); err != nil {
 		return nil, fmt.Errorf("error converting to %T: %w", obj, err)
 	}
 
+	if err := common.NormalizeReferences(ctx, reader, obj, nil); err != nil {
```
justinsb: Yay 👍

pkg/controller/direct/common/refs.go
yuwenma: There's no legacy references interfaces. The legacy references are all live under `apis/refs/<kind>.go`. cc @gemmahou  (who works on removing the legacy references).  

In this resource, its dependencies use the `ExternalNormalizer` interface. So we add this code snippet so the resource can call `NormalizeReference` to resolve references. 

The previous comment is about adding the Newest interface support in this function, so the new references who use the newest style can benefits from the  `NormalizeReference` as well.

apis/workflows/v1alpha1/workflow_types.go
yuwenma: No, IIUC the json name is the bridge that connects proto and krm. If we change it, then we have to manually write the mapper.

Basically the KRM field name (pascal-case) has to be the same as the protobuf field's (snake case) with Ref suffix.
https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/dev/tools/controllerbuilder/pkg/codegen/mappergenerator.go#L259

pkg/controller/direct/common/refs.go
yuwenma: I noticed some other problems in NormalizeReference (the function needs a refactoring first). Since this is orthogonal effort, I removed the call from this PR so we can continue the reference discussion and unblock the workflow promotion

apis/workflows/v1alpha1/workflow_types.go
```diff
@@ -44,51 +44,69 @@ type Parent struct {
 // +kcc:spec:proto=google.cloud.workflows.v1.Workflow
 type WorkflowsWorkflowSpec struct {
 	Parent `json:",inline"`
-
 	// Description of the workflow provided by the user.
-	// Must be at most 1000 unicode characters long.
+	//  Must be at most 1000 Unicode characters long.
+	//  This is a workflow-wide field and is not tied to a specific revision.
+	// +kcc:proto:field=google.cloud.workflows.v1.Workflow.description
```
gemmahou, line 50: nit: I thought we didn't want to make changes to the types file even they are description updates([previous comment](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5193#discussion_r2391615961)). Shall we make sure the field descriptions are up-to-date with the generated proto(I plan to make the updates for Compute types in a separate PR)? What's the best practice we should follow?

Edit: Noticed this is v1alpha1 resource so the types change should be fine.

yuwenma: Right, this is an alpha resource. We can change the API (and it seems to be better to make it more up-to-date)

apis/workflows/v1alpha1/workflow_types.go
justinsb: What if we change the json tag but not the go field name?  Then we can follow whatever pattern we want in our public API.

yuwenma: > What if we change the json tag but not the go field name? 

Then we cannot use the auto-mapper functions (but we can write the manual function if we want to use a custom name). This is the auto-mapper if using a different name https://screenshot.googleplex.com/4dhf3w8tMxNcMYz.png

https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5491 is to update the reference doc about the naming, and auto-mapper  when turning on the generate.sh.

justinsb: Can you try this: 
```
CryptoKeyNameRef *refs.KMSCryptoKeyRef `json:"kmsCryptoKeyRef,omitempty"`
```

yuwenma: I misread the original ask. The json and go field does not need to match, but this is alpha, isn't it better to keep them the same? Or is your concern about the crypto_key_"name" conflict with the "ref"?

PR #5309, opened 2025-10-05
Internal ranges - Add full testfixture coverage

katrielt:
- Add full test coverage for Network Connectivity Internal Ranges so it can be promoted to beta resource.
- Mark TargetCIDRRange immutable
- Fix update mask field names in mock controller

Mock vs. Real GCP Log Analysis for NetworkConnectivityInternalRange
  Affected Tests:
   * networkconnectivityinternalrange-minimal
   * networkconnectivityinternalrange-full-ipcidrrange
   * networkconnectivityinternalrange-full-prefix

  Key Discrepancies:
   * `compute#network` resource: The routingConfig field is missing from the mock response
     * Depndendency resource - Real API adds defaults fields that are not present in mock
   * Operation Metadata: The done and requestedCancellation fields are missing from the mock operation metadata
   * `InternalRange` resource:
       * The network field has an incorrect format (relative path instead of a full URL)
         * Network is correct in both, just displayed differently
       * The ipCidrRange field is not correctly populated in the response
         * Specific to the prefix test, the allocated range is not known in advance


test.log
yuwenma, line 1: nit: remove this file.

apis/networkconnectivity/v1alpha1/internalrange_types.go
```diff
@@ -73,6 +73,7 @@ type NetworkConnectivityInternalRangeSpec struct {
 
 	// Optional. Can be set to narrow down or pick a different address space while searching for a free range. If not set, defaults to the "10.0.0.0/8" address space. This can be used to search in other rfc-1918 address spaces like "172.16.0.0/12" and "192.168.0.0/16" or non-rfc-1918 address spaces used in the VPC.
 	// +kcc:proto:field=mockgcp.cloud.networkconnectivity.v1.InternalRange.target_cidr_range
+	// +immutable
```
yuwenma: This mark does not work for controller-gen. Please follow the guide about "immutability" https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/develop-resources/api-conventions/validations.md

mockgcp/mocknetworkconnectivity/internalranges.go
```diff
@@ -133,7 +133,7 @@ func (r *internalRanges) PatchProjectsLocationsInternalRange(ctx context.Context
 				obj.Network = patch.Network
 			case "peering":
 				obj.Peering = patch.Peering
-			case "target_cidr_range":
+			case "targetCidrRange":
```
yuwenma, line 136: This change conflicts with the other two changes in API and controller. Could you elaborate?

katrielt: This should probably be removed if the field is immutable or reverted if not. Pending previous comment to decide

apis/networkconnectivity/v1alpha1/internalrange_types.go
katrielt: I'll remove this. Looking at the doc you shared though, none of the options fit. The field is not marked as immutable in the proto and is optional.

I based my assumption on the fact that gcloud network-connectivity internal-ranges update does not let you update this value.

What would be the proper way in this case? Just leave it as-is?

PR #5302, opened 2025-10-03
refactor: update bigquerybiglake references to match apigee style

yuwenma:
### BRIEF Change description

Fixes #

#### Does this PR add something which needs to be 'release noted'?

```release-note

```

- [ ] Reviewer reviewed release note.

#### Additional documentation e.g., references, usage docs, etc.:

```docs

```

#### Intended Milestone

Please indicate the intended milestone. 
- [ ] Reviewer tagged PR with the actual milestone.

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #2758, opened 2024-09-20
fix: Move resolvers out of Adapter

gemmahou:
### Change description

Deleting a resource does not require resolving dependencies. However, if dependencies are deleted prior to the resource itself, it will prevent the deletion of that resource.

Therefore, we should move resolvers to their own function outside the Adapter. We can resolve them during the create and update only.

Fixes #2621 

### Tests you have done

- [x] Run `make ready-pr` to ensure this PR is ready for review.
- [x] Perform necessary E2E testing for changed resources.

PR #5313, opened 2025-10-06
feat: apply graceful-deleiton for apigeeEndpointAttachment

yuwenma:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -336,6 +336,14 @@ func (r *reconcileContext) doReconcile(ctx context.Context, u *unstructured.Unst
 	}
 	if adapteErr != nil {
 		if unwrappedErr, ok := lifecyclehandler.CausedByUnresolvableDeps(adapteErr); ok {
+			// If the resource is being deleted, an unresolvable dependency means
+			// we have an orphan because the model couldn't build an adapter
+			// from status and has now failed to build one from spec.
+			if !u.GetDeletionTimestamp().IsZero() {
+				logger.Info("could not resolve dependencies for deletion and status.externalRef is missing; resource may be orphaned in GCP", "resource", k8s.GetNamespacedName(u), "error", unwrappedErr)
+				// Remove the finalizer to unblock Kubernetes object deletion.
```
justinsb: All in favor of better logging, but removing the finalizer is a big change

tests/e2e/unified_test.go
```diff
@@ -411,6 +411,18 @@ func runScenario(ctx context.Context, t *testing.T, testPause bool, fixture reso
 					}
 				}
 
+				if fixture.DeleteDependencies != nil {
```
justinsb: I think this can just be a randomized test.  We can run all our tests again in a different mode where we don't delete objects in order.  In this mode we will not be able to compare the golden log, but we will still be able to verify that all objects get deleted (don't get stuck)

justinsb: (And we can start with just a script that runs a few fixtures in this randomized mode)

pkg/controller/direct/apigee/endpointattachment_controller.go
```diff
@@ -52,14 +52,43 @@ func (m *modelApigeeEndpointAttachment) AdapterForObject(ctx context.Context, re
...
+
 	i, err := obj.GetIdentity(ctx, reader)
 	if err != nil {
 		return nil, err
 	}
 	id := i.(*krm.ApigeeEndpointAttachmentIdentity)
 
 	// Get apigee GCP client
-	gcpClient, err := newGCPClient(ctx, m.config)
+	gcpClient, err = newGCPClient(ctx, m.config)
```
justinsb: I think we already have a gcpClient

pkg/controller/direct/apigee/endpointattachment_controller.go
```diff
@@ -52,14 +52,43 @@ func (m *modelApigeeEndpointAttachment) AdapterForObject(ctx context.Context, re
...
+
+	if !u.GetDeletionTimestamp().IsZero() {
+		externalRef, found, _ := unstructured.NestedString(u.Object, "status", "externalRef")
+		if found && externalRef != "" {
+			klog.V(2).Infof("handling deletion for %v with externalRef %v", k8s.GetNamespacedName(u), externalRef)
+			id := &krm.ApigeeEndpointAttachmentIdentity{}
+			if err := id.FromExternal(externalRef); err != nil {
+				return nil, fmt.Errorf("error parsing externalRef %q: %w", externalRef, err)
+			}
+			return &ApigeeEndpointAttachmentAdapter{
```
justinsb: I think you can just use an `else` statement to avoid constructing this object twice

pkg/controller/direct/apigee/endpointattachment_controller.go
```diff
@@ -52,14 +52,43 @@ func (m *modelApigeeEndpointAttachment) AdapterForObject(ctx context.Context, re
...
+				attachmentsClient: gcpClient.endpointsAttachmentsClient(),
+				operationsClient:  gcpClient.operationsClient(),
+				desired:           obj,
+			}, nil
+		}
+	}
+
+	// For create/update, or for deletion where status.externalRef is not set,
+	// we need to resolve dependencies.
+	if err := ResolveApigeeEndpointAttachmentRefs(ctx, reader, obj); err != nil {
```
justinsb, line 55: Would be nice to replace this with the standard resolver

pkg/controller/direct/apigee/endpointattachment_controller.go
```diff
@@ -52,14 +52,43 @@ func (m *modelApigeeEndpointAttachment) AdapterForObject(ctx context.Context, re
 		return nil, fmt.Errorf("error converting to %T: %w", obj, err)
 	}
 
+	gcpClient, err := newGCPClient(ctx, m.config)
+	if err != nil {
+		return nil, err
+	}
+
+	if !u.GetDeletionTimestamp().IsZero() {
```
justinsb: I wonder if this logic can just be in `obj.GetIdentity` (given that it already reads status.externalRef, and constructs the identity from spec, and currently compares them)

yuwenma: I tried that, but it is a more complicated code change. Actually, I plan to improve the GetIdentity once #5302 is merged, it should not need the GetParentIdentity, and the GetIdentity can be further unified.

pkg/controller/direct/apigee/endpointattachment_controller.go
yuwenma: Added a todo here.

pkg/controller/direct/directbase/directbase_controller.go
yuwenma: Done. Will keep the object.

pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -336,10 +336,25 @@ func (r *reconcileContext) doReconcile(ctx context.Context, u *unstructured.Unst
...
+				// GCP resource has been orphaned and requires manual cleanup.
+				// While removing the finalizer is a fallback to prevent the Kubernetes object from
+				// getting stuck, leaving it allows the user to recognize and address the orphaned resource.
+				// For more details, see docs/designs/graceful-orphan-deletion.md.
+				// return false, r.handleDeleted(ctx, u)
+			}
 			logger.Info(unwrappedErr.Error(), "resource", k8s.GetNamespacedName(u))
 			return r.handleUnresolvableDeps(ctx, u, unwrappedErr)
 		}
-		return false, r.handleUpdateFailed(ctx, u, adapteErr)
```
justinsb, line 342: Is this removal accidental, I don't follow why we want it

tests/e2e/unified_test.go
```diff
@@ -446,7 +446,9 @@ func runScenario(ctx context.Context, t *testing.T, testPause bool, fixture reso
 				if testPause {
 					opt.SkipWaitForDelete = true
 				}
-				if os.Getenv("GOLDEN_REQUEST_CHECKS") != "" {
+				if os.Getenv("RANDOMIZE_DELETION_ORDER") != "" {
+					opt.DeleteInOrder = false
+				} else if os.Getenv("GOLDEN_REQUEST_CHECKS") != "" {
```
justinsb, line 451: Should we just not set GOLDEN_REQUEST_CHECKS, rather than introducing another env var?  I do like the explicit nature of RANDOMIZE_DELETION_ORDER, but maybe we keep that in the scripts and not the go code?

justinsb: Thinking more, a good rule could be that we should expose DeleteInOrder and we should expose GOLDEN_REQUEST_CHECKS, and GOLDEN_REQUEST_CHECKS should _not_ change DeleteInOrder

pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -336,10 +336,25 @@ func (r *reconcileContext) doReconcile(ctx context.Context, u *unstructured.Unst
 	}
 	if adapteErr != nil {
 		if unwrappedErr, ok := lifecyclehandler.CausedByUnresolvableDeps(adapteErr); ok {
+			// If the resource is being deleted, an unresolvable dependency means
+			// we have an orphan because the model couldn't build an adapter
```
gemmahou, line 340: `If the resource is being deleted, an unresolvable dependency means we have an orphan because the model couldn't build an adapter`

I do not think it's always true - Only when there's an unresolvable **parent** dependency, we can safely skip the API delete call and remove the KRM object, as the KRM object is considered "orphaned".

If a non-parent dependency (e.g., a reference to another peer resource) becomes unresolvable, it should not block/skip the deletion of the target resource. The current approach is concerning because it seems to skip the deletion of the underlying GCP resource of the target resource(by not sending delete API call) regardless of parent or non-parent dependency is missing.

Ideally we should not resolve non-parent dependency in Adaptor and limit the resolution to only Create and Update operations(example [PR](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2758)).

pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -336,10 +336,25 @@ func (r *reconcileContext) doReconcile(ctx context.Context, u *unstructured.Unst
...
+			// from status and has now failed to build one from spec.
+			if !u.GetDeletionTimestamp().IsZero() {
+				logger.Info("could not resolve dependencies for deletion and status.externalRef is missing; resource may be orphaned in GCP", "resource", k8s.GetNamespacedName(u), "error", unwrappedErr)
```
gemmahou, line 343: `could not resolve dependencies for deletion and status.externalRef is missing; resource may be orphaned in GCP`

I feel a bit confusing when I read this log message - Maybe "Could not resolve parent for deletion because parent does not exist or its status.externalRef is missing. Resource is considered as orphaned"(assuming the underlying GCP resource was deleted when the parent was removed). 

Or maybe we could simply use what we have in TF controller: "resource has been orphaned; no API call necessary".

gemmahou (commented):
I was looking into another deletion related issue(5390, cc'd you as well) and jumped to this PR. Left some thoughts, feel free to let me know your opinions.

pkg/controller/direct/directbase/directbase_controller.go
gemmahou: I found some examples showing that non-parent dependencies are still being resolved in the Adaptor, and there might be other early direct controllers where we resolve non-parent dependencies within the Adaptor as well. This need to be fixed(and prevented from future direct controller development) before we can safely mark "orphaned" resources by this logic.

Edit: I created an GH issue to track the work and listed affected direct controllers: https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/5620

dev/ci/presubmits/tests-e2e-random-deletion
```diff
@@ -0,0 +1,29 @@
...
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+cd ${REPO_ROOT}/
+
+# Only run tests for Apigee resources as example.
+# Eventually we want to run this random deletion test for all resources.
+export ONLY_TEST_APIGROUPS="apigee.cnrm.cloud.google.com"
+
+
+export RANDOMIZE_DELETION_ORDER=1
```
cheftako, line 28: This feels like its providing the illusion of coverage without providing coverage. In other areas where actual coverage is so large that coverage isn't practical a fuzzer makes sense.

PR #5383, opened 2025-10-08
FirestoreDatabase: move to identity pattern, support export

justinsb:


pkg/controller/direct/firestore/firestoredatabase_controller.go
```diff
@@ -39,30 +39,25 @@ import (
 )
 
 const (
-	ctrlName      = "firestoredatabase-controller"
-	serviceDomain = "//firestore.googleapis.com"
+	ctrlName = "firestoredatabase-controller"
```
yuwenma, line 42: I think we don't need this.

pkg/controller/direct/firestore/firestoredatabase_controller.go
```diff
@@ -39,30 +39,25 @@ import (
...
 )
 
 func init() {
 	registry.RegisterModel(krm.FirestoreDatabaseGVK, NewModel)
 }
 
 func NewModel(ctx context.Context, config *config.ControllerConfig) (directbase.Model, error) {
```
yuwenma, line 49: you may want to rename this to NewKindModel so we can add more controllers to firestore in the future.

PR #5397, opened 2025-10-10
feat: Add API, fuzzer and mapper for BigTableTable

yuwenma:
### BRIEF Change description

This PR adds the API for BigTableTable (TF-based Beta) and guarantee backward compatibility.
It also fixes the following frictions when unifying the api mapper and fuzzer for the BigTable service:
* A bad naming for a reference field in BigTableCluster that mapper-generator cannot recoganize.
* Removed manual mapper for authorizedview
* Removed manual mapper for Backup
* Removed manual mapper for cluster
* Removed manual mapper for LogicalView
* Removed manual mapper for MaterializedView
* Removed stale fuzz test (roundtrip_test.go)
* Removed unnecessary manual functions in other resources 
* Add fuzzer for table 
* Fix LogicalView mapper and identity. The GCP introduces a special case that the Create/Update method  asks for a different proto (`LogicalViewInfo`) than the resource itself (`LogicalView`). Previous code introduces the KCC identity in the mapper and bypass the fuzz checks. 
* Fix Tabel legacy fields (renames, type changes, etc) to keep backward compatibility when migrate the legacy controller to direct (proto in 1.40)

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/apis/bigtable/v1alpha1/cluster_types.go
+++ b/apis/bigtable/v1alpha1/cluster_types.go
@@ -61,7 +61,7 @@ type Cluster_EncryptionConfig struct {
 	//  Values are of the form
 	//  `projects/{project}/locations/{location}/keyRings/{keyring}/cryptoKeys/{key}`
 	// +kcc:proto:field=google.bigtable.admin.v2.Cluster.EncryptionConfig.kms_key_name
-	KMSKeyNameRef *refs.KMSCryptoKeyRef `json:"kmsKeyNameRef,omitempty"`
+	KMSKeyNameRef *refs.KMSCryptoKeyRef `json:"kmsKeyRef,omitempty"`
 }
 
 // BigtableClusterSpec defines the desired state of BigtableCluster
--- a/pkg/controller/direct/bigtable/logicalview_controller.go
+++ b/pkg/controller/direct/bigtable/logicalview_controller.go
@@ -27,6 +27,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
 
+	// TODO: This gcp should not be used, it takes a different proto definition than the one defined in kcc apis.
 	gcp "cloud.google.com/go/bigtable"
 	bigtablepb "cloud.google.com/go/bigtable/admin/apiv2/adminpb"
 	"google.golang.org/api/option"
--- a/pkg/controller/direct/bigtable/table_fuzzer.go
+++ b/pkg/controller/direct/bigtable/table_fuzzer.go
@@ -36,8 +36,6 @@ func bigtableTableFuzzer() fuzztesting.KRMFuzzer {
 	f.SpecFields.Insert(".deletion_protection")
 	f.SpecFields.Insert(".row_key_schema")
 
-	f.StatusFields.Insert(".restore_info")
-
 	f.UnimplementedFields.Insert(".name")
 	f.UnimplementedFields.Insert(".automated_backup_policy")
 	f.UnimplementedFields.Insert(".granularity")
@@ -45,6 +43,7 @@ func bigtableTableFuzzer() fuzztesting.KRMFuzzer {
 	f.UnimplementedFields.Insert(".column_families")
 	f.UnimplementedFields.Insert(".row_key_schema")
 	f.UnimplementedFields.Insert(".change_stream_config") // The value has to be a duration
+	f.UnimplementedFields.Insert(".restore_info")         // Fields under restore_info are not supported yet
 
 	return f
 }
--- a/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_bigtableclusters.bigtable.cnrm.cloud.google.com.yaml
+++ b/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_bigtableclusters.bigtable.cnrm.cloud.google.com.yaml
@@ -112,7 +112,7 @@ spec:
                 description: Immutable. The encryption configuration for CMEK-protected
                   clusters.
                 properties:
-                  kmsKeyNameRef:
+                  kmsKeyRef:
                     description: 'Describes the Cloud KMS encryption key that will
                       be used to protect the destination Bigtable cluster. The requirements
                       for this key are: 1) The Cloud Bigtable service account associated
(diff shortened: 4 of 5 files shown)


apis/bigtable/v1alpha1/cluster_types.go
```diff
@@ -61,7 +61,7 @@ type Cluster_EncryptionConfig struct {
 	//  Values are of the form
 	//  `projects/{project}/locations/{location}/keyRings/{keyring}/cryptoKeys/{key}`
 	// +kcc:proto:field=google.bigtable.admin.v2.Cluster.EncryptionConfig.kms_key_name
-	KMSKeyRef *refs.KMSCryptoKeyRef `json:"kmsKeyRef,omitempty"`
+	KMSKeyNameRef *refs.KMSCryptoKeyRef `json:"kmsKeyNameRef,omitempty"`
```
justinsb: We need to write down (and ideally have a test) for our desired form of references.

pkg/controller/direct/bigtable/logicalview_controller.go
```diff
@@ -145,17 +145,34 @@ func (a *LogicalViewAdapter) Find(ctx context.Context) (bool, error) {
 	return true, nil
 }
 
+func convertLogicalViewToLogicalViewInfo(in *bigtablepb.LogicalView) *gcp.LogicalViewInfo {
```
justinsb, line 149: This needs a comment.  Is this because there is no (published) proto API?

pkg/controller/direct/bigtable/roundtrip_test.go
```diff
@@ -1,373 +0,0 @@
-// Copyright 2024 Google LLC
```
justinsb, line 1: Yay!

pkg/controller/direct/bigtable/table_fuzzer.go
```diff
@@ -0,0 +1,50 @@
...
+
+	f.StatusFields.Insert(".restore_info")
+
+	f.UnimplementedFields.Insert(".name")
+	f.UnimplementedFields.Insert(".automated_backup_policy")
+	f.UnimplementedFields.Insert(".granularity")
+	f.UnimplementedFields.Insert(".cluster_states")
+	f.UnimplementedFields.Insert(".column_families")
+	f.UnimplementedFields.Insert(".row_key_schema")
+	f.UnimplementedFields.Insert(".change_stream_config") // The value has to be a duration
```
justinsb, line 45: So FYI this is what the filter lets you fix.  You can change the value to be a duration

pkg/controller/direct/bigtable/table_mapper.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
+)
+
+func BigtableTableSpec_v1beta1_FromProto(mapCtx *direct.MapContext, in *pb.Table) *krm.BigtableTableSpec {
+	if in == nil {
+		return nil
+	}
+	out := &krm.BigtableTableSpec{}
+
+	// Note: Bigtable proto 1.38 -> 1.40 changed the ColumnFamily from a slice to a map
```
justinsb: Really?

https://pkg.go.dev/cloud.google.com/go/bigtable@v1.38.0/admin/apiv2/adminpb#Table

https://pkg.go.dev/cloud.google.com/go/bigtable@v1.40.0/admin/apiv2/adminpb#Table

pkg/controller/direct/bigtable/table_mapper.go
```diff
@@ -0,0 +1,142 @@
...
+			if v == nil {
+				continue
+			}
+			if cf := TableColumnFamily_v1beta1_FromProto(mapCtx, v); cf != nil {
+				out.ColumnFamily = append(out.ColumnFamily, cf)
+			}
+		}
+	}
+
+	// Note: Bigtable proto 1.38 -> 1.40 changed the ChangeStreamRetention from a single field to a struct
```
justinsb: I don't think this is accurate either

pkg/controller/direct/bigtable/table_mapper.go
```diff
@@ -0,0 +1,142 @@
...
+	if changeStreamConfig := ChangeStreamConfig_v1beta1_FromProto(mapCtx, in.GetChangeStreamConfig()); changeStreamConfig != nil {
+		out.ChangeStreamRetention = changeStreamConfig.RetentionPeriod
+	}
+
+	// Note: Bigtable proto 1.38 -> 1.40 changed the DeletionProtection type from string to bool; we handle the conversion.
```
justinsb: Nor this one

pkg/controller/direct/bigtable/logicalview_controller.go
yuwenma: Agree we should have a comment.  This PR is just to fix the previous mapper function which mixed the code for proto/krm mapping, two GCP proto mapping and some  identity assignment. We want the mapper to be the  standard KCC format so the fuzz test can verify.

I don't have a good answer about why the GCP server requires another proto `LogicalViewInfo`  for the `LogicalView` CREATE call. It looks like  the `LogicalView` is published.  
@georgecma Can you share the context so we can better deal with this special case?

pkg/controller/direct/bigtable/table_mapper.go
yuwenma: You are right, it is not from 1.38 (I deduced that from the TF field name but didn't verify) 
The problem is that this KCC field is called `ColumnFamily` (slice, TF-based) but the proto name is `ColumnFamilies` when we migrate them to direct, we have to manually convert them.  The claim about  "from 1.38 to 1.40" is not accurate, it comes from the `go.mod` change which could also come from the mockgcp bumping proto versions rather than from forking the Terraform-google-provider.

pkg/controller/direct/bigtable/logicalview_controller.go
georgecma: cc @bhshkh as our go client person

apis/bigtable/v1alpha1/cluster_types.go
yuwenma: It's written down in docs/ai/add-resource-references.md (to let Gemini discovering and defining a reference field).  We also have docs/develop-resources/api-conventions/resource-reference.md to talk about the reference field naming requirements.

I like that we use generate.sh to enforce the mapper-generator here and moving the code to automation. But this also sets a stricter naming rule for the KRM field name (that has to match the proto with "Ref" suffix). I've updated both files to include that. https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5491

FYI, this is a screenshot about what happens if we use the <Kind>Ref as the field name 
https://screenshot.googleplex.com/4dhf3w8tMxNcMYz.png

pkg/controller/direct/bigtable/logicalview_controller.go
justinsb: If the canonical representation uses ProtoA, but the create method uses ProtoB, then that's a perfectly valid reason for the controller to have to map from ProtoA to ProtoB.  I think just a comment saying that should be sufficient (e.g. "GET returns LogicalView, but POST requires LogicalViewInfo, so we use LogicalView as our canonical representation but map to LogicalViewInfo for create method" - assuming that is true)

justinsb (approved):

justinsb:
LGTM - please add that comment (but can be done in a later PR)

/approve
/lgtm

PR #5398, opened 2025-10-10
FirestoreBackupSchedule: CRD/Mapper/Fuzzer

justinsb:
- **FirestoreDatabaseBackupSchedule: CRD/Mapper/Fuzzer**
  

- **autogen: FirestoreDatabaseBackupSchedule**


pkg/controller/direct/firestore/firestorefield_fuzzer.go
```diff
@@ -32,6 +32,17 @@ func firestoreFieldFuzzer() fuzztesting.KRMFuzzer {
 		FirestoreFieldSpec_v1alpha1_FromProto, FirestoreFieldSpec_v1alpha1_ToProto,
 		FirestoreFieldObservedState_v1alpha1_FromProto, FirestoreFieldObservedState_v1alpha1_ToProto,
 	)
+	f.FilterSpec = func(in *pb.Field) {
```
yuwenma: Just want to learn for myself, what problem does this function solve?

justinsb: Generally filters here are an attempt to deal with special case values that are not valid (e.g. must follow a certain format).

It's a good question in this case, I'll add a comment at least.

justinsb: Going to split the filter out into its own PR, it doesn't look like I need it...

apis/firestore/v1alpha1/backupschedule_types.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+var FirestoreBackupScheduleGVK = GroupVersion.WithKind("FirestoreBackupSchedule")
+
+// FirestoreBackupScheduleSpec defines the desired state of FirestoreBackupSchedule
+// +kcc:spec:proto=google.firestore.admin.v1.BackupSchedule
+type FirestoreBackupScheduleSpec struct {
+	/* The database that this resource belongs to. */
+	// +required
+	DatabaseRef v1beta1.FirestoreDatabaseRef `json:"databaseRef"`
```
yuwenma, line 30: Is this the parent? I like the new naming convention `parentDatabaseRef`

justinsb: This one is alpha, and we can absolutely do it.  parentDatabaseRef is clearer here, though it's not too surprising.

Do we want to change all the _other_ types?  Would we rename `projectRef` -> `parentProjectRef`?  It feels like you're adding a lot of work, and if we're not consistent it may end up more confusing.

I don't mind either way, but do you want to create a test for whatever your choose?  I would imagine the rule would be either "no fields should be parent.*Ref" or "exactly one field should be parent.*Ref (with a known list of exceptions for top-level types)"

justinsb:
Thanks for review!

I'll work on a test for parent vs not in another PR, this is still alpha so we have time

/hold cancel

PR #5402, opened 2025-10-10
FirestoreDocument: CRD/Mapper/Fuzzer

justinsb:
- **fuzzers: support filtering of objects after generation**
  

- **controllerbuilder: support multiple services in mapper**
  

- **controllerbuilder: support multiple proto packages in generated mapper**

- **controllerbuilder: tolerate map[string]apiextensionsv1.JSON fields**
  

- **FirestoreDocument: CRD/Mapper/Fuzzer**
  

- **autogen: FirestoreDocument**

### DIFF
--- a/pkg/crd/fielddesc/fielddesc.go
+++ b/pkg/crd/fielddesc/fielddesc.go
@@ -126,8 +126,12 @@ func objectToDescription(props apiextensions.JSONSchemaProps, parent FieldDescri
 			"object":  true,
 		}
 		valueType := props.AdditionalProperties.Schema.Type
-		if _, ok := supportedTypes[valueType]; !ok {
-			panic("only support maps of boolean, integer, string, number, and object types")
+		if _, ok := supportedTypes[valueType]; ok {
+			fd.Type = fmt.Sprintf("map (key: string, value: %v)", valueType)
+		} else if props.AdditionalProperties.Schema.XPreserveUnknownFields != nil && *props.AdditionalProperties.Schema.XPreserveUnknownFields {
+			fd.Type = fmt.Sprintf("map (key: string, value: %v)", "any")
+		} else {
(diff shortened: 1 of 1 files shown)


justinsb:
/assign @cheftako 

Not as clean as it could be, because of the need to fix/improve our generators to cope with multiple proto namespaces and with the map[string]JSON

apis/firestore/v1alpha1/document_types.go
```diff
@@ -0,0 +1,107 @@
...
+// FirestoreDocumentSpec defines the desired state of FirestoreDocument
+// +kcc:spec:proto=google.firestore.v1.Document
+type FirestoreDocumentSpec struct {
+	// The FirestoreDocument name. If not given, the metadata.name will be used.
+	ResourceID *string `json:"resourceID,omitempty"`
+
+	// // The resource name of the document, for example
+	// //  `projects/{project_id}/databases/{database_id}/documents/{document_path}`.
+	// // +kcc:proto:field=google.firestore.v1.Document.name
+	// Name *string `json:"name,omitempty"`
```
cheftako, line 34: Not objecting to commenting out a field we don't want (?yet) in the CRD. Would be good to leave a bread crumb on why?

dev/tools/controllerbuilder/pkg/commands/generatemapper/generatemappercommand.go
```diff
@@ -128,14 +138,21 @@ func RunGenerateMapper(ctx context.Context, o *GenerateMapperOptions) error {
 	generatedFileAnnotation := &annotations.FileAnnotation{
 		Key: "+generated:mapper",
 		Attributes: map[string][]string{
-			"proto.service": {o.ServiceName},
+			"proto.service": o.ServiceNames,
```
cheftako, line 141: Seems odd to remove the curly braces on this line but not the next two.

PR #5408, opened 2025-10-10
fix: changing reconciliation interval should trigger reconciliation

### LINKED ISSUE #5405: Should be possible to restart reconciliation
We have had reports that setting the reconciliation interval to 0 (with `cnrm.cloud.google.com/reconcile-interval-in-seconds: "0"`) stops reconciliation (as expected), but that raising the value again does not cause reconciliation to resume.

This does not seem like the desired behaviour.

justinsb:
- **Add test for changing reconciliation interval annotation**
  Issue #5405
  

- **fix: changing reconciliation interval should trigger reconciliation**
  Fix #5405

### DIFF
--- a/pkg/controller/predicate/predicate.go
+++ b/pkg/controller/predicate/predicate.go
@@ -15,12 +15,15 @@
 package predicate
 
 import (
+	"context"
 	"reflect"
+	"strconv"
 
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/util/slice"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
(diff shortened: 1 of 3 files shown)


justinsb:
/remove-wip

/assign @cheftako 

I think we should make a decision on this behaviour

pkg/controller/predicate/predicate.go
```diff
@@ -65,6 +65,11 @@ func (UnderlyingResourceOutOfSyncPredicate) Update(e event.UpdateEvent) bool {
 		return true
 	}
 
+	// Changes to the reconcile interval annotation should trigger a reconcile
+	if e.ObjectOld.GetAnnotations()[k8s.ReconcileIntervalInSecondsAnnotation] != e.ObjectNew.GetAnnotations()[k8s.ReconcileIntervalInSecondsAnnotation] {
```
cheftako: Seems like this triggers a reconcile on ANY change of interval. It seems like it should only trigger a reconcile if objectNew interval is non zero?

cheftako: Otherwise looks good

cheftako: Rephrasing earlier comment. If the user is specifically setting the reconcile interval to zero, they are trying to suspend reconciliation. If the user is trying to suspend reconciliation, they I don't think we should trigger reconciliation.

justinsb: Good point, changed this.  But now a new problem:  does `reconcile-interval-in-seconds=0` mean "stop all reconciliation", or does it mean "no drift detection".  In other words, if the user changes a spec, should we reconcile?

My 2c: it's controlling drift detection.  If we set `reconcile-interval-in-seconds=60` but then change the spec, we don't wait 60 seconds.  So we we should turn off drift detection, but we should still reconcile on spec changes.

cheftako: Short version - we should still reconcile on spec changes.

Long versions - My guess is that this is mostly a control for ongoing aggregate quota usage. Going to 0 means don't spend quota to do drift detection for this resource. If the customer makes changes, they would still expect  it to be actuated, otherwise why make a change? Unless your making frequent spec changes, the quota usage is a drop in the bucket.

PR #5409, opened 2025-10-10
test: stabilize flaky firewallpolicyrule acquisition test

gemmahou:
### BRIEF Change description

Intend to stabilize flaky firewallpolicyrule acquisition test
Issue: https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/5410

I was not able to reproduce the test failure locally, but I did see the flakiness and failed several times on GH presubmit job.

I've triggered the job 3 times and they all passed: https://screenshot.googleplex.com/8HST4j8bqR3gcLZ.png

Here's the root cause and why it only happens on GH job that Gemini provided:
```
The extra GET request is a symptom of a subtle race condition between the Kubernetes API server's state and the controller's local cache.

Here's the sequence of events that causes it:

 1. Test Confirms Parent is Ready: The test script applies the parent ComputeFirewallPolicy and then waits for it to be
ready. This "wait" function checks the resource's status by polling the API server directly. Once it sees Ready=True, the
test proceeds.

 3. Child is Created Immediately: The test immediately applies the child ComputeFirewallPolicyRule.

 4. Controller Reads from Cache: The ComputeFirewallPolicyRule controller wakes up to reconcile the new child
resource. To get the parent's required ID, it reads the ComputeFirewallPolicy resource from its local, in-memory cache,
not from the API server.

 5. The Race: There is a very small, non-zero delay for the parent's updated status to propagate from the API server to
the child controller's local cache.
     * On your fast local machine: The cache usually syncs before the child controller needs it. The controller finds the
parent's ID in the cache on the first try.
     * On a slower GitHub runner: The child controller often reads its cache before the parent's status update has arrived.
It sees a stale version of the parent resource without the necessary ID.

When the controller reads the stale cache, its logic correctly determines the dependency isn't ready and requeues the
request. The reconciliation succeeds moments later, but this requeue-and-retry cycle alters the timing of operations. This
new timing causes a different pattern of HTTP requests (the extra GET) compared to the "perfect" run that was used to
generate the golden log file, resulting in the test failure.
```

While I'm not familiar with the concept of "controller's local cache", any further clarifications would be appreciated. 

This test started failing recently without any code changes to this resource. The only changes would be we have made
100+ presubmit jobs running in parallel, so it probably makes sense that:
```
On GitHub the environment is different. The underlying Linux kernel, its scheduler, and the load from other processes are
all variables. This can cause the Go runtime to schedule the goroutines in a slightly different order. All it takes is for the 
child FirewallPolicyRule controller's goroutine to be scheduled a few milliseconds before the parent's cache update is 
processed, and the bug is triggered.
``` 

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


tests/e2e/testdata/scenarios/acquisition/firewallpolicyrule/script.yaml
```diff
@@ -36,7 +36,15 @@ spec:
...
-TEST: APPLY
+# pausing the test execution, ensuring the controller's cache has time to sync with the API server before the child
+# resource is created.
+TEST: SLEEP
+apiVersion: compute.cnrm.cloud.google.com/v1beta1
+kind: ComputeFirewallPolicyRule
+metadata:
+  name: firewallpolicyrule-${uniqueId}
+
+---
```
acpana, line 47: Ah I didn't know we have `SLEEP` test case.

I think this works for now BUT Walter did ask me to use a wait.Poll for a similar-ish race on one of my tests! e.g.:

https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5288/files#diff-880e10d431db58a0b2411d1fb8dc426dca5aa8fda01ddfecebabb0761d30d0feR298-R303

I'm going to leave it up to him on whether to use it in this case!

tests/e2e/script_test.go
```diff
@@ -251,6 +251,8 @@ func TestE2EScript(t *testing.T) {
 						// Allow some time for reconcile
 						// Maybe we should instead wait for observedState
 						time.Sleep(2 * time.Second)
```
cheftako, line 253: wait.Poll is a better pattern that a plain time.Sleep. It allows us to wait longer in worst case, less in the normal case and have a more descriptive error about what went wrong.

tests/e2e/testdata/scenarios/acquisition/firewallpolicyrule/script.yaml
cheftako: +1. Sleep is a bit of an anti-pattern. It make the race condition less likely, usually at a prohibitive expense compared to value. wait.Poll or wait.Until usually work better. As above they allow a shorter average sleep with a longer maximal sleep. If the wait condition is set up correctly, they usually provide a better error message. (Eg. Waited 3o seconds for resource X to become 'ready' but it is still in condition 'y'.)

cheftako: Not wanting the perfect to be the enemy of the good. We should try to switch our standard fix pattern.

gemmahou: I'll hold off on this. I intend to update the use of `sleep` with either Alex's linked `waitForReconciliationAfterPatch` function above or the `watch for observedGeneration` functionality Justin mentioned.

PR #5416, opened 2025-10-13
feat: Add api, fuzzer and mapper for TagsTagBinding

yuwenma:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


apis/tags/v1beta1/tagbinding_identity.go
```diff
@@ -0,0 +1,83 @@
...
+		tagValue: tagValue,
+	}, nil
+}
+
+func resolveParent(ctx context.Context, reader client.Reader, obj *TagsTagBinding) (string, error) {
+	if obj.Spec.ParentRef.External != "" {
+		return obj.Spec.ParentRef.External, nil
+	}
+	if obj.Spec.ParentRef.Name != "" {
+		// TODO: This is not quite right, we need to resolve the project number.
```
justinsb: I have been calling common.NormalizeRefs first (and then we can treat External == "" as an error)

justinsb: Ah - a meta point - this is why it's problematic to introduce identity & reference here.  AFAICT this code is currently unused and therefore untested?

apis/tags/v1beta1/tagbinding_reference.go
```diff
@@ -0,0 +1,66 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ refsv1beta1.ExternalNormalizer = &TagsTagValueRef{}
+
+// NormalizedExternal provision the "External" value for other resource that depends on TagsTagValue.
+// If the "External" is given in the other resource's spec.TagsTagValueRef, the given value will be used.
+// Otherwise, the "Name" and "Namespace" will be used to query the actual TagsTagValue object from the cluster.
+func (r *TagsTagValueRef) NormalizedExternal(ctx context.Context, reader client.Reader, otherNamespace string) (string, error) {
+	if r.External != "" && r.Name != "" {
```
justinsb, line 49: Can we use the common function?

apis/tags/v1beta1/tagbinding_identity.go
yuwenma: Right, it is unused and untested. It would be changed in the next PR (controller)

yuwenma:
> CRD changes (or lack of changes!) look good
> 
> Identity/Reference code looks more problematic, and is not tested in this PR. Split into another PR. Or maybe we can find a way to test it now, if you really want to do it in this PR.

I want to merge this PR and have the code template fixed in a separate PR, and have this identity/reference fixed in the next controller PR. Theoretically the identity and reference in this PR shows that they come from the right place: generate.sh if the file does not exist., which uses the code template.

justinsb (approved):

PR #5420, opened 2025-10-13
feat:operator:kcc: MultiClusterLease integration

acpana:
* Integrates into operator and kcc https://github.com/gke-labs/multicluster-leader-election. 
* Automated testing is happening  in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5490

### DIFF
--- a/pkg/controller/kccmanager/kccmanager.go
+++ b/pkg/controller/kccmanager/kccmanager.go
@@ -47,6 +47,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/leaderelection"
 	"k8s.io/klog/v2"
 	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -99,11 +100,11 @@ type Config struct {
 	MultiClusterLease bool
 }
 
-func setUpMultiClusterLease(ctx context.Context, restConfig *rest.Config, opts manager.Options) (manager.Options, error) {
+func setUpMultiClusterLease(ctx context.Context, restConfig *rest.Config, scheme *runtime.Scheme) (*leaderelection.LeaderElectionConfig, error) {
 	// Create a temporary client to read the ConfigConnector object for leader election config.
-	c, err := crclient.New(restConfig, crclient.Options{Scheme: opts.Scheme})
+	c, err := crclient.New(restConfig, crclient.Options{Scheme: scheme})
 	if err != nil {
-		return manager.Options{}, fmt.Errorf("error creating temporary client: %w", err)
+		return nil, fmt.Errorf("error creating temporary client: %w", err)
 	}
 
 	// Get the ConfigConnector object.
@@ -125,15 +126,39 @@ func setUpMultiClusterLease(ctx context.Context, restConfig *rest.Config, opts m
 				leaseSpec.GlobalLockName,
 				15*time.Second,
 			)
-			opts.LeaderElectionResourceLock = leaseSpec.GlobalLockName
-			opts.LeaderElection = true
-			opts.LeaderElectionNamespace = leaseSpec.Namespace
-			opts.LeaderElectionID = leaseSpec.LeaseName
-			opts.LeaderElectionResourceLockInterface = lock
+			return &leaderelection.LeaderElectionConfig{
+				Lock:          lock,
+				LeaseDuration: 15 * time.Second,
+				RenewDeadline: 10 * time.Second,
+				RetryPeriod:   2 * time.Second,
+				Name:          leaseSpec.LeaseName,
+				Callbacks: leaderelection.LeaderCallbacks{
+					OnStoppedLeading: func() {
+						klog.ErrorS(nil, "leaderelection lost")
+						klog.FlushAndExit(klog.ExitFlushTimeout, 1)
+					},
+				},
+			}, nil
 		}
 	}
 
-	return opts, nil
+	return nil, nil
+}
+
+type leaderElectionManager struct {
+	manager.Manager
+	leConfig *leaderelection.LeaderElectionConfig
+}
+
+func (m *leaderElectionManager) Start(ctx context.Context) error {
+	m.leConfig.Callbacks.OnStartedLeading = func(ctx context.Context) {
+		klog.Infof("started leading")
+		if err := m.Manager.Start(ctx); err != nil {
+			klog.Fatalf("error running manager: %v", err)
+		}
+	}
+	leaderelection.RunOrDie(ctx, *m.leConfig)
+	return nil
 }
 
 // Creates a new controller-runtime manager.Manager and starts all of the KCC controllers pointed at the
@@ -158,11 +183,15 @@ func New(ctx context.Context, restConfig *rest.Config, cfg Config) (manager.Mana
 		return nil, fmt.Errorf("error adding schemes: %w", err)
 	}
 
+	var leConfig *leaderelection.LeaderElectionConfig
 	if cfg.MultiClusterLease {
-		opts, err = setUpMultiClusterLease(ctx, restConfig, opts)
+		leConfig, err = setUpMultiClusterLease(ctx, restConfig, opts.Scheme)
 		if err != nil {
 			return nil, fmt.Errorf("error setting up multi-cluster leader election: %w", err)
 		}
+	}
+	if leConfig != nil && opts.LeaderElection {
+		return nil, fmt.Errorf("error validating leader election config")
 	}
 
 	// only cache CC and CCC resources
@@ -269,6 +298,10 @@ func New(ctx context.Context, restConfig *rest.Config, cfg Config) (manager.Mana
 	// all our resources.
 	if err := registration.AddDefaultControllers(ctx, mgr, &rd, controllerConfig); err != nil {
 		return nil, fmt.Errorf("error adding registration controller: %w", err)
+	}
+
+	if leConfig != nil {
+		return &leaderElectionManager{Manager: mgr, leConfig: leConfig}, nil
 	}
 	return mgr, nil
 }


.github/license-lint-config.yml
```diff
@@ -41,4 +41,6 @@ allowlisted_modules:
   # Apache 2.0: https://github.com/apparentlymart/go-textseg/blob/5b41aa275ccaac4ccaa91729a8ee92e7eac9c728/LICENSE
   - github.com/apparentlymart/go-textseg/v13
   # Apache 2.0: https://github.com/apparentlymart/go-textseg/tree/72b78f42484ddc3f58858f794da1771fb9559ad0/LICENSE
-  - github.com/apparentlymart/go-textseg/v15
\ No newline at end of file
+  - github.com/apparentlymart/go-textseg/v15
+  # todo
+  - github.com/gke-labs/multicluster-leader-election
```
acpana: will need to specify the license in https://github.com/gke-labs/multicluster-leader-election and follow up here

cmd/manager/main.go
```diff
@@ -142,28 +154,12 @@ func main() {
...
-	mgr, err := kccmanager.New(ctx, restCfg, controllersCfg)
-	if err != nil {
-		return nil, fmt.Errorf("error creating manager: %w", err)
+	mgrErr := mgr.Start(stop)
+	if mgrErr != nil {
+		logging.Fatal(mgrErr, "error during manager execution.")
+	} else {
+		// err is nil
+		// todo acpana add more defense in depth here
+		logging.ExitInfo("might've lost leader election")
```
acpana: walter had suggested adding some defense in depth here where we as the kccmanager process check our leadership status and bail out if we are no longer a leader.

however, AFAICT, we get that guarantee for free if mgr.Start returns, meaning we are no longer a leader if err is nil.

go.mod
```diff
@@ -80,6 +80,7 @@ require (
 	github.com/cenkalti/backoff v2.2.1+incompatible
 	github.com/fatih/color v1.18.0
 	github.com/ghodss/yaml v1.0.0
+	github.com/gke-labs/multicluster-leader-election v0.0.0-20250923220528-0bf41dc7fecc
```
acpana: All go mod changes are coming out of there. Unfortunately, I think bumping the controller-runtime lib really upset the `pkg/cli/preview` as it introduces an interface change. I added an unimplemtend func for now for the preview cli.

cmd/manager/main.go
```diff
@@ -106,7 +106,19 @@ func main() {
 	// Set client site rate limiter to optimize the configconnector re-reconciliation performance.
 	ratelimiter.SetMasterRateLimiter(restCfg, rateLimitQps, rateLimitBurst)
 	logger.Info("Creating the manager")
-	mgr, err := newManager(ctx, restCfg, scopedNamespace, userProjectOverride, billingProject)
+	controllersCfg := kccmanager.Config{
+		ManagerOptions: manager.Options{
+			Cache: cache.Options{
+				DefaultNamespaces: map[string]cache.Config{
+					scopedNamespace: {},
```
gemmahou: nit: I see this is a copy of existing code, but I'm curious shouldn't this be scopedNamespace: scopedNamespace so it takes the flag value?

operator/pkg/controllers/configconnector/experiments.go
```diff
@@ -0,0 +1,86 @@
...
+		log.Info("enabling multi-cluster leader election for StatefulSet", "name", item.GetName())
+		if err := item.MutateContainers(func(container map[string]interface{}) error {
+			name, found, _ := unstructured.NestedString(container, "name")
+			if !found || name != "manager" {
+				return nil
+			}
+
+			// Add --multi-cluster-election=true flag
+			args, _, _ := unstructured.NestedStringSlice(container, "args")
+			args = append(args, "--multi-cluster-election=true")
```
gemmahou, line 78: QQ: just for self-education purpose, how this flag will be used to enable the feature? asking because I don't see we pass it anywhere, and seems like we check the existence of "cc.Spec.Experiments.LeaderElection" to determine whether to enable the feature(line 134-148 in kccmanager.go)

cmd/manager/main.go
acpana: actually no I think this is a great catch! I tried to do away with that level of indirection (newManager) func but I will revert that refactor to keep things less confusing and make sure I don't miss things like these!

.github/license-lint-config.yml
justinsb: I approved so we should be able to rebase

cmd/manager/main.go
justinsb: Is that actually what happens with the default lease objects?  If/when we lose the lease then Start simply returns?

operator/config/crd/bases/core.cnrm.cloud.google.com_configconnectors.yaml
```diff
@@ -66,6 +66,37 @@ spec:
                   It's recommended to use `googleServiceAccount` when running ConfigConnector in Google Kubernetes Engine (GKE) clusters with Workload Identity enabled.
                   This field cannot be specified together with `googleServiceAccount`.
                 type: string
+              experiments:
```
justinsb, line 69: Is this going to be per-namespace or per-cluster?

operator/pkg/apis/core/v1beta1/configconnector_types.go
```diff
@@ -66,6 +66,34 @@ type ConfigConnectorSpec struct {
...
+
+// MultiClusterLeaseSpec defines the configuration for a multi-cluster lease.
+type MultiClusterLeaseSpec struct {
+	// The name of the MultiClusterLease object that KCC will create.
+	LeaseName string `json:"leaseName"`
+	// The namespace where the MultiClusterLease object will be created.
+	Namespace string `json:"namespace"`
+	// The unique name of the global lock, which must be shared by all KCC replicas
+	// and workloads across all clusters that are part of the same election.
+	GlobalLockName string `json:"globalLockName"`
```
justinsb, line 97: We _could_ allow limited templating in this name, e.g. `gs://my-cluster-locks/kcc/prod/{namespace}`

operator/pkg/controllers/configconnector/experiments.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (r *Reconciler) applyMultiClusterLeaderElection(ctx context.Context, obj *manifest.Objects) error {
+	log := log.FromContext(ctx)
+	for _, item := range obj.Items {
+		if item.Kind != "StatefulSet" || !strings.HasPrefix(item.GetName(), "cnrm-controller-manager") {
```
justinsb: Nit: we might want to extract this if statement into a shared function, it's both a little less fragile / more maintainable and it documents more clearly what we are doing.

e.g. `IsControllerManagerStatefulSet`

operator/pkg/controllers/configconnector/experiments.go
```diff
@@ -0,0 +1,86 @@
...
+			continue
+		}
+
+		log.Info("enabling multi-cluster leader election for StatefulSet", "name", item.GetName())
+		if err := item.MutateContainers(func(container map[string]interface{}) error {
+			name, found, _ := unstructured.NestedString(container, "name")
+			if !found || name != "manager" {
```
justinsb: Nit: if !found, then name will be empty, so you don't need to check found

pkg/controller/kccmanager/kccmanager.go
```diff
@@ -36,20 +36,21 @@ import (
...
 	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
+	"k8s.io/klog/v2"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
-
-	// Register direct controllers
-	_ "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
```
justinsb, line 52: I think you need this (?).  If we don't, that's interesting and it should be a separate PR I think

pkg/controller/kccmanager/kccmanager.go
```diff
@@ -100,7 +101,9 @@ type Config struct {
 // This serves as the entry point for the in-cluster main and the Borg service main. Any changes made should be done
 // with care.
 func New(ctx context.Context, restConfig *rest.Config, cfg Config) (manager.Manager, error) {
+	krmtotf.SetUserAgentForTerraformProvider()
```
justinsb: Also looks unrelated to the MCL integration (?)

pkg/controller/kccmanager/kccmanager.go
```diff
@@ -113,6 +116,39 @@ func New(ctx context.Context, restConfig *rest.Config, cfg Config) (manager.Mana
 		return nil, fmt.Errorf("error adding schemes: %w", err)
 	}
 
+	// Create a temporary client to read the ConfigConnector object for leader election config.
```
justinsb, line 104: Hmmm, this is clever but I just don't know.  We should at least split it out into its own function to avoid accidentally reusing the client.

I think because we aren't going to watch this object, we should rely on the operator to manage the restarts / rollout.

Maybe invent something like a JDBC URL for leader election syntax to make it a one liner, e.g.

`gs://<bucket>/<path>?timeout=15s&namespace=foo`

I'm also not sure who should create the lease objects, should we simply accept a pointer to a pre-configured lease object?

go.mod
justinsb: That sounds OK.  It is a good reason _not_ to update library versions in libraries.  Also a good reason to stick to keep dependencies minimal in libraries.  I wonder if the MCL library should be using `dynamic.Interface` or a similar dependency-light library.  (Does one exist?)

cmd/manager/main.go
cheftako: Lets not conflate

cmd/manager/main.go
cheftako: I think we should be able to listen for a cancel on the context used for leader election and exist when it is called.

operator/config/crd/bases/core.cnrm.cloud.google.com_configconnectors.yaml
cheftako: Seems like we should start with per-cluster level. I think namespace (or sharding) is more complicated to build, maintain, configure and understand. Things like the syncer also get more complicated. Lets ship the easier version and then work on it.

pkg/controller/kccmanager/kccmanager.go
cheftako: +1

pkg/controller/kccmanager/kccmanager.go
cheftako: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/cloud-provider/app/controllermanager.go#L248 seems the current standard way to use leader election.

operator/pkg/controllers/configconnector/experiments.go
acpana: Good question! I have gated the leader election set up on the flag as well!

operator/pkg/apis/core/v1beta1/configconnector_types.go
acpana: We had chatted offline a bit about embedding the "cluster coordinates" in the lease identity. This will help the syncer in particular too.  

I think the templating of the lock here is somewhat orthogonal to that conversation and we can implement it (maybe as part of a follow on as we move towards per namespace leader election/ sharding ?)

cmd/manager/main.go
acpana: I've added a separate runnable to also kill the process!

go.mod
acpana: Downgraded mcl

pkg/controller/kccmanager/kccmanager.go
acpana: This should be reverted now

cmd/manager/main.go
```diff
@@ -69,6 +71,7 @@ func main() {
 	flag.IntVar(&pprofPort, "pprof-port", 6060, "The port that the pprof server binds to if enabled.")
 	flag.Float32Var(&rateLimitQps, "qps", 20.0, "The client-side token bucket rate limit qps.")
 	flag.IntVar(&rateLimitBurst, "burst", 30, "The client-side token bucket rate limit burst.")
+	flag.BoolVar(&multiClusterElection, "multi-cluster-election", false, "Enable multi-cluster leader election.")
```
cheftako: Standard gripe against boolean in apis. Would rather have an enum anticipating the likely direction of expansion. Eg multiClusterElectionType with values Disabled and Standard or V1.

operator/config/crd/bases/core.cnrm.cloud.google.com_configconnectors.yaml
```diff
@@ -66,6 +66,37 @@ spec:
...
+                properties:
+                  leaderElection:
+                    description: LeaderElection defines the configuration for multi
+                      cluster leader election.
+                    properties:
+                      multiClusterLease:
+                        description: MultiClusterLease defines configuration specific
+                          to multi-cluster leader election.
+                        properties:
+                          globalLockName:
```
cheftako, line 80: Can we default these?

acpana: added a default for the lock name -- not sure if it makes sense to default the name and namespace but we could to something like `cnrm-system` or `cnrm-mcl-system`

Issue #5334, opened 2025-10-07
feat: Add AnalyticsAccount as a KRM-native resource (Alpha)


yuwenma:
This task is to introduce the `AnalyticsAccount` resource as a new KRM-native resource in Alpha.

    This involves:
    - Creating the CRD and necessary API types.
    - Implementing the direct controller logic.
    - Adding comprehensive test fixtures and golden logs.
    - Add mockgcp to verify the fixtures.

maqiuyujoyce:
From the instructions (https://developers.google.com/analytics/devguides/config/admin/v1#account_provisioning), there is no way to know the account ID that was just created.

Here are more details:
1. Call [provisionAccountTicket](https://developers.google.com/analytics/devguides/config/admin/v1/rest/v1beta/accounts/provisionAccountTicket). Required fields. For example:
```
{
  "account": {
    "displayName": "test",
    "regionCode": "US"
  },
  "redirectUri": "http://www.google.com"
}
```
2. Get the returned account ticket ID, and use it to open this link: https://analytics.google.com/analytics/web/?provisioningSignup=false#/termsofservice/ACCOUNT_TICKET_ID.
3. The webpage contains some agreements for user to tick and to agree on. Once clicked `I agree`, it'll redirect to the redirectUri user configured.

In this process, I didn't see a way to retrieve the account ID.

I then listed the accounts using and got this:
```
{
  "accounts": [
    {
      "name": "accounts/[ACCOUNT_ID]",
      "createTime": "2025-10-08T03:27:59.219Z",
      "updateTime": "2025-10-08T03:27:59.219Z",
      "displayName": "test",
      "regionCode": "US"
    }
  ]
```
There is no information related to ticket ID above.

I also tried to create the same account again. It also passed and gave a different ticket ID this time.

I listed the accounts again, and there was still one account.

I tried to go to the webpage using the new ticket ID, this time when I accepted the agreement, I got the following error:
```
com.google.apps.framework.request.BadRequestException
	at com.google.analytics.boq.frontend.admin.service.actions.account.DeferredAccountListAction.lambda$createDeferredAccount$0(DeferredAccountListAction.java:128)
	at com.google.async.promisegraph.PromiseThrowables$CatchBlockFunction.apply(PromiseThrowables.java:378)
	at com.google.async.promisegraph.PromiseThrowables$CatchBlockFunction.apply(PromiseThrowables.java:358)
[...]

PR #5422, opened 2025-10-14
Support mockgcp for AnalyticsAccount

maqiuyujoyce:
### BRIEF Change description

Fixes part 1 of #5334 

Note that:
1. Analytics API doesn't support gcloud so we can't have gcloud test for it. Though the mock code is already verified in #5411. Splitting it out for review purposes.
2. We use the alpha version (googleapis/google/analytics/admin/v1alpha/*.proto) because only v1alpha is supported in the latest version of Cloud Go client (https://pkg.go.dev/cloud.google.com/go/analytics@v0.30.0/admin/apiv1alpha), and the proto version needs to match the client code version in the mockgcp.

```release-note
NONE
```

```docs

```

#### Intended Milestone

### Tests you have done


go.mod
yuwenma, line 1: Do you mean bumping the go mod in mockgcp (rather than the REPO_ROOT go.mod)?

mockgcp/Makefile
```diff
@@ -220,7 +220,8 @@ generate-grpc-for-google-protos:
 		./third_party/googleapis/google/cloud/recaptchaenterprise/v1/*.proto \
 		./third_party/googleapis/google/firestore/admin/v1/*.proto \
 		./third_party/googleapis/google/iam/v2/*.proto \
-		./third_party/googleapis/google/logging/v2/*.proto
+		./third_party/googleapis/google/logging/v2/*.proto \
+		./third_party/googleapis/google/analytics/admin/v1alpha/*.proto
```
yuwenma, line 224: Alpha is very experimental version. Suggest at least using v1beta?
https://developers.google.com/analytics/devguides/config/admin/v1/rpc/google.analytics.admin.v1beta

maqiuyujoyce: (Mentioned in description): We use the alpha version (googleapis/google/analytics/admin/v1alpha/*.proto) because only v1alpha is supported in the latest version of Cloud Go client (https://pkg.go.dev/cloud.google.com/go/analytics@v0.30.0/admin/apiv1alpha), and the proto version needs to match the client code version in the mockgcp.

We can bump to v1beta once the Go client supports it.

go.mod
maqiuyujoyce: Good point! Added the dependency in the mockgcp folder.

Tried to remove it from the root go.mod but it got added back automatically after `make ensure` so I assume that this indirect dependency is needed

yuwenma (commented):
/approve
/lgtm

PR #5084, opened 2025-08-29
[WIP] Update conductor to use gemini CLI

maqiuyujoyce:
### Change description

WIP. PR created for backing up purposes.

Fixes #

#### Does this PR add something which needs to be 'release noted'?

```release-note

```

- [ ] Reviewer reviewed release note.

#### Additional documentation e.g., references, usage docs, etc.:

```docs

```

#### Intended Milestone

Please indicate the intended milestone. 
- [ ] Reviewer tagged PR with the actual milestone.

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #5423, opened 2025-10-14
Support API, mapper and fuzzer for AnalyticsAccount

maqiuyujoyce:
### BRIEF Change description

Fixes part 2 of https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/5334

Note that additional fields and logics were implemented to support the CRUD workflow defined in the controller.

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -136,7 +136,6 @@ require (
 require go.opentelemetry.io/auto/sdk v1.1.0 // indirect
 
 require (
-	cloud.google.com/go/analytics v0.30.0 // indirect
 	cloud.google.com/go/osconfig v1.14.6 // indirect
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/detectors/gcp v1.27.0 // indirect
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/exporter/metric v0.53.0 // indirect


pkg/controller/direct/analytics/account_mapper.go
yuwenma, line 1: I see the mapper functions are all manually added. Can you try using the auto generated mapper? If not, what is the problem you see?

maqiuyujoyce: I was using conductor to generate it and realized that it is dependent on `controllerbuilder prompt` command, which relies on the old codebot logic and an discontinued version of gemini. I'm in the process of migrating the usage of this command in conductor to pure Gemini CLI ones.

Once that work is done, I'll switch back to generating mappers using automation.

FYI, a related PR #5084.

yuwenma: Could you use the controllerbuilder go code to generate the mapper? I noticed that manual mapper can always introduces frictions (e.g. nil pointer ref) and is hard to verify.

apis/analytics/v1alpha1/account_identity.go
yuwenma, line 1: Please add the generate.sh in this dir. You can look at example in apis/biglake/v1alpha1

maqiuyujoyce: Done.

pkg/controller/direct/analytics/account_mapper.go
maqiuyujoyce: Actually after re-reading your original comment, I noticed that I misunderstood - I thought you were talking about the fuzzers!

The mapper functions here were generated automatically, but because the types have to be manually constructed anyways (they require fields not existent in the original protobuf definition), I didn't create a commit for the manually generated one.

Added the empty generated mapper back.

apis/analytics/v1alpha1/generate.sh
yuwenma, line 1: (ah, I didn't know this works) But can you move this generate.sh to the v1alpha1 directory? Because the v1alpha1 and v1beta1 cannot use the same (yet) and have to be separate by the version dir.

maqiuyujoyce: Done!

apis/analytics/v1alpha1/account_identity.go
yuwenma, line 1: suggest remove identity and reference because they are not tested in this PR.

pkg/controller/direct/analytics/account_fuzzer.go
```diff
@@ -0,0 +1,44 @@
...
+
+	f.UnimplementedFields.Insert(".name")
+
+	f.SpecFields.Insert(".display_name")
+	f.SpecFields.Insert(".region_code")
+
+	f.StatusFields.Insert(".create_time")
+	f.StatusFields.Insert(".update_time")
+	f.StatusFields.Insert(".deleted")
+	f.StatusFields.Insert(".gmp_organization")
```
yuwenma, line 41: nit: I learnt that the f now has some top layer functions and we'd better use them. e.g. f.SpecField(".display_name"), f.Unimplemented_Identity

yuwenma (commented):
/approve

PR #5424, opened 2025-10-14
Support the controller for AnalyticsAccount

### LINKED ISSUE #5334: feat: Add AnalyticsAccount as a KRM-native resource (Alpha)

maqiuyujoyce:
### BRIEF Change description

Diffbase #5423 

Fixes #5334 

Analytics API doesn't have the endpoint to directly create an account. In order to create an Analytics account, user needs to call ProvisionAccountTicket() and then accept the ToS manually after constructing the link using the account ticket ID returned by ProvisionAccountTicket(). More details can be found in the comments in #5334.

To make this workflow work as transparently as possible, here are the customizations I did:

1. Added the `spec.redirectURI` and `status.observedState.accountTicketID` to support ProvisionAccountTicket() API call.
2. Used ProvisionAccountTicket() in Create and delayed setting of `status.externalRef` till we find a matching account in the ListAccounts() in Find.
3. Introduced the `ManualStepNotCompleted` error and make it a "dependency not ready" error. Explicitly asked users to accept the ToS when account ticket ID is found. Also made this error not a blocker for the resource deletion.

Picked the GRPC client because the REST client returned the following error repetitively even if I had already done what it suggested:
```
Error 403: Your application is authenticating by using local Application Default Credentials. The analyticsadmin.googleapis.com API requires a quota project, which is not set by default. To learn how to set your quota project, see https://cloud.google.com/docs/authentication/adc-troubleshooting/user-creds
```

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


pkg/controller/direct/analytics/account_controller.go
```diff
@@ -0,0 +1,336 @@
...
+func (m *modelAccount) AdapterForObject(ctx context.Context, reader client.Reader, u *unstructured.Unstructured) (directbase.Adapter, error) {
+	obj := &krm.AnalyticsAccount{}
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &obj); err != nil {
+		return nil, fmt.Errorf("error converting to %T: %w", obj, err)
+	}
+
+	id, err := krm.NewAccountIdentity(ctx, reader, obj)
+	if err != nil {
+		return nil, err
+	}
```
yuwenma: Can we use the identity.Identity interface? It would make the other resources easier to reference

pkg/controller/direct/analytics/account_controller.go
```diff
@@ -0,0 +1,336 @@
...
+		if accountTicketID != nil { // Reconciliation after creation ticket is provisioned.
+			namespacedName := types.NamespacedName{
+				Namespace: a.desired.Namespace,
+				Name:      a.desired.Name,
+			}
+			if *accountTicketID != "" {
+				return false, k8s.NewManualStepNotCompletedError(a.desired.GroupVersionKind(), namespacedName,
+					fmt.Sprintf("creation of a Google Analytics account is not completed. Go to ToS page "+
+						"https://analytics.google.com/analytics/web/?provisioningSignup=false#/termsofservice/%s and "+
+						"accept the Terms of Service", *accountTicketID))
```
yuwenma, line 172: I like the user-friendly message but I'm not sure if KCC should handle this case. For privacy and legal, it is normally out of the scope of the reconciler from the application layer. And It looks like the GCP should gives message on the trouble shooting message.

pkg/controller/direct/analytics/account_controller.go
```diff
@@ -0,0 +1,336 @@
...
+	mapCtx := &direct.MapContext{}
+
+	desiredPb := AnalyticsAccountSpec_ToProto(mapCtx, &a.desired.DeepCopy().Spec)
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+	desiredPb.Name = a.id.String()
+
+	paths := make(sets.Set[string])
+	// Option 1: This option is good for proto that has `field_mask` for output-only, immutable, required/optional.
```
yuwenma: can you clean up this code?

pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -354,6 +354,11 @@ func (r *reconcileContext) doReconcile(ctx context.Context, u *unstructured.Unst
 			// dependencies BUT the object is being deleted right now AND its dependecies have been deleted, then we will
 			// never encounter a "ready" condition for its dependencies
 			if !u.GetDeletionTimestamp().IsZero() {
+				// Edge case for AnalyticsAccount: When the manual step for the account creation is not completed, then
+				// the KRM object can be deleted without taking any actions because the account is never created.
+				if k8s.IsManualStepNotCompletedError(err) {
+					return false, r.handleDeleted(ctx, u)
+				}
```
yuwenma, line 361: This would be error prone if another reconciler use the generic ManualStep error and have different deletion behavior. I suggest we use deletion operation, and move the logic inside account_controller.go

pkg/k8s/errors.go
```diff
@@ -233,3 +233,35 @@ func (e *ImmutableFieldsMutationError) Error() string {
 func NewImmutableFieldsMutationError(immutableFields []string) *ImmutableFieldsMutationError {
 	return &ImmutableFieldsMutationError{immutableFields}
 }
+
+type ManualStepNotCompletedError struct {
+	resourceGVK schema.GroupVersionKind
+	resource    types.NamespacedName
+	message     string
+}
+
```
yuwenma, line 242: Can we introduce this generic error when another resource needs it? This give us a better signal to decide what should be generic and what should not.

pkg/controller/direct/analytics/account_controller.go
maqiuyujoyce: Done. Just curious: has our controller builder template be updated to use the new pattern?

pkg/controller/direct/analytics/account_controller.go
maqiuyujoyce: Done.

PR #5430, opened 2025-10-14
tests: reuse normalizer for exported KRM object

justinsb:
The object we export from the kube-apiserver has more information
than when we export the object, particularly because we do not export
status.

Reuse the normalizer for the kube-apiserver object, because many
fields will be more easily extracted from the status,
in particular identity fields.

### DIFF
--- a/tests/e2e/unified_test.go
+++ b/tests/e2e/unified_test.go
@@ -374,16 +374,42 @@ func runScenario(ctx context.Context, t *testing.T, testPause bool, fixture reso
 					folderID := h.FolderID()
 
 					for _, obj := range exportResources {
-						// Golden test exported GCP object
+						// Check the final state of the object in the kube-apiserver (and compare against golden file)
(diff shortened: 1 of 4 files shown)


justinsb:
/assign @yuwenma 

This is more useful for new tests, where we can extract normalizations from status.externalName (which is not in the exported object)

tests/e2e/unified_test.go
```diff
@@ -374,16 +374,42 @@ func runScenario(ctx context.Context, t *testing.T, testPause bool, fixture reso
 					folderID := h.FolderID()
 
 					for _, obj := range exportResources {
-						// Golden test exported GCP object
+						// Check the final state of the object in the kube-apiserver (and compare against golden file)
+						var normalizer *objectWalker
```
justinsb, line 378: We are doing these in the opposite order now, this code block has moved from lower down

PR #5432, opened 2025-10-14
tests: normalize request/response headers in logs

justinsb:


yuwenma (commented):
/approve

Looks good, I was expecting to see more golden log changes, but it seems only networkAttachments is affected 😆

PR #5435, opened 2025-10-15
Promote orgpolicycustomconstraint to beta

yuwenma:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/tests/apichecks/testdata/exceptions/missingfields.txt
+++ b/tests/apichecks/testdata/exceptions/missingfields.txt
@@ -3121,7 +3121,6 @@
 [missing_field] crd=notebookinstances.notebooks.cnrm.cloud.google.com version=v1beta1: field ".spec.vmImage.imageFamily" is not set in unstructured objects
 [missing_field] crd=notebookinstances.notebooks.cnrm.cloud.google.com version=v1beta1: field ".spec.vmImage.imageName" is not set in unstructured objects
 [missing_field] crd=notebookinstances.notebooks.cnrm.cloud.google.com version=v1beta1: field ".spec.vmImage.project" is not set in unstructured objects
-[missing_field] crd=orgpolicycustomconstraints.orgpolicy.cnrm.cloud.google.com version=v1beta1: field ".spec.description" is not set in unstructured objects
 [missing_field] crd=osconfigguestpolicies.osconfig.cnrm.cloud.google.com version=v1beta1: field ".spec.assignment.instanceNamePrefixes[]" is not set in unstructured objects
 [missing_field] crd=osconfigguestpolicies.osconfig.cnrm.cloud.google.com version=v1beta1: field ".spec.assignment.instances[].external" is not set in unstructured objects
 [missing_field] crd=osconfigguestpolicies.osconfig.cnrm.cloud.google.com version=v1beta1: field ".spec.assignment.instances[].namespace" is not set in unstructured objects
--- a/pkg/test/resourcefixture/testdata/basic/orgpolicy/v1beta1/orgpolicycustomconstraint/_generated_object_orgpolicycustomconstraint.golden.yaml
+++ b/pkg/test/resourcefixture/testdata/basic/orgpolicy/v1beta1/orgpolicycustomconstraint/_generated_object_orgpolicycustomconstraint.golden.yaml
@@ -12,6 +12,7 @@ metadata:
 spec:
   actionType: ALLOW
   condition: resource.management.autoUpgrade == true
+  description: updated OrgPolicy custom constraint
   displayName: KCC generated test CuOP for update
   methodTypes:
   - CREATE
--- a/pkg/test/resourcefixture/testdata/basic/orgpolicy/v1beta1/orgpolicycustomconstraint/_http.log
+++ b/pkg/test/resourcefixture/testdata/basic/orgpolicy/v1beta1/orgpolicycustomconstraint/_http.log
@@ -31,6 +31,7 @@ X-Goog-Request-Params: parent=organizations%2F${organizationID}
 {
   "actionType": 1,
   "condition": "resource.management.autoUpgrade == false",
+  "description": "A new OrgPolicy custom constraint",
   "displayName": "KCC generated test CuOP for create",
   "methodTypes": [
     1,
@@ -55,6 +56,7 @@ X-Xss-Protection: 0
 {
   "actionType": 1,
   "condition": "resource.management.autoUpgrade == false",
+  "description": "A new OrgPolicy custom constraint",
   "displayName": "KCC generated test CuOP for create",
   "methodTypes": [
     1,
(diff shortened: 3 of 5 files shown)


yuwenma:
issue #5429

apis/orgpolicy/v1alpha1/policy_types.go
```diff
@@ -176,3 +176,12 @@ type PolicySpec struct {
 	// +kcc:proto:field=google.cloud.orgpolicy.v2.PolicySpec.reset
 	Reset *bool `json:"reset,omitempty"`
 }
+
+// +kcc:observedstate:proto=google.cloud.orgpolicy.v2.PolicySpec
```
justinsb, line 180: Why?  This looks identical to the generated version

apis/orgpolicy/v1alpha1/types.generated.go
```diff
@@ -71,6 +70,17 @@ type PolicySpec_PolicyRule struct {
...
+	// Optional. Required for managed constraints if parameters are defined.
+	//  Passes parameter values when policy enforcement is enabled. Ensure that
+	//  parameter value types match those defined in the constraint definition.
+	//  For example:
+	//  {
+	//    "allowedLocations" : ["us-east1", "us-west1"],
+	//    "allowAll" : true
+	//  }
+	// +kcc:proto:field=google.cloud.orgpolicy.v2.PolicySpec.PolicyRule.parameters
+	Parameters map[string]string `json:"parameters,omitempty"`
```
justinsb: Looks like the type here is not `map[string]string`

justinsb: It's alpha, but it feels like we need to handle it because I don't see how we catch it in the beta promotion (although maybe the fuzz-test will fail?)

pkg/controller/direct/orgpolicy/policy_mappers.go
```diff
@@ -20,26 +20,86 @@ package orgpolicy
...
+}
+
+func Parameters_FromProto(mapCtx *direct.MapContext, in *structpb.Struct) map[string]string {
+	if in == nil {
+		return nil
+	}
+
+	out := map[string]string{}
+	for k, v := range in.AsMap() {
+		out[k] = v.(string)
```
justinsb: This will presumably panic with non-string values

apis/orgpolicy/v1alpha1/policy_types.go
yuwenma: the type.generated.go cannot generate this function. 

There's some flaws in the type-generator when we distinguish the Spec and ObservedState from the same proto.

Since we moved the spec part "PolicySpec" from types.generate.go to this manual file, and it has the marker `+kcc:observedstate:proto=google.cloud.orgpolicy.v2.PolicySpec`. The type generator skips the  ObservedState part as well.

I can fix it in a follow up PR, but because we want to enforce the generate.sh. This function has to be in the manual types.go

apis/orgpolicy/v1alpha1/types.generated.go
yuwenma: Good point! I trusted the type generator and didn't notice the type can actually be non-string.

yuwenma:
/hold cancel

I removed the parameter from alpha OrgPolicy.  I also found a bug in the mapper-generator #5436

pkg/controller/direct/orgpolicy/policy_mappers.go
```diff
@@ -20,26 +20,78 @@ package orgpolicy
...
+		return nil
+	}
+	out := &expr.Expr{}
+	out.Description = direct.ValueOf(in.Description)
+	out.Expression = direct.ValueOf(in.Expression)
+	out.Location = direct.ValueOf(in.Location)
+	out.Title = direct.ValueOf(in.Title)
+	return out
+}
+
```
yuwenma: The following 3 functions is to bypass a mapper-generator bug (fixed in a bug in the mapper-generator https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5436) 

The result is that the OrgPolicy cannot map a boolean field (if it is a `OneOf`) when set to False.

apis/orgpolicy/v1alpha1/policy_types.go
justinsb: OK, no worries, I assumed that because it was previously in generated that it was generated.  If that's not the case, then this makes sense!

pkg/controller/direct/orgpolicy/policy_mappers.go
justinsb: OK so if there are more changes needed in this PR let's hold off on approving.  I approved 5436 so let's let that merge, then we can rebase this one.

I personally think we could do the same thing with `map[string]string`

Another option is to cherry-pick 5436 into this one, then we can approve both at the same time.

tests/apichecks/testdata/exceptions/missingfields.txt
```diff
@@ -3121,6 +3121,7 @@
 [missing_field] crd=notebookinstances.notebooks.cnrm.cloud.google.com version=v1beta1: field ".spec.vmImage.imageFamily" is not set in unstructured objects
 [missing_field] crd=notebookinstances.notebooks.cnrm.cloud.google.com version=v1beta1: field ".spec.vmImage.imageName" is not set in unstructured objects
 [missing_field] crd=notebookinstances.notebooks.cnrm.cloud.google.com version=v1beta1: field ".spec.vmImage.project" is not set in unstructured objects
+[missing_field] crd=orgpolicycustomconstraints.orgpolicy.cnrm.cloud.google.com version=v1beta1: field ".spec.description" is not set in unstructured objects
```
justinsb: Is this one a hard one to add?

yuwenma: (I have the same question flowing in my head, then distracting to other PRs)

pkg/controller/direct/orgpolicy/policy_mappers.go
```diff
@@ -20,26 +20,59 @@ package orgpolicy
...
 	}
-	out := &krm.OrgPolicyPolicyObservedState{}
-	out.Spec = PolicySpecObservedState_FromProto(mapCtx, in.GetSpec())
-	out.DryRunSpec = PolicySpecObservedState_FromProto(mapCtx, in.GetDryRunSpec())
+	out := &krm.PolicySpec{}
+	// MISSING: Etag
+	// MISSING: UpdateTime
+	out.Rules = direct.Slice_FromProto(mapCtx, in.Rules, PolicySpec_PolicyRule_FromProto)
+	out.InheritFromParent = direct.LazyPtr(in.GetInheritFromParent())
+	// Expected code `in.GetReset()`, actual code `in.GetReset_()`
```
justinsb, line 38: Is `reset` a reserved keyword or something?  Do we know why this happened?

PR #5439, opened 2025-10-16
fix: skip sis defaulter for direct

acpana:
### BRIEF Change description

Don't apply the state into spec ApplyDefaults for direct resources.

Fixes #

Needed in order for resources with legacy annotations to be transitioned to direct from TF

```release-note
Ignore state into spec annotation  for any direct resources
```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/tests/e2e/testdata/scenarios/sql/merge-after-direct-default/_http00.log
+++ b/tests/e2e/testdata/scenarios/sql/merge-after-direct-default/_http00.log
@@ -1,0 +1,251 @@
+GET https://sqladmin.googleapis.com/sql/v1beta4/projects/${projectId}/instances/sqlinstance-${uniqueId}?alt=json&prettyPrint=false
+User-Agent: kcc/${kccVersion} (+https://github.com/GoogleCloudPlatform/k8s-config-connector) kcc/controller-manager/${kccVersion}
+
+404 Not Found
+Content-Type: application/json; charset=UTF-8
+Server: ESF
+Vary: Origin
+Vary: X-Origin
+Vary: Referer
+X-Content-Type-Options: nosniff
+X-Frame-Options: SAMEORIGIN
+X-Xss-Protection: 0
+
+{
+  "error": {
+    "code": 404,
+    "errors": [
+      {
+        "domain": "global",
+        "message": "The Cloud SQL instance does not exist.",
+        "reason": "instanceDoesNotExist"
+      }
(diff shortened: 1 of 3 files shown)


pkg/stateintospec/stateintospecdefaulter.go
```diff
@@ -55,6 +55,12 @@ func NewStateIntoSpecDefaulter(client client.Client) k8s.Defaulter {
 }
 
 func (v *StateIntoSpecDefaulter) ApplyDefaults(ctx context.Context, reconcilerType k8s.ReconcilerType, resource client.Object) (changed bool, err error) {
+	// Don't write the annotation for direct controllers
+	switch reconcilerType {
+	case k8s.ReconcilerTypeDirect:
+		return false, nil
```
cheftako, line 61: This seems correct but uninformative. Would be nice if when we ignored SIS:Merge that we logged that we were ignoring it.

acpana: For resources that attempt to change SIS:Merge and are not allowed to do it I do think we log it

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/5d74072f03ef4ef5856d1f733fc2c34e4b22f1f5/pkg/stateintospec/stateintospecdefaulter.go#L110-L120

moving this code block up here also protects us against this error:

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/5d74072f03ef4ef5856d1f733fc2c34e4b22f1f5/pkg/stateintospec/stateintospecdefaulter.go#L61-L66

where in 1.134 and later `SQLInstance`  no longer supports `merge` .

pkg/controller/dcl/controller.go
```diff
@@ -472,7 +472,7 @@ func (r *Reconciler) handleDefaults(ctx context.Context, u *unstructured.Unstruc
 	for _, defaulter := range r.defaulters {
 		changed, err := defaulter.ApplyDefaults(ctx, k8s.ReconcilerTypeDCL, u)
 		if err != nil {
-			return fmt.Errorf("applying defaults: %w", err)
+			return fmt.Errorf("applying defaults in the dcl controller: %w", err)
```
justinsb, line 475: Yes, but for something that we're planning on cherry-picking we normally want to keep it minimal.  I think we'll get away with this one though.

PR #5449, opened 2025-10-16
Change trigger on release tag creation workflow

xiaoweim:
### BRIEF Change description

The workflow did not get triggered when we merge release PRs. For example [Release PR 137](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5433), and [Github Action Tag Release](https://github.com/GoogleCloudPlatform/k8s-config-connector/actions/workflows/tag-release.yml) did not get triggered. 

I suspect it is related to this long standing Github issue: https://github.com/orgs/community/discussions/57830. When too many files are changes in one PR, the `paths` fails to check all the file change path to match to the workflow and trigger the necessary workflow.

```release-note
None
```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -19,8 +19,6 @@ on:
     branches:
       - master
       - 'release-*'
-    paths:
-      - 'version/VERSION'
 
 jobs:
   tag-release:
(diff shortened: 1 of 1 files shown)


.github/workflows/tag-release.yml
```diff
@@ -34,10 +32,20 @@ jobs:
         with:
           # We need to fetch all tags to see if the tag already exists.
           fetch-depth: 0
+      # Check if version/VERSION was changed
```
justinsb, line 35: I feel like we could probably put this into the script (if it's not already there), but ... not a blocker

xiaoweim: Oh actually yes we did put it in the script, so technically I can trigger this workflow on every push since it always verify the version in the `version/VERSION` file!

PR #5452, opened 2025-10-17
deletiondefender: use metadata-only client to read CRDs

justinsb:


pkg/controller/deletiondefender/controller.go
```diff
@@ -20,10 +20,10 @@ import (
 	"strings"
 
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
+	"k8s.io/client-go/metadata"
```
yuwenma, line 23: Good to know this!

PR #5456, opened 2025-10-17
Improve the cherry-pick script to make it more debuggable and robust

maqiuyujoyce:
### BRIEF Change description

Fix the rough edges in the cherry-pick script. It is a bit chatty right now but we can always add --verbose flag in the future.

* Print out detailed exception trace when the script fail.
* Clean up the existing branch and worktree before restart the cherrypicking - a bad state might exist due to previous failures.

```release-note
NONE
```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/dev/tasks/cherry-pick
+++ b/dev/tasks/cherry-pick
@@ -15,7 +15,13 @@ import json
 
 def run(cmd):
     print(f"Running command: {' '.join(cmd)}")
-    return subprocess.run(cmd, check=True, capture_output=True, text=True)
+    try:
+        return subprocess.run(cmd, check=True, capture_output=True, text=True)
+    except subprocess.CalledProcessError as e:
+        print(f"Command failed with exit code {e.returncode}")
+        print(f"Stderr: {e.stderr}")
+        print(f"Stdout: {e.stdout}")
+        raise
 
 
 def main():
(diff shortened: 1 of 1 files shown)


dev/tasks/cherry-pick
```diff
@@ -15,7 +15,13 @@ import json
 
 def run(cmd):
     print(f"Running command: {' '.join(cmd)}")
-    return subprocess.run(cmd, check=True, capture_output=True, text=True)
+    try:
+        return subprocess.run(cmd, check=True, capture_output=True, text=True)
+    except subprocess.CalledProcessError as e:
+        print(f"Command failed with exit code {e.returncode}")
+        print(f"Stderr: {e.stderr}")
```
xiaoweim, line 22: I like this for debugging! 👍

dev/tasks/cherry-pick
```diff
@@ -39,9 +45,41 @@ def main():
     github_username = run(["gh", "api", "user", "-q", ".login"]).stdout.strip()
     user_fork = f"https://github.com/{github_username}/k8s-config-connector"
 
+    # Clean up any leftover branches from previous runs
```
justinsb, line 48: This potentially involves data loss, so I don't think we should do it automatically.

I assume the problem is that when the cherry-pick fails to apply, we lose state.  Maybe we could handle that?  I think a reasonable step would be to print the commands that need to be done to finish the cherry pick (after fixing conflicts).

I think the k/k script pauses and expects you to fix the conflict in another terminal.

maqiuyujoyce: > I think a reasonable step would be to print the commands that need to be done to finish the cherry pick (after fixing conflicts).

Yes we can do that - though does it mean we shouldn't rerun the cherry-pick script multiple times?

I feel it should be okay as long as it's before PR creation because the impact is local. WDYT?

maqiuyujoyce:
/assign codebot-robot

Could you address the comments without touching existing commits?

PR #5458, opened 2025-10-17
recorder: reduce memory footprint of recorder

justinsb:


justinsb:
So the pprof stats look good:

`Showing nodes accounting for 8272.12kB, 74.16% of 11154.41kB total`

vs 

`Showing nodes accounting for 28638.17kB, 82.25% of 34820.34kB total`

I imagine there's some variation there, but the pprof clearly shows that we are no longer the GCP protos, and these consumed non-trivial amounts of RAM previously.

justinsb:
/assign @cheftako 

This includes #5540 , though I don't think we _need_ to merge them separately.

cheftako:
Surprised this doesn't have a merge conflict.
/lgtm
/approve

Issue #5419, opened 2025-10-13
refactor: use additive diffing (no side effects)


acpana:
Mostly from Walter:

> Return Values Directly: Go's multiple return values are a powerful feature. Prefer returning values directly rather than using "out parameters" (passing pointers to be modified). If you do not need values from an input parameter then it should probably be a return value.

The ask here is to actually report a diff in places like this as opposed to adding fields to the existing diff passed in as a parameter

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/e364c694b230a79aaa9f095d277a84bc29809a13/pkg/controller/direct/sql/sqlinstance_equality.go#L25

_Originally posted by @acpana in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5152#discussion_r2427365989_

PR #5462, opened 2025-10-17
refactor: report all diffs for SQLInstance

acpana:
### BRIEF Change description

Realized we are not actually reporting ALL the diffs for a SQLInstance.

Fixes https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/5419

It will improve the UX of preview, for one, and allow operators to see what the differences are between SQLInstance reconciles.

```release-note
Log all field diffs for `SQLInstance`.
```

```docs

```

#### Intended Milestone

Tagged with 1.140

### Tests you have done

### DIFF
--- a/pkg/controller/direct/sql/sqlinstance_equality.go
+++ b/pkg/controller/direct/sql/sqlinstance_equality.go
@@ -27,9 +27,6 @@ func DiffInstances(desired *api.DatabaseInstance, actual *api.DatabaseInstance) 
 	if desired == nil && actual == nil {
 		return diff
 	}
-	if !PointersMatch(desired, actual) {
-		return diff
-	}
 
 	if desired.DatabaseVersion != actual.DatabaseVersion {
 		diff.AddField(".databaseVersion", actual.DatabaseVersion, desired.DatabaseVersion)


pkg/controller/direct/sql/sqlinstance_equality.go
```diff
@@ -22,59 +22,52 @@ import (
...
-func InstancesMatch(desired *api.DatabaseInstance, actual *api.DatabaseInstance, diff *structuredreporting.Diff) bool {
+func DiffInstances(desired *api.DatabaseInstance, actual *api.DatabaseInstance) *structuredreporting.Diff {
+	diff := &structuredreporting.Diff{}
 	if desired == nil && actual == nil {
-		return true
+		return diff
 	}
 	if !PointersMatch(desired, actual) {
-		return false
+		return diff
```
cheftako: PointersMatch is a terrible name for that method. At this point we know that exactly one of desired/actual are not nil. Returning an empty diff does not seem correct. Also suspicious this does not conform to the old behavior.

cheftako: Would be good to add a test for each of desired and actual being nil.

pkg/controller/direct/sql/sqlinstance_equality.go
```diff
@@ -22,59 +22,49 @@ import (
...
 	if !ReplicaConfigurationsMatch(desired.ReplicaConfiguration, actual.ReplicaConfiguration) {
 		diff.AddField(".replicaConfiguration", actual.ReplicaConfiguration, desired.ReplicaConfiguration)
-		return false
 	}
 	// Ignore ReplicationCluster. It is not supported in KRM API.
 	// Ignore RootPassword. It is not exported.
-	if !SettingsMatch(desired.Settings, actual.Settings, diff) {
-		return false
-	}
+	diff.AddDiff(DiffSettings(desired.Settings, actual.Settings))
```
anhdle-sso, line 61: Why don't you pass the diff pointer to this function and keep adding to the same obj? You would not have to call AddDiff

acpana: I think @cheftako wanted us to move away from side effects like that 😛

pkg/controller/direct/sql/sqlinstance_equality.go
acpana: I've reworked this to remove the first pointers match!

pkg/controller/direct/sql/sqlinstance_equality.go
cheftako: +1. If a function name makes it clear that it mutates its inputs then we can discuss it. However the function DiffSettings suggests it does not mutate its inputs. AddDiff mya not modify its inputs but its at least clear its accumulating a result.

PR #5464, opened 2025-10-17
Fix: Correct Docs, prepare to support Logging Link on beta

tylerreidwaze:
### BRIEF Change description

Update some docs + 1 minor change with logging link on deletion

Bugs in LoggingLink resource


pkg/controller/direct/logging/mapper.generated.go
```diff
@@ -0,0 +1,83 @@
...
+func LoggingLinkSpec_v1alpha1_ToProto(mapCtx *direct.MapContext, in *krm.LoggingLinkSpec) *pb.Link {
+	if in == nil {
+		return nil
+	}
+	out := &pb.Link{}
+	// MISSING: Name
+	out.Description = direct.ValueOf(in.Description)
+	// MISSING: CreateTime
+	// MISSING: LifecycleState
+	// MISSING: BigqueryDataset
```
yuwenma: I think you want the BigQueryDataset to be assigned.

yuwenma: It looks like the API and mapper has some mismatch:
1. The KCC LoggingLink has `loggingBucketRef` field. I was expecting to see a `loggingBucket` field in the pb.Link (or a "// MISSING: ` comment)
2. The KCC LoggingLink does not have the `bigquerydataset` field, but it looks like the proto has one. Could you verify if the LoggingLink API is correct?

tylerreidwaze: A bit of gemini to help investigate this, but it all seems accurate and meshes with my understanding.  I think the autogenerate code is incorrectly assuming some fields and some of these fields are output only

> The KCC LoggingLink has loggingBucketRef field. I was expecting to see a loggingBucket field in the pb.Link (or a "// MISSING: ` comment)

The loggingBucketRef field in the KCC LoggingLink resource is a KCC-specific reference to a
  LoggingLogBucket. The underlying GCP Logging API's Link proto (defined in google.logging.v2.Link)
  does not have a direct loggingBucketRef or loggingBucket field.

  Instead, the pb.Link struct has a Name field, which is the full resource name of the link,
  including the bucket ID. The KCC controller is responsible for resolving the loggingBucketRef
  and constructing this Name field before sending the request to the GCP API.

  Therefore, loggingBucketRef is not expected to appear as a direct field in pb.Link or as a // 
  MISSING: comment in mapper.generated.go, as it's handled by the KCC controller's reference
  resolution logic, not a direct field-to-field mapping.

> The KCC LoggingLink does not have the bigquerydataset field, but it looks like the proto has one. Could you verify if the LoggingLink API is correct?

The bigquery_dataset field in the pb.Link proto is an output-only field, meaning it's populated by
   the GCP service upon resource creation and cannot be set by the user in the request.

  In KCC, output-only fields are placed in the status.observedState of the custom resource, not in
   the spec.

  Looking at the LoggingLink definition in apis/logging/v1alpha1/link_types.go, the
  BigQueryDataset field is correctly placed in LoggingLinkObservedState, which is part of the
  status. The LoggingLinkSpec does not and should not contain this field.

  The API definition is correct. The confusion arises because the auto-generated mapper code at
  pkg/controller/direct/logging/mapper.generated.go is not aware that bigquery_dataset is an
  output-only field, so it incorrectly flags it as // MISSING: in the spec mapping functions. This
   is a bug in the code generation process that can be fixed by adding the proper annotation to
  inform the generator that the field is output-only.

PR #5465, opened 2025-10-17
Refactor logmocks

tylerreidwaze:
### BRIEF Change description

Refactor log mocks into separate mock service for each api (configs and metrics)

This shared mock creates issues when trying to override certain APIs

### DIFF
--- a/mockgcp/mockbigtable/service.go
+++ b/mockgcp/mockbigtable/service.go
@@ -62,7 +62,7 @@ func (s *MockService) ExpectedHosts() []string {
 func (s *MockService) Register(grpcServer *grpc.Server) {
 	pb.RegisterBigtableInstanceAdminServer(grpcServer, &instanceAdminServer{MockService: s})
 	pb.RegisterBigtableTableAdminServer(grpcServer, &tableAdminServer{MockService: s})
-	s.operations.RegisterGRPCServices(grpcServer)
+	// s.operations.RegisterGRPCServices(grpcServer)
 }
 
 func (s *MockService) NewHTTPMux(ctx context.Context, conn *grpc.ClientConn) (http.Handler, error) {


mockgcp/mockbigtable/service.go
```diff
@@ -62,7 +62,7 @@ func (s *MockService) ExpectedHosts() []string {
 func (s *MockService) Register(grpcServer *grpc.Server) {
 	pb.RegisterBigtableInstanceAdminServer(grpcServer, &instanceAdminServer{MockService: s})
 	pb.RegisterBigtableTableAdminServer(grpcServer, &tableAdminServer{MockService: s})
-	s.operations.RegisterGRPCServices(grpcServer)
```
tylerreidwaze, line 65: This is an open Item I am looking for a resolution on.  The way the Mock library is currently written, I can't register the operations endpoint for BOTH bigtable and logging_config.  I had to comment out this to get my Mocks to run.

mockgcp/mock_http_roundtrip.go
```diff
@@ -63,6 +63,8 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/mockgkehub"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/mockgkemulticloud"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/mockkms"
+	_ "github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/mocklogging_config"
+	_ "github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/mocklogging_metrics"
```
yuwenma, line 67: These are already registered in the register.go file, so these 2 lines should be removed

mockgcp/mockbigtable/service.go
yuwenma: Could you further elaborate? I don't get how the mockbigtable affects mocklogging. FYI, there's a duel registration issue in the register.go and mock_http_roundtrip.go files (maybe that's the cause   of problem you meet)

yuwenma (commented):
Could you separate the PR into 3 smallers PRs, one for link, one for bucket and one for metrics? They are identical changes.

PR #5472, opened 2025-10-20
recorder: optimize memory footprint of namespace-name map

justinsb:
We create a specialized map type for storing keys based on namespace and name,
which reduces memory overhead compared to using generic map types,
assuming that namespaces contain multiple objects.

### DIFF
--- a/cmd/recorder/kube/namespacename_map.go
+++ b/cmd/recorder/kube/namespacename_map.go
@@ -1,0 +1,77 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package kube
+
+import "k8s.io/apimachinery/pkg/types"
+
+// NamespaceNameMap is a map from types.NamespacedName to some value of type T.
+// We specialize for this case because it lets us share the memory for the namespace string values.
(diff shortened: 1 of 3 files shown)


cmd/recorder/kube/namespacename_map.go
```diff
@@ -0,0 +1,74 @@
...
+// Delete deletes the value for the given namespace and name.
+func (m NamespaceNameMap[T]) Delete(key types.NamespacedName) {
+	namespaceObjects, found := m[key.Namespace]
+	if !found {
+		return
+	}
+	delete(namespaceObjects, key.Name)
+}
+
+// Snapshot returns a deep-copy of the map, as a naive map[types.NamespacedName]T.
```
yuwenma: Curious about the "naive".  Do you mean "native"?

cmd/recorder/kube/namespacename_map.go
```diff
@@ -0,0 +1,74 @@
...
+func (m NamespaceNameMap[T]) Snapshot() map[types.NamespacedName]T {
```
yuwenma: My read for "snapshot" is that it stamps the object at a specific state (and time), so the returned value cannot be changed externally. This is normally enforced by the Snapshot type (checksum, marshal, etc). But it seems to be not for this case. How about `DeepCopyNative`?

yuwenma (commented):
/approve
/lgtm
/hold 

2 quick questions, non blockers

cmd/recorder/kube/namespacename_map.go
justinsb: naive as opposed to the optimized structure we're implementing here

cmd/recorder/kube/namespacename_map.go
justinsb: Yeah, we don't lock so Snapshot is not great.  I think it's fine to return a mutable value from a Snapshot, but it should be at a point in time.  I thought about DeepCopy, but because we're changing the representation that is not a great name either.  I might just refactor to use Walk instead!

justinsb: Got rid of Snapshot; we want to avoid the memory allocation

PR #5480, opened 2025-10-20
chore: include signal in context-cancellation error

justinsb:
When the context is cancelled, it is helpful to have the error
indicate that the cancellation was due to a signal.

This change introduces new signal handling utilities
which record the signal that caused the context cancellation.


yuwenma:
Code looks good. I'm more curious about the rationale behind: Is this for debugging purpose (a real customer report about an ambiguous webhook termination) or more for development? Could you share the context?

m74278803-cmyk (commented):
Approvals

PR #5485, opened 2025-10-21
chore: update to use kube-apiserver 1.31 in tests

justinsb:


justinsb:
The PR that shows the problem is https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5473

PR #5491, opened 2025-10-21
doc: update reference field requirements

yuwenma:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/docs/develop-resources/api-conventions/resource-reference.md
+++ b/docs/develop-resources/api-conventions/resource-reference.md
@@ -1,182 +1,118 @@
 # Direct Resource Reference Guide
 
+## TL;DR
 
-# TL;DR
+Referencing GCP objects should define an API field that ends with `Ref` or `Refs` to indicate a dependency on another GCP resource. This document outlines the conventions for these reference fields, ensuring consistency and enabling automation via tooling.
 
-Referencing GCP objects should define a API field `spec.<Kind>Ref(s)` to reference the dependency of another GCP resource.  
+## Naming Conventions
 
+The naming of resource reference fields follows a clear and consistent pattern, mapping from the GCP proto field name to the KCC KRM field name and type. This convention is **required** for the `mapper-generator` to function correctly.
 
-# Basic rules
+### From GCP Proto to KRM
 
-If a Config Connector resource depends on another resource, it should define a `spec.<Kind>Ref` field, even if the dependent does not have a Config Connector resource yet (in which case only External is required).
+-   **GCP Proto Field:** The original field name in the GCP proto definition (e.g., `crypto_key_name`). This is typically in `snake_case`.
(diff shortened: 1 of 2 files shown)


docs/develop-resources/api-conventions/resource-reference.md
```diff
@@ -1,182 +1,118 @@
...
-Some Config Connector resources make the `resourceKind.external` to serve different usages.
+Use the `k8s.ReferenceNotFoundError` when the referenced Config Connector object is not found.
 
-For example, the `computeForwardingRule` has the `ComputeAddress` reference which `external` allows IP address value like `8.8.8.8 `(search `spec.ipAddress.addressRef.external` in this [page](https://cloud.google.com/config-connector/docs/reference/resource-docs/compute/computeforwardingrule)) 
+## Same Kind References
 
+For lists of references of the same kind:
 
-### Rule 1
+- **Rule 1: Avoid Mixed Forms:** Use either `external` or `name` for all references in a list, not a mix of both.
```
justinsb, line 109: Why?  And wrong doc?  This feels like advice to users rather than advice to developers

docs/develop-resources/api-conventions/resource-reference.md
```diff
@@ -1,182 +1,118 @@
...
+- **Rule 2: Form Switching:** Users should be able to switch between using `external` and `name`.
+- **Rule 3: Unique `external`:** Validate the uniqueness of `external` string values.
+- **Rule 4: Non-unique `name`:** Do not validate the uniqueness of `name`/`namespace` pairs. Instead, check the uniqueness of the GCP resources referenced by the `externalRef` field of the corresponding Config Connector objects.
```
justinsb, line 112: I don't think this is realistic.  We will normalize all references and then we can check uniqueness

docs/develop-resources/api-conventions/resource-reference.md
```diff
@@ -1,182 +1,118 @@
...
+- **Rule 5: Ordering:** Preserve the order of references as defined in the CRD, unless the GCP service prefers a sorted order.
```
justinsb: "in the CR" I believe, although "in the spec" would probably be clearer

docs/develop-resources/api-conventions/resource-reference.md
yuwenma: This is what we originally asked in SciFi guide, it defines the rules for developers to add the logic in their controller if applicable.  

This is not an ideal solution, but a tradeoff to use the simplest solution  to solve complicated problems: 
We need a way to identify duplicated items from the slice, so that we can tell if the config is changed, and whether a Update call should be made.

docs/develop-resources/api-conventions/resource-reference.md
yuwenma: Yes, this is exactly the same as what described here just in a different way.  When using`name/namespace`, the normalization works on `.status.externalRef`. 

This is written before we have a reference normalizer, even today we do not require controller to use the generic normalizer. So each controller can still have its own way to "normalize" the ref in theory.

Issue #5078, opened 2025-08-28
feat: Promote SpannerBackupSchedule to Beta (greenfield)


yuwenma:
This resource has controller not but full API coverage.

PR #5495, opened 2025-10-21
feat: Add references resolver in SpannerBackupsSchedule direct controller

gemmahou:
### BRIEF Change description

There was no references resolver in the BigQueryBackupsSchedule direct controller (Beta). This caused the KMSKeyRef to be skipped/ignored during the resource creation and update process.

I've added the resolver function to correctly process the KMSKeyRef value. I also updated our fixture test and verified the output against both realGCP and mockGCP.

Maybe related to https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/5078. cc @yuwenma 

Commit1: update SpannerBackupsSchedule direct controller and collect realGCP logs
Commit2: collect mockGCP logs. Note that all diffs between the mock and the real output come from the dependency SpannerDatabase/Instance, not the target resource BackupsSchedule. So it's safe to ignore them for now and improve the mock fidelity in future PRs.

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/controller/direct/spanner/backupschedule_resolverefs.go
+++ b/pkg/controller/direct/spanner/backupschedule_resolverefs.go
@@ -32,15 +32,13 @@ func resolveBackupScheduleRefs(ctx context.Context, reader client.Reader, obj *k
 	}
 
 	if obj.Spec.EncryptionConfig != nil && obj.Spec.EncryptionConfig.KMSKeyRefs != nil {
-		var keys []*refs.KMSCryptoKeyRef
-		for _, kmsKey := range obj.Spec.EncryptionConfig.KMSKeyRefs {
+		for i, kmsKey := range obj.Spec.EncryptionConfig.KMSKeyRefs {
 			key, err := refs.ResolveKMSCryptoKeyRef(ctx, reader, obj, kmsKey)
 			if err != nil {
 				return err
 			}
-			keys = append(keys, key)
+			obj.Spec.EncryptionConfig.KMSKeyRefs[i] = key
 		}
-		obj.Spec.EncryptionConfig.KMSKeyRefs = keys
 
 	}
 	return nil


mockgcp/mockspanner/admin/database.go
```diff
@@ -245,7 +245,7 @@ func (s *SpannerDatabaseV1) GetBackupSchedule(ctx context.Context, req *pb.GetBa
 
 	obj := &pb.BackupSchedule{}
 	if err := s.storage.Get(ctx, fqn, obj); err != nil {
-		return nil, err
+		return nil, status.Errorf(codes.NotFound, "Backup schedule not found: %s", fqn)
```
yuwenma: Is this the only possible error? I'm more curious about why changing it.

pkg/test/resourcefixture/testdata/basic/spanner/v1beta1/spannerbackupschedule/spannerbackupschedules-fully/dependencies.yaml
```diff
@@ -48,7 +48,7 @@ kind: KMSKeyRing
 metadata:
   name: kmskeyring-${uniqueId}
 spec:
-  location: us-central1
```
yuwenma, line 51: why do we change this?

pkg/test/resourcefixture/testdata/basic/spanner/v1beta1/spannerbackupschedule/spannerbackupschedules-fully/update.yaml
```diff
@@ -25,7 +25,7 @@ spec:
       text: 0 3 * * *
   retentionDuration: "1000000s"
   encryptionConfig:
-    encryptionType: "GOOGLE_DEFAULT_ENCRYPTION"
+    encryptionType: "CUSTOMER_MANAGED_ENCRYPTION"
```
yuwenma, line 28: same question: why do we change this value?

pkg/test/resourcefixture/testdata/basic/spanner/v1beta1/spannerbackupschedule/spannerbackupschedules-fully/create.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2022 Google LLC
+# Copyright 202 Google LLC
```
yuwenma: nit: can you fix this?

pkg/controller/direct/spanner/backupschedule_controller.go
yuwenma, line 1: The reason that the reference is not properly handled is because this resource miss the fuzzer test. Could you add it (as well as the generate.sh file)?

mockgcp/mockspanner/admin/database.go
gemmahou: It changes the mock server, to ensure mockGCP log matches the real.

pkg/test/resourcefixture/testdata/basic/spanner/v1beta1/spannerbackupschedule/spannerbackupschedules-fully/dependencies.yaml
gemmahou: When I ran the test against real GCP, I found that the key must be in the same location as the Spanner instance and database (which is us-west). It appears this error was not encountered previously because we never resolved the referenced key and never ran the test against a real GCP.

pkg/test/resourcefixture/testdata/basic/spanner/v1beta1/spannerbackupschedule/spannerbackupschedules-fully/update.yaml
gemmahou: Only CUSTOMER_MANAGED_ENCRYPTION supports kmskey, this is also mentioned in the field description: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/spanner/v1beta1/backupschedule_types.go#L71. Similarly we did not encounter this error before because we never resolved the referenced key and never ran the test against a real GCP.

pkg/controller/direct/spanner/backupschedule_controller.go
gemmahou: We have covered the encryption_config field in the fuzz test([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/c19f33c30f25e294a858168d342332b2572cd451/pkg/controller/direct/spanner/backupschedule_fuzzer.go#L38)). We also handle the reference conversion in the mappers, but this requires pre-resolving the references' external([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/c19f33c30f25e294a858168d342332b2572cd451/pkg/controller/direct/spanner/backupschedule_mapping.go#L49)). Failing to resolve the references' external will not cause the fuzz test to fail, because references' external default to be an empty string.

The reason is we did not verify the test against real GCP. Also, kmsKey field is missing in the mock golden log([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/c19f33c30f25e294a858168d342332b2572cd451/pkg/test/resourcefixture/testdata/basic/spanner/v1beta1/spannerbackupschedule/spannerbackupschedules-fully/_http.log#L605-L607)), but because the test passed, we may have ignored it

pkg/test/resourcefixture/testdata/basic/spanner/v1beta1/spannerbackupschedule/spannerbackupschedules-fully/update.yaml
yuwenma: Not really, the reason this was not encountered before is that this resource misses the fuzzer test (this could be because we don't have fuzzer test yet). Therefore, a more appropriate fix is to add the fuzzer test to guarantee *all* unmapped field, in lieu of this specific case.

pkg/test/resourcefixture/testdata/basic/spanner/v1beta1/spannerbackupschedule/spannerbackupschedules-fully/dependencies.yaml
yuwenma: If this is the reason, I suggest reflect that in the MockGCP (comments or/and code checks ). Otherwise, it is impossible to review and maintain.

This sounds like a good case to consider for LLM-driven PR reviews. If we reject the PRs with unrelated code changes, this PR would fail. If you have a reason, it'd better to be captured in the code (the file itself or somewhere).  cc @cheftak for other ideas.

mockgcp/mockspanner/admin/database.go
yuwenma: My original ask is if the `s.starge.Get` would return other errors (I think it will). Changing this error will hide those other cases (which normally indicates more severe and real problems).

pkg/controller/direct/spanner/backupschedule_resolverefs.go
```diff
@@ -0,0 +1,47 @@
...
+package spanner
+
+import (
+	"context"
+
+	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
+	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/spanner/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
```
yuwenma, line 24: Please use NormalizeReference function instead

pkg/controller/direct/spanner/backupschedule_controller.go
yuwenma: Yeah, this is about enforcing the reference normalization.  We should have a more generic approach to check if a resource has reference and has them normalized. Filed https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/5500

mockgcp/mockspanner/admin/database.go
gemmahou: Gotcha, updated in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5495/commits/2eae8d8040e4370d807c0896036bae2d8cff6c9b

pkg/test/resourcefixture/testdata/basic/spanner/v1beta1/spannerbackupschedule/spannerbackupschedules-fully/dependencies.yaml
gemmahou: Updated in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5495/commits/2eae8d8040e4370d807c0896036bae2d8cff6c9b

pkg/test/resourcefixture/testdata/basic/spanner/v1beta1/spannerbackupschedule/spannerbackupschedules-fully/update.yaml
gemmahou: Would you mind explain a bit what additional fuzz we want to add here, considering we already have the fuzz test to cover encryption_config(subfields encryption_type and kmskey will be covered as well). See my comment https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5495#discussion_r2449587162

pkg/controller/direct/spanner/backupschedule_resolverefs.go
gemmahou: CryptoKeyRef is still using the [old resolver](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/562ae865a36cfc80a7ec4554a2ea6e0fe26d9471/apis/refs/v1beta1/kmsrefs.go#L47). 

I do have an updated normalized function but it requires to use [KMSKeyRef_OneOf](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/562ae865a36cfc80a7ec4554a2ea6e0fe26d9471/apis/kms/v1beta1/kmskey_reference.go#L29)(support both manual and auto key), and requires changing to the KRM types. I plan to migrate all existing CryptoKeyRef to KMSKeyRef_OneOf gradually by services(timeline TBD), so let's keep using the old resolver for now.

pkg/controller/direct/spanner/backupschedule_resolverefs.go
```diff
@@ -0,0 +1,47 @@
...
+		key, err := refs.ResolveKMSCryptoKeyRef(ctx, reader, obj, obj.Spec.EncryptionConfig.KMSKeyRef)
+		if err != nil {
+			return err
+		}
+		obj.Spec.EncryptionConfig.KMSKeyRef = key
+	}
+
+	if obj.Spec.EncryptionConfig != nil && obj.Spec.EncryptionConfig.KMSKeyRefs != nil {
+		var keys []*refs.KMSCryptoKeyRef
+		for _, kmsKey := range obj.Spec.EncryptionConfig.KMSKeyRefs {
```
anhdle-sso: You can do this in place

PR #5497, opened 2025-10-22
tests: tf

acpana:
### BRIEF Change description

I got confused by this example when trying to use the controller overrides in my own testing.

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/operator/pkg/controllers/configconnectorcontext/configconnectorcontext_controller_test.go
+++ b/operator/pkg/controllers/configconnectorcontext/configconnectorcontext_controller_test.go
@@ -1412,17 +1412,17 @@ func TestControllerOverridesField(t *testing.T) {
 		Spec: corev1beta1.ConfigConnectorContextSpec{
 			Experiments: &corev1beta1.Experiments{
 				ControllerOverrides: map[string]k8sreconciler.ReconcilerType{
-					"BigQueryDataset.bigquery.cnrm.cloud.google.com":              "direct",    // default is terraform
-					"AlloyDBClusters.alloydb.cnrm.cloud.google.com":               "direct",    // default is terraform
-					"CloudIdentityGroups.cloudidentity.cnrm.cloud.google.com":     "terraform", // default is direct
-					"SQLInstances.sql.cnrm.cloud.google.com":                      "terraform", // default is direct
-					"CloudIdentityMembership.cloudidentity.cnrm.cloud.google.com": "dcl",       // default is direct
-					"UnknownResource.example.cnrm.cloud.google.com":               "direct",    // invalid resource
-					"AlloyDBInstance.alloydb.cnrm.cloud.google.com":               "unknown",   // invalid controller type
-					"example.cnrm.cloud.google.com":                               "unknown",   // invalid resource and invalid controller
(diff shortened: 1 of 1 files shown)


operator/pkg/controllers/configconnectorcontext/configconnectorcontext_controller_test.go
```diff
@@ -1412,17 +1412,17 @@ func TestControllerOverridesField(t *testing.T) {
...
-					"SQLInstances.sql.cnrm.cloud.google.com":                      "terraform", // default is direct
-					"CloudIdentityMembership.cloudidentity.cnrm.cloud.google.com": "dcl",       // default is direct
-					"UnknownResource.example.cnrm.cloud.google.com":               "direct",    // invalid resource
-					"AlloyDBInstance.alloydb.cnrm.cloud.google.com":               "unknown",   // invalid controller type
-					"example.cnrm.cloud.google.com":                               "unknown",   // invalid resource and invalid controller
-					"SpannerInstance.spanner.cnrm.cloud.google.com":               "terraform", // default is terraform
-					"GKEHubFeature.gkehub.cnrm.cloud.google.com":                  "dcl",       // default is dcl
-					"ComputeInstance.compute.cnrm.cloud.google.com":               "direct",    // direct is not supported for this resource
+					"BigQueryDataset.bigquery.cnrm.cloud.google.com":              "direct",  // default is terraform
+					"AlloyDBClusters.alloydb.cnrm.cloud.google.com":               "tf",      // default is terraform
```
yuwenma, line 1416: I would prefer "terraform" over "tf" for readability. It is fine to use tf in code, comments, or anywhere that does not reviewer/developer facing. But it is actually better to use "Terraform" here since it is for configuration purpose. UNLESS, we specifically want to hide the term "Terraform" for some reason.

yuwenma, line 1416: For the alloydb, if the value is the same as the default, should it be skipped ?

acpana: `terraform` is not a valid reconciler type in our routing logic 🙃 :

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/562ae865a36cfc80a7ec4554a2ea6e0fe26d9471/pkg/k8s/defaulter.go#L33

I discovered that while trying to write tests and got misdirected by this example. I don't disagree with your points I am just hoping this example doesn't confuse others!

acpana: I think this test just serializes these values it doesn't actually use any of the routing logic 🤔  see below

```
    unstructuredCCC, err := runtime.DefaultUnstructuredConverter.ToUnstructured(ccc)
	if err != nil {
		t.Fatalf("error converting to unstructured: %v", err)
	}
	_, found, err := unstructured.NestedMap(unstructuredCCC, "spec", "experiments", "controllerOverrides")
	if err != nil {
		t.Fatalf("error getting nested map: %v", err)
	}
	if !found {
		t.Fatalf("field .spec.experiments.controllerOverrides not found in unstructured object")
	}
```

PR #5513, opened 2025-10-24
test: Add tests for TF-based TagBinding

yuwenma:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/.github/workflows/ci-presubmit.yaml
+++ b/.github/workflows/ci-presubmit.yaml
@@ -1398,6 +1398,26 @@ jobs:
           path: /tmp/artifacts/
 
 
+  tests-e2e-fixtures-tags:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+      - name: "Run dev/ci/presubmits/tests-e2e-fixtures-tags"
+        run: |
+          ./dev/ci/presubmits/tests-e2e-fixtures-tags
+        env:
+          ARTIFACTS: /tmp/artifacts
+      - name: "Upload artifacts"
+        uses: actions/upload-artifact@v4
+        with:
+          name: artifacts-tests-e2e-fixtures-tags
+          path: /tmp/artifacts/
+
+
   tests-e2e-fixtures-vertexai:
     runs-on: ubuntu-latest
     timeout-minutes: 60


mockgcp/mockresourcemanager/service.go
```diff
@@ -112,5 +113,20 @@ func (s *MockService) NewHTTPMux(ctx context.Context, conn *grpc.ClientConn) (ht
...
-	return mux, nil
+	return &stripTrailingSlashHandler{mux}, nil
+}
+
+// stripTrailingSlashHandler is a middleware that removes trailing slashes from
+// the request URL path. This is necessary because the grpc-gateway mux performs
+// strict path matching, and clients (like gcloud) may inconsistently send
+// requests with a trailing slash. For example, this ensures that a request for
+// `/v3/tagBindings/` is correctly routed to the handler registered for
+// `/v3/tagBindings`.
```
yuwenma, line 124: TagBinding does not have a single "Get" call to retrieve a specific object. Instead, it calls "List" to get all the tagValues from a given parent

yuwenma: This is not the only case that "List" is used,  we can move this to the shared lib. WDYT? @justinsb

justinsb: Let's see where else it is needed.  We also have an implementation of GRPC-gateway functionality that doesn't use generated coming, needed for Firestore

mockgcp/mockresourcemanager/tagbinding.go
```diff
@@ -43,7 +41,7 @@ func (s *TagBindingsServer) normalizeParent(ctx context.Context, parent string)
 		projectWithNumber := fmt.Sprintf("//cloudresourcemanager.googleapis.com/projects/%d", project.Number)
 		return projectWithNumber, nil
 	} else {
-		return "", status.Errorf(codes.InvalidArgument, "invalid parent")
+		return parent, nil
```
justinsb, line 44: We'll have to see if this introduces regressions vs realgcp.  Not a blocker.

pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagbinding/tagbindingautogen/create.yaml
```diff
@@ -18,6 +18,7 @@ metadata:
   name: tagstagbinding-${uniqueId}
 spec:
   parentRef:
-    name: project-${uniqueId}
+    #  external: //cloudresourcemanager.googleapis.com/projects/938492827514 # If parent is project, This is the full resource name of a GCP project number not ID! 
+    external: //cloudresourcemanager.googleapis.com/projects/${projectNumber}
```
justinsb, line 22: The project ID case looks to be harder, so we should probably test that (?).  Not a blocker.

PR #5515, opened 2025-10-24
feat: Add direct controller for  TagsTagBinding (TF beta)

yuwenma:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


yuwenma:
Depends on #5513

issue #5516

apis/refs/v1beta1/projectref.go
```diff
@@ -189,6 +189,29 @@ func ResolveProjectID(ctx context.Context, reader client.Reader, obj *unstructur
 	return "", fmt.Errorf("cannot find project id for %v %v/%v", obj.GetKind(), obj.GetNamespace(), obj.GetName())
 }
 
+func ResolveProjectNumber(ctx context.Context, reader client.Reader, key types.NamespacedName) (string, error) {
```
yuwenma, line 192: The TagBinding uses the project number as its parent and composes the ID from it.

apis/tags/v1beta1/tagbinding_identity.go
```diff
@@ -13,3 +13,107 @@
...
+
+func (i *TagBindingIdentity) ParentWithFullURL() string {
+	return i.parent
+}
+
+func (i *TagBindingIdentity) TagValue() string {
+	return "tagValues/" + i.tagValue
+}
+
+func (i *TagBindingIdentity) FromExternal(ref string) error {
```
yuwenma, line 56: tagBinding ID is not named. It's a 1:1 mapping between a specific resource and TagValue. Thus, the ID  is in the form of "tagBindings/{{parent-full-url-with-url-escaped}}/tagValues/{{tagValueID}}"

apis/tags/v1beta1/tagbinding_parent.go
```diff
@@ -0,0 +1,95 @@
...
+
+package v1beta1
+
+import (
+	"fmt"
+	"strings"
+)
+
+// +k8s:deepcopy-gen=false
+type TagBindingParent interface {
```
yuwenma, line 28: TagBinding supports many parents. This is a new structure to make the parent extensible and can fit into the KCC reference solution.

pkg/controller/direct/tags/tagbinding_controller.go
```diff
@@ -0,0 +1,191 @@
...
+	actual    *resourcemanagerpb.TagBinding
+}
+
+var _ directbase.Adapter = &TagsTagBindingAdapter{}
+
+func (a *TagsTagBindingAdapter) Find(ctx context.Context) (bool, error) {
+	log := klog.FromContext(ctx)
+	log.V(2).Info("getting TagsTagBinding", "name", a.id)
+
+	req := &resourcemanagerpb.ListTagBindingsRequest{
```
yuwenma, line 106: Tagbinding has no single retrieval "GET", it only supports list.

pkg/controller/direct/tags/tagbinding_controller.go
```diff
@@ -0,0 +1,191 @@
...
+		Parent: a.id.ParentWithFullURL(),
+		// TODO: PageSize and PageToken
```
yuwenma: We should support this. But I want to do it in a follow-up
The default PageSize is 100, the server supports a maximum value of 300 (what TF-based used today).

pkg/controller/direct/tags/tagbinding_controller.go
```diff
@@ -0,0 +1,191 @@
...
+	if err != nil {
+		return false, fmt.Errorf("waiting for deletion of TagsTagBinding %s: %w", a.id, err)
+	}
+	log.V(2).Info("successfully deleted TagsTagBinding", "name", a.id)
+
+	return true, nil
+}
+
+func (a *TagsTagBindingAdapter) Export(ctx context.Context) (*unstructured.Unstructured, error) {
+	return nil, nil
```
yuwenma, line 190: we should implement the export (in a follow up)

apis/refs/v1beta1/interface.go
```diff
@@ -61,6 +61,18 @@ type Ref interface {
 	// value. If "Namespace" is not specified in the reference, the
 	// `defaultNamespace“ will be used instead.
 	Normalize(ctx context.Context, reader client.Reader, defaultNamespace string) error
+
+	// Used to get external from custom fields (managed by Terraform/DCL-based reconciler) if status.externalRef is not set.
```
justinsb: Looks like we need to decide on whether we want to add this to the permanent Ref interface.

I don't love the pattern, but another way to do this is to create a second interface, like `LegacyResourceRef`, and see if the Ref object can be cast to `LegacyResourceRef`

apis/refs/v1beta1/projectref.go
justinsb: We have a pattern "ConvertToProjectNumber" which we've been using in other places

apis/tags/v1beta1/tagbinding_identity.go
```diff
@@ -13,3 +13,107 @@
...
+	// TODO: parent can be other resources in addition to Project.
+	parent := &TagBindingProject{}
+	if err := parent.FromExternal(tokens[0]); err != nil {
+		return err
+	}
+	i.parent = parent.String()
+	if i.parent == "" {
+		return fmt.Errorf("parent was empty in external=%q", ref)
+	}
+	i.tagValue = tokens[1]
```
justinsb, line 75: So we need to check that these tokens do not include more tokens than we expected.  #5531 

Maybe we create a function / type for URL resolution?

apis/tags/v1beta1/tagbinding_identity.go
```diff
@@ -13,3 +13,107 @@
...
+	if obj.Spec.ResourceID != nil {
+		// This is a legacy special use. The prefix tagBinding is to support backward compatibility with existing resources.
+		externalRef := "tagBindings/" + *obj.Spec.ResourceID
+		identity := &TagBindingIdentity{}
+		err := identity.FromExternal(externalRef)
+		return identity, err
+
+	}
+
+	var parent string
```
justinsb: I think we should assume that `common.NormalizeReferences` has been called.

apis/tags/v1beta1/tagbinding_parent.go
```diff
@@ -0,0 +1,64 @@
...
+	Prefix() string
+
+	FromExternal(ref string) error
+}
+
+var ParentFullResourceMap = map[FullResourcePrefix]TagBindingParent{
+	ProjectPrefix: &TagBindingProject{},
+}
+
+type FullResourcePrefix string
```
justinsb, line 40: These need comments if they are exported, I think, because it isn't very obvious what they are.  Do they need to be exported?

apis/tags/v1beta1/tagbinding_parent.go
```diff
@@ -0,0 +1,64 @@
...
+type TagBindingProject struct {
+	ProjectNumber string
+}
+
+func (p *TagBindingProject) String() string {
+	return fmt.Sprintf("%s/%s", ProjectPrefix, p.ProjectNumber)
+}
+
+func (p *TagBindingProject) Prefix() string {
+	return fmt.Sprintf("%s", ProjectPrefix)
```
justinsb, line 58: Maybe just return the literal value, i.e. `return "//cloudresourcemanager.googleapis.com/projects"`

apis/tags/v1beta1/tagbinding_parent.go
```diff
@@ -0,0 +1,64 @@
...
+var _ TagBindingParent = &TagBindingProject{}
+
+// TagBindingProject represents a TagBinding parent of the form:
+// `//cloudresourcemanager.googleapis.com/projects/{{projectNumber}}`
+type TagBindingProject struct {
+	ProjectNumber string
+}
+
+func (p *TagBindingProject) String() string {
+	return fmt.Sprintf("%s/%s", ProjectPrefix, p.ProjectNumber)
```
justinsb, line 54: Maybe get the prefix by calling p.Prefix

pkg/controller/direct/tags/tagbinding_controller.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+func (m *modelTagsTagBinding) AdapterForURL(ctx context.Context, url string) (directbase.Adapter, error) {
+	return nil, nil
+}
+
+type TagsTagBindingAdapter struct {
+	id        *krm.TagBindingIdentity
+	gcpClient *resourcemanager.TagBindingsClient
+	desired   *krm.TagsTagBinding
```
justinsb: Is it cleaner if `desired` is stored as `resourcemanagerpb.TagBinding`?

pkg/controller/direct/tags/tagbinding_controller.go
justinsb: Is it a correctness issue, or an optimization?

pkg/controller/direct/tags/tagbinding_controller.go
```diff
@@ -0,0 +1,191 @@
...
+	if err != nil {
+		return fmt.Errorf("waiting for creation of TagsTagBinding %s: %w", a.id, err)
+	}
+
+	log.V(2).Info("successfully created TagsTagBinding", "name", a.id)
+
+	status := &krm.TagsTagBindingStatus{}
+	status.ExternalRef = direct.LazyPtr(a.id.String())
+
+	// TF-direct trims the prefix "tagBindings/", which is weird because then it is is invalid to retrieve the object. Shall we keep this behavior?
```
justinsb: What is "TF-direct"?

pkg/controller/direct/tags/tagbinding_controller.go
```diff
@@ -0,0 +1,191 @@
...
+	// This modified value is also the only accepted value in `spec.resourceID`.
+	status.Name = direct.LazyPtr(created.GetName())
+	return createOp.UpdateStatus(ctx, status, nil)
+}
+
+func (a *TagsTagBindingAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOperation) error {
+	log := klog.FromContext(ctx)
+	log.V(2).Info("TagsTagBinding update no-op", "name", a.id)
```
justinsb: Can we make this more verbose?  e.g. `TagsTagBinding does not have updateable fields; skipping update`

If I see this message, I don't know if this is something we need to implement or something normal

pkg/controller/direct/tags/tagbinding_controller.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	op, err := a.gcpClient.DeleteTagBinding(ctx, req)
+	if err != nil {
+		if direct.IsNotFound(err) {
+			log.V(2).Info("skipping delete for non-existent TagsTagBinding, assuming it was already deleted", "name", a.id)
+			return true, nil
+		}
+		return false, fmt.Errorf("deleting TagsTagBinding %s: %w", a.id, err)
+	}
+	err = op.Wait(ctx)
```
justinsb: Nit:
```
if err := op.Wait(ctx); err != nil {
```

pkg/controller/direct/tags/tagbinding_controller.go
justinsb: Can you create a github issue?

pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagbinding/tagbinding-project/_generated_object_tagbinding-project.golden.yaml
```diff
@@ -2,20 +2,19 @@ apiVersion: tags.cnrm.cloud.google.com/v1beta1
 kind: TagsTagBinding
 metadata:
   annotations:
+    alpha.cnrm.cloud.google.com/reconciler: direct
```
justinsb, line 5: Not for this PR, but I wonder if we should test this better now.  I feel like we can switch the controller to direct and update the golden logs, but we can still test the terraform behaviour, we just don't expect identical results on the http.log.  But we might expect that the final status is identical?

pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagbinding/tagbinding-project/_generated_object_tagbinding-project.golden.yaml
```diff
@@ -25,5 +24,6 @@ status:
     reason: UpToDate
     status: "True"
     type: Ready
-  name: '%2F%2Fcloudresourcemanager.googleapis.com%2Fprojects%2F${projectNumber}/tagValues/${tagValueID}'
-  observedGeneration: 2
+  externalRef: tagBindings/%2F%2Fcloudresourcemanager.googleapis.com%2Fprojects%2F${projectNumber}/tagValues/${tagValueID}
+  name: tagBindings/%2F%2Fcloudresourcemanager.googleapis.com%2Fprojects%2F${projectNumber}/tagValues/${tagValueID}
```
justinsb: This looks like a user-facing change to how they get the ID?

pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagbinding/tagbinding-project/_http.log
```diff
@@ -178,7 +178,25 @@ X-Xss-Protection: 0
 
 ---
 
-POST https://cloudresourcemanager.googleapis.com/v3/tagBindings?alt=json
+GET https://cloudresourcemanager.googleapis.com/v3/tagBindings?%24alt=json%3Benum-encoding%3Dint&parent=%2F%2Fcloudresourcemanager.googleapis.com%2Fprojects%2F${projectNumber}
```
justinsb: This makes sense.  Looks like Terraform just always creates the Tag, because it is immutable.  I think I prefer our approach.

pkg/controller/direct/tags/tagbinding_controller.go
yuwenma: Done.

pkg/controller/direct/tags/tagbinding_controller.go
yuwenma: Correctness. If the total tagBinding exceeds the PageSize, this may not find the tagbinding, and the Create call will return 409.

pkg/controller/direct/tags/tagbinding_controller.go
yuwenma: https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/5539

apis/refs/v1beta1/interface.go
yuwenma: May I ask the benefits of introducing a second interface? As greenfield uses `Ref` and brownfield uses `LegacyResourceRef`, how it helps on TF to direct migration, and the maintenance afterwards?  It sounds like we need to clean up the legacy code individually, which is another 300+ things to do.

apis/tags/v1beta1/tagbinding_identity.go
yuwenma: Good point! I think I did those check in the line above (line 63)

apis/tags/v1beta1/tagbinding_parent.go
yuwenma: Good question: I'm thinking about that. Basically this Prefix is the full resource name prefix. I want to move it to somewhere else. But I need some more exploration to see what's the proper way to do that.

yuwenma: I'll add the comment

apis/tags/v1beta1/tagbinding_identity.go
yuwenma: Right. This resource's parents can be any GCP service supported kind. I'd like to find out the best code structured, once I add more than one parent. Mind postponing the discussion to the next PR?

apis/tags/v1beta1/tagbinding_parent.go
yuwenma: Ditto, would you mind we holding the discussion to the next PR once there're 2 parents.

pkg/controller/direct/tags/tagbinding_controller.go
yuwenma: Fixed.

pkg/controller/direct/tags/tagbinding_controller.go
yuwenma: Make sense. Done

pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagbinding/tagbinding-project/_generated_object_tagbinding-project.golden.yaml
yuwenma: Not really. But I'm happy to discuss this (actually I left a comment in code about this).  
The tagBinding name *should* include `tagBindings`. The fact that TF-based reconciler drops it is kind of misleading and problematic: It's not the real value from gcp response.Name, and cannot be used in other places (gcloud, pantheon, etc) to retrieve the object. Considering this is an output-only in KCC status, I changed it to the actual gcp response.name

Meanwhile, the TF-based reconciler trims the prefix but adds it back when reading from `.spec.resourceID`. We also support this case (in this PR), if `.spec.resourceID` is configured, we also add back the prefix before using it. 

How about we fix this `.status.name` to the real value of the gcp name, and support both with and without prefix in `.spec.resourceID`?

pkg/controller/direct/tags/tagbinding_controller.go
yuwenma: Ah, good catch. In fact we don't need `desired`. The Create/Delete does not have payload

apis/tags/v1beta1/tagbinding_parent.go
yuwenma: Ditto, would you mind we holding the discussion to the next PR once there're 2 parents.

apis/refs/v1beta1/projectref.go
yuwenma: I saw that function but it does not work well with this resource, especially if we want to use the `identity.Identity` interface.

In short, we need to pass the projectMapper in Identity initialization (in `AdapterForObject`).

yuwenma: Besides, do you mind reminding me the problem of calling this function directly?

pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagbinding/tagbinding-project/_generated_object_tagbinding-project.golden.yaml
yuwenma: We could.  We could configure the fixture test to run  TF_based first, and then direct (with this annotation added on the fly), we could keep the _http.log for the direct controller, and have a separate _migration_diff.log to reflect the diff (TF vs Direct).

apis/refs/v1beta1/projectref.go
justinsb: Can you explain why you want to create a new pattern?

apis/tags/v1beta1/tagbinding_identity.go
justinsb: Line 63 splits on `/tagValues/`. not `/`.  So `foo/tagValues/foo/bar` goes wrong

apis/refs/v1beta1/projectref.go
yuwenma: I saw the pattern from this file is `ResolveProjectID` `ResolveProjectFromAnnotation` `ResolveProject`.  Are you the one adding those functions?  What would you recommend to use to get the ProjectNumber?

apis/tags/v1beta1/tagbinding_identity.go
yuwenma: kk. Updated.

pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagbinding/tagbinding-project/_generated_object_tagbinding-project.golden.yaml
justinsb: So the concern is that people that wanted the ID previously would compute `tagBindings/{{.status.name}}`.  Now they must just use `{{.status.name}}`

We can absolutely point them to status.externalRef as a better place to get the value.

pkg/controller/direct/tags/tagbinding_controller.go
justinsb: It _looks_ like ListTagBindings will handle that, because it returns an iterator?

pkg/controller/direct/tags/tagbinding_controller.go
justinsb: That would be nice info as a code comment

apis/tags/v1beta1/tagbinding_identity.go
justinsb: If common.NormalizeReferences has been called, we do not need to handle the case where we have to retrieve the Project object, the reference will have External.

The problem is that in the case where the reference is External, we do not convert to project number.

apis/tags/v1beta1/tagbinding_identity.go
```diff
@@ -13,3 +13,105 @@
...
+	return nil
+}
+
+var _ identity.Resource = &TagsTagBinding{}
+
+func (obj *TagsTagBinding) GetIdentity(ctx context.Context, reader client.Reader) (identity.Identity, error) {
+
+	if obj.Spec.ResourceID != nil {
+		// This is a legacy special use. The prefix tagBinding is to support backward compatibility with existing resources.
+		externalRef := "tagBindings/" + *obj.Spec.ResourceID
```
justinsb, line 88: Not a blocker, but I think we should not add `tagBindings/` if spec.resourceID already starts with `tagBindings/`; we can be liberal in what we accept here

apis/tags/v1beta1/tagbinding_identity.go
yuwenma: I used `common.NormalizeReferences` in the next PR https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5572/files#diff-e3dbee44fb4eb52af01a4805d5cac483c43213fdee726052ebf9908e8f6a14a4

The parent and identity are unified to an identit.Identity object.

pkg/controller/direct/tags/tagbinding_controller.go
yuwenma: Added the comments.

pkg/controller/direct/tags/tagbinding_controller.go
yuwenma: Sure. changed to use iterator to iterate all options.

pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagbinding/tagbinding-project/_generated_object_tagbinding-project.golden.yaml
yuwenma: Changed back to the existing format (no prefix). Also updated a description in the CRD field to recommend `status.externalRef`

Issue #5377, opened 2025-10-08
Support non-standard reference resolution


gemmahou:
### Describe your question

Current references resolver(normalizer) resolves references to a standard format, for example `projects/projectId/locations/us/resources/resourceId`. 

It's pretty common that the underlying API requires a different format: many compute resources require selfLink(full url), some requires the `ipAddress(8.8.8.8)` of ComputeAddress, or `version(1)` of SecretVersion, or `connectionName(${projectId}:us-central1:sqlinstance-activationpolicy-direct-${uniqueId})` of SQLInstance, or `email(gsa-${uniqueId}@${projectId}.iam.gserviceaccount.com)` of IAMServiceAccount.

Currently we implement special handlings to individual controllers, to convert resolved references to the non-standard format. Ref: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/develop-resources/api-conventions/resource-reference.md#ambiguous-external-usage

A question is can we implement a generic method that allows references to be resolved directly into the required format? This would eliminate the need for individual controller hacks (example PR: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5035). The `NormalizeOnTemplate` function would allow controllers to simply specify a template(tpl) to get the correct output.

As a followup, we can then create a centralized mapping between controllers and their non-standard `tpl` specifications. This would serve as a single source of documentation for all special references and potentially allow us to auto-generate the non-standard resolver logic in each controller.

PR #5517, opened 2025-10-24
feat: Make `Ref` interface fit TF/DCL migrated resources

yuwenma:
### BRIEF Change description

We want to implement the `Ref` interface for all resource reference normalization. However, the current `Ref` assumes the Reference resource is managed by direct controller, which only looks at the `.status.external` field. However, the reality is the opposite, most references are for the TF/DCL controller managed resources, that does not use `.status.externalRef`. This change is to extend the `Ref` to use a custom field path to fit the broader case.

* If your pull request is to address an open issue, indicate it by specifying the
issue number: 
* If your pull request fixes an issue which has not been filed, please file the
issue and put the number here.

Related to #5515

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/apis/firestore/v1beta1/database_reference.go
+++ b/apis/firestore/v1beta1/database_reference.go
@@ -68,3 +68,7 @@ func (r *FirestoreDatabaseRef) ValidateExternal(ref string) error {
 func (r *FirestoreDatabaseRef) Normalize(ctx context.Context, reader client.Reader, defaultNamespace string) error {
 	return refsv1beta1.Normalize(ctx, reader, r, defaultNamespace)
 }
+
+func (r *FirestoreDatabaseRef) GetExternalFromCustomFields() []string {
+	return nil
+}


apis/refs/v1beta1/interface.go
```diff
@@ -61,6 +61,18 @@ type Ref interface {
...
+	// This is used in two ways:
+	// 1. A one-off solution to support migration from Terraform/DCL-based reconciler to direct reconciler.
+	// 2. A fallback mechanism to support backward compatibility for existing resources created by Terraform/DCL-based reconciler.
+	//
+	// Returns a slice of field paths (in string format) to try to get the external reference from.
+	// Each field path is represented as a slice of strings, where each string is a field name in the path.
+	//
+	// For example, if the external reference can be found in spec.fieldA.subFieldB,
+	// the returned value should be []string{"spec", "fieldA", "subFieldB"}.
+	GetExternalFromCustomFields() []string
```
gemmahou, line 75: I saw this PR is related to references and have some thoughts I'd like to share, as I was working on something similar in #5035 and I logged an issue #5377 to address this.

1. I really like the function to now resolve references directly from their fields!!
2. Should this function be limited to TF/DCL-based resources, or will it be used for pure direct resources as well?
  - If limited to the TF/DCL, adding it to the main interface might create unnecessary work (implementing a nil return for all pure resources). Currently we are implementing custom `Normalize`([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/apigee/v1beta1/organization_reference.go#L72-L104)), which serves the same goal.
  - I feel pure direct resources should also be able to be referenced by a custom field, instead of only leverage `status.externalRef`(i.e. we could make it default to `status.externelRef`, also allow other fields). This is not only for backward compatibility, but to align with the underlying API design. We are doing refining work case-by-case today; As we have the`GetExternalFromCustomFields` function now, can we use it in pure direct resource as well, to eliminate the needs of special handling?
3. Reference like ComputeAddressRef, when referenced by ComputeNetwork, it uses its `selfLink` field, when referenced by ComputeForwardingRule, it uses its `address` field. So it might be hard to set only one target field for a reference. I feel it's easier to handle the target field configuration in individual controllers, instead of in the references.
4. We'll also need to update `ValidateExternal`(or the enforced format validation on `xxxRef.external` field), as the output from `GetExternalFromCustomFields` may not conform to the previously expected normalized format.

If we agree to support resolving custom fields for references, I'd like to share more of my design.

yuwenma: I want to enforce direct controller to use externalref as the single source of truth. The new normalization considers the case about switch back and forth among different controllers (which is what I don't want to support originally tbh) but it does save a lot of effort to fix processing and communication issues on each single reference.

Regarding the case that multiple fields could be used, I thought about that. It is rare that they can all be used directly as external (may need some modification), so it's better to have the reference write its own Normalize than using this one. 

Compute is a special case, let's handle it separately. I feel like it would add many noise to the problem that is not repeatable on any other services

apis/refs/v1beta1/interface.go
```diff
@@ -85,7 +97,14 @@ func Normalize(ctx context.Context, reader client.Reader, ref Ref, defaultNamesp
 			return fmt.Errorf("reading status.externalRef: %w", err)
 		}
 		if externalRef == "" {
-			return k8s.NewReferenceNotReadyError(u.GroupVersionKind(), key)
+			// Try to get external from legacy fields if status.externalRef is not set.
```
justinsb, line 100: Wouldn't the ref just not use `Normalize` in this case?

apis/refs/v1beta1/interface.go
gemmahou: Thanks for looking into this and the clarification. However, I personally feel multi-field references are _not uncommon_ and are not limited to ComputeAddress. I used Gemini to query TF servicemappings and be able to get some multi-field/multi-template references, I listed a few that seems a bit complex:

Note:  `{{value}}` in valueTemplate is pointing to `targetField`. If No `targetField` is specified, `spec.resourceID` is the default target field.

```  
ComputeAddress
   - `targetField: address` is used by:
       - ComputeForwardingRule
       - ComputeInstance
       - ComputeInstanceTemplate
       - DNSRecordSet
       - ServiceDirectoryEndpoint
   - `targetField: self_link` is used by:
       - ComputeInterconnectAttachment
   - No `targetField` is specified for the reference in:
       - ServiceNetworkingConnection
SecretManagerSecretVersion
   - `targetField: name` (no valueTemplate) is used by:
       - CloudBuildTrigger (in cloudbuild.yaml)
   - `targetField: version` (no valueTemplate) is used by:
       - RunJob (in run.yaml)
       - RunService (in run.yaml)  
SQLInstance
   - (no `targetField` or `valueTemplate`) is used by:
       - SQLDatabase
       - SQLSSLCert
       - SQLUser
   - `targetField: connection_name` is used by:
       - RunJob
       - RunService
StorageBucket
   - (no `targetField` or `valueTemplate`) is used by:
       - ComputeBackendBucket
       - StorageBucketAccessControl
       - StorageDefaultObjectAccessControl
       - StorageNotification
       - StorageTransferJob
   - `valueTemplate: storage.googleapis.com/{{value}}` is used by:
       - LoggingLogSink
```
So if we want to support this feature to resolve refs by their target fields, we should consider configure the target fields in individual controllers code, instead of in references code. Because I'm worried if we establish the pattern(interface), it will be hard to change later. Or maybe we could just override Normalize([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/apigee/v1beta1/organization_reference.go#L72-L104)) function to handle all of them for now, as Justin pointed.

apis/refs/v1beta1/interface.go
yuwenma: There are 300+ of them not using Normalize, and the custom code is mostly repeated except these few lines.

justinsb: OK, how about a special `NormalizeWithFieldPath` function (and it can be marked deprecated)?  We want it to drop out once the last TF resource is done.

gemmahou: For reference, I had a NormalizeOnTemplate (see implementation [here](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5035/files#diff-39e83069bbca639a3e846abda2ef4b103bc0360dce3183652d3b606de8bf1111R91)). This function allows individual controllers to pass a template (including field paths) when resolving references. It not only resolves references by fields but can also resolve them into a special format(aiming to address valueTemplate feature in TF servicemappings).

We could implement a `NormalizeWithFieldPath` for simplicity now, and expand its flexibility later.

Correct me but I feel we cannot deprecate this function, even after the TF/DCL migration. Take `TagsTagValue` as example, resolve the `TagsTagValueRef` using its `name` field is required by the API, though it might violate our standard `Normalize`. Until the API are updated to accept our standard normalized values, `NormalizeWithFieldPath` is needed. 

I also plan to refactor all special handlings we have so far to use `NormalizeWithFieldPath`/`NormalizeOnTemplate`, so we don't have special handlings everywhere in the code, making it much easier to track and maintain.

yuwenma: I got your point and like `NormalizeWithFieldPath`, @justinsb 
in fact this PR makes the work simpler (once the last TF migration is done): Remove `GetExternalFromCustomFields`  from interface, and update `Normalize`. It's done. (leave the 300+ `GetExternalFromCustomFields` as orphan code to clean up). 

 If defining the `NormalizeWithFieldPath`, basically each resource's normalize function is in the critical path and  needs to be changed. 

Reply to @gemmahou  Your proposal also mixed the legacy with the desired form (KRM-native, only from `.status.externalRef`), we want to get rid of that (ideally), or mitigate the impact by explicitly showing it in code.

yuwenma: also, I updated this PR to apply the new function to Firestore. (@justinsb you move so fast!)

PR #5519, opened 2025-10-25
tests: verify parent refs in CRDs

justinsb:
We have not historically used parentFooRef for parent references in CRDs,
preferring fooRef.

Given the number of existing references, even if parentFooRef is more
correct, it will be a very inconsistent user experience.


justinsb:
So we're trying to decide which way to standardize here.

If we go with "parents should just be fooRef" there are 10 exceptions today; 5 alpha, 5 beta (biglaketable, datacatalogpolicytag, monitoringgroup, tagstagbinding, tagstagvalue).  This is counting resources that a field named something like parentRef, parentFooRef, parentFooRefs etc.

If we go with "parents should be parentFooRef" there are 443 exceptions today; 214 alpha; 229 beta.  This is counting resources that have no fields like parentRef, parentFooRef, parentFooRefs.

cc @yuwenma @cheftako

tests/apichecks/crds_test.go
```diff
@@ -204,6 +196,79 @@ func TestCRDsDoNotHaveFooUrlRef(t *testing.T) {
...
+	if strings.HasSuffix(fieldPath, "Ref") {
+		isRef = true
+	}
+	if strings.HasSuffix(fieldPath, "Refs[]") || strings.HasSuffix(fieldPath, "Refs") {
+		isRef = true
+	}
+	return isRef
+}
+
+// CRDs should not have parentFooRef fields; use fooRef even for parent references.
```
cheftako, line 211: I'm torn. I think field names should be descriptive to help customers use the field. That is why we add ref to the field name in the first place. Its a hint to the customer that it is a reference to another KCC object which can be used to reference a GCP object. So we clearly recognize that in some circumstances we allow field names to be modified. 

However I think there are 2 reasons to keep such modifications to a minimum. The first is that we do not have first line control of the namespace, the service team does. We do not determine if an additonal (to foo) field called parentFoo would be added to the structure. The service team makes that call and we get no call in it. Very secondarily if we call it parentFoo and a user/support person tries to ask a question about parentFoo it will cause confusion. The closer we keep the field name to the original the easier it is for users to map fields, understand meaning and support themselves.

yuwenma: I'm also confused. In the previous discussion, I thought you suggest to use `parent<Kind>Ref`.  https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5302#discussion_r2408428534 

For API readability, I would vote for `.spec.parent.<Kind>Ref`,  but  KCC already take the naming convention `projectRef` and `location` as individual top-layer fields. Does this API check come from that?

justinsb: This PR is trying to figure out what it should be, it's not obvious what we should choose.

yuwenma: I missed your previous message about the stats. Then I agree with this check.

PR #5520, opened 2025-10-27
FirestoreDocument: create controller

justinsb:
- **FirestoreDocument: controller**
- **autogen: FirestoreDocument controller**


pkg/controller/direct/firestore/firestoredocument_controller.go
```diff
@@ -0,0 +1,278 @@
...
+
+func init() {
+	registry.RegisterModel(krm.FirestoreDocumentGVK, NewFirestoreDocumentModel)
+}
+
+func NewFirestoreDocumentModel(ctx context.Context, config *config.ControllerConfig) (directbase.Model, error) {
+	return &firestoreDocumentModel{config: config}, nil
+}
+
+var _ directbase.Model = &firestoreDocumentModel{}
```
cheftako, line 46: Nit: This sort of invariant seems better placed before any of the functions.

pkg/controller/direct/firestore/firestoredocument_controller.go
```diff
@@ -0,0 +1,278 @@
...
+	fqn := a.id.String()
+
+	log := klog.FromContext(ctx)
+	log.V(2).Info("updating FirestoreDocument", "name", fqn)
+
+	resource := direct.ProtoClone(a.desired)
+
+	latest := a.actual
+
+	changedFieldPaths := changedFieldPaths(a.actual, a.desired)
```
cheftako, line 178: Future: I think we need to start passing up the call stack why update is called. Wondering if it makes sense to put changeFieldsPaths into the ctx so safer upgrade and other mechanisms can determine if/why upgrade was called.

pkg/controller/direct/firestore/firestoredocument_controller.go
```diff
@@ -0,0 +1,278 @@
...
+	req := &pb.DeleteDocumentRequest{
+		Name: fqn,
+	}
+	if err := a.firestoreClient.DeleteDocument(ctx, req); err != nil {
+		if direct.IsNotFound(err) {
+			// Return success if not found (assume it was already deleted).
+			log.V(2).Info("skipping delete for non-existent FirestoreDocument, assuming it was already deleted", "name", fqn)
+			return true, nil
+		}
+		return false, fmt.Errorf("deleting FirestoreDocument %s: %w", fqn, err)
```
cheftako, line 253: Nit: is there an edge case around getting an error calling delete on a document which is being deleted but is not yet gone?

pkg/test/resourcefixture/testdata/basic/firestore/v1alpha1/firestoredocument/firestoredocument-full/dependencies.yaml
```diff
@@ -0,0 +1,23 @@
+# Copyright 2024 Google LLC
```
cheftako, line 1: Nit: Wrong copyright year.

pkg/test/resourcefixture/testdata/basic/firestore/v1alpha1/firestoredocument/firestoredocument-full/update.yaml
```diff
@@ -0,0 +1,32 @@
+# Copyright 2024 Google LLC
```
cheftako, line 1: Ditto

pkg/controller/direct/firestore/firestoredocument_controller.go
justinsb: Good call, will move it to immediately follow the type declaration, before the functions.

pkg/controller/direct/firestore/firestoredocument_controller.go
justinsb: Ah - we do have structuredreporting for this, I'l send a PR to surface it!

pkg/controller/direct/firestore/firestoredocument_controller.go
justinsb: We haven't seen one, there's no LRO so I would expect this call would either succeed once the document was deleted, or return an error in which case we would retry.

Issue #5318, opened 2025-10-06
New Resource: FirestoreDocument


justinsb:
REST: https://cloud.google.com/firestore/docs/reference/rest/v1/projects.databases.documents
Terraform: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/firestore_document

justinsb:
Note it's not entirely clear whether documents should be managed through KCC.  Is there a use case?

PR #5521, opened 2025-10-27
FirestoreDocument mockgcp support

justinsb:
- **tests: Simple standin for grpc-gateway**
- **FirestoreDocument: mockgcp support**
- **autogen: FirestoreDocument tests**

This one is interesting because grpc-gateway doesn't fully support streaming methods, but we can (relatively) easily replace grpc-gateway for the FirestoreDocument mock.

The replacement is also a little easier to work with, we don't need to compile any protos at all!


justinsb:
Issue #5318

mockgcp/common/httptogrpc/pathmatcher.go
```diff
@@ -0,0 +1,272 @@
...
+
+// componentMatcher matches a component of a path expression
+type componentMatcher interface {
+	// Match attempts to match the given tokens
+	// If matched, it populates matches with any named captures
+	// It returns true if matched, false otherwise
+	Match(tokens []string, matches map[string]string) bool
+
+	// FixedLength returns the fixed length of this component, if applicable
+	// If the component is variable length, the second return value is false
```
cheftako, line 151: Nit: I think the comment would be more clear it is said something like length is undefined for variable length, (fixed == false).

mockgcp/mockfirestore/document.go
```diff
@@ -0,0 +1,187 @@
+// Copyright 2024 Google LLC
```
cheftako, line 1: Nit: 2025

PR #5530, opened 2025-10-28
fix logginglink API (alpha), and add mapper and fuzzer

yuwenma:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/apis/logging/v1alpha1/link_types.go
+++ b/apis/logging/v1alpha1/link_types.go
@@ -80,7 +80,7 @@ type LoggingLinkObservedState struct {
 	//  linked to. This dataset will also have BigQuery Views corresponding to the
 	//  LogViews in the bucket.
 	// +kcc:proto:field=google.logging.v2.Link.bigquery_dataset
-	BigqueryDataset *BigQueryDatasetObservedState `json:"bigqueryDataset,omitempty"`
+	BigqueryDataset *BigQueryDatasetObservedState `json:"bigQueryDataset,omitempty"`
 }
 
 // +genclient
--- a/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_logginglinks.logging.cnrm.cloud.google.com.yaml
+++ b/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_logginglinks.logging.cnrm.cloud.google.com.yaml
@@ -144,7 +144,7 @@ spec:
                 description: ObservedState is the state of the resource as most recently
                   observed in GCP.
                 properties:
-                  bigqueryDataset:
+                  bigQueryDataset:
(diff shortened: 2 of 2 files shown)


apis/logging/v1alpha1/link_types.go
```diff
@@ -25,16 +25,28 @@ var LoggingLinkGVK = GroupVersion.WithKind("LoggingLink")
...
 
 	// Describes this link.
+	//
 	//  The maximum length of the description is 8000 characters.
+	// +kcc:proto:field=google.logging.v2.Link.description
 	Description *string `json:"description,omitempty"`
 
-	// The LoggingLogBucket that this Link is associated with
-	LoggingLogBucketRef *loggingv1beta1.LoggingLogBucketRef `json:"loggingLogBucketRef,omitempty"`
+	// NOTYET:  This field requires further investigation.
```
justinsb, line 41: Who added this field? We should copy them on the review

Also I don't see a controller change.  If the service is creating the BigQuery dataset, then that's legitimate I think.

yuwenma: I guess there's a misread about the code change (due to the git-diff). This comment is for bq dataset, which does not exist in the CRD yet. The deleted code is moved above (as the parent).

yuwenma: Regarding the BG dataset itself, @tylerreidwaze has confirmed that having loggingLink to create the BQ dataset is the expected feature, https://cloud.google.com/blog/products/devops-sre/introducing-cloud-loggings-log-analytics-powered-by-big-query.

justinsb: Ah, got it - thank you!

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_logginglinks.logging.cnrm.cloud.google.com.yaml
```diff
@@ -141,7 +144,7 @@ spec:
                 description: ObservedState is the state of the resource as most recently
                   observed in GCP.
                 properties:
-                  bigQueryDataset:
+                  bigqueryDataset:
```
justinsb: So this is alpha so I'm not going to block on it, but I think we have named our type `BigQueryDataset` not `BigqueryDataset`, so I think we want to put this back.  (And we might need a test to ensure consistency going forwards)

yuwenma: Fixed in this PR

justinsb (approved):

PR #5538, opened 2025-10-29
mockgcp: tests & fixes for compute backendService

justinsb:
- **mockgcp: add test for compute backend-service**
  

- **autogen: golden output for compute backend-service**
  

- **mockgcp: more fidelity for compute backendService**


justinsb:
/assign @cheftako 

The bulk of the lines of code here are in the 3rd commit, which is just updating our golden tests to accurately reflect GCP

PR #5560, opened 2025-10-31
Updating BigtableMaterializedView Resource

shavonz:
- Adding mapper between MaterializedViewSpec and MaterializedViewInfo
- Removing unused types from bigtablematerializedview

### DIFF
--- a/pkg/controller/direct/bigtable/mapper.generated.go
+++ b/pkg/controller/direct/bigtable/mapper.generated.go
@@ -495,23 +495,13 @@ func BigtableMaterializedViewSpec_v1alpha1_FromProto(mapCtx *direct.MapContext, 
 	if in == nil {
 		return nil
 	}
-	out := &krmbigtablev1alpha1.BigtableMaterializedViewSpec{}
-	// MISSING: Name
-	out.Query = direct.LazyPtr(in.GetQuery())
-	// MISSING: Etag
-	out.DeletionProtection = direct.LazyPtr(in.GetDeletionProtection())
-	return out
+	return BigtableMaterializedViewInfo_ToBigtableMaterializedViewSpec(mapCtx, BigtableMaterializedView_ToBigtableMaterializedViewInfo(mapCtx, in))
 }
 func BigtableMaterializedViewSpec_v1alpha1_ToProto(mapCtx *direct.MapContext, in *krmbigtablev1alpha1.BigtableMaterializedViewSpec) *pb.MaterializedView {
 	if in == nil {
 		return nil
 	}
-	out := &pb.MaterializedView{}
-	// MISSING: Name
-	out.Query = direct.ValueOf(in.Query)
-	// MISSING: Etag
-	out.DeletionProtection = direct.ValueOf(in.DeletionProtection)
-	return out
+	return BigtableMaterializedViewInfo_ToBigtableMaterializedView(mapCtx, BigtableMaterializedViewSpec_ToMaterializedViewInfo(mapCtx, in))
 }
 func ChangeStreamConfig_v1beta1_FromProto(mapCtx *direct.MapContext, in *pb.ChangeStreamConfig) *krm.ChangeStreamConfig {
 	if in == nil {


pkg/controller/direct/bigtable/materializedview_mapper.go
yuwenma, line 1: We need the autogenerated mapper to verify the initial mapping logic, could you revert the change in mapper.generated.go and build another mapping between the `MaterializedView` object and `MaterializedViewInfo`?

shavonz: I'm not sure I understand.

The functions in `mapper.generated.go` were removed because we are removing the unused `ObservedState`  from `BigtableMaterializedViewStatus`.

Should `ObservedState` not be removed?

yuwenma: From the (removed) autogenerated mapper, it tells that the materializedViewObservedState has an `etag` field (meaning the corresponding proto contains etag), but when I looked at the [godoc materializedViewInfo](https://pkg.go.dev/cloud.google.com/go/bigtable#MaterializedViewInfo), I don't see the `etag`. Could you help me understand this diff? Does it infer the object and the object info is not a 1 to 1 mapping here?

This is one reason we want to keep the observed state (even if empty), that it helps us discover issues like this.

shavonz: MaterializedViewInfo is what we use to communicate with the go client I believe. The actual underlying MaterializedView proto does have an `etag` field. 
https://docs.cloud.google.com/nodejs/docs/reference/bigtable/latest/bigtable/protos.google.bigtable.admin.v2.materializedview

yuwenma: Unfortunately, this is a red flag that shows KCC would support two protos and two mappings for the same KCC resource, which is not allowed. That the two protos diverge in this field is an evidence showing the rationale.  

I think what you really want is to have the KCC resource generated from the `materializedviewInfo` proto directly rather than the `materializedview` proto. Could you change this line https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/bigtable/v1alpha1/generate.sh#L30  to `--resource BigtableLogicalView:LogicalViewInfo`, it should generate the *right api and mapper without additional manual changes*  (You would need to remove the existing mapper logic to avoid the auto-code generator bypass those fields.) Let me know if you need further info on how to do it.

georgecma: On a tangential note, our other KCC resources are tied to their proto counterparts as well (even if they have an Info object: https://pkg.go.dev/cloud.google.com/go/bigtable#AuthorizedViewInfo). 

It looks like refactoring is in order for all of our resources if we want to move off of the current paradigm - maybe that can be planned work to unblock this PR?

shavonz: IIUC, to populate this type within KCC, we need `materializedviewinfo` to be a proto, but it's only a [typed struct](https://pkg.go.dev/cloud.google.com/go/bigtable#MaterializedViewInfo) exposed from the go client:


For reference, this was my attempt.
```
❯ go run . generate-types -s bigtable --resource BigtableMaterializedViewInfo:MaterializedViewInfo
I1111 21:23:04.910821  726787 generatetypescommand.go:133] "visiting proto" name="bigtable.MaterializedViewInfo"
Error: failed to find the proto message bigtable.MaterializedViewInfo: proto: not found
Usage:
   generate-types [flags]

Flags:
  -h, --help                            help for generate-types
      --output-api string               base directory for writing APIs (default "/usr/local/google/home/shavonz/go/src/github.com/shavonz/k8s-config-connector/apis/")
      --resource StorageBucket:Bucket   the KRM Kind and the equivalent proto resource separated with a colon.  e.g. for resource google.storage.v1.Bucket, the flag should be StorageBucket:Bucket.  Can be specified multiple times. (default )
  -s, --service string                  the GCP service name
      --skip-scaffold-files             skip generating scaffold files (types, refs, and identity) for all resources generated by this command

Global Flags:
  -v, --api-version string         the KRM API version. used to import the KRM API
      --config string              path to service config file, the config file will override other flags
      --proto-source-path string   path to (compiled) proto for APIs (default "/usr/local/google/home/shavonz/go/src/github.com/shavonz/k8s-config-connector/.build/googleapis.pb")

failed to find the proto message bigtable.MaterializedViewInfo: proto: not found
exit status 1
```

yuwenma, line 1: Can we use the KCC provided autogenerated-mapper function (in mapper.generated.go), and add the manual mapper between the two protos (`MaterializedView` vs `MaterializedViewInfo`)?  It is clearer to split the focus so we can guarantee the code is reviewed and validated in the right scope:  
-  KCC can review and maintain the KRM and Proto mapper logic (based on its auto-generated framework)
- BigTable ( @georgecma)  can review the mapping between the two protos bigtable owned.

shavonz: Moved custom mappers into #5663

PR #5565, opened 2025-11-03
fix: turn off compression for Prometheus metrics to reduce memory usage

justinsb:
flate compression used by default in Prometheus metrics handler can consume significant memory.


justinsb:
Memory is now: "...8687.12kB total"

(It was around 11000 previous)

PR #5569, opened 2025-11-03
tests: create scenario test for clearing sql maintenanceVersion

justinsb:
- **tests: create scenario test for clearing sql maintenanceVersion**
  

- **autogen: golden output for   tests/e2e/testdata/scenarios/sql/unmanage-maintenanceversion**

PR #5571, opened 2025-11-03
mockgcp: block empty sqlinstance patches

justinsb:


justinsb:
Also, we're making the wrong request on create!

tests/e2e/testdata/scenarios/sql/unmanage-maintenanceversion/_http00.log
acpana, line 1: `"maintenanceVersion": "NOT_VALID_VALUE"` should not be going through 😢

acpana: Ah I see, I forgot to add the annotation on CREATE here too!

https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/HEAD/tests/e2e/testdata/scenarios/sql/unmanage-maintenanceversion/script.yaml#L18-L20

I can do that as a follow up or if you want to do it and update the log, lmk!

justinsb: I don't think the annotation should interfere with Create.

We'll clean it up on https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5569 .  Right now I think it's doing the right thing.  We'll have #5569 recognize this well-known value and block it, and we'll probably have to switch this test to _not_ specify a maintenanceVersion.

PR #5572, opened 2025-11-03
feat: Support StorageBucketRef and RunJobRef as parent for TagBinding

yuwenma:
### BRIEF Change description

This PR propose the CRD to solve a problem described below. 
Add parents that requires talking to the regional GCP host. 
- Add fixture tests for StorageBucket parent and RunJob parent
- Record real gcp and mockgcp. 

issue #5300

Depends on #5515   (last 3 git-commits)

### Problem

The TagBinding parent could be [a list of different resources](https://docs.cloud.google.com/resource-manager/docs/tags/tags-supported-services) in their full resource name format (e.g. `tagBindings///storage.googleapis.com/projects/_/buckets/tagbinding-sb-${uniqueId}/tagValues/${tagValueID}`) 
Existing TagsTagBinding (TF Beta) only supports Organization, Folder and Project, defining the CRD as `.spec.parentRef.external`. 

This conflicts with the KCC typed reference fields since it does not tell the parent type.  It also makes the parent unextendable and unmanageable because the parent reference needs to map from the KCC object to the *full resource name* format in the "external" field.    

Example to use Project as parent via `external` 
```yaml
apiVersion: tags.cnrm.cloud.google.com/v1beta1
kind: TagsTagBinding
spec:
    parentRef: 
        external: //cloudresourcemanager.googleapis.com/projects/${projectNumber} // has to be projectNumber, ProjectID is not right
```

Example to use Project as parent via reference to a KCC project  
```yaml
apiVersion: tags.cnrm.cloud.google.com/v1beta1
kind: TagsTagBinding
spec:
    parentRef: 
        name: projects/${projectID} 
```

Example to use Organization as parent via reference to a KCC project  
```yaml
apiVersion: tags.cnrm.cloud.google.com/v1beta1
kind: TagsTagBinding
spec:
    parentRef: 
        name: organization/${orgID} 
```

### Solution

KCC direct controller continues support the existing  Organization, Folder and Project parent, and continue supporting the `.spec.parentRef.external` field as plain value. It introduces a list of type-specific CRD field under existing `.spec.parentRef` for the new parent reference (direct controller), and extend the  full resource name to support the new parent.   

If new parent is added, user can either refer to the KCC parent CR object, the KCC parent external identity, or the full resource name. 
 
Example to use new RunJob as parent via reference to a KCC project  (standard KRM-based reference ID)
```yaml
apiVersion: tags.cnrm.cloud.google.com/v1beta1
kind: TagsTagBinding
spec:
    parentRef: 
        runJobRef:
                name: rj-${uniqueId} 
    location:  us-central1 // (New feature) talks to  regional host
```

Example to use new RunJob as parent via reference to a KCC project   (standard KRM-based reference ID)
```yaml
apiVersion: tags.cnrm.cloud.google.com/v1beta1
kind: TagsTagBinding
spec:
    parentRef: 
        runJobRef:
                external: projects/${projectId}/locations/${location}/jobs/{$job}  # regular KCC external identity 
    location:  us-central1 // (New feature) talks to  regional host
```

Example to use  RunJob as parent via full resource name  (backward compatible to TF-beta)
```yaml
apiVersion: tags.cnrm.cloud.google.com/v1beta1
kind: TagsTagBinding
spec:
    parentRef: 
        external: //run.googleapis.com/projects/${projectId}/locations/${location}/jobs/{$job} # extension for backward compatibility purpose 
    location:  us-central1
```

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


apis/run/v1beta1/job_reference.go
```diff
@@ -70,12 +72,23 @@ func (r *JobRef) NormalizedExternal(ctx context.Context, reader client.Reader, o
...
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &obj); err != nil {
+		return "", fmt.Errorf("error converting to %T: %w", obj, err)
 	}
-	if actualExternalRef == "" {
+	var actualExternalRef string
+	// Get external from status.externalRef. This is the most trustworthy place.
+	if obj.Status.ExternalRef != nil {
+		actualExternalRef = direct.ValueOf(obj.Status.ExternalRef)
+	} else if obj.Spec.ResourceID != nil {
+		projects, err := refsv1beta1.ResolveProject(ctx, reader, otherNamespace, obj.Spec.ProjectRef)
```
justinsb, line 84: Nit: project

apis/tags/v1beta1/tagbinding_parent.go
```diff
@@ -0,0 +1,126 @@
...
+// +k8s:deepcopy-gen=false
+type TagBindingParent struct {
+	parent identity.Identity
+}
+
+func (p *TagBindingParent) String() string {
+	var fullResourceName string
+	switch p := p.parent.(type) {
+	case *refsv1beta1.Project:
+		fullResourceName = ProjectPrefix + "/" + p.ProjectID
```
justinsb, line 47: Nit: given you always add "/" to ProjectPrefix / BucketPrefix/ RunJobPrefix, consider including it in the constant

apis/tags/v1beta1/tagbinding_parent.go
```diff
@@ -0,0 +1,126 @@
...
+	case *storagev1beta1.StorageBucketIdentity:
+		// TagBindings on buckets have a special parent format with `projects/_`.
+		// The identity's ID() method returns just the bucket name.
+		fullResourceName = BucketPrefix + "/" + p.ID()
+	case *runv1beta1.JobIdentity:
+		fullResourceName = RunJobPrefix + "/" + p.String()
+	default:
+		fullResourceName = ""
```
justinsb, line 55: I suggest `klog.Fatalf` in this case

apis/tags/v1beta1/tagbinding_parent.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+const (
+	ProjectPrefix = "//cloudresourcemanager.googleapis.com/projects"
+	// OrganizationPrefix = "//cloudresourcemanager.googleapis.com/organizations"
+	BucketPrefix = "//storage.googleapis.com/projects/_/buckets"
+	RunJobPrefix = "//run.googleapis.com"
+)
+
+var _ identity.Identity = &TagBindingParent{}
```
justinsb, line 36: I wonder if we should work towards adding a `ServiceName()` to identity.Identity

apis/tags/v1beta1/tagbinding_parent.go
```diff
@@ -0,0 +1,126 @@
...
+func (r *ParentRef) Normalize(ctx context.Context, reader client.Reader, namespace string) error {
+	// Validate the full resource name.
+	if r.External != "" {
+		parent := &TagBindingParent{}
+		return parent.FromExternal(r.External)
+	}
+
+	switch {
+	case r.RunJobRef != nil:
+		external, err := r.RunJobRef.NormalizedExternal(ctx, reader, namespace)
```
justinsb, line 91: Doesn't our NormalizeReferences function already do this?

We will need to look through each of the refs checking external, but we won't need to call normalize on each.  (And do we have a function which can give us an identity.Identity from a Ref?)

apis/tags/v1beta1/tagbinding_types.go
```diff
@@ -38,15 +41,29 @@ type TagsTagBindingSpec struct {
...
+
+	// +kcc:ref=StorageBucket
+	StorageBucketRef *storagev1beta1.StorageBucketRef `json:"storageBucketRef,omitempty"`
+
+	// +kcc:ref=RunJob
+	RunJobRef *runv1beta1.JobRef `json:"runJobRef,omitempty"`
+}
+
+type ResourceManagerRef struct {
+	// Name of the referent. Can be Project or Organization. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
```
justinsb, line 62: I don't think "Can be Project or Organization" is correct?

apis/tags/v1beta1/tagbinding_types.go
```diff
@@ -24,6 +26,7 @@ var TagsTagBindingGVK = GroupVersion.WithKind("TagsTagBinding")
 // TagsTagBindingSpec defines the desired state of TagsTagBinding
 // +kcc:spec:proto=google.cloud.resourcemanager.v3.TagBinding
 type TagsTagBindingSpec struct {
+	Location *string `json:"location,omitempty"`
 	// +kcc:ref=Project
 	ParentRef *ParentRef `json:"parentRef"`
```
justinsb, line 31: It's unfortunate that this is called parentRef and not projectRef or parentProjectRef

apis/tags/v1beta1/tagbinding_types.go
```diff
@@ -38,15 +41,29 @@ type TagsTagBindingSpec struct {
...
+	// - `//cloudresourcemanager.googleapis.com/projects/{{ProjectNumber}}`,
+	// - `//cloudresourcemanager.googleapis.com/organzations/{{OrgID}}`,
+	// - `//storage.googleapis.com/projects/_/buckets/{{BucketName}}`,
+	External string `json:"external,omitempty"`
+
+	// The ResourceManager parent. Can be Project or Organization.
+	ResourceManagerRef `json:",inline"`
+
+	// +kcc:ref=StorageBucket
+	StorageBucketRef *storagev1beta1.StorageBucketRef `json:"storageBucketRef,omitempty"`
```
justinsb, line 55: So this isn't one of KCC's patterns, as far as I know.  Having project fields be at a different level of nesting is surprising.

We have a few patterns:

1) Generic parentRef - add kind to parentRef.  kind defaults to project

2) Typed refs where you can only set one.  So we would have spec.parentRef for project, spec.runJobRef for runJob, spec.storageBucketRef etc

2b) Because spec.parentRef for project is unintuitive, you can also add spec.projectRef (and again it would be an error to set both spec.parentRef and spec.projectRef)

apis/tags/v1beta1/tagvalue_reference.go
```diff
@@ -27,3 +47,74 @@ type TagsTagValueRef struct {
...
+				return k8s.NewReferenceNotReadyError(u.GroupVersionKind(), key)
+			}
+			if externalRef, _, err = unstructured.NestedString(u.Object, fieldPaths...); err != nil {
+				return k8s.NewReferenceNotReadyError(u.GroupVersionKind(), key)
+			}
+		}
+		r.SetExternal(externalRef)
+	}
+
+	return r.ValidateExternal(r.GetExternal())
```
justinsb, line 119: If we wanted a function to map from Ref -> identity.Identity, it feels like it could share much of the logic here.

mockgcp/mockresourcemanager/tagbinding.go
```diff
@@ -33,21 +33,27 @@ import (
...
+		return parent, nil
+	}
+	pTokens := strings.Split(tokens[1], "/")
+	if len(pTokens) < 1 {
+		return parent, nil
+	}
+	projectIdentifier := pTokens[0]
+	project, err := s.Projects.GetProjectByIDOrNumber(projectIdentifier)
+	if err != nil {
 		return parent, nil
```
justinsb, line 49: Nit: you probably want to log here (or just return the error, it's a mock after all), because otherwise it's going to be hard to figure out what is going on.

mockgcp/mockresourcemanager/tagbinding.go
```diff
@@ -57,16 +63,17 @@ func (s *TagBindingsServer) CreateTagBinding(ctx context.Context, req *pb.Create
...
-	parent := req.GetTagBinding().GetParent()
-	normalizedParent, err := s.normalizeParent(parent)
+	normalizedParent, err := s.normalizeParent(req.GetTagBinding().GetParent())
 	if err != nil {
 		return nil, err
 	}
 	obj.Parent = normalizedParent
 
-	obj.Name = fmt.Sprintf("tagBindings/%s/tagValues/%s", url.PathEscape(obj.Parent), strings.TrimPrefix(tagValue.Name, "tagValues/"))
+	// The name in the response should use the original parent string (with project ID).
```
justinsb, line 72: Isn't normalizedParent the number?

pkg/controller/direct/common/refs.go
```diff
@@ -138,8 +144,8 @@ func (r *refNormalizer) VisitField(path string, v any) error {
 		alertChart.AlertPolicyRef.External = refined
 	}
 
-	if alertChart, ok := v.(*krm.IncidentList); ok {
-		for i, policyRef := range alertChart.PolicyRefs {
+	if incidentList, ok := v.(*krm.IncidentList); ok {
```
justinsb, line 147: 👍

pkg/controller/direct/tags/tagbinding_controller.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+func (m *modelTagsTagBinding) client(ctx context.Context, location string) (*resourcemanager.TagBindingsClient, error) {
+	var opts []option.ClientOption
+	opts, err := m.config.RESTClientOptions()
+	if err != nil {
+		return nil, err
+	}
+	if location != "" {
+		opts = append(opts, option.WithEndpoint(fmt.Sprintf("%s-cloudresourcemanager.googleapis.com:443", location)))
```
justinsb, line 61: This is presumably why it is adding the port!

justinsb: Docs say:

> For a gRPC client, the port number is typically included in the endpoint. Example: "us-central1-speech.googleapis.com:443".
> 
> For a REST client, the port number is typically not included. Example: "https://speech.googleapis.com/".

pkg/controller/direct/tags/tagbinding_controller.go
```diff
@@ -0,0 +1,197 @@
...
+type modelTagsTagBinding struct {
+	config config.ControllerConfig
+}
+
+func (m *modelTagsTagBinding) client(ctx context.Context, location string) (*resourcemanager.TagBindingsClient, error) {
+	var opts []option.ClientOption
+	opts, err := m.config.RESTClientOptions()
+	if err != nil {
+		return nil, err
+	}
```
justinsb, line 59: A comment or link would be great on the location logic, as to why we need to do this - I've not seen this before.

apis/tags/v1beta1/tagbinding_types.go
```diff
@@ -24,6 +26,7 @@ var TagsTagBindingGVK = GroupVersion.WithKind("TagsTagBinding")
 // TagsTagBindingSpec defines the desired state of TagsTagBinding
 // +kcc:spec:proto=google.cloud.resourcemanager.v3.TagBinding
 type TagsTagBindingSpec struct {
+	Location *string `json:"location,omitempty"`
```
justinsb, line 29: Can we have a comment here (not least because it makes it into the CRD + docs)

I think this switches into a regional tag binding?  And maybe we have to match the resource, or something like that?

pkg/controller/direct/tags/tagbinding_controller.go
```diff
@@ -0,0 +1,197 @@
...
+
+var _ directbase.Adapter = &TagsTagBindingAdapter{}
+
+func (a *TagsTagBindingAdapter) Find(ctx context.Context) (bool, error) {
+	log := klog.FromContext(ctx)
+	log.V(2).Info("getting TagsTagBinding", "name", a.id)
+
+	req := &resourcemanagerpb.ListTagBindingsRequest{
+		Parent: a.id.Parent().String(),
+		// TODO: PageSize and PageToken
```
justinsb, line 114: As before

pkg/controller/direct/tags/tagbinding_controller.go
```diff
@@ -0,0 +1,197 @@
...
+	log.V(2).Info("creating TagsTagBinding", "name", a.id)
+
+	req := &resourcemanagerpb.CreateTagBindingRequest{
+		TagBinding: &resourcemanagerpb.TagBinding{
+			Parent:   a.id.Parent().String(),
+			TagValue: a.id.TagValue(),
+		},
+	}
+	op, err := a.gcpClient.CreateTagBinding(ctx, req)
+
```
justinsb, line 146: Nit: don't put a line break here, if you want a line break put it after you handle `err`

pkg/controller/direct/tags/tagbinding_controller.go
```diff
@@ -0,0 +1,197 @@
...
+	if err != nil {
+		return fmt.Errorf("waiting for creation of TagsTagBinding %s: %w", a.id, err)
+	}
+
+	log.V(2).Info("successfully created TagsTagBinding", "name", a.id)
+
+	status := &krm.TagsTagBindingStatus{}
+	status.ExternalRef = direct.LazyPtr(a.id.String())
+
+	// TF-based trims the prefix "tagBindings/", which is weird because then it is is invalid to retrieve the object. Shall we keep this behavior?
```
justinsb, line 160: > Shall we keep this behavior?
Yes, because someone may be depending on it

pkg/controller/direct/tags/tagbinding_controller.go
```diff
@@ -0,0 +1,197 @@
...
+	// This modified value is also the only accepted value in `spec.resourceID`.
```
justinsb, line 161: That we could address!

pkg/controller/direct/tags/tagbinding_controller.go
```diff
@@ -0,0 +1,197 @@
...
+	status.Name = direct.LazyPtr(created.GetName())
+	return createOp.UpdateStatus(ctx, status, nil)
+}
+
+func (a *TagsTagBindingAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOperation) error {
+	log := klog.FromContext(ctx)
+	log.V(2).Info("TagsTagBinding does not have updateable fields; skipping update", "name", a.id)
```
justinsb, line 168: Not for this PR, but we need to figure out how to handle this.  We probably should post to status, which can be as simple as returning an error.

I suspect we won't reach this point though.

PR #5577, opened 2025-11-04
feat: unmanage all `settings` fields, `maintenanceVersion`

acpana:
Note, this PR builds on top of #5288

### BRIEF Change description

Further add support to `unamange` the `settings` and `maintenanceVersion` fields.

Without this change, a CX using `state-into-spec: merge` for SQLInstance pre-direct migration may not be able to unmanage these fields.

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/controller/direct/sql/sqlinstance_controller.go
+++ b/pkg/controller/direct/sql/sqlinstance_controller.go
@@ -75,6 +75,8 @@ type FieldMetadata struct {
 
 	// preserveActualValue contains the specific logic to copy this field's value
 	// from the live GCP resource (`actual`) to the outgoing API request payload (`out`).
+
+	// if a field is nil in `out`, we treat it as a no op
 	preserveActualValue func(out *api.DatabaseInstance, actual *api.DatabaseInstance)
 }
 
@@ -93,7 +95,7 @@ var supportedUnmanageableFields = map[string]*FieldMetadata{
 		preserveActualValue: func(out *api.DatabaseInstance, actual *api.DatabaseInstance) {
 			if actual.Settings != nil {
 				if out.Settings == nil {
-					out.Settings = &api.Settings{}
+					return
 				}
 				out.Settings.Edition = actual.Settings.Edition
 			}
@@ -108,7 +110,7 @@ var supportedUnmanageableFields = map[string]*FieldMetadata{
 		preserveActualValue: func(out *api.DatabaseInstance, actual *api.DatabaseInstance) {
 			if actual.Settings != nil {
 				if out.Settings == nil {
-					out.Settings = &api.Settings{}
+					return
 				}
 				out.Settings.ActivationPolicy = actual.Settings.ActivationPolicy
 			}
@@ -118,7 +120,7 @@ var supportedUnmanageableFields = map[string]*FieldMetadata{
 		preserveActualValue: func(out *api.DatabaseInstance, actual *api.DatabaseInstance) {
 			if actual.Settings != nil {
 				if out.Settings == nil {
-					out.Settings = &api.Settings{}
+					return
 				}
 				out.Settings.BackupConfiguration = actual.Settings.BackupConfiguration
 			}
@@ -128,7 +130,7 @@ var supportedUnmanageableFields = map[string]*FieldMetadata{
 		preserveActualValue: func(out *api.DatabaseInstance, actual *api.DatabaseInstance) {
 			if actual.Settings != nil {
 				if out.Settings == nil {
-					out.Settings = &api.Settings{}
+					return
 				}
 				out.Settings.ConnectorEnforcement = actual.Settings.ConnectorEnforcement
 			}
@@ -138,7 +140,7 @@ var supportedUnmanageableFields = map[string]*FieldMetadata{
 		preserveActualValue: func(out *api.DatabaseInstance, actual *api.DatabaseInstance) {
 			if actual.Settings != nil {
 				if out.Settings == nil {
-					out.Settings = &api.Settings{}
+					return
 				}
 				out.Settings.DataCacheConfig = actual.Settings.DataCacheConfig
 			}
@@ -148,7 +150,7 @@ var supportedUnmanageableFields = map[string]*FieldMetadata{
 		preserveActualValue: func(out *api.DatabaseInstance, actual *api.DatabaseInstance) {
 			if actual.Settings != nil {
 				if out.Settings == nil {
-					out.Settings = &api.Settings{}
+					return
 				}
 				out.Settings.StorageAutoResize = actual.Settings.StorageAutoResize
 			}
@@ -158,7 +160,7 @@ var supportedUnmanageableFields = map[string]*FieldMetadata{
 		preserveActualValue: func(out *api.DatabaseInstance, actual *api.DatabaseInstance) {
 			if actual.Settings != nil {
 				if out.Settings == nil {
-					out.Settings = &api.Settings{}
+					return
 				}
 				out.Settings.StorageAutoResizeLimit = actual.Settings.StorageAutoResizeLimit
 			}
@@ -168,7 +170,7 @@ var supportedUnmanageableFields = map[string]*FieldMetadata{
 		preserveActualValue: func(out *api.DatabaseInstance, actual *api.DatabaseInstance) {
 			if actual.Settings != nil {
 				if out.Settings == nil {
-					out.Settings = &api.Settings{}
+					return
 				}
 				out.Settings.DataDiskSizeGb = actual.Settings.DataDiskSizeGb
 			}
@@ -178,7 +180,7 @@ var supportedUnmanageableFields = map[string]*FieldMetadata{
 		preserveActualValue: func(out *api.DatabaseInstance, actual *api.DatabaseInstance) {
 			if actual.Settings != nil {
 				if out.Settings == nil {
-					out.Settings = &api.Settings{}
+					return
 				}
 				out.Settings.DataDiskType = actual.Settings.DataDiskType
 			}
@@ -188,7 +190,7 @@ var supportedUnmanageableFields = map[string]*FieldMetadata{
 		preserveActualValue: func(out *api.DatabaseInstance, actual *api.DatabaseInstance) {
 			if actual.Settings != nil {
 				if out.Settings == nil {
-					out.Settings = &api.Settings{}
+					return
 				}
 				out.Settings.IpConfiguration = actual.Settings.IpConfiguration
 			}
@@ -198,7 +200,7 @@ var supportedUnmanageableFields = map[string]*FieldMetadata{
 		preserveActualValue: func(out *api.DatabaseInstance, actual *api.DatabaseInstance) {
 			if actual.Settings != nil {
 				if out.Settings == nil {
-					out.Settings = &api.Settings{}
+					return
 				}
 				out.Settings.LocationPreference = actual.Settings.LocationPreference
 			}
@@ -208,7 +210,7 @@ var supportedUnmanageableFields = map[string]*FieldMetadata{
 		preserveActualValue: func(out *api.DatabaseInstance, actual *api.DatabaseInstance) {
 			if actual.Settings != nil {
 				if out.Settings == nil {
-					out.Settings = &api.Settings{}
+					return
 				}
 				out.Settings.PricingPlan = actual.Settings.PricingPlan
 			}
@@ -218,10 +220,10 @@ var supportedUnmanageableFields = map[string]*FieldMetadata{
 		preserveActualValue: func(out *api.DatabaseInstance, actual *api.DatabaseInstance) {
 			if actual.Settings != nil && actual.Settings.BackupConfiguration != nil {
 				if out.Settings == nil {
-					out.Settings = &api.Settings{}
+					return
 				}
 				if out.Settings.BackupConfiguration == nil {
-					out.Settings.BackupConfiguration = &api.BackupConfiguration{}
+					return
 				}
 				out.Settings.BackupConfiguration.BackupRetentionSettings = actual.Settings.BackupConfiguration.BackupRetentionSettings
 			}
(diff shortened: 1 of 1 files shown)


tests/e2e/testdata/scenarios/sql/unmanage-edition/script.yaml
gemmahou, line 50: why we are specifying two duplicated fields here?

tests/e2e/testdata/scenarios/sql/unmanage-edition/script.yaml
```diff
@@ -66,4 +65,17 @@ spec:
   region: us-central1
   settings:
     tier: db-g1-small
-  resourceID: sqlinstance-${uniqueId}
\ No newline at end of file
+  resourceID: sqlinstance-${uniqueId}
+---
+# 04
```
gemmahou, line 70: I guess this TOUCH is to trigger a reconciliation, can we add this to 03(remove the annotation and trigger the reconciliation) directly?

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -38,11 +38,64 @@ import (
...
+					out.Settings = &api.Settings{}
+				}
+				out.Settings.Edition = actual.Settings.Edition
+			}
+		},
+	},
+	settingsFieldPath: {
+		paths: []string{"spec", "settings"},
+		preserveActualValue: func(out *api.DatabaseInstance, actual *api.DatabaseInstance) {
+			out.Settings = actual.Settings
```
gemmahou, line 106: Just curious, why we have a separate handling for settings.editions, but  not settings.tier?  How about if user wants to unmanage `tier` but not `editions`?

tests/e2e/testdata/scenarios/sql/unmanage-maintenanceversion/script.yaml
```diff
@@ -0,0 +1,42 @@
...
+apiVersion: sql.cnrm.cloud.google.com/v1beta1
+kind: SQLInstance
+metadata:
+  name: sqlinstance-${uniqueId}
+spec:
+  databaseVersion: POSTGRES_15
+  region: us-central1
+  settings:
+    tier: db-g1-small
+  maintenanceVersion: "NOT_VALID_VALUE"
```
gemmahou, line 25: Is this intended to be "NOT_VALID_VALUE" during creation?

tests/e2e/testdata/scenarios/sql/unmanage-edition/script.yaml
acpana: Ah that was a test we added to make sure that they annotation parsing doesn't blow up!

tests/e2e/testdata/scenarios/sql/unmanage-edition/script.yaml
acpana: Unfortunately, I was not successful to trigger the reconcile without the TOUCH force reconciliation.

pkg/controller/direct/sql/sqlinstance_controller.go
acpana: Chatted with walter offline. I think now we have a control for all settings subfields too.

tests/e2e/testdata/scenarios/sql/unmanage-maintenanceversion/script.yaml
acpana: Yes! We don't want to see this appear in the `_http00.log`

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -105,17 +422,22 @@ func (m *sqlInstanceModel) AdapterForObject(ctx context.Context, kube client.Rea
...
-			if field != unmanageEditionAnnotationValue && field != unmanageInstanceTypeAnnotationValue {
-				return nil, fmt.Errorf("unmanaging field `%s` is not supported, supported fields are: %s", field, []string{unmanageEditionAnnotationValue, unmanageInstanceTypeAnnotationValue})
+		unmanagedFieldsList := strings.Split(unmanaged, ",")
+		for _, fieldPath := range unmanagedFieldsList {
+			field, supported := supportedUnmanageableFields[fieldPath]
+			if !supported {
+				return nil, fmt.Errorf("unmanaging field `%s` is not supported, supported fields are: %v", fieldPath, sortedUnmanageableFieldPaths)
 			}
+			// Mark the field as unmanaged for this resource instance
+			field.isUnmanaged = true
```
cheftako, line 439: Future: There are probably a few optimizations we can do here. If spec.settings is set there is no point in carrying spec.settings.x. We might be able to infer spec.settings being set if all spec.settings.x are specified. The last however seems a bad idea as it is asking to have a breakage if we ever add a new field to spec.settings.

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -38,12 +39,328 @@ import (
...
+	maintenanceVersionFieldPath: {
+		preserveActualValue: func(out *api.DatabaseInstance, actual *api.DatabaseInstance) {
+			out.MaintenanceVersion = actual.MaintenanceVersion
+		},
+	},
+	editionFieldPath: {
+		preserveActualValue: func(out *api.DatabaseInstance, actual *api.DatabaseInstance) {
+			if actual.Settings != nil {
+				if out.Settings == nil {
+					out.Settings = &api.Settings{}
```
cheftako: Silly question but if the customer is not specifying the settings, do we need to bother setting it as we are here? Can we just let it pass through as empty? The case we are worried about is SIS:Merge in which case out will be populated.

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -38,12 +39,330 @@ import (
...
+
+	// preserveActualValue contains the specific logic to copy this field's value
+	// from the live GCP resource (`actual`) to the outgoing API request payload (`out`).
+
+	// if a field is nil in `out`, we treat it as a no op
+	preserveActualValue func(out *api.DatabaseInstance, actual *api.DatabaseInstance)
+}
+
+var supportedUnmanageableFields = map[string]*FieldMetadata{
+	instanceTypeFieldPath: {
```
justinsb, line 84: Nit: I personally think if a constant is single-use, it probably should be inline.  Because now I have to go check whether each constant lines up.

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -38,12 +39,330 @@ import (
...
+	maintenanceVersionFieldPath: {
+		preserveActualValue: func(out *api.DatabaseInstance, actual *api.DatabaseInstance) {
+			out.MaintenanceVersion = actual.MaintenanceVersion
+		},
+	},
+	editionFieldPath: {
+		preserveActualValue: func(out *api.DatabaseInstance, actual *api.DatabaseInstance) {
+			if actual.Settings != nil {
+				if out.Settings == nil {
+					return
```
justinsb, line 98: I don't think this is right.  If there are no settings, we will no longer ignore the value as the user requested.

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -38,12 +39,330 @@ import (
...
+			}
+		},
+	},
+}
+
+var sortedUnmanageableFieldPaths []string
+
 func init() {
+	// Pre-sort the keys for stable error messages.
+	for path := range supportedUnmanageableFields {
```
justinsb, line 361: Nit: This seems like premature optimization?

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -105,17 +424,22 @@ func (m *sqlInstanceModel) AdapterForObject(ctx context.Context, kube client.Rea
 		sqlOperationsClient: gcpClient.sqlOperationsClient(),
 		sqlInstancesClient:  gcpClient.sqlInstancesClient(),
 		sqlUsersClient:      gcpClient.sqlUsersClient(),
+		fieldMeta:           make(map[string]*FieldMetadata),
 	}
 
-	unmanaged, ok := obj.GetAnnotations()["cnrm.cloud.google.com/unmanaged"]
+	unmanaged, ok := obj.GetAnnotations()[k8s.UnmanagedFieldsList]
 	if ok && unmanaged != "" {
```
justinsb, line 431: Nit: `unmanaged := obj.GetAnnotations()[key]` would work just as well, if the key isn't in the map we get the empty string

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -105,17 +424,22 @@ func (m *sqlInstanceModel) AdapterForObject(ctx context.Context, kube client.Rea
...
-		for _, field := range unmanagedFields {
-			if field != unmanageEditionAnnotationValue && field != unmanageInstanceTypeAnnotationValue {
-				return nil, fmt.Errorf("unmanaging field `%s` is not supported, supported fields are: %s", field, []string{unmanageEditionAnnotationValue, unmanageInstanceTypeAnnotationValue})
+		unmanagedFieldsList := strings.Split(unmanaged, ",")
+		for _, fieldPath := range unmanagedFieldsList {
+			field, supported := supportedUnmanageableFields[fieldPath]
+			if !supported {
+				return nil, fmt.Errorf("unmanaging field `%s` is not supported, supported fields are: %v", fieldPath, sortedUnmanageableFieldPaths)
 			}
+			// Mark the field as unmanaged for this resource instance
```
justinsb, line 438: It looks like we are changing the shared `*FieldMetadata`

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -170,7 +494,13 @@ func (a *sqlInstanceAdapter) Create(ctx context.Context, createOp *directbase.Cr
 		if err := a.insertInstance(ctx, u, log); err != nil {
 			return err
 		}
-		if maintenanceVersion != "" {
+
+		isMaintenanceVersionUnamanaged := false
```
justinsb, line 498: I don't think _other_ values are ignored on create.  One of the problems with "unmanaged" is that it's sort of ambiguous.  Do we actually mean "ignore-after-create"?  Do we mean "always-ignore-even-on-create"? etc 


Also a nit: there's a typo here

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -316,6 +638,16 @@ func (a *sqlInstanceAdapter) Update(ctx context.Context, updateOp *directbase.Up
 		a.actual = updated
 	}
 
+	isEditionUnmanaged := false
```
justinsb, line 641: Do we still need this?  I would expect if the edition is unmanaged that the values should match

tests/e2e/script_test.go
```diff
@@ -588,6 +595,45 @@ func readObject(h *create.Harness, gvk schema.GroupVersionKind, namespace, name
 	return obj
 }
 
+// waitForReconciliationAfterPatch waits for a reconciliation to complete after an object has been patched.
+// It does this by waiting for the resourceVersion to change twice: once from the user's patch,
```
justinsb, line 599: I think I asked this before, but I guess it got lost in rebases: this feels fragile.  Should we wait for status.observedGeneration instead?

pkg/controller/direct/sql/sqlinstance_controller.go
acpana: I think we still need to do a bit of extra work as `edition` needs to be patched separately from other fields.

pkg/controller/direct/sql/sqlinstance_controller.go
acpana: When providing a `maintenanceVersion` field on CREATE, we have to break down the operation into a CREATE (without the field) and a PATCH (with the field).

Unmanaging maintenanceVersion on CREATE sounds like a very unique edge case but I thought it would be worth addressing it. AFAICT, a spec for that would look like:

```
apiVersion: sql.cnrm.cloud.google.com/v1beta1
kind: SQLInstance
metadata:
  name: sqlinstance-${uniqueId}
  annotations:
    cnrm.cloud.google.com/unmanaged: "spec.maintenanceVersion"
spec:
  ...
  maintenanceVersion: "NOT_VALID_VALUE"
  ```
  
I think arguably, we could err out (cannot unmanage and specify maintenanceVersion on CREATE) but 🤷🏼 

At least this was my read on this situation, happy to change course.

justinsb: But what's the point of specifying `spec.maintenanceVersion: NOT_VALID_VALUE` if we ignore that value on create _and_ update?

pkg/controller/direct/sql/sqlinstance_controller.go
justinsb: Right, I'm saying shouldn't the values match if the edition is unmanaged?  That's how it works for other fields, IIUC

pkg/controller/direct/sql/sqlinstance_controller.go
cheftako: I asked for this :P 

If out.Settings is nil then we know we are not in the SIS:Merge case. This is the case where there user has requested we ignore the field but also did not provide the field at all. The previous code attempted to ignore the field by setting the field to the GCP value. Right now we know not providing the field path causes the value to float and I believe the tests Alex added confirm that.

pkg/controller/direct/sql/sqlinstance_controller.go
cheftako: Good catch. I think isUnanaged only ever has a value of true in this code but if we ever allow a false this will be a problem.

tests/e2e/script_test.go
acpana: Yes I'm sorry I dropped the ball there 🙏🏼 I can wait on https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5569 or I can steal it from there?

pkg/controller/direct/sql/sqlinstance_controller.go
acpana: Hm, I think we still need the separate code block for when the edition is unmanaged and the field is in the spec.

pkg/controller/direct/sql/sqlinstance_controller.go
acpana: Chatting offline, we're making a change such that `maintenanceVersion` cannot be unmanaged on CREATE. xref #5591

pkg/controller/direct/sql/sqlinstance_controller.go
acpana: addressed in #5591

pkg/controller/direct/sql/sqlinstance_controller.go
acpana: I wanted this so the error message for when a field is not supported to be unmanaged was consistent 🙏🏼

pkg/controller/direct/sql/sqlinstance_controller.go
acpana: From offline chatter, I reverted this in #5591

PR #5583, opened 2025-11-05
Fix Bugs with Logging Link + Implement Mock

tylerreidwaze:
Thie feature implements the mock for logging link as well as fixes some bugs within the LoggingLink logic.

The next step will be to elevate this resource to Beta, but we can do that in a follow-up PR


mockgcp/mocklogging/logbucket.go
```diff
@@ -107,7 +107,8 @@ func (s *configServiceV2) GetBucket(ctx context.Context, req *pb.GetBucketReques
 	obj := &pb.LogBucket{}
 	if err := s.storage.Get(ctx, fqn, obj); err != nil {
 		if status.Code(err) == codes.NotFound {
-			return nil, status.Errorf(codes.NotFound, "Bucket `%s` does not exist", name.BucketName)
+			return nil, status.Errorf(codes.NotFound, "Bucket `%s` in location `%s` does not exist", name.BucketName, name.location)
```
tylerreidwaze: This change breaks some other tests, but this is a proper return value from the GCP API.

What should I do here?  Do we re-record GCP for the failing tests?  Ignore this small difference for my test?

pkg/controller/direct/logging/mapper.go
yuwenma, line 1: I look that you add the inline cmd to explain specialties. The main reason we prefer the autogenerated mapper is that it makes the  GCP "special" fields become very obvious. What's more, it gets rid of some common pointer and type bugs SWE introduced. 

That's said, the manual mapper function is more a exception list, that it only deals with special fields or types. And an ideal and clean mapper would to have all code comes from the auto-mapper.   Based on this, may I ask what the exception is for these manual mappers?

tylerreidwaze: Done.  Using autogenerated now

mockgcp/mocklogging/logbucket.go
tylerreidwaze: Overrode my mock to exclude this small difference

pkg/test/resourcefixture/testdata/basic/logging/v1alpha1/logginglink/basiclink/_http.log
yuwenma, line 1: Super nit: rename the test dir from basiclink to `logginglink-basic` because the fixture test runs by matching the partially directory names, running  `hack/compare fxitures/basiclink` may run other tests in the future.

This is not a blocker to this PR. Feel free to resolve in follow-up PRs

PR #5589, opened 2025-11-06
feat: Make NormalizeReference a generic solution 

yuwenma:
### BRIEF Change description

- Add a new interface for reference that needs additional refinement 
- Make `NormalizeReference` supports `Ref` and `ExternalNormalizer` by default
- Fix monitoring services log message, move it to its own package. 

Why we need this interface rather than use existing `Ref` or ` ExternalNormalizer` that normalize the reference?
* Because the projectID comes from the caller function `refNormalize`,  that it cannot be passed into the normalize function (If we do, we need  to change the other ~200 references)  

Why we cannot keep the special logic inside `NormalizeReference` function (as it is before) but use a new interface? 
* Because otherwise we cannot extend  the function NormalizeReference to other existing 200 resources (and the future 800 resources) which implements the `ExternalNormalizer` interface, since the monitoring will break (its normalize function breaks its own refine function as the interface level Normalize is called) 

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/apis/monitoring/v1beta1/alertpolicy_reference.go
+++ b/apis/monitoring/v1beta1/alertpolicy_reference.go
@@ -92,15 +92,13 @@ func (r *MonitoringAlertPolicyRef) NormalizedExternal(ctx context.Context, reade
 	if err != nil || name == "" {
 		return "", fmt.Errorf("cannot get name for referenced %s %v (status.name is empty)", u.GetKind(), u.GetNamespace())
 	}
-	return r.External, err
+	r.External = name
+	return r.External, nil
 }
 
 var _ refsv1beta1.RefinerWithProjectID = &MonitoringAlertPolicyRef{}
 
 func (r *MonitoringAlertPolicyRef) RefineWithProjectID(projectID, path string) error {
-
-	//	.Spec.ColumnLayout.Columns[].Widgets[].AlertChart.AlertPolicyRef
-
 	tokens := strings.Split(r.External, "/")
 
 	// AlertChart has a specific format requirement:
--- a/pkg/controller/direct/common/refs.go
+++ b/pkg/controller/direct/common/refs.go
@@ -114,17 +114,22 @@ type refNormalizer struct {
 }
 
 func (r *refNormalizer) VisitField(path string, v any) error {
+	// For those references that implements the refs.ExternalNormalizer.
+	// Good for all types, even if not map to a KCC object, like ProjectAndLocationRef).
 	if ref, ok := v.(refs.ExternalNormalizer); ok {
 		if _, err := ref.NormalizedExternal(r.ctx, r.kube, r.src.GetNamespace()); err != nil {
 			return fmt.Errorf("error normalizing reference at path %q: %w", path, err)
 		}
 	}
+	// For those references that implements the refs.Ref.
+	// Good for type that maps to a KRM CR object.
 	if ref, ok := v.(refs.Ref); ok {
 		if err := ref.Normalize(r.ctx, r.kube, r.src.GetNamespace()); err != nil {
 			return fmt.Errorf("error normalizing reference at path %q: %w", path, err)
 		}
 	}
 
+	// Good for types that needs refinement for a specific API field after generic normalization.
 	if ref, ok := v.(refs.RefinerWithProjectID); ok {
 		if err := ref.RefineWithProjectID(r.project.ProjectID, path); err != nil {
 			return fmt.Errorf("error refine the project ID at path %q: %w", path, err)
@@ -142,6 +147,7 @@ func (r *refNormalizer) VisitField(path string, v any) error {
 		}
 	}
 
+	// For legacy ProjectRef (with Kind)
 	if projectRef, ok := v.(*refs.ProjectRef); ok {
 		if ref, err := normalizeProjectRef(r.ctx, r.kube, r.src, projectRef); err != nil {
 			return err


pkg/controller/direct/common/refs.go
```diff
@@ -165,42 +152,3 @@ func (r *refNormalizer) VisitField(path string, v any) error {
 
 	return nil
 }
-
-// RefineMonitoringAlertPolicyRefForAlertChart refine the alertPolicy format because MonitoringDashboard's AlertChart has a specific format requirement:
-// "The format is: projects/[PROJECT_ID_OR_NUMBER]/alertPolicies/[ALERT_POLICY_ID]"
-func (r *refNormalizer) RefineMonitoringAlertPolicyRefForAlertChart(external string) (string, error) {
```
yuwenma, line 171: `RefineMonitoringAlertPolicyRefForAlertChart` and `RefineMonitoringAlertPolicyRefForIncidentList` are moved to Monitoring reference file.

apis/refs/v1beta1/interface.go
```diff
@@ -92,3 +92,7 @@ func Normalize(ctx context.Context, reader client.Reader, ref Ref, defaultNamesp
 
 	return ref.ValidateExternal(ref.GetExternal())
 }
+
+type RefinerWithProjectID interface {
```
yuwenma, line 96: Why we need this interface rather than use existing `Ref` or ` ExternalNormalizer` that normalize the reference?
* Because the projectID comes from the caller function `refNormalize`,  that it cannot be passed into the normalize function (If we do, we need  to change the other ~200 references)  

Why we cannot keep the special logic inside `NormalizeReference` function (as it is before) but use a new interface? 
* Because otherwise we cannot extend  the function NormalizeReference to other existing 200 resources (and the future 800 resources) which implements the `ExternalNormalizer` interface, since the monitoring will break itself: normalize (via parent) -->refined (via parent) --> normalize(self, fail normalization)

justinsb: So no objection to a second interface. 

* I think the function should be named NormalizeReference<foo>, not Refine<foo>
* I think the function should take a NormalizeReferenceOptions struct, for extensibility.  So maybe NormalizeReferenceWithOptions?

BTW I don't think it's a huge deal to fix these interfaces (I think it's a search and replace with some regex), but I also have no problem with doing it more incrementally by introducing a V2 interface.

pkg/controller/direct/common/refs.go
```diff
@@ -114,47 +114,40 @@ type refNormalizer struct {
...
+	// Good for all types, even if not map to a KCC object, like ProjectAndLocationRef).
+	if ref, ok := v.(refs.ExternalNormalizer); ok {
+		if _, err := ref.NormalizedExternal(r.ctx, r.kube, r.src.GetNamespace()); err != nil {
+			return fmt.Errorf("error normalizing reference at path %q: %w", path, err)
 		}
 	}
-	if alertChart, ok := v.(*krm.AlertChart); ok {
-		if r.project == nil {
-			return fmt.Errorf("must specify project for alertChart references")
+	// For those references that implements the refs.Ref.
```
justinsb, line 124: If we're adding a third interface, we should deprecate at least one of these two older interfaces!

pkg/controller/direct/common/refs.go
```diff
@@ -114,47 +114,40 @@ type refNormalizer struct {
...
+	if ref, ok := v.(refs.RefinerWithProjectID); ok {
+		if err := ref.RefineWithProjectID(r.project.ProjectID, path); err != nil {
+			return fmt.Errorf("error refine the project ID at path %q: %w", path, err)
 		}
-		alertChart.AlertPolicyRef.External = refined
 	}
 
-	if alertChart, ok := v.(*krm.IncidentList); ok {
-		for i, policyRef := range alertChart.PolicyRefs {
-			if r.project == nil {
```
justinsb, line 143: I think we have lost this check (?)

apis/monitoring/v1beta1/alertpolicy_reference.go
```diff
@@ -95,3 +95,43 @@ func (r *MonitoringAlertPolicyRef) NormalizedExternal(ctx context.Context, reade
...
+		// projects/{{projectId}}/alertPolicies/{{alertPolicyId}}
+		if tokens[1] != projectID {
+			return fmt.Errorf("resolve alertPolicy (%q) in AlertChat was not in same project", r.External)
+		}
+		return nil
+	}
+
+	// IncidentList has a specific format requirement:
+	// "Don't include the project ID prefix in the policy name. For example, use `alertPolicies/utilization`."
+	if strings.HasSuffix(path, ".IncidentList.AlertPolicyRef") {
```
justinsb, line 121: Can we fix this in the mapper instead?

apis/monitoring/v1beta1/alertpolicy_reference.go
```diff
@@ -95,3 +95,43 @@ func (r *MonitoringAlertPolicyRef) NormalizedExternal(ctx context.Context, reade
...
+			if tokens[1] != projectID {
+				return fmt.Errorf("resolve alertPolicy (%q) in incidentList was not in same project", r.External)
+			}
+		}
+		prefix := fmt.Sprintf("projects/%s/", projectID)
+		r.External = strings.TrimPrefix(r.External, prefix)
+		return nil
+	}
+
+	return nil
```
justinsb, line 136: I don't think we should silently not be doing anything

pkg/controller/direct/common/refs.go
```diff
@@ -114,47 +114,40 @@ type refNormalizer struct {
...
-			} else {
-				logsPanel.ResourceNames[i] = *ref
-			}
+	// For those references that implements the refs.ExternalNormalizer.
+	// Good for all types, even if not map to a KCC object, like ProjectAndLocationRef).
+	if ref, ok := v.(refs.ExternalNormalizer); ok {
+		if _, err := ref.NormalizedExternal(r.ctx, r.kube, r.src.GetNamespace()); err != nil {
+			return fmt.Errorf("error normalizing reference at path %q: %w", path, err)
 		}
 	}
```
justinsb, line 123: Ah, I think we should return after calling the resolver / normalizer function.  i.e. we should not be calling Refine after calling NormalizedExternal.

pkg/controller/direct/common/refs.go
yuwenma: `refs.Ref` requires GVKNN, but it works well with Gemini. 
`refs.ExternalNormalizer` allows more generic reference (and no GVKNN requirements), but it does not work well with `identity.Identity`.  
I'm actually okay with supporting both. But not a third one.

apis/monitoring/v1beta1/alertpolicy_reference.go
yuwenma: Sure. But the scope seems to go with normalization not mapper.

apis/monitoring/v1beta1/alertpolicy_reference.go
yuwenma: The Refine is only for special case. The rest of the work is covered by those two ref normalizers.

apis/monitoring/v1beta1/alertpolicy_reference.go
justinsb: I think normalization should take it to normalized form (only external set, in a format we define, including project id vs number).  Then if that isn't the format the API expects, we transform that in the controller or the mapper.

I think either we visit the fields in the controller before calling the mapper, or we embed it into the mapper.  You're right, the mapper is probably not a good place because the mapper is generated.

pkg/controller/direct/common/refs.go
justinsb: This PR is adding the third interface.

Looks like we should aim to converge on refs.Ref?

(And then this PR would be creating refs.RefWithOptions, which embed refs.Ref and adds NormalizeWithOptions)

PR #5591, opened 2025-11-06
refactor:fix: 5577 followups

acpana:
A few of the follow ups from Justin's review of #5577

Needs #5571

### DIFF
--- a/pkg/controller/direct/sql/sqlinstance_controller.go
+++ b/pkg/controller/direct/sql/sqlinstance_controller.go
@@ -397,8 +397,8 @@ func (m *sqlInstanceModel) AdapterForObject(ctx context.Context, kube client.Rea
 		fieldMeta:           make(map[string]*FieldMetadata),
 	}
 
-	unmanaged, ok := obj.GetAnnotations()[k8s.UnmanagedFieldsList]
-	if ok && unmanaged != "" {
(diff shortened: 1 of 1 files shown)


pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -691,15 +659,15 @@ func (a *sqlInstanceAdapter) Update(ctx context.Context, updateOp *directbase.Up
 		}
 	}
 
-	isMaintenanceVersionUnamanaged := false
-	if mvField, ok := a.fieldMeta[maintenanceVersionFieldPath]; ok {
-		isMaintenanceVersionUnamanaged = mvField.isUnmanaged
+	isMaintenanceVersionUnmanaged := false
```
justinsb, line 662: I think it's harmless but do we need this?  If maintenanceVersion is unmanaged then I would expect desired.Spec.MaintenanceVersion should == actual.MaintenanceVersion, because it would have been set above.

But maybe not a problem to fix in a cherry-pick.

acpana: > , because it would have been set above.

Sorry I'm confused where would the maintenanceVersion be set prior to this block in the Update func?

PR #5593, opened 2025-11-07
mockgcp: more fidelity for tags and buckets

justinsb:
- **mockgcp: more fidelity for tags and buckets**
  

- **autogen: mockgcp fixes for cloudresourcemanager & storage**

### DIFF
--- a/mockgcp/mockresourcemanager/service.go
+++ b/mockgcp/mockresourcemanager/service.go
@@ -28,12 +28,12 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/common/projects"
 	pb_v1 "github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/generated/mockgcp/cloud/resourcemanager/v1"
 	pb_v3 "github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/generated/mockgcp/cloud/resourcemanager/v3"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/mockgcpregistry"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/pkg/storage"
 )
 
 func init() {
-	// Note: we need to publish our projects store, so we build this one specially
-	// mockgcpregistry.Register(New)
+	mockgcpregistry.Register(New)
 }
 
 // MockService represents a mocked privateca service.
@@ -68,7 +68,7 @@ type TagBindingsServer struct {
 }
 
 // New creates a MockService.
-func New(env *common.MockEnvironment, storage storage.Storage) *MockService {
+func New(env *common.MockEnvironment, storage storage.Storage) mockgcpregistry.MockService {
 	s := &MockService{
 		MockEnvironment: env,
 		storage:         storage,
@@ -80,6 +80,9 @@ func New(env *common.MockEnvironment, storage storage.Storage) *MockService {
 	s.tagKeys = &TagKeys{MockService: s}
 	s.tagValues = &TagValues{MockService: s}
 	s.tagBindings = &TagBindingsServer{MockService: s}
+
+	env.Projects = s.projectsInternal
+
 	return s
 }
 
--- a/mockgcp/mock_http_roundtrip.go
+++ b/mockgcp/mock_http_roundtrip.go
@@ -73,7 +73,6 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/mockpubsublite"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/mockrecaptchaenterprise"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/mockredis"
-	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/mockresourcemanager"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/mocksecretmanager"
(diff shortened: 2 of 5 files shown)


mockgcp/common/httpmux/mux.go
```diff
@@ -32,6 +32,9 @@ type Options struct {
 	// Some older APIs do this (e.g. cloudbilling)
 	// While it likely doesn't matter, it makes golden testing easier to match.
 	EmitUnpopulated bool
+
+	// UnescapingMode controls how URL path parameters are unescaped.
+	UnescapingMode runtime.UnescapingMode
```
yuwenma, line 37: oh! This is a good idea.

mockgcp/mockresourcemanager/service.go
```diff
@@ -31,6 +31,11 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/pkg/storage"
 )
 
+func init() {
+	// Note: we need to publish our projects store, so we build this one specially
```
yuwenma: I don't fully get this comment, do you mind elaborating a little bit more? Tag and projects are both under resourcemanager service, Do you mean we want the `storage.Storage` for project to be readable from other places?

mockgcp/mockstorage/managedfolder.go
```diff
@@ -57,29 +75,117 @@ func (s *managedFolders) ListManagedFolders(ctx context.Context, req *pb.ListMan
...
+type managedFolderName struct {
+	Bucket string
+	Folder string
+}
+
+func (n *managedFolderName) String() string {
+	return fmt.Sprintf("buckets/%s/managedFolders/%s", n.Bucket, n.Folder)
+}
+
+// // parseFolderName parses a string into an folderName.
```
yuwenma: shall we clean up this?

mockgcp/mockstorage/normalize.go
```diff
@@ -18,21 +18,39 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/mockgcpregistry"
 )
 
+const PlaceholderTimestamp = "2024-04-01T12:34:56.123456Z"
+
 var _ mockgcpregistry.SupportsNormalization = &MockService{}
 
 func (s *MockService) ConfigureVisitor(url string, replacements mockgcpregistry.NormalizingVisitor) {
 	// Bucket
```
yuwenma, line 26: nit: remove `// Bucket`

mockgcp/mockstorage/objects.go
```diff
@@ -44,7 +44,7 @@ func (s *objects) ListObjects(ctx context.Context, req *pb.ListObjectsRequest) (
 func (s *objects) GetObject(ctx context.Context, req *pb.GetObjectRequest) (*pb.Object, error) {
 	// A stub implementation, just to support deletion (for now)
 
-	httpmux.SetExpiresHeader(ctx, time.Now())
+	// httpmux.SetExpiresHeader(ctx, time.Now())
```
yuwenma: Is this intended? Do you mind adding  a comment to explain what problem it gives, since the other objects still use setExpiresHeader

mockgcp/mockstorage/bucket.go
```diff
@@ -113,19 +127,30 @@ func (s *buckets) InsertBucket(ctx context.Context, req *pb.InsertBucketRequest)
...
 
+	obj.Generation = PtrTo(time.Now().UnixNano())
+
+	switch req.GetBucket().GetLocation() {
+	case "us-central1":
+		obj.Location = PtrTo("US-CENTRAL1")
+		obj.LocationType = PtrTo("region")
+		obj.Rpo = nil
+		obj.SatisfiesPZI = PtrTo(true)
+	}
```
yuwenma: Can we extend this to support some more location value and type?  Or have some way to make it more discoverable that StorageBucket can only have `us-central1` if not using the default multi-region US.

mockgcp/common/httpmux/mux.go
justinsb: Yeah, I think there's another PR that sets this but at this point we probably need to opt-in service by service.  It was messing up one of the folder requests.

mockgcp/mockstorage/managedfolder.go
justinsb: Good call - removed!

mockgcp/mockstorage/normalize.go
justinsb: This delineates the normalizations that apply to bucket resources from those that apply to other resources.

mockgcp/mockstorage/objects.go
justinsb: Removed!

justinsb: Ah, I see.  The real GCP logs don't seem to override this header, so we don't want to set it in the mocks.

mockgcp/mockstorage/bucket.go
justinsb: It's a mock, so we generally aim to emulate the behavior we know about.  But I generalized to handle locations that "look like" regions.

tests/e2e/script_test.go
```diff
@@ -353,7 +353,7 @@ func TestE2EScript(t *testing.T) {
 						appliedObjects = append(appliedObjects, obj)
 
 					default:
-						t.Errorf("unknown TEST command %q", testCommand)
+						t.Errorf("FAIL: unknown TEST command %q", testCommand)
```
justinsb, line 356: So this error will come up when we search for it :-)

mockgcp/mockresourcemanager/service.go
justinsb: So I just made it follow the pattern instead :-)

PR #5596, opened 2025-11-07
TagsTagValue: CRD/Mapper/Fuzzer

justinsb:
- **TagsTagValue: CRD/Mapper/Fuzzer**
  

- **autogen: TagsTagValue: CRD/Mapper/Fuzzer**


apis/tags/v1beta1/tagkey_types.go
```diff
@@ -0,0 +1,152 @@
...
+	ResourceID *string `json:"resourceID,omitempty"`
+
+	// Immutable. The resource name of the TagKey's parent. A TagKey can be
+	//  parented by an Organization or a Project. For a TagKey parented by an
+	//  Organization, its parent must be in the form `organizations/{org_id}`. For
+	//  a TagKey parented by a Project, its parent can be in the form
+	//  `projects/{project_id}` or `projects/{project_number}`.
+	// +kcc:proto:field=google.cloud.resourcemanager.v3.TagKey.parent
+	// +required
+	Parent *string `json:"parent,omitempty"`
```
cheftako, line 42: Should we have a ref for this?

cheftako: I mean in addition to the Parent which I think we are stuck with for compatibility reasons?

justinsb: Yes.  I suggest we have this PR be 1:1 with terraform (other than status.externalRef and status.observedGeneration going to int64).  But let's track as https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/5613

cheftako: +1. Happy to keep it 1:1 until we make direct default. This was clearly a miss on the original TF.

apis/tags/v1beta1/tagvalue_types.go
```diff
@@ -0,0 +1,138 @@
...
+
+// TagsTagValueObservedState is the state of the TagsTagValue resource as most recently observed in GCP.
+// +kcc:observedstate:proto=google.cloud.resourcemanager.v3.TagValue
+type TagsTagValueObservedState struct {
+	// NOTYET: ETAG
+	// // Optional. Entity tag which users can pass to prevent race conditions. This
+	// //  field is always set in server responses. See UpdateTagValueRequest for
+	// //  details.
+	// // +kcc:proto:field=google.cloud.resourcemanager.v3.TagValue.etag
+	// Etag *string `json:"etag,omitempty"`
```
cheftako, line 101: Nit: Currently this only seems to be set to go into the status. As such it seems safe. The description seems confusing to users since the cannot/should not "pass to prevent race conditions" fields in the status.

PR #5597, opened 2025-11-07
[wip][Safer Upgrade] Run kccmanager with all supported reconciler types within preview mode

anhdle-sso:
### BRIEF Change description

- Run all supported controller type in static config for each resource. After running each controller type, we only log the error. At the end of parent controller reconcile loop, we return empty result and nil error to avoid requeuing.
- Add rate limiters to the GCP client with qps and burst as flags for the CLI. Each API (url.Host) has its own rate limiter.
- Update output file formatting.
- Within streaming_informer, wait for cache to sync before getting from cache. This is to avoid not getting CC or CCC objects. Since we only run the reconcile loop once, this would not effect performance much.
- Make the preview CLI to listen to SIGTERM or SIGINT. This is for gracefull shutdown when running as a Job within GKE cluster.

TEST:
1. Running the preview tool with small cluster.
```
Namespace        Group                            Kind              Total   Default Controller Type   TF-Good   TF-Bad   Direct-Good   Direct-Bad   DCL-Good   DCL-Bad   Custom-Good   Custom-Bad
config-control   bigquery.cnrm.cloud.google.com   BigQueryTable     1       tf                        1         0        0             1            NA         NA        NA            NA
config-control   bigquery.cnrm.cloud.google.com   BigQueryDataset   1       tf                        1         0        1             0            NA         NA        NA            NA
config-control   spanner.cnrm.cloud.google.com    SpannerInstance   1       tf                        1         0        1             0            NA         NA        NA            NA
config-control   storage.cnrm.cloud.google.com    StorageBucket     2       tf                        1         1        NA            NA           NA         NA        NA            NA
config-control   alloydb.cnrm.cloud.google.com    AlloyDBInstance   1       direct                    NA        NA       1             0            NA         NA        NA            NA
```
 The `Total` column is the the count of resources within group-namespace-kind.
 The `TF-Good` column is the number of resources successfully reconciled by TF controller.
 The `TF-Bad` column is the number of resources unsuccessfully reconciled by TF controller (unexpected GCP request or not fully reconciled).
NA means this Group Kind is not supported by this controller type.
So on for other types of controller.
When all objects are reconciled, `Total = TF-Good + TF-Bad = Direct-Good + Direct-Bad etc.`

2. Benchmark on a cluster with ~7k resource with (1500 in each of the IAM kinds).
 QPS 20 Burst 20: 4 minutes 57.00 seconds
 QPS 10 Burst 10: 8 minutes 22.86 seconds

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/cli/preview/recorder.go
+++ b/pkg/cli/preview/recorder.go
@@ -414,7 +414,8 @@ func (r *Recorder) PreloadGKNN(ctx context.Context, config *rest.Config) error {
 			}
 			config, err := resourceControllerConfig.GetControllersForGVK(gvk)
 			if err != nil {
-				return fmt.Errorf("error getting controller config found for GroupKind %v", gvk.GroupKind())
+				klog.Warningf("error getting controller config found for GroupKind %v", gvk.GroupKind())
+				continue
 			}
 			for _, resource := range resources.Items {
 				r.ReconciledResources[GKNN{


pkg/cli/preview/rate_limiter.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"fmt"
+	"net/url"
+
+	"golang.org/x/time/rate"
+)
+
+func (c *interceptingGCPClient) getOrCreateRateLimiter(u *url.URL) (*rate.Limiter, error) {
+	if u == nil {
+		return nil, fmt.Errorf("nil URL")
```
justinsb, line 26: Nit: I do think a panic or klog.Fatalf is fine here.  If we hit this something has gone very wrong.

pkg/cli/preview/recorder.go
```diff
@@ -206,19 +209,30 @@ func (r *Recorder) recordReconcileEnd(ctx context.Context, u *unstructured.Unstr
...
 	r.reconcileTrackerMutex.Lock()
-	defer r.reconcileTrackerMutex.Unlock()
-	r.ReconciledResources[gknn] = true
-	r.RemainResourcesCount--
+	r.ReconciledResources[gknn][t] = true
+	r.reconcileTrackerMutex.Unlock()
+
+	if r.GKNNDoneReconcile(gknn) {
+		r.reconcileTrackerMutex.Lock()
+		r.RemainResourcesCount--
```
justinsb, line 222: This field feels superfluous, it's usually better to "waste" some CPU time and avoid storing things twice, but ... not a blocker

pkg/cli/preview/result.go
```diff
@@ -20,74 +20,182 @@ import (
...
 	"strings"
 	"text/tabwriter"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/resourceconfig"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
+type GKNSummary struct {
+	// number of resouces successfully reconciled by TF reconciler
```
justinsb, line 30: Nit: typos in resources

pkg/cli/preview/result.go
```diff
@@ -20,74 +20,182 @@ import (
...
+		dcl = fmt.Sprintf("%d", *s.dclCount)
+		dclBad = fmt.Sprintf("%d", total-*s.dclCount)
+	}
+	custom := "NA"
+	customBad := "NA"
+	if s.customCount != nil {
+		custom = fmt.Sprintf("%d", *s.customCount)
+		customBad = fmt.Sprintf("%d", total-*s.customCount)
+	}
+	return fmt.Sprintf("%s\t%s\t%s\t%s\t%s\t%s\t%s\t%s", tf, tfBad, direct, directBad, dcl, dclBad, custom, customBad)
```
justinsb, line 65: Nit: The order should be direct then TF then DCL, so that the columns "drop off" from the right hand side and don't get renumbered over time.

pkg/cli/preview/result.go
```diff
@@ -20,74 +20,182 @@ import (
...
 type PreviewSummary struct {
-	totalGood int
-	totalBad  int
-	// namespace -> group -> kind -> []name
-	goodGKNN map[string]map[string]map[string][]string
-	badGKNN  map[string]map[string]map[string][]string
-	reports  map[GKNN][]string
+	// namespace -> group -> kind -> GKNSummary
+	// Count the number of resources successfully reconciled by each reconciler type
+	summary map[string]map[string]map[string]*GKNSummary
```
justinsb, line 77: Nit: Why not `map[GKN]*GKNSummary`

pkg/cli/preview/result.go
```diff
@@ -20,74 +20,182 @@ import (
...
+		countPtr = &summary.tfCount
+	case k8s.ReconcilerTypeDCL:
+		countPtr = &summary.dclCount
+	case k8s.ReconcilerTypeDirect:
+		countPtr = &summary.directCount
+	default:
+		countPtr = &summary.customCount
+	}
+
+	if r.good {
```
justinsb, line 112: Suggestion for the future: `type PreviewSummary` has probably hit the limit of what we can sustainably maintain.  You may want to just try a totally different approach.  Splitting it into a type with a few methods and a private data representation is a good first step there.

pkg/cli/preview/streaming_informer.go
```diff
@@ -288,7 +288,9 @@ func (i *streamingInformer) Get(ctx context.Context, key client.ObjectKey, obj c
 	if len(opts) != 0 {
 		return fmt.Errorf("options not implemented: %v", opts)
 	}
-
+	if !i.WaitForCacheSync(ctx) {
```
justinsb, line 291: Ah - nice!

justinsb: (Although I'm surprised that we hit this ... are we missing a wait in our real controllers?  Or does the normal informer implicitly wait?)

pkg/controller/parent/controller.go
```diff
@@ -265,3 +270,66 @@ func (r *ParentReconciler) determineControllerType(ctx context.Context, u *unstr
 	}
 	return config.DefaultController, nil
 }
+
+// ReconcilePreviewMode reconciles the resource in preview mode
```
justinsb, line 274: Blocker: this doesn't feel maintainable, we now have two Reconcile methods.

Can we inject a fake CCC or use a similar existing mechanism?

justinsb:
So one blocker, sadly right at the end (two Reconcile methods).

I suggest breaking this into a set of PRs, one for each logical idea.  For example qps/burst feels very separable.  Then it's much easier to review and you get your code merged much faster.  Right now when we find an approach that doesn't need two Reconcile methods, we also need to re-review burst/qps (and the printing changes etc)

You could even split out the printing changes into their own PRs, even though everything will only be reconciled with one reconciler until the "try both reconcilers" PR lands.

I do think that the presentation could also be more user-focused.  I think the user wants to know whether there's a problem if they upgrade KCC, and separately whether there's a longer-term problem if the switch to direct controllers.

PR #5598, opened 2025-11-08
chore: normalize unusual value in FirestoreDocument fuzzer

justinsb:


justinsb:
Was seeing occasional failures in our fuzzer from this case, where the wrapper was non-nil but the oneof was nil.

/assign @cheftako

PR #5599, opened 2025-11-08
mockgcp: improve fidelity for artifactregistry

justinsb:


justinsb:
Not terribly important, but came up in the context of another resource IIRC.  Should all be test code.

/assign @cheftako

PR #5606, opened 2025-11-10
feat: Add references resolver in ComposerEnvironment direct controller 

gemmahou:
### BRIEF Change description

Resolve references for ComposerEnvironment and add full test case.

Related issue: https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/5500

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

Ran test against realGCP and confirmed it passed. Excluded realGCP log from this PR because it contained lots of real IAM policy and member information of my google project.

### DIFF
--- a/pkg/controller/direct/composer/environment_resolverefs.go
+++ b/pkg/controller/direct/composer/environment_resolverefs.go
@@ -16,10 +16,11 @@ package composer
 
 import (
 	"context"
+	"strings"
+
 	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/composer/v1beta1"
 	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


yuwenma:
/hold

This will introduce another bug to ComposerEnvironment because the  reference Compute resources has different project in `spec` and `status` (ProjectNumber vs ProjectID) and could end up as infinite update calls to the ComposerEnvironment server.

Meanwhile, this PR depends on the decisions on using `NormalizeReference`.

maqiuyujoyce:
> > This will introduce another bug to ComposerEnvironment because the reference Compute resources has different project in `spec` and `status` (ProjectNumber vs ProjectID) and could end up as infinite update calls to the ComposerEnvironment server.
> 
> Why would the referenced ComputeNetwork and ComputeSubnetwork cause issue in ComposerEnvironment, as they are all resolved to the expected values(`projects/{projectId}/global/networks/{networkId}` and `projects/{projectId}/regions/{regionId}/subnetworks/{subnetworkId}`)?

@gemmahou , do you mind first running against real API and then mock API to verify the mock works as expected?
I think that'll also help validate the returned network and subnetwork in Composer Environment contains project ID but not project number.

PR #5614, opened 2025-11-12
TagKey: implement direct controller

justinsb:


mockgcp/common/httpmux/mux.go
```diff
@@ -170,7 +170,6 @@ func (m *ServeMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		if k == "$alt" {
 			for _, v := range values {
 				if v == "json;enum-encoding=int" {
-					klog.Infof("found %q=%q, will convert to Accept header", k, v)
```
justinsb, line 173: It was just spammy

pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagkey/tagkeyorgautogen/_generated_object_tagkeyorgautogen.golden.yaml
```diff
@@ -3,7 +3,6 @@ kind: TagsTagKey
 metadata:
   annotations:
     cnrm.cloud.google.com/management-conflict-prevention-policy: none
-    cnrm.cloud.google.com/state-into-spec: absent
```
justinsb, line 6: This file shows the change switching from TF to direct 🎉

pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagkey/tagkeyorgautogen/_final_object_old_controller.golden.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: tags.cnrm.cloud.google.com/v1beta1
```
justinsb, line 1: This file _should_ be the same as the old _generated_object file (but github doesn't show it)

pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagkey/tagkeyorgautogen/_http.log
```diff
@@ -1,4 +1,4 @@
-POST https://cloudresourcemanager.googleapis.com/v3/tagKeys?alt=json
+POST https://cloudresourcemanager.googleapis.com/v3/tagKeys?%24alt=json%3Benum-encoding%3Dint
```
justinsb, line 1: This file shows the http changes going from TF to direct; pretty minimal 🎉

pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagkey/tagkeyorgautogen/_http_old_controller.log
```diff
@@ -0,0 +1,146 @@
+POST https://cloudresourcemanager.googleapis.com/v3/tagKeys?alt=json
```
justinsb, line 1: This file should be the same as the old _http.log, but again git / github doesn't show it :-(

justinsb: `git log -C -p` does show it (-C shows copies)

```
...
diff --git a/pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagkey/tagkeyorgautogen/_http.log b/pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagkey/tagkeyorgautogen/_http_old_controller.log
similarity index 100%
copy from pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagkey/tagkeyorgautogen/_http.log
copy to pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagkey/tagkeyorgautogen/_http_old_controller.log
...
```

pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagkey/tagkeyprojectautogen/_http.log
```diff
@@ -1,10 +1,35 @@
-POST https://cloudresourcemanager.googleapis.com/v3/tagKeys?alt=json
+GET https://cloudresourcemanager.googleapis.com/v3/projects/${projectId}?%24alt=json%3Benum-encoding%3Dint
```
justinsb, line 1: I'm not entirely sure how tf doesn't need to look up the project id -> project number, I'm guessing it just doesn't check.

This diff gets smaller if we turn on project caching #5660

tests/e2e/unified_test.go
```diff
@@ -84,27 +84,27 @@ func TestAllInSeries(t *testing.T) {
 			// TODO(b/259496928): Randomize the resource names for parallel execution when/if needed.
 
 			t.Run(sampleKey.Name, func(t *testing.T) {
-				ctx := addTestTimeout(ctx, t, subtestTimeout, sampleKey.Name)
+				ctx := addTestTimeout(ctx, t, subtestTimeout, sampleKey.TestKey)
```
justinsb, line 87: The change to move from Name -> TestKey is what enables us to better check the full test path, and are split out in #5672 if we want to review them separately.

pkg/controller/direct/directbase/operations.go
```diff
@@ -98,6 +99,23 @@ func (o *CreateOperation) RecordUpdatingEvent() {
 	r.Event(o.object, corev1.EventTypeNormal, k8s.Updating, k8s.UpdatingMessage)
 }
 
+func (o *CreateOperation) SetSpecResourceID(ctx context.Context, resourceID string) error {
+	log := klog.FromContext(ctx)
+
+	u := o.GetUnstructured()
+	log.Info("setting resourceID after creation", "resourceID", resourceID)
+
+	if err := unstructured.SetNestedField(u.Object, resourceID, "spec", "resourceID"); err != nil {
```
yuwenma, line 113: Do you mind adding more context about this "controller writes to spec" logic? We believe we have tried very hard to not do this (even for `.spec.resourceID`).

justinsb: Have we written down the policy?

yuwenma: We have the doc about why not use `state-into-spec`, and because some resourceID is server-generated, so that field falls into the same category.  Can I ask what this policy should for? Maybe I miss a broader case. But basically I think if we do not open the door, it does not need explain.

justinsb: I think you make a good point on spec.resourceID.  Ideally we would not write back to spec.  I thought we had to for compatibility.  But in general, if you're catching things in code review, we want to make sure that these things are tested and/or written down.  If they're written down we can probably have the AI do an automated review, and testing is obviously much more accurate but we also have to explicitly enumerate all the cases, whereas with AI maybe we could make a more generic statement.

cc @cheftako for overlap with reviewbot

docs/ai/server-generated-id.md
```diff
@@ -0,0 +1,19 @@
+When a resource has a server-generated ID, we need to store the generated ID into the object.
```
justinsb: I created a doc to describe what I think @yuwenma is proposing for how we should handle spec.resourceID.  There are definitely trade-offs here, so we should decide that this is worth it.

@yuwenma 
@cheftako

docs/ai/server-generated-id.md
```diff
@@ -0,0 +1,19 @@
...
+
+Legacy controllers would write back to spec.resourceID, which is the same field that users can specify if they want to adopt
+an existing resource.  However, this makes the field ownership of the spec.resourceID field unclear: did the user specify it, or did KCC write it back?
```
cheftako: +1. We are no longer able to determine/infer if the object is expected to exist or not.

docs/ai/server-generated-id.md
```diff
@@ -0,0 +1,19 @@
...
+
+Instead, we now write to status.externalRef.  The downside is that this is technically a breaking change between terraform and direct controllers,
```
cheftako: Pedantic: I don't think writing to status.externalRef is a breaking change. Its failing to write to spec.resourceID which is the breaking change.

docs/ai/server-generated-id.md
```diff
@@ -0,0 +1,19 @@
...
+so if we create an object with the direct controller we can't go back to using the terraform controller (it will create a second object).
+We might also break user workflows if they are expecting to see the ID in the existing spec.resourceID field.
+
+We could instead write to both spec.resourceID _and_ status.externalRef.  If we did this then we could move between terraform and direct freely,
```
cheftako: +1. I think there are 2 flavors of this. We could write(/read) from both in Direct or we could write(/read) from both in TF.

docs/ai/server-generated-id.md
```diff
@@ -0,0 +1,19 @@
...
+and we would not break existing workflows.  The downside is that we do carry forward the field ownership ambiguity.
+Perhaps we could address this as part of moving from v1beta1 to v1.
```
cheftako: Is there a way we could tie this to SIS:Merge? (As a possible alternate solution)

docs/ai/server-generated-id.md
```diff
@@ -0,0 +1,19 @@
+When a resource has a server-generated ID, we need to store the generated ID into the object.
+
+Legacy controllers would write back to spec.resourceID, which is the same field that users can specify if they want to adopt
+an existing resource.  However, this makes the field ownership of the spec.resourceID field unclear: did the user specify it, or did KCC write it back?
+
+Instead, we now write to status.externalRef.  The downside is that this is technically a breaking change between terraform and direct controllers,
+so if we create an object with the direct controller we can't go back to using the terraform controller (it will create a second object).
```
yuwenma: +1. Yes this is the problem. Our previous decision to not write back to `.spec.resourceID` is under the condition that the TF/DCL to direct is a one-way path, and users take their own risk to switch back to Terraform. I know SQL forces the condition resets to some degree. But with the preview tool, do we still want to officially support the switching-back to TF path? That means many other issues will come up as well (diff between TF and direct on gcp default values, opt-in to TF special handling for certain fields ,etc), and these are larger cost.  
 
FYI,  regarding switching back to TF will create a second object, it is not true for `SecretManangerSecretVersion`. Just tested this resource, it passes TF (created by direct controller, that filled in `.status.externalRef`) without a new object, and the `.spec.resourceID` is refilled by TF controller to the right value.

cheftako: >under the condition that the TF/DCL to direct is a one-way path

Sadly(?) I don't think that the Direct to TF switch was ever a one-way path. [Compatability Version](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/features/compatabilityversion.md) means that at least since 1.126 users have been able to switch to older behavior. We also know that on older versions that 1.126, setting the label on the CRD would revert the behavior.

docs/ai/server-generated-id.md
justinsb: Good point, I will clarify.  I meant that _not_ writing to spec.resourceID is the breaking change, that does not come across.

docs/ai/server-generated-id.md
justinsb: If we just wrote in TF to both places, we would need to get everyone using an updated version of the TF controllers, and then move to Direct.

I do like reading in TF, though that is probably a more complicated change.  But it would work...

justinsb: (Though it does not address other readers of spec.resourceID - other controllers or other tooling)

docs/ai/server-generated-id.md
justinsb: We could, but I think it puts a lot of work on the user.  I think writing to spec.resourceID and status.externalRef breaks the least things, we just have to swallow our pride on writing to spec.resourceID for a little longer (for terraform resources with server-generated IDs)

docs/ai/server-generated-id.md
justinsb: > Our previous decision to not write back to .spec.resourceID

Did we write this down in the repo?

I do think that switching back and forth between terraform and direct is likely best avoided, but I also think we can aim for compatibility.  I think that requires writing to both spec.resourceID and status.externalRef.

I am going to create another version of this doc basically reversing the position, saying we should write to spec.resourceID and status.externalRef.

justinsb: >  it is not true for SecretManangerSecretVersion. Just tested this resource, it passes TF (created by direct controller, that filled in .status.externalRef) without a new object, and the .spec.resourceID is refilled by TF controller to the right value.

Do we know how this happened?

docs/ai/server-generated-id.md
```diff
@@ -0,0 +1,30 @@
...
+Legacy controllers would write this ID to spec.resourceID,
+which is the same field that users can specify if they want to adopt an existing resource.
+However, this makes the field ownership of the spec.resourceID field unclear: did the user specify it, or did KCC write it back?
+
+For greenfield controllers, we are introducing status.externalRef.
+This field will consistently give the GCP resource ID, as well as indicating whether a resource is under KCC control.
+
+The question is whether we should also write to spec.resourceID.
+To avoid a behavioural change, when moving from terraform to direct, we will write to both status.externalRef _and_ spec.resourceID.
+We want to avoid two problems:
```
yuwenma, line 13: nit: it's a little bit implicit about why behavior change matters? It is because we want to support users to switch back and forth among controllers. If only from TF to direct, there's no duplicate resource creation and we are fine.

docs/ai/server-generated-id.md
```diff
@@ -0,0 +1,30 @@
...
+
+* It should be possible to create a GCP resource with the direct controller and then revert back to the terraform controller (without creating a new GCP resource)
+
+* Tooling or other controllers should continue to see the resourceID in the existing spec.resourceID field.
+
+Even if we could address the first problem (for example by teaching terraform to read status.externalRef),
+the only way we have to solve the user expectations problem is with time (a "deprecation" period).
+
+Instead, we accept that although we would prefer never to write to spec, we will continue
+to write spec.resourceID for server-generated-id resources that we are migrating from DCL or Terraform to Direct.
```
yuwenma, line 23: What if the TF-controller can handle the migration without `.spec.resourceID` (like SecretManagerSecretVersion, no new object is created and TF fill back the `.spec.resourceID` using the existing ID).

 Then it is unnecessary for direct-controller to write back to `.spec.resourceID`.

docs/ai/server-generated-id.md
```diff
@@ -0,0 +1,30 @@
...
+We will not (currently) write to spec for non-server-generated-id resources, nor for greenfield Direct controllers.
+Nor are we intending to "open the door" for writing fields _other_ than spec.resourceID from direct controllers.
```
yuwenma, line 25: 💯 💯 💯 💯

docs/ai/server-generated-id.md
yuwenma: Actually I think we shouldn't have direct controllers to write back to `.spec.resourceID` if TF can deal with it. Because that's a cleaner path for v1.

apis/tags/v1beta1/tagkey_identity.go
```diff
@@ -43,38 +43,59 @@ func (i *TagsTagKeyIdentity) String() string {
...
+	// Get desired resource ID
+	resourceID := common.ValueOf(obj.Spec.ResourceID)
 
-	newIdentity.TagKey = common.ValueOf(obj.Spec.ResourceID)
-	if newIdentity.TagKey == "" {
-		newIdentity.TagKey = obj.GetName()
-	}
-	if newIdentity.TagKey == "" {
-		return nil, fmt.Errorf("cannot resolve resource ID")
+	// Server-generated ID; do not fallback to name
```
cheftako, line 67: **LATER**: It seems like we should keep metadata such as `func (obj *TagsTagKey) UsesServerGeneratedId() { return true }` Eventually some of this code could then be made to be part of the framework.

apis/tags/v1beta1/tagkey_identity.go
```diff
@@ -43,38 +43,59 @@ func (i *TagsTagKeyIdentity) String() string {
...
 		if err := statusIdentity.FromExternal(externalRef); err != nil {
 			return nil, fmt.Errorf("cannot parse existing externalRef=%q: %w", externalRef, err)
 		}
-		if statusIdentity.String() != newIdentity.String() {
-			return nil, fmt.Errorf("existing externalRef=%q does not match the identity resolved from spec: %q", externalRef, newIdentity.String())
+	}
+
+	if specIdentity != nil {
+		if statusIdentity != nil && statusIdentity.String() != specIdentity.String() {
+			return nil, fmt.Errorf("existing externalRef=%q does not match the identity resolved from spec: %q", externalRef, specIdentity.String())
```
cheftako, line 91: **LATER** : This seems like an assertion more than a simple Error. It seems like we should consider finding a way to signal this back. Propagating a higher level error (Eg on the CCC), sending an email or at least some sort of metric indicating a violated assertion.

docs/ai/server-generated-id.md
```diff
@@ -0,0 +1,30 @@
+When a resource has a server-generated ID, we need to store the generated ID into the kubernetes object,
+so that we can associate the GCP object with the kubernetes object on subsequent reconciliations.
+
+Legacy controllers would write this ID to spec.resourceID,
+which is the same field that users can specify if they want to adopt an existing resource.
+However, this makes the field ownership of the spec.resourceID field unclear: did the user specify it, or did KCC write it back?
+
+For greenfield controllers, we are introducing status.externalRef.
+This field will consistently give the GCP resource ID, as well as indicating whether a resource is under KCC control.
```
cheftako, line 9: **ALL** new controllers should be writing to status.externalRef.

docs/ai/server-generated-id.md
```diff
@@ -0,0 +1,30 @@
...
+
+The question is whether we should also write to spec.resourceID.
+To avoid a behavioural change, when moving from terraform to direct, we will write to both status.externalRef _and_ spec.resourceID.
+We want to avoid two problems:
+
+* It should be possible to create a GCP resource with the direct controller and then revert back to the terraform controller (without creating a new GCP resource)
+
+* Tooling or other controllers should continue to see the resourceID in the existing spec.resourceID field.
```
cheftako, line 17: While we are encouraging ALL KCC users to switch their usage to status.externalRef, we should not be tieing their migration of when to instantaneously switch their application for a specific field to a particular KCC update. Especially as that KCC update may be outside their control and may occur without their being warned.

docs/ai/server-generated-id.md
cheftako: Not clear how long we would have to write to .spec.resourceID. At least a year? Not sure we want to put a hold on TF -> Direct controllers for a year.

docs/ai/server-generated-id.md
```diff
@@ -0,0 +1,30 @@
...
+
+Even if we could address the first problem (for example by teaching terraform to read status.externalRef),
+the only way we have to solve the user expectations problem is with time (a "deprecation" period).
+
+Instead, we accept that although we would prefer never to write to spec, we will continue
+to write spec.resourceID for server-generated-id resources that we are migrating from DCL or Terraform to Direct.
+We will not (currently) write to spec for non-server-generated-id resources, nor for greenfield Direct controllers.
```
cheftako, line 24: This is true but the "(currently)" seems more irresolute than I think we should be.  I would rather this was something like "We do not intend to write to spec for non-server-generated-id resources, nor for greenfield Direct controllers unless there is a very compelling reason."

docs/ai/server-generated-id.md
cheftako: Again seems a bit irresolute. "This edge case for writing to spec should not be used as an excuse for writing any other spec fields."

pkg/controller/direct/directbase/operations.go
cheftako: I believe this has not  been written down with the docs/ai/server-generated-id.md change.

pkg/controller/direct/tags/tagkey_controller.go
```diff
@@ -0,0 +1,353 @@
...
+
+func (m *TagsTagKeyModel) AdapterForObject(ctx context.Context, reader client.Reader, u *unstructured.Unstructured) (directbase.Adapter, error) {
+	tagKeysClient, err := newTagKeysClient(ctx, m.config)
+	if err != nil {
+		return nil, err
+	}
+
+	obj := &krm.TagsTagKey{}
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &obj); err != nil {
+		return nil, fmt.Errorf("error converting to %T: %w", obj, err)
```
cheftako, line 63: Don't we know we are converting to TagsTagKey? I would think that u.Object is the more interesting thing to log.

pkg/controller/direct/tags/tagkey_controller.go
```diff
@@ -0,0 +1,353 @@
...
+
+	if err := common.NormalizeReferences(ctx, reader, obj, nil); err != nil {
+		return nil, fmt.Errorf("normalizing references: %w", err)
+	}
+
+	// normalize parent (sadly not a ref, for legacy reasons)
+	{
+		parent := direct.ValueOf(obj.Spec.Parent)
+		tokens := strings.Split(parent, "/")
+		if len(tokens) == 2 && tokens[0] == "projects" {
```
cheftako, line 74: NIT: len(tokens) == 2 needs to always be true. If we check that first (once) we could probably give a more targeted error and have less complicated conditional statements.

pkg/controller/direct/tags/tagkey_controller.go
```diff
@@ -0,0 +1,353 @@
...
+		latest = updated
+	}
+
+	return a.updateStatus(ctx, updateOp, latest)
+}
+
+func (a *TagsTagKeyAdapter) updateStatus(ctx context.Context, op directbase.Operation, latest *pb.TagKey) error {
+	status := &krm.TagsTagKeyStatus{}
+
+	// NOTYET: observedState
```
cheftako, line 240: NIT: would be nice to indicate when, rather than just NOTYET. Also would be nice to have something to alert us when we should come back and fix.

pkg/controller/direct/tags/tagkey_controller.go
```diff
@@ -0,0 +1,353 @@
...
+	// 		return mapCtx.Err()
+	// 	}
+	// }
+
+	status.ExternalRef = direct.PtrTo(latest.GetName())
+
+	// Legacy status fields
+	status.Name = direct.PtrTo(strings.TrimPrefix(latest.GetName(), "tagKeys/"))
+	status.NamespacedName = direct.PtrTo(latest.GetNamespacedName())
+	status.CreateTime = direct.PtrTo(latest.GetCreateTime().AsTime().Format("2006-01-02T15:04:05Z07:00"))
```
cheftako, line 254: Where is "2006-01-02T15:04:05Z07:00" coming from? This isn't a test, this doesn't seem like it belongs in the actual controller.

apis/tags/v1beta1/tagkey_identity.go
justinsb: Agreed, we do this in almost our controllers though.  It is a bad situation though, it's likely if we pick either value we will "leak" the resource corresponding to the other value.

docs/ai/server-generated-id.md
justinsb: +1

apis/tags/v1beta1/tagkey_identity.go
cheftako: For **later** a metric seems like something we could alert on. Essentially keep an assertions violated metric with a tag indicating a bounded set of assertions. If it ever goes non zero, we have a real problem we need to investigate.

pkg/controller/direct/tags/tagkey_controller.go
justinsb: Yes, but I'm trying to optimize for self-contained conditional statements rather than for efficiency here.  We end up with a pattern-matching expression on each line, and IMO it's relatively easy to scan and see which case we're handling in a long list.

pkg/controller/direct/tags/tagkey_controller.go
justinsb: Yes; it's of a compromise here because we don't concretely know when to add it, we just know we want to remain terraform compatible as far as possible right now.

pkg/controller/direct/tags/tagkey_controller.go
justinsb: So this is go's way of specifying timestamps.  Java and the Java ecosystem had already taken all the other ways, so they had to invent this one :-)

I honestly don't remember writing it, I _think_ I copied it over from the existing direct controller (which we are replacing here)

pkg/controller/direct/tags/tagkey_controller.go
justinsb: It's a good point.  But the type of u.Object is always going to be unstructured, so there's not a lot of value there either.  We use %T to avoid having to include TagsTagKey in the format string:  I have a hope in the back of my mind that we can move this to a common utility function, or just have the common reconciler convert to TagsTagKey once generics are a bit more mature.

justinsb:
I had queued up some replies but didn't have good wifi on the plane.  I don't think there are any code changes that are needed now, but will push a new PR with the docs changes to tighten that up!

PR #5615, opened 2025-11-12
FirestoreDocument: fix numeric precision handling

justinsb:
The fuzzer found an issue where large integers were being converted to floats
and losing precision. This change improves the handling of numeric types during
the mapping process to preserve integer precision.

### DIFF
--- a/pkg/controller/direct/firestore/firestoredocument_mappers.go
+++ b/pkg/controller/direct/firestore/firestoredocument_mappers.go
@@ -15,7 +15,11 @@
 package firestore
 
 import (
+	"bytes"
 	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
 
 	pb "cloud.google.com/go/firestore/apiv1/firestorepb"
 	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/firestore/v1alpha1"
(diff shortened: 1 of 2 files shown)


pkg/controller/direct/firestore/firestoredocument_mappers.go
```diff
@@ -35,16 +39,123 @@ func FirestoreDocumentSpec_v1alpha1_FromProto(mapCtx *direct.MapContext, in *pb.
 	for k, v := range in.Fields {
 		outV := Field_FromProto(mapCtx, v)
 
-		j, err := json.Marshal(outV)
+		j, err := toJSON(outV)
 		if err != nil {
-			return nil
+			mapCtx.Errorf("failed to marshal field in FirestoreDocument: %v", err)
```
cheftako: Would having the k/v help here?

pkg/controller/direct/firestore/firestoredocument_mappers.go
```diff
@@ -35,16 +39,123 @@ func FirestoreDocumentSpec_v1alpha1_FromProto(mapCtx *direct.MapContext, in *pb.
...
+		render = buffer.String()
+		return apiextensionsv1.JSON{Raw: buffer.Bytes()}, nil
+	case float32, float64:
+		// Force float to be formatted with scientific notation,
+		// so we don't convert back to int accidentally.
+		var buffer bytes.Buffer
+		fmt.Fprintf(&buffer, "%g", in)
+		render = buffer.String()
+		return apiextensionsv1.JSON{Raw: buffer.Bytes()}, nil
+	}
```
cheftako, line 102: Are we deliberately letting unrecognized types fall through? Even if we are I think a 'default:' with a comment that says that is easier to understand.

justinsb: Good call!

pkg/controller/direct/firestore/firestoredocument_mappers.go
justinsb: Good call!

justinsb:
Thanks @cheftako !

PR #5622, opened 2025-11-13
chore: use service normalize in tests on kube objects

justinsb:
- **chore: use service normalize in tests on kube objects**
  

- **autogen: update golden output for sharing normalizer**

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/managedkafka/v1beta1/managedkafkacluster/managedkafkacluster-basic/_generated_object_managedkafkacluster-basic.golden.yaml
+++ b/pkg/test/resourcefixture/testdata/basic/managedkafka/v1beta1/managedkafkacluster/managedkafkacluster-basic/_generated_object_managedkafkacluster-basic.golden.yaml
@@ -17,9 +17,9 @@ spec:
     accessConfig:
       networkConfigs:
       - subnetworkRef:
-          name: computesubnetwork-${uniqueId}-1
+          name: ${subnetworkID}
       - subnetworkRef:
-          name: computesubnetwork-${uniqueId}-2
+          name: ${subnetworkID}
   location: us-central1
   projectRef:
     external: ${projectId}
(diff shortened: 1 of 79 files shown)


justinsb:
/assign @cheftako 

Sorry that this one is verbose.  By extracting normalization values from the http logs, we can better (and more easily) normalize the KRM object.  Right now I am just adding the normalizer to the ones we use for KRM objects, but I think we could actually _probably_ replace the existing KRM normalization logic entirely - or at least the bulk of it!

justinsb:
This PR blocks a bunch of other tests, which blocks a bunch of controller.

pkg/test/resourcefixture/testdata/basic/managedkafka/v1beta1/managedkafkacluster/managedkafkacluster-basic/_generated_object_managedkafkacluster-basic.golden.yaml
```diff
@@ -17,9 +17,9 @@ spec:
     accessConfig:
       networkConfigs:
       - subnetworkRef:
-          name: computesubnetwork-${uniqueId}-1
+          name: ${subnetworkID}
       - subnetworkRef:
-          name: computesubnetwork-${uniqueId}-2
+          name: ${subnetworkID}
```
yuwenma, line 22: I like the ID normalization in general, only that it makes the code looks like a Magic and make those slice field (like this) looks wrong.  What do you think?

justinsb: What do you suggest?

yuwenma: I suggest we add a doc to explain the rationale and expected behavior so we don't need to worry about giving solutions for all the corner cases.

PR #5623, opened 2025-11-13
fix: Change single fields to references for BatchJob 

yuwenma:
### BRIEF Change description

This PR is to prepare BatchJob for Beta promotion.
Next step is to add more fixture test coverage to verify the spec fields.  

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


pkg/controller/direct/batch/job_mapper.go
```diff
@@ -61,3 +61,22 @@ func AllocationPolicy_ToProto(mapCtx *direct.MapContext, in *krm.AllocationPolic
...
+func AllocationPolicy_Disk_ToProto(mapCtx *direct.MapContext, in *krm.AllocationPolicy_Disk) *pb.AllocationPolicy_Disk {
+	if in == nil {
+		return nil
+	}
+	out := &pb.AllocationPolicy_Disk{}
+	if in.ImageRef != nil {
+		out.DataSource = &pb.AllocationPolicy_Disk_Image{
+			Image: in.ImageRef.External,
+		}
+	}
```
yuwenma, line 87: The proto has `dataSource` as an interface.

apis/batch/v1alpha1/job_types.go
```diff
@@ -148,8 +149,7 @@ type AllocationPolicy_Disk struct {
 	//  * `batch-debian`: use Batch Debian images.
```
justinsb: The short names are going to be interesting!

apis/compute/v1beta1/computeimage_identity.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/GoogleCloudPlatform/k8s-config-connector/apis/common/identity"
+)
+
+// TODO: We should verify the ComputeImageIdentity before use.
+// Currently this struct is not for the ComputeImage direct controller but only used as reference validation (FromExternal)
+
+var _ identity.Identity = &ComputeImageIdentity{}
+
+type ComputeImageIdentity struct {
+	// +required
```
justinsb: This is not a KRM type, we probably don't want +required (?)

apis/compute/v1beta1/computeimage_identity.go
```diff
@@ -0,0 +1,53 @@
...
+	return fmt.Sprintf("projects/%s/global/images/%s", i.ProjectID, i.Name)
+}
+
+func (i *ComputeImageIdentity) FromExternal(s string) error {
+	if s == "" {
+		return fmt.Errorf("value cannot be empty")
+	}
+	tokens := strings.Split(s, "/")
+	// image name can contain "/"
+	if tokens[0] != "projects" || tokens[2] != "global" || tokens[3] != "images" {
```
justinsb: Must check len here

apis/compute/v1beta1/computeimage_identity.go
```diff
@@ -0,0 +1,53 @@
...
+func (i *ComputeImageIdentity) String() string {
+	return fmt.Sprintf("projects/%s/global/images/%s", i.ProjectID, i.Name)
+}
+
+func (i *ComputeImageIdentity) FromExternal(s string) error {
+	if s == "" {
+		return fmt.Errorf("value cannot be empty")
+	}
+	tokens := strings.Split(s, "/")
+	// image name can contain "/"
```
justinsb: Really?  Do you have an example.. I though if there was a slash it was the image family

apis/compute/v1beta1/computeimage_reference.go
```diff
@@ -0,0 +1,111 @@
...
+		return fmt.Errorf("reading status.externalRef: %w", err)
+	}
+	if externalRef == "" {
+		if externalRef, _, err = unstructured.NestedString(u.Object, "status", "selfLink"); err != nil {
+			return err
+		}
+		if externalRef == "" {
+			return k8s.NewReferenceNotReadyError(u.GroupVersionKind(), key)
+		}
+		externalRef = strings.TrimPrefix(externalRef, "https://www.googleapis.com/compute/v1/")
```
justinsb, line 105: Do we do any validation here (or on status.externalRef?)

pkg/controller/direct/batch/job_controller.go
```diff
@@ -96,6 +96,10 @@ func (m *jobModel) AdapterForObject(ctx context.Context, reader client.Reader, u
 		return nil, err
 	}
 
+	// TODO: The ComputeNetworkRef and ComputeSubnetworkRef misses normalization.
```
justinsb, line 99: Is this because they don't support the new interfaces?

pkg/controller/direct/batch/job_mapper.go
justinsb: You could try settting the field name as dataSourceRef, see if the mapper is then auto-generated.  But ... not a blocker.

apis/batch/v1alpha1/job_types.go
justinsb: (Not a blocker)

yuwenma: I can remove this special value right now, and add the support later (with computeImage Reference Refinement)

pkg/controller/direct/batch/job_mapper.go
yuwenma: Done. Thanks

pkg/controller/direct/batch/job_controller.go
yuwenma: Yes.

apis/compute/v1beta1/computeimage_reference.go
yuwenma: Added validation (but we don't have direct controller for this resource yet)

apis/compute/v1beta1/computeimage_identity.go
yuwenma: Good catch.

apis/compute/v1beta1/computeimage_identity.go
yuwenma: Yes, that's the image family. My original attempt is to make it just simple so I can move forward with the actual resource `BatchJob`, and don't need to deal with the "zone" in image family (was confused by the gcloud "zone" flag, `gcloud compute images describe-from-family IMAGE_FAMILY_NAME  --project=IMAGE_PROJECT --zone=ZONE`, thought the image family supports locations)

I took a deeper look at the image URL.  I can support both image and image family (seems to be more preferred).

apis/compute/v1beta1/computeimage_reference.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	u := &unstructured.Unstructured{}
+	u.SetGroupVersionKind(r.GetGVK())
+	if err := reader.Get(ctx, key, u); err != nil {
+		if apierrors.IsNotFound(err) {
+			return k8s.NewReferenceNotFoundError(u.GroupVersionKind(), key)
+		}
+		return fmt.Errorf("reading referenced %s %s: %w", r.GetGVK(), key, err)
+	}
+	// Get external from status.externalRef. This is the most trustworthy place.
```
justinsb: Not a blocker, but I think you mean "this is our new preferred location".  They should be equally trustworthy if populated, and I think selfLink is more likely to be populated.

pkg/controller/direct/batch/mapper.generated.go
```diff
@@ -96,9 +96,7 @@ func AllocationPolicy_Disk_ToProto(mapCtx *direct.MapContext, in *krm.Allocation
 		return nil
 	}
 	out := &pb.AllocationPolicy_Disk{}
-	if oneof := AllocationPolicy_Disk_Image_ToProto(mapCtx, in.Image); oneof != nil {
-		out.DataSource = oneof
-	}
+	// MISSING: Image
```
justinsb: I meant to rename the go field but not the json tag, because I think this trick lets the generated mappers work correctly.

pkg/controller/direct/batch/mapper.generated.go
```diff
@@ -84,7 +84,7 @@ func AllocationPolicy_Disk_FromProto(mapCtx *direct.MapContext, in *pb.Allocatio
 		return nil
 	}
 	out := &krm.AllocationPolicy_Disk{}
-	out.Image = direct.LazyPtr(in.GetImage())
+	// MISSING: Image
```
justinsb: Do we need a fuzzer?

yuwenma: Already have the fuzzer added in the previous PR. https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/batch/job_fuzzer.go

pkg/controller/direct/batch/mapper.generated.go
yuwenma: Let me correct my previous message: Neither `ImageRef` nor `DataSourceRef` in the KRM go field can have mapper automatically generate the functions. This is because our mapper generator cannot handle interface (flattened) and reference (checking APIs) at the same time yet ( this is doable, so I can improve this in a separate PR).

Regarding this resource, the field `DataSource` is an interface, which has two objects implementing this interface. Both objects has a single field ([`AllocationPolicy_Disk_Snapshot` with `snapshot` field, and  `AllocationPolicy_Disk_Image` with `image` field](https://pkg.go.dev/cloud.google.com/go/batch@v1.13.0/apiv1/batchpb#AllocationPolicy_Disk_Snapshot)). 
In the job_types.go, they are flattened. 

I reverted this PR with the original attempt to use ImageRef and a manual mapper.

apis/compute/v1beta1/computeimage_identity.go
```diff
@@ -0,0 +1,64 @@
...
+package v1beta1
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/apis/common/identity"
+)
+
+// TODO: We should verify the ComputeImageIdentity before use.
```
justinsb, line 24: Nit: I don't think you need this statement, we aren't going to do anything differently because it is here.

apis/compute/v1beta1/computeimage_reference.go
```diff
@@ -0,0 +1,109 @@
...
+	if err := reader.Get(ctx, key, u); err != nil {
+		if apierrors.IsNotFound(err) {
+			return k8s.NewReferenceNotFoundError(u.GroupVersionKind(), key)
+		}
+		return fmt.Errorf("reading referenced %s %s: %w", r.GetGVK(), key, err)
+	}
+	// `status.externalRef` is the preferred field to store externalRef. Since ComputeImage is not yet migrated to
+	// direct controller, so this field does not exist. We will use `.status.selfLink` instead.
+	externalRef, _, err := unstructured.NestedString(u.Object, "status", "externalRef")
+	if err != nil || externalRef == "" {
```
justinsb, line 98: Nit: I think you can just check `if externalRef == ""`, you're not using `err`

pkg/controller/direct/batch/job_mapper.go
```diff
@@ -61,3 +62,34 @@ func AllocationPolicy_ToProto(mapCtx *direct.MapContext, in *krm.AllocationPolic
 	out.Tags = in.Tags
 	return out
 }
+func AllocationPolicy_Disk_FromProto(mapCtx *direct.MapContext, in *pb.AllocationPolicy_Disk) *krm.AllocationPolicy_Disk {
```
justinsb, line 65: I still think this looks like it should be auto-generated, but let's merge and then I'll check what I'm missing.

justinsb: Ah - saw your note above about the oneof - yes, you're right!  It got close, but not right, so 👍 to just manually generating

PR #5626, opened 2025-11-13
FirestoreDatabase: add support for delete protection state

justinsb:
- **FirestoreDatabase: add support for delete_protection_state**
- **chore: FirestoreDatabase controller: use status helpers etc**

### DIFF
--- a/pkg/controller/direct/firestore/firestoredatabase_controller.go
+++ b/pkg/controller/direct/firestore/firestoredatabase_controller.go
@@ -28,17 +28,13 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
 
 	apiv1 "cloud.google.com/go/firestore/apiv1/admin"
-	firestorepb "cloud.google.com/go/firestore/apiv1/admin/adminpb"
+	pb "cloud.google.com/go/firestore/apiv1/admin/adminpb"
 	"google.golang.org/protobuf/types/known/fieldmaskpb"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-)
-
-const (
-	ctrlName = "firestoredatabase-controller"
 )
 
 func init() {
@@ -106,13 +102,13 @@ type Adapter struct {
 	id                   *krm.FirestoreDatabaseIdentity
 	firestoreAdminClient *apiv1.FirestoreAdminClient
 	desired              *krm.FirestoreDatabase
-	actual               *firestorepb.Database
+	actual               *pb.Database
 }
 
 var _ directbase.Adapter = &Adapter{}
 
 func (a *Adapter) Find(ctx context.Context) (bool, error) {
-	log := klog.FromContext(ctx).WithName(ctrlName)
+	log := klog.FromContext(ctx)
 
 	if a.id == nil {
 		return false, nil
@@ -121,8 +117,8 @@ func (a *Adapter) Find(ctx context.Context) (bool, error) {
 	fqn := a.id.String()
 
 	log.V(2).Info("getting FirestoreDatabase", "name", fqn)
-	req := &firestorepb.GetDatabaseRequest{Name: fqn}
-	firestoredatabasepb, err := a.firestoreAdminClient.GetDatabase(ctx, req)
+	req := &pb.GetDatabaseRequest{Name: fqn}
+	actual, err := a.firestoreAdminClient.GetDatabase(ctx, req)
 	if err != nil {
 		if direct.IsNotFound(err) {
 			return false, nil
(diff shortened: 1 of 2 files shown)


apis/firestore/v1beta1/database_types.go
```diff
@@ -42,6 +42,51 @@ type FirestoreDatabaseSpec struct {
 	// Whether to enable the PITR feature on this database.
 	// See https://cloud.google.com/firestore/docs/reference/rest/v1/projects.databases#pointintimerecoveryenablement for more info.
 	PointInTimeRecoveryEnablement *string `json:"pointInTimeRecoveryEnablement,omitempty"`
+
+	// State of delete protection for the database.
```
cheftako, line 46: Nit: Should we indicate its optional in the description?

apis/firestore/v1beta1/database_types.go
```diff
@@ -42,6 +42,51 @@ type FirestoreDatabaseSpec struct {
...
+	// // +kcc:proto:field=google.firestore.admin.v1.Database.tags
+	// Tags map[string]string `json:"tags,omitempty"`
+
+	// // This checksum is computed by the server based on the value of other
+	// //  fields, and may be sent on update and delete requests to ensure the
+	// //  client has an up-to-date value before proceeding.
+	// // +kcc:proto:field=google.firestore.admin.v1.Database.etag
+	// Etag *string `json:"etag,omitempty"`
+
+	// // Immutable. The edition of the database.
```
cheftako, line 86: Nit: Should this be a status field rather than a spec field?

docs/ai/implement-fuzzer-field.md
```diff
@@ -0,0 +1,17 @@
+# How to add an unimplemented field in a fuzzer
```
cheftako, line 1: Comment: Ideally we should have added this in a separate PR. Not sure we would want to revert this if we reverted the rest of the PR.

justinsb: Agree, and I'm likely being lazy.  It is nice though to show the prompt & the output in the same PR, IMO.

apis/firestore/v1beta1/database_types.go
justinsb: We tend to document (inconsistently) when fields are required, with the implication that everything else is optional.

pkg/controller/direct/firestore/firestoredatabase_controller.go
```diff
@@ -107,13 +102,13 @@ type Adapter struct {
...
 	desired              *krm.FirestoreDatabase
-	actual               *firestorepb.Database
+	actual               *pb.Database
 }
 
 var _ directbase.Adapter = &Adapter{}
 
 func (a *Adapter) Find(ctx context.Context) (bool, error) {
-	log := klog.FromContext(ctx).WithName(ctrlName)
+	log := klog.FromContext(ctx)
```
cheftako, line 111: Question: Is there a reason we are moving away from using WithName()? It actually seems like it would be helpful.

apis/firestore/v1beta1/database_types.go
justinsb: Probably, but it's commented out so I don't think it matters for now.

I've added them commented out here because it is very convenient to "grab" these fields when they are first generated, because after that the main type is no longer generated.  So my workflow for a new type is typically to add it generate.sh, run generate.sh, copy the generated fields into spec/observedstate and then comment out the ones I don't want right away (which are _normally_ few, but in this case we're dealing with terraform decisions)

pkg/controller/direct/firestore/firestoredatabase_controller.go
justinsb: It's already in the context in the logger.

If it isn't, we would add it in the base controller, as that knows the GVK, the controller name, etc etc

pkg/controller/direct/firestore/firestoredatabase_defaults.go
```diff
@@ -18,7 +18,13 @@ import pb "cloud.google.com/go/firestore/apiv1/admin/adminpb"
 
 func ApplyFirestoreDatabaseDefaults(in *pb.Database) {
 	// Set default values to make sure firestore database is "declarative-friendly".
-	in.Type = pb.Database_FIRESTORE_NATIVE
-	in.AppEngineIntegrationMode = pb.Database_DISABLED
-	in.DeleteProtectionState = pb.Database_DELETE_PROTECTION_DISABLED
+	if in.Type == pb.Database_DATABASE_TYPE_UNSPECIFIED {
```
cheftako, line 21: Question: Wondering if this should be a 3-way merge? Spec value takes priority otherwise use the existing GCP value otherwise use this default?

justinsb: I think we would do that with our "ignore" annotation, otherwise this sounds like a new behaviour

cheftako: If we're going to default like this should it at least be marked as the default value in the CRD? This feels like we are determining a default value the we are hiding both from the end user and the service team.

cheftako: Discussed this should probably be expected rather than default. Also would be good as its expected values for it to not be called from the create flow. Happy to merge and fix these in a follow up.

PR #5630, opened 2025-11-14
feat: Improve test coverage for BatchJob 

yuwenma:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/batch/v1alpha1/batchjob/batchjob-network/_generated_object_batchjob-network.golden.yaml
+++ b/pkg/test/resourcefixture/testdata/basic/batch/v1alpha1/batchjob/batchjob-network/_generated_object_batchjob-network.golden.yaml
@@ -35,10 +35,10 @@ spec:
     network:
       networkInterfaces:
       - networkRef:
-          name: batchjob-dep-network-${uniqueId}
+          name: ${networkID}
         noExternalIPAddress: false
         subnetworkRef:
-          name: batchjob-dep-subnetwork-${uniqueId}
+          name: ${subnetworkID}
   location: us-central1
   logsPolicy:
     destination: CLOUD_LOGGING


yuwenma:
Depend on #5623

pkg/test/resourcefixture/testdata/basic/batch/v1alpha1/batchjob/batchjob-network/create.yaml
```diff
@@ -0,0 +1,50 @@
...
+  name: batchjob-network-${uniqueId}
+spec:
+  projectRef:
+    external: ${projectId}
+  location: us-central1
+  priority: 10
+  allocationPolicy:
+    location:
+      allowedLocations:
+        - "zones/us-central1-a"
```
justinsb, line 13: Is the zones prefix required?  I guess we can always relax it over time anyway, so not a blocker...

pkg/test/resourcefixture/testdata/basic/batch/v1alpha1/batchjob/batchjob-network/_http.log
```diff
@@ -0,0 +1,742 @@
...
+      }
+    ],
+    "location": {
+      "allowedLocations": [
+        "zones/us-central1-a"
+      ]
+    },
+    "network": {
+      "networkInterfaces": [
+        {}
```
justinsb, line 343: Randomly found this (was looking for noExternalIPAddresses, because it's a funky field name).  I think this is a bug, we have specified a networkInterfaceRef?

Assuming this is a bug, I'm also not sure how we should find these bugs.  Maybe we should implement export, and look for fields that "go missing"?  Though it's going to be tricky because refs will always be external on export, so it's not going to be a trivial comparison.

pkg/test/resourcefixture/testdata/basic/batch/v1alpha1/batchjob/batchjob-network/create.yaml
yuwenma: I ran the test against real GCP.

pkg/test/resourcefixture/testdata/basic/batch/v1alpha1/batchjob/batchjob-network/_http.log
yuwenma: Just to clarify, do you mean we want to "abstract" the `networkInterface` to a intermediate struct in KCC, in the form of a special reference? Right now, this field itself is not a ref but contains 2 references `ComputeNetworkRef` and `ComputeSubNetworkRef`. So it actually matches the current requirement.  

I took a look at https://docs.cloud.google.com/compute/docs/networking/update-network-interfaces#rest. The networkInterface (vNIC) indeed can be considered as a whole. However, this is still a virtual concept and have no corresponding REST APIs yet. @justinsb Could you share more insights about the suggestion?

yuwenma: I reran the test on HEAD, here's a fix https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5950. (not sure how the log diff can pass presubmit in this PR, but I see what you mean the "bug" is)

PR #5636, opened 2025-11-14
TagsTagBinding: create direct controller

justinsb:
- **TagsTagBinding: create direct controller**

-  **tests: add TagsTagBinding to direct tests**

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagbinding/tagbindingautogen/_final_object.diff
+++ b/pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagbinding/tagbindingautogen/_final_object.diff
@@ -1,0 +1,4 @@
+6d5
+<     cnrm.cloud.google.com/state-into-spec: absent
+27a27
+>   externalRef: tagBindings/%2F%2Fcloudresourcemanager.googleapis.com%2Fprojects%2F${projectNumber}/tagValues/${tagValueID}
--- a/pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagbinding/tagbindingautogen/_final_object_old_controller.golden.yaml
+++ b/pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagbinding/tagbindingautogen/_final_object_old_controller.golden.yaml
@@ -3,6 +3,7 @@ kind: TagsTagBinding
 metadata:
   annotations:
     cnrm.cloud.google.com/management-conflict-prevention-policy: none
+    cnrm.cloud.google.com/state-into-spec: absent
   finalizers:
   - cnrm.cloud.google.com/finalizer
   - cnrm.cloud.google.com/deletion-defender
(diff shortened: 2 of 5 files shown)


apis/tags/v1beta1/tagbinding_identity.go
```diff
@@ -13,3 +13,97 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/apis/common"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/apis/common/identity"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// TagsTagBindingIdentityURL is the format for the externalRef of a TagsTagBinding.
```
cheftako, line 28: NIT: Same as TagsTagBinding. I think its useful to comment some metadata about {{tagBinding}} being both service generated and universal. 

Also wonder if it would be nice to leave a comment link to https://docs.cloud.google.com/asset-inventory/docs/asset-names

justinsb: Good call on links; I do put the link later on (when we parse it) but I think this is also indicative that we could probably rationalize a lot of this.  (Particularly once you see the parsing for TagsLocationTagBinding where we have to parse multiple external forms)

Actually.... this one is different.  The API uses a different from Cloud Asset Inventory

pkg/controller/direct/tags/tagbinding_controller.go
```diff
@@ -0,0 +1,314 @@
...
+
+func (a *TagsTagBindingAdapter) Find(ctx context.Context) (bool, error) {
+	log := klog.FromContext(ctx)
+
+	if a.id == nil {
+		return false, nil
+	}
+
+	fqn := a.id.String()
+	log.V(0).Info("getting TagsTagBinding", "name", fqn)
```
cheftako, line 136: NIT: At v(0) are we concerned about filling the logs, especially in something like a multi-controller reconcile competition?

pkg/controller/direct/tags/tagbinding_controller.go
```diff
@@ -0,0 +1,314 @@
...
+		spec := TagsTagBindingSpec_FromProto(mapCtx, a.actual)
+		if mapCtx.Err() != nil {
+			return nil, mapCtx.Err()
+		}
+		obj.Spec = *spec
+	}
+
+	obj.SetGroupVersionKind(krm.TagsTagBindingGVK)
+	name := a.actual.TagValueNamespacedName
+	name = strings.ReplaceAll(name, "/", "_")
```
cheftako, line 254: COMMENT: Assuming this is reversable because the "namespace" is a camel cased enumeration and the "id" is service generated, so "_" should not occur in either.

justinsb: It's more that we need _a_ name that is semi-unique and a valid k8s object name.

pkg/controller/direct/tags/tagbinding_controller.go
justinsb: Good call, I'll lower.  I wonder if I can have Gemini find instances like this....

(We're still _way_ less verbose than the legacy logging, but this is just noise)

PR #5638, opened 2025-11-15
TagsTagValue: create direct controller

justinsb:
* tests: fix identification of PrimaryResource
* tests: test tagstagkey in direct
* TagsTagValue: create direct controller

### DIFF
--- a/mockgcp/mockresourcemanager/tagbinding.go
+++ b/mockgcp/mockresourcemanager/tagbinding.go
@@ -25,6 +25,8 @@ import (
 	"strings"
 
 	lropb "cloud.google.com/go/longrunning/autogen/longrunningpb"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/emptypb"
 
@@ -46,11 +48,33 @@ func (s *TagBindingsServer) normalizeParent(parent string) (string, error) {
 }
 
 func (s *TagBindingsServer) CreateTagBinding(ctx context.Context, req *pb.CreateTagBindingRequest) (*lropb.Operation, error) {
-	tagValue, err := s.tagValues.GetNamespacedTagValue(ctx, &pb.GetNamespacedTagValueRequest{
-		Name: req.GetTagBinding().GetTagValueNamespacedName(),
-	})
-	if err != nil {
-		return nil, err
+	var tagValue *pb.TagValue
+
+	// For methods that support TagValue namespaced name, only one of
+	// tag_value_namespaced_name or tag_value may be filled. Requests with both
+	// fields will be rejected.
+	if req.GetTagBinding().GetTagValue() != "" && req.GetTagBinding().GetTagValueNamespacedName() != "" {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid request: both TagValue and TagValueNamespacedName are set")
+	}
+
+	if req.GetTagBinding().GetTagValue() != "" {
+		found, err := s.tagValues.GetTagValue(ctx, &pb.GetTagValueRequest{
+			Name: req.GetTagBinding().GetTagValue(),
+		})
+		if err != nil {
+			return nil, err
+		}
+		tagValue = found
+	} else if req.GetTagBinding().GetTagValueNamespacedName() != "" {
+		found, err := s.tagValues.GetNamespacedTagValue(ctx, &pb.GetNamespacedTagValueRequest{
+			Name: req.GetTagBinding().GetTagValueNamespacedName(),
+		})
+		if err != nil {
+			return nil, err
+		}
+		tagValue = found
+	} else {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid request: must specify TagValue or TagValueNamespacedName")
 	}
 
 	obj := proto.Clone(req.TagBinding).(*pb.TagBinding)


apis/tags/v1beta1/tagvalue_identity.go
```diff
@@ -0,0 +1,100 @@
...
+	// TagsTagValueIdentityURL is the format for the externalRef of a TagsTagValue.
+	TagsTagValueIdentityURL = "tagValues/{{tagValue}}"
+)
+
+var _ identity.Identity = &TagsTagValueIdentity{}
+
+// TagsTagValueIdentity represents the identity of a TagsTagValue.
+// +k8s:deepcopy-gen=false
+type TagsTagValueIdentity struct {
+	TagValue string
```
cheftako, line 37: COMMENT?: TAG_VALUE_ID: A globally unique, system-generated numeric identifier.

pkg/controller/direct/tags/diff.go
```diff
@@ -28,6 +32,29 @@ type FieldChange struct {
...
+	diff := &structuredreporting.Diff{}
+
+	var paths []string
+	fields := actual.Type().Descriptor().Fields()
+	for i := 0; i < fields.Len(); i++ {
+		path := string(fields.Get(i).Name())
+		fieldDiff, err := fieldHasChanged(path, desired, actual)
+		if err != nil {
+			log.Error(err, "error determining if field has changed", "field", path)
+			// If we can't determine if the field has changed, include it in the update.
```
cheftako: NIT: This means that fieldHasChanged() must return a valid fieldDiff on error. That appears to be true now. We should probably have a comment on fieldHasChanged() which specifies that as a requirement.

pkg/controller/direct/tags/tagvalue_controller.go
```diff
@@ -0,0 +1,316 @@
...
+	req := &pb.CreateTagValueRequest{
+		TagValue: direct.ProtoClone(a.desired),
+	}
+
+	op, err := a.tagValuesClient.CreateTagValue(ctx, req)
+	if err != nil {
+		return fmt.Errorf("creating TagsTagValue: %w", err)
+	}
+
+	created, err := op.Wait(ctx)
```
cheftako, line 164: FUTURE: It would be good to set a status while we wait. Especially on a resource with a server generated ID. If something happens (like the pod dieing) while we are waiting we will end up with an orphaned resource.

pkg/controller/direct/tags/tagvalue_controller.go
```diff
@@ -0,0 +1,316 @@
...
+	if err != nil {
+		return fmt.Errorf("waiting for creation of TagsTagValue: %w", err)
+	}
+	log.V(0).Info("created TagsTagValue", "name", created.GetName())
+
+	// For compatibility, we set spec.resourceID after creation because this is a server-generated-id resource that we are migrating from terraform/DCL.
+	// More info in docs/ai/server-generated-id.md
+	resourceID := strings.TrimPrefix(created.GetName(), "tagValues/")
+	if err := createOp.SetSpecResourceID(ctx, resourceID); err != nil {
+		return err
```
cheftako, line 174: We need to return an error. However failing to updateStatus means that we *will* orphan an instance of the resource. Would it make sense to updateStatus first (or as well?) as that might mean we do not always orphan the resource?

pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagvalue/tagvalueproject/_generated_object_tagvalueproject.golden.yaml
```diff
@@ -3,7 +3,6 @@ kind: TagsTagValue
 metadata:
   annotations:
     cnrm.cloud.google.com/management-conflict-prevention-policy: none
-    cnrm.cloud.google.com/state-into-spec: absent
```
cheftako, line 6: Do we know why SIS:Absent disappeared from this golden?

pkg/controller/direct/tags/tagvalue_controller.go
justinsb: Opened #5778 to track, I think this is a cross cutting problem

pkg/controller/direct/tags/diff.go
justinsb: Ugh ... good point, that was not intentional

pkg/controller/direct/tags/tagvalue_controller.go
cheftako: It is cross cutting for all the resources in https://docs.cloud.google.com/config-connector/docs/how-to/managing-deleting-resources#service-generated-resource-id. Also worth noting that we should add this resource to that list.

pkg/controller/direct/tags/tagvalue_controller.go
justinsb: Tracking in #5778

pkg/test/resourcefixture/testdata/basic/tags/v1beta1/tagstagvalue/tagvalueproject/_generated_object_tagvalueproject.golden.yaml
justinsb: This is a deliberate design choice.  Direct controllers ignore s-i-s.  One day we will only have direct controllers, and we have previously chosen not to write these annotations.

We do run the tests twice: this is the "main" version which is the direct controller, we also have _final_object_old_controller.yaml which is the TF/DCL version.  (Note the file names also need fixing, but ... hours in the day)

pkg/controller/direct/tags/diff.go
```diff
@@ -28,33 +32,60 @@ type FieldChange struct {
...
+			continue
+		}
+		diff.AddField(fieldDiff.FieldPath, fieldDiff.ActualValue, fieldDiff.DesiredValue)
+		paths = append(paths, fieldDiff.FieldPath)
+	}
+
+	return diff, &fieldmaskpb.FieldMask{Paths: paths}, nil
+}
+
+// fieldHasChanged compares the field at fieldPath in desired and actual messages.
```
justinsb, line 53: I cleaned up the contract here, to hopefully be more obvious.

PR #5640, opened 2025-11-15
chore: skip autogenerated files when generating mappers

justinsb:
This avoids a failure case where the autogenerated files are broken
and cause the mapper generation to fail, because it can't load the
package AST.


dev/tools/controllerbuilder/pkg/gocode/ast.go
```diff
@@ -73,7 +76,23 @@ type StructField struct {
 func LoadPackage(goPackage string, path string) (*Package, error) {
 	fileSet := token.NewFileSet()
 	mode := parser.ParseComments
-	var filter func(fs.FileInfo) bool
+	filter := func(info fs.FileInfo) bool {
+		p := filepath.Join(path, info.Name())
+		// Skip test files (for speed)
+		if strings.HasSuffix(info.Name(), "_test.go") {
+			return false
```
cheftako, line 83: Nit: Seems odd that filter false means to filter. Confirmed that it works that way but ...

dev/tools/controllerbuilder/pkg/gocode/ast.go
```diff
@@ -73,7 +76,23 @@ type StructField struct {
...
+		}
+		// Skip autogenerated files (in case they are broken)
+		b, err := os.ReadFile(p)
+		if err != nil {
+			klog.Warningf("failed to read file %q: %v", p, err)
```
cheftako: Question: If err, I would expect b to be empty.  If b is empty it will not 'Contains()' any string. That implies we will return true. Is that what we want?

dev/tools/controllerbuilder/pkg/gocode/ast.go
```diff
@@ -73,7 +76,23 @@ type StructField struct {
...
+		}
+		if strings.Contains(string(b), "//go:build !ignore_autogenerated") {
+			return false
+		}
+		return true
+	}
+
```
cheftako, line 96: Nit: parser.ParseDir(fileSet, path, filter, mode) is deprecated.

justinsb: Yeah, but the new version is _much_ slower

dev/tools/controllerbuilder/pkg/gocode/ast.go
justinsb: We probably should early-return, but I think we want to visit the file.  We could panic I guess, but we can only true or false, and we are not able to exclude the file.

dev/tools/controllerbuilder/pkg/gocode/ast.go
justinsb: Yeah, we could maybe rename filter to `visitFilePredicate` or something?

justinsb: Switched

dev/tools/controllerbuilder/pkg/gocode/ast.go
justinsb: Switched to panic

dev/tools/controllerbuilder/pkg/gocode/ast.go
justinsb: Added a comment

PR #5649, opened 2025-11-17
chore: clean up prompt annotations

justinsb:
- **chore: clean up prompt annotations**

### DIFF
--- a/mockgcp/mockstorage/folder.go
+++ b/mockgcp/mockstorage/folder.go
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// +tool:mockgcp-support
-// proto.service: google.storage.v1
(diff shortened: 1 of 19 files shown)


mockgcp/mockstorage/folder.go
```diff
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// +tool:mockgcp-support
-// proto.service: google.storage.v1
+// +tool:mockgcp-support-but-no-proto
```
yuwenma, line 15: If it is the case, shall we not add this comment?

justinsb: The idea is that we can create a separate example set of training data.  They don't follow the same patterns as ones with protos, but they probably follow the same patterns as ones without protos!

PR #5650, opened 2025-11-17
mockgcp: initial support for sourcerepo

justinsb:
- **mockgcp: create test for gcloud source repos**
  

- **autogen: golden output for gcloud source repo test**
  

- **mockgcp: initial support for sourcerepo**
  

- **autogen: generated sourcerepo bindings**


mockgcp/mocksourcerepo/service.go
```diff
@@ -0,0 +1,82 @@
+// Copyright 2024 Google LLC
```
yuwenma, line 1: nit: 2025

mockgcp/mocksourcerepo/utils.go
yuwenma, line 1: I think we have a few computeEtag and ProtoClone under different mockgcp/mockservice, maybe it's a good time to use the shared one rather than adding a new file.

justinsb: Yeah, let's do that in follow on.  I had forgotten where we put these!

mockgcp/mocksourcerepo/service.go
justinsb: Will fix in follow on (or hack up the AI to replace the wrong copyright years!)

PR #5655, opened 2025-11-17
mockgcp: create tests for accesscontextmanager levels & policies

justinsb:
- **mockgcp: create tests for accesscontextmanager levels & policies**
- **autogen: golden output for gcloud access-context-manager tests**
- **tests: add support for organization to mockgcp tests**
- **mockgcp: support for accesscontextmanager levels & policies**

### DIFF
--- a/mockgcp/mockaccesscontextmanager/normalize.go
+++ b/mockgcp/mockaccesscontextmanager/normalize.go
@@ -22,6 +22,8 @@ import (
 
 const TimePlaceholder = "2024-04-01T12:34:56.123456Z"
 
+const EtagPlaceholder = "abcdef0123A="
+
 var _ mockgcpregistry.SupportsNormalization = &MockService{}
 
 func (s *MockService) ConfigureVisitor(url string, replacements mockgcpregistry.NormalizingVisitor) {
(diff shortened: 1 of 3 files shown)


justinsb:
/assign @cheftako 

Turning the crank on mockgcp!

mockgcp/mockaccesscontextmanager/accesspolicy.go
```diff
@@ -0,0 +1,201 @@
...
+	lroPrefix := "operations/" + fqn + "/delete/{{operationID}}"
+	return s.operations.DoneLRO(ctx, lroPrefix, metadata, &emptypb.Empty{})
+}
+
+func (s *AccessContextManagerV1) ListAccessPolicies(ctx context.Context, req *pb.ListAccessPoliciesRequest) (*pb.ListAccessPoliciesResponse, error) {
+	if req.GetParent() == "" {
+		return nil, status.Errorf(codes.InvalidArgument, "parent is required")
+	}
+
+	name, err := s.parseAccessPolicyName(req.GetParent() + "/accessPolicies/dummy")
```
yuwenma: This seems to be not right because the `parseAccessPolicyName` requires the token size to be 2, but  req.GetParent() + "/accessPolicies/dummy" will give 4 tokens, like `organizations/<id>/accessPolicies/dummy`

justinsb: Good point - fixed up.  We don't have test coverage information, it might be interesting to add this to our arsenal.  I don't know that we would care too much about the mock being wrong, but maybe we should not implement functions in the mock unless we actually need them.  The AI did this one, because sometimes it is needed by gcloud, but I guess not here.

justinsb: I added test coverage for listing :-)

PR #5658, opened 2025-11-18
fix: don't replace namespaced with tagKeyID/tagValueID

justinsb:
Both of these have URL paths for a search operation that looks
like a GET operation, which we need to special case to avoid
replacing the "namespaced" value with a placeholder.


mockgcp/mockresourcemanager/normalize.go
yuwenma, line 1: This change makes sense, but I saw #5656 also has the _http.log fixed without changing the normalize.go (and the presubmit passes). Is there a duplicated normalization?

justinsb: Let's go with that one then

PR #5660, opened 2025-11-18
chore: add caching to the project mapper

justinsb:
This both cuts down on API calls, but it also speeds up
real-world operation.

### DIFF
--- a/dev/tools/controllerbuilder/pkg/gocode/ast.go
+++ b/dev/tools/controllerbuilder/pkg/gocode/ast.go
@@ -322,6 +322,21 @@ func toGoType(t ast.Expr) (string, error) {
 		}
 		return "map[" + k + "]" + v, nil
 
+	case *ast.IndexListExpr:
+		s, err := toGoType(t.X)
+		if err != nil {
+			return "", err
+		}
+		var indices []string
+		for _, index := range t.Indices {
+			ixStr, err := toGoType(index)
+			if err != nil {
+				return "", err
+			}
+			indices = append(indices, ixStr)
+		}
+		return s + "[" + strings.Join(indices, ", ") + "]", nil
+
 	default:
 		return "", fmt.Errorf("unhandled field type %T, %+v", t, t)
 	}


pkg/test/resourcefixture/testdata/basic/vertexai/v1beta1/vertexaimetadatastore/metadatastorebasic/_http.log
```diff
@@ -141,83 +141,6 @@ X-Xss-Protection: 0
...
-  "createTime": "2024-04-01T12:34:56.123456Z",
-  "etag": "abcdef0123A=",
-  "name": "projects/${projectNumber}",
-  "projectId": "${projectId}",
-  "state": 1
-}
-
----
-
-GET https://us-central1-aiplatform.googleapis.com/v1beta1/projects/${projectNumber}/locations/us-central1/metadataStores/metadatastore${uniqueId}?%24alt=json%3Benum-encoding%3Dint
```
justinsb, line 169: I think this request is probably still happening, just that our golden tests are eliding repeated requests, and there's no longer an intermediate project request

apis/common/projects/mapper.go
```diff
@@ -56,18 +87,95 @@ func (m *ProjectMapper) LookupProjectNumber(ctx context.Context, projectID strin
...
+
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	info := c.projectsByID[projectID]
+	if info != nil && time.Since(info.timestamp) > c.expiration {
+		info = nil
+	}
+	if info == nil {
+		projectInfo, err := c.getProject(ctx, "projects/"+projectID)
```
cheftako: Not a fan of making RPC calls while holding locks. I think its ok to break this into a couple of locked blocks with the getProject between.

apis/common/projects/mapper.go
```diff
@@ -56,18 +87,95 @@ func (m *ProjectMapper) LookupProjectNumber(ctx context.Context, projectID strin
...
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	info := c.projectsByNumber[projectNumber]
+	if info != nil && time.Since(info.timestamp) > c.expiration {
+		info = nil
+	}
+	if info == nil {
+		name := "projects/" + strconv.FormatInt(projectNumber, 10)
+		projectInfo, err := c.getProject(ctx, name)
```
cheftako: Ditto

pkg/concurrent/collections.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// Set sets the value for the given key in the map.
+func (m *ExpiringMap[K, V]) Set(k K, v V) {
+	m.inner.Set(k, v)
+}
+
+// ConcurrentMap is a thread-safe map with lazy computation capabilities.
+type ConcurrentMap[K comparable, V any] struct {
+	mutex sync.Mutex
```
cheftako, line 55: Nit: optional's lock is safe from being copied because you've made it package private (I too can troll) but this lock is not safe from being copied.

pkg/concurrent/collections.go
```diff
@@ -0,0 +1,158 @@
...
+// GetOrCompute retrieves the value for the given key from the map,
+// or computes it using the provided function if not present.
+func (m *ConcurrentMap[K, V]) GetOrCompute(key K, compute func() (V, error)) (V, error) {
+	m.mutex.Lock()
+	value := m.data[key]
+	if value == nil {
+		value = newEmptyOptional[V]()
+		m.data[key] = value
+	}
+	m.mutex.Unlock()
```
cheftako, line 135: Nit: highly unlikely but I think it is possible for us to panic before we get to this unlock. May not care as we are probably about to crash in that case.

pkg/concurrent/collections.go
```diff
@@ -0,0 +1,158 @@
...
+// getOrCompute retrieves the value if set, or computes it using the provided function.
+func (ov *optional[V]) getOrCompute(compute func() (V, error)) (V, error) {
+	ov.mutex.Lock()
+	defer ov.mutex.Unlock()
+
+	if ov.valueIsSet {
+		return ov.value, nil
+	}
+
+	value, err := compute()
```
cheftako, line 149: Nit: We are down the a very local lock to hold while making a RPC call but we are still holding a lock while making a RPC call. If we really want a single initialization to occur Once() still has the problem of pausing all the callers but its a little clearer what is happening. Otherwise we may be better of letting multiple goroutines call compute(). If 1 gets stuck on a 99+% latency call the rest get to continue. We expect all of them to return the same value but we can validate that when we attempt the set.

cheftako: Actually not convinced that Once is a good idea with a RPC in case of error. Still not sure I'm comfortable with holding a lock and performing a RPC.

justinsb: I'll add a timeout to the RPC call.  We want to avoid concurrent requests, and I think that means a lock

pkg/concurrent/collections.go
justinsb: I think because we're not calling out to other code we should accept this risk.  I don't know of a way to write this in go without introducing another function so it will be quite hard to read.

pkg/concurrent/collections.go
justinsb: True, but mutex embeds noCopy and so `go vet` will complain bitterly if you try.  Ask me how I know :-)

PR #5662, opened 2025-11-18
feat: API, Fuzzer, Mapper for AssuredWorkLoadsWorkload

yuwenma:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/controller/direct/assuredworkloads/workload_fuzzer.go
+++ b/pkg/controller/direct/assuredworkloads/workload_fuzzer.go
@@ -33,14 +33,14 @@ func assuredWorkloadsWorkloadFuzzer() fuzztesting.KRMFuzzer {
 		AssuredWorkloadsWorkloadObservedState_FromProto, AssuredWorkloadsWorkloadObservedState_ToProto,
 	)
 
-	f.SpecFields.Insert(".billing_account")
-	f.SpecFields.Insert(".resource_settings")
+	f.SpecField(".billing_account")
+	f.SpecField(".resource_settings")
 
-	f.StatusFields.Insert(".create_time")
-	f.StatusFields.Insert(".resources")
-	f.StatusFields.Insert(".kaj_enrollment_state")
-	f.StatusFields.Insert(".saa_enrollment_response")
-	f.StatusFields.Insert(".compliant_but_disallowed_services")
+	f.StatusField(".create_time")
+	f.StatusField(".resources")
+	f.StatusField(".kaj_enrollment_state")
+	f.StatusField(".saa_enrollment_response")
+	f.StatusField(".compliant_but_disallowed_services")
 	f.Unimplemented_NotYetTriaged(".provisioned_resources_parent")
 	f.Unimplemented_NotYetTriaged(".kms_settings")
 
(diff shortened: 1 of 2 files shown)


pkg/controller/direct/assuredworkloads/workload_mapper.go
```diff
@@ -0,0 +1,55 @@
...
+	out := &krm.Workload_SaaEnrollmentResponse{}
+	out.SetupErrors = []string{}
+	for _, e := range in.SetupErrors {
+		out.SetupErrors = append(out.SetupErrors, e.String())
+	}
+	out.SetupStatus = direct.LazyPtr(in.SetupStatus.String())
+	return out
+}
+
+func Workload_SaaEnrollmentResponse_ToProto(mapCtx *direct.MapContext, in *krm.Workload_SaaEnrollmentResponse) *pb.Workload_SaaEnrollmentResponse {
```
yuwenma, line 43: This manual mapper (and above) is because the proto defines a typed  int32 variable and customized the "String" function.

pkg/controller/direct/assuredworkloads/workload_fuzzer.go
```diff
@@ -0,0 +1,52 @@
...
+	fuzztesting.RegisterKRMFuzzer(assuredWorkloadsWorkloadFuzzer())
+}
+
+func assuredWorkloadsWorkloadFuzzer() fuzztesting.KRMFuzzer {
+	f := fuzztesting.NewKRMTypedFuzzer(&pb.Workload{},
+		AssuredWorkloadsWorkloadSpec_FromProto, AssuredWorkloadsWorkloadSpec_ToProto,
+		AssuredWorkloadsWorkloadObservedState_FromProto, AssuredWorkloadsWorkloadObservedState_ToProto,
+	)
+
+	f.SpecFields.Insert(".billing_account")
```
justinsb: Nit: we should use `f.SpecField(` etc

pkg/controller/direct/assuredworkloads/workload_mapper.go
```diff
@@ -0,0 +1,55 @@
...
+	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/assuredworkloads/v1alpha1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
+)
+
+func Workload_SaaEnrollmentResponse_FromProto(mapCtx *direct.MapContext, in *pb.Workload_SaaEnrollmentResponse) *krm.Workload_SaaEnrollmentResponse {
+	if in == nil {
+		return nil
+	}
+	out := &krm.Workload_SaaEnrollmentResponse{}
+	out.SetupErrors = []string{}
```
justinsb: You probably don't want to do this, you probably want to ~make an array of the same size~ explicitly check nil, otherwise `nil` and `[]` both map to `[]`

pkg/controller/direct/assuredworkloads/workload_mapper.go
justinsb: Not sure what you mean - it looks like a normal enum

yuwenma: Yes. Just a marker on why we cannot use the auto mapper function

yuwenma (commented):
This is ready for another look @justinsb  🙇

PR #5664, opened 2025-11-18
feat: API, Fuzzer, Mapper for CertificateManagerCertificateIssuanceConfig

yuwenma:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


apis/certificatemanager/v1alpha1/dnsauthorization_types.go
yuwenma, line 1: This resource is already in Beta, but the alpha is not removed. Clean up the code and use `internal.cloud.google.com/additional-versions=v1alpha1` in beta to keep the alpha CRD the same as the beta one.

apis/certificatemanager/v1beta1/dnsauthorization_types.go
```diff
@@ -58,7 +58,7 @@ type CertificateManagerDNSAuthorizationStatus struct {
 	to DNS configuration for the authorization to be usable by
 	certificate. */
 	// +optional
-	DnsResourceRecord []DnsAuthorization_DnsResourceRecord `json:"dnsResourceRecord,omitempty"`
+	DnsResourceRecord []DNSAuthorization_DNSResourceRecordObservedState `json:"dnsResourceRecord,omitempty"`
```
yuwenma: Because this field is in `.status` directly rather than `.status.observedState`, the manual mapper has to exist to fix the mapping (the mapper itself has bugs, see fix below).

apis/certificatemanager/v1alpha1/dnsauthorization_types.go
justinsb: Can you split into another PR?

pkg/controller/direct/certificatemanager/dnsauthorization_mappings.go
```diff
@@ -20,9 +20,14 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
 )
 
-func CertificateManagerDNSAuthorizationStatusObservedState_FromProto(mapCtx *direct.MapContext, created *certificatemanagerpb.DnsAuthorization) *krm.CertificateManagerDNSAuthorizationStatus {
-	status := &krm.CertificateManagerDNSAuthorizationStatus{
-		DnsResourceRecord: []krm.DnsAuthorization_DnsResourceRecord{direct.ValueOf(DnsAuthorization_DnsResourceRecord_FromProto(mapCtx, created.GetDnsResourceRecord()))},
+func CertificateManagerDNSAuthorizationStatusObservedState_FromProto(mapCtx *direct.MapContext, in *certificatemanagerpb.DnsAuthorization) *krm.CertificateManagerDNSAuthorizationStatus {
```
justinsb: You _should_ be able to add a `// +kcc:status:proto=google.cloud...` annotation and it _should_ come closer.  That said I normally just map these fields manually in an updateStatus function.

justinsb (commented):
Looks good, except that dnsauthorization is mixed into this PR

apis/certificatemanager/v1alpha1/dnsauthorization_types.go
yuwenma: Yes, I should. #5689 I discovered more and more problems in the other resource while developing the original one.

pkg/controller/direct/certificatemanager/dnsauthorization_mappings.go
yuwenma: I like that! The only thing is that I also need to change the `[]DNSAuthorization_DNSResourceRecordObservedState` to `[]*DNSAuthorization_DNSResourceRecordObservedState`, which I think is fine.

PR #5672, opened 2025-11-18
tests: prefer using directory instead of Name for tests

justinsb:
The full path is much less fragile (less chance of collisions),
and is much more descriptive (being the full path).


justinsb:
/assign @cheftako 

Setting us up for #5674

PR #5673, opened 2025-11-18
chore: don't allow two controllers to register for the same kind

justinsb:


pkg/controller/direct/registry/registry.go
```diff
@@ -89,6 +89,9 @@ func RegisterModel(gvk schema.GroupVersionKind, modelFn ModelFactoryFunc) {
 	if singleton.registrations == nil {
 		singleton.registrations = make(map[schema.GroupKind]*registration)
 	}
+	if singleton.registrations[gvk.GroupKind()] != nil {
+		klog.Fatalf("Model for %s already registered", gvk.GroupKind())
```
cheftako, line 93: Nit: Would be nice if modelFn would better identify what it was. We could both ensure it wasn't the same function registered twice. If they are different then it might better indicate where from.

PR #5700, opened 2025-11-21
PubsubLiteSubscription: CRD/Mapper/Fuzzer

justinsb:
- **PubsubLiteSubscription: CRD/Mapper/Fuzzer**
  

- **autogen: PubsubLiteSubscription CRD/Mapper/Fuzzer**
  

- **controllerbuilder: limit number of examples we send to 32**
  This avoids exceeding token limits when generating prompts for the LLM.
  
- **tests: flag more references, based on their descriptions**


apis/pubsublite/v1alpha1/subscription_types.go
```diff
@@ -0,0 +1,133 @@
...
+	// Required. Defines the parent path of the resource.
+	// +required
+	ProjectRef *parent.ProjectRef `json:"projectRef,omitempty"`
+
+	// The region of the pubsub lite topic.
+	Region *string `json:"region,omitempty"`
+
+	// Immutable. A reference to a Topic resource.
+	// +required
+	Topic *string `json:"topic,omitempty"`
```
yuwenma, line 42: This should be a reference. do you want to try the @docs/ai/add-resource-references.md? I used that to detect ref field and add the _reference.go files  now

apis/pubsublite/v1alpha1/subscription_types.go
```diff
@@ -0,0 +1,133 @@
...
+
+	// The zone of the pubsub lite topic.
+	// +required
+	Zone *string `json:"zone,omitempty"`
```
yuwenma, line 46: If Topic is a reference, I think we don't need this zone.

dev/tools/controllerbuilder/pkg/toolbot/csv.go
yuwenma, line 1: shall this file change go with the PR?  If so, do you mind pointing me to how it is used  (just for self learning) ?

justinsb: This is the prompt tool that we've been using e.g. `cat prompt | controllerbuilder prompt --model gemini-3-pro-preview > apis/appengine/v1alpha1/firewallrule_types.go`

I included it here because I needed it to generate this PR, and often it's handy to see the output alongside the change, but in the output is not as revealing (other than that it is pretty good!)

apis/pubsublite/v1alpha1/subscription_types.go
justinsb: It's an existing CRD (albeit alpha), so step 1 is to bring it to direct.

apis/pubsublite/v1alpha1/subscription_types.go
justinsb: Yeah, I'm hoping we can make this much better once it is direct.

apis/pubsublite/v1alpha1/subscription_types.go
yuwenma: ok, then can we improve the API check to catch this as an exception ( in the "missing_reference" list)? Otherwise, we are highly likely to miss this case since the follow-up PRs won't touch this code if in happy path.

justinsb:
/assign @cheftako 

This was built with LLMs

PR #5704, opened 2025-11-21
chore: improve type and fuzzer instructions

yuwenma:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


docs/develop-resources/scenarios/new-resource.md
```diff
@@ -6,14 +6,19 @@ This guide outlines the structured, three-PR process for adding a new Config Con
...
 
-### 1. Generate Initial API, CRD, and Mapper
+### 1. Generate Initial CRD, Mapper and Fuzzer
 
 This is the first step to create the basic building blocks from the GCP API definition.
 
 *   **Gemini Prompt:**
     ```bash
-    "Add generate.sh for apis/<service>/v1alpha1, similar to the example in apis/bigquerybiglake/v1alpha1/generate.sh, then run the new generate.sh."
+    "Add generate.sh for apis/<service>/v1alpha1, look at examples in apis/bigquerybiglake/v1alpha1/generate.sh, then run the generate.sh."
```
justinsb, line 15: I like the more direct instructions

dev/tools/controllerbuilder/template/apis/types.go
```diff
@@ -33,15 +33,14 @@ package {{ .Version }}
...
-	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/apis/k8s/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 var {{ .Kind }}GVK = GroupVersion.WithKind("{{ .Kind }}")
 
 // {{ .Kind }}Spec defines the desired state of {{ .Kind }}
 {{- if .KindProtoTag }}
-// +kcc:spec:proto={{ .KindProtoTag }}
+// TODO: Move kcc:spec:proto={{ .KindProtoTag }} content to Spec, and modify the reference field.  Once done, add +kcc:spec:proto={{ .KindProtoTag }} and rerun generate.sh
```
justinsb, line 43: Are we doing this so that we keep generating the types, so we can copy them for longer?

I was thinking about this, it is surprising to the LLMs (and maybe to people) that the types go away.  Maybe it would be better if we generated the types but commented them out, or generated them into a separate file.  Personally I think from the point of view of stable git diffs, we should generate them into the same file but comment them out (with /* ... */ style comments, to avoid changing indentation)

pkg/controller/direct/bigquerybiglake/table_fuzzer.go
```diff
@@ -33,15 +33,15 @@ func bigLakeTableFuzzer() fuzztesting.KRMFuzzer {
 		BigLakeTableObservedState_v1beta1_FromProto, BigLakeTableObservedState_v1beta1_ToProto,
 	)
 
-	f.SpecFields.Insert(".type")
-	f.SpecFields.Insert(".hive_options")
+	f.SpecField(".type")
```
justinsb, line 36: These changes are good but they don't seem to have any relation to the markdown changes?

One thing I've been experimenting with is creating an Epic and subissues and having gemini tackle it unattended.  In this case the Epic would be "Fuzzers should use helpers e.g. SpecField instead of SpecFields.Insert" and then the subissues would probably be per-package (?).  It breaks the work down into sensible chunks that gemini can tackle, and that is easy to review.

dev/tools/controllerbuilder/template/apis/types.go
cheftako: +1

PR #5708, opened 2025-11-21
chore: apply ref and identity to dataplex

yuwenma:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/dataplex/v1alpha1/dataplextask/dataplextask-minimal/_http.log
+++ b/pkg/test/resourcefixture/testdata/basic/dataplex/v1alpha1/dataplextask/dataplextask-minimal/_http.log
@@ -14,7 +14,11 @@ GRPC /google.cloud.dataplex.v1.DataplexService/CreateLake
 
 {
   "lake": {
-    "description": "Initial description"
+    "description": "Initial description",
+    "labels": {
+      "cnrm-test": "true",
+      "managed-by-cnrm": "true"
+    }
   },
   "lakeId": "dataplexlake-${uniqueId}",
   "parent": "projects/${projectId}/locations/us-central1"
@@ -57,6 +61,10 @@ OK
     "assetStatus": {},
     "createTime": "2024-04-01T12:34:56.123456Z",
     "description": "Initial description",
+    "labels": {
+      "cnrm-test": "true",
+      "managed-by-cnrm": "true"
+    },
     "metastore": {},
     "metastoreStatus": {
       "state": "NONE",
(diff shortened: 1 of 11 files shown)


apis/dataplex/v1alpha1/entrygroup_identity.go
```diff
@@ -39,80 +46,57 @@ func (i *EntryGroupIdentity) ID() string {
...
+func (i *EntryGroupIdentity) Parent() *parent.ProjectAndLocationParent {
 	return i.parent
 }
 
-// No changes needed as EntryGroupParent fields and resource pattern are correct.
-type EntryGroupParent struct {
-	ProjectID string
-	Location  string
+func (i *EntryGroupIdentity) FromExternal(ref string) error {
+	tokens := strings.Split(ref, "/entryGroups/")
```
justinsb, line 54: Nit: I have a PopURLComponent function (or something like that) that pops off the last two components of the URL, verifying that the "key" is what we expect.  It'll land in some random PR, but I think I like that pattern so it's worth taking a look at.  We want to seed the AI with good examples, what you have here is definitely better so not a blocker

apis/dataplex/v1alpha1/entrygroup_types.go
```diff
@@ -15,26 +15,21 @@
...
-	ProjectRef *refv1beta1.ProjectRef `json:"projectRef"`
-	// +required
-	Location string `json:"location"`
-}
-
 // DataplexEntryGroupSpec defines the desired state of DataplexEntryGroup
 // +kcc:spec:proto=google.cloud.dataplex.v1.EntryGroup
 type DataplexEntryGroupSpec struct {
-	DataplexEntryGroupParent `json:",inline"`
+	ParentRef *parent.ProjectAndLocationRef `json:",inline"`
```
justinsb, line 31: I would still argue that it's simpler just to put the fields in here directly, but ... not a blocker

pkg/controller/direct/dataplex/entrytype_controller.go
```diff
@@ -60,14 +63,30 @@ func (m *entryTypeModel) AdapterForObject(ctx context.Context, reader client.Rea
...
+	if mapCtx.Err() != nil {
+		return nil, mapCtx.Err()
+	}
+	desired.Labels = label.NewGCPLabelsFromK8sLabels(u.GetLabels())
+
+	idI, err := obj.GetIdentity(ctx, reader)
 	if err != nil {
 		return nil, err
 	}
+	id, ok := idI.(*krm.EntryTypeIdentity)
```
justinsb, line 82: Yeah, this is annoying.  I haven't figured out a nice idiomatic way with the same of generics being what it is.  I typically just do an unchecked cast (`id.(*krm.EntryTypeIdentity)`) - if it's wrong it'll blow up in testing.

pkg/test/resourcefixture/testdata/basic/dataplex/v1alpha1/dataplextask/dataplextask-minimal/_http.log
```diff
@@ -406,6 +442,10 @@ OK
   },
   "createTime": "2024-04-01T12:34:56.123456Z",
   "description": "Initial description",
+  "labels": {
```
justinsb, line 445: It feels like a bug that we were able to merge controllers with "broken" labels handling.

PR #5712, opened 2025-11-24
FirestoreDatabase: fix defaulting

justinsb:
We want to simulate the server-defaulting logic to avoid spurious diffs,
not to send values the users did not specify.

### DIFF
--- a/pkg/controller/direct/firestore/firestoredatabase_controller.go
+++ b/pkg/controller/direct/firestore/firestoredatabase_controller.go
@@ -26,6 +26,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/structuredreporting"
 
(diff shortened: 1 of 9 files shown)


pkg/controller/direct/firestore/firestoredatabase_controller.go
```diff
@@ -190,25 +188,35 @@ func (a *Adapter) Update(ctx context.Context, updateOp *directbase.UpdateOperati
 	}
 	desired.Name = fqn
 
-	// Apply default values.
-	ApplyFirestoreDatabaseDefaults(desired)
+	// Simulate server-side defaulting, so we don't issue updates for fields that
+	// are different only because of defaulting.
+	expected := direct.ProtoClone(desired)
```
yuwenma, line 203: Awesome! I forgot we have the ProneClone. I was using `proto.Clone` directly but hit a panic when the proto contains unexported fields.

pkg/controller/direct/firestore/firestoredatabase_controller.go
```diff
@@ -190,25 +188,35 @@ func (a *Adapter) Update(ctx context.Context, updateOp *directbase.UpdateOperati
...
+	ApplyServerSideDefaults(expected)
```
yuwenma, line 204: How about `proto.Merge`?

justinsb: I don't think it does the same thing

PR #5717, opened 2025-11-25
feat: Simplify cookie usage for direct controllers (RunJob as example)

yuwenma:
### Why we want this change:
The current cookie logic is spread in controller Create/Update, confused Gemini for different scenarios. This logic is less error prone and is much easier to use because it abstracts the cookie part to its own library.  

This PR also fixes 2 issues in RunJob. one is the missing externalRef in update, the other is the lastModifiedCookie is not a md hash (but a plain struct)

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/controller/direct/common/hash.go
+++ b/pkg/controller/direct/common/hash.go
@@ -16,24 +16,27 @@
 package common
 
 import (
+	"crypto/md5"
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
 
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
 	"google.golang.org/genproto/googleapis/api/annotations"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/reflect/protoreflect"
 	"google.golang.org/protobuf/types/known/fieldmaskpb"
+	"k8s.io/klog/v2"
 )
 
-// HashProto calculates a hash of a proto message.
+// hashProto calculates a hash of a proto message.
 // We use this to detect changes to the GCP resource.
-func HashProto(obj proto.Message) (string, error) {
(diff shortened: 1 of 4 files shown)


pkg/controller/direct/common/hash.go
```diff
@@ -114,23 +129,22 @@ type Cookie struct {
...
+	return fmt.Sprintf("%x", md5.Sum(b))
 }
 
-// ParseCookie parses a cookie string.
-func ParseCookie(s string) (*Cookie, error) {
-	cookie := &Cookie{}
-	if err := json.Unmarshal([]byte(s), cookie); err != nil {
-		return nil, fmt.Errorf("error unmarshalling cookie: %w", err)
+func (c *Cookie) Equal(lastModifiedCookie *string) bool {
+	if lastModifiedCookie == nil {
```
justinsb, line 142: I believe `direct.ValueOf(lastModifiedCookie)` returns "" when lastModifiedCookie is nil

pkg/controller/direct/run/job_controller.go
```diff
@@ -208,39 +201,21 @@ func (a *JobAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOper
...
-				return updateOp.UpdateStatus(ctx, status, nil)
-			}
-		}
+	if currentCookie.Equal(a.lastModifiedCookie) {
+		log.V(2).Info("resource is up to date", "name", a.id)
+		return a.updateStatus(ctx, a.actual, updateOp)
 	}
 
-	a.desired.Name = a.actual.Name
+	// We need to set the name for the update request, but we don't want to modify a.desired
```
cheftako, line 213: We should be generating a structured diff and reporting it when we decide we need to do an actual update.

justinsb: Good call!

PR #5729, opened 2025-12-01
feat: support map type with key string and value message

### LINKED ISSUE #6944: tool: support generating map type with key string and value message
### Checklist

The existing GoTypeForField code only supported map types where the value was a primitive type (string or int64).

### Additional information

_No response_

### Importance

_No response_

SotaSato-stst:
### BRIEF Change description

**Enable support for generating Go types for `map<string, message>` fields in Protocol Buffer definitions.**

This change modifies the `GoTypeForField` function within the `codegen` package to correctly map Proto map fields where the key is a `string` and the value is a `message` type (i.e., another nested struct), converting them to `map[string]MyStructName` in Go.

The existing `GoTypeForField` code only supported map types where the value was a primitive type (`string` or `int64`). 
This modification is necessary to allow the code generator to successfully process a wider range of standard Proto definitions, specifically those involving complex map structures.

The core change is in the `GoTypeForField` function. When detecting `protoreflect.MessageKind` as the map value, it now correctly determines the corresponding Go struct name (including the `ObservedState` suffix if `isTransitiveOutput` is true) and formats the output as `map[string]StructName`.

### Tests you have done

### Example Generated Output
```go
// Map from column family name to the columns in this family to be included
//  in the AuthorizedView.
// +kcc:proto:field=google.bigtable.admin.v2.AuthorizedView.SubsetView.family_subsets
FamilySubsets map[string]*AuthorizedView_FamilySubsets `json:"familySubsets,omitempty"`

### CODE REVIEW

**Comment by SotaSato-stst** [comment]
/assign @acpana @yuwenma
It's ready for review. I'm not sure who the ideal reviewer is for this change, but cloud you review this PR?

**Comment by SotaSato-stst** [comment]
/assign @gemmahou @cheftako 
Could you review this PR?
We have forked Config Connector for internal use, and we would like to contribute our changes back to the upstream repository.

File: dev/tools/controllerbuilder/pkg/codegen/typegenerator.go [hunk 1]
```diff
@@ -243,13 +243,28 @@ func GoTypeForField(field protoreflect.FieldDescriptor, isTransitiveOutput bool)
...
 			return "map[string]int64", nil
+		} else if valueKind == protoreflect.MessageKind {
+			var valueGoType string
+			if isTransitiveOutput {
+				valueGoType = goNameForOutputProtoMessage(valueField.Message())
+			} else {
+				valueGoType = GoNameForProtoMessage(valueField.Message())
+			}
+
+			return fmt.Sprintf("map[string]*%s", valueGoType), nil
```
gemmahou: The code unconditionally adds a pointer (`*`) to the value type.
For `google.protobuf.Struct`, `GoNameForProtoMessage` returns `apiextensionsv1.JSON`.
Existing code (lines 295-297 in the original file) explicitly removes the pointer for `apiextensionsv1.JSON` to produce `apiextensionsv1.JSON` instead of `*apiextensionsv1.JSON`.
This change will generate `map[string]*apiextensionsv1.JSON`. Please verify if the pointer is intended here, or if we should apply the same special casing to avoid it.

+1

gemmahou: Similar to the above, for `google.protobuf.Timestamp` which maps to `string` (via `protoMessagesNotMappedToGoStruct`), this will generate `map[string]*string`. While using a pointer for nullable fields in a map is generally correct for proto semantics, please confirm this is the desired Go type for these specific KRM mappings.

While having `google.protobuf.Timestamp` in map is very rare, the value should not be a pointer as the comment above.

dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
```diff
@@ -243,13 +243,28 @@ func GoTypeForField(field protoreflect.FieldDescriptor, isTransitiveOutput bool)
...
+		}
+
+		if valueKind == protoreflect.StringKind {
 			return "map[string]string", nil
-		} else if keyKind == protoreflect.StringKind && valueKind == protoreflect.Int64Kind {
+		} else if valueKind == protoreflect.Int64Kind {
 			return "map[string]int64", nil
+		} else if valueKind == protoreflect.MessageKind {
+			var valueGoType string
+			if isTransitiveOutput {
```
gemmahou: nit: Consider using a `switch` statement for `valueKind` to improve readability and extensibility, especially if more types need to be supported in the future (e.g., `BytesKind`).

gemmahou: nit: Since this logic for determining the Go type name (lines 260-264) is very similar to the logic for non-map fields below (lines 272-276), consider extracting it into a helper function to avoid code duplication.

dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
SotaSato-stst: I originally used pointers to reduce copying overhead, but I wasn't aware of the KRM convention. I'll update the code to remove the pointers for these types to follow the standard pattern.

SotaSato-stst:
@gemmahou
Thank you for the quick review. I'm sorry for the delay in getting these fixes back to you. Could you please take another look?

apis/gkehub/v1beta1/types.generated.go
```diff
@@ -216,8 +216,11 @@ type MonitoringConfig struct {
 
 // +kcc:proto=google.cloud.gkehub.policycontroller.v1beta.PolicyContentSpec
 type PolicyContentSpec struct {
-
-	// TODO: unsupported map type with key string and value message
+	// map of bundle name to BundleInstallSpec. The bundle name maps to the
+	//  `bundleName` key in the `policycontroller.gke.io/constraintData` annotation
+	//  on a constraint.
+	// +kcc:proto:field=google.cloud.gkehub.policycontroller.v1beta.PolicyContentSpec.bundles
+	Bundles map[string]BundleInstallSpec `json:"bundles,omitempty"`
```
cheftako, line 223: Why isn't this creating a 'bundles' field at the same level as 'templateLibrary` in https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/v1.145.0/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_gkehubfeaturememberships.gkehub.cnrm.cloud.google.com.yaml#L640 ?

gemmahou:
Closing this one as it's out-dated. Let's followup on issue #6944.

PR #5739, opened 2025-12-03
chore: Remove repeated getIamPolicy requests in golden logs

gemmahou:
### BRIEF Change description

Remove repeated getIamPolicy requests in golden logs.

Our current test logic removes GET requests from the logs. However, the `getIamPolicy` requests use an unusual pattern: `POST https://cloudresourcemanager.googleapis.com/v1/projects/${projectId}:getIamPolicy?alt=json&prettyPrint=false`, so they are not being filtered from the golden logs, making log comparison difficult.

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


tests/e2e/httplog.go
```diff
@@ -62,6 +62,11 @@ func RemoveExtraEvents(events test.LogEntries) test.LogEntries {
 					return true
 				}
 			}
+			// Remove repeated getIamPolicy requests, the request looks like
+			// POST https://cloudresourcemanager.googleapis.com/v1/projects/${projectId}:getIamPolicy?alt=json&prettyPrint=false
+			if r.Method == "POST" {
+				return strings.Contains(r.URL, "getIamPolicy")
+			}
```
acpana, line 69: /hold

I think this is a "side effect"/ sympton of using the current hand rolled controllers. I believe that if we fully move over to the direct way of doing IAM, these repeated GETs in POSTs 🙃 would get removed by themselves. I wouldn't mind being overruled here but I'd say let's keep these for now! I

gemmahou: Thanks for the clarification! I'll keep this PR opened(on hold) and check back once the IAM migration is done!

PR #5740, opened 2025-12-03
feat(operator): Add metadataHost field for custom metadata server configuration

### LINKED ISSUE #5741: Config Connector will fail on IPv6-only GKE clusters due to hardcoded metadata IP
## Describe the bug

Config Connector will fail to start on IPv6-only GKE clusters because the controller attempts to reach the GCP metadata server at the hardcoded IPv4 address `169.254.169.254`, which is unreachable from pods with only IPv6 addresses.

## Expected behavior

Config Connector should provide a way to configure the metadata server address, allowing it to work in IPv6-only environments and other scenarios requiring custom metadata endpoints.

## Environment

- Kubernetes cluster type: GKE (IPv6-only, when available)
- Config Connector version: latest

## Anticipated error

When IPv6-only GKE clusters become available, Config Connector pods will fail with:

```
dial tcp 169.254.169.254:80: connect: network is unreachable
```

## Additional context

The GCP client libraries already support the `GCE_METADATA_HOST` environment variable to override the metadata server address. This allows using:
- `metadata.google.internal` (DNS-based, works on both IPv4 and IPv6)
- `fd20:ce::254` (GCP's IPv6 metadata address)

### Use cases beyond IPv6

This configuration option would also benefit:
- **Metadata proxy configurations**: Organizations using proxies for security auditing, credential scoping, or SSRF protection (e.g., [k8s-metadata-proxy](https://github.com/GoogleCloudPlatform/k8s-metadata-proxy))
- **Custom credential vending**: Integration with tools like HashiCorp Vault
- **Dual-stack clusters**: Environments preferring IPv6 connectivity

### Proposed solution

Add a `spec.metadataHost` field to the ConfigConnector CRD that, when set, injects the `GCE_METADATA_HOST` environment variable into controller containers.

```yaml
kind: ConfigConnector
metadata:
spec:
  mode: cluster
  googleServiceAccount: "..."
  metadataHost: "metadata.google.internal"  # Proposed new field
```

GraceAtwood:
Fixes #5741

## Summary

This PR adds a new `spec.metadataHost` field to the ControllerResource CRD that allows operators to specify a custom GCP metadata server hostname for Config Connector controllers.

## Problem

The GCP metadata server is currently accessed via the hardcoded IPv4 link-local address `169.254.169.254`. This will cause issues in IPv6-only GKE clusters where pods only have IPv6 addresses and cannot reach IPv4 endpoints:

```
dial tcp 169.254.169.254:80: connect: network is unreachable
```

Additionally, there are legitimate use cases for overriding the metadata endpoint beyond IPv6 support.

## Solution

When `spec.metadataHost` is set on a ControllerResource CR, the operator injects the `GCE_METADATA_HOST` environment variable into the controller's containers. The underlying GCP client libraries (`golang.org/x/oauth2/google` and `cloud.google.com/go/compute/metadata`) already support this env var.

The field is placed on ControllerResource (rather than ConfigConnector) to keep all controller pod resource modifications in one place, consistent with the existing pattern for container resources, replicas, and VPA settings.

## Usage

```yaml
kind: ControllerResource
metadata:
  name: cnrm-controller-manager
spec:
  metadataHost: "metadata.google.internal"
---
kind: ControllerResource
metadata:
  name: cnrm-webhook-manager
spec:
  metadataHost: "metadata.google.internal"
```

## Use Cases

- **IPv6-only GKE clusters**: When these become available, pods will need to use DNS (`metadata.google.internal`) or an IPv6 address (`fd20:ce::254`) instead of the IPv4 link-local address
- **Dual-stack clusters preferring IPv6**: Similar connectivity considerations
- **Metadata proxy configurations**: Organizations using metadata proxies for security auditing, credential scoping, or SSRF protection (e.g., [k8s-metadata-proxy](https://github.com/GoogleCloudPlatform/k8s-metadata-proxy))
- **Custom credential vending**: Integration with tools like HashiCorp Vault that can act as metadata proxies

## Changes

- `operator/pkg/apis/core/customize/v1beta1/controllerresource_types.go`: Added `MetadataHost` field to ControllerResourceSpec
- `operator/pkg/controllers/configconnector/configconnector_controller.go`: Call ApplyMetadataHost in applyControllerResourceCR
- `operator/pkg/controllers/utils.go`: Added ApplyMetadataHost function
- `operator/pkg/controllers/configconnector/metadata_host_test.go`: Updated tests for new location

### DIFF
--- a/operator/config/crd/bases/customize.core.cnrm.cloud.google.com_controllerresources.yaml
+++ b/operator/config/crd/bases/customize.core.cnrm.cloud.google.com_controllerresources.yaml
@@ -97,6 +97,11 @@ spec:
                   - resources
                   type: object
                 type: array
+              metadataHost:
+                description: |-
+                  MetadataHost overrides the GCP metadata server hostname (injected as GCE_METADATA_HOST).
+                  Useful for IPv6-only clusters where the default 169.254.169.254 is unreachable.
+                type: string
               replicas:
                 description: |-
                   The number of desired replicas of the config connector controller.
--- a/operator/pkg/apis/core/customize/v1alpha1/controllerresource_types.go
+++ b/operator/pkg/apis/core/customize/v1alpha1/controllerresource_types.go
@@ -50,6 +50,10 @@ type ControllerResourceSpec struct {
 	// +optional
 	// +kubebuilder:default=Disabled
 	VerticalPodAutoscalerMode *v1beta1.VPAMode `json:"verticalPodAutoscalerMode,omitempty"`
+	// MetadataHost overrides the GCP metadata server hostname (injected as GCE_METADATA_HOST).
+	// Useful for IPv6-only clusters where the default 169.254.169.254 is unreachable.
+	// +optional
+	MetadataHost string `json:"metadataHost,omitempty"`
 }
 
 // ContainerResourceSpec is the specification of the resource customization for a container of


operator/pkg/controllers/configconnector/metadata_host.go
```diff
@@ -0,0 +1,105 @@
...
+	if cc.Spec.MetadataHost == "" {
+		return nil
+	}
+
+	log := log.FromContext(ctx)
+	log.Info("applying GCE_METADATA_HOST environment variable", "metadataHost", cc.Spec.MetadataHost)
+
+	for _, item := range m.Items {
+		// Apply to StatefulSets (controller-manager, deletiondefender, etc.)
+		if item.Kind != "StatefulSet" && item.Kind != "Deployment" {
```
marqc: Should the list also include `DaemonSet`?

nit: the comment only mentions Statefulsets, but code also allows other kinds.

operator/pkg/controllers/configconnector/metadata_host.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// Check if GCE_METADATA_HOST is already set
+	for _, e := range existingEnv {
+		envMap, ok := e.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		name, _, _ := unstructured.NestedString(envMap, "name")
+		if name == "GCE_METADATA_HOST" {
+			// Already set, update the value
```
marqc: Is this actually needed? This change would be safer if it only sets the field when it doesn't exist. Mutating existing fields seems more error prone.

operator/pkg/controllers/configconnector/metadata_host.go
```diff
@@ -0,0 +1,105 @@
...
+			if err := unstructured.SetNestedField(envMap, metadataHost, "value"); err != nil {
+				return fmt.Errorf("failed to update GCE_METADATA_HOST value: %w", err)
+			}
+			return unstructured.SetNestedSlice(container, existingEnv, "env")
```
marqc: nit: for consistency you can wrap the error message in the same way as you do when you add env in line 100

marqc (commented):
Generally LGTM. Thank you for your contribution Grace! I left few comments.

operator/pkg/controllers/configconnector/metadata_host.go
GraceAtwood: Yep you're right. made that change sorry about that

operator/pkg/controllers/configconnector/metadata_host.go
GraceAtwood: Yeah that's way smarter 😅 just returns nil now if it's already set.

operator/pkg/controllers/configconnector/metadata_host.go
GraceAtwood: Since we're exiting early in this case, i just deleted this whole portion now

operator/pkg/apis/core/v1beta1/configconnector_types.go
```diff
@@ -69,6 +69,14 @@ type ConfigConnectorSpec struct {
...
 	// ConfigConnector specific experiments
 	Experiments *CCExperiments `json:"experiments,omitempty"`
+
+	// MetadataHost specifies the hostname to use for GCP metadata server requests.
+	// This is useful for IPv6-only GKE clusters where the default metadata IP (169.254.169.254)
+	// is not reachable. Set this to "metadata.google.internal" for IPv6 environments.
+	// When set, this value is injected as the GCE_METADATA_HOST environment variable
+	// into the controller manager and related containers.
+	//+kubebuilder:validation:Optional
+	MetadataHost string `json:"metadataHost,omitempty"`
```
xiaoweim: Can we move this field to the [controllerresource](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/operator/pkg/apis/core/customize/v1beta1/controllerresource_types.go#L40)? We want to keep the override/modification to controller pod resources in one place. 

The configconnector controller will handle the controller resource CR https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/operator/pkg/controllers/configconnector/configconnector_controller.go#L653

GraceAtwood: Yep I'll get that moved now thanks for the review

GraceAtwood: @xiaoweim ok I think I moved it correctly? mind giving it another look?

xiaoweim:
Defer for approval

/assign @cheftako

xiaoweim:
Could you please fix the unit test and make sure the v1alpha1 and v1beta1 version of `controllerresource` remain the same? Probably need to add the override field in v1alpha1 too.

PR #5746, opened 2025-12-04
bug: tolerate SQLInstance edition not being set

justinsb:
We need to default/normalize both desired and actual, it looks like.

### DIFF
--- a/pkg/controller/direct/sql/sqlinstance_controller.go
+++ b/pkg/controller/direct/sql/sqlinstance_controller.go
@@ -610,11 +610,11 @@ func (a *sqlInstanceAdapter) Update(ctx context.Context, updateOp *directbase.Up
 	if editionField, ok := a.fieldMeta["spec.settings.edition"]; ok {
 		isEditionUnmanaged = editionField.isUnmanaged
 	}
-	isSettingsUnamanged := false
+	isSettingsUnmanaged := false
 	if settingsField, ok := a.fieldMeta["spec.settings"]; ok {
-		isSettingsUnamanged = settingsField.isUnmanaged
+		isSettingsUnmanaged = settingsField.isUnmanaged
 	}
-	isEditionUnmanaged = isEditionUnmanaged || isSettingsUnamanged
+	isEditionUnmanaged = isEditionUnmanaged || isSettingsUnmanaged
 
 	// Next, handle database edition updates
 	if !isEditionUnmanaged {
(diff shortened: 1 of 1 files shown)


pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -623,7 +623,12 @@ func (a *sqlInstanceAdapter) Update(ctx context.Context, updateOp *directbase.Up
 			desiredEdition = *a.desired.Spec.Settings.Edition
 		}
 
-		if desiredEdition != a.actual.Settings.Edition {
+		actualEdition := "ENTERPRISE" // Default value
```
cheftako, line 626: I think we need to determine the right behavior for MySQL 8.4 which defaults to ENTERPRISE_PLUS.

cheftako: Can we add a test for MySQL 8.4 which doesn't have edition defined but does have machine series set to C4A (https://docs.cloud.google.com/sql/docs/mysql/machine-series-overview) ? I don't think that your allowed to downgrade in that case. Worried this may cause a re-reconcile issue.

justinsb: Good idea: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5814 (starting with mockgcp)

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -610,11 +610,11 @@ func (a *sqlInstanceAdapter) Update(ctx context.Context, updateOp *directbase.Up
 	if editionField, ok := a.fieldMeta["spec.settings.edition"]; ok {
 		isEditionUnmanaged = editionField.isUnmanaged
 	}
-	isSettingsUnamanged := false
+	isSettingsUnmanaged := false
```
cheftako, line 613: Agent: Good catch on fixing this typo!

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -623,7 +623,12 @@ func (a *sqlInstanceAdapter) Update(ctx context.Context, updateOp *directbase.Up
 			desiredEdition = *a.desired.Spec.Settings.Edition
 		}
```
cheftako, line 624: Agent: While this default is correct, the same hardcoded string `"ENTERPRISE"` is used on line 621 for defaulting the `desiredEdition`.

To improve maintainability and avoid magic strings, please define a constant for this default value (e.g., `const defaultSQLEdition = "ENTERPRISE"`) at the package level and use it in both places.

pkg/controller/direct/sql/sqlinstance_controller.go
```diff
@@ -623,7 +623,12 @@ func (a *sqlInstanceAdapter) Update(ctx context.Context, updateOp *directbase.Up
...
 
-		if desiredEdition != a.actual.Settings.Edition {
+		actualEdition := "ENTERPRISE" // Default value
+		if a.actual.Settings.Edition != "" {
+			actualEdition = a.actual.Settings.Edition
```
cheftako, line 628: Agent: To prevent regressions, could you please add a unit test that specifically covers this case? The test should verify that no update is triggered when the `edition` is omitted from the spec and the actual resource in GCP has the default "ENTERPRISE" edition.

cheftako: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5814 (starting with mockgcp)

cheftako: #5814 is merged and #5746 has been rebased on the change.

PR #5750, opened 2025-12-04
mockgcp: allow UpdateDashboard requests without an etag

justinsb:
Ideally we would always do updates with an etag, but gcloud seems to send updates without etags sometimes.


justinsb:
Actually let's hold this one, I think it was probably #5766 

/hold

PR #5752, opened 2025-12-04
mockgcp: switch edgecontainer to use grpctohttp

justinsb:
- **mockgcp: bring edgecontainercluster to latest approaches**
  

- **mockgcp: switch edgecontainer to use grpctohttp**


justinsb:
/assign @cheftako

Not particularly urgent, but one we can probably roll out across the codebase using gemini, giving this as an example (at least the first two commits)

PR #5759, opened 2025-12-05
versionbump: golangci to v2.7.1

justinsb:
This blocks updating our golang version.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -12,61 +12,59 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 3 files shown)


.golangci.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2024 Google LLC
```
cheftako, line 1: Don't think we should be updating the copyright.

justinsb: Fixed.  (The configuration format for golangci changed, which is what triggered this mess)

PR #5766, opened 2025-12-05
tests: fix sed expression to tolerate slashes in dashboard etag

justinsb:
Our sed expression was not handling slashes in the etag correctly.

Update the sed delimiter to '@' to avoid conflicts with slashes.


justinsb:
/assign @cheftako 

I made a mistake here that is causing flakes (when the etag contains a slash)

PR #5771, opened 2025-12-06
tests: write diff files when running two controllers

justinsb:
This makes it easier to see the differences.

### DIFF
--- a/tests/e2e/unified_test.go
+++ b/tests/e2e/unified_test.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"hash/fnv"
 	"io"
@@ -25,6 +26,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"slices"
 	"strconv"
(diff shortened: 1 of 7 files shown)


justinsb:
/assign @cheftako 

WDYT ... worth it for easier reviewing?

tests/e2e/unified_test.go
```diff
@@ -310,6 +314,65 @@ func testFixturesInSeries(ctx context.Context, t *testing.T, scenarioOptions Sce
...
+				// This is expected when files differ, so do not treat as an error
+			} else {
+				t.Errorf("error running diff command: %v", err)
+			}
+		}
+	}
+
+	// _http.log
+	{
+		oldPath := filepath.Join(dir, "_http_old_controller.log")
```
cheftako, line 352: When are the "old_controller" files reset?
I like the idea but the validity/correctness of this approach seems to rely on the lifecycle of these files. 
Seems like the following should be true.
- At PR time "old_controller" + "diff" == "new"
- We should probably have pre submit check that the above is true.
- ?Post submit we should set "old_controller" == "new".

justinsb: So old controller == tf/dcl.  I think we would remove "old" when we remove tf/dcl for that kind.

Currently there's a hacky way we opt-in based on the test path, just so we can ease into this.  But ideally we will opt-in all controllers that have both DCL + direct or TF + direct.

I think if we make these into golden files we'll get the presubmit checks, I agree that will be better

PR #5789, opened 2025-12-09
tests: fix edgecontainer test

justinsb:
Setting the resourceID caused the two projects to be the same project.

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/edgecontainer/v1beta1/edgecontainercluster/_http.log
+++ b/pkg/test/resourcefixture/testdata/basic/edgecontainer/v1beta1/edgecontainercluster/_http.log
@@ -345,10 +345,10 @@ X-Xss-Protection: 0
 
 ---
 
-GET https://cloudresourcemanager.googleapis.com/v1/projects/${projectId}?alt=json&prettyPrint=false
+GET https://cloudresourcemanager.googleapis.com/v1/projects/project-${uniqueId}?alt=json&prettyPrint=false
 User-Agent: google-api-go-client/0.5 kcc/${kccVersion} (+https://github.com/GoogleCloudPlatform/k8s-config-connector) kcc/controller-manager/${kccVersion}
 
-200 OK
+403 Forbidden
 Content-Type: application/json; charset=UTF-8
 Server: ESF
 Vary: Origin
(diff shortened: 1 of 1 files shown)


pkg/test/resourcefixture/testdata/basic/edgecontainer/v1beta1/edgecontainercluster/dependencies.yaml
```diff
@@ -37,7 +37,7 @@ metadata:
   annotations:
     cnrm.cloud.google.com/deletion-policy: abandon
 spec:
-  resourceID: ${projectId}
```
gemmahou, line 40: `resourceID` seems unnecessary here, since it is the same as the `metadata.name`, and we are not acquiring any pre-existing project. Can we remove `resourceID`, and change the` name` field to `project-${uniqueId}` to avoid confusion? 

Furthermore, I do not think we want to keep this random test project, can we remove the `deletion-policy: abandon`?

justinsb: Good call.  Removed the abandon-on-delete annotation and spec.resourceID

pkg/test/resourcefixture/testdata/basic/edgecontainer/v1beta1/edgecontainercluster/_http.log
```diff
@@ -345,10 +345,10 @@ X-Xss-Protection: 0
 
 ---
 
-GET https://cloudresourcemanager.googleapis.com/v1/projects/${projectId}?alt=json&prettyPrint=false
+GET https://cloudresourcemanager.googleapis.com/v1/projects/project-${uniqueId}?alt=json&prettyPrint=false
 User-Agent: google-api-go-client/0.5 kcc/${kccVersion} (+https://github.com/GoogleCloudPlatform/k8s-config-connector) kcc/controller-manager/${kccVersion}
 
-200 OK
+403 Forbidden
```
cheftako, line 351: NIT: is this supposed to be a 403 for the test? This makes me wonder if we hit some sort of error in the recording?

justinsb: It's the project creation flow - we first GET to check that the project does not yet exist.  On creation, the project does not exist, and we return a 403 instead of a 404 because security.

PR #5796, opened 2025-12-09
TagsLocationTagBinding: CRD/Mapper/Fuzzer

justinsb:
- **TagsLocationTagBinding: CRD/Mapper/Fuzzer**
  

- **autogen: TagsLocationTagBinding CRD/Mapper/Fuzzer**


apis/tags/v1alpha1/tagbinding_types.go
```diff
@@ -0,0 +1,102 @@
...
+
+// TagsLocationTagBindingSpec defines the desired state of TagsLocationTagBinding
+// +kcc:spec:proto=google.cloud.resourcemanager.v3.TagBinding
+type TagsLocationTagBindingSpec struct {
+	// +kcc:ref=Project
+	ParentRef *v1beta1.TagsTagBindingParentRef `json:"parentRef"`
+
+	// The location for the resource being tagged.
+	// +required
+	Location string `json:"location"`
```
cheftako, line 33: Should this be a *string ? I realize its required but seems like switching to pointer allows for things like the field becoming optional later. Might be easier to just be consistent on this.

justinsb:
> @justinsb Its alpha so lets merge but I do think we should consider making Location a '*string'.

Ah - I see!  It's required and "" is not valid, but yes, agree that `*string` is probably something we could just standardize on, rather than thinking about it every time

PR #5797, opened 2025-12-10
tests: kccmanager election smoke test

acpana:
Adds a smoke test for validating basic leader election interaction with the kccmanager!

### DIFF
--- a/pkg/controller/kccmanager/kccmanager.go
+++ b/pkg/controller/kccmanager/kccmanager.go
@@ -99,48 +99,75 @@ type Config struct {
 
 	// Configure manager to participate in leader election if MultiClusterLease is enabled.
 	MultiClusterLease bool
+
+	// SkipControllerRegistration is true if we should skip registering the default controllers
+	// used for testing purposes to avoid controller name conflicts when creating multiple managers
+	SkipControllerRegistration bool
+
+	// MultiClusterLeaseConfig is the configuration for the multi-cluster lease.
+	// If specified, this configuration takes precedence over the ConfigConnector object.
+	// This is primarily used for testing to simulate multiple clusters.
+	MultiClusterLeaseConfig *operatorv1beta1.MultiClusterLeaseSpec
 }
 
-func setUpMultiClusterLease(ctx context.Context, restConfig *rest.Config, scheme *runtime.Scheme) (*leaderelection.LeaderElectionConfig, error) {
-	// Create a temporary client to read the ConfigConnector object for leader election config.
-	c, err := crclient.New(restConfig, crclient.Options{Scheme: scheme})
-	if err != nil {
-		return nil, fmt.Errorf("error creating temporary client: %w", err)
-	}
+func setUpMultiClusterLease(ctx context.Context, restConfig *rest.Config, scheme *runtime.Scheme, explicitConfig *operatorv1beta1.MultiClusterLeaseSpec, existOnLeadershipLoss bool) (*leaderelection.LeaderElectionConfig, error) {
+	var leaseSpec *operatorv1beta1.MultiClusterLeaseSpec
 
-	// Get the ConfigConnector object.
-	cc := &operatorv1beta1.ConfigConnector{}
-	ccName := types.NamespacedName{Name: "configconnector.core.cnrm.cloud.google.com"}
-	klog.Infof("checking for ConfigConnector object")
-	if err := c.Get(ctx, ccName, cc); err != nil {
-		// If the ConfigConnector object is not found, proceed with default leader election.
-		klog.Infof("ConfigConnector object not found, using default in-cluster leader election")
+	if explicitConfig != nil {
+		klog.Infof("using explicit multi-cluster lease config")
+		leaseSpec = explicitConfig
 	} else {
+		// Create a temporary client to read the ConfigConnector object for leader election config.
+		c, err := crclient.New(restConfig, crclient.Options{Scheme: scheme})
+		if err != nil {
+			return nil, fmt.Errorf("error creating temporary client: %w", err)
+		}
+
+		// Get the ConfigConnector object.
+		cc := &operatorv1beta1.ConfigConnector{}
+		ccName := types.NamespacedName{Name: "configconnector.core.cnrm.cloud.google.com"}
+		klog.Infof("checking for ConfigConnector object")
+		if err := c.Get(ctx, ccName, cc); err != nil {
+			// If the ConfigConnector object is not found, proceed with default leader election.
+			klog.Infof("ConfigConnector object not found, using default in-cluster leader election")
+			return nil, nil
+		}
 		klog.Infof("found ConfigConnector object")
 		if cc.Spec.Experiments != nil && cc.Spec.Experiments.LeaderElection != nil && cc.Spec.Experiments.LeaderElection.MultiClusterLease != nil {
 			klog.Infof("multi-cluster leader election is configured")
-			leaseSpec := cc.Spec.Experiments.LeaderElection.MultiClusterLease
-			lock := mcleclient.New(
-				c,
-				leaseSpec.LeaseName,
-				leaseSpec.Namespace,
-				leaseSpec.GlobalLockName,
-				15*time.Second,
-			)
-			return &leaderelection.LeaderElectionConfig{
-				Lock:          lock,
-				LeaseDuration: 15 * time.Second,
-				RenewDeadline: 10 * time.Second,
-				RetryPeriod:   2 * time.Second,
-				Name:          leaseSpec.LeaseName,
-				Callbacks: leaderelection.LeaderCallbacks{
-					OnStoppedLeading: func() {
-						klog.ErrorS(nil, "leaderelection lost")
-						klog.FlushAndExit(klog.ExitFlushTimeout, 1)
-					},
-				},
-			}, nil
+			leaseSpec = cc.Spec.Experiments.LeaderElection.MultiClusterLease
 		}
+	}
+
+	if leaseSpec != nil {
+		// Create a new client for the lock to ensure it uses the correct configuration
+		c, err := crclient.New(restConfig, crclient.Options{Scheme: scheme})
+		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/kccmanager/kccmanager.go
```diff
@@ -99,48 +99,77 @@ type Config struct {
...
+	// SkipControllerRegistration is true if we should skip registering the default controllers
+	// used for testing purposes to avoid controller name conflicts when creating multiple managers
+	// SkipControllerRegistration is true if we should skip registering the default controllers
+	// used for testing purposes to avoid controller name conflicts when creating multiple managers
+	SkipControllerRegistration bool
+
+	// MultiClusterLeaseConfig is the configuration for the multi-cluster lease.
+	// If specified, this configuration takes precedence over the ConfigConnector object.
+	// This is primarily used for testing to simulate multiple clusters.
+	MultiClusterLeaseConfig *operatorv1beta1.MultiClusterLeaseSpec
```
acpana, line 110: Not a big fan of these "test knobs" but I'm pretty excited to fairly test the kccmanager in isolation for leadership!

- no controllers
- no "fake" auth for TF/ DCL

It gives me hope that we can further build on top of this for whenever we need to test some kccmanager function that is not necessarily TF or DCL related.

pkg/controller/kccmanager/kccmanager_multicluster_test.go
```diff
@@ -0,0 +1,305 @@
...
+			return false, nil
+		}
+		if lease.Spec.HolderIdentity != nil && *lease.Spec.HolderIdentity != "" {
+			return true, nil
+		}
+		return false, nil
+	})
+}
+
+func waitForObject(ctx context.Context, c client.Client, nn types.NamespacedName, obj client.Object, timeout time.Duration) error {
```
xiaoweim, line 251: Can we also use `wait.PollUntilContextTimeout` here too?

pkg/controller/kccmanager/kccmanager.go
xiaoweim: +1 not a big fan on adding test-specific fields to a production config. 

Not a blocker: maybe we could have a separate test-only config or something similar in the future?

pkg/controller/kccmanager/kccmanager_multicluster_test.go
acpana: replaced internals!

pkg/controller/kccmanager/kccmanager.go
maqiuyujoyce: Hmmm, can we have a test-only interface? Unless you're confirmed to split out the test-only logic in a quarter (so that we still have the context), I'd rather we do it right in the beginning to avoid polluting this interface.

Here is what Gemini suggested:

```
  Here's how you can implement it:

  1. Define an Internal Configuration Struct

  Create an unexported (internal) struct that holds both your public Config and any test-specific fields.

    1 // In pkg/controller/kccmanager/kccmanager.go
    2 
    3 // Config is the public configuration struct for kccmanager.
    4 // Keep this struct clean and free of test-only fields.
    5 type Config struct {
    6     ManagerOptions         manager.Options
    7     UserProjectOverride    string
    8     BillingProject         string
    9     GCPAccessToken         string
   10     EnableMetricsTransport bool
   11     MultiClusterLease      bool
   12 }
   13 
   14 // internalConfig is used internally by New to aggregate all configuration,
   15 // including test-only overrides.
   16 type internalConfig struct {
   17     Config // Embed the public Config struct
   18 
   19     // Test-only fields:
   20     skipControllerRegistration bool
   21     multiClusterLeaseConfig    *operatorv1beta1.MultiClusterLeaseSpec
   22     existOnLeadershipLoss      bool // Control the exit behavior
   23 }
   24 
   25 // Option defines a functional option for the New constructor.
   26 type Option func(*internalConfig)

  2. Create Test-Specific Functional Options

  These public functions allow tests to set the internal-only fields.

    1 // WithTestSkipControllerRegistration allows skipping controller registration for testing.
    2 func WithTestSkipControllerRegistration(skip bool) Option {
    3     return func(ic *internalConfig) {
    4         ic.skipControllerRegistration = skip
    5     }
    6 }
    7 
    8 // WithTestMultiClusterLeaseConfig allows injecting a MultiClusterLeaseSpec for testing.
    9 func WithTestMultiClusterLeaseConfig(spec *operatorv1beta1.MultiClusterLeaseSpec) Option {
   10     return func(ic *internalConfig) {
   11         ic.multiClusterLeaseConfig = spec
   12     }
   13 }
   14 
   15 // WithTestExistOnLeadershipLoss controls the exit behavior on leadership loss for testing.
   16 func WithTestExistOnLeadershipLoss(exist bool) Option {
   17     return func(ic *internalConfig) {
   18         ic.existOnLeadershipLoss = exist
   19     }
   20 }

  3. Update the New Constructor

  Modify New to accept these functional options. Inside New, apply the options to the internalConfig and then use those fields for your logic.

    1 // New creates a new kccmanager.Manager.
    2 // It accepts optional functional options for advanced/testing scenarios.
    3 func New(ctx context.Context, restConfig *rest.Config, cfg Config, opts ...Option) (manager.Manager, error) {
    4     // Initialize the internalConfig with the public Config and default test values.
    5     ic := &internalConfig{
    6         Config: cfg,
    7         // Set production-like defaults for test-only fields.
    8         // If not explicitly overridden by options, these defaults apply.
    9         skipControllerRegistration: false, // In production, we *don't* skip controller registration
   10         multiClusterLeaseConfig:    nil,   // In production, we *don't* inject a static config
   11         existOnLeadershipLoss:      true,  // In production, we *do* exit on leadership loss
   12     }
   13 
   14     // Apply any provided functional options, overriding the defaults.
   15     for _, opt := range opts {
   16         opt(ic)
   17     }
   18 
   19     // ... (rest of your existing New function logic) ...
   20 
   21     // Use ic.Config.ManagerOptions for manager creation
   22     opts := manager.Options{
   23         Scheme: ic.Config.ManagerOptions.Scheme,
   24         Metrics: ic.Config.ManagerOptions.Metrics,
   25         HealthProbeBindAddress: ic.Config.ManagerOptions.HealthProbeBindAddress,
   26         LeaderElection: ic.Config.ManagerOptions.LeaderElection,
   27         LeaderElectionID: ic.Config.ManagerOptions.LeaderElectionID,
   28     }
   29     mgr, err := manager.New(restConfig, opts)
   30     if err != nil {
   31         return nil, fmt.Errorf("error creating new manager: %w", err)
   32     }
   33 
   34     var leConfig *leaderelection.LeaderElectionConfig
   35     if ic.Config.MultiClusterLease {
   36         // Use ic.multiClusterLeaseConfig and ic.existOnLeadershipLoss for setUpMultiClusterLease
   37         leConfig, err = setUpMultiClusterLease(ctx, restConfig, mgr.GetScheme(), ic.multiClusterLeaseConfig, ic.existOnLeadershipLoss)
   38         if err != nil {
   39             return nil, fmt.Errorf("error setting up multi-cluster leader election: %w", err)
   40         }
   41     }
   42 
   43     // Conditionally register controllers based on ic.skipControllerRegistration
   44     if !ic.skipControllerRegistration {
   45         // ... (existing TF, DCL, and direct controller bootstrapping logic) ...
   46         // Ensure this part only runs if not skipped by the test option.
   47         // It's important to move all of this logic inside this 'if' block.
   48     }
   49 
   50     // ... (rest of the New function) ...
   51 
   52     return mgr, nil
   53 }

  4. Update Your Test Code

  In your kccmanager_multicluster_test.go, your newTestManagerWithConfig helper would now look like this:

    1 func newTestManagerWithConfig(t *testing.T, cfg *rest.Config, scheme *runtime.Scheme, identity string, namespace string) (manager.Manager, *startSignal) {
    2     t.Helper()
    3 
    4     mclConfig := &operatorv1beta1.MultiClusterLeaseSpec{
    5         LeaseName:      testLeaseName,
    6         Namespace:      namespace,
    7         GlobalLockName: identity,
    8     }
    9 
   10     // Construct the base Config (production-like)
   11     kccCfg := kccmanager.Config{
   12         ManagerOptions: manager.Options{
   13             Scheme:                 scheme,
   14             Metrics:                metricsserver.Options{BindAddress: "0"},
   15             HealthProbeBindAddress: "0",
   16             LeaderElection:         false, // Default to false in tests, then override with MultiClusterLease
   17             LeaderElectionID:       identity,
   18         },
   19         MultiClusterLease: true, // Enable MultiClusterLease for the test scenario
   20     }
   21 
   22     // Call New with functional options to inject test-specific behaviors
   23     mgr, err := kccmanager.New(
   24         context.Background(),
   25         cfg,
   26         kccCfg,
   27         kccmanager.WithTestSkipControllerRegistration(true),       // Skip heavy prod initializations
   28         kccmanager.WithTestMultiClusterLeaseConfig(mclConfig),     // Inject test lease config
   29         kccmanager.WithTestExistOnLeadershipLoss(false),           // Prevent test environment from exiting
   30     )
   31     if err != nil {
   32         t.Fatalf("error creating new manager for %s: %v", identity, err)
   33     }
   34 
   35     signal := &startSignal{started: make(chan struct{})}
   36     if err := mgr.Add(signal); err != nil {
   37         t.Fatalf("error adding start signal to manager %s: %v", identity, err)
   38     }
   39 
   40     return mgr, signal
   41 }

  Benefits of this "Cheap" Approach:

   * Clean Public API: Your kccmanager.Config struct remains exactly as a production user would expect it, without any fields like SkipControllerRegistration.
   * Minimal Impact: The core New function is modified, but its public signature for common use cases (New(ctx, restConfig, cfg)) is largely unchanged; you just add opts ...Option to the end.
   * Clear Separation: Test-only concerns are clearly demarcated within their own Option functions and only applied when explicitly invoked by tests.
   * Go Idiomatic: This is a widely accepted and understood pattern in Go for handling optional parameters to constructors.

```

acpana: I think this is a splendid idea. Chatting offline, I will follow up in a separate PR!

PR #5798, opened 2025-12-10
TagsLocationTagBinding: direct controller

justinsb:
- **TagsLocationTagBinding: direct controller**
  

- **TagsLocationTagBinding: add support for binding to StorageBucket**
  

- **autogen: golden output for direct TagsLocationTagBucket**

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/tags/v1alpha1/tagslocationtagbinding/tagslocationtagbinding_bucket/_final_object.diff
+++ b/pkg/test/resourcefixture/testdata/basic/tags/v1alpha1/tagslocationtagbinding/tagslocationtagbinding_bucket/_final_object.diff
@@ -1,0 +1,4 @@
+6d5
+<     cnrm.cloud.google.com/state-into-spec: absent
+28a28
+>   externalRef: tagBindings/%2F%2Fstorage.googleapis.com%2Fprojects%2F_%2Fbuckets%2Fstoragebucket-${uniqueId}/tagValues/${tagValueID}
--- a/pkg/test/resourcefixture/testdata/basic/tags/v1alpha1/tagslocationtagbinding/tagslocationtagbinding_bucket/_final_object_old_controller.golden.yaml
+++ b/pkg/test/resourcefixture/testdata/basic/tags/v1alpha1/tagslocationtagbinding/tagslocationtagbinding_bucket/_final_object_old_controller.golden.yaml
@@ -3,6 +3,7 @@ kind: TagsLocationTagBinding
 metadata:
   annotations:
     cnrm.cloud.google.com/management-conflict-prevention-policy: none
+    cnrm.cloud.google.com/state-into-spec: absent
   finalizers:
   - cnrm.cloud.google.com/finalizer
   - cnrm.cloud.google.com/deletion-defender
@@ -25,6 +26,5 @@ status:
     reason: UpToDate
     status: "True"
     type: Ready
-  externalRef: tagBindings/%2F%2Fstorage.googleapis.com%2Fprojects%2F_%2Fbuckets%2Fstoragebucket-${uniqueId}/tagValues/${tagValueID}
   name: tagBindings/%2F%2Fstorage.googleapis.com%2Fprojects%2F_%2Fbuckets%2Fstoragebucket-${uniqueId}/tagValues/${tagValueID}
   observedGeneration: 2
--- a/pkg/test/resourcefixture/testdata/basic/tags/v1alpha1/tagslocationtagbinding/tagslocationtagbinding_bucket/_http.diff
+++ b/pkg/test/resourcefixture/testdata/basic/tags/v1alpha1/tagslocationtagbinding/tagslocationtagbinding_bucket/_http.diff
@@ -1,0 +1,14 @@
+326c326
+< POST https://us-east4-cloudresourcemanager.googleapis.com/v3/tagBindings?alt=json
+---
+> POST https://us-east4-cloudresourcemanager.googleapis.com/v3/tagBindings?%24alt=json%3Benum-encoding%3Dint
+358c358
+< GET https://us-east4-cloudresourcemanager.googleapis.com/v3/tagBindings/?alt=json&pageSize=300&parent=%2F%2Fstorage.googleapis.com%2Fprojects%2F_%2Fbuckets%2Fstoragebucket-${uniqueId}
+---
+> GET https://us-east4-cloudresourcemanager.googleapis.com/v3/tagBindings?%24alt=json%3Benum-encoding%3Dint&parent=%2F%2Fstorage.googleapis.com%2Fprojects%2F_%2Fbuckets%2Fstoragebucket-${uniqueId}
+384c384
+< DELETE https://us-east4-cloudresourcemanager.googleapis.com/v3/tagBindings/%2F%2Fstorage.googleapis.com%2Fprojects%2F_%2Fbuckets%2Fstoragebucket-${uniqueId}/tagValues/${tagValueID}?alt=json
+---
+> DELETE https://us-east4-cloudresourcemanager.googleapis.com/v3/tagBindings/%252F%252Fstorage.googleapis.com%252Fprojects%252F_%252Fbuckets%252Fstoragebucket-${uniqueId}/tagValues/${tagValueID}?%24alt=json%3Benum-encoding%3Dint
+386a387
+> X-Goog-Request-Params: name=tagBindings%2F%252F%252Fstorage.googleapis.com%252Fprojects%252F_%252Fbuckets%252Fstoragebucket-${uniqueId}%2FtagValues%2F${tagValueID}
--- a/pkg/test/resourcefixture/testdata/basic/tags/v1alpha1/tagslocationtagbinding/tagslocationtagbinding_bucket/_http_old_controller.log
+++ b/pkg/test/resourcefixture/testdata/basic/tags/v1alpha1/tagslocationtagbinding/tagslocationtagbinding_bucket/_http_old_controller.log
@@ -323,7 +323,7 @@ X-Xss-Protection: 0
 
 ---
 
-POST https://us-east4-cloudresourcemanager.googleapis.com/v3/tagBindings?%24alt=json%3Benum-encoding%3Dint
+POST https://us-east4-cloudresourcemanager.googleapis.com/v3/tagBindings?alt=json
 Content-Type: application/json
 User-Agent: kcc/${kccVersion} (+https://github.com/GoogleCloudPlatform/k8s-config-connector) kcc/controller-manager/${kccVersion}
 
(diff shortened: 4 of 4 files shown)


apis/tags/v1beta1/tagbinding_parentref.go
```diff
@@ -62,39 +69,51 @@ func (r *TagsTagBindingParentRef) SetExternal(ref string) {
...
+		return err
+	}
+
+	// Include the service qualification in case of ambiguities
+	r.External = "//" + service + "/" + id.GetExternal()
+
+	return nil
+}
+
+func (r *TagsTagBindingParentRef) resolveReference() (string, refs.Ref) {
```
justinsb, line 99: Introducing this to allow for more referenced types without too much code repetition.  I think it could also clean up how we handle Export

apis/tags/v1beta1/tagbinding_parentref.go
```diff
@@ -62,39 +69,51 @@ func (r *TagsTagBindingParentRef) SetExternal(ref string) {
...
+	}
+
+	if err := id.Normalize(ctx, reader, defaultNamespace); err != nil {
+		return err
+	}
+
+	// Include the service qualification in case of ambiguities
+	r.External = "//" + service + "/" + id.GetExternal()
+
+	return nil
```
cheftako, line 96: AGENT: The reconstruction of `r.External` here seems to have a bug. `id.GetExternal()` can return a full resource path that already includes the service name (e.g., `//storage.googleapis.com/...`). Concatenating it with the `service` variable will result in a duplicated service prefix and a malformed URL (e.g., `//service.com///service.com/resource`).

The logic should be corrected to ensure a valid, non-duplicated resource name is generated.

apis/tags/v1beta1/tagbinding_parentref.go
```diff
@@ -62,39 +69,51 @@ func (r *TagsTagBindingParentRef) SetExternal(ref string) {
...
+		service := "storage.googleapis.com"
+
+		tokens := strings.Split(suffix, "/")
+		if len(tokens) == 4 && tokens[0] == "projects" && tokens[2] == "buckets" {
+			return service, &storage.StorageBucketRef{
+				External:  suffix,
+				Name:      r.Name,
+				Namespace: r.Namespace,
+			}
 		}
```
cheftako, line 110: AGENT: Parsing the `External` field using `strings.LastIndex` and `strings.HasPrefix` is fragile. This could easily break with different or more complex resource name formats.

A regular expression would be a more robust and readable way to extract the service and resource parts from the full resource name string, similar to the approach in `tagslocationtagbinding_controller.go`.

SAMPLE: https://go.dev/play/p/S5eenFoo6Fz

apis/tags/v1beta1/tagbinding_parentref.go
```diff
@@ -62,39 +69,51 @@ func (r *TagsTagBindingParentRef) SetExternal(ref string) {
...
+	}
+
+	// Include the service qualification in case of ambiguities
+	r.External = "//" + service + "/" + id.GetExternal()
+
+	return nil
+}
+
+func (r *TagsTagBindingParentRef) resolveReference() (string, refs.Ref) {
+	if suffix, ok := strings.CutPrefix(r.External, "//storage.googleapis.com/"); ok {
```
cheftako, line 100: AGENT: This `resolveReference` function contains logic for parsing full resource names. Similar parsing logic exists in the `parseParent` function in `pkg/controller/direct/tags/tagslocationtagbinding_controller.go`. To avoid code duplication and potential for future drift, this parsing logic should be consolidated into a shared utility function.

apis/tags/v1beta1/tagbinding_parentref.go
```diff
@@ -62,39 +69,51 @@ func (r *TagsTagBindingParentRef) SetExternal(ref string) {
...
-	case refs.ProjectGVK:
-		id := &refs.ProjectRef{
-			Name:      r.Name,
-			Namespace: r.Namespace,
-			External:  r.External,
-		}
-		if err := id.Normalize(ctx, reader, defaultNamespace); err != nil {
-			return err
+	service, id := r.resolveReference()
+	if id == nil {
```
cheftako: This `if/else if` chain for determining the reference type is hard to maintain. A `map[string]refs.Ref` or a similar data structure could make this logic more data-driven and easier to extend with new types in the future.

apis/tags/v1beta1/tagbinding_parentref.go
cheftako: There's a potential bug here. `id.GetExternal()` can return a full resource path that already includes the service (e.g., `//storage.googleapis.com/...`). Prepending `//` and the service name again will result in a malformed URL like `//service.com///service.com/resource`. The logic needs to handle this to avoid creating invalid external references.

apis/tags/v1beta1/tagbinding_parentref.go
cheftako: This function `resolveReference` has very similar logic to `parseParent` in the controller. This duplication should be avoided by creating a shared utility function for parsing parent resource strings.

apis/tags/v1beta1/tagbinding_parentref.go
cheftako: Parsing the `External` field using `strings.LastIndex` and `strings.Split` is fragile and can easily break with different or more complex resource name formats.

A regular expression would be a more robust and readable way to extract the service and resource parts from the full resource name string, similar to the approach in `tagslocationtagbinding_controller.go`.

apis/tags/v1beta1/tagbinding_parentref.go
cheftako, line 85: AGENT: This `if/else if` chain for determining the reference type based on the `Kind` is hard to maintain and extend. A `map[string]refs.Ref` or a similar data structure could make this logic more data-driven and easier to add new supported parent types in the future.

apis/tags/v1beta1/tagbinding_parentref.go
```diff
@@ -62,39 +69,51 @@ func (r *TagsTagBindingParentRef) SetExternal(ref string) {
...
+			}
 		}
-		// Include the service qualification in case of ambiguities
-		r.External = "//cloudresourcemanager.googleapis.com/" + id.External
-		return nil
-	default:
-		return fmt.Errorf("unsupported gvk for reference %v", gvk)
+		return "", nil
+	}
+
```
cheftako, line 113: AGENT: The resource part of the name is reconstructed with `strings.Join(parts[3:], "/")`. This is also part of the fragile parsing logic. A regex with named capture groups would be a much safer and more readable way to extract these components.

apis/tags/v1beta1/tagbinding_parentref.go
```diff
@@ -62,39 +69,51 @@ func (r *TagsTagBindingParentRef) SetExternal(ref string) {
...
+				Namespace: r.Namespace,
+			}
 		}
-		// Include the service qualification in case of ambiguities
-		r.External = "//cloudresourcemanager.googleapis.com/" + id.External
-		return nil
-	default:
-		return fmt.Errorf("unsupported gvk for reference %v", gvk)
+		return "", nil
+	}
```
cheftako, line 112: AGENT: Accessing `parts[2]` assumes the `External` string will always conform to the `//service/path` format and produce at least three parts when split. If the string is malformed (e.g., `//foo`), this will cause a runtime panic.

Please add a check to validate the length of the `parts` slice before accessing its elements to prevent this.

apis/tags/v1beta1/tagbinding_parentref.go
cheftako: The resource part of the name is reconstructed using `strings.Join(parts[3:], "/")`. This is part of a fragile parsing scheme that assumes a simple, `/`-delimited structure. A regex with named capture groups would be a much safer and more readable way to extract these components.

apis/tags/v1beta1/tagbinding_parentref.go
```diff
@@ -62,39 +69,51 @@ func (r *TagsTagBindingParentRef) SetExternal(ref string) {
...
+func (r *TagsTagBindingParentRef) resolveReference() (string, refs.Ref) {
+	if suffix, ok := strings.CutPrefix(r.External, "//storage.googleapis.com/"); ok {
+		service := "storage.googleapis.com"
+
+		tokens := strings.Split(suffix, "/")
+		if len(tokens) == 4 && tokens[0] == "projects" && tokens[2] == "buckets" {
+			return service, &storage.StorageBucketRef{
+				External:  suffix,
+				Name:      r.Name,
+				Namespace: r.Namespace,
```
cheftako, line 108: AGENT: The error message `unknown format for external reference` is not very informative. It should include the value of `r.External` to aid in debugging.

cheftako (commented):
Most of the comments do not seem blocking. I would like to see the possible recursive external field value definition concerned addresssed.

AGENT: The PR is a good step towards managing `TagsLocationTagBinding` resources directly and extending tag support to `StorageBucket`. However, several critical issues need to be addressed before it can be merged. These include a bug in how external references are normalized, which could lead to incorrect GCP API calls, and performance concerns due to repeated regex compilations and fragile string parsing logic. Most significantly, there is a major API design inconsistency in the new `TagsLocationTagBinding` CRD, as it uses a raw string for the parent reference instead of KCC's standard structured `Ref` types. Please review the detailed comments for specific suggestions on how to resolve these issues.

apis/tags/v1beta1/tagbinding_parentref.go
justinsb: I don't believe this is true, but I think there's a valid point here: we are not enforcing that External has been fully-qualified or has been normalized or that the normalized format includes (or does not include) the service prefix

apis/tags/v1beta1/tagbinding_parentref.go
justinsb: I don't think this comment is on the right line.  And the current PR does not include LastIndex (though it does include HasPrefix).  Maybe the agent was confused by a rebase?

apis/tags/v1beta1/tagbinding_parentref.go
justinsb: I think I agree with this, it is now clear that we could have a centralized registry of "resolvers" that given an external string could convert it into a Ref or an ID.  Do you want to have a go at this @cheftako, or shall I ask gemini to?

(I think we probably want to base that work on this PR, as this PR is where the patterns become more clear, at least IMO)

apis/tags/v1beta1/tagbinding_parentref.go
justinsb: I think this is looking at an old version of the code.  We've moved it to resolveReference and cleaned it up a bit, but I do agree that this could be refactored further into a central "registry"

apis/tags/v1beta1/tagbinding_parentref.go
justinsb: Gemini might think it can cope with regexes, but I think regexes are actually quite hard to maintain (personally) and they have a noticeable memory footprint.  I would personally ban regexes for things that can be trivially parsed, and URLs are designed to be trivially parsed.

apis/tags/v1beta1/tagbinding_parentref.go
justinsb: This comment is on the wrong line.  I can't find the referenced code.  Agree that we should not check parts[2] without checking len >= 3, but I don't think we do.

apis/tags/v1beta1/tagbinding_parentref.go
justinsb: So we do call Normalize right above.

I think we should have ValidateExternal return `(identity.Identity, error)`, and Normalize also return `(identity.Identity, error)`.  Then I think identity.Identity should implement `QualifiedName() string` (which returns something like `//storage.googleapis.com/buckets/foo`), `UnqualifiedName() string` (which returns something like buckets/foo) and probably `Service() string`.  We could also have it return a struct which returns those 3 things, but I would argue that the identity.Identity _is_ that object.

I propose we do this as part of the "normalizer registry" work, though IDK if it should be a separate PR.  WDYT @cheftako

cheftako: +1.

PR #5799, opened 2025-12-10
fix the commit on which the release branch was created on

xiaoweim:
### BRIEF Change description

Include all three commits related to release to create the release branch.

The issue was spotted on the 140 release and we cherry picked the two remaining commits: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5790

```release-note
NONE
```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/dev/tasks/auto-tag-on-version-update.sh
+++ b/dev/tasks/auto-tag-on-version-update.sh
@@ -41,7 +41,43 @@ echo "Tag v$VERSION does not exist. Proceeding."
 # We tag the end of the release PR (which includes 3 commits), not just the first commit that bumped the version.
 # We assume this script runs on the tip of the release branch/PR.
 COMMIT_HASH=$(git rev-parse HEAD)
+if [ -z "${COMMIT_HASH}" ]; then
+  echo "ERROR: Could not determine commit hash."
+  exit 1
+fi
 echo "Using HEAD as commit to tag: ${COMMIT_HASH}"
+
+# 4. Verify the commit messages of the last 3 commits.
+# Expected order (oldest to newest):
+# 1. Release <VERSION>
+# 2. Update alpha CRDs for Release <VERSION>
+# 3. Update golden files for operator controllers
+
+# HEAD~2 (First commit)
+MSG_FIRST=$(git log --format=%s -n 1 HEAD~2)
+EXPECTED_FIRST="Release ${VERSION}"
+if [ "${MSG_FIRST}" != "${EXPECTED_FIRST}" ]; then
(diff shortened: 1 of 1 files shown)


dev/tasks/auto-tag-on-version-update.sh
```diff
@@ -37,13 +37,11 @@ if [ "$(git tag -l "v${VERSION}")" ]; then
...
-if [ -z "$COMMIT_HASH" ]; then
-  echo "ERROR: Could not find a commit for ${VERSION_FILE}."
-  exit 1
-fi
-echo "Found commit to tag: ${COMMIT_HASH}"
+# 3. Use the current commit (HEAD) as the tag target.
+# We tag the end of the release PR (which includes 3 commits), not just the first commit that bumped the version.
+# We assume this script runs on the tip of the release branch/PR.
+COMMIT_HASH=$(git rev-parse HEAD)
+echo "Using HEAD as commit to tag: ${COMMIT_HASH}"
```
cheftako, line 48: I would still check for not set and error out.

dev/tasks/auto-tag-on-version-update.sh
acpana, line 1: I recall there was a version of this (or maybe separate script entirely) where we checked the commits messages too. Do we want that? It would make the process a bit more cumbersome.

xiaoweim: We tried to verify the commit message in another script. But good idea to double check, I have updated the script to add the verification check.

dev/tasks/auto-tag-on-version-update.sh
```diff
@@ -37,13 +37,47 @@ if [ "$(git tag -l "v${VERSION}")" ]; then
...
+# Expected order (oldest to newest):
+# 1. Release <VERSION>
+# 2. Update alpha CRDs for Release <VERSION>
+# 3. Update golden files for operator controllers
+
+# HEAD~2 (First commit)
+MSG_FIRST=$(git log --format=%s -n 1 HEAD~2)
+EXPECTED_FIRST="Release ${VERSION}"
+if [ "${MSG_FIRST}" != "${EXPECTED_FIRST}" ]; then
+  echo "ERROR: HEAD~2 commit message '${MSG_FIRST}' does not match expected '${EXPECTED_FIRST}'"
```
cheftako, line 60: NIT: We should have a runbook (or similar) for the release eng to follow when the last 3 commits are not what it is expected. Not holding the PR for this but we should think through and capture the steps soon.

PR #5800, opened 2025-12-10
fix: ignore gpu_zonal_redundancy_disabled in RunJob fuzzer

justinsb:
- **fix: ignore gpu_zonal_redundancy_disabled in RunJob fuzzer**


pkg/controller/direct/run/job_fuzzer.go
```diff
@@ -71,6 +71,7 @@ func runJobFuzzer() fuzztesting.KRMFuzzer {
 	f.Unimplemented_NotYetTriaged(".template.template.containers[].startup_probe.http_get")
 	f.Unimplemented_NotYetTriaged(".template.template.containers[].liveness_probe.grpc")
 	f.Unimplemented_NotYetTriaged(".template.template.containers[].liveness_probe.http_get")
+	f.Unimplemented_NotYetTriaged(".template.template.gpu_zonal_redundancy_disabled")
```
cheftako, line 74: Do we have a bug(s) to resolve the Unimplemented?

justinsb: So the idea is that these actually serve as the todo list, a bit like ratcheting.go

PR #5803, opened 2025-12-10
fix: limit memory usage in go fuzz tests

justinsb:
fuzzing is very parallel and generates a lot of GC pressure,
so limit the memory usage to avoid OOMs (particularly in github actions).

### DIFF
--- a/dev/ci/presubmits/fuzz-roundtrippers
+++ b/dev/ci/presubmits/fuzz-roundtrippers
@@ -20,6 +20,8 @@ set -o pipefail
 REPO_ROOT="$(git rev-parse --show-toplevel)"
 cd ${REPO_ROOT}
 
+export GOMEMLIMIT=2GiB # Avoid OOMs, particularly in github actions
+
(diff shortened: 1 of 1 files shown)


dev/ci/presubmits/fuzz-roundtrippers
```diff
@@ -20,6 +20,8 @@ set -o pipefail
 REPO_ROOT="$(git rev-parse --show-toplevel)"
 cd ${REPO_ROOT}
 
+export GOMEMLIMIT=2GiB # Avoid OOMs, particularly in github actions
```
cheftako, line 23: CURIOUS: Where did the 2GiB figure come from? It is documented in GitHub actions?

justinsb: I pulled it out of ... thin air

justinsb: I think the runners actually have 8GB or 16GB, but these fuzzers shouldn't actually be consuming much memory, we just need to encourage go to do GC...  The problem is that the fuzzer is basically allocating memory as fast as it can, and I guess we just aren't prioritizing GC

PR #5804, opened 2025-12-10
mockgcp: implement field filter in serviceusage API

justinsb:


mockgcp/mockserviceusage/serviceusagev1.go
```diff
@@ -273,6 +275,47 @@ func (s *ServiceUsageV1) ListServices(ctx context.Context, req *pb.ListServicesR
 		})
 	}
 
+	// Implement field filtering
+	// TODO: Can we mame this generic?  Is it used in other APIs?
```
cheftako, line 279: mame? 
make?

mockgcp/mockserviceusage/serviceusagev1.go
```diff
@@ -273,6 +275,47 @@ func (s *ServiceUsageV1) ListServices(ctx context.Context, req *pb.ListServicesR
...
+	{
+		md, _ := metadata.FromIncomingContext(ctx)
+		query := ""
+		if md != nil {
+			for _, v := range md.Get("query") {
+				query = v
```
cheftako, line 285: If query shows up more than once, it might be nice to get a warning. Especially as all but the last will be ignored.

mockgcp/mockserviceusage/serviceusagev1.go
```diff
@@ -273,6 +275,47 @@ func (s *ServiceUsageV1) ListServices(ctx context.Context, req *pb.ListServicesR
...
+			}
+		}
+		fields := ""
+		if query != "" {
+			q, err := url.ParseQuery(query)
```
cheftako, line 290: Would it make sense to try to ParseQuery() before overriding an earlier query string?

cheftako:
/lgtm
/approve
Several nits nut none blocking

mockgcp/mockserviceusage/serviceusagev1.go
```diff
@@ -273,6 +275,47 @@ func (s *ServiceUsageV1) ListServices(ctx context.Context, req *pb.ListServicesR
...
+				return nil, status.Errorf(codes.InvalidArgument, "unable to parse query %q: %v", query, err)
+			}
+			fields = q.Get("fields")
+		}
+		if fields != "" {
+			include := make(map[string]bool)
+			for _, field := range strings.Split(fields, ",") {
+				include[field] = true
+			}
+
```
cheftako: The implementation uses hardcoded field names like `"services/name"`. This is brittle. If the protobuf definition for `Service` changes, this code will need manual updates.

mockgcp/mockserviceusage/serviceusagev1.go
```diff
@@ -273,6 +275,47 @@ func (s *ServiceUsageV1) ListServices(ctx context.Context, req *pb.ListServicesR
...
+			q, err := url.ParseQuery(query)
+			if err != nil {
+				return nil, status.Errorf(codes.InvalidArgument, "unable to parse query %q: %v", query, err)
+			}
+			fields = q.Get("fields")
+		}
+		if fields != "" {
+			include := make(map[string]bool)
+			for _, field := range strings.Split(fields, ",") {
+				include[field] = true
```
cheftako: This implementation only supports a flat list of fields. Real field masks support nesting, e.g., `services(name,parent)`. A simple `strings.Split` won't handle this. While maybe not required for the immediate need, it makes the mock less realistic.

mockgcp/mockserviceusage/serviceusagev1.go
```diff
@@ -273,6 +275,47 @@ func (s *ServiceUsageV1) ListServices(ctx context.Context, req *pb.ListServicesR
...
+		if query != "" {
+			q, err := url.ParseQuery(query)
+			if err != nil {
+				return nil, status.Errorf(codes.InvalidArgument, "unable to parse query %q: %v", query, err)
+			}
+			fields = q.Get("fields")
+		}
+		if fields != "" {
+			include := make(map[string]bool)
+			for _, field := range strings.Split(fields, ",") {
```
cheftako: The logic doesn't handle the case where a parent field is requested (e.g., `fields=services`). In a real API, this would typically include all sub-fields of `services`. Here, it would include nothing because there is no exact match for `"services"`.

mockgcp/mockserviceusage/serviceusagev1.go
justinsb: Will fix!

mockgcp/mockserviceusage/serviceusagev1.go
justinsb: Will make it an error if > 1; https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5826

mockgcp/mockserviceusage/serviceusagev1.go
justinsb: Should be moot if we treat multiple values as an error, I believe

PR #5810, opened 2025-12-11
chore: add fuzz unit test and fix a few fuzz edge cases

justinsb:


pkg/test/fuzz/generate.go
```diff
@@ -67,8 +67,8 @@ func fillWithRandom0(t *testing.T, randStream *rand.Rand, msg protoreflect.Messa
 		field := fields.Get(i)
 
 		if field.IsList() {
-			count := randStream.Intn(10)
-			// Bias heavily to zero
+			count := randStream.Intn(20)
+			// Bias heavily to zero; we don't want to create infinitely large messages
```
cheftako, line 71: Would it make sense to find a more curved distribution system? 0 seems to just not modify the field at all. It would seem like something like 60% 0, 25% 1, 10% 2,  4% v3 and 1% 4, might give better testing while helping with message size?

PR #5814, opened 2025-12-12
mockgcp: tests for mysql 8.4 / c4a family

justinsb:
- **mockgcp: add test for mysql 8.4 / c4a family**
  

- **autogen: golden output for mockgcp mysql84 test**
  

- **mockgcp: more fidelity for mocksql 8.4**


cheftako:
Thanks for adding this test case Justin!

Issue #5778, opened 2025-12-08
Prioritize writing back server-generated resource ID


justinsb:
For server-generated resources, if we don't write back the server-generated ID, we leak the resource.

We should try _very_ hard not to do this, therefore.

1) Should update ID as soon as possible (before waiting for LRO, ideally)

2) Should have some retries / blind writes when updating spec.resourceID and status

2b) If we fail to update spec.resourceID, we should probably also try updating status.

In particular this likely affects TagsTagKey, TagsTagValue, TagsTagBinding

justinsb:
This may also be a good time to figure out how to better update the results of an LRO.  Ideally we would return from the Reconcile method but "know" that we are waiting on an LRO, and reschedule ourselves appropriately.

PR #5815, opened 2025-12-12
feature: enableCiliumClusterwideNetworkPolicy in ContainerCluster

justinsb:
- **terraform: add support for enable_cilium_clusterwide_network_policy to google_container_cluster**
  

- **mockgcp: add support for enableCiliumClusterwideNetworkPolicy to container cluster**
  

- **autogen: generation for enableCiliumClusterwideNetworkPolicy in ContainerCluster**
  

- **tests: verify enableCiliumClusterwideNetworkPolicy in ContainerCluster**
  

- **autogen: golden output for containercluster test**

### DIFF
--- a/pkg/test/resourcefixture/testdata/basic/container/v1beta1/containercluster/containercluster-dataplanev2/_generated_object_containercluster-dataplanev2.golden.yaml
+++ b/pkg/test/resourcefixture/testdata/basic/container/v1beta1/containercluster/containercluster-dataplanev2/_generated_object_containercluster-dataplanev2.golden.yaml
@@ -28,6 +28,7 @@ spec:
       minimum: 100
       resourceType: memory
   datapathProvider: ADVANCED_DATAPATH
+  enableCiliumClusterwideNetworkPolicy: false
   initialNodeCount: 1
   ipAllocationPolicy:
     clusterIpv4CidrBlock: /20
--- a/pkg/test/resourcefixture/testdata/basic/container/v1beta1/containercluster/containercluster-dataplanev2/_http.log
+++ b/pkg/test/resourcefixture/testdata/basic/container/v1beta1/containercluster/containercluster-dataplanev2/_http.log
@@ -376,7 +376,8 @@ User-Agent: google-api-go-client/0.5 kcc/${kccVersion} (+https://github.com/Goog
     "name": "cluster-sample-${uniqueId}",
     "network": "projects/${projectId}/global/networks/${networkID}",
     "networkConfig": {
-      "datapathProvider": "ADVANCED_DATAPATH"
+      "datapathProvider": "ADVANCED_DATAPATH",
+      "enableCiliumClusterwideNetworkPolicy": true
     },
     "networkPolicy": {},
     "nodeConfig": {
@@ -636,6 +637,7 @@ X-Xss-Protection: 0
   "network": "${networkID}",
   "networkConfig": {
     "datapathProvider": "ADVANCED_DATAPATH",
+    "enableCiliumClusterwideNetworkPolicy": true,
     "network": "projects/${projectId}/global/networks/${networkID}",
     "serviceExternalIpsConfig": {},
     "subnetwork": "projects/${projectId}/regions/us-central1/subnetworks/${subnetworkID}"
@@ -995,6 +997,7 @@ X-Xss-Protection: 0
   "network": "${networkID}",
   "networkConfig": {
     "datapathProvider": "ADVANCED_DATAPATH",
+    "enableCiliumClusterwideNetworkPolicy": true,
     "network": "projects/${projectId}/global/networks/${networkID}",
     "serviceExternalIpsConfig": {},
     "subnetwork": "projects/${projectId}/regions/us-central1/subnetworks/${subnetworkID}"
@@ -1354,6 +1357,7 @@ X-Xss-Protection: 0
   "network": "${networkID}",
   "networkConfig": {
     "datapathProvider": "ADVANCED_DATAPATH",
+    "enableCiliumClusterwideNetworkPolicy": true,
     "network": "projects/${projectId}/global/networks/${networkID}",
     "serviceExternalIpsConfig": {},
     "subnetwork": "projects/${projectId}/regions/us-central1/subnetworks/${subnetworkID}"
@@ -1713,6 +1717,7 @@ X-Xss-Protection: 0
   "network": "${networkID}",
   "networkConfig": {
     "datapathProvider": "ADVANCED_DATAPATH",
+    "enableCiliumClusterwideNetworkPolicy": true,
     "network": "projects/${projectId}/global/networks/${networkID}",
     "serviceExternalIpsConfig": {},
     "subnetwork": "projects/${projectId}/regions/us-central1/subnetworks/${subnetworkID}"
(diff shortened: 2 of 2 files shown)


config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_containerclusters.container.cnrm.cloud.google.com.yaml
```diff
@@ -566,6 +566,10 @@ spec:
                   Enable Binary Authorization for this cluster. If enabled, all container
                   images will be validated by Google Binary Authorization.
                 type: boolean
+              enableCiliumClusterwideNetworkPolicy:
+                description: Whether Cilium cluster-wide network policy is enabled
```
cheftako, line 570: AGENT: For better user clarity, please consider enhancing the description to mention that this feature requires the `datapathProvider` to be `ADVANCED_DATAPATH` (Dataplane V2).

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -1923,6 +1923,12 @@ func ResourceContainerCluster() *schema.Resource {
 				Description: `Whether multi-networking is enabled for this cluster.`,
 				Default:     false,
 			},
+			"enable_cilium_clusterwide_network_policy": {
+				Type:        schema.TypeBool,
+				Optional:    true,
```
cheftako, line 1928: AGENT: To ensure consistency and help users, could you update the `Description` here as well to note the dependency on Dataplane V2 (`ADVANCED_DATAPATH`)?

mockgcp/mockcontainer/cluster.go
```diff
@@ -318,6 +318,14 @@ func (s *ClusterManagerV1) UpdateCluster(ctx context.Context, req *pb.UpdateClus
 		update.DesiredDefaultEnablePrivateNodes = nil
 	}
 
+	if update.DesiredEnableCiliumClusterwideNetworkPolicy != nil {
+		if obj.NetworkConfig == nil {
```
cheftako, line 322: AGENT: The mock logic correctly applies the update. To make tests more robust, consider adding validation to the mock that rejects enabling this policy if Dataplane V2 is not configured on the mock cluster object, mirroring the expected behavior of the real API.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -3131,6 +3141,23 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
...
+			Update: &container.ClusterUpdate{
+				DesiredEnableCiliumClusterwideNetworkPolicy: enabled,
+				ForceSendFields:                             []string{"DesiredEnableCiliumClusterwideNetworkPolicy"},
+			},
+		}
+		updateF := updateFunc(req, "updating cilium clusterwide network policy")
+		// Call update serially.
+		if err := transport_tpg.LockedCall(lockKey, updateF); err != nil {
+			return err
+		}
```
cheftako, line 3156: AGENT: The log message is helpful. To improve traceability in environments with many clusters, could you include the project and location of the cluster in this log message?

pkg/test/resourcefixture/testdata/basic/container/v1beta1/containercluster/containercluster-dataplanev2/create.yaml
```diff
@@ -20,6 +20,7 @@ metadata:
   name: cluster-sample-${uniqueId}
 spec:
   datapathProvider: ADVANCED_DATAPATH
+  enableCiliumClusterwideNetworkPolicy: true
```
cheftako, line 23: AGENT: The tests for `true` and `false` states are great. To ensure the default behavior is also covered, please add a test case where `enableCiliumClusterwideNetworkPolicy` is omitted from the manifest to verify it defaults to `false` as intended.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -3131,6 +3141,23 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
 		log.Printf("[INFO] GKE cluster %s L4 ILB Subsetting has been updated to %v", d.Id(), enabled)
 	}
 
+	if d.HasChange("enable_cilium_clusterwide_network_policy") {
+		enabled := d.Get("enable_cilium_clusterwide_network_policy").(bool)
+		req := &container.UpdateClusterRequest{
+			Update: &container.ClusterUpdate{
+				DesiredEnableCiliumClusterwideNetworkPolicy: enabled,
```
cheftako, line 3148: AGENT: Using `ForceSendFields` here is the correct approach for an optional boolean. Does the GKE API guarantee that updates to network configuration are atomic? It's worth confirming that there are no partial update failure modes if other network-related fields were being changed simultaneously.

COUNTER: Given the are an eventually consistent SOT with periodic rereconciliation not sure I am currently worried about the agent's concerns.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -2680,6 +2687,9 @@ func resourceContainerClusterRead(d *schema.ResourceData, meta interface{}) erro
 	if err := d.Set("enable_multi_networking", cluster.NetworkConfig.EnableMultiNetworking); err != nil {
 		return fmt.Errorf("Error setting enable_multi_networking: %s", err)
```
cheftako, line 2688: AGENT: The error message here is a bit generic. Please consider wrapping the error with more context, such as the cluster name, to make debugging easier if this `d.Set` call fails. For example: `fmt.Errorf("Error setting enable_cilium_clusterwide_network_policy for cluster %s: %w", d.Id(), err)`.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -1923,6 +1923,12 @@ func ResourceContainerCluster() *schema.Resource {
 				Description: `Whether multi-networking is enabled for this cluster.`,
 				Default:     false,
 			},
+			"enable_cilium_clusterwide_network_policy": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: `Whether Cilium cluster-wide network policy is enabled on this cluster.`,
```
cheftako, line 1929: AGENT: The `Default: false` is a sensible choice. Was it confirmed that this matches the GKE API's default behavior if the field is omitted entirely from a cluster creation request? A mismatch could lead to unexpected diffs on the first apply.

mockgcp/mockcontainer/cluster.go
```diff
@@ -318,6 +318,14 @@ func (s *ClusterManagerV1) UpdateCluster(ctx context.Context, req *pb.UpdateClus
 		update.DesiredDefaultEnablePrivateNodes = nil
 	}
```
cheftako, line 320: AGENT: The mock correctly creates a `NetworkConfig` if it's `nil`. This is good defensive programming for the mock's internal state.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -3131,6 +3141,23 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
 		log.Printf("[INFO] GKE cluster %s L4 ILB Subsetting has been updated to %v", d.Id(), enabled)
 	}
 
+	if d.HasChange("enable_cilium_clusterwide_network_policy") {
```
cheftako, line 3144: CONCERN: Are we recording in the status that we have begun reconciliation? Do partial updates and waiting/erroring out concerns me if we do not record that we have begun processing.

cheftako (commented):
AGERNT: This PR adds support for `enable_cilium_clusterwide_network_policy` in `ContainerCluster`. The changes are well-structured, consistent with the existing codebase, and include necessary test coverage for create and update scenarios. A few minor points could be addressed to improve the implementation, specifically enhancing the descriptions in the CRD and Terraform schema for better user clarity and refining the mock server logic to be more representative of actual GKE API validations.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_containerclusters.container.cnrm.cloud.google.com.yaml
justinsb: I believe this is inherited from the terraform provider code, and that is inherited from the google apis.  I don't think we want to be in the business of "fixing" GCP APIs, even if this is a valid observation.

mockgcp/mockcontainer/cluster.go
justinsb: That's reasonable, I'll add it to the list of mockgcp gaps for container clusters!

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
justinsb: We should tell the agent to not suggest as many changes in vendored code, where it is important that we stick close to upstream, for better or worse

pkg/test/resourcefixture/testdata/basic/container/v1beta1/containercluster/containercluster-dataplanev2/create.yaml
justinsb: I don't think we want to aim for full test coverage, it will be unmanageable.  That said, this case might be worthwhile.  I'm wondering if we should do an update fuzzer.

The fuzzer should work on the principle of path invariance: no matter what sequence of changes we made to get to a particular spec, if the status is ready then the state of the resource in GCP should be the same (after normalization)

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
justinsb: We need to dial down the agent on vendored code where we don't set the quality bar.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
justinsb: vendored

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
justinsb: vendored

mockgcp/mockcontainer/cluster.go
justinsb: Thanks!  But maybe we should also tell the agent not to worry as much about mockgcp, our standard for mockgcp should be that it gives the same output as realgcp (as determined by the overnight tests)

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
justinsb: We should, and I don't think we are for direct controllers (where we would fix it).  I added a note to https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/5778#issuecomment-3670993505

pkg/test/resourcefixture/testdata/basic/container/v1beta1/containercluster/containercluster-dataplanev2/create.yaml
justinsb: Captured this in https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/5846

PR #5818, opened 2025-12-12
docs: park current state of MCL

acpana:
Parking the current state of the multi cluster support in KCC. Things may change (rapidly) down the road!

### DIFF
--- a/docs/features/index.md
+++ b/docs/features/index.md
@@ -2,3 +2,4 @@
 
 * [Opt in to enable experimental direct controllers](./optin.md)
 * [Pause actuation of resources onto the cloud provider](./pause.md)
+* [Alpha: Multi Cluster KCC](./mcl.md)
--- a/docs/features/mcl.md
+++ b/docs/features/mcl.md
@@ -1,0 +1,85 @@
+# Setting up multi cluster support for KCC
+
+> NOTE: This feature is in alpha stage and is actively being worked on.
+
+Imagine we have two Kubernetes clusters, `mcl-cluster-1` and `mcl-cluster-2`. We want to run KCC in both for high availability, but only one instance should be the active leader at any time.
+
+### Setup
+
+#### 1. Cluster Lock: Powered GCS Bucket
+
+```sh
+# This is a one-time setup
+gcloud storage buckets create gs://kcc-global-leader-lock
+```
+
+#### 2. The `multiclusterlease-controller` (MCL Controller)
+
+Deploy the MCL controller to **both** `mcl-cluster-1` and `mcl-cluster-2`. This controller's job is to watch for `MultiClusterLease` CRs and contend for the lock in the GCS bucket.
+
+```yaml
+# Deployed in BOTH mcl-cluster-1 and mcl-cluster-2
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: multiclusterlease-controller-manager
+  namespace: multiclusterlease-system
+spec:
+  replicas: 1
+  template:
+    spec:
+      # This controller needs GCP credentials to talk to the bucket.
+      # This is best done using Workload Identity, which binds the Kubernetes
+      # Service Account (KSA) used by this pod to a Google Service Account (GSA).
+      # Ensure the KSA (e.g., 'default' in 'multiclusterlease-system') is
+      # configured for Workload Identity.
+      serviceAccountName: default
+      containers:
+      - name: manager
+        image: gcr.io/multiclusterlease/controller:latest
+        command:
+        - /manager
+        args:
+        - --gcs-bucket=kcc-global-leader-lock
+```
+
+At this point, we have an election controller running in each cluster, ready to act on behalf of any client that creates a `MultiClusterLease` resource.
+
+#### 3. KCC integration & E2E Walkthrough
+
+Now, let's set up the lease objects for when the two KCC pods start up in `mcl-cluster-1` and `mcl-cluster-2`.
+In each of your clusters, configure your `ConfigConnector` objects such that you specify the name, namespace
+where you want the lease objects to be created, and the name of the global lock you set up before. Example:
+
+```yaml
+apiVersion: core.cnrm.cloud.google.com/v1beta1
+kind: ConfigConnector
+metadata:
+  name: configconnector.core.cnrm.cloud.google.com
+spec:
+  mode: cluster
+  experiments:
+    leaderElection:
+      multiClusterLease:
+        leaseName: kcc-mcl-cluster-1-lease # kcc-mcl-cluster-2-lease for your other cluster
+        namespace: kcc-system
+        globalLockName: kcc-global-leader-lock
+```
+
+At this point, you will also need to create the MultiClusterLease CRs for now. In the future, the KCC manager or
+the operator may manage the lifecycle of these resources.
+
+```yaml
+# In mcl-cluster-1
+apiVersion: multicluster.core.cnrm.cloud.google.com/v1alpha1
+kind: MultiClusterLease
+metadata:
+  name:  kcc-mcl-cluster-1-lease
+  namespace: kcc-system
+spec:
+  leaseDurationSeconds: 60
+status: {} # Status is empty for now
+```
+
+At this point when the KCC manager pods start, they would start with leader election configured. One pod will
+be elected leader and the other will exit. 


docs/features/mcl.md
```diff
@@ -0,0 +1,85 @@
...
+metadata:
+  name: configconnector.core.cnrm.cloud.google.com
+spec:
+  mode: cluster
+  experiments:
+    leaderElection:
+      multiClusterLease:
+        leaseName: kcc-mcl-cluster-1-lease # kcc-mcl-cluster-2-lease for your other cluster
+        namespace: kcc-system
+        globalLockName: kcc-global-leader-lock
```
acpana, line 66: from offline conversations, we don't want to expose this. Rather have the kccmanager fetch the lock name from the spec of the MCL!

docs/features/index.md
```diff
@@ -1,4 +1,5 @@
 # Contents
 
 * [Opt in to enable experimental direct controllers](./optin.md)
-* [Pause actuation of resources onto the cloud provider](./pause.md)
\ No newline at end of file
+* [Pause actuation of resources onto the cloud provider](./pause.md)
```
cheftako, line 4: cloud provider seems unnecessarily indirect. 'Pause actuation of GCP resources' seems more clear and succinct.

docs/features/mcl.md
```diff
@@ -0,0 +1,85 @@
+# Setting up multi cluster support for KCC
+
+> NOTE: This feature is in alpha stage and is actively being worked on.
+
+Imagine we have two Kubernetes clusters, `mcl-cluster-1` and `mcl-cluster-2`. We want to run KCC in both for high availability, but only one instance should be the active leader at any time.
```
cheftako, line 5: Imagine?

For this example we have two...

cheftako: in both ?what? 
I assume clusters.

cheftako: instance of ?what? 
Its ok to repeat the word cluster.

docs/features/mcl.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Setup
```
cheftako, line 7: Transition. You've described what you want and are now giving instructions. I can assume the are to do what you described but please provide than transition.

cheftako, line 7: Transition. You've described what you want and are now giving instructions. I can assume the are to do what you described but please provide than transition.

docs/features/mcl.md
```diff
@@ -0,0 +1,85 @@
...
+
+#### 1. Cluster Lock: Powered GCS Bucket
+
```
cheftako, line 10: Please provide some context. You are creating the GCS Bucket to act as the backing store for the state of the leader election. This is a pluggable component ....

docs/features/mcl.md
```diff
@@ -0,0 +1,85 @@
...
+```sh
+# This is a one-time setup
+gcloud storage buckets create gs://kcc-global-leader-lock
+```
+
+#### 2. The `multiclusterlease-controller` (MCL Controller)
+
+Deploy the MCL controller to **both** `mcl-cluster-1` and `mcl-cluster-2`. This controller's job is to watch for `MultiClusterLease` CRs and contend for the lock in the GCS bucket.
```
cheftako, line 18: the GCS bucket named 'kcc-global-leader-lock'.

docs/features/mcl.md
```diff
@@ -0,0 +1,85 @@
...
+      containers:
+      - name: manager
+        image: gcr.io/multiclusterlease/controller:latest
+        command:
+        - /manager
+        args:
+        - --gcs-bucket=kcc-global-leader-lock
+```
+
+At this point, we have an election controller running in each cluster, ready to act on behalf of any client that creates a `MultiClusterLease` resource.
```
cheftako, line 46: Where does the 'election controller' come in to this. I couldn't find a definition of it any earlier in the doc. Seems like you mean the 'multiclusterlease-controller' but why change the name?

docs/features/mcl.md
```diff
@@ -0,0 +1,85 @@
...
+
+#### 3. KCC integration & E2E Walkthrough
+
+Now, let's set up the lease objects for when the two KCC pods start up in `mcl-cluster-1` and `mcl-cluster-2`.
```
cheftako, line 50: Pedantic: You are not configuring/setting up the lease objects. You are configuring the service to use the lease objects.

docs/features/mcl.md
```diff
@@ -0,0 +1,85 @@
...
+  mode: cluster
+  experiments:
+    leaderElection:
+      multiClusterLease:
+        leaseName: kcc-mcl-cluster-1-lease # kcc-mcl-cluster-2-lease for your other cluster
+        namespace: kcc-system
+        globalLockName: kcc-global-leader-lock
+```
+
+At this point, you will also need to create the MultiClusterLease CRs for now. In the future, the KCC manager or
```
cheftako, line 69: Now you seem to be setting up the lease object.

cheftako: Also lose the "At this point".

docs/features/mcl.md
```diff
@@ -0,0 +1,85 @@
...
+kind: MultiClusterLease
+metadata:
+  name:  kcc-mcl-cluster-1-lease
+  namespace: kcc-system
+spec:
+  leaseDurationSeconds: 60
+status: {} # Status is empty for now
+```
+
+At this point when the KCC manager pods start, they would start with leader election configured. One pod will
```
cheftako, line 84: Ditto

cheftako: The other should not exit but go into stand-by mode.

acpana: the hot stand by should not be bouncing; a stack dump would indicate it's trying to acquire the lock

docs/features/mcl.md
```diff
@@ -0,0 +1,85 @@
...
+  namespace: multiclusterlease-system
+spec:
+  replicas: 1
+  template:
+    spec:
+      # This controller needs GCP credentials to talk to the bucket.
+      # This is best done using Workload Identity, which binds the Kubernetes
+      # Service Account (KSA) used by this pod to a Google Service Account (GSA).
+      # Ensure the KSA (e.g., 'default' in 'multiclusterlease-system') is
+      # configured for Workload Identity.
```
cheftako, line 35: AGENT: Using the `:latest` tag for container images is generally discouraged in production environments as it can lead to unpredictable behavior when the image is updated. It would be better to use a specific version tag to ensure deterministic deployments.

TAKO: +1

docs/features/mcl.md
cheftako, line 5: AGENT: The phrasing "Imagine we have" is a bit too informal for technical documentation. Consider rephrasing to something more direct, for example: "This guide describes a high-availability setup for KCC across two Kubernetes clusters, `mcl-cluster-1` and `mcl-cluster-2`."

docs/features/mcl.md
cheftako, line 18: AGENT: The name `multiclusterlease-controller` is used here, but the deployment metadata name is `multiclusterlease-controller-manager`. While the controller binary might be different, this could be confusing. It would be good to maintain consistency in naming or clarify the distinction.

docs/features/mcl.md
```diff
@@ -0,0 +1,85 @@
...
+  experiments:
+    leaderElection:
+      multiClusterLease:
+        leaseName: kcc-mcl-cluster-1-lease # kcc-mcl-cluster-2-lease for your other cluster
+        namespace: kcc-system
+        globalLockName: kcc-global-leader-lock
+```
+
+At this point, you will also need to create the MultiClusterLease CRs for now. In the future, the KCC manager or
+the operator may manage the lifecycle of these resources.
```
cheftako, line 70: AGENT: "for now" sounds temporary. It would be better to state the current requirement clearly. Suggestion: "You must create the `MultiClusterLease` CR manually. In the future, the KCC manager or the operator may manage the lifecycle of these resources."

docs/features/mcl.md
```diff
@@ -0,0 +1,85 @@
...
+metadata:
+  name:  kcc-mcl-cluster-1-lease
+  namespace: kcc-system
+spec:
+  leaseDurationSeconds: 60
+status: {} # Status is empty for now
+```
+
+At this point when the KCC manager pods start, they would start with leader election configured. One pod will
+be elected leader and the other will exit.
```
cheftako, line 85: AGENT: The documentation ends abruptly. It would be helpful to describe the expected outcome more completely. What happens to the pod that doesn't acquire the lease? Does it exit, or does it stay in a standby mode? Clarifying this would give users a better understanding of the system's behavior.

docs/features/mcl.md
```diff
@@ -0,0 +1,85 @@
...
+> NOTE: This feature is in alpha stage and is actively being worked on.
+
+Imagine we have two Kubernetes clusters, `mcl-cluster-1` and `mcl-cluster-2`. We want to run KCC in both for high availability, but only one instance should be the active leader at any time.
+
+### Setup
+
+#### 1. Cluster Lock: Powered GCS Bucket
+
+```sh
+# This is a one-time setup
```
cheftako, line 12: AGENT: The documentation should specify the required IAM permissions for the Google Service Account that will be used by the `multiclusterlease-controller`. Without this information, users will have to guess the necessary permissions, which could lead to overly permissive setups. Please add a section on the required IAM roles (e.g., `roles/storage.objectAdmin` on the bucket).

docs/features/mcl.md
```diff
@@ -0,0 +1,85 @@
...
+  name: multiclusterlease-controller-manager
+  namespace: multiclusterlease-system
+spec:
+  replicas: 1
+  template:
+    spec:
+      # This controller needs GCP credentials to talk to the bucket.
+      # This is best done using Workload Identity, which binds the Kubernetes
+      # Service Account (KSA) used by this pod to a Google Service Account (GSA).
+      # Ensure the KSA (e.g., 'default' in 'multiclusterlease-system') is
```
cheftako, line 34: AGENT: Using the `default` service account is not a recommended security practice. It's better to create a dedicated Kubernetes Service Account for the `multiclusterlease-controller-manager` deployment to ensure it has only the permissions it needs. The documentation should be updated to reflect this best practice.

docs/features/mcl.md
```diff
@@ -0,0 +1,85 @@
...
+      # configured for Workload Identity.
+      serviceAccountName: default
+      containers:
+      - name: manager
+        image: gcr.io/multiclusterlease/controller:latest
+        command:
+        - /manager
+        args:
+        - --gcs-bucket=kcc-global-leader-lock
+```
```
cheftako, line 44: AGENT: The `--gcs-bucket` argument hardcodes the bucket name `kcc-global-leader-lock`. If a user creates a bucket with a different name, this example will fail. It would be clearer to use a placeholder like `[YOUR_GCS_BUCKET_NAME]` and explicitly instruct the user to replace it with the name of the bucket they created in the first step.

docs/features/mcl.md
```diff
@@ -0,0 +1,85 @@
...
+kind: ConfigConnector
+metadata:
+  name: configconnector.core.cnrm.cloud.google.com
+spec:
+  mode: cluster
+  experiments:
+    leaderElection:
+      multiClusterLease:
+        leaseName: kcc-mcl-cluster-1-lease # kcc-mcl-cluster-2-lease for your other cluster
+        namespace: kcc-system
```
cheftako, line 65: AGENT: While the comment `# kcc-mcl-cluster-2-lease for your other cluster` is helpful, the documentation would be more complete and less error-prone if it included the full YAML configuration example for the `ConfigConnector` resource in the second cluster.

docs/features/mcl.md
cheftako, line 70: AGENT: The user is instructed to create a `MultiClusterLease` CR, but it's not clear how the `MultiClusterLease` CRD is installed in the cluster. Please add instructions on how to install the CRD or clarify if it's bundled with the `multiclusterlease-controller`.

docs/features/mcl.md
```diff
@@ -0,0 +1,85 @@
...
+
+```yaml
+# In mcl-cluster-1
+apiVersion: multicluster.core.cnrm.cloud.google.com/v1alpha1
+kind: MultiClusterLease
+metadata:
+  name:  kcc-mcl-cluster-1-lease
+  namespace: kcc-system
+spec:
+  leaseDurationSeconds: 60
```
cheftako, line 80: AGENT: The comment `# Status is empty for now` is noted. However, it would be beneficial for users to understand what to expect from the `status` field in the future. Providing a brief description of the information it will eventually contain (e.g., current lease holder, lease expiration time) would add valuable context.

docs/features/mcl.md
gemmahou: Noticed that we have a refactor in #5827 so this example yaml is out-dated. Let's make sure we update it before merging!

PR #5821, opened 2025-12-15
chore: don't generate test for apiextensions.k8s.io group

justinsb:


dev/tasks/generate-ci-cd-jobs
```diff
@@ -27,9 +27,13 @@ for group in $(rgrep --no-filename '^[ ]*group:' config/crds | sort | uniq | awk
   if [[ "${group}" == "" ]]; then
     continue
   fi
-  
+  if [[ "${group}" == "apiextensions.k8s.io" ]]; then
```
cheftako: `if [[ "${group}" != *.cnrm.cloud.google.com ]] ; then` would more closely match the statement in the echo.

justinsb: Good shout, will fix.

PR #5827, opened 2025-12-16
refactor: rename field 

acpana:
Renaming the field from `globalLockName` (which is a lie, that's not what the field is), to the candidate identity field!


cheftako:
Change should be safe as this experiment has not been rolled out yet. However shortly we will need to stop making these sort of changes to it.

PR #5829, opened 2025-12-16
refactor: internal test config

acpana:
Follow up on https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5797#discussion_r2615936392

### DIFF
--- a/pkg/controller/kccmanager/kccmanager.go
+++ b/pkg/controller/kccmanager/kccmanager.go
@@ -100,14 +100,23 @@ type Config struct {
 	// Configure manager to participate in leader election if MultiClusterLease is enabled.
 	MultiClusterLease bool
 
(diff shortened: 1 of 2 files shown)


pkg/controller/kccmanager/kccmanager.go
```diff
@@ -100,14 +100,22 @@ type Config struct {
 	// Configure manager to participate in leader election if MultiClusterLease is enabled.
 	MultiClusterLease bool
 
-	// SkipControllerRegistration is true if we should skip registering the default controllers
+	testConfig
```
maqiuyujoyce, line 104: Could you add a comment saying this is used for smoke test only (if we don't also use it for unit testing)?

Also do you think we will do further refactoring in the future to make controllerRegistration and multiClusterLeaseConfig configurable in general?

PR #5809, opened 2025-12-11
Foundational API and Scaffolding for Parameter Manager Parameter

dhavalbhensdadiya-crest:
### BRIEF Change description

This pull request is PR#1 for the support of new resource parameter manager parameter. It is created on the basis of the steps documented [here](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/develop-resources/ai/scenarios/new-resource.md)

Fixes #
This is for ticket - https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/5785. However, not adding it to fixes as I assume it can be closed once the resource is upgraded to beta.
#### WHY do we need this change?
This is for adding support of new resource: Parameter Manager - Parameter
#### Special notes for your reviewer:
I have updated the documentation because it has come to my attention that fuzzer tests are not executed if an entry is not made in the `register.go` file.
#### Does this PR add something which needs to be 'release noted'?

```release-note

```

- [ ] Reviewer reviewed release note.

#### Additional documentation e.g., references, usage docs, etc.:

```docs

```

#### Intended Milestone

Please indicate the intended milestone. 
- [ ] Reviewer tagged PR with the actual milestone.

### Tests you have done

- [x] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #5833, opened 2025-12-17
Added mockgcp server for Parameter Manager Parameter resource

dhavalbhensdadiya-crest:
### BRIEF Change description
This PR adds the support of mock gcp server for parameter manager. I have also added mock CRUD functions for parameter resource

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/mockgcp/mockparametermanager/parameter.go
+++ b/mockgcp/mockparametermanager/parameter.go
@@ -62,6 +62,7 @@ func (s *ParameterManagerV1) CreateParameter(ctx context.Context, req *pb.Create
 		return nil, err
 	}
 
+	obj.Labels = nil // Post response does not include labels
 	return obj, nil
 }
 
@@ -123,6 +124,7 @@ func (s *ParameterManagerV1) UpdateParameter(ctx context.Context, req *pb.Update
 	if err := s.storage.Update(ctx, fqn, updated); err != nil {
 		return nil, err
 	}
+	updated.Labels = nil // Patch response does not include labels
 	return updated, nil
 }
 


maqiuyujoyce:
This PR shouldn't need to be dependent on the #5809. Do you mind cleaning up the relevant commits?

Besides, you'll want to rebase against master to solve some conflicts.

dhavalbhensdadiya-crest:
> This PR shouldn't need to be dependent on the #5809. Do you mind cleaning up the relevant commits?
> 
> Besides, you'll want to rebase against master to solve some conflicts.

Hi @maqiuyujoyce ,
I have rebased the PR to only have the relevant commits and resolved the merge conflicts. Thanks

mockgcp/mockparametermanager/parameter.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		return nil, err
+	}
+	fqn := name.String()
+	existing := &pb.Parameter{}
+	if err := s.storage.Get(ctx, fqn, existing); err != nil {
+		return nil, err
+	}
+
+	updated := ProtoClone(existing)
```
maqiuyujoyce: You should update `UpdateTime` here to reflect the modification time.
```go
updated.UpdateTime = timestamppb.Now()
```

mockgcp/mockparametermanager/parameter.go
```diff
@@ -0,0 +1,206 @@
...
+			return nil, status.Errorf(codes.NotFound, "Parameter [%s] not found.", fqn)
+		}
+		return nil, err
+	}
+
+	return &parameter, nil
+}
+
+// ProtoClone is a type-safe wrapper around proto.Clone
+func ProtoClone[T proto.Message](t T) T {
```
maqiuyujoyce: Please remove it and use the standard pattern proto.Clone(existing).(*pb.Parameter) inline.

This generic `ProtoClone` helper seems redundant if it's only used in this package. `proto.Clone` is sufficient, or this should be in a common utility package if used widely.

mockgcp/mockparametermanager/parameter.go
```diff
@@ -0,0 +1,206 @@
...
+
+		name := &parameterName{
+			Project:       project,
+			Location:      tokens[3],
+			ParameterName: tokens[5],
+		}
+
+		return name, nil
+	} else {
+		return nil, status.Errorf(codes.InvalidArgument, "name %q is not valid", name)
```
maqiuyujoyce: Nit: The error message could be more descriptive about *why* it is invalid (e.g. "expected format: projects/.../locations/.../parameters/...").

mockgcp/mockparametermanager/parameter.go
```diff
@@ -0,0 +1,206 @@
...
+	// obj.CreateTime = timestamppb.Now()
+	if err := s.storage.Create(ctx, fqn, obj); err != nil {
+		return nil, err
+	}
+
+	return obj, nil
+}
+
+// Gets metadata for a given [Parameter][google.cloud.parametermanager.v1.Parameter].
+func (s *ParameterManagerV1) GetParameter(ctx context.Context, req *pb.GetParameterRequest) (*pb.Parameter, error) {
```
maqiuyujoyce: In `GetParameter`, you parse the name but don't seem to use the parsed project/location to validate they exist (mock-wise).
This is fine as `s.storage.Get(ctx, fqn, ...)` handles the retrieval by key.

mockgcp/mockparametermanager/service.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (s *MockService) ExpectedHosts() []string {
+	return []string{"parametermanager.googleapis.com"}
+}
+
+func (s *MockService) Register(grpcServer *grpc.Server) {
+	pb.RegisterParameterManagerServer(grpcServer, s.v1)
+}
+
+func (s *MockService) NewHTTPMux(ctx context.Context, conn *grpc.ClientConn) (http.Handler, error) {
```
maqiuyujoyce, line 60: The service is not registered with the `mockgcp` framework. You need to add an `init()` function to register it.

```go
func init() {
    mockgcpregistry.Register(New)
}
```
For example: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/mockgcp/mockalloydb/service.go#L32

mockgcp/mockparametermanager/parameter.go
```diff
@@ -0,0 +1,206 @@
...
+	name := parameterName{
+		Project:       project,
+		Location:      location,
+		ParameterName: parameterID,
+	}
+	fqn := name.String()
+
+	obj := proto.Clone(req.Parameter).(*pb.Parameter)
+	obj.Name = fqn
+	// obj.CreateTime = timestamppb.Now()
```
maqiuyujoyce: Please uncomment and implement `CreateTime`.
Controllers often rely on this field.
You will need to import `google.golang.org/protobuf/types/known/timestamppb` and use `timestamppb.Now()`.

mockgcp/mockparametermanager/parameter.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	updated := ProtoClone(existing)
+	updated.Name = name.String()
+
+	// Required. The update mask applies to the resource.
+	paths := req.GetUpdateMask().GetPaths()
+	if len(paths) == 0 {
+		return nil, status.Errorf(codes.InvalidArgument, "update_mask is required")
+	}
```
maqiuyujoyce, line 108: Could you ensure the update of kmsKey is covered in testdata?

If the request comes via gRPC (which mockgcp serves), the FieldMask paths are typically the proto field names (snake_case), i.e., `kms_key`.

However, mockgcp also serves HTTP requests via the gRPC-Gateway. The Gateway might transform JSON (camelCase) to Proto (snake_case) or vice-versa.

So we need to have the test to verify which is the right pattern for field path.

mockgcp/mockparametermanager/parameter.go
maqiuyujoyce: The `UpdateTime` must be updated here to reflect the modification.
```go
updated.UpdateTime = timestamppb.Now()
```

mockgcp/mockparametermanager/parameter.go
maqiuyujoyce: This generic `ProtoClone` helper seems unnecessary as `proto.Clone` is standard.
You can simply use `proto.Clone(existing).(*pb.Parameter)` inline or just keep the type assertion.

mockgcp/mockparametermanager/parameter.go
```diff
@@ -0,0 +1,206 @@
...
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/common/projects"
+	pb "github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/generated/mockgcp/cloud/parametermanager/v1"
+)
+
+type ParameterManagerV1 struct {
+	*MockService
+
+	pb.UnimplementedParameterManagerServer
+}
```
maqiuyujoyce: This `parseKey` method is defined on `*MockService` but does not use the receiver `s`.
It should probably be a standalone helper function or a private function in the package.

mockgcp/mockparametermanager/parameter.go
```diff
@@ -0,0 +1,206 @@
...
+
+func (s *MockService) parseKey(key string, input string) (string, error) {
```
maqiuyujoyce: The error return here is always `nil`. Either implement actual error checking (e.g., if key not found) or remove the error return.
Currently, it swallows potential parsing issues silently.

mockgcp/mockparametermanager/parameter.go
```diff
@@ -0,0 +1,206 @@
...
+	existing := &pb.Parameter{}
+	if err := s.storage.Get(ctx, fqn, existing); err != nil {
+		return nil, err
+	}
+
+	updated := ProtoClone(existing)
+	updated.Name = name.String()
+
+	// Required. The update mask applies to the resource.
+	paths := req.GetUpdateMask().GetPaths()
```
maqiuyujoyce: According to AIP-134, if `update_mask` is empty (and the verb is Update), it should typically be treated as a full replacement of all supported fields.
Returning `InvalidArgument` is non-standard unless this specific API strictly requires a mask.
Consider supporting empty mask or confirming this behavior.

mockgcp/mockparametermanager/parameter.go
maqiuyujoyce: The error message `name %q is not valid` is a bit generic.
It would be helpful to include the expected format in the error message, e.g., "expected projects/.../locations/.../parameters/...".

mockgcp/mockparametermanager/parameter.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+func (n *parameterName) String() string {
+	return "projects/" + n.Project.ID + "/locations/" + n.Location + "/parameters/" + n.ParameterName
+}
+
+// parseParameterName parses a string into a parameterName.
+// The expected form is projects/<projectID>/locations/<location>/parameters/<parameterName>
+func (s *MockService) parseParameterName(name string) (*parameterName, error) {
+	tokens := strings.Split(name, "/")
```
maqiuyujoyce: The hardcoded indices (`tokens[0]`, `tokens[2]`) assume a specific path length and structure.
While correct for the expected valid path, using a `path.Match` or `regexp` approach might be more robust and readable for resource name parsing.

mockgcp/mockparametermanager/parameter.go
```diff
@@ -0,0 +1,206 @@
...
+	return &parameter, nil
+}
+
+// ProtoClone is a type-safe wrapper around proto.Clone
+func ProtoClone[T proto.Message](t T) T {
+	return proto.Clone(t).(T)
+}
+
+// Update metadata for a given [Parameter][google.cloud.parametermanager.v1.Parameter].
+func (s *ParameterManagerV1) UpdateParameter(ctx context.Context, req *pb.UpdateParameterRequest) (*pb.Parameter, error) {
```
maqiuyujoyce: Please add a check for `req.Parameter == nil` here as well to avoid panic when accessing `req.Parameter.Name` or `req.Parameter.GetLabels()`.

mockgcp/mockparametermanager/parameter.go
```diff
@@ -0,0 +1,206 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mockparametermanager
+
+import (
+	"context"
+	"strings"
+
```
maqiuyujoyce, line 20: You need to add `"google.golang.org/protobuf/types/known/timestamppb"` to the imports to support `CreateTime` and `UpdateTime`.

mockgcp/mockparametermanager/parameter.go
```diff
@@ -0,0 +1,206 @@
...
+	// Required. The update mask applies to the resource.
+	paths := req.GetUpdateMask().GetPaths()
+	if len(paths) == 0 {
+		return nil, status.Errorf(codes.InvalidArgument, "update_mask is required")
+	}
+	for _, path := range paths {
+		switch path {
+		case "kmsKey":
+			if kmsKey := req.Parameter.GetKmsKey(); kmsKey != "" {
+				updated.KmsKey = &kmsKey
```
maqiuyujoyce: If `req.Parameter.GetKmsKey()` returns an empty string (default), this logic sets `updated.KmsKey` to `nil`.
If the user explicitly sent `kmsKey: ""` intending to clear it, this is correct.
However, if they sent nothing (nil) but included `kmsKey` in the mask, `GetKmsKey()` also returns `""`.
This ambiguity is inherent in `Get*` accessors for optional fields, but combined with the mask, the logic `!= ""` implies we can never set the field to explicit empty string. Verify if empty string is a valid value for `KmsKey`.

mockgcp/mockparametermanager/parameter.go
```diff
@@ -0,0 +1,206 @@
...
+			} else {
+				updated.KmsKey = nil
+			}
+		case "labels":
+			updated.Labels = req.Parameter.GetLabels()
+		default:
+			return nil, status.Errorf(codes.InvalidArgument, "update_mask path %q not valid", path)
```
maqiuyujoyce: Does `Parameter` resource have a `description` field?
If so, it should probably be handled here or explicitly ignored.
If `kmsKey` and `labels` are the only mutable fields, this is correct.

mockgcp/mockparametermanager/parameter.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+// Gets metadata for a given [Parameter][google.cloud.parametermanager.v1.Parameter].
+func (s *ParameterManagerV1) GetParameter(ctx context.Context, req *pb.GetParameterRequest) (*pb.Parameter, error) {
+	name, err := s.parseParameterName(req.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	var parameter pb.Parameter
```
maqiuyujoyce: If `s.storage.Get` returns an error other than `NotFound` (e.g., internal error), this block returns it directly.
Ensure that `s.storage` returns errors compatible with gRPC status codes, or wrap them here to ensure the client receives a proper gRPC error code.

mockgcp/mockparametermanager/parameter.go
```diff
@@ -0,0 +1,206 @@
...
+	return updated, nil
+}
+
+// Deletes a [Parameter][google.cloud.parametermanager.v1.Parameter].
+func (s *ParameterManagerV1) DeleteParameter(ctx context.Context, req *pb.DeleteParameterRequest) (*emptypb.Empty, error) {
+	name, err := s.parseParameterName(req.Name)
+	if err != nil {
+		return nil, err
+	}
+
```
maqiuyujoyce: In `DeleteParameter`, if `s.storage.Delete` fails because the resource is not found, `Delete` should typically return `NotFound` (AIP-135).
Ensure `s.storage.Delete` returns a `NotFound` error in that case, or handle it here.

mockgcp/register.go
```diff
@@ -46,6 +46,7 @@ import (
 	_ "github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/mocknetworksecurity"
 	_ "github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/mocknetworkservices"
 	_ "github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/mockorgpolicy"
+	_ "github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/mockparametermanager"
```
maqiuyujoyce: The import is present, which is good.
However, it will only be effective once the `init()` function is added to `mockparametermanager/service.go`.

mockgcp/mockparametermanager/parameter.go
maqiuyujoyce: I think we don't need this function. Please see my other comment under CreateParameter(). 

This parseKey implementation is brittle because it searches for the key anywhere in the path string, potentially leading to incorrect matches if resource IDs contain the key name. It also fails to enforce the expected hierarchical structure of the resource name (e.g., that projects is the first segment).

Since req.Parent for CreateParameter has a well-defined format (`projects/*/locations/*`), you should parse it explicitly like you do in `parseParameterName`, or create a helper that strictly validates the indices.

mockgcp/mockparametermanager/parameter.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+// Creates a new [Parameter][google.cloud.parametermanager.v1.Parameter].
+func (s *ParameterManagerV1) CreateParameter(ctx context.Context, req *pb.CreateParameterRequest) (*pb.Parameter, error) {
+	parameterID := req.ParameterId
+	if parameterID == "" {
+		return nil, status.Errorf(codes.InvalidArgument, "ParameterId is required")
+	}
+
+	projectStr, err := s.parseKey("projects", req.Parent)
```
maqiuyujoyce: The usage of `s.parseKey` to extract `project` and `location` from `req.Parent` is unsafe. `parseKey` performs loose matching that can be fooled by resource IDs containing segment names (e.g. `projects`) and ignores the expected resource hierarchy.

Please replace the calls to `parseKey` with a strict parsing helper similar to `parseParameterName`.

Specifically, you should parse `req.Parent` by splitting the string and validating the indices:
```
  tokens := strings.Split(req.Parent, "/")
  if len(tokens) == 4 && tokens[0] == "projects" && tokens[2] == "locations" {
      projectID = tokens[1]
      location = tokens[3]
  } else {
      return nil, status.Errorf(codes.InvalidArgument, "parent %q is not valid", req.Parent)
  }
```

This ensures we validate the parent format `projects/<project>/locations/<location>` correctly and consistently with other mock services.

mockgcp/mockparametermanager/parameter.go
maqiuyujoyce: You should also set `UpdateTime` to the same value as `CreateTime` during creation.

mockgcp/mockparametermanager/service.go
dhavalbhensdadiya-crest: Implemented similar to mockalloydb. My previous implementation had an entry in mock_http_roundtrip.go incorporated in PR3.

mockgcp/mockparametermanager/parameter.go
dhavalbhensdadiya-crest: Done

mockgcp/mockparametermanager/parameter.go
dhavalbhensdadiya-crest: Yes. It is covered.

mockgcp/mockparametermanager/parameter.go
dhavalbhensdadiya-crest: Done

mockgcp/mockparametermanager/parameter.go
dhavalbhensdadiya-crest: Done

mockgcp/mockparametermanager/parameter.go
dhavalbhensdadiya-crest: Removed

mockgcp/mockparametermanager/parameter.go
dhavalbhensdadiya-crest: Replaced. Reused parseParameterName

mockgcp/mockparametermanager/parameter.go
dhavalbhensdadiya-crest: Done

dhavalbhensdadiya-crest: Done

mockgcp/mockparametermanager/parameter.go
dhavalbhensdadiya-crest: Used generic

maqiuyujoyce:
Looks like test-mockgcp failed with the following error:
```
2026-02-04T21:56:33.4432437Z === NAME  TestScripts/mockparametermanager/testdata/parameter/crud
2026-02-04T21:56:33.4433160Z     utils.go:150: FAIL: unexpected diff in /home/runner/work/k8s-config-connector/k8s-config-connector/mockgcp/mockparametermanager/testdata/parameter/crud/_http.log:   (
2026-02-04T21:56:33.4433265Z           	"""
2026-02-04T21:56:33.4433482Z           	... // 26 identical lines
2026-02-04T21:56:33.4433641Z           	  "format": "JSON",
2026-02-04T21:56:33.4434358Z           	  "kmsKey": "projects/${projectId}/locations/global/keyRings/test-keyring-${uniqueId}/cryptoKeys/test-key-${uniqueId}",
2026-02-04T21:56:33.4434475Z         + 	  "labels": {
2026-02-04T21:56:33.4434608Z         + 	    "key": "value"
2026-02-04T21:56:33.4434701Z         + 	  },
2026-02-04T21:56:33.4435257Z           	  "name": "projects/${projectId}/locations/global/parameters/test-${uniqueId}",
2026-02-04T21:56:33.4435415Z           	  "policyMember": {
2026-02-04T21:56:33.4435578Z           	... // 70 identical lines
2026-02-04T21:56:33.4435729Z           	  "format": "JSON",
2026-02-04T21:56:33.4436436Z           	  "kmsKey": "projects/${projectId}/locations/global/keyRings/test-keyring-${uniqueId}/cryptoKeys/test-key2-${uniqueId}",
2026-02-04T21:56:33.4436549Z         + 	  "labels": {
2026-02-04T21:56:33.4436677Z         + 	    "key": "value2"
2026-02-04T21:56:33.4436773Z         + 	  },
2026-02-04T21:56:33.4437214Z           	  "name": "projects/${projectId}/locations/global/parameters/test-${uniqueId}",
2026-02-04T21:56:33.4437368Z           	  "policyMember": {
2026-02-04T21:56:33.4437522Z           	... // 24 identical lines
2026-02-04T21:56:33.4437621Z           	"""
2026-02-04T21:56:33.4437709Z           )
```

Looks like the original _http.log does not contain the labels field in the POST or PATCH responses? Did you get the _http.log via running against the real API? If so, could you update the mockgcp to match this behavior?

maqiuyujoyce (approved):
/lgtm
/approve

dhavalbhensdadiya-crest:
Hi @maqiuyujoyce ,
The PR merge failed due to below errors in pubsub:
`
=== RUN   TestAllInSeries/fixtures/pubsubschema#01
    ratcheting.go:495: skipping re-reconciliation test for gvk pubsub.cnrm.cloud.google.com/v1beta1, Kind=PubSubSchema
    unified_test.go:275: not yet using SSA for create of resources in group "pubsub.cnrm.cloud.google.com"
    harness.go:243: starting envtest apiserver
    harness.go:246: error starting test environment: unable to start control plane itself: failed to start the controlplane. retried 5 times: exec: "etcd": executable file not found in $PATH
`

Do you think this will be resolved by retrying the merge or is there any other action required from my end?

PR #5841, opened 2025-12-17
Validate reference docs before the release

maqiuyujoyce:
### BRIEF Change description

Ensure we validate the resource reference docs before the release.

Note that currently we do have a bunch of incorrect URLs defined in the reference doc:
```
--- FAIL: TestReferenceDocConsistency (13.57s)
    doc_validation_test.go:170: Validating external URLs in template files...
    doc_validation_test.go:209: Found 357 unique URLs to validate in templates.
    doc_validation_test.go:271: URL https://cloud.google.com/resource-manager/reference/rest/v2/folders returned status 404 (referenced in: /usr/local/google/home/maqiuyu/go/src/github.com/maqiuyujoyce/4-k8s-config-connector/scripts/generate-google3-docs/resource-reference/templates/resourcemanager_folder.tmpl)
    doc_validation_test.go:271: URL https://cloud.google.com/iam/reference/rest/v1/iamPolicies returned status 404 (referenced in: /usr/local/google/home/maqiuyu/go/src/github.com/maqiuyujoyce/4-k8s-config-connector/scripts/generate-google3-docs/resource-reference/templates/iam_iamauditconfig.tmpl)
    doc_validation_test.go:271: URL https://cloud.google.com/dataform/docs/api/reference/rest/v1beta1/projects.locations.repositoriess returned status 404 (referenced in: /usr/local/google/home/maqiuyu/go/src/github.com/maqiuyujoyce/4-k8s-config-connector/scripts/generate-google3-docs/resource-reference/templates/dataform_dataformrepository.tmpl)
    doc_validation_test.go:271: URL https://cloud.google.com/iam/reference/rest/v1/projects.roles returned status 404 (referenced in: /usr/local/google/home/maqiuyu/go/src/github.com/maqiuyujoyce/4-k8s-config-connector/scripts/generate-google3-docs/resource-reference/templates/iam_iamcustomrole.tmpl)
    doc_validation_test.go:271: URL https://cloud.google.com/resource-manager/reference/rest/v1/projects returned status 404 (referenced in: /usr/local/google/home/maqiuyu/go/src/github.com/maqiuyujoyce/4-k8s-config-connector/scripts/generate-google3-docs/resource-reference/templates/resourcemanager_project.tmpl)
    doc_validation_test.go:271: URL https://cloud.google.com/resource-manager/reference/rest/v1/Policy returned status 404 (referenced in: /usr/local/google/home/maqiuyu/go/src/github.com/maqiuyujoyce/4-k8s-config-connector/scripts/generate-google3-docs/resource-reference/templates/resourcemanager_resourcemanagerpolicy.tmpl)
    doc_validation_test.go:271: URL https://cloud.google.com/iam/reference/rest/v1/projects.serviceAccounts returned status 404 (referenced in: /usr/local/google/home/maqiuyu/go/src/github.com/maqiuyujoyce/4-k8s-config-connector/scripts/generate-google3-docs/resource-reference/templates/iam_iamserviceaccount.tmpl)
    doc_validation_test.go:271: URL https://cloud.google.com/iam/reference/rest/v1/projects.serviceAccounts.keys returned status 404 (referenced in: /usr/local/google/home/maqiuyu/go/src/github.com/maqiuyujoyce/4-k8s-config-connector/scripts/generate-google3-docs/resource-reference/templates/iam_iamserviceaccountkey.tmpl)
    doc_validation_test.go:271: URL https://cloud.google.com/resource-manager/reference/rest/v1/liens returned status 404 (referenced in: /usr/local/google/home/maqiuyu/go/src/github.com/maqiuyujoyce/4-k8s-config-connector/scripts/generate-google3-docs/resource-reference/templates/resourcemanager_resourcemanagerlien.tmpl)
    doc_validation_test.go:277: Found 9 invalid URLs in templates
FAIL
FAIL	github.com/GoogleCloudPlatform/k8s-config-connector/scripts/generate-google3-docs/resource-reference	13.768s
FAIL
```

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/scripts/generate-google3-docs/resource-reference/doc_validation_test.go
+++ b/scripts/generate-google3-docs/resource-reference/doc_validation_test.go
@@ -1,4 +1,4 @@
-// Copyright 2022 Google LLC
+// Copyright 2026 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -15,7 +15,7 @@
 package main
 
 import (
-	"io/ioutil"
+	"fmt"
 	"net/http"
 	"os"
 	"path/filepath"
@@ -29,6 +29,10 @@ import (
 
 	"gopkg.in/yaml.v2"
 )
+
+// The path to Config Connector resource reference docs starts with
+// "/config-connector/docs/reference/resource-docs/".
+var docPathPrefix = "/config-connector/docs/reference/resource-docs/"
 
 type TOC struct {
 	TOC []TOCItem `yaml:"toc"`
@@ -69,8 +73,7 @@ func TestReferenceDocConsistency(t *testing.T) {
 				return err
 			}
 			// Normalize to the format used in _toc.yaml and overview.md
-			// They use /config-connector/docs/reference/resource-docs/...
-			webPath := "/config-connector/docs/reference/resource-docs/" + relPath
+			webPath := fmt.Sprintf("%s%s", docPathPrefix, relPath)
 			generatedFiles[webPath] = true
 		}
 		return nil
@@ -80,7 +83,7 @@ func TestReferenceDocConsistency(t *testing.T) {
 	}
 
 	// 2. Validate _toc.yaml
-	tocData, err := ioutil.ReadFile(tocPath)
+	tocData, err := os.ReadFile(tocPath)
 	if err != nil {
 		t.Fatalf("error reading _toc.yaml: %v", err)
 	}
@@ -100,7 +103,7 @@ func TestReferenceDocConsistency(t *testing.T) {
 	}
 
 	// 3. Validate overview.md
-	overviewData, err := ioutil.ReadFile(overviewPath)
+	overviewData, err := os.ReadFile(overviewPath)
 	if err != nil {
 		t.Fatalf("error reading overview.md: %v", err)
 	}
@@ -128,7 +131,9 @@ func TestReferenceDocConsistency(t *testing.T) {
 	}
 
 	// 5. Validate URLs in templates (Optional/integration check)
-	if os.Getenv("VALIDATE_URLS") == "true" {
+	if os.Getenv("VALIDATE_URLS") != "true" {
+		t.Log("Skipping URL validation: VALIDATE_URLS not set to true")
+	} else {
 		validateURLsInTemplates(t, templatesDir)
 	}
 }
@@ -144,28 +149,6 @@ func validatePath(t *testing.T, path string, generatedFiles, seenPaths map[strin
 	}
 }
 
-var (
-	// A list of regex patterns for URLs that should be skipped during validation.
-	skippedURLPatterns = []*regexp.Regexp{
-		regexp.MustCompile(`^https?://example\.com`),
-		regexp.MustCompile(`^https?://test\.com`),
-		regexp.MustCompile(`^http://localhost`),
-		regexp.MustCompile(`^http://metadata/`),
-		regexp.MustCompile(`\$\{.*?\}`),               // Templating variables
-		regexp.MustCompile(`\{\%.*?\%\}|\{\{.*?\}\}`), // Jinja/Go template syntax
-		regexp.MustCompile(`\[.*?\]`),                 // Placeholders
-	}
-)
-
-func shouldSkipURLValidation(url string) bool {
-	for _, pattern := range skippedURLPatterns {
-		if pattern.MatchString(url) {
-			return true
-		}
-	}
-	return false
-}
-
 func validateURLsInTemplates(t *testing.T, templatesDir string) {
 	t.Log("Validating external URLs in template files...")
 
@@ -178,7 +161,7 @@ func validateURLsInTemplates(t *testing.T, templatesDir string) {
 			return err
 		}
 		if !info.IsDir() && strings.HasSuffix(info.Name(), ".tmpl") {
-			content, err := ioutil.ReadFile(path)
+			content, err := os.ReadFile(path)
 			if err != nil {
 				return err
 			}
@@ -186,9 +169,6 @@ func validateURLsInTemplates(t *testing.T, templatesDir string) {
 			matches := hrefRegex.FindAllStringSubmatch(string(content), -1)
 			for _, match := range matches {
 				url := match[1]
-				if shouldSkipURLValidation(url) {
-					continue
-				}
 
 				// Construct full URL for relative paths
 				if strings.HasPrefix(url, "/") {
@@ -219,17 +199,18 @@ func validateURLsInTemplates(t *testing.T, templatesDir string) {
 	concurrency := 20
 	var wg sync.WaitGroup
 
+	client := &http.Client{
+		Timeout: 10 * time.Second,
+		Transport: &http.Transport{
+			MaxIdleConns:        concurrency,
+			MaxIdleConnsPerHost: concurrency,
+		},
+	}
+
 	for i := 0; i < concurrency; i++ {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
-			client := &http.Client{
-				Timeout: 10 * time.Second,
-				Transport: &http.Transport{
-					DisableKeepAlives: true,
-					MaxIdleConns:      5,
-				},
-			}
 
 			for u := range jobs {
 				res := result{url: u}


dev/release/generate-release.sh
```diff
@@ -49,14 +49,18 @@ if ! (go test ./pkg/controllers/...); then
...
-  
+
   echo "Retrying unit tests..."
   go test ./pkg/controllers/...
 fi
 
+echo "Validating resource reference docs..."
+cd ..
+# If failed, fix the issues in the resource reference docs.
+VALIDATE_URLS="true" go test ./scripts/generate-google3-docs/...
```
xiaoweim, line 62: Could you add a bit more context here to let the user know what to do if it fails?

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
```diff
@@ -0,0 +1,279 @@
...
+	// 4. Verify all generated files are referenced
+	for path := range generatedFiles {
+		if !tocPaths[path] {
+			t.Errorf("generated file %s is not referenced in _toc.yaml", path)
+		}
+		if !overviewPaths[path] {
+			t.Errorf("generated file %s is not referenced in overview.md", path)
+		}
+	}
+
```
xiaoweim, line 132: Non-blocking: I have recently worked to create new reference documents and received lots of feedback on the Google Standard preferred syntax to use for this type of documents. If we can maybe add a function, for example, `validateSyntax` here, it would save a lot of trouble for PR reviews in Google Internal. (I can add those once this is merged!)

maqiuyujoyce: Yeah, feel free to add more validation tests for resource docs when this process is established! I added some thoughts in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5841#issuecomment-3688528582.

dev/release/generate-release.sh
maqiuyujoyce: Done.

xiaoweim:
> For example, the field descriptions in the CRD are a big part of the reference doc. So fixing the doc itself works around the syntax issues but doesn't solve it. You'll want to make sure the field descriptions in the CRDs are correct. This means if there are any validations, we should do it at the time when the CRD is added/updated.

Agreed! We should run validation against the descriptions when new CRDs are added, that would be the most efficient way and save the most tine.


> Meanwhile, we didn't manually add those field descriptions, we mostly "inherited" them. They are from (1) the Terraform schema, (2) the DCL schema, (3) the protobuf from googleapis GitHub repo and (4) the templated descriptions, and only a smart part of them were (5) manually added. Among all these sources, we have full control of (1), (2), (4) and (5). But (3) the protobuf from googleapis GitHub repo is tricky because we periodically sync against master HEAD of googleapis GitHub repo to get the latest protobuf of the APIs, and we also re-generate the Go types for the resources from time to time. If there are syntax errors in the protobuf, we'll have to manually fix them repetitively if we only fix them in our resource Go types. Because of that, I think improving the doc quality for those fields is a bit out of scope for our team. Instead, this should probably be a GCP level effort to maintain the quality of the field docs in the protobuf.

Thanks for explainng the source of all the field descriptions where we get from. I agree that (3) should not be fix manually and repetitively by us, we need to have another discussion on how to handle (3).

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
```diff
@@ -0,0 +1,279 @@
+// Copyright 2022 Google LLC
```
cheftako: 2025

cheftako: AGENT: Please update the copyright year to 2026.

TAKO: Okay the agent got it more correct than I did.

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
```diff
@@ -0,0 +1,279 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"io/ioutil"
```
cheftako: AGENT: The `io/ioutil` package has been deprecated since Go 1.16. Please use the `io` and `os` packages instead (e.g., `os.ReadFile`, `io.ReadAll`).

TAKO: +1 as per https://pkg.go.dev/io/ioutil

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
```diff
@@ -0,0 +1,279 @@
...
+			return err
+		}
+		if !info.IsDir() && strings.HasSuffix(info.Name(), ".md") {
+			relPath, err := filepath.Rel(generatedDocsDir, path)
+			if err != nil {
+				return err
+			}
+			// Normalize to the format used in _toc.yaml and overview.md
+			// They use /config-connector/docs/reference/resource-docs/...
+			webPath := "/config-connector/docs/reference/resource-docs/" + relPath
```
cheftako: AGENT: The prefix `/config-connector/docs/reference/resource-docs/` is hardcoded here.
Ensure this matches the intended public documentation structure. If the base URL changes, this validation will fail.

TAKO: We've seen this move. I'd suggest making this a const near the top of the file. Easier to find and change when it does move.

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
```diff
@@ -0,0 +1,279 @@
...
+			generatedFiles[webPath] = true
+		}
+		return nil
+	})
+	if err != nil {
+		t.Fatalf("error walking generated docs dir: %v", err)
+	}
+
+	// 2. Validate _toc.yaml
+	tocData, err := ioutil.ReadFile(tocPath)
```
cheftako: AGENT: Replace `ioutil.ReadFile` with `os.ReadFile`.

TAKO: +1

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
```diff
@@ -0,0 +1,279 @@
...
+		if item.Path != "" {
+			validatePath(t, item.Path, generatedFiles, tocPaths, "_toc.yaml")
+		}
+		for _, section := range item.Section {
+			validatePath(t, section.Path, generatedFiles, tocPaths, "_toc.yaml")
+		}
+	}
+
+	// 3. Validate overview.md
+	overviewData, err := ioutil.ReadFile(overviewPath)
```
cheftako: AGENT: Replace `ioutil.ReadFile` with `os.ReadFile`.

TAKO: +1

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
```diff
@@ -0,0 +1,279 @@
...
+		regexp.MustCompile(`^https?://test\.com`),
+		regexp.MustCompile(`^http://localhost`),
+		regexp.MustCompile(`^http://metadata/`),
+		regexp.MustCompile(`\$\{.*?\}`),               // Templating variables
+		regexp.MustCompile(`\{\%.*?\%\}|\{\{.*?\}\}`), // Jinja/Go template syntax
+		regexp.MustCompile(`\[.*?\]`),                 // Placeholders
+	}
+)
+
+func shouldSkipURLValidation(url string) bool {
```
cheftako: AGENT: Use `t.Skip("Skipping URL validation...")` instead of `return`. This makes it clear in the test output that the test was skipped intentionally.

TAKO: Sounds better.

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
```diff
@@ -0,0 +1,279 @@
...
+					url = "https://cloud.google.com" + url
+				}
+
+				if strings.HasPrefix(url, "http") {
+					uniqueURLs[url] = append(uniqueURLs[url], path)
+				}
+			}
+		}
+		return nil
+	})
```
cheftako, line 184: AGENT: Replace `ioutil.ReadFile` with `os.ReadFile`.

TAKO: +1

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
```diff
@@ -0,0 +1,279 @@
...
+	if err != nil {
+		t.Fatalf("error collecting URLs from templates: %v", err)
+	}
+
+	t.Logf("Found %d unique URLs to validate in templates.", len(uniqueURLs))
+
+	type result struct {
+		url  string
+		err  error
+		code int
```
cheftako, line 194: AGENT: The regex `https?://[^\s)"]+` is simple and efficient. Just double-check if it needs to handle other common trailing punctuation in your docs (like comma `,` or period `.`) which are handled by `TrimRight` below, or markdown link syntaxes. It seems robust enough for now.

TAKO: I suspect we have a comment alignment problem.

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
```diff
@@ -0,0 +1,279 @@
...
+	jobs := make(chan string, len(uniqueURLs))
+	results := make(chan result, len(uniqueURLs))
+	concurrency := 20
+	var wg sync.WaitGroup
+
+	for i := 0; i < concurrency; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			client := &http.Client{
```
cheftako: AGENT: The `http.Client` is created inside `checkURL`, which is called for every URL.
This prevents connection reuse (keep-alive). Move the client creation outside the loop/function (e.g., create once and pass it in) to improve performance.

TAKO: https://pkg.go.dev/net/http#hdr-Clients_and_Transports states that "Clients and Transports are safe for concurrent use by multiple goroutines and for efficiency should only be created once and re-used." So I would give it a try.

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
```diff
@@ -0,0 +1,279 @@
...
+
+	if !generatedFiles[path] {
+		t.Errorf("path %s referenced in %s does not exist in generated output", path, source)
+	}
+}
+
+var (
+	// A list of regex patterns for URLs that should be skipped during validation.
+	skippedURLPatterns = []*regexp.Regexp{
+		regexp.MustCompile(`^https?://example\.com`),
```
cheftako: We should probably always use ^https? unless you have reason to believe the given url should only be skipped when its http.  Also its probably more efficient to combine like patterns.
Eg. '^https?://(example\.com|test\.com|localhost|metadata/)'

Also is there are a reason only metadata requires a trailing slash '/'?

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
```diff
@@ -0,0 +1,279 @@
...
+		regexp.MustCompile(`^https?://test\.com`),
+		regexp.MustCompile(`^http://localhost`),
+		regexp.MustCompile(`^http://metadata/`),
+		regexp.MustCompile(`\$\{.*?\}`),               // Templating variables
```
cheftako: Concerned that these last 3 regexes do not use a start of line indicator '^'. They will pretty much match ANY line containing square brackets [], including this one.

https://go.dev/play/p/mQQpeZu7zjC

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
maqiuyujoyce: Actually we should not use t.Skip() inside shouldSkipURLValidation in this context as t.Skip() only helps skip subtests within a loop using t.Run(), but not a normal loop (so no subtests).

Here is the reasoning (polished by gemini):

1. `t.Skip()` terminates the test: Calling t.Skip() doesn't just "skip this iteration of the loop"; it immediately stops the execution of the entire test function (or subtest). If we used it inside shouldSkipURLValidation, the first time the code encountered a URL like http://localhost, the entire test would stop, and no subsequent URLs would be validated.
2. It's a helper function: shouldSkipURLValidation is currently a pure helper function returning a boolean. Passing t into it just to skip logically breaks the flow of "iterating and filtering."

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
maqiuyujoyce: Hmmm, could you elaborate on what you suggested here?

Regarding the Agent feedback, it is a good general reminder for URL parsing, especially from unstructured text. However, for this specific implementation, it used `href="([^"]+)"` to extract URLs from HTML href attributes in .tmpl files, and then constructed the URL within the test before validate it.

So I think the current logic is sufficient. The agent's suggestions would be more relevant if we were extracting URLs from plain text directly.

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
maqiuyujoyce: Done.

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
maqiuyujoyce: Done.

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
maqiuyujoyce: Done.

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
maqiuyujoyce: Done.

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
maqiuyujoyce: Done.

maqiuyujoyce:
I removed the entire local url validation step as I think (and verified) no href attributes in the current templates contain the patterns that were being skipped, and there shouldn't be in the future.

Not skipping them is actually a good validation to ensure they don't exist.

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
maqiuyujoyce: After some rethinking, I removed the function. See https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5841#issuecomment-3712554730.

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
maqiuyujoyce: Done.

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
maqiuyujoyce: Done.

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
maqiuyujoyce: Makes sense. Though I removed this variable as it is no longer needed. See https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5841#issuecomment-3712554730.

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
maqiuyujoyce: Actually these regexes were used to skip the URLs that contain templating variables, etc, so not using `^` and `$` was intended.

Though this variable is removed now. See https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5841#issuecomment-3712554730.

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
cheftako: I liked most of what the agent suggested. In this case I wasn't sure why the comment was here or what it was refering to. Kept it in place as a reference for the agent dev team to look at what happened. Nothing for you to do on this one Joyce.

scripts/generate-google3-docs/resource-reference/doc_validation_test.go
```diff
@@ -0,0 +1,260 @@
...
+				return err
+			}
+
+			matches := hrefRegex.FindAllStringSubmatch(string(content), -1)
+			for _, match := range matches {
+				url := match[1]
+
+				// Construct full URL for relative paths
+				if strings.HasPrefix(url, "/") {
+					url = "https://cloud.google.com" + url
```
xiaoweim, line 175: some documentation start with "https://docs.cloud.google.com/"

PR #5842, opened 2025-12-17
feat(operator): Add metadataHost field for custom metadata server configuration #5740

### LINKED ISSUE #5741: Config Connector will fail on IPv6-only GKE clusters due to hardcoded metadata IP

cheftako:
Fixes #5741

## Summary

This PR adds a new `spec.metadataHost` field to the ControllerResource CRD that allows operators to specify a custom GCP metadata server hostname for Config Connector controllers.

## Problem

The GCP metadata server is currently accessed via the hardcoded IPv4 link-local address `169.254.169.254`. This will cause issues in IPv6-only GKE clusters where pods only have IPv6 addresses and cannot reach IPv4 endpoints:

```
dial tcp 169.254.169.254:80: connect: network is unreachable
```

Additionally, there are legitimate use cases for overriding the metadata endpoint beyond IPv6 support.

## Solution

When `spec.metadataHost` is set on a ControllerResource CR, the operator injects the `GCE_METADATA_HOST` environment variable into the controller's containers. The underlying GCP client libraries (`golang.org/x/oauth2/google` and `cloud.google.com/go/compute/metadata`) already support this env var.

The field is placed on ControllerResource (rather than ConfigConnector) to keep all controller pod resource modifications in one place, consistent with the existing pattern for container resources, replicas, and VPA settings.

## Usage

```yaml
kind: ControllerResource
metadata:
  name: cnrm-controller-manager
spec:
  metadataHost: "metadata.google.internal"
---
kind: ControllerResource
metadata:
  name: cnrm-webhook-manager
spec:
  metadataHost: "metadata.google.internal"
```

## Use Cases

- **IPv6-only GKE clusters**: When these become available, pods will need to use DNS (`metadata.google.internal`) or an IPv6 address (`fd20:ce::254`) instead of the IPv4 link-local address
- **Dual-stack clusters preferring IPv6**: Similar connectivity considerations
- **Metadata proxy configurations**: Organizations using metadata proxies for security auditing, credential scoping, or SSRF protection (e.g., [k8s-metadata-proxy](https://github.com/GoogleCloudPlatform/k8s-metadata-proxy))
- **Custom credential vending**: Integration with tools like HashiCorp Vault that can act as metadata proxies

## Changes

- `operator/pkg/apis/core/customize/v1beta1/controllerresource_types.go`: Added `MetadataHost` field to ControllerResourceSpec
- `operator/pkg/controllers/configconnector/configconnector_controller.go`: Call ApplyMetadataHost in applyControllerResourceCR
- `operator/pkg/controllers/utils.go`: Added ApplyMetadataHost function
- `operator/pkg/controllers/configconnector/metadata_host_test.go`: Updated tests for new location
-

### DIFF
--- a/tests/apichecks/testdata/exceptions/missingfields.txt
+++ b/tests/apichecks/testdata/exceptions/missingfields.txt
@@ -1465,6 +1465,7 @@
 [missing_field] crd=controllerreconcilers.customize.core.cnrm.cloud.google.com version=v1beta1: field ".spec.rateLimit.qps" is not set in unstructured objects
 [missing_field] crd=controllerresources.customize.core.cnrm.cloud.google.com version=v1beta1: field ".spec.containers[].name" is not set in unstructured objects
 [missing_field] crd=controllerresources.customize.core.cnrm.cloud.google.com version=v1beta1: field ".spec.containers[].resources.limits" is not set in unstructured objects
+[missing_field] crd=controllerresources.customize.core.cnrm.cloud.google.com version=v1beta1: field ".spec.metadataHost" is not set in unstructured objects
 [missing_field] crd=controllerresources.customize.core.cnrm.cloud.google.com version=v1beta1: field ".spec.containers[].resources.requests" is not set in unstructured objects
 [missing_field] crd=controllerresources.customize.core.cnrm.cloud.google.com version=v1beta1: field ".spec.replicas" is not set in unstructured objects
 [missing_field] crd=controllerresources.customize.core.cnrm.cloud.google.com version=v1beta1: field ".spec.verticalPodAutoscalerMode" is not set in unstructured objects


cheftako:
Closing in favor of #5844

PR #5852, opened 2025-12-19
tests: rerecon iampolicy

acpana:
While I can't turn on the full re-recon test yet, this will suffice for now!


tests/e2e/testdata/scenarios/iam/iampolicy/_http04.log
```diff
@@ -0,0 +1,69 @@
+POST https://cloudresourcemanager.googleapis.com/v1/projects/iampartialpolicy-dep-project:getIamPolicy?alt=json&prettyPrint=false
+Content-Type: application/json
+User-Agent: google-api-go-client/0.5 kcc/${kccVersion} (+https://github.com/GoogleCloudPlatform/k8s-config-connector) kcc/controller-manager/${kccVersion}
```
acpana, line 3: note this and the other http call are actually "gets"!

PR #5856, opened 2025-12-19
chore: cleanup legacy secret to use pattern more similar to refs

justinsb:
- **chore: cleanup legacy secret to use pattern more similar to refs**


apis/refs/v1beta1/secret/legacy.go
```diff
@@ -38,22 +38,71 @@ type Legacy struct {
...
+	r.Value = value
+	r.ValueFrom = nil
+	return nil
+}
+
+// ReadSecretValue returns the secret value, resolving the value if needed.
+func (r *Legacy) ReadSecretValue(ctx context.Context, fieldPath string, namespace string, reader client.Reader) (*string, error) {
+	value := r.Value
+	valueFrom := r.ValueFrom
+	if value != nil && valueFrom != nil {
```
cheftako, line 58: This is not thread safe with NormalizeSecret (if we care)

apis/refs/v1beta1/secret/legacy.go
```diff
@@ -38,22 +38,71 @@ type Legacy struct {
...
+	if valueFrom != nil {
+		if valueFrom.SecretKeyRef == nil {
+			return nil, fmt.Errorf("'%s.valueFrom.secretRef' "+
+				"should be configured", fieldPath)
+		}
+		secretValue, err := valueFrom.SecretKeyRef.ReadSecretValue(ctx, reader, namespace)
+		if err != nil {
+			return nil, err
+		}
+		value := string(secretValue)
```
cheftako, line 74: Is there a reason we leaving the Legacy with valueFrom set and value not? Seems like this may cause us to repeatedly look up the secret value.

pkg/controller/direct/alloydb/cluster_controller.go
```diff
@@ -294,9 +294,10 @@ func (a *ClusterAdapter) resolveInitialUserPasswordField(ctx context.Context) er
 	}
 
 	// Resolve sensitive field 'spec.initialUser.password' when it is set.
-	if err := direct.ResolveSensitiveField(ctx, obj.Spec.InitialUser.Password, "spec.initialUser.password", obj.Namespace, a.reader); err != nil {
+	if err := obj.Spec.InitialUser.Password.NormalizeSecret(ctx, "spec.initialUser.password", obj.Namespace, a.reader); err != nil {
```
cheftako, line 297: I realize its to late to remove 'initialUser.password.value' but do we have any checks to prevent this mistake from being repeated on other resources? Also should we update the description to indicate that users should use 'initialUser.password.valueFrom' instead?

apis/refs/v1beta1/secret/legacy.go
justinsb: NormalizeSecret is not thread-safe because it mutates in-place.  We normally do that on a mutable copy of the object that we have mapped from unstructured.  I wish go had `const` :-)

apis/refs/v1beta1/secret/legacy.go
justinsb: So the refs pattern would probably use `NormalizeSecret` and not have `ReadSecretValue`.  Even `NormalizeSecret` normally operates on a copy, so we are going to look up the secret repeatedly (i.e. once per Reconcile).

WDYT about marking ReadSecretValue as deprecated?  There's some risk that we write back to the object if we call NormalizeSecret, but if we were to do that accidentally we should catch it in our tests, and we don't do it on any other references so ....

apis/refs/v1beta1/secret/legacy.go
```diff
@@ -38,22 +38,71 @@ type Legacy struct {
...
+	return nil, fmt.Errorf("at least one of '%s.value' and '%s.valueFrom' "+
+		"should be configured: neither is configured", fieldPath, fieldPath)
+}
+
 // +kubebuilder:object:generate:=true
 type LegacyValueFrom struct {
 	/* Reference to a value with the given key in the given Secret in the resource's namespace. */
 	// +optional
-	SecretKeyRef *deprecatedrefs.SecretKeyRef `json:"secretKeyRef,omitempty"`
+	SecretKeyRef *LegacyValueFromSecretKeyRef `json:"secretKeyRef,omitempty"`
```
justinsb, line 85: This was the big thing, I wanted to stop using the generated client as that's just too circular.

pkg/controller/direct/alloydb/cluster_controller.go
justinsb: We can add a check in our API checks (`go test -v ./tests/apichecks/...`) and we can deprecate the whole type `LegacyValueFromSecretKeyRef` (and friends).

But it is much more convenient for users to be able to specify the secret inline.  But maybe we should make the less secure things not super-easy :-)

cheftako: +1

apis/refs/v1beta1/secret/legacy.go
cheftako: +1. I think we should probably have some sort of checks in mocks and controllers which errors out if we *ever* attempt to update any KRM secret.value field. Also we should probably try to avoid ever adding a new secret.value field. "it is much more convenient for users to be able to specify the secret inline." Its nice for demos. I think however its enough of an anti-pattern that we should look to avoid it going forward.

cheftako: > once per Reconcile

I'm fine with once per reconcile. That should mean once per 10 minutes (normally). I was more worried that over time that once becomes twice, thrice, etc

PR #5857, opened 2025-12-19
chore: use generator for logging v1beta1

justinsb:
- **chore: use generator for logging v1beta1**
  

- **autogen: generators for logging v1beta1**

- **chore: generate additional messages in generator**


config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_logginglogmetrics.logging.cnrm.cloud.google.com.yaml
```diff
@@ -70,7 +70,6 @@ spec:
                     properties:
                       bounds:
                         description: The values must be monotonically increasing.
-                        format: double
```
cheftako, line 73: Is it ok to lose this?

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_logginglogmetrics.logging.cnrm.cloud.google.com.yaml
```diff
@@ -80,31 +79,27 @@ spec:
                     properties:
                       growthFactor:
                         description: Must be greater than 1.
-                        format: double
                         type: number
                       numFiniteBuckets:
                         description: Must be greater than 0.
-                        format: int64
+                        format: int32
```
cheftako, line 85: I don't think down-scaling is technically safe. I guess we can make an argument that the number of buckets must be less than 2,147,483,647.

apis/logging/v1beta1/types.generated.go
```diff
@@ -0,0 +1,88 @@
...
+// +generated:types
+// krm.group: logging.cnrm.cloud.google.com
+// krm.version: v1beta1
+// proto.service: google.logging.v2
+// resource: LoggingLogMetric:LogMetric
+
+package v1beta1
+
+// +kcc:proto=google.api.Distribution.BucketOptions
+type Distribution_BucketOptions struct {
```
cheftako, line 24: AGENT: Generated types like `Distribution_BucketOptions` are introduced. The naming convention is verbose but clear.

apis/logging/v1beta1/zz_generated.deepcopy.go
```diff
@@ -25,6 +25,146 @@ import (
 	runtime "k8s.io/apimachinery/pkg/runtime"
 )
 
+// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
+func (in *Distribution_BucketOptions) DeepCopyInto(out *Distribution_BucketOptions) {
```
cheftako, line 29: AGENT: DeepCopy functions correctly generated for the new types.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_logginglogmetrics.logging.cnrm.cloud.google.com.yaml
```diff
@@ -80,31 +79,27 @@ spec:
                     properties:
                       growthFactor:
                         description: Must be greater than 1.
-                        format: double
                         type: number
                       numFiniteBuckets:
```
cheftako, line 83: AGENT: Breaking Change: `format` changed from `int64` to `int32`. While this matches the proto definition, verify that this won't cause validation issues for existing resources that might have been created with `int64` assumptions (though typically compatible in JSON).

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_logginglogmetrics.logging.cnrm.cloud.google.com.yaml
```diff
@@ -80,31 +79,27 @@ spec:
...
-                        format: double
                         type: number
                     type: object
                   linearBuckets:
                     description: The linear bucket.
                     properties:
                       numFiniteBuckets:
                         description: Must be greater than 0.
-                        format: int64
+                        format: int32
```
cheftako, line 96: AGENT: Same `int32` change for `linearBuckets.numFiniteBuckets`.

pkg/controller/direct/logging/mapper.generated.go
```diff
@@ -14,61 +14,168 @@
 
 // +generated:mapper
 // krm.group: logging.cnrm.cloud.google.com
-// krm.version: v1alpha1
+// krm.version: v1beta1
 // proto.service: google.logging.v2
 
 package logging
```
cheftako, line 21: AGENT: The mapper file now imports both `v1alpha1` and `v1beta1` APIs, reflecting that `LoggingLogMetric` is migrating while other resources (like `LoggingLink`?) might still be on `v1alpha1`.

pkg/controller/direct/logging/mapper.generated.go
```diff
@@ -14,61 +14,168 @@
...
 	return out
 }
-func LoggingLinkObservedState_FromProto(mapCtx *direct.MapContext, in *pb.Link) *krm.LoggingLinkObservedState {
+func Distribution_BucketOptions_FromProto(mapCtx *direct.MapContext, in *distributionpb.Distribution_BucketOptions) *krm.Distribution_BucketOptions {
 	if in == nil {
 		return nil
 	}
-	out := &krm.LoggingLinkObservedState{}
+	out := &krm.Distribution_BucketOptions{}
+	out.LinearBuckets = Distribution_BucketOptions_Linear_FromProto(mapCtx, in.GetLinearBuckets())
```
cheftako, line 70: AGENT: Generated mapper for `Distribution_BucketOptions` handles the oneof fields (linear, exponential, explicit) correctly.

cheftako (commented):
AGENT: The PR looks solid and correctly adopts the generator pattern for `LoggingLogMetric`. 

Points to note:
1.  **CRD Schema Change**: The switch to `int32` for `numFiniteBuckets` is correct per proto but strictly breaking for the CRD.
2.  **Filter Field**: The change to `*string` improves "optional vs empty" semantics but requires careful handling (which seems covered).
3.  **Tooling**: The improvements to `controllerbuilder` (cycle detection, explicit proto inclusion) are valuable.

Overall, a good cleanup and modernization.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_logginglogmetrics.logging.cnrm.cloud.google.com.yaml
justinsb: I checked this - we are using float64 (which is double precision).  controller-tools has an interesting position on floats in general: https://github.com/kubernetes-sigs/controller-tools/issues/245

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_logginglogmetrics.logging.cnrm.cloud.google.com.yaml
justinsb: Yes, I think int64 -> int32 is safe for these cases because nobody is going to have more than 2^31 in practice, and if they did it wouldn't work in the proto (which is int32).  We just lost the int64 vs int32 information when we were trying to go through terraform/dcl

apis/logging/v1beta1/types.generated.go
justinsb: I think we need to tell the agent to back off on generated files.

apis/logging/v1beta1/zz_generated.deepcopy.go
justinsb: This is pure noise ... looks like we've set up the agent to feel like it has to comment.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_logginglogmetrics.logging.cnrm.cloud.google.com.yaml
justinsb: I think number is floating point, not an integer.  But given the agent thinks it is an integer, we should make sure there's a test with a (non integer) floating point value

pkg/controller/direct/logging/mapper.generated.go
justinsb: That's right, but feels like noise

pkg/controller/direct/logging/mapper.generated.go
justinsb: Seems like noise, given it's generated

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_logginglogmetrics.logging.cnrm.cloud.google.com.yaml
yuwenma: (good to learn this!) Thanks for asking and answering.

justinsb:
codebot-robot did it

/close

PR #5864, opened 2025-12-19
Add structured diff reporting to SpannerBackupSchedule direct controller

anhdle-sso:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/controller/direct/spanner/backupschedule_controller.go
+++ b/pkg/controller/direct/spanner/backupschedule_controller.go
@@ -24,6 +24,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/common"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/structuredreporting"
 
 	gcp "cloud.google.com/go/spanner/admin/database/apiv1"
 	spannerbackupschedulespb "cloud.google.com/go/spanner/admin/database/apiv1/databasepb"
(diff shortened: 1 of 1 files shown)


pkg/controller/direct/spanner/backupschedule_controller.go
```diff
@@ -188,6 +189,11 @@ func (a *BackupScheduleAdapter) Update(ctx context.Context, updateOp *directbase
 	if desiredPb.EncryptionConfig == nil {
 		paths = paths.Delete(defaultOnEmptyFields...)
 	}
+	report := &structuredreporting.Diff{Object: updateOp.GetUnstructured()}
+	for path := range paths {
```
xiaoweim, line 197: Can we do the following?

```
if len(paths) > 0 {
	for path := range paths {
		report.AddField(path, nil, nil)
	}
	structuredreporting.ReportDiff(ctx, report)
}
```

pkg/controller/direct/spanner/backupschedule_controller.go
```diff
@@ -188,6 +189,11 @@ func (a *BackupScheduleAdapter) Update(ctx context.Context, updateOp *directbase
...
+		report.AddField(path, nil, nil)
```
xiaoweim, line 198: Same here, I do not get the two nil here for the old and new values, it's not informative in the report.

pkg/controller/direct/spanner/backupschedule_controller.go
anhdle-sso: We already had the the if statement len(paths) == 0 above which will terminate the function already.

pkg/controller/direct/spanner/backupschedule_controller.go
anhdle-sso: For the current ability of the CompareProtoMessage function, this is the best we can do for now. For preview purposes, I only need the field name as the actual value can be PII. I intend to add more feature to CompareProtoMessage but that would be a different PR.

pkg/controller/direct/spanner/backupschedule_controller.go
anhdle-sso: done

PR #5870, opened 2025-12-20
Add structured diff reporting to AssetFeed direct controller

anhdle-sso:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/controller/direct/asset/feed_controller.go
+++ b/pkg/controller/direct/asset/feed_controller.go
@@ -42,6 +42,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/common"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/structuredreporting"
 )
 
 func init() {
@@ -244,6 +245,12 @@ func (a *feedAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOpe
 		}
 		return updateOp.UpdateStatus(ctx, status, nil)
 	}
+
+	report := &structuredreporting.Diff{Object: updateOp.GetUnstructured()}
+	for path := range paths {
+		report.AddField(path, nil, nil)
+	}
+	structuredreporting.ReportDiff(ctx, report)
 
 	log.V(2).Info("updating asset feed fields", "paths", sets.List(paths))
 


pkg/controller/direct/asset/feed_controller.go
```diff
@@ -243,6 +244,12 @@ func (a *feedAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOpe
 		return updateOp.UpdateStatus(ctx, status, nil)
 	}
 
+	report := &structuredreporting.Diff{Object: updateOp.GetUnstructured()}
+	for path := range paths {
+		report.AddField(path, nil, nil)
```
xiaoweim, line 251: AGENT: Same here on the two nil for old and new values to be added to the report.

Xiaowei: Okay for now, can you add a TODO here so that we update it in the future.

pkg/controller/direct/asset/feed_controller.go
```diff
@@ -243,6 +244,12 @@ func (a *feedAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOpe
 		return updateOp.UpdateStatus(ctx, status, nil)
 	}
 
+	report := &structuredreporting.Diff{Object: updateOp.GetUnstructured()}
```
xiaoweim, line 249: AGENT: Iterating over `paths` (which is a map/set) results in non-deterministic order.
Consider using `sets.List(paths)` to iterate over the paths in a sorted order. This sorted list can also be reused for the log statement below to avoid sorting twice.

Xiaowei: Nice to have.

pkg/controller/direct/asset/feed_controller.go
```diff
@@ -243,6 +244,12 @@ func (a *feedAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOpe
...
+	for path := range paths {
```
xiaoweim, line 250: AGENT: To address the missing values, you could use reflection to retrieve the field values from `resource` (desired) and `a.actual` (actual) using the `path`.
Since `common.CompareProtoMessage` only returns paths, you might need a helper function (e.g., in `pkg/controller/direct/common`) to extract the values if you want to populate them here.

Xiaowei: Currently you are using `nil` and `nil` for the old and new values, so we probably do not need to extract the values for now. But in the future, when we need to populate the actual values, it is worth considering how to do that.

pkg/controller/direct/asset/feed_controller.go
justinsb: Or you could create a new method on the Diff, like `report.AddField_NoValuesYet`, mark it deprecated and leave a comment as to why in practice we have to use it, but we would like to eliminate it.

xiaoweim: +1

cheftako (commented):
Thanks Anh

PR #5871, opened 2025-12-20
Add structured diff reporting to AssetSavedQuery direct controller

anhdle-sso:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/controller/direct/asset/savedquery_controller.go
+++ b/pkg/controller/direct/asset/savedquery_controller.go
@@ -45,6 +45,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/common"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/structuredreporting"
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


pkg/controller/direct/asset/savedquery_controller.go
```diff
@@ -253,6 +254,12 @@ func (a *savedQueryAdapter) Update(ctx context.Context, updateOp *directbase.Upd
 		return updateOp.UpdateStatus(ctx, status, nil)
 	}
 
+	report := &structuredreporting.Diff{Object: updateOp.GetUnstructured()}
+	for path := range mutablePathsSet {
+		report.AddField(path, nil, nil)
```
xiaoweim, line 261: AGENT: Passing `nil, nil` for the old and new values limits the usefulness of the structured diff. 
While I understand `mutablePathsSet` only provides the field names, extracting and including the actual values from `a.actual` and `resource` (e.g., via a helper or reflection if appropriate) would significantly improve the debuggability and audit value of these reports.

Xiaowei: Can we add TODO here to update the `nil, nil` in the future?

pkg/controller/direct/asset/savedquery_controller.go
```diff
@@ -253,6 +254,12 @@ func (a *savedQueryAdapter) Update(ctx context.Context, updateOp *directbase.Upd
...
+	}
+	structuredreporting.ReportDiff(ctx, report)
```
xiaoweim, line 263: AGENT: Please verify if `structuredreporting.ReportDiff` returns an error. 
If it does, it should be handled or logged (e.g., as a warning) to ensure we are aware if reporting mechanism fails, without necessarily blocking the reconciliation.

Xiaowei: +1. It would be good to verify if `ReportDiff` returns an error.

xiaoweim (commented):
The structured diff reporting is a welcome addition.
Overall, the implementation looks correct in terms of placement within the `Update` loop.
I've left comments about potentially improving the data quality in the report (values vs just paths) and verifying error handling.

pkg/controller/direct/asset/savedquery_controller.go
anhdle-sso: structuredreporting.ReportDiff does not return an error

pkg/controller/direct/asset/savedquery_controller.go
anhdle-sso: For now, we don't want to report the actual value as it can contain senstive information. The listen can print the object value to cloud log which then exposed PII

cheftako (approved):
Looks good.

PR #5873, opened 2025-12-20
Add structured diff reporting to BigtableAppProfile direct controller

anhdle-sso:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/controller/direct/bigtable/appprofile_controller.go
+++ b/pkg/controller/direct/bigtable/appprofile_controller.go
@@ -24,6 +24,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/structuredreporting"
 
 	gcp "cloud.google.com/go/bigtable"
 	bigtablepb "cloud.google.com/go/bigtable/admin/apiv2/adminpb"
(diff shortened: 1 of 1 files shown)


xiaoweim:
It would be beneficial to add a unit test for the structured diff reporting in the `BigtableAppProfile` controller. This test should specifically:
- Verify Expected Paths: Ensure that `ReportDiff` is called with the correct field paths, such as multi_cluster_routing_use_any and single_cluster_routing, which should match the GCP API field names
- Confirm No-Op Behavior: Verify that when the actual state from GCP and the desired state from the Kubernetes spec are identical, no diffs are added to the report and hasChanges remains false
- Validate Change Capture: Confirm that changes in specific fields (like description or standard_isolation) are correctly captured and reflected in the structuredreporting.Diff

anhdle-sso:
> The resource is bigtable app profile from the PR, the `AppProfileIdentity` in the PR title is a little misleading.

Updated the PR title

anhdle-sso:
> It would be beneficial to add a unit test for the structured diff reporting in the `BigtableAppProfile` controller. This test should specifically:
> 
> * Verify Expected Paths: Ensure that `ReportDiff` is called with the correct field paths, such as multi_cluster_routing_use_any and single_cluster_routing, which should match the GCP API field names
> * Confirm No-Op Behavior: Verify that when the actual state from GCP and the desired state from the Kubernetes spec are identical, no diffs are added to the report and hasChanges remains false
> * Validate Change Capture: Confirm that changes in specific fields (like description or standard_isolation) are correctly captured and reflected in the structuredreporting.Diff

I will add unittest in another PR if needed.

pkg/controller/direct/bigtable/appprofile_controller.go
```diff
@@ -229,30 +230,37 @@ func (a *BigtableAppProfileAdapter) Update(ctx context.Context, updateOp *direct
...
 	}
 
+	report := &structuredreporting.Diff{Object: updateOp.GetUnstructured()}
+
 	if desired.Spec.Description != nil && !cmp.Equal(resource.Description, a.actual.Description) {
+		report.AddField("description", a.actual.Description, resource.Description)
 		fieldsToUpdate.Description = resource.Description
 		hasChanges = true
 	}
 	if desired.Spec.MultiClusterRoutingUseAny != nil && !cmp.Equal(resource.GetMultiClusterRoutingUseAny(), a.actual.GetMultiClusterRoutingUseAny(), cmpopts.IgnoreUnexported(bigtablepb.AppProfile_MultiClusterRoutingUseAny{})) {
```
cheftako, line 240: AGENT: You are using the snake_case proto field name "multi_cluster_routing_use_any" here (and similarly for other fields).
Since the `report.Object` is the KRM resource, using the camelCase KRM field name (e.g., "multiClusterRoutingUseAny") might be more intuitive for users reading the report, as it maps directly to their YAML configuration.
However, if the project convention is to strictly use proto field names for these IDs, then this is consistent.

TAKO: Comment seems to be 1 line too early. I don't really care which case we use for the report. I do care that we are consistent. Please make sure that *all* reports use the same case.

pkg/controller/direct/bigtable/appprofile_controller.go
```diff
@@ -229,30 +230,37 @@ func (a *BigtableAppProfileAdapter) Update(ctx context.Context, updateOp *direct
...
+		report.AddField("multi_cluster_routing_use_any", a.actual.GetMultiClusterRoutingUseAny(), resource.GetMultiClusterRoutingUseAny())
```
cheftako, line 241: AGENT: Consider using the KRM field name `multiClusterRoutingUseAny` (camelCase) here instead of snake_case. This matches the user's YAML configuration, making the diff report easier to understand.

TAKO: Again I do not care strongly. I do care that its consistent. I would also suggest that it may have a point. Customers are used to dealing with YAML. Having the field name in a form which can be used to search the YAML is probably helpful.

pkg/controller/direct/bigtable/appprofile_controller.go
```diff
@@ -229,30 +230,37 @@ func (a *BigtableAppProfileAdapter) Update(ctx context.Context, updateOp *direct
...
 		fieldsToUpdate.RoutingConfig = &gcp.MultiClusterRoutingUseAnyConfig{
 			ClusterIDs: resource.GetMultiClusterRoutingUseAny().ClusterIds,
 		}
 		hasChanges = true
 	}
 	if desired.Spec.SingleClusterRouting != nil && !cmp.Equal(resource.GetSingleClusterRouting(), a.actual.GetSingleClusterRouting(), cmpopts.IgnoreUnexported(bigtablepb.AppProfile_SingleClusterRouting{})) {
+		report.AddField("single_cluster_routing", a.actual.GetSingleClusterRouting(), resource.GetSingleClusterRouting())
```
cheftako, line 248: AGENT: Similarly, consider using `singleClusterRouting` to align with the KRM field name.

TAKO: As above.

cheftako (commented):
AGENT: The PR looks solid. It correctly introduces structured diff reporting for the `BigtableAppProfile` resource, covering the mutable fields.

I have a suggestion regarding the naming convention of the diff fields: using KRM-style camelCase keys (e.g., `multiClusterRoutingUseAny`) might be more intuitive for users debugging their YAML, as opposed to the internal snake_case proto field names. Additionally, I have a note on the behavior for `oneof` fields.

TAKO: Having the field in the same form you need to do a search in YAML is pretty compelling.

PR #5882, opened 2025-12-20
Add structured diff reporting to NotebookInstance direct controller

anhdle-sso:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/controller/direct/notebooks/instance_controller.go
+++ b/pkg/controller/direct/notebooks/instance_controller.go
@@ -25,6 +25,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/common"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/structuredreporting"
 
 	gcp "cloud.google.com/go/notebooks/apiv1"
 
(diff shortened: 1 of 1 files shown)


pkg/controller/direct/notebooks/instance_controller.go
```diff
@@ -177,6 +178,13 @@ func (a *InstanceAdapter) Update(ctx context.Context, updateOp *directbase.Updat
 	if len(paths) == 0 {
 		log.V(2).Info("no field needs update", "name", a.id)
 	}
+
+	report := &structuredreporting.Diff{Object: updateOp.GetUnstructured()}
+	for path := range paths {
```
gemmahou, line 183: Agent: If `paths` is a map-based set (like `sets.String`), iterating over it directly will result in non-deterministic order. 
It might be better to sort the paths to ensure consistent reporting/logging, unless `structuredreporting.Diff` handles sorting internally.

Gemma: I do not think we worry about the order here.

pkg/controller/direct/notebooks/instance_controller.go
```diff
@@ -177,6 +178,13 @@ func (a *InstanceAdapter) Update(ctx context.Context, updateOp *directbase.Updat
...
+		report.AddField(path, nil, nil)
```
gemmahou, line 184: Agent: We are passing `nil` for both the actual and expected values.
While knowing which fields changed is useful, having the values would be much more valuable for debugging. 
Is it feasible to populate the values here, or is the intention solely to track the field names?

Gemma: discussed offline, this will be handled separately and we are using nil for now

PR #5883, opened 2025-12-20
Add structured diff reporting to RedisCluster direct controller

anhdle-sso:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/controller/direct/redis/cluster/rediscluster_controller.go
+++ b/pkg/controller/direct/redis/cluster/rediscluster_controller.go
@@ -35,6 +35,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/common"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/structuredreporting"
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


pkg/controller/direct/redis/cluster/rediscluster_controller.go
```diff
@@ -259,28 +260,35 @@ func (a *redisClusterAdapter) Update(ctx context.Context, updateOp *directbase.U
...
+		report.AddField("deletion_protection_enabled", a.actual.DeletionProtectionEnabled, a.desired.DeletionProtectionEnabled)
 	}
 
 	if !proto.Equal(a.desired.PersistenceConfig, a.actual.PersistenceConfig) {
+		report.AddField("persistence_config", a.actual.PersistenceConfig, a.desired.PersistenceConfig)
 		updateMask.Paths = append(updateMask.Paths, "persistence_config")
 	}
 
 	if !reflect.DeepEqual(a.desired.GetRedisConfigs(), a.actual.GetRedisConfigs()) {
+		report.AddField("redis_configs", a.actual.GetRedisConfigs(), a.desired.GetRedisConfigs())
```
xiaoweim, line 286: AGENT: Please verify if `redis_configs` contains any sensitive information (e.g., specific Redis configuration parameters that might be considered secrets). If so, we should ensure they are redacted before passing them to `report.AddField`, as this will likely be logged.

Xiaowei: +1, we do not want to expose sensitive information, it would be good to double check.

pkg/controller/direct/redis/cluster/rediscluster_controller.go
```diff
@@ -259,28 +260,35 @@ func (a *redisClusterAdapter) Update(ctx context.Context, updateOp *directbase.U
 
 	// TODO: What if a different field (immutability again)
```
xiaoweim, line 263: AGENT:
Nit: The pattern of manually updating `updateMask` and adding to `report` is repeated for every field. Consider using a helper function or a struct to map fields to their proto values and update both the mask and report together. This would reduce code duplication and the risk of forgetting one when adding new fields.

Xiaowei: we can ignore this, I think this is not a change for this PR.

xiaoweim (commented):
AGENT: The changes correctly implement structured diff reporting for the `RedisCluster` resource. 

I have a couple of comments regarding potential security implications of logging `redis_configs` and a suggestion for maintainability.

pkg/controller/direct/redis/cluster/rediscluster_controller.go
anhdle-sso: it is map[string]string, probably the reason why repo agent mark this. it is not marked as sensitive in our CRD or GCP doc (https://docs.cloud.google.com/memorystore/docs/cluster/reference/rest/v1beta1/projects.locations.clusters#Cluster)

PR #5884, opened 2025-12-20
Add structured diff reporting to WorkstationConfig direct controller

anhdle-sso:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/controller/direct/workstations/config_controller.go
+++ b/pkg/controller/direct/workstations/config_controller.go
@@ -25,6 +25,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/fuzztesting"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/structuredreporting"
 
 	gcp "cloud.google.com/go/workstations/apiv1"
 	pb "cloud.google.com/go/workstations/apiv1/workstationspb"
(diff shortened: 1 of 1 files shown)


pkg/controller/direct/workstations/config_controller.go
```diff
@@ -235,6 +236,13 @@ func (a *WorkstationConfigAdapter) Update(ctx context.Context, updateOp *directb
 		}
 		return updateOp.UpdateStatus(ctx, status, nil)
 	}
+
```
xiaoweim, line 241: AGENT: `paths` is a map (set), so the iteration order is non-deterministic.
Consider using `sortedPaths := sets.List(paths)` here to ensure the reported fields are always in the same order.
Additionally, you can reuse `sortedPaths` in the `UpdateMask` field below (line 246) to avoid sorting the keys twice.

Xiaowei: Nice to have and maybe not for this PR.

pkg/controller/direct/workstations/config_controller.go
```diff
@@ -235,6 +236,13 @@ func (a *WorkstationConfigAdapter) Update(ctx context.Context, updateOp *directb
...
+	report := &structuredreporting.Diff{Object: updateOp.GetUnstructured()}
```
xiaoweim, line 242: AGENT: Passing `nil, nil` means the report will only show which fields changed but not the values. 
While this is acceptable given `CompareProtoMessage` only returns paths, it's worth noting this limitation. If accessing the values becomes feasible in the future, it would greatly enhance the report's utility.

Xiaowei: Okay for dry run for now, need to change the "nil, nil" for future. Non-blocking.

xiaoweim (commented):
The changes look good and functionally correct. I have a minor suggestion regarding the iteration order of the paths for deterministic reporting and potential optimization, and a note on the diff details.

PR #5885, opened 2025-12-20
Add structured diff reporting to Workstation direct controller

anhdle-sso:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/controller/direct/workstations/workstation_controller_test.go
+++ b/pkg/controller/direct/workstations/workstation_controller_test.go
@@ -1,0 +1,238 @@
+/*
+Copyright 2024 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package workstations
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"strings"
+	"testing"
+
+	gcp "cloud.google.com/go/workstations/apiv1"
+	"cloud.google.com/go/workstations/apiv1/workstationspb"
+	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/workstations/v1beta1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/lifecyclehandler"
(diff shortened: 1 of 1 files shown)


pkg/controller/direct/workstations/workstation_controller.go
```diff
@@ -216,6 +217,13 @@ func (a *WorkstationAdapter) Update(ctx context.Context, updateOp *directbase.Up
 		}
 		return updateOp.UpdateStatus(ctx, status, nil)
 	}
+
+	report := &structuredreporting.Diff{Object: updateOp.GetUnstructured()}
```
cheftako, line 221: Can we please add a test for the results of using common.CompareProtoMessage + structuredreporting on workstation? I would expect a test that 2 separate (but identical) instances do not generate any reported changes. Ideally we should do things like perturb list ordering for lists where the ordering does not convey meaning (Eg labels). I would also like to see at least 1 example where there is a diff.

anhdle-sso: yes, I will add the test in a separate PR

pkg/controller/direct/workstations/workstation_controller.go
```diff
@@ -216,6 +217,13 @@ func (a *WorkstationAdapter) Update(ctx context.Context, updateOp *directbase.Up
...
+	for path := range paths {
+		report.AddField(path, nil, nil)
```
cheftako, line 223: AGENT: Passing `nil, nil` for the old and new values is consistent with `common.CompareProtoMessage` returning only paths. However, this limits the utility of the report. If feasible, consider looking up the actual values from `a.actual` and `desiredPb` using the path to provide a richer diff in the future.

pkg/controller/direct/workstations/workstation_controller.go
cheftako, line 221: AGENT: Please add a unit test (e.g., in `workstation_controller_test.go`) that verifies `ReportDiff` is called with the expected paths. It is important to confirm that identical protos result in no reported diffs and that changes are correctly captured.

pkg/controller/direct/workstations/workstation_controller_test.go
```diff
@@ -0,0 +1,238 @@
...
+		httpClient := &http.Client{Transport: mockHTTP}
+		gcpClient, _ := gcp.NewRESTClient(ctx, option.WithHTTPClient(httpClient), option.WithoutAuthentication())
+
+		desired := &krm.Workstation{
+			Spec: krm.WorkstationSpec{
+				DisplayName: direct.PtrTo("new-name"),
+				Labels: []krm.WorkstationLabel{
+					{Key: "key1", Value: "value1"},
+					{Key: "key2", Value: "value2"},
+				},
```
gemmahou, line 180: nit: Should we include the `Annotation` field in the test? Since it's also an array, I’m unsure if element order matters.

If we want to avoid redundant tests for fields with identical types and behaviors, we could add a comment to clarify the scope. For example:
```
// String fields (e.g., displayName, xxx): testing displayName only.
// Unordered arrays (e.g., labels, annotations): testing labels only.
// Ordered arrays (e.g., yyy, zzz): testing yyy only.
```

gemmahou (commented):
/lgtm

PR #5892, opened 2025-12-23
chore: gemini settings.json expects strings for env vars

justinsb:


justinsb:
Without this we get this error:

```
Invalid configuration in <...>/.gemini/settings.json:

Error in: mcpServers.github.env.GITHUB_READ-ONLY
    Expected string, received number
Expected: string, but received: number

Please fix the configuration and try again.
See: https://github.com/google-gemini/gemini-cli/blob/main/docs/get-started/configuration.md
```

justinsb:
/assign @cheftako

Unblocks using latest gemini, seems to only be a problem with version 0.22.2 (and later, I presume)

PR #5670, opened 2025-11-18
feat: Add API, Fuzzer, Mapper  for CloudDMSMigrationJob

yuwenma:
### BRIEF Change description

Fixes #

#### Does this PR add something which needs to be 'release noted'?

```release-note

```

- [ ] Reviewer reviewed release note.

#### Additional documentation e.g., references, usage docs, etc.:

```docs

```

#### Intended Milestone

Please indicate the intended milestone. 
- [ ] Reviewer tagged PR with the actual milestone.

### Tests you have done

- [ ] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #5308, opened 2025-10-04
Add Bigtablematerializedview Controller

shavonz:
### BRIEF Change description

This changset:
- Adds the controller for bigtablematerializedview
- Adds mockgcp support for bigtablematerializedview

#### Does this PR add something which needs to be 'release noted'?

```release-note

```

- [ ] Reviewer reviewed release note.

#### Additional documentation e.g., references, usage docs, etc.:

```docs

```

#### Intended Milestone

Please indicate the intended milestone. 
- [ ] Reviewer tagged PR with the actual milestone.

### Tests you have done

Ran:
```
dev/tasks/record-gcp mockbigtable/testdata/materializedview/crud
dev/tasks/compare-mock mockbigtable/testdata/materializedview/crud
```

- [x] Run `make ready-pr` to ensure this PR is ready for review.
- [ ] Perform necessary E2E testing for changed resources.

PR #5893, opened 2025-12-23
Added controller and fixture tests for Parameter Manager Parameter

### LINKED ISSUE #5785: Add support of new resource ParameterManagerParameter
### Checklist

The goal of this ticket is to add the support of Parameter Manager Parameter as a new resource in KCC.

### Additional information

Docs: https://docs.cloud.google.com/secret-manager/parameter-manager/docs/overview

Go reference: https://pkg.go.dev/cloud.google.com/go/parametermanager/apiv1/parametermanagerpb#Parameter

Google APIs proto: https://github.com/googleapis/googleapis/tree/master/google/cloud/parametermanager/v1

### Importance

_No response_

dhavalbhensdadiya-crest:
### BRIEF Change description
Added the controller and fixture tests.

Fixes #5785 

This change adds the support of new resource parameter manager.

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/controller/direct/parametermanager/parameter_controller.go
+++ b/pkg/controller/direct/parametermanager/parameter_controller.go
@@ -210,6 +210,12 @@ func (a *ParameterAdapter) Update(ctx context.Context, updateOp *directbase.Upda
 	if !reflect.DeepEqual(a.actual.GetLabels(), resource.GetLabels()) {
 		paths = append(paths, "labels")
 	}
+	// Add to update_mask only if format is specified
+	if desired.Spec.Format != nil {
+		if !reflect.DeepEqual(a.actual.GetFormat(), resource.GetFormat()) {
+			paths = append(paths, "format")
+		}
+	}
 
 	if len(paths) == 0 {
 		log.V(2).Info("no field needs update", "name", a.id)


maqiuyujoyce:
Hi @dhavalbhensdadiya-crest , thanks a lot for supporting the resource! Could you split this PR into the follow sub PRs for easier review and tracking?

* crud, log and MockGCP (e.g. https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5716)
* API, Mapper and fuzzer (e.g. https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5670)
* controller (*_identity, *_reference), and fixture tests (e.g. https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5308)

The first two don't need to be depend on each other.

dhavalbhensdadiya-crest:
Hi @maqiuyujoyce ,
Thanks for picking up the review of Parameter support. I have assigned you to https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5809 which is the first PR. 

I have raised 3 PR for Parameter which are built on top of each other.
1. https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5809 - API, Mapper, and Fuzzer.
2. https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5833 - Mock gcp server and mock functions
3. https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5893 - Controller and Fixture tests

Once each PR is merged, I plan to take the pull in the subsequent PRs so that only relevant changes are visible to the reviewer. Do let me know if this works or if you need me to raise separate PRs with only limited files. Thanks

dhavalbhensdadiya-crest:
Hi @maqiuyujoyce ,
I have rebased this PR on master branch, and also regenerated the real and mockgcp logs for test cases in resourcefixture. 
This PR is ready for your review now.

pkg/controller/direct/parametermanager/parameter_controller.go
```diff
@@ -0,0 +1,297 @@
...
+
+	id, err := obj.GetIdentity(ctx, reader)
+	if err != nil {
+		return nil, err
+	}
+
+	format := direct.ValueOf(obj.Spec.Format)
+
+	if format == "" {
+		obj.Spec.Format = direct.LazyPtr("UNFORMATTED")
```
maqiuyujoyce: The validation for `Format` is done here in the controller.
It is recommended to also add validation markers (e.g. `// +kubebuilder:validation:Enum=UNFORMATTED;JSON;YAML`) in the CRD (in `types.go`) to enforce this at the API level.

pkg/controller/direct/parametermanager/parameter_controller.go
```diff
@@ -0,0 +1,297 @@
...
+
+	mapCtx := &direct.MapContext{}
+
+	copied := obj.DeepCopy()
+
+	if err := m.normalizeKMSKeyRef(ctx, reader, obj, copied); err != nil {
+		return nil, err
+	}
+
+	desired := ParameterManagerParameterSpec_ToProto(mapCtx, &copied.Spec)
```
maqiuyujoyce: `ParameterManagerParameterSpec_ToProto` is used here but is not defined in the provided diff. 
Please ensure `pkg/controller/direct/parametermanager/parameter_mappings.go` is included in the PR.

pkg/controller/direct/parametermanager/parameter_controller.go
```diff
@@ -0,0 +1,297 @@
...
+		Parameter:   a.desired,
+		ParameterId: a.id.ID(),
+	}
+	created, err := a.gcpClient.CreateParameter(ctx, req)
+	if err != nil {
+		return fmt.Errorf("creating Parameter %s: %w", a.id, err)
+	}
+	log.V(2).Info("successfully created Parameter", "name", a.id)
+
+	mapCtx := &direct.MapContext{}
```
maqiuyujoyce: `ParameterManagerParameterObservedState_FromProto` is used here but is not defined in the provided diff.

pkg/controller/direct/parametermanager/parameter_controller.go
```diff
@@ -0,0 +1,297 @@
...
+	return &modelParameter{config: *config}, nil
+}
+
+var _ directbase.Model = &modelParameter{}
+
+type modelParameter struct {
+	config config.ControllerConfig
+}
+
+func (m *modelParameter) client(ctx context.Context, location string) (*gcp.Client, error) {
```
maqiuyujoyce: The regional endpoint construction `parametermanager.%s.rep.googleapis.com:443` matches the expected pattern for Parameter Manager.

apis/parametermanager/v1alpha1/parameter_reference.go
```diff
@@ -0,0 +1,70 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ refsv1beta1.Ref = &ParameterRef{}
+
+// ParameterRef is a reference to a ParameterManagerParameter resource.
+type ParameterRef struct {
+	// A reference to an externally managed ParameterManagerParameter resource.
+	// Should be in the format "projects/{{projectID}}/locations/{{location}}/catalogs/{{catalogID}}".
```
maqiuyujoyce: Copy-paste error: `catalogs/{{catalogID}}` should likely be `parameters/{{parameterID}}`.

config/samples/resources/parametermanagerparameter/globalparameter/parametermanager_v1alpha1_parametermanagerparameter.yaml
```diff
@@ -0,0 +1,26 @@
...
+kind: ParameterManagerParameter
+metadata:
+  labels:
+    label-one: value-one
+  name: parametermanagerparameter-sample
+spec:
+  projectRef:
+    external: projects/${PROJECT_ID?}
+  location: global
+  format: JSON
```
maqiuyujoyce, line 25: Nit: Trailing whitespace.

pkg/controller/direct/parametermanager/parameter_controller.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+	externalRef := a.actual.GetName()
+	var id *krm.ParameterIdentity
+	if err := id.FromExternal(externalRef); err != nil {
+		return nil, fmt.Errorf("parsing external ref %q: %w", externalRef, err)
+	}
+
+	obj.Spec.ProjectAndLocationRef.ProjectRef = &refs.ProjectRef{External: a.id.Parent().ProjectID}
+	obj.Spec.ProjectAndLocationRef.Location = a.id.Parent().Location
+
```
maqiuyujoyce: `obj.Spec.ProjectAndLocationRef.ProjectRef` implies `ProjectAndLocationRef` exists in Spec. Cannot verify without types file.

apis/parametermanager/v1alpha1/parameter_identity.go
```diff
@@ -0,0 +1,97 @@
...
+	parameter.id = common.ValueOf(obj.Spec.ResourceID)
+	if parameter.id == "" {
+		parameter.id = obj.GetName()
+	}
+	if parameter.id == "" {
+		return nil, fmt.Errorf("cannot resolve resource ID")
+	}
+
+	// Validate against the ID stored in status.externalRef, if any
+	externalRef := common.ValueOf(obj.Status.ExternalRef)
```
maqiuyujoyce: `obj.Status.ExternalRef`. Ensure this field exists in the Status struct.

pkg/controller/direct/parametermanager/parameter_controller.go
```diff
@@ -0,0 +1,297 @@
...
+	obj.Spec.ResourceID = direct.LazyPtr(a.id.ID())
+	uObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
+	if err != nil {
+		return nil, err
+	}
+
```
maqiuyujoyce: Ensure `ResourceID` is a field in the Spec.

pkg/controller/direct/parametermanager/parameter_controller.go
```diff
@@ -0,0 +1,297 @@
...
+	u.SetName(a.id.ID())
+	u.SetGroupVersionKind(krm.ParameterManagerParameterGVK)
+
+	u.Object = uObj
+	return u, nil
+}
+
```
maqiuyujoyce: `krm.ParameterManagerParameterGVK` usage.

config/samples/resources/parametermanagerparameter/regionalparameter/parametermanager_v1alpha1_parametermanagerparameter_regional.yaml
```diff
@@ -0,0 +1,27 @@
...
+  labels:
+    label-one: value-one
+  name: parametermanagerparameter-regional-sample
+spec:
+  projectRef:
+    external: projects/${PROJECT_ID?}
+  location: us-central1
+  format: YAML
+  kmsKeyRef:
+    name: kmscryptokey-sample
```
maqiuyujoyce: You'll want to add a manifest file for this referenced resource as well.

pkg/controller/direct/parametermanager/parameter_controller.go
```diff
@@ -0,0 +1,297 @@
...
+func (m *modelParameter) client(ctx context.Context, location string) (*gcp.Client, error) {
+	var opts []option.ClientOption
+	opts, err := m.config.RESTClientOptions()
+	if err != nil {
+		return nil, err
+	}
+
+	// Add regional endpoint if location is specified
```
maqiuyujoyce: Nit: Because location is a required field (https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/parametermanager/v1alpha1/parameter_types.go#L30), it is always specified. Maybe use more clear description like `Add regional endpoint if a region is specified`.

pkg/controller/direct/parametermanager/parameter_controller.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	id, err := obj.GetIdentity(ctx, reader)
+	if err != nil {
+		return nil, err
+	}
+
+	format := direct.ValueOf(obj.Spec.Format)
+
+	if format == "" {
```
maqiuyujoyce: We don't want to do the defaulting here because users may apply a KCC resource for either creation or acquisition, and in the acquisition use case, this field will be determined by the underlying GCP resource if it is not specified in the KRM resource.

Or if this field should be required in general, then we should require it at the CRD level (in the Go types).

pkg/controller/direct/parametermanager/parameter_controller.go
```diff
@@ -0,0 +1,297 @@
...
+		obj.Spec.Format = direct.LazyPtr("UNFORMATTED")
+	} else {
+		if format != "UNFORMATTED" && format != "JSON" && format != "YAML" {
```
maqiuyujoyce: We don't want to do the error check here because the GCP API may extend the support to more formats.

pkg/controller/direct/parametermanager/parameter_controller.go
```diff
@@ -0,0 +1,297 @@
...
+	desired := ParameterManagerParameterSpec_ToProto(mapCtx, &copied.Spec)
+	if mapCtx.Err() != nil {
+		return nil, mapCtx.Err()
+	}
+
+	// if the proto `desired` has field "labels". we should do `desired.Labels = label.NewGCPLabelsFromK8sLabels(u.GetLabels())
+	desired.Labels = label.NewGCPLabelsFromK8sLabels(u.GetLabels())
+
+	location := obj.Spec.ProjectAndLocationRef.Location
+	if location == "" {
```
maqiuyujoyce: No need. Location is defined at https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/parametermanager/v1alpha1/parameter_types.go#L30, and it is a required field.

pkg/controller/direct/parametermanager/parameter_controller.go
```diff
@@ -0,0 +1,297 @@
...
+		if format != "UNFORMATTED" && format != "JSON" && format != "YAML" {
+			return nil, fmt.Errorf("invalid format %q, only UNFORMATTED, JSON, and YAML are supported", format)
+		}
+	}
+
+	mapCtx := &direct.MapContext{}
+
+	copied := obj.DeepCopy()
+
+	if err := m.normalizeKMSKeyRef(ctx, reader, obj, copied); err != nil {
```
maqiuyujoyce: Should be called in the beginning of Create() and Update() instead of here.

pkg/controller/direct/parametermanager/parameter_controller.go
```diff
@@ -0,0 +1,297 @@
...
+	paths := make(sets.Set[string])
+	var err error
+	a.desired.Name = a.id.String()
+	paths, err = common.CompareProtoMessage(a.desired, a.actual, common.BasicDiff)
+	if err != nil {
+		return err
+	}
+
+	if len(paths) == 0 {
+		log.V(2).Info("no field needs update", "name", a.id)
```
maqiuyujoyce, line 221: You'll want to cover the acquisition use case here as well. See https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/alloydb/instance_controller.go#L286 as an example.

pkg/controller/direct/parametermanager/parameter_controller.go
```diff
@@ -0,0 +1,297 @@
...
+	updated, err := a.gcpClient.UpdateParameter(ctx, req)
+	if err != nil {
+		return fmt.Errorf("updating Parameter %s: %w", a.id, err)
+	}
+
+	log.V(2).Info("successfully updated Parameter", "name", a.id)
+
+	mapCtx := &direct.MapContext{}
+
+	status := &krm.ParameterManagerParameterStatus{}
```
maqiuyujoyce, line 248: You'll want to cover the acquisition use case here as well. See https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/alloydb/instance_controller.go#L327 as an example.

tests/e2e/normalize.go
```diff
@@ -224,6 +224,9 @@ func buildKRMNormalizer(t *testing.T, u *unstructured.Unstructured, project test
 	// Specific to Secret Manager
 	visitor.replacePaths[".spec.expireTime"] = "2025-10-03T15:01:23Z"
 
+	// Specific to Parameter Manager
+	visitor.replacePaths[".status.observedState.policyMember.iamPolicyUidPrincipal"] = "principal://parametermanager.googleapis.com/projects/${projectNumber}/uid/locations/global/parameters/${uniqueId}"
```
maqiuyujoyce: Could you move the normalization code to be under the corresponding mock folder? See https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/mockgcp/mockiam/normalize.go as an example.

tests/e2e/normalize_legacy.go
```diff
@@ -249,6 +249,10 @@ func LegacyNormalize(t *testing.T, h *create.Harness, project testgcp.GCPProject
 	addReplacement("response.executionStatus.updateTime", "2024-04-01T12:34:56.123456Z")
 	addReplacement("response.executionStatus.latestJob.uid", "0123456789abcdef")
 	addReplacement("executionStatus.latestJob.uid", "0123456789abcdef")
+
+	// Specific to Parameter Manager
```
maqiuyujoyce: ditto

pkg/controller/direct/parametermanager/parameter_controller.go
dhavalbhensdadiya-crest: Hello @maqiuyujoyce ,
I have made this to be similar to the Parameter Manager sdk ([parametermanagerpb](https://pkg.go.dev/cloud.google.com/go/parametermanager/apiv1/parametermanagerpb#Parameter)). In case format is not specified, it defaults to UNFORMATTED for the end user.

Do let me know if you would like for the KCC codebase to make the format mandatory. Thanks

pkg/controller/direct/parametermanager/parameter_controller.go
dhavalbhensdadiya-crest: Hi @maqiuyujoyce ,
I am a bit confused about this part. Since the observed state and the externalref are already set in the Create method, is this extra check required in the update method? Since this is new resource, we would not have the case of customer switching to Direct controller.

pkg/controller/direct/parametermanager/parameter_controller.go
dhavalbhensdadiya-crest: Hi @maqiuyujoyce ,
I have moved all the common logic to AdapterForObject. This is similar to `pkg/controller/direct/batch/job_controller.go`, and  `pkg/controller/direct/dataplex/dataplexlake_controller.go`

apis/parametermanager/v1alpha1/parameter_reference.go
dhavalbhensdadiya-crest: Updated. Thanks

config/samples/resources/parametermanagerparameter/globalparameter/parametermanager_v1alpha1_parametermanagerparameter.yaml
dhavalbhensdadiya-crest: Updated. Thanks

config/samples/resources/parametermanagerparameter/regionalparameter/parametermanager_v1alpha1_parametermanagerparameter_regional.yaml
dhavalbhensdadiya-crest: Added manifests to create kmscryptokey-sample

pkg/controller/direct/parametermanager/parameter_controller.go
dhavalbhensdadiya-crest: Updated

pkg/controller/direct/parametermanager/parameter_controller.go
dhavalbhensdadiya-crest: Removed the check

pkg/controller/direct/parametermanager/parameter_controller.go
dhavalbhensdadiya-crest: Removed the check

tests/e2e/normalize.go
dhavalbhensdadiya-crest: Done

tests/e2e/normalize_legacy.go
dhavalbhensdadiya-crest: Done

pkg/controller/direct/parametermanager/parameter_controller.go
```diff
@@ -0,0 +1,292 @@
...
+
+func (m *modelParameter) client(ctx context.Context, location string) (*gcp.Client, error) {
+	var opts []option.ClientOption
+	opts, err := m.config.RESTClientOptions()
+	if err != nil {
+		return nil, err
+	}
+
+	// Add regional endpoint if region is specified
+	if location != "" && location != "global" {
```
maqiuyujoyce, line 63: Just curious, what will be the error like if the region is not valid but still gets used to construct the endpoint? Asking because if the error message coming back from `parametermanager.INVALID_REGION.rep.googleapis.com:443` is too confusing then we may want to do some basic validation of the region here.

pkg/controller/direct/parametermanager/parameter_controller.go
maqiuyujoyce: According to the protobuf, the default value is not `unformatted`, but `format_unspecified`: https://github.com/googleapis/googleapis/blob/master/google/cloud/parametermanager/v1/service.proto#L150

For enum mapping, you don't need to set zero value here, because [Enum_ToProto function](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/maputils.go#L93) used in the [mapper](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/parametermanager/parameter_mapper.go#L35) handles it properly.

pkg/controller/direct/parametermanager/parameter_controller.go
maqiuyujoyce: To call normalizeKMSKeyRef() here successfully, the dependent KMSCryptoKey must exist. However, during deletion time, the KMSCryptoKey may be deleted faster than this ParameterManagerParameter, and we don't want to be blocked on the non-existent KMSCryptoKey.

That's why we only want to call this normalize function in the beginning of Create() and Update().

And thank you for bringing up the example you checked. We will need to clean them up whenever possible.

pkg/controller/direct/parametermanager/parameter_controller.go
maqiuyujoyce: > Since the observed state and the externalref are already set in the Create method

This is not always true. We support both resource creation and [acquisition](https://docs.cloud.google.com/config-connector/docs/how-to/managing-deleting-resources#acquiring_an_existing_resource). When acquiring an existing resource (i.e. first reconciliation after creating the KRM object), controller calls Update() instead of Create(), and we still need to set the status even if we don't do any update.

tests/e2e/normalize_legacy.go
```diff
@@ -249,6 +249,7 @@ func LegacyNormalize(t *testing.T, h *create.Harness, project testgcp.GCPProject
 	addReplacement("response.executionStatus.updateTime", "2024-04-01T12:34:56.123456Z")
 	addReplacement("response.executionStatus.latestJob.uid", "0123456789abcdef")
 	addReplacement("executionStatus.latestJob.uid", "0123456789abcdef")
+
```
maqiuyujoyce: Nit: Clean up the newline here.

maqiuyujoyce (commented):
Thanks for addressing the comments! Let me know if you still have any questions.

pkg/controller/direct/parametermanager/parameter_controller.go
dhavalbhensdadiya-crest: Update call failed: error creating: creating Parameter projects/migrationsource-392805/locations/us-cen1/parameters/global-1: Post "https://parametermanager.us-cen1.rep.googleapis.com:443/v1/projects/migrationsource-392805/locations/us-cen1/parameters?%24alt=json%3Benum-encoding%3Dint&parameterId=global-1": dial tcp: lookup parametermanager.us-cen1.rep.googleapis.com on 34.118.224.10:53: no such host

pkg/controller/direct/parametermanager/parameter_controller.go
dhavalbhensdadiya-crest: Hello @maqiuyujoyce ,
`format_unspecified` seems to be an internal only value. The Parameter Manager UI does not have any such option. Also, if we create any parameter without providing any format, it will be passed as `format_unspecified`, but the parameter that is created, will have a format value of `unformatted`. Due to this, when we run any unrelated update operation (e.g. labels), the update call will fail as it will try to update the format from `unformatted` to `format_unspecified` which is not allowed.

I tried it out, and I get an error as shown below. In my opinion, we should default the format to "unformatted" if it is not specified to avoid this issue. Let me know your thoughts. Thanks

Normal   UpToDate      51s   parametermanagerparameter-controller  The resource is up to date
  Warning  UpdateFailed  1s    parametermanagerparameter-controller  Update call failed: error updating: updating Parameter projects/migrationsource-392805/locations/global/parameters/global-1: googleapi: Error 400: Request is trying to modify an immutable field format.
error details: name = ErrorInfo reason = IMMUTABLE_FIELD domain = parametermanager.googleapis.com metadata = map[path:format]
error details: name = LocalizedMessage locale = en-US msg = Request is trying to modify an immutable field format.

pkg/controller/direct/parametermanager/parameter_controller.go
dhavalbhensdadiya-crest: Thanks for the explanation. Updated as requested.

pkg/controller/direct/parametermanager/parameter_controller.go
dhavalbhensdadiya-crest: Thanks for the explanation. Updated as requested.

pkg/controller/direct/parametermanager/parameter_controller.go
dhavalbhensdadiya-crest: Thanks for the explanation. Updated as requested.

tests/e2e/normalize_legacy.go
dhavalbhensdadiya-crest: updated

pkg/controller/direct/parametermanager/parameter_controller.go
maqiuyujoyce: Thanks a lot for experimenting with it. Let's start with no validation for now, and if we receive feedback about this error message being confusing then we can make improvements accordingly.

pkg/controller/direct/parametermanager/parameter_controller.go
maqiuyujoyce: Thank you for all the details! This indeed look like an important issue in this controller so let's address it before we merge the PR.

First, in general, we want to avoid a customized defaulting logic in the controller unless we confirm it is a true edge case.

Then, about the Update error.
> when we run any unrelated update operation (e.g. labels), the update call will fail as it will try to update the format from unformatted to format_unspecified which is not allowed.

If a field is unset in the desired manifest, we'll use the live state of that field to compare the diff. So this error basically means one of the following based on the update logic:

(1) The returned `paths` in `paths, err = common.CompareProtoMessage(a.desired, a.actual, common.BasicDiff)` incorrectly contains the `format` field. <- If this is the root cause, then we can customize the diffing logic to address it.

(2) The API didn't properly filter fields to be updated using the values in `updateMask`. <- If this is the root cause, then it is a true edge case.

Could you print out the value of `updateMask` before it is passed to the Update Request so that we understand what's wrong here? E.g. with `log.Info("updateMask before sending to the API", "updateMask", fmt.Sprintf("%+v", updateMask))` you can find this log in the pod log.

maqiuyujoyce (commented):
Thanks for the addressing the comments! The PR is almost ready to go though I have one concern about the update logic for the `format` field.

pkg/controller/direct/parametermanager/parameter_controller.go
dhavalbhensdadiya-crest: Hello @maqiuyujoyce ,
I have done this analysis previously. The order of operations is as follows:

1. User creates a manifest file where format is not specified
2. Since format is not specified, the generated proto files default to the format "PARAMTER_FORMAT_UNSPECIFIED"
3. When the parameter is created, it is created with the format "UNFORMATTED". This is because "PARAMTER_FORMAT_UNSPECIFIED" is not an actual format and the actual default format of the created parameter is "UNFORMATTED"
4. User updates the Labels in the manifest file
5. The diffing mechanism compares the GCP Object with the manifest file and it will identify that Labels and Format have been updated. This is because the generated proto file will have the format as "PARAMTER_FORMAT_UNSPECIFIED" and the actual GCP object will have the format "UNFORMATTED". 
6. Update will fail because the format is a non mutable field.

Solutions:
1. If format is not specified, use "UNFORMATTED" by default. This matches to what actually happens to a parameter in GCP. This will also work for acquisition because while format is optional for user to specify, a parameter created in GCP will always have a format. This is the solution I had implemented previously.
2. Update the current diffing mechanism. Instead of comparing the complete objects, compare only labels and kms keys which are mutable objects. I have made a commit with this solution as I understand this is your preference. Do let me know if there is any other solution or if I misunderstood your comment about custom diffing logic. Thanks

pkg/controller/direct/parametermanager/parameter_controller.go
```diff
@@ -203,12 +202,13 @@ func (a *ParameterAdapter) Update(ctx context.Context, updateOp *directbase.Upda
...
-	paths, err = common.CompareProtoMessage(resource, a.actual, common.BasicDiff)
-	if err != nil {
-		return err
+	paths := []string{}
+	if !reflect.DeepEqual(a.actual.GetKmsKey(), resource.GetKmsKey()) {
+		paths = append(paths, "kms_key")
+	}
+	if !reflect.DeepEqual(a.actual.GetLabels(), resource.GetLabels()) {
+		paths = append(paths, "labels")
 	}
```
maqiuyujoyce, line 219: I think we still want to compare the live state and desired state of the `format` field here. We shouldn't silently ignore the change of `format` field.

dhavalbhensdadiya-crest: Hi @maqiuyujoyce ,
This specific code change is related to the discussion about not having a default value for format in the controller logic. There is an edge case in the parameter manager logic, where it has 4 ENUM values in the proto file, but a created parameter can only have 3 ENUM values. Due to this when a format value is not specified, the proto will default to "PARAMTER_FORMAT_UNSPECIFIED", but the gcp will default to "UNFORMATTED" as it converts "PARAMTER_FORMAT_UNSPECIFIED" to "UNFORMATTED" when the parameter is created. This will lead to a format mismatch in update scenario. Linking the thread which has a more detailed discussion. - https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5893#discussion_r2785256470

maqiuyujoyce: Thank you for the elaboration! Though the way we compare the diffs between the desired state and the actual state is: We only do comparison when the desired state is configured. So for example, for the format field, I'd do something like:

```
if desired.Spec.Format != nil {
	if !reflect.DeepEqual(a.actual.GetFormat(), resource.GetFormat()) {
		paths = append(paths, "format")
	}
}
```

And the same should be applied to other spec fields.

dhavalbhensdadiya-crest: Thanks @maqiuyujoyce . This method will work.

1. Format - Will check for nil before comparison
2. KMSKeyRef - Should not check for nil. nil is a valid choice which means customer would like to opt out of the custom kms key usage
3. Labels - Not a spec field. Obtained for metadata. 

Hence, adding a check for format

dhavalbhensdadiya-crest: Hi @maqiuyujoyce ,
I have updated the code to check for format and tested it. Thanks

Issue #5894, opened 2025-12-23
Code refactor: Do not override logger name with WithName


justinsb:
Particularly in our direct controllers, we should not override the logger name (using WithName).

So when we see \`log := klog.FromContext(ctx).WithName(ctrlName)\`, that should simply be \`log := klog.FromContext(ctx)\`.

Calling \`WithName\` extends/replaces the name in the context, and the name in the context is already the controller name.

PR #5896, opened 2025-12-23
chore: create linter to avoid log.FromContext(ctx).WithName()

justinsb:
Issue #5894 , but adding a linter.


anhdle-sso:
+1 on the change. I also want to add a linter that make sure the users add klog Verbosity level when calling log.Info. This would help filter out not needed log. WDYT?

PR #5906, opened 2025-12-23
mockgcp: move container to httptogrpc

justinsb:
- **mockgcp: move container to httptogrpc**


justinsb:
Let's do this in #8298

/close

PR #5910, opened 2025-12-24
chore: update GCP dependencies

justinsb:
- **chore: don't try to autoupdate kompanion or testdata**
  

- **autogen: dev/tasks/update-gcp-dependencies**


justinsb:
/assign @cheftako 

This shows how we detect fields today.  To automate adding the fuzz ignore lines, I think we want to use a technique like I am proposing in #5897

PR #5916, opened 2025-12-24
chore: fix github-mcp-server invocation

justinsb:


dev/mcp/github-mcp-server
```diff
@@ -0,0 +1,56 @@
...
+    GOWORK=off go build -o ${MCP_SERVER_PATH} ./cmd/github-mcp-server > /dev/null 2>&1
+fi
+
+# Ensure GITHUB_PERSONAL_ACCESS_TOKEN is set, using gh CLI
+if [[ -z "${GITHUB_PERSONAL_ACCESS_TOKEN:-}" ]]; then
+    GITHUB_PERSONAL_ACCESS_TOKEN=$(gh auth token)
+    export GITHUB_PERSONAL_ACCESS_TOKEN
+fi
+
+exec ${MCP_SERVER_PATH} $@
```
cheftako, line 56: AGENT: It is safer to quote the variables here to ensure arguments with spaces are handled correctly:
```bash
exec "${MCP_SERVER_PATH}" "$@"
```

TAKO: Wondering if this suggests we should add one of the shell linters. Pretty sure they also complain on this. Lets deal with this in a follow up.

dev/mcp/github-mcp-server
```diff
@@ -0,0 +1,56 @@
...
+    git checkout ${VERSION} > /dev/null 2>&1
+    cd ..
+    else
+    git clone --branch ${VERSION} --single-branch --depth 1 https://github.com/github/github-mcp-server.git > /dev/null 2>&1
+    fi
+
+    cd github-mcp-server
+
+    # Build the github-mcp-server binary
+    GOWORK=off go build -o ${MCP_SERVER_PATH} ./cmd/github-mcp-server > /dev/null 2>&1
```
cheftako, line 47: Suppressing all output with `> /dev/null 2>&1` will make it difficult to diagnose build failures. Consider removing this redirection or only redirecting stdout, letting stderr show through.

dev/mcp/github-mcp-server
```diff
@@ -0,0 +1,56 @@
...
+fi
+
+# Ensure GITHUB_PERSONAL_ACCESS_TOKEN is set, using gh CLI
+if [[ -z "${GITHUB_PERSONAL_ACCESS_TOKEN:-}" ]]; then
+    GITHUB_PERSONAL_ACCESS_TOKEN=$(gh auth token)
```
cheftako, line 52: AGENT: It might be helpful to check if the `gh` CLI is installed (e.g., using `command -v gh`) before running it. This would allow for a clear error message if the tool is missing, rather than a potentially confusing shell error.

.gemini/settings.json
```diff
@@ -2,12 +2,12 @@
...
     "mcpServers": {
         "github": {
-            "command": "/server/github-mcp-server",
+            "command": "dev/mcp/github-mcp-server",
             "args": [
                 "stdio"
             ],
             "env": {
-                "GITHUB_READ-ONLY": "1"
+                "GITHUB_READ_ONLY": "1"
```
cheftako, line 10: AGENT: Good fix on the environment variable name. `GITHUB_READ_ONLY` is the correct convention and likely what the server expects.

TAKO: Looks like they made a backward incompatible change with https://github.com/github/github-mcp-server/pull/1067. Given that this is now correct.

dev/mcp/github-mcp-server
```diff
@@ -0,0 +1,56 @@
...
+MCP_SERVER_PATH="${BINDIR}/github-mcp-server"
+
+if [[ ! -f "${MCP_SERVER_PATH}" ]]; then
+    mkdir -p .build/mcp/src
+    cd .build/mcp/src
+
+    # Check out specific version of github-mcp-server
+    if [ -d "github-mcp-server" ]; then
+    cd github-mcp-server
+    git fetch origin ${VERSION} > /dev/null 2>&1
```
cheftako, line 37: AGENT: Since `VERSION` appears to be a tag, using `git fetch origin tag ${VERSION}` could be more specific, ensuring only the tag is fetched. However, the current command is also functional.

dev/mcp/github-mcp-server
```diff
@@ -0,0 +1,56 @@
...
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+cd ${REPO_ROOT}
+
+VERSION="v0.26.3"
```
cheftako, line 23: AGENT: Hardcoding the version here is fine for now, but if this tool becomes widely used, consider moving this to a configuration file or environment variable to make updates easier without modifying the script.

.gemini/settings.json
```diff
@@ -2,12 +2,12 @@
 {
     "mcpServers": {
         "github": {
-            "command": "/server/github-mcp-server",
```
cheftako, line 5: @barney-s I originally had this set based on the sandbox path. Sandbox is no longer functional but not sure if this affects the repo-agent.

cheftako (commented):
AGENT: Overall, this is a solid improvement to the developer workflow. The wrapper script abstracts away the binary management effectively.

A few minor suggestions to improve robustness:
1.  Quote variables in the `exec` command to safely handle arguments with spaces.
2.  Avoid suppressing build output to aid in debugging if the build fails.
3.  Add a check for the `gh` CLI availability.

Issue #5915, opened 2025-12-24
mockgcp netapp: move to cloud.google.com/go and httptogrpc


justinsb:
Sub-task of #5913 for mockgcp/mocknetapp

* Move to cloud.google.com/go/netapp (replace generated proto binding imports with "official" cloud.google.com/go proto imports)
* Replace grpc-gateway with httptogrpc

PR #5918, opened 2025-12-24
netapp backup policy: fixes to mockgcp and controller defaulting

justinsb:
- **mockgcp netapp: move to cloud.google.com/go and httptogrpc**
  

- **chore: improve service discovery in httptogrpc**
  

- **mockgcp: more fidelity for netapp backuppolicy**
  

- **fix: implement defaults for netapp backuppolicy**

### DIFF
--- a/pkg/controller/direct/netapp/backuppolicy_controller.go
+++ b/pkg/controller/direct/netapp/backuppolicy_controller.go
@@ -171,9 +171,7 @@ func (a *BackupPolicyAdapter) Update(ctx context.Context, updateOp *directbase.U
 	}
 	desiredPb.Name = a.id.String()
 
-	paths := make(sets.Set[string])
-	var err error
-	paths, err = common.CompareProtoMessage(desiredPb, a.actual, common.BasicDiff)
+	paths, err := computeDiffBackupPolicy(mapCtx, a.actual, desiredPb)
 	if err != nil {
 		return err
 	}
@@ -183,7 +181,7 @@ func (a *BackupPolicyAdapter) Update(ctx context.Context, updateOp *directbase.U
 		return nil
 	}
 	updateMask := &fieldmaskpb.FieldMask{
-		Paths: sets.List(paths),
+		Paths: paths,
 	}
 
 	req := &netapppb.UpdateBackupPolicyRequest{
@@ -257,4 +255,24 @@ func (a *BackupPolicyAdapter) Delete(ctx context.Context, deleteOp *directbase.D
 		return false, fmt.Errorf("waiting delete BackupPolicy %s: %w", a.id, err)
 	}
 	return true, nil
+}
+
+func computeDiffBackupPolicy(mapCtx *direct.MapContext, actual, desired *netapppb.BackupPolicy) ([]string, error) {
+	desired = direct.ProtoClone(desired)
+	actual = direct.ProtoClone(actual)
+
+	populateServerSideDefaults := func(o *netapppb.BackupPolicy) {
+		if o.Enabled == nil {
+			o.Enabled = direct.PtrTo(true)
+		}
+	}
+
+	populateServerSideDefaults(desired)
+	populateServerSideDefaults(actual)
+
+	paths, err := common.CompareProtoMessage(desired, actual, common.BasicDiff)
+	if err != nil {
+		return nil, err
+	}
+	return sets.List(paths), nil
 }
--- a/pkg/test/resourcefixture/testdata/basic/netapp/v1alpha1/netappbackuppolicy/netappbackuppolicy-basic/_http.log
+++ b/pkg/test/resourcefixture/testdata/basic/netapp/v1alpha1/netappbackuppolicy/netappbackuppolicy-basic/_http.log
@@ -126,7 +126,7 @@ X-Xss-Protection: 0
 
 ---
 
-PATCH https://netapp.googleapis.com/v1/projects/${projectId}/locations/us-west2/backupPolicies/netappbackuppolicy-minimal-${uniqueId}?%24alt=json%3Benum-encoding%3Dint&updateMask=dailyBackupLimit%2Cenabled
+PATCH https://netapp.googleapis.com/v1/projects/${projectId}/locations/us-west2/backupPolicies/netappbackuppolicy-minimal-${uniqueId}?%24alt=json%3Benum-encoding%3Dint&updateMask=dailyBackupLimit
 Content-Type: application/json
 User-Agent: kcc/${kccVersion} (+https://github.com/GoogleCloudPlatform/k8s-config-connector) kcc/controller-manager/${kccVersion}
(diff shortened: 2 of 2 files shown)


justinsb:
This is issue #5915

It builds on #5917

pkg/test/resourcefixture/testdata/basic/netapp/v1alpha1/netappbackuppolicy/netappbackuppolicy-basic/_http.log
yuwenma, line 1: The  http log looks so different. It seems to be a big improvements about defaulting 👍 . However, I don't fully get how it ends up with some additional GET calls. Could you elaborate on that? Also, would be great if you can upload the real gcp log (or any other approach) to verify the change better matches the real log.

pkg/controller/direct/netapp/backuppolicy_controller.go
```diff
@@ -258,3 +256,23 @@ func (a *BackupPolicyAdapter) Delete(ctx context.Context, deleteOp *directbase.D
 	}
 	return true, nil
 }
+
+func computeDiffBackupPolicy(mapCtx *direct.MapContext, actual, desired *netapppb.BackupPolicy) ([]string, error) {
+	desired = direct.ProtoClone(desired)
+	actual = direct.ProtoClone(actual)
+
+	populateServerSideDefaults := func(o *netapppb.BackupPolicy) {
```
cheftako, line 264: AGENT: `mockgcp` (and presumably the real API) appears to default `Description`, `MonthlyBackupLimit`, and `WeeklyBackupLimit` as well.
Should we include them here to ensure we don't detect a spurious diff if the server defaults these fields and the KRM object leaves them unset?

mockgcp/common/httptogrpc/service.go
```diff
@@ -23,11 +23,68 @@ import (
...
+	}
+}
+
+type serviceOptions struct {
+	ServiceName string
+}
+
+// addServiceWithOptions adds a gRPC service (with all the methods) to the mux.
+func (m *grpcMux) addServiceWithOptions(client any, options serviceOptions) {
+	if options.ServiceName == "" {
```
cheftako, line 53: AGENT: This calls `klog.Fatalf` which will terminate the test runner/process.
Since this is a library function, it might be safer to return an error and let the caller decide how to handle it, even if `AddService` signature doesn't currently support error returns (perhaps the interface needs updating or this helper should verify before adding).

TAKO: Been through this before. This is in a mock, as such its test code. Not concerned about it affecting the live system. As before mild concern about it making some failures harder to debug. However not a blocking issue.

mockgcp/common/httptogrpc/service.go
```diff
@@ -54,49 +110,67 @@ func (m *grpcMux) AddService(client any) {
...
 
-	s, err := newGRPCService(client, matchingServices[0])
-	if err != nil {
-		klog.Fatalf("adding grpc service: %v", err)
+	for k := range matchingServices {
+		return k
 	}
 
-	m.services = append(m.services, s)
+	klog.Fatalf("cannot match service for %T", client)
```
cheftako, line 172: AGENT: Similar to above, `klog.Fatalf` here is quite aggressive.
If `mockgcp` is used in a context where we want to recover or fail a single test instead of the whole suite, `Fatalf` is problematic.

TAKO: Similar to above.

mockgcp/mocknetapp/backuppolicy.go
```diff
@@ -46,6 +47,9 @@ func (s *backupVaultsService) GetBackupPolicy(ctx context.Context, req *pb.GetBa
 
 	obj := &pb.BackupPolicy{}
 	if err := s.storage.Get(ctx, fqn, obj); err != nil {
+		if status.Code(err) == codes.NotFound {
+			return nil, status.Errorf(codes.NotFound, "Resource '%v' was not found", fqn)
```
cheftako, line 51: AGENT: The error message format has changed from `backupPolicy "..." not found` to `Resource '...' was not found`.
Please verify if this matches the actual Google Cloud NetApp API error message format exactly, as tests might rely on string matching.

TAKO: Again this is in the mock code. I agree that it needs to match GCP API. Assuming it does, thats why we made the change. Given that any code controller or test which relies on the old string needs to be fixed.

mockgcp/common/httptogrpc/dynamicgrpcgateway.go
```diff
@@ -34,7 +34,7 @@ import (
 // Mux is the primary interface for mapping HTTP requests to gRPC method calls.
 type Mux interface {
 	http.Handler
-	AddService(client any)
+	AddService(client any, opt ...ServiceOption)
```
cheftako, line 37: AGENT: Modifying the `Mux` interface to accept `opt ...ServiceOption` is good for extensibility.
Ensure all existing implementations of `Mux` (if any others exist) are updated.

mockgcp/mocknetapp/backuppolicy.go
```diff
@@ -121,6 +128,21 @@ func (s *backupVaultsService) UpdateBackupPolicy(ctx context.Context, req *pb.Up
...
+	if obj.Description == nil {
+		obj.Description = mocks.PtrTo("")
+	}
+	if obj.Enabled == nil {
+		obj.Enabled = mocks.PtrTo(true)
+	}
+	if obj.MonthlyBackupLimit == nil {
+		obj.MonthlyBackupLimit = mocks.PtrTo(int32(0))
+	}
+	if obj.WeeklyBackupLimit == nil {
```
cheftako, line 141: AGENT: Does `populateDefaultsForBackupPolicy` need to be called in `UpdateBackupPolicy` as well?
If the user sends an update with some fields unset, does the server apply defaults or retain existing values?
If it retains existing, then no change needed. If it resets to default, we might need it.

pkg/controller/direct/netapp/backuppolicy_controller.go
```diff
@@ -258,3 +256,23 @@ func (a *BackupPolicyAdapter) Delete(ctx context.Context, deleteOp *directbase.D
...
+		if o.Enabled == nil {
+			o.Enabled = direct.PtrTo(true)
+		}
+	}
+
+	populateServerSideDefaults(desired)
+	populateServerSideDefaults(actual)
```
cheftako, line 271: TAKO: Can we chat on this? Having to populate defaults on both desired and actual really starts feeling like we are making a mistake. I can't help this should be handled by making sure the diff does the right thing, rather than fudging the data to try to get the diff to return the right answer.

cheftako (commented):
AGENT: The move to `cloud.google.com/go` and `httptogrpc` for NetApp is a great improvement for maintainability and consistency. The improved LRO simulation (using `StartLRO` instead of `DoneLRO`) adds valuable fidelity to the mocks, which explains the updated HTTP logs.

I have a few comments regarding the robustness of the new `httptogrpc` service discovery and the completeness of the controller's defaulting logic.

PR #5919, opened 2025-12-24
chore: improve service discovery in httptogrpc

justinsb:
This is part of #5918 (which motivates it), but our automatic service discovery could not "find" the proto service when the service was in a different file to the messages.

### DIFF
--- a/mockgcp/common/httptogrpc/dynamicgrpcgateway.go
+++ b/mockgcp/common/httptogrpc/dynamicgrpcgateway.go
@@ -34,7 +34,7 @@ import (
 // Mux is the primary interface for mapping HTTP requests to gRPC method calls.
 type Mux interface {
 	http.Handler
-	AddService(client any)
+	AddService(client any, opt ...ServiceOption)
 }
 
 // grpcMux implements Mux.
--- a/mockgcp/common/httptogrpc/service.go
+++ b/mockgcp/common/httptogrpc/service.go
@@ -23,11 +23,68 @@ import (
 	"google.golang.org/genproto/googleapis/api/annotations"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/reflect/protoregistry"
 	"k8s.io/klog/v2"
 )
 
 // AddService adds a gRPC service (with all the methods) to the mux.
-func (m *grpcMux) AddService(client any) {
+func (m *grpcMux) AddService(client any, options ...ServiceOption) {
+	var opt serviceOptions
+	for _, option := range options {
+		option(&opt)
+	}
+	m.addServiceWithOptions(client, opt)
+}
+
+type ServiceOption func(*serviceOptions)
+
+func WithServiceName(name string) ServiceOption {
+	return func(o *serviceOptions) {
+		o.ServiceName = name
+	}
+}
+
+type serviceOptions struct {
+	ServiceName string
+}
+
+// addServiceWithOptions adds a gRPC service (with all the methods) to the mux.
+func (m *grpcMux) addServiceWithOptions(client any, options serviceOptions) {
+	if options.ServiceName == "" {
+		options.ServiceName = findServiceName(client)
+	}
+
+	serviceName := protoreflect.FullName(options.ServiceName)
+
+	var matchingServices []protoreflect.ServiceDescriptor
+	protoregistry.GlobalFiles.RangeFiles(func(fd protoreflect.FileDescriptor) bool {
+		services := fd.Services()
+		for i := range services.Len() {
+			service := services.Get(i)
+			if service.FullName() != serviceName {
+				continue
+			}
+			matchingServices = append(matchingServices, service)
+		}
+		return true
+	})
+
+	if len(matchingServices) == 0 {
+		klog.Fatalf("could not find proto service %q", serviceName)
+	}
+
+	if len(matchingServices) > 1 {
+		klog.Fatalf("found multiple matching services %q", serviceName)
+	}
+
+	s, err := newGRPCService(client, matchingServices[0])
+	if err != nil {
+		klog.Fatalf("adding grpc service: %v", err)
+	}
+	m.services = append(m.services, s)
+}
+
+func findServiceName(client any) string {
 	var protoMessage proto.Message
 	protoMessageType := reflect.TypeOf(&protoMessage).Elem()
 
@@ -36,8 +93,7 @@ func (m *grpcMux) AddService(client any) {
 	goTypeMethodNames := make(map[string]bool)
 
 	var discoveredProtobufTypes []reflect.Type
-	n := goType.NumMethod()
-	for i := 0; i < n; i++ {
+	for i := range goType.NumMethod() {
 		method := goType.Method(i)
 		goTypeMethodNames[method.Name] = true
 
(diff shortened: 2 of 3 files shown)


mockgcp/common/httptogrpc/service.go
```diff
@@ -54,49 +110,67 @@ func (m *grpcMux) AddService(client any) {
...
+	matchingServices := make(map[string]protoreflect.ServiceDescriptor)
 	for _, protoType := range discoveredProtobufTypes {
 		msg := reflect.New(protoType).Elem().Interface()
 		md := msg.(proto.Message).ProtoReflect().Descriptor()
-		fd := md.ParentFile()
 
-		services := fd.Services()
-		for i := range services.Len() {
-			service := services.Get(i)
+		protoregistry.GlobalFiles.RangeFilesByPackage(md.ParentFile().FullName(), func(fd protoreflect.FileDescriptor) bool {
```
cheftako, line 118: AGENT: `md.ParentFile().FullName()` returns the file path (e.g., "google/identity/accesscontextmanager/v1/service.proto"), but `protoregistry.GlobalFiles.RangeFilesByPackage` expects a package name (e.g., "google.identity.accesscontextmanager.v1").

Using the file path as the package name will likely result in zero matches, breaking the automatic service discovery for services that don't explicitly provide the name.

Should this be `md.ParentFile().Package()`?

mockgcp/common/httptogrpc/service.go
```diff
@@ -23,11 +23,68 @@ import (
 	"google.golang.org/genproto/googleapis/api/annotations"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/reflect/protoregistry"
 	"k8s.io/klog/v2"
 )
 
 // AddService adds a gRPC service (with all the methods) to the mux.
```
cheftako, line 30: AGENT: The `ServiceOption` pattern is a good addition here. It allows for future extensibility without breaking the `AddService` signature further (though this change itself breaks the interface).

mockgcp/common/httptogrpc/service.go
```diff
@@ -23,11 +23,68 @@ import (
...
+			if service.FullName() != serviceName {
+				continue
+			}
+			matchingServices = append(matchingServices, service)
+		}
+		return true
+	})
+
+	if len(matchingServices) == 0 {
+		klog.Fatalf("could not find proto service %q", serviceName)
```
cheftako, line 73: AGENT: `klog.Fatalf` inside `findServiceName` (which is called by `AddService`) might be too aggressive if this is used as a library. It would panic the entire application if a service cannot be matched. 

While acceptable for a test binary, consider if returning an error up the stack would be better, although `AddService` currently doesn't return an error.

TAKO: I would agree with the agent here IF it was in an actual controller. This appears to be in the mock. A little worried this might be hard to debug but the failure is guaranteed to be in test.

mockgcp/common/httptogrpc/service.go
```diff
@@ -23,11 +23,68 @@ import (
...
+	}
+
+	s, err := newGRPCService(client, matchingServices[0])
+	if err != nil {
+		klog.Fatalf("adding grpc service: %v", err)
+	}
+	m.services = append(m.services, s)
+}
+
+func findServiceName(client any) string {
```
cheftako, line 87: AGENT: The logic here iterates over `discoveredProtobufTypes` and for each type, calls `RangeFilesByPackage`.
If multiple types belong to the same package, this will redundantly scan the same package multiple times.

Consider collecting unique package names first, or using a visited map for packages, to improve startup performance slightly.

mockgcp/common/httptogrpc/service.go
```diff
@@ -54,49 +110,67 @@ func (m *grpcMux) AddService(client any) {
 	}
```
cheftako, line 110: AGENT: The logic `!goTypeMethodNames[string(method.Name())]` assumes a strict 1:1 mapping where every method in the Proto Service MUST exist in the Go client.

Is it guaranteed that the Go client (interface/struct) has exactly the same methods? Sometimes generated clients might have helper methods or the proto might have methods not yet generated/supported?

Just double checking if this strictness is intended.

TAKO: I think this is a pre-existing strictness. I do not believe the PR is making this more strict/worse.

mockgcp/common/httptogrpc/service.go
```diff
@@ -54,49 +110,67 @@ func (m *grpcMux) AddService(client any) {
...
 
-			// TODO: Is there a better way to match this?
-			if isMatch {
-				matchingServices = append(matchingServices, service)
+				// We match by looking at the method names.  There may be a better way to do this, but we haven't found one yet!
+				isMatch := true
+
+				protoMethods := service.Methods()
+				protoMethodNames := make(map[string]bool)
+				for j := range protoMethods.Len() {
```
cheftako, line 133: AGENT: Similarly here, this checks that every method in the Go client exists in the Proto Service.

`reflect.TypeOf(client)` returns the concrete type (not the interface, unless `client` is nil). If the generated client struct has any extra exported methods (e.g. `Close`, `String`), this matching will fail.

mockgcp/common/httptogrpc/dynamicgrpcgateway.go
```diff
@@ -34,7 +34,7 @@ import (
 // Mux is the primary interface for mapping HTTP requests to gRPC method calls.
 type Mux interface {
 	http.Handler
-	AddService(client any)
+	AddService(client any, opt ...ServiceOption)
```
cheftako, line 37: AGENT: This change to `AddService` breaks the `Mux` interface for any external implementations.
If `mockgcp` is consumed by others who implement `Mux`, this is a breaking change.
Given this seems internal to the project's test suite, it's likely fine.

mockgcp/common/httptogrpc/service.go
```diff
@@ -54,49 +110,67 @@ func (m *grpcMux) AddService(client any) {
...
+			return true
+		})
 	}
 
-	if len(matchingServices) == 0 {
-		klog.Fatalf("cannot match service for %v", goType.Name())
-	}
 	if len(matchingServices) > 1 {
-		klog.Fatalf("found multiple matching service for %v", goType.Name())
+		for k, v := range matchingServices {
```
cheftako, line 162: AGENT: If `matchingServices` has > 1 entries, `klog.Infof` prints them.
It would be helpful to also print the `client` type name in the loop or in the final Fatalf message (which you do).

mockgcp/mockaccesscontextmanager/service.go
```diff
@@ -74,7 +74,7 @@ func (s *MockService) NewHTTPMux(ctx context.Context, conn *grpc.ClientConn) (ht
 		return nil, fmt.Errorf("error building grpc service: %w", err)
 	}
 
-	grpcMux.AddService(pb.NewAccessContextManagerClient(conn))
+	grpcMux.AddService(pb.NewAccessContextManagerClient(conn), httptogrpc.WithServiceName("google.identity.accesscontextmanager.v1.AccessContextManager"))
```
cheftako, line 77: AGENT: Good usage of the new option. This bypasses the potentially buggy `findServiceName` logic for this specific service.

cheftako (commented):
AGENT: The introduction of `ServiceOption` is a great improvement for flexibility.
However, I have a strong concern about the usage of `RangeFilesByPackage` with `fd.FullName()`.
Please verify the detailed comments below.

mockgcp/common/httptogrpc/service.go
justinsb: Yeah, I don't love it, but because it's part of mock initialization it feels like if something goes wrong here it is a code bug - there should be no external dependencies (e.g. network calls)

mockgcp/common/httptogrpc/service.go
justinsb: We have to match go methods to proto methods.  Strict works.  We should ask the agent to send a PR to make it better :-)

mockgcp/common/httptogrpc/service.go
justinsb: Correct.  But it works, and there is a comment saying it's not ideal, just the best way we know of.

mockgcp/common/httptogrpc/dynamicgrpcgateway.go
justinsb: Right, I don't think httptogrpc is yet being used outside of KCC.  Also maybe this shouldn't even be an interface.

mockgcp/mockaccesscontextmanager/service.go
justinsb: Right, that's the plan.  I want to start encouraging passing the servicename, but it should be optional right now.  We could make it mandatory, I guess, it shouldn't be too hard to specify.

mockgcp/common/httptogrpc/service.go
justinsb: From https://pkg.go.dev/google.golang.org/protobuf/reflect/protoreflect#Descriptor: `// FullName is the fully-qualified name of the declaration.`

So I think the assumption that FullName returns a path is incorrect.

`md.ParentFile().Package()` might be clearer though.

mockgcp/common/httptogrpc/service.go
justinsb: The idea is that we print the client type name once, at the end.  Printing the client type every time was just TMI

mockgcp/common/httptogrpc/service.go
justinsb: I think I'll do a follow up to optimize the efficiency of findServiceName

PR #5921, opened 2025-12-26
release improvements

xiaoweim:
## Release Script Improvements (`generate-release.sh`)

*Version Validation*: Added validation to ensure NEW_VERSION is strictly greater than STALE_VERSION to prevent accidental downgrades or duplicate versions.
*Branch Management*: Added checks for existing release branches with an interactive prompt to delete/recreate them if needed.
*Robustness*: Improved error handling for unit tests and added retries after golden file updates. Fixed directory context issues when running go test and `make fmt`.

## Tagging & Diff enhancements

*Manual Recovery*: Updated `auto-tag-on-version-update.sh` to output helpful manual tagging commands when commit message verification fails.
`Flexible Commit Parsing`: Updated `generate-crd-diff.sh` to handle various combinations of optional release commits (e.g., handling missing "Format code" or "Update golden files" commits gracefully).

## Refactoring

   * Moved `filter_version_changes.py` and `deploy-release-manifest` from `dev/tasks/ to dev/release/` to centralize all release-related logic in one directory.

Verification was done via the following ways:
  * Verified that `generate-release.sh` correctly fails when `NEW_VERSION <= STALE_VERSION`.
  * Verified that `generate-crd-diff.sh` identifies the correct CRD commit hash across different commit sequences.
  * Verified file paths after moving scripts to `dev/release/`

```release-note
None
```

```docs

```

#### Intended Milestone

### Tests you have done


dev/release/generate-release.sh
```diff
@@ -24,9 +24,39 @@ if [ -z "$NEW_VERSION" ]; then
...
+# We expect STALE_VERSION to be smaller, so it should be first in sort -V output.
+SORTED_VERSIONS=$(printf "%s\n%s" "$NEW_VERSION" "$STALE_VERSION" | sort -V)
+SMALLEST_VERSION=$(echo "$SORTED_VERSIONS" | head -n1)
+
+if [ "$SMALLEST_VERSION" == "$NEW_VERSION" ]; then
+     echo "Error: New version ($NEW_VERSION) is not greater than stale version ($STALE_VERSION)."
+     exit 1
+fi
+
 # Step 2: Create Release Branch
```
maqiuyujoyce, line 47: Agent:
>Consider validating that `NEW_VERSION` conforms to a valid format (e.g., semantic versioning) or safe git branch name characters. This prevents invalid branch names or issues in downstream commands if the input contains unexpected characters.

Nice to have.

dev/release/generate-release.sh
```diff
@@ -24,9 +24,39 @@ if [ -z "$NEW_VERSION" ]; then
   exit 1
 fi
 
+if [ "$NEW_VERSION" == "$STALE_VERSION" ]; then
```
maqiuyujoyce, line 32: Just curious: What's the stale version in this context? And do we need to provide it when generating the release? If so, why?

dev/release/generate-release.sh
```diff
@@ -24,9 +24,39 @@ if [ -z "$NEW_VERSION" ]; then
...
+    echo "Error: New version ($NEW_VERSION) is the same as the stale version ($STALE_VERSION)."
+    exit 1
+fi
+
+# sort -V sorts version numbers. If the first line is the new VERSION, then it is not strictly greater than STALE_VERSION.
```
maqiuyujoyce, line 37: From agent:
> `sort -V` is a GNU extension. If this script is expected to run on non-GNU systems (e.g., standard macOS), this might fail or behave unexpectedly. Verify compatibility.

I'm not opinionated about it.

dev/release/generate-release.sh
```diff
@@ -24,9 +24,39 @@ if [ -z "$NEW_VERSION" ]; then
...
+     exit 1
+fi
+
 # Step 2: Create Release Branch
-echo "Creating release branch release-${NEW_VERSION}..."
-git checkout -b "release-${NEW_VERSION}"
+RELEASE_BRANCH="release-${NEW_VERSION}"
+if git rev-parse --verify "${RELEASE_BRANCH}" >/dev/null 2>&1; then
+    echo "Release branch ${RELEASE_BRANCH} already exists."
+    read -p "Do you want to delete the existing branch and recreate it? (y/N): " response
```
maqiuyujoyce: When is it safe to delete and recreate the release branch? I feel this could be a risky decision. Could you share more context?

maqiuyujoyce: From agent:
> This introduces an interactive prompt. If this script is run in a CI environment or non-interactive shell, `read` might fail or hang. Consider checking for an interactive terminal or adding a flag to bypass this prompt.

I think this is a reasonable concern.

dev/release/generate-release.sh
```diff
@@ -24,9 +24,39 @@ if [ -z "$NEW_VERSION" ]; then
...
+    if [[ "$response" =~ ^[Yy]$ ]]; then
```
maqiuyujoyce: From the agent:
> The regex `^[Yy]$` matches only "y" or "Y". It would be more user-friendly to allow "yes" or "Yes" as well (e.g., `^[Yy].*`).

I think this is nice to have.

dev/release/generate-release.sh
```diff
@@ -44,14 +74,18 @@ git commit -m "Update alpha CRDs for Release ${NEW_VERSION}"
 echo "Running unit tests..."
 cd operator
 # We use an if statement to handle the failure case without exiting due to set -e
-if ! (go test ./pkg/controllers/...); then
+if ! (cd operator && go test ./pkg/controllers/...); then
```
maqiuyujoyce, line 73: From the agent:
>You are already in the `operator` directory due to the `cd operator` command earlier in the script (visible in the context). Executing `cd operator` again inside this subshell will fail (unless there is a nested `operator` directory), causing this condition to always evaluate to true (treating it as a test failure). You should remove the `cd operator` from this subshell.

I'm agreeing with the agent unless I missed anything.

dev/tasks/auto-tag-on-version-update.sh
```diff
@@ -82,6 +82,10 @@ if [ "${MSG}" = "${EXPECTED_CRDS}" ]; then
 else
   echo "ERROR: Expected commit message '${EXPECTED_CRDS}' at ${CURRENT_REF}, but found '${MSG}'"
   echo "The release PR must include an '${EXPECTED_CRDS}' commit."
+  echo "Please check manually if the version bump is valid."
```
maqiuyujoyce, line 85: How should the release driver do this manual check?

dev/tasks/auto-tag-on-version-update.sh
```diff
@@ -82,6 +82,10 @@ if [ "${MSG}" = "${EXPECTED_CRDS}" ]; then
...
+  echo "If it is valid, you can create the tag manually:"
+  echo "  git tag v${VERSION}"
+  echo "  git push origin v${VERSION}"
```
maqiuyujoyce, line 88: From the agent:
> Hardcoding `origin` assumes the user's remote is named `origin`. This works for most but might be confusing if the user is working with a different remote setup (e.g. `upstream`).

I'm curious, what this origin should be in this command?

dev/release/generate-release.sh
xiaoweim: We only keep one version of the CRD and other manifests in the KCC repo, and we always keep the latest version. So when we generate a new release, we will remove the older version CRD and manifests to avoid confusion. That is what stale version is.

dev/tasks/auto-tag-on-version-update.sh
xiaoweim: My apologies, this should be `remote`!

dev/release/generate-release.sh
xiaoweim: I would prefer it to always re-create the branch when there is an existing one. Since we are committing all the changes in the release branch, to re-use an existing one, we face the risk of committing residual and unwanted changes together with the correct release changes. It is better to start from fresh and generate all the release changes.

dev/tasks/auto-tag-on-version-update.sh
xiaoweim: The release driver should verify that the CRD versions in the `crds/` folder are all updated to the latest release version, this is the case where the commit to update all the CRD versions is missing and that is a required commit before we generate a new release.

PR #5926, opened 2025-12-29
chore: split mockgcp/git.versions from apis/git.versions

justinsb:
Forcing them to update at the same time makes it harder to update each version.


yuwenma (commented):
/lgtm
/approve
Agree!

PR #5937, opened 2026-01-02
mockgcp: better normalization for sql instance ipAddresses

justinsb:


justinsb:
Issue #5940

mockgcp/mockgcpregistry/interfaces.go
```diff
@@ -79,4 +79,7 @@ type Event interface {
 
 	// VisitResponseStringValues calls the callback for each string-typed value found in the response object (if any)
 	VisitResponseStringValues(callback func(path string, value string))
+
+	// ParseResponseInto attempts to parse the response into the provided dest value
+	ParseResponseInto(dest any) bool
```
yuwenma, line 84: Cool!

mockgcp/mocksql/normalize.go
yuwenma, line 1: Just out of curiosity, how did you discover this case?  It feels so hard to reverse engineer a service full of exceptions. If you have any trick, could you also share that in `mockgcp/docs/ai/create-mock`?

PR #5960, opened 2026-01-06
chore: make output from dev/tasks/generate-types-and-mappers much shorter

justinsb:
This otherwise eats LLM context.

Also encourage the LLM to run the script when changing API types.

### DIFF
--- a/dev/tools/controllerbuilder/pkg/codegen/generatorbase.go
+++ b/dev/tools/controllerbuilder/pkg/codegen/generatorbase.go
@@ -142,7 +142,7 @@ func (f *generatedFile) Write(addCopyright bool, writeEmptyFiles bool) error {
 		return fmt.Errorf("running goimports on %q: %w", p, err)
 	}
 
-	klog.Infof("writing file %v", p)
+	klog.V(2).Infof("writing file %v", p)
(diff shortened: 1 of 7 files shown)


dev/tools/controllerbuilder/pkg/codegen/generatorbase.go
```diff
@@ -142,7 +142,7 @@ func (f *generatedFile) Write(addCopyright bool, writeEmptyFiles bool) error {
 		return fmt.Errorf("running goimports on %q: %w", p, err)
 	}
 
-	klog.Infof("writing file %v", p)
+	klog.V(2).Infof("writing file %v", p)
```
cheftako, line 145: Is V(1) the default? For some reason I thought V(2) was the standard default setting (Eg https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md) ?

justinsb: V(0) is the default in klog.  Most people run e.g. kube-apiserver with v=2, but arguably that's just overkill.

I am using V(1) when it is something "surprising / not the common path" and V(2) when it is just totally normal progress reports.

PR #5961, opened 2026-01-06
chore: compute generation should include mappers

justinsb:
- **chore: compute generation should include mappers**
  

- **autogen: chore: compute generation should include mappers**

### DIFF
--- a/pkg/controller/direct/compute/mapper.generated.go
+++ b/pkg/controller/direct/compute/mapper.generated.go
@@ -21,6 +21,7 @@ package compute
 
 import (
 	pb "cloud.google.com/go/compute/apiv1/computepb"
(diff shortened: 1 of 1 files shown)


pkg/controller/direct/compute/mapper.generated.go
```diff
@@ -107,6 +108,86 @@ func ForwardingruleServiceDirectoryRegistrations_ToProto(mapCtx *direct.MapConte
 	// MISSING: ServiceDirectoryRegion
 	return out
 }
+func Interconnect_FromProto(mapCtx *direct.MapContext, in *pb.Interconnect) *krmcomputev1alpha1.Interconnect {
```
cheftako, line 111: Would be nice to have spaces between functions.

justinsb: Yes, but it's auto-generated so it would be a big diff

PR #5966, opened 2026-01-07
chore: mappers should be aware of proto2

justinsb:
- **chore: mappers should be aware of proto2**
- **chore: remove proto2 special-casing from mappers for compute APIs**


yuwenma (commented):
/approve
/lgtm

Thanks! I like the code-generator + example style. We should make this the standard for AI PR review.

PR #5967, opened 2026-01-07
tests: bail on status leadership mismatch

acpana:
This change builds on #5886 in order to make testing easier!

### DIFF
--- a/pkg/controller/kccmanager/kccmanager_multicluster_test.go
+++ b/pkg/controller/kccmanager/kccmanager_multicluster_test.go
@@ -376,110 +376,133 @@ func TestSplitBrainProtection(t *testing.T) {
 		t.Fatalf("error creating client: %v", err)
 	}
 
-	ns := "test-ns-splitbrain"
-	leaseName := "test-lease"
-	imposterIdentity := "imposter-manager"
-	localIdentity := "my-manager"
-
-	// Create namespace to be safe
-	namespaceObj := &corev1.Namespace{
-		ObjectMeta: metav1.ObjectMeta{Name: ns},
-	}
-	if err := k8sClient.Create(ctx, namespaceObj); err != nil {
-		t.Fatalf("error creating namespace: %v", err)
-	}
-
-	leaseDuration := int32(15)
-	renewTime := metav1.MicroTime{Time: time.Now()}
-	// 1. Create the MultiClusterLease with a DIFFERENT global holder
-	mcl := &mclv1alpha1.MultiClusterLease{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      leaseName,
-			Namespace: ns,
+	tests := []struct {
+		name                 string
+		globalHolderIdentity *string
+		expectedErrorSnippet string
+	}{
+		{
+			name:                 "Different global holder",
+			globalHolderIdentity: func() *string { s := "imposter-manager"; return &s }(),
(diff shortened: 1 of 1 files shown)


pkg/controller/kccmanager/kccmanager_multicluster_test.go
gemmahou, line 1: /lgtm

nit: Do we need a test case for the scenario where GlobalHolderIdentity is nil? (e.g. the MultiClusterLease object fails to reconcile, and has empty status)

acpana: done!

pkg/controller/kccmanager/kccmanager.go
```diff
@@ -139,7 +140,7 @@ func setUpMultiClusterLease(ctx context.Context, restConfig *rest.Config, scheme
 		if err := c.Get(ctx, ccName, cc); err != nil {
 			// If the ConfigConnector object is not found, proceed with default leader election.
 			klog.Infof("ConfigConnector object not found, using default in-cluster leader election")
```
maqiuyujoyce, line 142: Not a blocker: I'd log the actual error here in case it is an edge case.

pkg/controller/kccmanager/kccmanager.go
```diff
@@ -176,22 +177,51 @@ func setUpMultiClusterLease(ctx context.Context, restConfig *rest.Config, scheme
...
 
 func (m *leaderElectionManager) Start(ctx context.Context) error {
 	m.leConfig.Callbacks.OnStartedLeading = func(ctx context.Context) {
 		klog.Infof("started leading; identity: %s", m.leConfig.Lock.Identity())
+
+		// Explicitly fetch the MultiClusterLease object to verify its status.
+		lease := &mclv1alpha1.MultiClusterLease{}
+		leaseName := types.NamespacedName{Namespace: m.mclConfig.Namespace, Name: m.mclConfig.LeaseName}
+		if err := m.Manager.GetAPIReader().Get(ctx, leaseName, lease); err != nil {
+			m.fatal("error fetching MultiClusterLease %s: %v", leaseName, err)
```
maqiuyujoyce, line 212: Agent mentioned the following and I'm curious too as it looks that we'll implement onFatal() separately.
> Does `m.fatal` ensure that the process exits with a non-zero status code?
In the default implementation it calls `klog.Fatalf`, which does.
Just verifying that this aligns with the "bail" requirement.

pkg/controller/kccmanager/kccmanager.go
```diff
@@ -176,22 +177,51 @@ func setUpMultiClusterLease(ctx context.Context, restConfig *rest.Config, scheme
...
+	klog.Fatalf(format, args...)
 }
 
 func (m *leaderElectionManager) Start(ctx context.Context) error {
 	m.leConfig.Callbacks.OnStartedLeading = func(ctx context.Context) {
 		klog.Infof("started leading; identity: %s", m.leConfig.Lock.Identity())
+
+		// Explicitly fetch the MultiClusterLease object to verify its status.
+		lease := &mclv1alpha1.MultiClusterLease{}
+		leaseName := types.NamespacedName{Namespace: m.mclConfig.Namespace, Name: m.mclConfig.LeaseName}
```
maqiuyujoyce, line 210: Agent mentioned the following. I'm not super opinionated on it.

> `m.mclConfig` is assumed to be non-nil here. While `New` ensures it is populated when `leConfig` is present, adding a sanity check or comment in `Start` or `New` might be safer against future refactors.

PR #5974, opened 2026-01-07
chore: automatically build/update googleapis.pb in scripts

justinsb:
This is because otherwise the LLM has to figure out how to do. this. The trick here is to make it fast in the no-change case, so we can invoke it everywhere.


yuwenma:
/lgtm
/approve 

Make sense.

PR #5975, opened 2026-01-07
chore: fix Makefile to compile kustomize itself

justinsb:
This avoids having to preinstall kustomize binary, which may not be present
or may be the wrong version.


acpana:
nice!

/lgtm
/approve

PR #5981, opened 2026-01-08
chore: create tooling to easily compare CRDs

justinsb:
The LLMs are having trouble with bigger CRDs, so create
a simple context-efficient command that ignores descriptions.

### DIFF
--- a/dev/tasks/diff-crds
+++ b/dev/tasks/diff-crds
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# This tool shows diffs between the simple schema of all CRD files that have changed in the current branch compared to upstream/master.
+# Example usage: dev/tasks/diff-crds
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+cd ${REPO_ROOT}
+
+for CRD_FILE in $(git diff --name-only upstream/master -- config/crds); do
+  echo "************************************************************************"
+  echo "Diffs for CRD: ${CRD_FILE}"
+
+  git show upstream/master:${CRD_FILE} | go run ./dev/tools/crd-to-simple-schema --diff-crd-file="${CRD_FILE}" --flatten --ignore-integer-type-differences
+done
--- a/dev/tools/crd-to-simple-schema/main.go
+++ b/dev/tools/crd-to-simple-schema/main.go
@@ -1,0 +1,299 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"io"
+	"os"
+	"sort"
+
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/yaml"
+)
+
+func main() {
+	ctx := context.Background()
+
+	var opt ConvertOptions
+
+	flag.StringVar(&opt.Version, "version", "", "CRD version to convert")
+	flag.StringVar(&opt.CRDFile, "crd-file", "", "CRD file to convert; if empty, reads from stdin")
+	flag.StringVar(&opt.DiffCRDFile, "diff-crd-file", "", "CRD file to compare against; if empty, no comparison is done")
+	flag.BoolVar(&opt.Flatten, "flatten", false, "Flatten output to path=type lines (easier for diffing)")
+	flag.BoolVar(&opt.IgnoreIntegerTypeDifferences, "ignore-integer-type-differences", opt.IgnoreIntegerTypeDifferences, "Treat int32 and int64 as equivalent to integer when diffing.")
+
+	flag.Parse()
+
+	if err := Run(ctx, opt, os.Stdout); err != nil {
+		fmt.Fprintf(os.Stderr, "%v\n", err)
+		os.Exit(1)
+	}
+}
+
+type ConvertOptions struct {
+	// CRD file to convert; if empty, reads from stdin
+	CRDFile string
+
+	// CRD file to compare against; if empty, no comparison is done
+	DiffCRDFile string
+
+	// CRD version to convert; if empty, uses the first version found
+	Version string
+
+	// Flatten output to path=type lines (easier for diffing)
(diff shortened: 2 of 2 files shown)


dev/tasks/diff-crds
```diff
@@ -0,0 +1,31 @@
...
+set -o nounset
+set -o pipefail
+
+# This tool shows diffs between the simple schema of all CRD files that have changed in the current branch compared to upstream/master.
+# Example usage: dev/tasks/diff-crds
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+cd ${REPO_ROOT}
+
+for CRD_FILE in $(git diff --name-only upstream/master -- config/crds); do
```
cheftako, line 26: AGENT: Iterating with `$(git diff ...)` will break if any filenames contain spaces. 
Consider using `git diff ... -z` or a `while read` loop to handle filenames safely.

TAKO: Maybe we should just insist (have a policy for CRDs) files not have spaces in them?

dev/tasks/diff-crds
```diff
@@ -0,0 +1,31 @@
...
+  echo "************************************************************************"
+  echo "Diffs for CRD: ${CRD_FILE}"
+
+  git show upstream/master:${CRD_FILE} | go run ./dev/tools/crd-to-simple-schema --diff-crd-file="${CRD_FILE}" --flatten --ignore-integer-type-differences
```
cheftako, line 30: AGENT: Running `go run` inside the loop is inefficient as it recompiles the tool for every CRD file.
It would be better to build the tool once before the loop (e.g. `go build -o ...`) and then execute the binary.

TAKO: +1. Not blocking but I do think it would be good to have a check that the executable exists in the path before running and then use the built version.

cheftako, line 30: AGENT: This relies on the remote being named `upstream`, which might not be true for all users.
Also, if `${CRD_FILE}` is a newly added file, `git show upstream/master:${CRD_FILE}` will fail, causing the script to exit immediately due to `set -o errexit`.
You should check if the file exists in the base branch before attempting to show it.

TAKO: Would be good to check this assumption at the beginning. I'd suggest checking against the result of 'git remote -v' but that does not seem to work with git worktrees. Having said that git show upstream/master also does not work on git worktrees. So maybe we should have the check and add a line recommending you not do this on a worktree.

dev/tools/crd-to-simple-schema/main.go
```diff
@@ -0,0 +1,299 @@
...
+		// Funky sort that tries to match up + and - lines
+		sort.Slice(diff, func(i, j int) bool {
+			s0 := diff[i]
+			s1 := diff[j]
+			if s0[1:] < s1[1:] {
+				return true
+			}
+			if s0[1:] > s1[1:] {
+				return false
+			}
```
cheftako, line 147: AGENT: This sort logic returns `false` if `s0` starts with `-`, meaning `-` lines are considered "greater" than `+` lines (if content matches).
This results in `+` lines appearing before `-` lines. Standard diff convention is usually `-` then `+`.

TAKO: The minus sign check is not a stable less than function. Not sure if it matters for this. More to the point however I think we only want to sort by 'k' not 'v'. Should we taking the substring up to the '=' sign?

dev/tools/crd-to-simple-schema/main.go
```diff
@@ -0,0 +1,299 @@
...
+			fmt.Fprintln(out, line)
+		}
+		return nil
+	}
+
+	if opt.Flatten {
+		lines := make(map[string]string)
+		flatten("", schema, lines)
+
+		for _, line := range lines {
```
cheftako, line 164: AGENT: Iterating over a map `range lines` produces output in a random order. 
For a tool intended to generate schemas or diffs, deterministic output is crucial. 
You should collect the keys and sort them before printing.

TAKO: +1

dev/tools/crd-to-simple-schema/main.go
```diff
@@ -0,0 +1,299 @@
...
+			fmt.Fprintln(out, line)
```
cheftako, line 165: AGENT: This loop prints only the value (the type string) and ignores the key (the path).
This renders the output useless for identifying which field has which type.
It should likely be `fmt.Fprintf(out, "%s=%s\n", k, v)` (after fixing the map iteration to use keys).

dev/tools/crd-to-simple-schema/main.go
```diff
@@ -0,0 +1,299 @@
...
+	simple := walk(schema)
+
+	return simple, nil
+}
+
+func flatten(path string, schema any, out map[string]string) {
+	switch schema := schema.(type) {
+	case map[string]any:
+		for k, v := range schema {
+			flatten(path+"."+k, v, out)
```
cheftako, line 217: AGENT: Constructing the path as `path+"."+k` results in a leading dot for top-level fields (e.g., `.spec.foo`).
While not technically wrong, it's cleaner to handle the empty `path` case to avoid the leading dot (e.g. `if path == "" { return k }`).

TAKO: That is fine.

dev/tools/crd-to-simple-schema/main.go
```diff
@@ -0,0 +1,299 @@
...
+		}
+
+	case []any:
+		for _, v := range schema {
+			flatten(fmt.Sprintf("%s[]", path), v, out)
+		}
+	case string:
+		out[path] = schema
+	default:
+		klog.Fatalf("unhandled type %T", schema)
```
cheftako, line 227: AGENT: Using `klog.Fatalf` here will terminate the entire program immediately.
It would be better to return an error from `flatten` (and `walk`) so that `main` can handle it gracefully or print a proper error message.

TAKO: Tool rather than prod system. I think its fine. I would say we may want to consider investing in errors that can help an LLM diagnose and fix the issue.

dev/tools/crd-to-simple-schema/main.go
```diff
@@ -0,0 +1,299 @@
...
+	if _, err := out.Write(res); err != nil {
+		return fmt.Errorf("writing output: %w", err)
+	}
+	return nil
+}
+
+func buildSimpleSchema(data []byte, opt ConvertOptions) (any, error) {
+	var crd apiextensionsv1.CustomResourceDefinition
+	if err := yaml.Unmarshal(data, &crd); err != nil {
+		return nil, fmt.Errorf("unmarshaling CRD: %w", err)
```
cheftako, line 183: AGENT: `yaml.Unmarshal` here will only decode the first document if the input contains multiple YAML documents (or a stream).
If CRD files might contain multiple documents, you should use a decoder loop.

TAKO: Might be good to document that but I think its ok.

cheftako (commented):
AGENT: Overall, the idea is solid and helps solving the problem of reviewing large CRD changes.
The implementation needs some fixes to be robust and correct. 
Specifically, the non-diff output is currently incorrect, and the script needs better error handling.

TAKO: Can we get a sample output to help understand if it does what we want?

dev/tasks/diff-crds
justinsb: They won't have spaces in them, but I might just rewrite in python!

dev/tasks/diff-crds
justinsb: `go run` is fast (enough) now.  It didn't used to be the case, but it is good enough now IMO.

justinsb: Sounds like python time!

dev/tools/crd-to-simple-schema/main.go
justinsb: Yeah, we've used the `.spec.foo` format elsewhere

dev/tools/crd-to-simple-schema/main.go
justinsb: Yes, this one indicates that we have some code to write, so klog.Fatalf is appropriate.  Agree that we can improve the message

dev/tools/crd-to-simple-schema/main.go
justinsb: Will fix

dev/tools/crd-to-simple-schema/main.go
justinsb: Will fix

dev/tasks/diff-crds
justinsb: The default is upstream/master, but it's handy to be able to pass in a base (including a tag!), so I added that

dev/tools/crd-to-simple-schema/main.go
justinsb: Cleaned up in #6013

PR #5983, opened 2026-01-08
feat: statusz endpoint

acpana:
Show:

```
$ curl http://localhost:8082/statusz
KCC Manager Status: RUNNING
---------------------------
Version:       1.141.0
Billing Project:
User Project Override: false
MultiCluster Lease:    false
Mode:                  Cluster
Actuation Mode:
Time:                  2026-01-08T20:43:23Z
```

### BRIEF Change description

First pass to add a `/statusz` endpoint.

Enable human operators to query manager pods internal status

```release-note
Add a `/statusz` endpoint for manager pods.
```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -61,6 +61,7 @@ func main() {
 		rateLimitQps             float32
 		rateLimitBurst           int
 		leaderElectionMode       string
+		statuszPort              int
 	)
 	flag.StringVar(&prometheusScrapeEndpoint, "prometheus-scrape-endpoint", ":8888", "configure the Prometheus scrape endpoint; :8888 as default")
 	flag.BoolVar(&controllermetrics.ResourceNameLabel, "resource-name-label", false, "option to enable the resource name label on some Prometheus metrics; false by default")
@@ -69,6 +70,7 @@ func main() {
 	flag.StringVar(&scopedNamespace, "scoped-namespace", "", "scope controllers to only watch resources in the specified namespace; if unspecified, controllers will run in cluster scope")
 	flag.BoolVar(&enablePprof, "enable-pprof", false, "Enable the pprof server.")
 	flag.IntVar(&pprofPort, "pprof-port", 6060, "The port that the pprof server binds to if enabled.")
+	flag.IntVar(&statuszPort, "statusz-port", 8082, "The port for the statusz endpoint.")
 	flag.Float32Var(&rateLimitQps, "qps", 20.0, "The client-side token bucket rate limit qps.")
 	flag.IntVar(&rateLimitBurst, "burst", 30, "The client-side token bucket rate limit burst.")
 	flag.StringVar(&leaderElectionMode, "leader-election-type", "disabled", "Leader election mode. One of: default, multicluster.")
@@ -116,7 +118,7 @@ func main() {
 	// Set client site rate limiter to optimize the configconnector re-reconciliation performance.
 	ratelimiter.SetMasterRateLimiter(restCfg, rateLimitQps, rateLimitBurst)
 	logger.Info("Creating the manager")
-	mgr, err := newManager(ctx, restCfg, scopedNamespace, userProjectOverride, billingProject, multiClusterElection)
+	mgr, err := newManager(ctx, restCfg, scopedNamespace, userProjectOverride, billingProject, multiClusterElection, statuszPort)
 	if err != nil {
 		logging.Fatal(err, "error creating the manager")
 	}
@@ -172,7 +174,7 @@ func main() {
 	logging.ExitInfo("main.go finished execution; exiting ...")
 }
 
-func newManager(ctx context.Context, restCfg *rest.Config, scopedNamespace string, userProjectOverride bool, billingProject string, multiclusterlease bool) (manager.Manager, error) {
+func newManager(ctx context.Context, restCfg *rest.Config, scopedNamespace string, userProjectOverride bool, billingProject string, multiclusterlease bool, statuszPort int) (manager.Manager, error) {
 	krmtotf.SetUserAgentForTerraformProvider()
 	controllersCfg := kccmanager.Config{
 		ManagerOptions: manager.Options{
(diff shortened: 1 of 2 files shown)


pkg/controller/kccmanager/kccmanager.go
```diff
@@ -302,6 +310,78 @@ func New(ctx context.Context, restConfig *rest.Config, cfg Config) (manager.Mana
 			return nil, err
 		}
 
+		// Register /statusz Endpoint
+		if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
```
cheftako, line 314: AGENT: The inline anonymous function for the statusz server is getting quite large.

Consider extracting this logic into a separate function (e.g., `startStatuszServer`) or a dedicated struct. This would improve the readability of the `New` function and make it easier to unit test the statusz logic in isolation.

TAKO: +1. No good reason to have an anonymous method.

pkg/controller/kccmanager/kccmanager.go
```diff
@@ -302,6 +310,78 @@ func New(ctx context.Context, restConfig *rest.Config, cfg Config) (manager.Mana
...
+				fmt.Fprintf(w, "Mode:                  %s\n", mode)
+
+				// Fetch Actuation Status
+				client := mgr.GetClient()
+				var actuationStatus string = "Unknown"
+
+				if cfg.ScopedNamespace == "" {
+					// Cluster Mode: Check ConfigConnector
+					cc := &operatorv1beta1.ConfigConnector{}
+					if err := client.Get(ctx, types.NamespacedName{Name: operatorv1beta1.ConfigConnectorAllowedName}, cc); err != nil {
```
cheftako, line 340: AGENT: It is recommended to use the request context `r.Context()` instead of the manager's context `ctx` for operations performed within the HTTP handler.

Using `r.Context()` ensures that if the incoming HTTP request is cancelled or times out, the downstream Kubernetes client operation is also cancelled, preventing unnecessary work.

TAKO: +1

pkg/controller/kccmanager/kccmanager.go
```diff
@@ -302,6 +310,78 @@ func New(ctx context.Context, restConfig *rest.Config, cfg Config) (manager.Mana
...
+						actuationStatus = fmt.Sprintf("Error fetching CC: %v", err)
+					} else {
+						actuationStatus = string(cc.Spec.Actuation)
+					}
+				} else {
+					// Namespaced Mode: Check ConfigConnectorContext
+					ccc := &operatorv1beta1.ConfigConnectorContext{}
+					if err := client.Get(ctx, types.NamespacedName{Name: operatorv1beta1.ConfigConnectorContextAllowedName, Namespace: cfg.ScopedNamespace}, ccc); err != nil {
```
cheftako, line 348: AGENT: Same as above: please use `r.Context()` here for the client call.

TAKO: +1

pkg/controller/kccmanager/kccmanager.go
```diff
@@ -302,6 +310,78 @@ func New(ctx context.Context, restConfig *rest.Config, cfg Config) (manager.Mana
...
+						actuationStatus = fmt.Sprintf("Error fetching CCC: %v", err)
+					} else {
+						actuationStatus = string(ccc.Spec.Actuation)
+					}
+				}
+				fmt.Fprintf(w, "Actuation Mode:        %s\n", actuationStatus)
+				fmt.Fprintf(w, "Time:                  %s\n", time.Now().Format(time.RFC3339))
+			})
+
+			port := cfg.StatuszPort
```
cheftako, line 358: AGENT: This logic overrides the zero value (0) to 8082. While this ensures a default, it prevents the caller from explicitly specifying port 0 (which usually implies a dynamically assigned port) or potentially disabling the port if that was intended.

Since `main.go` already sets the default flag value to 8082, `cfg.StatuszPort` will likely already be 8082 unless explicitly set to 0. Consider if 0 should be a valid configuration.

cmd/manager/main.go
```diff
@@ -172,7 +174,7 @@ func main() {
 	logging.ExitInfo("main.go finished execution; exiting ...")
 }
 
-func newManager(ctx context.Context, restCfg *rest.Config, scopedNamespace string, userProjectOverride bool, billingProject string, multiclusterlease bool) (manager.Manager, error) {
+func newManager(ctx context.Context, restCfg *rest.Config, scopedNamespace string, userProjectOverride bool, billingProject string, multiclusterlease bool, statuszPort int) (manager.Manager, error) {
```
cheftako, line 177: AGENT: The `newManager` function signature is growing quite large (7 arguments).

Consider refactoring this to accept a configuration struct or using the functional options pattern. This makes the function easier to extend in the future without breaking signatures.

TAKO: +1. Not blocking. Builder pattern can be another approach.

pkg/controller/kccmanager/kccmanager.go
cheftako, line 314: AGENT: This PR adds significant logic in the handler (fetching K8s resources, formatting) but appears to lack unit tests.

Please add a test case (likely in `kccmanager_test.go` or a new test file) that exercises the `/statusz` handler, mocks the client, and asserts on the returned status format and HTTP codes.

TAKO: See below. For /status endpoint less concerned. For an actual /status*z* I think this is critical.

pkg/controller/kccmanager/kccmanager.go
```diff
@@ -302,6 +310,78 @@ func New(ctx context.Context, restConfig *rest.Config, cfg Config) (manager.Mana
 			return nil, err
 		}
 
+		// Register /statusz Endpoint
+		if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
+			mux := http.NewServeMux()
+			mux.HandleFunc("/statusz", func(w http.ResponseWriter, r *http.Request) {
+				w.Header().Set("Content-Type", "text/plain")
+				w.WriteHeader(http.StatusOK)
+				fmt.Fprintf(w, "KCC Manager Status: RUNNING\n")
```
cheftako, line 319: Please see https://kubernetes.io/docs/reference/instrumentation/zpages/. Z pages are meant to be machine readable, structured output. It should have a schema and be in json. If you want a more human readable and less rigid status endpoint, that should just be /status.

cheftako: Also we have a pause feature. So I don't thing "RUNNING" is a given?

PR #5985, opened 2026-01-08
tests: fix fuzz path for dashboard xy_chart

justinsb:


pkg/controller/direct/monitoring/dashboard_fuzzer.go
```diff
@@ -85,10 +90,10 @@ func dashboardFuzzer() fuzztesting.KRMFuzzer {
 		fuzzer.UnimplementedFields.Insert(widgetPath + ".time_series_table.data_sets[].time_series_query.time_series_filter_ratio.statistical_time_series_filter")
 		fuzzer.UnimplementedFields.Insert(widgetPath + ".time_series_table.data_sets[].time_series_query.time_series_filter_ratio.pick_time_series_filter.interval")
 
-		fuzzer.UnimplementedFields.Insert(widgetPath + ".xy_chart.time_series_query.time_series_filter.statistical_time_series_filter")
-		fuzzer.UnimplementedFields.Insert(widgetPath + ".xy_chart.time_series_query.time_series_filter.pick_time_series_filter.interval")
-		fuzzer.UnimplementedFields.Insert(widgetPath + ".xy_chart.time_series_query.time_series_filter_ratio.statistical_time_series_filter")
-		fuzzer.UnimplementedFields.Insert(widgetPath + ".xy_chart.time_series_query.time_series_filter_ratio.pick_time_series_filter.interval")
+		fuzzer.UnimplementedFields.Insert(widgetPath + ".xy_chart.data_sets[].time_series_query.time_series_filter.statistical_time_series_filter")
```
cheftako, line 93: AGENT: Good catch. The insertion of `.data_sets[]` correctly matches the `XyChart` proto structure where `time_series_query` is nested inside `dataSets`.

TAKO: Is there a way we could have caught the path error earlier?

pkg/controller/direct/networkmanagement/connectivitytest_fuzzer.go
```diff
@@ -13,8 +13,9 @@
 // limitations under the License.
 
 // +tool:fuzz-gen
-// proto.message: google.cloud.kms.v1.ImportJob
-// api.group: kms.cnrm.cloud.google.com
+// proto.message: google.cloud.networkmanagement.v1.ConnectivityTest
```
cheftako, line 16: AGENT: Thanks for fixing the copy-paste error here. It definitely shouldn't be pointing to KMS ImportJob.

TAKO: Same question. Is there a way we could have caught this earlier? Linter?

pkg/controller/direct/monitoring/dashboard_fuzzer.go
justinsb: Good call... we could probably check that quite easily actually, just by walking the proto MessageDescriptor

pkg/controller/direct/networkmanagement/connectivitytest_fuzzer.go
justinsb: I think we could ask gemini to make sure that all fuzzers have these annotations, fixing them if needed

PR #5989, opened 2026-01-08
feat: no manual editing to  Status_FromProto(ToProto)

yuwenma:
### BRIEF Change description
Without this change, the mapper generator cannot create the `Status_FromProto` and `Status_ToProto` functions. It ends up as we have to manually do the work whenever the GCP resource uses `google.rpc.Status`.

This PR patches both type-generator and mapper-generator to generated mappers for `google.rpc.Status` (and use a shared `Status` struct)   

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


pkg/controller/direct/maputils.go
```diff
@@ -579,3 +583,56 @@ func Struct_ToProto(mapCtx *MapContext, in *apiextensionsv1.JSON) *structpb.Stru
...
+	if len(in.Details) == 0 {
+		return out
+	}
+	detailsOut := make([]common.Any, 0)
+	for _, d := range in.Details {
+		if d == nil {
+			continue
+		}
+		dOut := common.Any{
+			TypeURL: LazyPtr(d.TypeUrl),
```
cheftako, line 604: AGENT: For `networkmanagement`, the `Any.Value` field was previously mapped to `*string` (decoding bytes to string).
Now it is mapped to `[]byte`, which will result in a base64-encoded string in the JSON output.
This is a breaking change for clients reading the `details` field. Please confirm this is intentional.

apis/networkmanagement/v1alpha1/connectivitytest_types.go
```diff
@@ -336,7 +337,7 @@ type ProbingDetailsObservedState struct {
 
 	// Details about an internal failure or the cancellation of active probing.
 	// +kcc:proto:field=google.cloud.networkmanagement.v1.ProbingDetails.error
-	Error *StatusObservedState `json:"error,omitempty"`
+	Error *common.Status `json:"error,omitempty"`
```
cheftako: AGENT: Using `common.Status` here brings in the `common.Any` type for `Details`.
As noted, `common.Any` uses `[]byte` for `Value`, whereas the previous local `Any` struct used `*string`.
This confirms the schema change visible in the CRD diff.

dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
```diff
@@ -133,6 +133,7 @@ func (g *TypeGenerator) WriteVisitedMessages() error {
 			FileName:  "types.generated.go",
 		}
 		out := g.getOutputFile(k)
+		out.addImport("common", "github.com/GoogleCloudPlatform/k8s-config-connector/apis/common")
```
cheftako: This unconditionally adds the `common` import to all visited messages.
Ensure that `goimports` or similar tooling is run to remove this import if the generated file doesn't actually use `common.Status` or `common.Any` (to avoid "unused import" errors).

apis/common/googleapi.go
```diff
@@ -0,0 +1,72 @@
...
+	//    URL, or have them precompiled into a binary to avoid any
+	//    lookup. Therefore, binary compatibility needs to be preserved
+	//    on changes to types. (Use versioned type names to manage
+	//    breaking changes.)
+	//
+	//  Note: this functionality is not currently available in the official
+	//  protobuf release, and it is not used for type URLs beginning with
+	//  type.googleapis.com.
+	//
+	//  Schemes other than `http`, `https` (or the empty scheme) might be
```
cheftako: Changing `Value` to `[]byte` means it will be serialized as a Base64 string in JSON.
If the previous generated code used `*string` (and not `[]byte`), verify if the previous serialization was also Base64 or a raw string.
If it was a raw string, this is a breaking change for clients reading the `details` field.

dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
cheftako: AGENT: This unconditionally adds the `apis/common` import.
If a generated file does not use `common.Status` or `common.Any` (e.g. it only has other types), this will result in an "unused import" compilation error unless `goimports` is run afterwards.
Consider checking if the message actually uses `Status` or `Any` before adding the import.

TAKO: +1

apis/common/googleapi.go
```diff
@@ -0,0 +1,72 @@
+// Copyright 2024 Google LLC
```
cheftako: AGENT: Please update the copyright year to 2026.

TAKO: +1

apis/common/zz_generated.deepcopy.go
```diff
@@ -0,0 +1,76 @@
+//go:build !ignore_autogenerated
```
cheftako, line 1: AGENT: Please update the copyright year to 2026 (currently 2020).

TAKO: At a minimum we should fix the generation for new files.

apis/networkmanagement/v1alpha1/connectivitytest_types.go
cheftako: AGENT: **Breaking Change**: `Error` now uses `common.Status`, which uses `common.Any`. `common.Any.Value` is `[]byte` (Base64 in JSON).
Previously, this was `*string` (String in JSON). This breaks clients expecting a plain string in `details[].value`.

pkg/controller/direct/maputils.go
```diff
@@ -579,3 +583,56 @@ func Struct_ToProto(mapCtx *MapContext, in *apiextensionsv1.JSON) *structpb.Stru
...
+		return nil
+	}
+	out := &statuspb.Status{}
+	out.Code = ValueOf(in.Code)
+	out.Message = ValueOf(in.Message)
+
+	if len(in.Details) == 0 {
+		return out
+	}
+	detailsOut := make([]*anypb.Any, 0)
```
cheftako: AGENT: Performance: Preallocate the slice.
`detailsOut := make([]*anypb.Any, 0, len(in.Details))`

TAKO: At best a nice to have.

apis/common/googleapi.go
```diff
@@ -0,0 +1,72 @@
...
+	//  type.googleapis.com.
+	//
+	//  Schemes other than `http`, `https` (or the empty scheme) might be
+	//  used with implementation specific semantics.
+	// +kcc:proto:field=google.protobuf.Any.type_url
+	TypeURL *string `json:"typeURL,omitempty"`
+
+	// Must be a valid serialized protocol buffer of the above specified type.
+	// +kcc:proto:field=google.protobuf.Any.value
+	Value []byte `json:"value,omitempty"`
```
cheftako: AGENT: Changing Value to []byte means it will be serialized as a Base64 string in JSON.
If the previous generated code used *string (and not []byte), verify if the previous serialization was also Base64 or a raw string.
If it was a raw string, this is a breaking change for clients reading the details field.

Defining `Value` as `[]byte` is standard for `google.protobuf.Any`, but note that this enforces Base64 encoding in JSON, differing from some previous custom mappings.

TAKO: Again good thing to confirm.

apis/common/googleapi.go
```diff
@@ -0,0 +1,72 @@
...
+	//    breaking changes.)
+	//
+	//  Note: this functionality is not currently available in the official
+	//  protobuf release, and it is not used for type URLs beginning with
+	//  type.googleapis.com.
+	//
+	//  Schemes other than `http`, `https` (or the empty scheme) might be
+	//  used with implementation specific semantics.
+	// +kcc:proto:field=google.protobuf.Any.type_url
+	TypeURL *string `json:"typeURL,omitempty"`
```
cheftako: AGENT: The JSON tag typeURL is likely a change from previous generated code which typically used typeUrl (camelCase) to match the proto field name type_url.
Please confirm if existing resources (like NetworkManagement or BigQueryReservation) used typeUrl. If so, this is a breaking change for clients.

TAKO: Seems a good thing to confirm.

cheftako (commented):
AGENT: The centralization of `google.rpc.Status` and `google.protobuf.Any` types is a positive step for maintainability and structure.

However, I have identified breaking changes that need addressing:
*   **`NetworkManagement` ConnectivityTest:** The `Any` value exposed at `status.error.details[].value` has changed from a string to bytes (base64 encoded in JSON). Please confirm if this breaking change is intended, even for a `v1alpha1` resource.
*   **`Any` Schema:** There appear to be changes affecting the `Any.TypeURL` JSON tag as well.

Additionally, I noted a few minor issues regarding copyright headers, potential unused imports, and slice allocation optimizations. Aside from these points, the code generation updates and common type definitions look correct.

apis/common/googleapi.go
yuwenma: Confirmed this is not a breaking change because the CRD is not changed.

dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
yuwenma: Sure. I've updated the change to make the code cleaner. But this won't cause problems because we indeed call goimports afterwards :)

pkg/controller/direct/maputils.go
yuwenma: Good call! reverted the change in NetworkManagementt.

apis/common/googleapi.go
```diff
@@ -0,0 +1,72 @@
...
+	//  user-facing error message should be localized and sent in the
+	//  [google.rpc.Status.details][google.rpc.Status.details] field, or localized
+	//  by the client.
+	// +kcc:proto:field=google.rpc.Status.message
+	Message *string `json:"message,omitempty"`
+
+	// A list of messages that carry the error details.  There is a common set of
+	//  message types for APIs to use.
+	// +kcc:proto:field=google.rpc.Status.details
+	Details []Any `json:"details,omitempty"`
```
justinsb, line 34: So I don't think we want Details (at least not by default), because it's not going to be human-readable.  We can always add it later if needed.

apis/common/googleapi.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// +kcc:proto=google.protobuf.Any
+type Any struct {
```
justinsb, line 38: Let's drop Any, for now

justinsb:
This looks useful, but I don't think we want to map Details because the Any will not be human-readable.  @codebot-robot do you want to take over?  We'll need a rebase it looks like

PR #5998, opened 2026-01-12
Added initial IAM ServiceAccountKey types.

cheftako:
### BRIEF Change description

Added identity and reference objects.
Updated mappers to get compiles working.
A few tweaks to get generation working.

Fixes #

```release-note
NONE
```

```docs
NONE
```

#### Intended Milestone

### Tests you have done


apis/iam/v1beta1/serviceaccountkey_types.go
```diff
@@ -0,0 +1,115 @@
...
+	// +optional
+	ObservedGeneration *int64 `json:"observedGeneration,omitempty"`
+
+	// Immutable. The name used for this key pair.
+	// +optional
+	Name *string `json:"name,omitempty"`
+
+	// The private key in JSON format, base64 encoded. This is what you normally get as a file when creating service account keys through the CLI or web console. This is only populated when creating a new key.
+	// +optional
+	PrivateKey *string `json:"privateKey,omitempty"`
```
acpana, line 64: agent + acpana (AA): A V2 of this resource here would probably better hide this right?

apis/iam/v1beta1/serviceaccountkey_types.go
```diff
@@ -0,0 +1,115 @@
...
+	// +optional
+	KeyAlgorithm *string `json:"keyAlgorithm,omitempty"`
+
+	// Immutable. The output format of the private key. TYPE_GOOGLE_CREDENTIALS_FILE is the default output format.
+	// +optional
+	PrivateKeyType *string `json:"privateKeyType,omitempty"`
+
+	// Immutable. A field that allows clients to upload their own public key. If set, use this public key data to create a service account key for given service account. Please note, the expected format for this field is a base64 encoded X509_PEM.
+	// +optional
+	PublicKeyData *string `json:"publicKeyData,omitempty"`
```
acpana, line 42: AA big nit: validation for the format of the field.

apis/iam/v1beta1/serviceaccountkey_types.go
```diff
@@ -0,0 +1,115 @@
...
+// IAMServiceAccountKeySpec defines the desired state of IAMServiceAccountKey
+// +kcc:spec:proto=google.iam.admin.v1.ServiceAccountKey
+type IAMServiceAccountKeySpec struct {
+	// Immutable. The Service Account to create a key for.
+	// +required
+	ServiceAccountRef *refs.IAMServiceAccountRef `json:"serviceAccountRef"`
+
+	// Immutable. The algorithm used to generate the key, used only on create. KEY_ALG_RSA_2048 is the default algorithm. Valid values are: "KEY_ALG_RSA_1024", "KEY_ALG_RSA_2048".
+	// +optional
+	KeyAlgorithm *string `json:"keyAlgorithm,omitempty"`
```
acpana, line 34: AA nit: enum validation for the format of the field.

And for the PrivateKeyType, PublicKeyType etc.

apis/iam/v1beta1/serviceaccountkey_types.go
```diff
@@ -0,0 +1,115 @@
...
+	// +optional
+	PrivateKey *string `json:"privateKey,omitempty"`
+
+	// Immutable. The public key, base64 encoded.
+	// +optional
+	PublicKey *string `json:"publicKey,omitempty"`
+
+	// The key can be used after this timestamp. A timestamp in RFC3339 UTC "Zulu" format, accurate to nanoseconds. Example: "2014-10-02T15:01:23.045123456Z".
+	// +optional
+	ValidAfter *string `json:"validAfter,omitempty"`
```
acpana, line 72: AGENT: Please confirm that the time format returned by the GCP API strictly matches the `date-time` format expected by Kubernetes serialization if you were to use `metav1.Time`. Since you are using `string`, it's fine, but consistent formatting is important for clients.

apis/iam/v1beta1/serviceaccountkey_identity.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	// Resolve the service account reference
+	serviceAccountRef := *obj.Spec.ServiceAccountRef
+	if err := serviceAccountRef.Resolve(ctx, reader, obj); err != nil {
+		return nil, fmt.Errorf("cannot resolve spec.serviceAccountRef: %w", err)
+	}
+
+	// The resolved External should be an email address
+	serviceAccountID := serviceAccountRef.External
```
acpana: AGENT: Ensure that `serviceAccountRef.External` is strictly the email address (the Service Account ID) and not the full resource name (which includes `projects/...`). If `External` contains the full name, the `String()` method on `ServiceAccountKeyParent` (line 50) will construct an invalid parent path (nested `projects/`).

apis/iam/v1beta1/serviceaccountkey_reference.go
```diff
@@ -0,0 +1,83 @@
...
+	// From the Config Connector object
+	if r.Namespace == "" {
+		r.Namespace = otherNamespace
+	}
+	key := types.NamespacedName{Name: r.Name, Namespace: r.Namespace}
+	u := &unstructured.Unstructured{}
+	u.SetGroupVersionKind(IAMServiceAccountKeyGVK)
+	if err := reader.Get(ctx, key, u); err != nil {
+		if apierrors.IsNotFound(err) {
+			return "", k8s.NewReferenceNotFoundError(u.GroupVersionKind(), key)
```
acpana: AA: If the object exists but `status.name` is empty (e.g. not yet reconciled), this returns a `ReferenceNotFoundError`. `NewReferenceNotReadyError` (used below) seems more appropriate if the object exists but isn't ready.

dev/tools/controllerbuilder/pkg/gocode/ast.go
```diff
@@ -239,12 +239,20 @@ func (p *Package) addStruct(name *ast.Ident, def *ast.StructType, comments []ast
 		tokens := strings.Split(s, ".")
 		if len(tokens) > 1 {
 			packageName := tokens[0]
-			for _, imp := range p.Imports {
-				if imp.Alias == packageName {
-					structField.GoPackage = imp.GoPackage
-					break
+			// if packageName == "apiextensionsv1" {
```
acpana, line 242: AGENT: Please remove this commented-out code block before merging.

acpana: it's fine to leave it in as far as this is tooling code.

pkg/controller/direct/iam/mappers.go
```diff
@@ -42,3 +46,170 @@ func Expr_ToProto(mapCtx *direct.MapContext, in *krmv1alpha1.Expr) *exprpb.Expr
 	out.Location = direct.ValueOf(in.Location)
 	return out
 }
+
+func IAMServiceAccountKeySpec_FromProto(mapCtx *direct.MapContext, in *adminpb.ServiceAccountKey) *krm.IAMServiceAccountKeySpec {
+	if in == nil {
+		return nil
+	}
+	out := &krm.IAMServiceAccountKeySpec{}
+	out.KeyAlgorithm = direct.Enum_FromProto(mapCtx, in.GetKeyAlgorithm())
```
acpana, line 55: AGENT: For `PublicKeyData`, you cast `in.GetPublicKeyData()` (which returns `[]byte` in generated protos usually) to `string`. Ensure that the proto field is indeed `bytes` and that converting it to string is the desired behavior for the `Spec`. 

The Spec defines it as: `PublicKeyData *string`. 
GCP `ServiceAccountKey` proto defines `public_key_data` as `bytes`.

If it's arbitrary binary data (DER/PEM), treating it as a Go string is mostly fine but `[]byte` in the Spec (which marshals to base64 string in JSON) might be more idiomatic for K8s if it's binary. However, if it's PEM (text), string is correct. The comment in Spec says "base64 encoded X509_PEM".

dev/tools/controllerbuilder/pkg/protoapi/loader.go
```diff
@@ -79,3 +79,14 @@ func (p *Proto) GetFileDescriptorByPackage(protoPackage string) ([]protoreflect.
 	}
 	return files, nil
 }
+
+func (p *Proto) GetAllFileDescriptors() ([]protoreflect.FileDescriptor, error) {
```
acpana, line 83: AGENT: This new method `GetAllFileDescriptors` is added but I don't see it used in the diff. Is it used in the modified `controllerbuilder` code?

apis/iam/v1beta1/serviceaccountkey_types.go
cheftako: Not in a useful way. Hiding the types seems antithetical to the goal of generating a types file which is compatible with the older TF controller. A v2 would still have to be compatible but it would be harder to compare.

apis/iam/v1beta1/serviceaccountkey_types.go
cheftako: What validation? This should probably be discussed in the CRD. However there is no format validation defined there. Adding validation is generally not backward compatible.

apis/iam/v1beta1/serviceaccountkey_types.go
cheftako: What validation? This should probably be discussed in the CRD. However there is no format validation defined there. Adding validation is generally not backward compatible.

apis/iam/v1beta1/serviceaccountkey_types.go
cheftako: What validation? This should probably be discussed in the CRD. However there is no format validation defined there. Adding validation is generally not backward compatible.

apis/iam/v1beta1/serviceaccountkey_identity.go
cheftako: What validation? This should probably be discussed in the CRD. However there is no format validation defined there. Adding validation is generally not backward compatible.

apis/iam/v1beta1/serviceaccountkey_reference.go
cheftako: Should switch this to use gcpurls.

dev/tools/controllerbuilder/pkg/protoapi/loader.go
cheftako: Its a debug utility. I was calling it to determine the file list being used. Should not be called except in a debug situation.

apis/iam/v1beta1/serviceaccountkey_identity.go
```diff
@@ -0,0 +1,196 @@
...
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/apis/common/identity"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/gcpurls"
+)
+
+var (
+	_ identity.IdentityV2 = &ServiceAccountKeyIdentity{}
+	_ identity.IdentityV2 = &ServiceAccountKeyParent{}
+
+	serviceAccountKeyFormat = gcpurls.Template[ServiceAccountKeyIdentity](
```
justinsb: So I don't think we need ServiceAccountKeyParent at all, we just have

```
type ServiceAccountIdentity struct {
	Project string
	ServiceAccount string
}
```

and

```
type ServiceAccountKeyIdentity struct {
	Project string
	ServiceAccount string
    Key string
}
```

apis/iam/v1beta1/serviceaccountkey_reference.go
```diff
@@ -0,0 +1,144 @@
...
+func (r *ServiceAccountKeyRef) Normalize(ctx context.Context, reader client.Reader, defaultNamespace string) error {
+	fallback := func(u *unstructured.Unstructured) string {
+		resourceID, err := refsv1beta1.GetResourceID(u)
+		if err != nil || resourceID == "" {
+			// ServiceAccountKye is currently non-adoptable.
+			// As such we expect the key ref to not work, for now.
+			return ""
+		}
+
+		externelRef, present, err := unstructured.NestedString(u.Object, "spec", "serviceAccountRef", "external")
```
justinsb, line 89: I would take a look at getIdentityFromBigQueryAnalyticsHubListingSpec for what I think is a more maintainable pattern for resolving parent refs (it's not perfect, it feels a little verbose, but it reuses the existing code)

apis/iam/v1beta1/serviceaccountkey_types.go
```diff
@@ -0,0 +1,115 @@
...
+	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/apis/k8s/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var IAMServiceAccountKeyGVK = GroupVersion.WithKind("IAMServiceAccountKey")
+
+// IAMServiceAccountKeySpec defines the desired state of IAMServiceAccountKey
+// +kcc:spec:proto=google.iam.admin.v1.ServiceAccountKey
+type IAMServiceAccountKeySpec struct {
```
justinsb, line 27: For anything with an existing terraform type, I believe the "merge criteria" is that the CRD doesn't change (much), and now we have `dev/tasks/diff-crds` to make that easy to read.  So if you wanted to split this one out into its own PR it might be simpler all round. (Similarly you could use the "canned issue" to do identity for ServiceAccountKey, if you wanted)

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_iamserviceaccountkeys.iam.cnrm.cloud.google.com.yaml
```diff
@@ -170,9 +181,3 @@ spec:
     storage: true
     subresources:
```
justinsb, line 182: Diff looks good (and we should probably figure out how to run dev/tasks/diff-crds on every PR!)

pkg/gcpurls/template.go
```diff
@@ -53,7 +53,7 @@ type URLTemplate[T any] struct {
 	host                string
 	template            string
 	segments            []string
-	segmentToFieldIndex []int
+	segmentToFieldIndex [][]int
```
justinsb: If this is to support {a.b} in the template, I'm not sure it's worth the complexity either here or in the templates.

apis/iam/v1beta1/serviceaccountkey_identity.go
cheftako: I guess a matter of taste. I prefer
```
type ServiceAccountIdentity struct {
	Project string
	ServiceAccount string
}
```
and
```
type ServiceAccountKeyIdentity struct {
    ServiceAccountIdentity
    Key string
}
```

apis/iam/v1beta1/serviceaccountkey_reference.go
cheftako: Dislike all of these. Somewhat irrelevant for now an this is a *currently* non adoptable (https://docs.cloud.google.com/config-connector/docs/how-to/managing-deleting-resources#resources_that_cannot_be_acquired) Would not block on this for now. We should come up with a patter we like.

apis/iam/v1beta1/serviceaccountkey_types.go
cheftako: +1. For the comparison I have been looking at the diff on the actual CRD below.

pkg/gcpurls/template.go
cheftako: It is. Lets chat. I agree here things are more complicated. It is localized and I think it gets us more consistency and makes the appropriate pieces as building blocks.

pkg/gcpurls/template.go
```diff
@@ -48,68 +48,52 @@ func register(t RegisteredTemplate) {
...
 	var zero T
 	typ := reflect.TypeOf(zero)
 	if typ.Kind() != reflect.Struct {
 		panic(fmt.Sprintf("type %T must be a struct", zero))
 	}
-
-	t.segmentToFieldIndex = make([]int, len(t.segments))
-	for i, segment := range t.segments {
+	// Generate a map of field name to index/path
+	nameMap := make(map[string][]int)
```
justinsb, line 74: I feel like we can probably have the LLM refactor this, as it's quite hard to read, but I think it's right!

PR #6008, opened 2026-01-13
fix: sanitize managed fields metadata to remove leaked status subresource

maqiuyujoyce:
### BRIEF Change description

This change:
1. Adds SanitizeSpecManagedFields to explicitly clear subresource from entries that manage `f:spec`.
2. Applies this sanitization in `LifecycleHandler.updateStatus` and `LifecycleHandler.updateAPIServer` immediately after unmarshalling the response via `util.Marshal()`.
3. Adds unit tests covering this corruption scenario.

Fixes: b/465380187

`json.Unmarshal` merges the input JSON into the existing struct. When reusing a `k8s.Resource` object across reconciliations, if a previous operation (like a status update) added a managedFields entry with `subresource: status`, and the subsequent `json.Unmarshal` (for a spec update) touches the same fields, the `subresource: status` field persists in the entry even if the new input didn't have it.

This causes incorrect managedFields metadata where spec updates are attributed to the status subresource, leading to 'Location must be set' errors when reconciling a CloudSchedulerJob because the controller ignores spec fields managed by status.

```release-note
Fixes a bug where managedFields metadata corruption could prevent resources from reconciling correctly
```

```docs

```

#### Intended Milestone

### Tests you have done

Created local build and ran E2E test to verify the managed fields stay correct all the time.

### DIFF
--- a/pkg/k8s/managedfields.go
+++ b/pkg/k8s/managedfields.go
@@ -167,14 +167,29 @@ func GetK8sManagedFields(u *unstructured.Unstructured) (*fieldpath.Set, error) {
 
 // SanitizeSpecManagedFields ensures that managed fields entries for spec fields do not have
 // Subresource field configured.
+// This function modifies the resource in-place.
+// This is needed to clean up the leaked `subresource: status` when unmarshalling the correct resource into existing
+// structs (b/465380187).
 func SanitizeSpecManagedFields(r *Resource) {
+	if r == nil || r.ObjectMeta.ManagedFields == nil {
+		return
+	}
 	for i := range r.ObjectMeta.ManagedFields {
 		if r.ObjectMeta.ManagedFields[i].FieldsV1 == nil {
 			continue
 		}
 		var fields map[string]interface{}
+		if len(r.ObjectMeta.ManagedFields[i].FieldsV1.Raw) == 0 {
+			continue
+		}
+		if !bytes.Contains(r.ObjectMeta.ManagedFields[i].FieldsV1.Raw, []byte("\"f:spec\"")) {
+			continue
+		}
+		// json.Unmarshal is called here to strictly check if "f:spec" is a top-level key.
+		// While unmarshalling adds some overhead, it is necessary for correctness (vs simple string matching),
+		// and the cost is acceptable given the low frequency of calls (per status update) and typical small size of managed fields.
 		if err := json.Unmarshal(r.ObjectMeta.ManagedFields[i].FieldsV1.Raw, &fields); err != nil {
-			klog.Warningf("managedfields: error unmarshalling FieldsV1.Raw for manager %v: %v", r.ObjectMeta.ManagedFields[i].Manager, err)
+			klog.V(2).Infof("managedfields: error unmarshalling FieldsV1.Raw for manager %v: %v", r.ObjectMeta.ManagedFields[i].Manager, err)
 			continue
 		}
 		if _, ok := fields["f:spec"]; ok {
--- a/pkg/k8s/managedfields_test.go
+++ b/pkg/k8s/managedfields_test.go
@@ -16,7 +16,6 @@ package k8s_test
 
 import (
 	"encoding/json"
-	"reflect"
 	"testing"
 
 	corekccv1alpha1 "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/apis/core/v1alpha1"
@@ -24,7 +23,9 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test"
 	testk8s "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/test/k8s"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/util"
 
+	"github.com/google/go-cmp/cmp"
 	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/structured-merge-diff/v4/fieldpath"
@@ -735,7 +736,7 @@ func mapToManagedFieldEntry(t *testing.T, manager string, fields map[string]inte
 	}
 }
 
-func TestSanitizeManagedFields(t *testing.T) {
+func TestSanitizeSpecManagedFields(t *testing.T) {
 	tests := []struct {
 		name                string
 		managedFieldEntries []v1.ManagedFieldsEntry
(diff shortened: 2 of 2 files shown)


pkg/controller/lifecyclehandler/handler.go
acpana, line 1: > This causes incorrect managedFields metadata where spec updates are attributed to the status subresource, leading to 'Location must be set' errors when reconciling a CloudSchedulerJob because the controller ignores spec fields managed by status.

This doesn't have to be part of this PR, but I think an e2e looking fixture test where we see the "Location must be set" error go away with this change would be really valuable!

maqiuyujoyce: Thanks for the suggestion!
One tricky thing is that this test is not consistently reproducible. I usually needed to try multiple times throughout an hour or two to successfully reproduce the issue. For now I rely on manual test and I feel we may not want to spend more time to figure out how to consistently reproduce it.

pkg/k8s/managedfields.go
```diff
@@ -163,6 +165,24 @@ func GetK8sManagedFields(u *unstructured.Unstructured) (*fieldpath.Set, error) {
 	return nil, nil
 }
 
+// SanitizeSpecManagedFields ensures that managed fields entries for spec fields do not have
+// Subresource field configured.
```
cheftako, line 169: AGENT: Consider checking if `r` or `r.ObjectMeta.ManagedFields` is nil or empty at the start of the function to return early and avoid unnecessary processing.

TAKO: +1, or `r.ObjectMeta`. Not sure if it saves on processing but it seems cleaner/safer.

pkg/k8s/managedfields.go
```diff
@@ -163,6 +165,24 @@ func GetK8sManagedFields(u *unstructured.Unstructured) (*fieldpath.Set, error) {
...
+func SanitizeSpecManagedFields(r *Resource) {
+	for i := range r.ObjectMeta.ManagedFields {
+		if r.ObjectMeta.ManagedFields[i].FieldsV1 == nil {
+			continue
+		}
+		var fields map[string]interface{}
```
cheftako, line 181: AGENT: If FieldsV1 is not nil but Raw is empty, json.Unmarshal will return an error. You might want to check len(r.ObjectMeta.ManagedFields[i].FieldsV1.Raw) > 0 before unmarshalling.

TAKO: Can you take a look at this?

AGENT: `json.Unmarshal` is called inside the loop. While likely acceptable, for resources with many managed fields, this adds overhead. Since we only check for the top-level `f:spec` key, ensure this performance cost is acceptable for the frequency of these calls.

TAKO: Not too concerned about this one.

pkg/k8s/managedfields.go
```diff
@@ -163,6 +165,24 @@ func GetK8sManagedFields(u *unstructured.Unstructured) (*fieldpath.Set, error) {
...
+		if err := json.Unmarshal(r.ObjectMeta.ManagedFields[i].FieldsV1.Raw, &fields); err != nil {
```
cheftako, line 191: AGENT: This warning log could be noisy if a resource has persistently malformed `managedFields`. Consider using `klog.V(2)` or similar to avoid spamming production logs.

TAKO: Wish we had an edge detection logger. (Log the first time it happens not all). Wondering if the Raw check above would be enough to keep occurrences of the log message down. For now I think we are better of going with the V(2).

pkg/k8s/managedfields.go
```diff
@@ -163,6 +165,24 @@ func GetK8sManagedFields(u *unstructured.Unstructured) (*fieldpath.Set, error) {
 	return nil, nil
 }
 
+// SanitizeSpecManagedFields ensures that managed fields entries for spec fields do not have
+// Subresource field configured.
+func SanitizeSpecManagedFields(r *Resource) {
```
cheftako: If `FieldsV1` is not nil but `Raw` is empty, `json.Unmarshal` will return an error. You might want to check `len(r.ObjectMeta.ManagedFields[i].FieldsV1.Raw) > 0` before unmarshalling.

pkg/k8s/managedfields.go
```diff
@@ -163,6 +165,24 @@ func GetK8sManagedFields(u *unstructured.Unstructured) (*fieldpath.Set, error) {
...
+	for i := range r.ObjectMeta.ManagedFields {
+		if r.ObjectMeta.ManagedFields[i].FieldsV1 == nil {
+			continue
+		}
+		var fields map[string]interface{}
+		if err := json.Unmarshal(r.ObjectMeta.ManagedFields[i].FieldsV1.Raw, &fields); err != nil {
+			klog.Warningf("managedfields: error unmarshalling FieldsV1.Raw for manager %v: %v", r.ObjectMeta.ManagedFields[i].Manager, err)
+			continue
+		}
```
cheftako, line 194: AGENT: By clearing the `Subresource`, if this entry also legitimately manages `f:status` (in a mixed scenario), the ownership of `f:status` effectively moves to the main resource. Please confirm this side effect is intended and safe.

pkg/k8s/managedfields.go
```diff
@@ -163,6 +165,24 @@ func GetK8sManagedFields(u *unstructured.Unstructured) (*fieldpath.Set, error) {
 	return nil, nil
 }
 
+// SanitizeSpecManagedFields ensures that managed fields entries for spec fields do not have
```
cheftako, line 168: AGENT: It would be helpful to add a comment explaining *why* this sanitization is needed (referencing the leaked status subresource issue) for future maintainers.

TAKO: +1

pkg/controller/lifecyclehandler/handler.go
```diff
@@ -78,6 +78,7 @@ func (r *LifecycleHandler) updateStatus(ctx context.Context, resource *k8s.Resou
 	if err := util.Marshal(u, resource); err != nil {
 		return err
 	}
+	k8s.SanitizeSpecManagedFields(resource)
```
cheftako, line 81: AGENT: `SanitizeSpecManagedFields` modifies `resource` in place. Ensure `resource` is not shared or used elsewhere where the original `managedFields` state might be expected.

TAKO: I'm guessing you intend to modify in place. However I prefer the golang assign back paradigm, as it makes it more clear what is going on. I think it would be better to make the SanitizeSpecManagedFields non-mutating/idempotent and then assign back to change the field. So the call would be more like  `resource = k8s.SanitizeSpecManagedFields(resource)` similar to `slice = append(slice, element1, element2, ...)`

There are some perf issues with this but I think the readability/maintainability may be worth the cost. Not currently viewing this as blocking.

pkg/controller/lifecyclehandler/handler.go
```diff
@@ -110,6 +111,7 @@ func (r *LifecycleHandler) updateAPIServer(ctx context.Context, resource *k8s.Re
 	if err := util.Marshal(u, resource); err != nil {
 		return fmt.Errorf("error syncing updated resource metadata: %w", err)
 	}
+	k8s.SanitizeSpecManagedFields(resource)
```
cheftako, line 114: AGENT: Similar to the comment in `updateStatus`, verify that modifying `resource` here does not have unintended side effects on other consumers of this object.

TAKO: Yup, similar to updateStatus.

pkg/k8s/managedfields_test.go
```diff
@@ -16,6 +16,7 @@ package k8s_test
 
 import (
 	"encoding/json"
+	"reflect"
```
cheftako: AGENT: You are adding `reflect` here. If the codebase uses `github.com/google/go-cmp/cmp` (which is common in K8s projects), consider using it instead of `reflect.DeepEqual` for more readable diffs on failure.

TAKO: +1

pkg/k8s/managedfields_test.go
```diff
@@ -733,3 +734,127 @@ func mapToManagedFieldEntry(t *testing.T, manager string, fields map[string]inte
 		FieldsV1:   &v1.FieldsV1{Raw: b},
 	}
 }
+
+func TestSanitizeManagedFields(t *testing.T) {
```
cheftako: AGENT: Consider adding a test case where `FieldsV1` contains `f:spec` but is deeply nested or formatted differently (though `json.Unmarshal` handles formatting) to ensure robustness.

TAKO: +1

pkg/controller/lifecyclehandler/handler.go
maqiuyujoyce: I think this is a reasonable concern.

For now, we want to do a minimized fix, and the in-place modification is consistent with the existing updateStatus / updateAPIServer semantics (which mutate the input resource) so I won't be concerned about it.

The next improvement would be that we stop doing in-place modification.

pkg/controller/lifecyclehandler/handler.go
maqiuyujoyce: Similar with my response above.

pkg/k8s/managedfields.go
maqiuyujoyce: Done.

pkg/k8s/managedfields.go
maqiuyujoyce: Done.

r.ObjectMeta is not a pointer so this one is not checked.

pkg/k8s/managedfields.go
maqiuyujoyce: > AGENT: If FieldsV1 is not nil but Raw is empty, json.Unmarshal will return an error. You might want to check len(r.ObjectMeta.ManagedFields[i].FieldsV1.Raw) > 0 before unmarshalling.
> 
>TAKO: Can you take a look at this?

Done.

> AGENT: json.Unmarshal is called inside the loop. While likely acceptable, for resources with many managed fields, this adds overhead. Since we only check for the top-level f:spec key, ensure this performance cost is acceptable for the frequency of these calls.

We already checked the top level key. One improvement would be checking whether raw contains "f:spec" keyword.

pkg/k8s/managedfields.go
maqiuyujoyce: Done.

pkg/k8s/managedfields.go
maqiuyujoyce: `f:spec` should never have `subresource: status`. So even if the entry has `f:status`, it's still valid to cleanup subresource field.

pkg/k8s/managedfields_test.go
maqiuyujoyce: Done.

pkg/k8s/managedfields_test.go
maqiuyujoyce: Done.

cheftako:
@maqiuyujoyce Can we get this back-ported to the 1.134 branch?

pkg/controller/lifecyclehandler/handler.go
maqiuyujoyce: Actually I added "TestUnmarshalMergeBehavior" to simulate the error and the fix.

PR #6028, opened 2026-01-14
feat: Unify Composer Enviroment Reference, Identity and fix labels

yuwenma:
### BRIEF Change description

Fixes #

Previously we have agreed to unify the interface uses as 
1. Implement identity.Identity in *_identity.go 
2. Use refs.Ref instead of ExternalNormalizer in *_reference.go  
3. Simplify the comment ProjectRef and Location to use the wrapper parent.ProjectAndLocationRef.

This PR also fixes the fixture tests naming to follow the enforced convention.

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/crds/batch_v1alpha1_batchjob.yaml
+++ b/crds/batch_v1alpha1_batchjob.yaml
@@ -1,4 +1,4 @@
-# Copyright 2025 Google LLC
+# Copyright 2026 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/crds/bigquerybiglake_v1alpha1_biglakecatalog.yaml
+++ b/crds/bigquerybiglake_v1alpha1_biglakecatalog.yaml
@@ -1,4 +1,4 @@
-# Copyright 2025 Google LLC
+# Copyright 2026 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/crds/certificatemanager_v1alpha1_certificatemanagercertificateissuanceconfig.yaml
+++ b/crds/certificatemanager_v1alpha1_certificatemanagercertificateissuanceconfig.yaml
@@ -1,4 +1,4 @@
-# Copyright 2025 Google LLC
+# Copyright 2026 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/crds/clouddms_v1alpha1_clouddmsmigrationjob.yaml
+++ b/crds/clouddms_v1alpha1_clouddmsmigrationjob.yaml
@@ -1,4 +1,4 @@
-# Copyright 2025 Google LLC
+# Copyright 2026 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/crds/composer_v1beta1_composerenvironment.yaml
+++ b/crds/composer_v1beta1_composerenvironment.yaml
@@ -1,4 +1,4 @@
-# Copyright 2025 Google LLC
+# Copyright 2026 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/crds/configdelivery_v1alpha1_configdeliveryresourcebundle.yaml
+++ b/crds/configdelivery_v1alpha1_configdeliveryresourcebundle.yaml
@@ -1,4 +1,4 @@
-# Copyright 2025 Google LLC
+# Copyright 2026 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/crds/dataplex_v1alpha1_dataplexentrygroup.yaml
+++ b/crds/dataplex_v1alpha1_dataplexentrygroup.yaml
@@ -1,4 +1,4 @@
-# Copyright 2025 Google LLC
+# Copyright 2026 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/crds/dataplex_v1alpha1_dataplexentrytype.yaml
+++ b/crds/dataplex_v1alpha1_dataplexentrytype.yaml
@@ -1,4 +1,4 @@
-# Copyright 2025 Google LLC
+# Copyright 2026 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/crds/dataplex_v1alpha1_dataplexlake.yaml
+++ b/crds/dataplex_v1alpha1_dataplexlake.yaml
@@ -1,4 +1,4 @@
-# Copyright 2025 Google LLC
+# Copyright 2026 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.


apis/composer/v1beta1/environment_types.go
```diff
@@ -26,20 +27,11 @@ import (
...
-	// Required.
-	Location string `json:"location"`
-}
-
 // ComposerEnvironmentSpec defines the desired state of ComposerEnvironment
 // +kcc:spec:proto=google.cloud.orchestration.airflow.service.v1.Environment
 type ComposerEnvironmentSpec struct {
-	Parent `json:",inline"`
+	// Required. Defines the parent path of the resource.
+	ParentRef *parent.ProjectAndLocationRef `json:",inline"`
```
cheftako, line 34: AGENT: Changing `Parent` (struct) to `ParentRef` (pointer) is a breaking change for Go clients. Ensure this is acceptable.

TAKO: Technically correct. Not sure we intend anyone to be using our controller code as a library. The "good" news is if they do pull this in as a library, changing the name should break they build. That should make it clear they need to do a manual review of change.

AGENT: Also, `parent.ProjectAndLocationRef` does not seem to carry the `x-kubernetes-validations: rule="self == oldSelf"` for `location` that was previously present in the `Parent` struct. Removing this allows users to mutate `location`, which likely isn't supported by the API and will fail at runtime.

TAKO: It would be nice to mark each of the keys as immutable.

apis/composer/v1beta1/environment_identity.go
```diff
@@ -39,79 +42,57 @@ func (i *EnvironmentIdentity) ID() string {
...
-func NewEnvironmentIdentity(ctx context.Context, reader client.Reader, obj *ComposerEnvironment) (*EnvironmentIdentity, error) {
+func (obj *ComposerEnvironment) GetIdentity(ctx context.Context, reader client.Reader) (identity.Identity, error) {
+	environment := &EnvironmentIdentity{
+		parent: &parent.ProjectAndLocationParent{},
+	}
 
-	// Get Parent
-	projectRef, err := refsv1beta1.ResolveProject(ctx, reader, obj.GetNamespace(), obj.Spec.ProjectRef)
-	if err != nil {
+	// Resolve user-configured Parent
```
cheftako, line 83: AGENT: `obj.Spec.ParentRef` is now a pointer. If it is nil, this call will panic. Please add a nil check.

TAKO: +1

apis/composer/v1beta1/environment_identity.go
```diff
@@ -39,79 +42,57 @@ func (i *EnvironmentIdentity) ID() string {
...
 
-func (i *EnvironmentIdentity) Parent() *EnvironmentParent {
+func (i *EnvironmentIdentity) Parent() *parent.ProjectAndLocationParent {
 	return i.parent
 }
 
-type EnvironmentParent struct {
-	ProjectID string
-	Location  string
+func (i *EnvironmentIdentity) FromExternal(ref string) error {
```
cheftako, line 60: AGENT: This parsing logic assumes the external reference strictly follows the `.../environments/...` pattern. Ensure this is robust enough for all valid resource IDs, or that invalid formats are handled gracefully without confusing error messages.

TAKO: https://docs.cloud.google.com/asset-inventory/docs/asset-names for composer supports that //composer.googleapis.com/projects/PROJECT_ID/locations/LOCATION/environments/ENVIRONMENT is the correct for for a composer environment. It would be nice to see tests for error handling for bad parses.

pkg/controller/direct/composer/environment_controller.go
```diff
@@ -70,11 +72,28 @@ func (m *modelEnvironment) AdapterForObject(ctx context.Context, reader client.R
 		return nil, fmt.Errorf("error converting to %T: %w", obj, err)
 	}
 
-	id, err := krm.NewEnvironmentIdentity(ctx, reader, obj)
+	i, err := obj.GetIdentity(ctx, reader)
```
cheftako, line 75: AGENT: This type assertion `i.(*krm.EnvironmentIdentity)` is unchecked. While likely safe given the current codebase, it's safer to check the boolean result to prevent potential panics during future refactors or if `GetIdentity` behavior changes.

TAKO: I think this belongs on line 79. It looks like the intent was to check the boolean on a type assertion but there is a bug on capturing the boolean value.

pkg/controller/direct/composer/environment_controller.go
```diff
@@ -70,11 +72,28 @@ func (m *modelEnvironment) AdapterForObject(ctx context.Context, reader client.R
...
+		return nil, err
+	}
 
+	mapCtx := &direct.MapContext{}
+	copied := obj.DeepCopy()
+	desired := ComposerEnvironmentSpec_ToProto(mapCtx, &copied.Spec)
+	if mapCtx.Err() != nil {
+		return nil, mapCtx.Err()
+	}
+	if len(desired.Labels) == 0 {
```
cheftako: AGENT: The logic `len(desired.Labels) == 0` covers both "nil labels" and "empty labels map". If a user explicitly sets `labels: {}` to clear labels, this logic will overwrite it with `metadata.labels`. Consider checking if `desired.Labels` is nil if preserving an explicit empty map is important.

TAKO: +1 and I think it would be good to add an explicit test via YAML for this.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_composerenvironments.composer.cnrm.cloud.google.com.yaml
```diff
@@ -990,14 +990,9 @@ spec:
                    size.
                 type: object
               location:
-                description: Immutable. The name of the location where the Environment
```
cheftako, line 993: The immutability validation for `location` has been removed. This is a regression in validation strictness. If the underlying API does not support moving environments, it is better to fail fast here.

apis/composer/v1beta1/environment_identity.go
```diff
@@ -39,79 +42,57 @@ func (i *EnvironmentIdentity) ID() string {
...
-		actualParent, actualResourceID, err := ParseEnvironmentExternal(externalRef)
-		if err != nil {
-			return nil, err
+		statusIdentity := &EnvironmentIdentity{}
+		if err := statusIdentity.FromExternal(externalRef); err != nil {
+			return nil, fmt.Errorf("cannot parse existing externalRef=%q: %w", externalRef, err)
 		}
-		if actualParent.ProjectID != projectID {
-			return nil, fmt.Errorf("spec.projectRef changed, expect %s, got %s", actualParent.ProjectID, projectID)
+		if statusIdentity.String() != environment.String() {
```
cheftako, line 107: AGENT: Good check to ensure the external reference in status matches the spec. This prevents "hijacking" of the identity.

TAKO: We should probably have a document for users who hit this issue and the error should point to the doc. Otherwise I'm not sure the customer will have any clue how to resolve the problem they have found themselves in.

pkg/controller/direct/composer/environment_controller.go
```diff
@@ -70,11 +72,28 @@ func (m *modelEnvironment) AdapterForObject(ctx context.Context, reader client.R
...
+	i, err := obj.GetIdentity(ctx, reader)
 	if err != nil {
 		return nil, err
 	}
+	id := i.(*krm.EnvironmentIdentity)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := common.NormalizeReferences(ctx, reader, obj, nil); err != nil {
```
cheftako, line 84: AGENT: Passing `nil` for `ResolveOptions` in `NormalizeReferences`. Ensure that no specific options (like `CallerProjectID`) are needed for this resource.

TAKO: Worth checking.

apis/composer/v1beta1/environment_identity.go
cheftako: TAKO: I think this should be switched to use gcpurls ala https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/6014 or https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/6074

apis/composer/v1beta1/environment_reference.go
```diff
@@ -35,49 +32,40 @@ type EnvironmentRef struct {
 	// Should be in the format "projects/{{projectID}}/locations/{{location}}/environments/{{environmentID}}".
 	External string `json:"external,omitempty"`
 
-	// The name of a ComposerEnvironment resource.
+	// The name of a ComposerEnvironment resource.                                                                                          │
```
cheftako: TAKO: Why are we ending the comment with a pipe '|' ?

pkg/test/resourcefixture/testdata/basic/composer/v1beta1/composerenvironment/composerenvironment-minimal/_generated_object_composerenvironment-minimal.golden.yaml
```diff
@@ -4,14 +4,16 @@ metadata:
   finalizers:
   - cnrm.cloud.google.com/finalizer
   - cnrm.cloud.google.com/deletion-defender
-  generation: 2
+  generation: 1
```
cheftako, line 7: TAKO: Are we sure observed generation is being properly incremented with changes?

apis/composer/v1beta1/environment_types.go
yuwenma: I modified the `location` to be immutable (projectRef is mutable, as user can either use `.external` or the KCC `.name`/`.namespace` combination). This changes some other CRDs since the parent is a shared struct.

apis/composer/v1beta1/environment_identity.go
yuwenma: Ok, we should update the KCC example resource BigLake to use gcpurls as well. Otherwise future PRs could learn from the outdated example. I can do that in a separate PR.

pkg/test/resourcefixture/testdata/basic/composer/v1beta1/composerenvironment/composerenvironment-minimal/_generated_object_composerenvironment-minimal.golden.yaml
yuwenma: Ah, I see what happens. The original test also contains update.yaml, which makes the generation to be 2. As I resolve merge conflict, and move the test files to a new dir (to meet a new presubmit check about fixture test path name convention), the update.yaml was "renamed" to be the dependencies.yaml.

pkg/controller/direct/composer/environment_controller.go
yuwenma: Passing nil is fine and we don't need the CallerProjectID here because the projectRef has been resolved/normalized ahead (as part of the identity initialization).

apis/composer/v1beta1/environment_identity.go
yuwenma: The error will show in their CR `.status`.

PR #6103, opened 2026-01-16
chore: fix a few gofmt / addlicense issues

justinsb:
These escaped our machinery, but they were flagged by the ap tool.


yuwenma (commented):
/lgtm
/approve

PR #6127, opened 2026-01-18
Add jsonunmarshalreuse linter to detect unsafe unmarshal reuse

maqiuyujoyce:
### BRIEF Change description

Added a new custom linter, jsonunmarshalreuse, to detect potentially unsafe or confusing reuse of non-empty variables when calling json.Unmarshal() (and pkg/util.Marshal()), and added a tool to filter linter output to only report violations on changed lines.

Motivation:
Reusing a non-empty variable as the destination for json.Unmarshal can lead to unexpected behavior:
   * Slices: The destination slice is reset to length 0 before appending, causing existing elements to be lost (overwritten).
   * Maps & Structs: The new data is merged into the existing object, which can lead to mixed state if not intended.

This linter aims to catch these patterns to prevent subtle bugs in future PRs. Here is an example: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/6470

```release-note
NONE
```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/dev/linters/jsonunmarshalreuse/analyzer.go
+++ b/dev/linters/jsonunmarshalreuse/analyzer.go
@@ -45,26 +45,33 @@ func run(pass *analysis.Pass) (interface{}, error) {
 		call := n.(*ast.CallExpr)
 
 		// Check function call identifier
-		sel, ok := call.Fun.(*ast.SelectorExpr)
-		if !ok {
+		var funName string
+		var funObj types.Object
+		isTarget := false
+
+		if sel, ok := call.Fun.(*ast.SelectorExpr); ok {
+			funName = sel.Sel.Name
+			funObj = pass.TypesInfo.ObjectOf(sel.Sel)
+		} else if ident, ok := call.Fun.(*ast.Ident); ok {
+			funName = ident.Name
+			funObj = pass.TypesInfo.ObjectOf(ident)
+		} else {
 			return
 		}
 
-		isTarget := false
-
 		// Target 1: encoding/json.Unmarshal
-		if sel.Sel.Name == "Unmarshal" {
-			if obj := pass.TypesInfo.ObjectOf(sel.Sel); obj != nil {
-				if pkg := obj.Pkg(); pkg != nil && pkg.Path() == "encoding/json" {
+		if funName == "Unmarshal" {
+			if funObj != nil {
+				if pkg := funObj.Pkg(); pkg != nil && pkg.Path() == "encoding/json" {
 					isTarget = true
 				}
 			}
 		}
 
 		// Target 2: pkg/util.Marshal (internally performs json.Unmarshal into its second argument)
-		if !isTarget && sel.Sel.Name == "Marshal" {
-			if obj := pass.TypesInfo.ObjectOf(sel.Sel); obj != nil {
-				if pkg := obj.Pkg(); pkg != nil && strings.HasSuffix(pkg.Path(), "/pkg/util") {
+		if !isTarget && funName == "Marshal" {
+			if funObj != nil {
+				if pkg := funObj.Pkg(); pkg != nil && strings.HasSuffix(pkg.Path(), "/pkg/util") {
 					isTarget = true
 				}
 			}


dev/linters/jsonunmarshalreuse/analyzer.go
```diff
@@ -0,0 +1,218 @@
...
+				if pkg := obj.Pkg(); pkg != nil && pkg.Path() == "encoding/json" {
+					isTarget = true
+				}
+			}
+		}
+
+		// Target 2: pkg/util.Marshal
+		if !isTarget && sel.Sel.Name == "Marshal" {
+			if obj := pass.TypesInfo.ObjectOf(sel.Sel); obj != nil {
+				if pkg := obj.Pkg(); pkg != nil && pkg.Path() == "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/util" {
```
maqiuyujoyce: Hardcoding the package path `github.com/GoogleCloudPlatform/k8s-config-connector/pkg/util` makes this linter brittle to project renames or forks. Consider making this configurable or deriving it relative to the project root if possible.

dev/linters/jsonunmarshalreuse/analyzer.go
```diff
@@ -0,0 +1,218 @@
...
+							// Should not happen for valid Go code, but be safe.
+							allInitializedFieldsIgnored = false
+							break
+						}
+						if !allInitializedFieldsIgnored {
+							break
+						}
+					}
+					if !allInitializedFieldsIgnored {
+						return true, "potential reuse of non-empty variable in json.Unmarshal/util.Marshal; consider using an empty literal or nil"
```
maqiuyujoyce: For slices, `json.Unmarshal` resets the slice length to zero before appending, so it effectively overwrites rather than merges (unlike maps/structs).
While initializing a slice with values before unmarshalling is wasteful and confusing, the message "leading to merging" is technically incorrect for slices.
Consider creating a distinct message for slices, e.g., "potential reuse of non-empty slice; existing elements will be lost".

dev/linters/jsonunmarshalreuse/analyzer.go
```diff
@@ -0,0 +1,218 @@
...
+			return false, ""
+		}
+
+		// Helper to check if a make() call is problematic
+		checkMakeCall := func(makeCall *ast.CallExpr) (bool, string) {
+			if fun, isFun := makeCall.Fun.(*ast.Ident); isFun && fun.Name == "make" {
+				if len(makeCall.Args) >= 2 {
+					// The second argument of make is the length
+					// (Note: for maps, this is capacity, not length, so it's always safe)
+					if basicLit, isBasicLit := makeCall.Args[1].(*ast.BasicLit); isBasicLit && basicLit.Kind == token.INT {
```
maqiuyujoyce: This only checks for `token.INT` literals. It will miss constants (e.g., `const N = 5; make([]T, N)`) or simple expressions.
Consider using `pass.TypesInfo.Types[basicLit].Value` to resolve constant values if you want to catch `make([]T, CONST)`.

dev/linters/jsonunmarshalreuse/analyzer.go
```diff
@@ -0,0 +1,218 @@
...
+						if length, err := strconv.Atoi(basicLit.Value); err == nil && length > 0 {
+							return true, "potential reuse of variable created with non-zero length in json.Unmarshal/util.Marshal; consider using make([]T, 0) or nil"
```
maqiuyujoyce: Similar to the comment above, for `make([]T, len)`, `json.Unmarshal` will reset the length to 0. The risk is wastefulness, not merging.

dev/linters/jsonunmarshalreuse/testdata/src/a/bad.go
```diff
@@ -0,0 +1,70 @@
...
+
+func BadUnmarshalUsage() {
+	var data []byte = []byte(`["a", "b"]`)
+	var s []string 
+	
+	if err := json.Unmarshal(data, &s); err != nil { 
+		log.Fatal(err)
+	}
+
+	// Case 1: make with len > 0
```
maqiuyujoyce, line 30: Consider adding a test case using a constant for the length, e.g., `const N = 5; make([]string, N)`, to document whether the linter currently catches or ignores it.

dev/tools/lint-filter/main.go
```diff
@@ -0,0 +1,180 @@
...
+	Start int
+	Count int
+}
+
+// FileChanges maps filenames to their changed line ranges
+type FileChanges map[string][]LineRange
+
+func main() {
+	changes, err := getChanges()
+	if err != nil {
```
maqiuyujoyce, line 37: If `getChanges` fails, printing the error to stderr and proceeding to lint *all* files is a safe fallback. Good choice.

dev/tools/lint-filter/main.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+}
+
+func getChanges() (FileChanges, error) {
+	// Try to find a merge base or common ancestor.
+	// In strict CI, origin/master is usually available.
+	base := "origin/master"
+
+	// Check if origin/master exists
+	cmdCheck := exec.Command("git", "rev-parse", "--verify", "origin/master")
```
maqiuyujoyce: `origin/master` might not be available in all CI environments (e.g., shallow clones or detached HEADs without remote tracking).
Consider allowing the base ref to be configured via an environment variable (e.g., `BASE_REF` or `GIT_BASE`) to support various CI setups more flexibly.

dev/tools/lint-filter/main.go
```diff
@@ -0,0 +1,180 @@
...
+
+	cmd := exec.Command("git", "diff", "--unified=0", base)
+	out, err := cmd.Output()
+	if err != nil {
+		return nil, err
+	}
+
+	changes := make(FileChanges)
+	var currentFile string
+
```
maqiuyujoyce, line 99: `strings.Split` on the entire output might allocate significantly if the diff is huge.
Since you are processing line by line, consider using `bufio.Scanner` here as well for memory efficiency.

dev/tools/lint-filter/main.go
```diff
@@ -0,0 +1,180 @@
...
+			if !strings.HasPrefix(newRange, "+") {
+				continue // Should start with +
+			}
+			newRange = strings.TrimPrefix(newRange, "+")
+
+			start := 0
+			count := 1
+
+			if strings.Contains(newRange, ",") {
+				sub := strings.Split(newRange, ",")
```
maqiuyujoyce, line 124: `strconv.Atoi` error is ignored here. If parsing fails, `count` remains 1. This seems safe but explicit error handling or logging might be helpful for debugging the tool itself.

dev/tools/lint-filter/main.go
```diff
@@ -0,0 +1,180 @@
...
+				start, _ = strconv.Atoi(sub[0])
+				count, _ = strconv.Atoi(sub[1])
+			} else {
```
maqiuyujoyce, line 135: Similarly, ignoring the error for `start` means it defaults to 0. If `newRange` is malformed, this might lead to incorrect filtering.

dev/tools/lint-filter/main.go
```diff
@@ -0,0 +1,180 @@
...
+			if count > 0 {
+				changes[currentFile] = append(changes[currentFile], LineRange{Start: start, Count: count})
+			}
+		}
+	}
+	return changes, nil
+}
+
+func isRelevant(line string, changes FileChanges) bool {
+	// "go run" might print "exit status N" to stderr, which is part of the input to lint-filter.
```
maqiuyujoyce, line 160: Checking for exact string "exit status 3" might be brittle if the exit code changes or if `go run` output format varies.
Consider `strings.HasPrefix(line, "exit status ")` or ignoring lines that don't match the expected `file:line:col` pattern in `isRelevant` (which you already do partially).

dev/tools/lint-filter/main.go
```diff
@@ -0,0 +1,180 @@
...
+	// We need to ignore "exit status 3", as it means violations found.
+	if line == "exit status 3" {
+		return false
+	}
+
```
maqiuyujoyce, line 165: `strings.SplitN(line, ":", 4)` assumes the filename doesn't contain colons before the line number.
While standard for Go tools, if `isRelevant` encounters a line like `file:with:colons.go:10:...`, it might parse incorrectly.
Your current logic falls back to `return true` (pass through) if `Atoi` fails, which is a good failsafe.

dev/linters/jsonunmarshalreuse/analyzer.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+
+		// 2. If it's an identifier, check its declaration
+		if ident, isIdent := expr.(*ast.Ident); isIdent {
+			if obj := ident.Obj; obj != nil && obj.Kind == ast.Var {
+				if decl, ok := obj.Decl.(*ast.ValueSpec); ok {
+					// Check initialization values
+					for _, value := range decl.Values {
+						if problem, msg := isProblematic(value); problem {
+							pass.Reportf(call.Pos(), "%s", msg)
```
maqiuyujoyce, line 138: Reporting at `call.Pos()` is good.
If the variable definition is far away, it might be helpful to include the position of the variable definition in the error message (if available from `ident.Obj`) to help the user locate the "reuse" source.

dev/linters/jsonunmarshalreuse/analyzer.go
```diff
@@ -0,0 +1,218 @@
...
+							allInitializedFieldsIgnored = false
+							break
+						}
+
+						foundField := false
+						for i := 0; i < structType.NumFields(); i++ {
+							field := structType.Field(i)
+							if field.Name() == fieldName {
+								foundField = true
+								jsonTag := reflect.StructTag(structType.Tag(i)).Get("json")
```
maqiuyujoyce: This correctly handles `json:"-"`. Note that `json:"-,omitempty"` means the field is named "-", so strictly checking `== "-"` is correct for ignoring fields. Good.

dev/linters/jsonunmarshalreuse/analyzer.go
```diff
@@ -0,0 +1,218 @@
...
+		// Helper to check if a CompositeLit is problematic
+		checkCompositeLit := func(lit *ast.CompositeLit) (bool, string) {
+			// Check if it's a struct literal
+			if structType, ok := pass.TypesInfo.TypeOf(lit).Underlying().(*types.Struct); ok {
+				if len(lit.Elts) > 0 {
+					allInitializedFieldsIgnored := true
+					for _, elt := range lit.Elts {
+						var fieldName string
+						if kv, isKv := elt.(*ast.KeyValueExpr); isKv {
+							if ident, isIdent := kv.Key.(*ast.Ident); isIdent {
```
maqiuyujoyce: The logic here assumes that if *any* initialized field is NOT ignored, it's a problem. This is correct and safe.

maqiuyujoyce:
All the comments above are from repo agent.

It also had the following summary:

> Overall, this is a great addition to the tooling.
>
> The custom linter helps prevent subtle bugs with json.Unmarshal merging behavior.
I have pointed out a few edge cases in the analyzer logic (specifically regarding slices and constants) and some > robustness improvements for the lint-filter tool.
>
>Please consider the feedback regarding the error message wording for slices.

dev/tools/lint-filter/main.go
maqiuyujoyce: Refactored and now it's using bufio.NewScanner for line-by-line processing of git diff output.

dev/tools/lint-filter/main.go
maqiuyujoyce: Supported LINT_FILTER_GIT_BASE environment variable for caller to specify the git reference for diffing if needed.

dev/tools/lint-filter/main.go
maqiuyujoyce: Ack.

dev/tools/lint-filter/main.go
maqiuyujoyce: Added explicit handling of potential errors from strconv.Atoi.

dev/tools/lint-filter/main.go
maqiuyujoyce: Done as invalid range is now skipped.

dev/tools/lint-filter/main.go
maqiuyujoyce: We chose to explicitly target exit status 3 because it is the specific exit code used by multichecker to signal 'violations found'. Filtering other exit statuses (like exit status 1 or exit status 2) carries the risk of suppressing genuine runtime errors, build failures, or panics from the linter tool itself, which we want to see reported.

While we considered a broader prefix match like strings.HasPrefix(line, "exit status "), we decided that being specific to the known 'success-with-violations' code is safer to avoid masking unexpected failures. The rest of the line processing logic already handles non-diagnostic lines by passing them through, ensuring that any other error output remains visible.

dev/tools/lint-filter/main.go
maqiuyujoyce: It's a good point regarding filenames containing colons. However, because this is a rare case and Go filenames typically do not contain colons (and it's generally discouraged), we don't explicitly handle this edge case in the parsing logic for now. The existing failsafe ensures that if such a filename were encountered, the line would be passed through for manual review rather than being silently suppressed.

dev/linters/jsonunmarshalreuse/testdata/src/a/bad.go
maqiuyujoyce: Done.

dev/linters/jsonunmarshalreuse/analyzer.go
maqiuyujoyce: Done.

dev/linters/jsonunmarshalreuse/analyzer.go
maqiuyujoyce: Ack.

dev/linters/jsonunmarshalreuse/analyzer.go
maqiuyujoyce: Ack.

dev/linters/jsonunmarshalreuse/analyzer.go
maqiuyujoyce: Updated.

dev/linters/jsonunmarshalreuse/analyzer.go
maqiuyujoyce: Updated.

dev/linters/jsonunmarshalreuse/analyzer.go
maqiuyujoyce: Ack.

dev/linters/jsonunmarshalreuse/analyzer.go
maqiuyujoyce: Done.

dev/tools/lint-filter/main.go
```diff
@@ -0,0 +1,209 @@
...
+// FileChanges maps filenames to their changed line ranges
+type FileChanges map[string][]LineRange
+
+func main() {
+	changes, err := getChanges()
+	if err != nil {
+		// If git fails (e.g. no git repo, detached head issues), we warn but pass everything through?
+		// Or fail? Let's fail safe: pass everything through if we can't detect changes (e.g. initial commit or CI weirdness),
+		// OR simpler: just print error and exit.
+		// For now, let's print error and output nothing (assume filtering failed implies safe to ignore or we want to fix env).
```
acpana: Are these gemini thoughts? I tend to delete/ pair these down to what's important

maqiuyujoyce: Yep, those are Gemini thoughts! Let me clean them up.

maqiuyujoyce: Done.

maqiuyujoyce:
> I like the AST approach to adding this linter. I'm not too sold on the git diff parsing in `lint-filter` but I'm willing to try it!

I enabled the linter only for the newly changed code because we have existing anti-patterns that we actually want to support (as part of the tests). You can find more details at https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/6470. In the future, we might be able to support a exemption list for this linter / linters in general.

PR #6191, opened 2026-01-20
type-generator should match spec / status when generating a message

justinsb:
We don't want ObservedState messages to prevent generation of the Spec message,
and vice-versa.

Also introduce some simple caching.


config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_dataplexentrytypes.dataplex.cnrm.cloud.google.com.yaml
yuwenma, line 1: Is this CRD change relevant to the PR? If not, can we drop them?

yuwenma (commented):
/approve

Good catch and fix, thank you! Could you resolve the compile error?

PR #6194, opened 2026-01-20
fix: Don't mark Kind default in TagBinding ParentRef

justinsb:
Setting the default means that we actually populate the value as Project,
but I think we want to keep it unset, because Kind can be optional
if External is specified.

### DIFF
--- a/apis/tags/v1beta1/tagbinding_parentref.go
+++ b/apis/tags/v1beta1/tagbinding_parentref.go
@@ -29,10 +29,9 @@ import (
 // ParentRef is a reference to a parent resource.
 // +kcc:ref=Project;StorageBucket
 type TagsTagBindingParentRef struct {
-	// Kind of the referent.
+	// Kind to which we are binding the tag.  Defaults to Project if not specified.
 	// +optional
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:default=Project
 	Kind string `json:"kind,omitempty"`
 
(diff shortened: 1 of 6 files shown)


apis/tags/v1beta1/tagbinding_parentref.go
```diff
@@ -29,10 +29,9 @@ import (
 // ParentRef is a reference to a parent resource.
 // +kcc:ref=Project;StorageBucket
 type TagsTagBindingParentRef struct {
-	// Kind of the referent.
+	// Kind to which we are binding the tag.  Defaults to Project if not specified.
```
yuwenma, line 32: Shall we add that "Organization is allowed without 'kind' being specified".  Because the existing TF-based TagBinding also allows parent to be an Organization, and since this is supported before the "kind" field is added. This  description makes it look like KCC is not backward compatible for the org case any more. 

Here's an test PR to show how organization is allowed from the single `.spec.parentRef.external` field       
https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5542/files#diff-c13994765c1cc4d47db95e8f8a8741efaa3dce91e736838918aa37a2430b52ae.  Just to call it out since it surprised me at first.

yuwenma (commented):
/approve
/lgtm
/hold  One backward compatibility question to confirm, otherwise look good). Feel free to unhold if the "org as default" has been handled in the code.

PR #6246, opened 2026-01-23
Add reconcilerOverride to preview pkg

### LINKED ISSUE #5980: SaferUpgrade: Add reconciler type override to preview
### Checklist

Add flag so we can set reconciler type for Group Kind during preview run such as

`--reconciler-type-override BigQueryTable.bigquery.cnrm.cloud.google.com:direct`

When the flag is set, all BigQueryTable resource will be reconciled by direct reconciler instead of TF reconciler. The flag can be used multiple times for multiple resources within one run.

### Additional information

_No response_

### Importance

The feature allow us to run different reconciler types with different clusters to help migrating existing DCL/TF  reconciler to direct.

anhdle-sso:
# Refactor Preview CLI to Support Reconciler Overrides
## Description
This PR refactors the Preview CLI to introduce a flexible `ObjectTransformer` mechanism, enabling type-safe reconciler overrides. It extracts the override logic into a dedicated transformer, allowing `ConfigConnectorContext` resources to be dynamically modified during the preview interception process with specific controller overrides (e.g., forcing `direct` or `dcl` mode).
## Changes
- **Transformer Implementation**: Added [pkg/cli/preview/transformer.go] with `newReconcilerOverrideTransformer` to handle `ConfigConnectorContext` modifications.
- **Preview Integration**: Updated [pkg/cli/preview/preview.go] to initialize the transformer in [NewPreviewInstance] and pass it to the interceptor.
- **Interceptor Update**: Modified [pkg/cli/preview/intercepting_kube_client.go] to accept and apply object transformers.
- **Informer Support**: Updated [pkg/cli/preview/streaming_informer.go] to apply object transformers during `OnWatchAdd` events.
- **Testing**: Added [pkg/cli/preview/transformer_test.go] with unit tests covering generic object transformation and specific reconciler override scenarios.

Fixes #5980

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/cli/preview/streaming_informer.go
+++ b/pkg/cli/preview/streaming_informer.go
@@ -67,6 +67,7 @@ func (o *objects) OnListObject(ctx context.Context, obj Object, isInInitialList 
 		for _, transformer := range objectTransformers {
 			if err := transformer(ctx, clientObj); err != nil {
 				klog.Errorf("transformer failed: %v", err)
+				return fmt.Errorf("transformer failed: %w", err)
 			}
 		}
 	}


pkg/cli/preview/intercepting_kube_client.go
```diff
@@ -41,11 +41,12 @@ type interceptingKubeClient struct {
 	upstreamClient     *StreamingClient
 	upstreamRestMapper meta.RESTMapper
 
-	recorder *Recorder
+	reconcilerOverride map[string]string
```
justinsb: This feels very specific to this particular use case.

I think implementing a "filteringKubeClient" that accepts a function that can transform objects is probably a more sustainable pattern.

I'd be OK with having a filter function in newInterceptingKubeClient, but it feels like we could do better.

We could also implement the functionality to just work on a copy of the objects, like in the second PR

anhdle-sso: I added more robust function options for kubeclient

pkg/cli/preview/transformer.go
```diff
@@ -0,0 +1,45 @@
...
+	"context"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	corev1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/apis/core/v1beta1"
+	k8s "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
+)
+
+func newReconcilerOverrideTransformer(reconcilerOverride map[schema.GroupKind]k8s.ReconcilerType) ObjectTransformer {
```
xiaoweim: This function creates a transformer that applies overrides to any `ConfigConnectorContext` without checking the `PreviewInstance.Namespace`.

pkg/cli/preview/preview.go
```diff
@@ -85,7 +93,9 @@ func NewPreviewInstance(recorder *Recorder, options PreviewInstanceOptions) (*Pr
 		upstreamGCPHTTPClient = http.DefaultClient
 	}
 
-	hookKube, err := newInterceptingKubeClient(recorder, upstreamRESTConfig)
+	hookKube, err := newInterceptingKubeClient(recorder, upstreamRESTConfig, []ObjectTransformer{
```
xiaoweim: Namespace is not passed in here too, so any override would apply to all namespace? IIUC `PreviewInstance.Namespace` is supposed to allow namespace level override.

pkg/cli/preview/transformer.go
xiaoweim: AGENT: The `ObjectTransformer` function signature implies in-place mutation of `obj`. While this appears safe in the current usage (as `streamingInformer` stores the result immediately after), it's worth noting that if `obj` were shared or cached upstream, this could have side effects. If strict safety is required, consider working on a deep copy, though for the preview CLI this is likely acceptable.

Xiaowei: Nice to have.

pkg/cli/preview/intercepting_kube_client.go
```diff
@@ -72,6 +76,7 @@ func newInterceptingKubeClient(recorder *Recorder, upstreamRestConfig *rest.Conf
 		upstreamClient:     upstreamClient,
 		upstreamRestMapper: upstreamRestMapper,
 		recorder:           recorder,
+		objectTransformers: objectTransformers,
```
xiaoweim: AGENT: Consider making a defensive copy of the `objectTransformers` slice here (e.g., `append([]ObjectTransformer(nil), objectTransformers...)`). If the caller modifies the backing array of the passed slice later, it could affect the client's state, although unlikely in the current call site.

Xiaowei: Nice to have.

pkg/cli/preview/streaming_informer.go
```diff
@@ -54,7 +54,15 @@ type objects struct {
 }
 
 // OnListObject is called from list, the lock should be held
-func (o *objects) OnListObject(obj Object, isInInitialList bool, eventHandlerRegistrations []*eventHandlerRegistration) {
+func (o *objects) OnListObject(ctx context.Context, obj Object, isInInitialList bool, eventHandlerRegistrations []*eventHandlerRegistration, objectTransformers []ObjectTransformer) error {
+	if clientObj, ok := obj.(client.Object); ok {
+		for _, transformer := range objectTransformers {
```
xiaoweim, line 67: AGENT: If a transformer returns an error, `OnListObject` returns it immediately, which will likely abort the entire `List` operation. Ensure that transformers (and future implementations) are designed to only return errors for fatal conditions that should stop the preview.

Xiaowei: +1. We can have better error handling here.

pkg/cli/preview/streaming_informer.go
```diff
@@ -134,7 +153,12 @@ func (i *streamingInformer) runOnce(ctx context.Context) error {
 	}
 
 	i.hasSynced.Store(true)
-	watchListener := &watchListener{objects: &i.objects, eventHandlerRegistrations: i.eventHandlerRegistrations}
+	watchListener := &watchListener{
```
xiaoweim, line 173: We are capturing `ctx` in the `watchListener`. Ensure that the lifecycle of this `ctx` (passed from `runOnce`) covers the entire duration of the `Watch` call. Typically `Watch` blocks until context cancellation or error, so this should be correct, but valid to double-check.

pkg/cli/preview/preview.go
```diff
@@ -85,7 +93,9 @@ func NewPreviewInstance(recorder *Recorder, options PreviewInstanceOptions) (*Pr
 		upstreamGCPHTTPClient = http.DefaultClient
 	}
```
xiaoweim, line 96: We are creating a new slice literal here `[]ObjectTransformer{...}`. This is good practice as it isolates the slice from `options` (though `options` doesn't have transformers).

pkg/cli/preview/transformer.go
```diff
@@ -0,0 +1,45 @@
...
+	return func(ctx context.Context, obj client.Object) error {
+		if ccc, ok := obj.(*corev1beta1.ConfigConnectorContext); ok {
+			if len(reconcilerOverride) == 0 {
+				return nil
+			}
+			if ccc.Spec.Experiments == nil {
+				ccc.Spec.Experiments = &corev1beta1.Experiments{}
+			}
+			if ccc.Spec.Experiments.ControllerOverrides == nil {
+				ccc.Spec.Experiments.ControllerOverrides = make(map[string]k8s.ReconcilerType)
```
xiaoweim, line 43: AGENT: Good use of `make` to ensure `ControllerOverrides` is initialized before assignment.

Xiaowei: 💯

pkg/cli/preview/intercepting_kube_client.go
```diff
@@ -118,7 +123,7 @@ func (c *interceptingKubeClient) NewCache(restConfig *rest.Config, opts cache.Op
 		break
 	}
 
-	return newInterceptingControllerRuntimeCache(c.upstreamClient, typeStore, namespace)
+	return newInterceptingControllerRuntimeCache(c.upstreamClient, typeStore, namespace, c.objectTransformers)
```
xiaoweim, line 126: The `objectTransformers` are passed down to the cache and then to the informers. Since the transformers are functions and stateless (in this PR), sharing them is fine.

pkg/cli/preview/preview.go
anhdle-sso: done, added the namespace check to the transformer. Though it is might not be necessary, we already restrict to only get CCC of the exact namespace (if specified).

pkg/cli/preview/transformer.go
anhdle-sso: done

pkg/cli/preview/streaming_informer.go
anhdle-sso: done

pkg/cli/preview/preview.go
anhdle-sso: done

pkg/cli/preview/transformer.go
anhdle-sso: fixed

pkg/cli/preview/streaming_informer.go
```diff
@@ -54,7 +54,18 @@ type objects struct {
...
-func (o *objects) OnListObject(obj Object, isInInitialList bool, eventHandlerRegistrations []*eventHandlerRegistration) {
+func (o *objects) OnListObject(ctx context.Context, obj Object, isInInitialList bool, eventHandlerRegistrations []*eventHandlerRegistration, objectTransformers []ObjectTransformer) error {
+	if clientObj, ok := obj.(client.Object); ok {
+		// Deep copy to avoid mutating the original object if it's shared
+		clientObj = clientObj.DeepCopyObject().(client.Object)
+		obj = clientObj
+		for _, transformer := range objectTransformers {
+			if err := transformer(ctx, clientObj); err != nil {
+				klog.Errorf("transformer failed: %v", err)
+			}
```
cheftako, line 71: AGENT: Transformation errors are currently logged and suppressed here. If a transformer fails (e.g., critical override failure), the object continues processing in its original or partially modified state. 

Since the function signature was updated to return `error`, consider whether these errors should be propagated to abort the operation, or if the signature implies a contract that isn't fully utilized yet.

TAKO: Make sure you know how to debug. Assuming we have access to the log, this is probably fine.

pkg/cli/preview/streaming_informer.go
```diff
@@ -54,7 +54,18 @@ type objects struct {
 }
 
 // OnListObject is called from list, the lock should be held
-func (o *objects) OnListObject(obj Object, isInInitialList bool, eventHandlerRegistrations []*eventHandlerRegistration) {
+func (o *objects) OnListObject(ctx context.Context, obj Object, isInInitialList bool, eventHandlerRegistrations []*eventHandlerRegistration, objectTransformers []ObjectTransformer) error {
+	if clientObj, ok := obj.(client.Object); ok {
+		// Deep copy to avoid mutating the original object if it's shared
+		clientObj = clientObj.DeepCopyObject().(client.Object)
+		obj = clientObj
+		for _, transformer := range objectTransformers {
```
cheftako, line 67: AGENT: `DeepCopyObject()` returns a `runtime.Object`. Asserting it directly to `client.Object` (`.(client.Object)`) is generally safe for Kubernetes resources but could panic if the underlying type doesn't implement `client.Object`.

Consider using a safe type assertion (`if cast, ok := ...; !ok { ... }`) to prevent potential runtime panics in the CLI.

TAKO: +1

pkg/cli/preview/streaming_informer.go
```diff
@@ -79,16 +102,18 @@ func (o *objects) OnWatchAdd(obj Object, eventHandlerRegistrations []*eventHandl
 	for _, handler := range eventHandlerRegistrations {
 		handler.handler.OnAdd(obj, false)
 	}
+	return nil
 }
```
cheftako, line 117: AGENT: Similar to `OnListObject`, transformer errors are suppressed in `OnWatchAdd`. If the intention is to allow partial failures, the logging is appropriate, but be aware that this might lead to inconsistent state in the preview if overrides fail silently.

TAKO: Ditto

pkg/cli/preview/streaming_informer.go
```diff
@@ -54,7 +54,18 @@ type objects struct {
...
 
 // OnListObject is called from list, the lock should be held
-func (o *objects) OnListObject(obj Object, isInInitialList bool, eventHandlerRegistrations []*eventHandlerRegistration) {
+func (o *objects) OnListObject(ctx context.Context, obj Object, isInInitialList bool, eventHandlerRegistrations []*eventHandlerRegistration, objectTransformers []ObjectTransformer) error {
+	if clientObj, ok := obj.(client.Object); ok {
+		// Deep copy to avoid mutating the original object if it's shared
+		clientObj = clientObj.DeepCopyObject().(client.Object)
```
cheftako: AGENT: Performing a deep copy on every object during `List` and `Watch` ensures thread safety and isolation for transformers, which is good. However, be mindful of the performance impact if the preview CLI is used against clusters with a very large number of resources.

TAKO: Not that concerned given this is a preview tool.

pkg/cli/preview/transformer.go
```diff
@@ -0,0 +1,48 @@
...
+	"context"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	corev1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/apis/core/v1beta1"
+	k8s "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
+)
+
+func newReconcilerOverrideTransformer(namespace string, reconcilerOverride map[schema.GroupKind]k8s.ReconcilerType) ObjectTransformer {
```
cheftako, line 29: AGENT: The `reconcilerOverride` map is captured by reference in the returned closure. If the caller modifies the map after creating the transformer, it will affect the transformer's behavior. 

While unlikely in the current usage, creating a copy of the map (`maps.Clone` or manual copy) inside this factory function would be safer and more robust.

TAKO: Can we have a better name which describe what this transformer does? Eg setAllControllerTransformer. Not sure we should be modifying the map, so the concern may be somewhat academic.

pkg/cli/preview/preview.go
cheftako: AGENT: The `newReconcilerOverrideTransformer` is currently hardcoded here. 

To make `PreviewInstance` more extensible for future transformation needs, consider allowing `PreviewInstanceOptions` to accept a slice of `ObjectTransformer`s, which can then be combined with the reconciler override transformer.

pkg/cli/preview/transformer.go
```diff
@@ -0,0 +1,48 @@
...
+	return func(ctx context.Context, obj client.Object) error {
+		if ccc, ok := obj.(*corev1beta1.ConfigConnectorContext); ok {
+			if namespace != "" && ccc.GetNamespace() != namespace {
```
cheftako, line 33: AGENT: `ccc.GetNamespace()` is called on `*corev1beta1.ConfigConnectorContext`. Ensure that `ConfigConnectorContext` generated code implements `client.Object` (specifically `GetNamespace`). It typically does via embedding `ObjectMeta`, but good to double-check given the type assertion above.

TAKO. Good habit.

pkg/cli/preview/streaming_informer.go
```diff
@@ -63,10 +74,22 @@ func (o *objects) OnListObject(obj Object, isInInitialList bool, eventHandlerReg
 	for _, handler := range eventHandlerRegistrations {
 		handler.handler.OnAdd(obj, false)
 	}
+	return nil
```
cheftako, line 83: AGENT: The function returns `nil` at the end, making the `error` return type in the signature effectively unused for error propagation. If error suppression is the intended design, this is fine, but it leaves the API slightly ambiguous.

TAKO: What is the plan?

pkg/cli/preview/preview.go
```diff
@@ -44,9 +46,13 @@ type PreviewInstance struct {
 	hookKube *interceptingKubeClient
 	recorder *Recorder
 
+	// Namespace is the namespace of the cluster to preview
```
cheftako, line 49: AGENT: `ReconcilerOverride` is exported in the `PreviewInstance` struct. If this field is intended to be purely configuration status, this is fine. If it's not needed after initialization, you might remove it to keep the struct cleaner.

TAKO: Why is it public? This field seems like it should be package/private.

pkg/cli/preview/preview.go
cheftako: TAKO: Can we separately initialize the transformer slice and then pass that in as a variable to make the code easier to read?

pkg/cli/preview/streaming_informer.go
```diff
@@ -134,7 +159,12 @@ func (i *streamingInformer) runOnce(ctx context.Context) error {
 	}
 
 	i.hasSynced.Store(true)
-	watchListener := &watchListener{objects: &i.objects, eventHandlerRegistrations: i.eventHandlerRegistrations}
+	watchListener := &watchListener{
+		objects:                   &i.objects,
+		eventHandlerRegistrations: i.eventHandlerRegistrations,
+		objectTransformers:        i.objectTransformers,
+		ctx:                       ctx,
```
cheftako: Quibble. context per the API is always supposed to be the first parameter. (https://pkg.go.dev/context)

pkg/cli/preview/streaming_informer.go
```diff
@@ -193,14 +230,15 @@ func (i *listListener) OnListEnd() {
 type watchListener struct {
 	objects                   *objects
 	eventHandlerRegistrations []*eventHandlerRegistration
+	objectTransformers        []ObjectTransformer
+	ctx                       context.Context
```
cheftako: Setting the context on the object is a bad idea. See https://pkg.go.dev/context. Can we not do this? contexts are only supposed to had request lifetimes and setting on struct is asking to have a stale instance.

pkg/cli/preview/transformer.go
```diff
@@ -0,0 +1,48 @@
+// Copyright 2025 Google LLC
```
cheftako: 2026

cheftako (commented):
AGENT: Overall, the changes look good and provide the requested functionality. 

I have pointed out a few areas for improvement:
1.  **Error Handling**: Transformer errors are logged but not propagated, which might mask configuration issues.
2.  **Safety**: A safe cast is recommended when deep-copying objects to avoid potential panics.
3.  **Extensibility**: Consider allowing `PreviewInstanceOptions` to accept generic transformers for future flexibility.
4.  **Concurrency**: Defensive copying of the override map in the transformer factory is checking for best practices.

pkg/cli/preview/transformer.go
anhdle-sso: done

pkg/cli/preview/transformer.go
anhdle-sso: done

pkg/cli/preview/preview.go
anhdle-sso: done, I make the transformer as an option. We can add the transformer preview instance creation

pkg/cli/preview/transformer.go
anhdle-sso: the function name is correct as it takes in the list of overrides.

pkg/cli/preview/streaming_informer.go
anhdle-sso: done

pkg/cli/preview/preview.go
anhdle-sso: Removed this field. Added transformer slices field to be more flexible

pkg/cli/preview/streaming_informer.go
anhdle-sso: I allow partial failures. As we combine to result of the 2 reconciliations (1 normal, 1 direct), if this has error, the result will be missing and the log would show what the issue was. We can make the code change to address that. I do not want one error on one object to break the result of other resources.

pkg/cli/preview/streaming_informer.go
anhdle-sso: done

pkg/cli/preview/streaming_informer.go
anhdle-sso: I am ok with partial failure with logging to point out the failure.

pkg/cli/preview/preview.go
```diff
@@ -98,6 +105,7 @@ func NewPreviewInstance(recorder *Recorder, options PreviewInstanceOptions) (*Pr
 	i.hookKube = hookKube
 	i.recorder = recorder
 	i.Namespace = options.Namespace
+	i.ObjectTransformers = options.ObjectTransformers
```
cheftako: AGENT: `i.ObjectTransformers` is assigned the slice from `options.ObjectTransformers`. Since slices are reference types, modifications to the underlying array in `options` (if reused) would affect the instance. Consider `slices.Clone` or `append` if you want to isolate the `PreviewInstance` state.

TAKO: More concerned outside of preview but may still be good to be in the habit of cloning.

pkg/cli/preview/streaming_informer.go
```diff
@@ -169,8 +214,10 @@ func (i *streamingInformer) doList(ctx context.Context) (*ListMetadata, error) {
 
 // listListener is a listener for list operations.
 type listListener struct {
+	ctx                       context.Context
```
cheftako: AGENT: Storing `context.Context` in the struct (`listListener`) is generally an anti-pattern. Since you successfully updated `WatchListener` to accept `ctx`, consider updating the `ListListener` interface (and `OnListObject`) to also accept `context.Context` to maintain consistency and avoid storing request-scoped context in the struct.

TAKO: +1. From https://go.dev/blog/context-and-structs "Contexts should not be stored inside a struct type, but instead passed to each function that needs it."

cheftako (commented):
Couple of minor points.

pkg/cli/preview/streaming_informer.go
anhdle-sso: fixed

pkg/cli/preview/preview.go
anhdle-sso: fixed

pkg/cli/preview/streaming_informer.go
```diff
@@ -54,7 +54,23 @@ type objects struct {
...
+	if clientObj, ok := obj.(client.Object); ok {
+		// Deep copy to avoid mutating the original object if it's shared
+		copied := clientObj.DeepCopyObject()
+		var okCopy bool
+		clientObj, okCopy = copied.(client.Object)
+		if !okCopy {
+			return fmt.Errorf("DeepCopyObject returned %T, expected client.Object", copied)
+		}
+		obj = clientObj
+		for _, transformer := range objectTransformers {
```
cheftako, line 67: Transformer errors silently swallowed — streaming_informer.go (Medium)

  In OnListObject and OnWatchAdd, errors from transformers are only logged:

  for _, transformer := range objectTransformers {
      if err := transformer(ctx, clientObj); err != nil {
          klog.Errorf("transformer failed: %v", err)
      }
  }

  The object ends up stored in the cache and passed to event handlers in a
  potentially inconsistent state. Since transformers here are critical
  (injecting controller overrides), a silent failure means preview runs with
  incorrect behavior with no indication to the caller.

  Recommendation: Return the error (return fmt.Errorf("transformer failed: %w",
  err)) or explicitly document the swallow-and-continue decision.

anhdle-sso: done

cheftako (approved):
Looks good.

PR #6260, opened 2026-01-23
Add CompareProtoMessageStructuredDiff for detailed diff reporting

### LINKED ISSUE #6247: feature: new CompareProtoMessage to return both paths and structured diff
Create CompareProtoMessageDiff to return both paths and structured diff when compare 2 proto message. Use `CompareProtoMessage` in pkg/controller/direct/common/compare.go as base line

anhdle-sso:
# Add CompareProtoMessageStructuredDiff for detailed diff reporting

## Description
This PR introduces `CompareProtoMessageStructuredDiff` in `pkg/controller/direct/common`, a new utility function that compares two proto messages and returns:
1. A set of changed field paths (compatible with the existing `CompareProtoMessage`).
2. A `*structuredreporting.Diff` object containing detailed information about the changes (old vs new values) for each field.

This enhancement supports richer diff reporting mechanisms, allowing consumers to process and display granular difference details rather than just field paths.

## Key Changes
- **New Function**: `CompareProtoMessageStructuredDiff` recursively compares proto messages, accumulating both modified field paths and structured diff data.
- **Structured Reporting**: Integrates with `structuredreporting.DiffField` to capture `Old` and `New` values for changed fields.
- **Robustness**: Handles nested messages and correctly reconstructs dot-notated field paths during recursion.

## Testing
- Added `pkg/controller/direct/common/compare_test.go`.
- Implemented `TestCompareProtoMessageStructuredDiff` covering:
    - Identical messages (no diffs).
    - Simple scalar value changes.
    - Nested message changes (using `descriptorpb` to avoid complex dependencies).
- Verified that both the returned path set and the structured diff fields match expectations.

Fixes #6247

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/controller/direct/common/compare_test.go
+++ b/pkg/controller/direct/common/compare_test.go
@@ -23,16 +23,35 @@ import (
 	"google.golang.org/protobuf/reflect/protoreflect"
 	"google.golang.org/protobuf/testing/protocmp"
 	"google.golang.org/protobuf/types/descriptorpb"
+	"google.golang.org/protobuf/types/known/durationpb"
 	"google.golang.org/protobuf/types/known/structpb"
+	"google.golang.org/protobuf/types/known/timestamppb"
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 func protoMapToMap(m protoreflect.Map) map[string]interface{} {
 	res := make(map[string]interface{})
 	m.Range(func(k protoreflect.MapKey, v protoreflect.Value) bool {
-		res[k.String()] = v.Message().Interface()
+		if mv, ok := v.Interface().(protoreflect.Message); ok {
+			res[k.String()] = mv.Interface()
+		} else {
+			res[k.String()] = v.Interface()
+		}
 		return true
 	})
+	return res
+}
+
+func protoListToList(l protoreflect.List) []interface{} {
+	res := make([]interface{}, l.Len())
+	for i := 0; i < l.Len(); i++ {
+		v := l.Get(i)
+		if m, ok := v.Interface().(protoreflect.Message); ok {
+			res[i] = m.Interface()
+		} else {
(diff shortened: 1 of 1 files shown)


pkg/controller/direct/common/compare_test.go
```diff
@@ -0,0 +1,160 @@
...
+			name:      "different string value",
+			msgA:      fooVal,
+			msgB:      barVal,
+			wantPaths: sets.New("string_value"),
+			wantDiffs: []structuredreporting.DiffField{
+				{ID: "string_value", Old: "bar", New: "foo"},
+			},
+		},
+		{
+			name:      "nested message value",
```
cheftako, line 131: You should also have a descA compare to descA test case.

pkg/controller/direct/common/compare_test.go
```diff
@@ -0,0 +1,160 @@
...
+	res := make(map[string]interface{})
+	m.Range(func(k protoreflect.MapKey, v protoreflect.Value) bool {
+		res[k.String()] = v.Message().Interface()
+		return true
+	})
+	return res
+}
+
+func TestCompareProtoMessageStructuredDiff(t *testing.T) {
+	fooVal, err := structpb.NewValue("foo")
```
cheftako, line 59: Can we add same to same comparison which is not the same memory? (verify value vs memory comparitor correctness)

It would be good to have some list tests.
In fact I would ask your LLM to add test cases to improve coverage.

pkg/controller/direct/common/compare_test.go
anhdle-sso: done

pkg/controller/direct/common/compare_test.go
anhdle-sso: done

Issue #6318, opened 2026-01-27
Support CloudDeployTarget in v1alpha1 direct


katrielt:
Support CloudDeployTarget in v1alpha with a direct controller.

This is an issue to to aggregate the PRs required

@yuwenma FYI.

PR #6358, opened 2026-01-29
feat: add CloudBuildWorkerPool identity and reference types

katrielt:
### BRIEF Change description
add CloudBuildWorkerPool identity and reference types

Part of the work required for #6318

Unblocks the proper on boarding of CloudDeployTarget

### DIFF
--- a/apis/cloudbuild/v1beta1/workerpool_identity.go
+++ b/apis/cloudbuild/v1beta1/workerpool_identity.go
@@ -1,0 +1,113 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1beta1
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/apis/common"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/apis/common/identity"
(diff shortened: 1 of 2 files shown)


apis/cloudbuild/v1beta1/workerpool_identity.go
```diff
@@ -0,0 +1,113 @@
...
+	_ identity.Resource   = &CloudBuildWorkerPool{}
+)
+
+var CloudBuildWorkerPoolIdentityFormat = gcpurls.Template[CloudBuildWorkerPoolIdentity]("cloudbuild.googleapis.com", "projects/{project}/locations/{location}/workerPools/{resourceID}")
+
+// +k8s:deepcopy-gen=false
+type CloudBuildWorkerPoolIdentity struct {
+	Project    string
+	Location   string
+	ResourceID string
```
justinsb, line 38: Super nit: I have been starting to call this WorkerPool or WorkerPoolID.  I just find it easier personally, when using it.  In no way a blocker though!

apis/cloudbuild/v1beta1/workerpool_identity.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	// Get Location
+	location := obj.Spec.Location
+	if location == "" {
+		return nil, fmt.Errorf("cannot resolve location")
+	}
+
+	// Get Project
+	projectID := obj.GetAnnotations()["cnrm.cloud.google.com/project-id"]
```
justinsb, line 79: I think we have a helper for this somewhere, but not a blocker (we can always ask the robots to tidy this when we need it)

apis/cloudbuild/v1beta1/workerpool_reference.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func (r *CloudBuildWorkerPoolRef) Normalize(ctx context.Context, reader client.Reader, defaultNamespace string) error {
+	fallback := func(u *unstructured.Unstructured) string {
+		resourceID, err := refs.GetResourceID(u)
+		if err != nil {
+			return ""
+		}
+
+		location, err := refs.GetLocation(u)
```
justinsb, line 97: Maybe this is the helper I was thinking of.  Not a blocker anyway!

PR #6372, opened 2026-01-30
Support defaultComputeClassConfig in ContainerCluster

### LINKED ISSUE #5900: ContainerCluster: support EnableDefaultComputeClass
A customer has requested that we add support for "EnableDefaultComputeClass" to ContainerCluster

maqiuyujoyce:
### BRIEF Change description

Fixes #5900

Supported `spec.clusterAutoscaling.defaultComputeClassConfig` field in ContainerCluster resource.

While in the issue, the requested feature is "EnableDefaultComputeClass" (a single field), the corresponding feature in the GKE API is represented by DefaultComputeClassConfig struct (https://docs.cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters#defaultcomputeclassconfig). I think we should not do redesign for the GCP API, so I supported this feature based on the GKE API definition.

```release-note
Supported `spec.clusterAutoscaling.defaultComputeClassConfig` field in ContainerCluster resource.
```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_containerclusters.container.cnrm.cloud.google.com.yaml
+++ b/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_containerclusters.container.cnrm.cloud.google.com.yaml
@@ -415,6 +415,8 @@ spec:
                       enabled:
                         description: Enables default compute class.
                         type: boolean
+                    required:
+                    - enabled
                     type: object
                   enabled:
                     description: Whether node auto-provisioning is enabled. Resource


third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -6350,6 +6391,18 @@ func BinaryAuthorizationDiffSuppress(k, old, new string, r *schema.ResourceData)
 	return false
 }
 
+func DefaultComputeClassConfigDiffSuppress(k, old, new string, r *schema.ResourceData) bool {
+	// An empty config is equivalent to a config with enabled set to false.
+	if k == "default_compute_class_config.#" && old == "1" && new == "0" {
+		o, _ := r.GetChange("default_compute_class_config.0.enabled")
```
maqiuyujoyce: CRITICAL: `default_compute_class_config` is a nested field within `cluster_autoscaling`. `r.GetChange` operates on the full resource state and requires the absolute path to the field. Accessing it as `default_compute_class_config.0.enabled` will likely return nil because it ignores the `cluster_autoscaling.0.` prefix.

This will cause `o` to be nil, and the subsequent `o.(bool)` type assertion will panic.

You should construct the correct path, possibly by inspecting `k` (which contains the full path) or by hardcoding `cluster_autoscaling.0.default_compute_class_config.0.enabled` (though `cluster_autoscaling` being a list makes dynamic construction safer).

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -6350,6 +6391,18 @@ func BinaryAuthorizationDiffSuppress(k, old, new string, r *schema.ResourceData)
...
+		if !o.(bool) {
+			return true
```
maqiuyujoyce, line 6399: Related to the path issue: You should strictly check if `o` is nil before type asserting `o.(bool)` to prevent runtime panics.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -3146,7 +3162,7 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
 		req := &container.UpdateClusterRequest{
 			Update: &container.ClusterUpdate{
 				DesiredEnableCiliumClusterwideNetworkPolicy: enabled,
-				ForceSendFields:                             []string{"DesiredEnableCiliumClusterwideNetworkPolicy"},
+				ForceSendFields: []string{"DesiredEnableCiliumClusterwideNetworkPolicy"},
```
maqiuyujoyce, line 3165: NIT: Unrelated whitespace change.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -745,6 +745,22 @@ func ResourceContainerCluster() *schema.Resource {
 							ValidateFunc:     validation.StringInSlice([]string{"BALANCED", "OPTIMIZE_UTILIZATION"}, false),
 							Description:      `Configuration options for the Autoscaling profile feature, which lets you choose whether the cluster autoscaler should optimize for resource utilization or resource availability when deciding to remove nodes from a cluster. Can be BALANCED or OPTIMIZE_UTILIZATION. Defaults to BALANCED.`,
 						},
+						"default_compute_class_config": {
+							Type:             schema.TypeList,
+							Optional:         true,
+							DiffSuppressFunc: DefaultComputeClassConfigDiffSuppress,
```
maqiuyujoyce: The schema definition correctly uses `MaxItems: 1` to model the API's singleton configuration.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -745,6 +745,22 @@ func ResourceContainerCluster() *schema.Resource {
...
+							MaxItems:         1,
+							Description:      "Default compute class is a configuration for default compute class.",
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"enabled": {
+										Type:        schema.TypeBool,
```
maqiuyujoyce, line 757: `Optional: true` implies that if the field is omitted, it takes a zero value (false). Explicitly stating the default behavior in the description would be helpful for users.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -745,6 +745,22 @@ func ResourceContainerCluster() *schema.Resource {
 							ValidateFunc:     validation.StringInSlice([]string{"BALANCED", "OPTIMIZE_UTILIZATION"}, false),
 							Description:      `Configuration options for the Autoscaling profile feature, which lets you choose whether the cluster autoscaler should optimize for resource utilization or resource availability when deciding to remove nodes from a cluster. Can be BALANCED or OPTIMIZE_UTILIZATION. Defaults to BALANCED.`,
 						},
+						"default_compute_class_config": {
+							Type:             schema.TypeList,
```
maqiuyujoyce, line 749: NIT: "Default compute class is a configuration for default compute class." is slightly repetitive. Consider something like "Configuration options for the default compute class."

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_containerclusters.container.cnrm.cloud.google.com.yaml
```diff
@@ -408,6 +408,14 @@ spec:
                       when deciding to remove nodes from a cluster. Can be BALANCED
                       or OPTIMIZE_UTILIZATION. Defaults to BALANCED.
                     type: string
+                  defaultComputeClassConfig:
```
maqiuyujoyce, line 411: The CRD structure correctly mirrors the Terraform schema changes.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -5854,10 +5882,23 @@ func flattenClusterAutoscaling(a *container.ClusterAutoscaling) []map[string]int
 		r["enabled"] = false
 	}
 	r["autoscaling_profile"] = a.AutoscalingProfile
+	if a.DefaultComputeClassConfig != nil {
+		r["default_compute_class_config"] = flattenDefaultComputeClassConfig(a.DefaultComputeClassConfig)
```
maqiuyujoyce, line 5886: The flattening logic is correctly integrated into `flattenClusterAutoscaling`.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -5854,10 +5882,23 @@ func flattenClusterAutoscaling(a *container.ClusterAutoscaling) []map[string]int
...
+	}
 
 	return []map[string]interface{}{r}
 }
 
+func flattenDefaultComputeClassConfig(c *container.DefaultComputeClassConfig) []map[string]interface{} {
```
maqiuyujoyce, line 5892: Correctly handling `c == nil` to return an empty list, which matches Terraform's expectation for missing optional blocks.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -6350,6 +6391,18 @@ func BinaryAuthorizationDiffSuppress(k, old, new string, r *schema.ResourceData)
 	return false
 }
 
+func DefaultComputeClassConfigDiffSuppress(k, old, new string, r *schema.ResourceData) bool {
```
maqiuyujoyce: The diff suppress function logic (ignoring 1->0 transition if the old value was `enabled: false`) seems sound for a toggle field, effectively treating "missing" as "disabled".

pkg/test/resourcefixture/testdata/basic/container/v1beta1/containercluster/containercluster-defaultcomputeclassconfig/create.yaml
```diff
@@ -0,0 +1,24 @@
...
+apiVersion: container.cnrm.cloud.google.com/v1beta1
+kind: ContainerCluster
+metadata:
+  name: containercluster-${uniqueId}
+spec:
+  location: us-central1-a
+  initialNodeCount: 1
+  clusterAutoscaling:
+    defaultComputeClassConfig:
+      enabled: true
```
maqiuyujoyce, line 24: Good coverage: explicitly enabling the feature in `create.yaml`.

pkg/test/resourcefixture/testdata/basic/container/v1beta1/containercluster/containercluster-defaultcomputeclassconfig/update.yaml
```diff
@@ -0,0 +1,24 @@
...
+apiVersion: container.cnrm.cloud.google.com/v1beta1
+kind: ContainerCluster
+metadata:
+  name: containercluster-${uniqueId}
+spec:
+  location: us-central1-a
+  initialNodeCount: 1
+  clusterAutoscaling:
+    defaultComputeClassConfig:
+      enabled: false
```
maqiuyujoyce, line 24: Good coverage: explicitly disabling the feature in `update.yaml`.

pkg/test/resourcefixture/testdata/basic/container/v1beta1/containercluster/containercluster-defaultcomputeclassconfig/_generated_object_containercluster-defaultcomputeclassconfig.golden.yaml
```diff
@@ -0,0 +1,46 @@
...
+  - cnrm.cloud.google.com/deletion-defender
+  generation: 3
+  labels:
+    cnrm-test: "true"
+  name: containercluster-${uniqueId}
+  namespace: ${uniqueId}
+spec:
+  clusterAutoscaling:
+    defaultComputeClassConfig:
+      enabled: false
```
maqiuyujoyce, line 21: The golden file showing `enabled: false` confirms that the flattening logic correctly interprets the API response (even if the API omits the field, it defaults to false).

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -745,6 +745,22 @@ func ResourceContainerCluster() *schema.Resource {
...
+							Type:             schema.TypeList,
+							Optional:         true,
+							DiffSuppressFunc: DefaultComputeClassConfigDiffSuppress,
+							MaxItems:         1,
+							Description:      "Default compute class is a configuration for default compute class.",
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"enabled": {
+										Type:        schema.TypeBool,
+										Optional:    true,
```
maqiuyujoyce: NIT: Description could be more direct, e.g., "Whether the default compute class is enabled."

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -6350,6 +6391,18 @@ func BinaryAuthorizationDiffSuppress(k, old, new string, r *schema.ResourceData)
 	return false
 }
 
+func DefaultComputeClassConfigDiffSuppress(k, old, new string, r *schema.ResourceData) bool {
+	// An empty config is equivalent to a config with enabled set to false.
```
maqiuyujoyce, line 6395: The check `old == "1" && new == "0"` correctly targets the deletion case.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -6350,6 +6391,18 @@ func BinaryAuthorizationDiffSuppress(k, old, new string, r *schema.ResourceData)
 	return false
 }
```
maqiuyujoyce, line 6393: The key `k` passed to `DiffSuppressFunc` includes the full path (e.g., `cluster_autoscaling.0.default_compute_class_config.#`). Checking for exact equality `== "default_compute_class_config.#"` will fail.
Use `strings.HasSuffix(k, "default_compute_class_config.#")` to correctly identify the field.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: CRITICAL: `r.GetChange` expects the full path to the attribute, but `default_compute_class_config.0.enabled` is a relative path.
This call will likely return `nil` (or not find the value), leading to a panic in the type assertion below.
You must construct the full path using the prefix derived from `k`.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce, line 6395: CRITICAL: Unsafe type assertion `o.(bool)`. If `r.GetChange` returns `nil` (which is possible if the path is wrong or the field is missing), this will panic.
Always check if the returned value is non-nil before asserting its type.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -6350,6 +6391,18 @@ func BinaryAuthorizationDiffSuppress(k, old, new string, r *schema.ResourceData)
 	return false
```
maqiuyujoyce, line 6391: Style: Helper functions like this should generally be unexported (start with a lowercase letter) unless they are intended to be used by other resources.
Suggestion: Rename to `defaultComputeClassConfigDiffSuppress`.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -4560,10 +4576,22 @@ func expandClusterAutoscaling(configured interface{}, d *schema.ResourceData) *c
 		EnableNodeAutoprovisioning:       config["enabled"].(bool),
 		ResourceLimits:                   resourceLimits,
 		AutoscalingProfile:               config["autoscaling_profile"].(string),
+		DefaultComputeClassConfig:        expandDefaultComputeClassConfig(config["default_compute_class_config"], d),
 		AutoprovisioningNodePoolDefaults: expandAutoProvisioningDefaults(config["auto_provisioning_defaults"], d),
 	}
 }
 
+func expandDefaultComputeClassConfig(configured interface{}, d *schema.ResourceData) *container.DefaultComputeClassConfig {
+	l := configured.([]interface{})
```
maqiuyujoyce, line 4585: Unsafe type assertion. `configured` is an `interface{}`, and while `d.Get` usually returns the correct type, it is safer to use the comma-ok idiom:
`l, ok := configured.([]interface{})`
This prevents a panic if `configured` is somehow nil or the wrong type.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -4560,10 +4576,22 @@ func expandClusterAutoscaling(configured interface{}, d *schema.ResourceData) *c
 		EnableNodeAutoprovisioning:       config["enabled"].(bool),
 		ResourceLimits:                   resourceLimits,
 		AutoscalingProfile:               config["autoscaling_profile"].(string),
+		DefaultComputeClassConfig:        expandDefaultComputeClassConfig(config["default_compute_class_config"], d),
```
maqiuyujoyce, line 4579: The `d` parameter (schema.ResourceData) is passed but not used in `expandDefaultComputeClassConfig`.
You can remove it from the function signature if it's not needed.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -2166,17 +2182,17 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 		ClusterTelemetry: expandClusterTelemetry(d.Get("cluster_telemetry")),
 		EnableTpu:        d.Get("enable_tpu").(bool),
 		NetworkConfig: &container.NetworkConfig{
```
maqiuyujoyce, line 2184: Maintenance: This block contains unrelated formatting changes (re-alignment of struct fields).
Please revert these lines to match the upstream file state. This minimizes the diff and makes future rebases easier.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce, line 3165: Maintenance: Unrelated formatting change. Please revert.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -745,6 +745,22 @@ func ResourceContainerCluster() *schema.Resource {
 							ValidateFunc:     validation.StringInSlice([]string{"BALANCED", "OPTIMIZE_UTILIZATION"}, false),
 							Description:      `Configuration options for the Autoscaling profile feature, which lets you choose whether the cluster autoscaler should optimize for resource utilization or resource availability when deciding to remove nodes from a cluster. Can be BALANCED or OPTIMIZE_UTILIZATION. Defaults to BALANCED.`,
 						},
+						"default_compute_class_config": {
+							Type:             schema.TypeList,
+							Optional:         true,
+							DiffSuppressFunc: DefaultComputeClassConfigDiffSuppress,
+							MaxItems:         1,
+							Description:      "Default compute class is a configuration for default compute class.",
```
maqiuyujoyce, line 753: Doc: "Default compute class is a configuration for default compute class." is tautological.
Suggestion: "Configuration for the default compute class."

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -745,6 +745,22 @@ func ResourceContainerCluster() *schema.Resource {
...
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"enabled": {
+										Type:        schema.TypeBool,
+										Optional:    true,
+										Description: "Enables default compute class.",
```
maqiuyujoyce, line 759: Doc: Suggestion: "Whether the default compute class is enabled."

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_containerclusters.container.cnrm.cloud.google.com.yaml
```diff
@@ -408,6 +408,14 @@ spec:
...
+                    description: Default compute class is a configuration for default
```
maqiuyujoyce, line 412: Please update the description to match the suggested Go code description:
"Configuration for the default compute class."

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_containerclusters.container.cnrm.cloud.google.com.yaml
```diff
@@ -408,6 +408,14 @@ spec:
...
+                      compute class.
+                    properties:
+                      enabled:
+                        description: Enables default compute class.
```
maqiuyujoyce, line 416: Suggestion: "Whether the default compute class is enabled."

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -6350,6 +6391,18 @@ func BinaryAuthorizationDiffSuppress(k, old, new string, r *schema.ResourceData)
 	return false
 }
 
+func DefaultComputeClassConfigDiffSuppress(k, old, new string, r *schema.ResourceData) bool {
+	// An empty config is equivalent to a config with enabled set to false.
+	if k == "default_compute_class_config.#" && old == "1" && new == "0" {
+		o, _ := r.GetChange("default_compute_class_config.0.enabled")
+		if !o.(bool) {
+			return true
+		}
```
maqiuyujoyce, line 6400: The `DiffSuppressFunc` is intended to handle the case where the field is missing in config but present (as disabled) in state.
The current tests (`update.yaml`) explicitly set `enabled: false`.
Please add a test case where the `defaultComputeClassConfig` block is entirely removed from the YAML to verify that the suppress function works as intended (i.e., no diff is generated).

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -5854,10 +5882,23 @@ func flattenClusterAutoscaling(a *container.ClusterAutoscaling) []map[string]int
...
+	if a.DefaultComputeClassConfig != nil {
+		r["default_compute_class_config"] = flattenDefaultComputeClassConfig(a.DefaultComputeClassConfig)
+	}
 
 	return []map[string]interface{}{r}
 }
 
+func flattenDefaultComputeClassConfig(c *container.DefaultComputeClassConfig) []map[string]interface{} {
+	result := []map[string]interface{}{}
+	if c != nil {
```
maqiuyujoyce, line 5894: Nit: Using `make(map[string]interface{})` or `make([]map[string]interface{}, 0, 1)` is slightly more idiomatic than `[]map[string]interface{}{}`, but this is minor.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -4560,10 +4576,22 @@ func expandClusterAutoscaling(configured interface{}, d *schema.ResourceData) *c
...
 		AutoprovisioningNodePoolDefaults: expandAutoProvisioningDefaults(config["auto_provisioning_defaults"], d),
 	}
 }
 
+func expandDefaultComputeClassConfig(configured interface{}, d *schema.ResourceData) *container.DefaultComputeClassConfig {
+	l := configured.([]interface{})
+	if len(l) == 0 || l[0] == nil {
+		return nil
+	}
+	config := l[0].(map[string]interface{})
```
maqiuyujoyce, line 4589: Safety: Ensure `config["enabled"]` exists and is a bool. `d.Get` populates defaults, so it should be fine, but explicit checks or safe casting `v, ok := config["enabled"].(bool)` are preferred in robust provider code.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce, line 5886: The nil check for `a.DefaultComputeClassConfig` is correct. If the API returns nil, this field will be omitted from the map, resulting in `nil` in the TF state, which matches the empty config case.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce, line 757: Since `enabled` is `Optional` and has no `Default` specified in schema, Terraform will default it to the zero value (false) if missing. Ensure this aligns with the API behavior (which seems to default to false based on the golden files).

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: `MaxItems: 1` correctly enforces the singleton nature of this configuration block.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce, line 6391: The suppress function is added but I don't see any unit tests for the suppress logic itself.
Consider adding a unit test for `DefaultComputeClassConfigDiffSuppress` in `resource_container_cluster_test.go` if possible, covering the path resolution and logic.

maqiuyujoyce (commented):
*This is the first round review comments from repo AGENT. I'll share my detailed thoughts in the replies.*

Overall, the logic for adding the new field is sound. However, the `DiffSuppressFunc` and expansion functions need to be corrected to prevent potential panics and ensure correctness. Please also address the minor nits regarding descriptions and revert any unrelated formatting changes to keep the codebase maintainable.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: I think it is a good concern, though it is not concerning based on how Terraform lib is implemented and used.

Before this function is called, many steps have been conducted to ensure the passed in value is no empty and matches the schema, so as long as the code logic here matches the schema, it should not panic.

Besides, this is a well-adopted pattern in the Terraform library.

I'll keep the code as is.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Same answer as above.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: This is the go formatting fix that the previous code change missed.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Ack.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: We want to reduce manual work and inconsistency so the description is a copy of what defined in https://docs.cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters#Cluster.ClusterAutoscaling.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Good point about the description, though as we want to reduce manual work and inconsistency so the description is a copy of what defined in https://docs.cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters#Cluster.ClusterAutoscaling.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_containerclusters.container.cnrm.cloud.google.com.yaml
maqiuyujoyce: Ack.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Ack.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Ack.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Ack.

pkg/test/resourcefixture/testdata/basic/container/v1beta1/containercluster/containercluster-defaultcomputeclassconfig/create.yaml
maqiuyujoyce: Ack.

pkg/test/resourcefixture/testdata/basic/container/v1beta1/containercluster/containercluster-defaultcomputeclassconfig/update.yaml
maqiuyujoyce: Ack.

pkg/test/resourcefixture/testdata/basic/container/v1beta1/containercluster/containercluster-defaultcomputeclassconfig/_generated_object_containercluster-defaultcomputeclassconfig.golden.yaml
maqiuyujoyce: Ack.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Agreed. Though we want to reduce manual work and inconsistency so the description is a copy of what defined in https://docs.cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters#Cluster.ClusterAutoscaling.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Ack.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Ack.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Verified this behavior is aligned.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Ack.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Good point! Will fix.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Will fix.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Good point. Though it is an established pattern and common practice to assume the value is properly populated and typed so I'll keep the code as is for now.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: This is the standard go formatting of the file.

I'll keep it as is.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Will update.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Removing a field in a TF-based KCC resource is considered unmanaging the resource, so this field cannot be unset. Adding one more test is nice to have but not blocking.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_containerclusters.container.cnrm.cloud.google.com.yaml
maqiuyujoyce: Good point about the description, though as we want to reduce manual work and inconsistency so the description is a copy of what defined in https://docs.cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters#Cluster.ClusterAutoscaling.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_containerclusters.container.cnrm.cloud.google.com.yaml
maqiuyujoyce: Good point about the description, though as we want to reduce manual work and inconsistency so the description is a copy of what defined in https://docs.cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters#Cluster.ClusterAutoscaling.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Good point about the description, though as we want to reduce manual work and inconsistency so the description is a copy of what defined in https://docs.cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters#Cluster.ClusterAutoscaling.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Good point about the description, though as we want to reduce manual work and inconsistency so the description is a copy of what defined in https://docs.cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters#Cluster.ClusterAutoscaling.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: This happened as a result of doing go formatting of this file.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: This is a valid concern. Regarding this code in particular, direct type assertion is a common pattern in TF lib and the types should be verified before passed to this function. Also the function design in general doesn't provide a channel to raise the type assertion errors. So I'll just leave it as is.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: In TF lib, it's a common pattern for expansion functions to include d *schema.ResourceData in their signature, and passing it to the function even if not immediately used. This ensures a consistent interface across related functions and provides easy access to other resource data if the logic needs to expand in the future (e.g., for cross-field validation).

It is generally more aligned with the existing design pattern and safer for future development to keep `d` in the function signature of expandDefaultComputeClassConfig.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Ack.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Good point though can be done later when needed.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Good point! Though I think this is an existing pattern in TF lib and I'll keep as is for now.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Done.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Done.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
maqiuyujoyce: Done.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
gemmahou, line 1: General question: ContainerCluster is still TF based, why don't we patch the existing [TF code](https://github.com/gemmahou/terraform-provider-google-beta/blob/main/google-beta/services/container/resource_container_cluster.go), specifically, the field is "default_compute_class_enabled"(not default_compute_class_config)

maqiuyujoyce: I'm preparing our resource for the future direct migration, i.e. having generated API based on googleapis proto for all the newly patched fields (https://github.com/googleapis/googleapis/blob/master/google/container/v1beta1/cluster_service.proto#L5879). And the TF provider is not really aligning with the proto design.

pkg/test/resourcefixture/testdata/basic/container/v1beta1/containercluster/containercluster-defaultcomputeclassconfig/_http.log
```diff
@@ -503,10 +491,10 @@ X-Xss-Protection: 0
 
 ---
 
-GET https://compute.googleapis.com/compute/v1/projects/${projectId}/zones/us-central1-a/instanceGroupManagers/gke-containercluster-azk-default-pool-ab7b082a-grp?alt=json&prettyPrint=false
+GET https://compute.googleapis.com/compute/v1/projects/${projectId}/zones/us-central1-a/instanceGroupManagers/gke-containercluster-abcdef-default-pool-grp?alt=json&prettyPrint=false
 User-Agent: google-api-go-client/0.5 kcc/${kccVersion} (+https://github.com/GoogleCloudPlatform/k8s-config-connector) kcc/controller-manager/${kccVersion}
 
-200 OK
+404 Not Found
```
gemmahou, line 497: no action needed for now: I guess this is because our mock doesn't automatically create a default nodepool during cluster creation.

maqiuyujoyce: Actually mock container cluster does create a default nodepool. It's probably that this nodepool doesn't create  corresponding Compute InstanceGroupManagers in the backend.

PR #6409, opened 2026-02-05
Migrate the preview cmd to config connector cli

anhdle-sso:
# Migrate the preview cmd to config connector cli
## Description
This PR migrates the `preview` command into the main Config Connector CLI (`pkg/cli`). It registers `preview` as a subcommand of `config-connector`, allowing users to invoke the preview functionality (e.g., `config-connector preview`) directly.
## Changes
*   **pkg/cli/cmd/preview.go**: Added registration for the `preview` subcommand.
*   **pkg/cli/cmd/preview/execute.go**: Implemented the execution logic for the preview command, including setting up the preview instance and handling results.
*   **pkg/cli/cmd/preview/parameters.go**: Defined the flags and parameters for the preview command.
*   **pkg/cli/cmd/root.go**: Added `preview` to the list of root commands.
*   **experiments/kompanion/main.go**: Updated to reflect the CLI refactoring.
## Verification
*   Verified that `config-connector preview` can be invoked.

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


pkg/cli/cmd/preview.go
```diff
@@ -0,0 +1,62 @@
...
+
+const (
+	previewExamples = `
+	# preview Config Connector resources
+	config-connector preview
+	`
+)
+
+var (
+	previewParams = parameters.Parameters{}
```
gemmahou: AGENT: `previewParams` is defined as a package-level variable. While this is common in some Cobra implementations, it prevents the command from being re-entrant and can cause side effects in tests if the command is executed multiple times. Consider using a struct to hold the command state or defining this variable locally within the command creation logic.

pkg/cli/cmd/preview.go
```diff
@@ -0,0 +1,62 @@
...
+		Hidden:  true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return preview.Execute(cmd.Context(), &previewParams)
+		},
+		Args: cobra.ExactArgs(0),
+	}
+)
+
+func init() {
+	// TODO: Add scope
```
gemmahou: AGENT: This `TODO: Add scope` in `init()` seems misplaced or redundant, as scope handling appears to be inside `execute.go`. If this refers to a missing flag, it should be clarified.

pkg/cli/cmd/preview.go
```diff
@@ -0,0 +1,62 @@
...
+	previewCmd.Flags().StringVarP(&previewParams.Kubeconfig, previewKubeconfigFlag, "", previewParams.Kubeconfig, "path to the kubeconfig file.")
+	previewCmd.Flags().IntVarP(&previewParams.Timeout, previewTimeoutFlag, "", 15, "timeout in minutes. Default to 15 minutes.")
+	previewCmd.Flags().StringVarP(&previewParams.ReportNamePrefix, previewReportNamePrefixFlag, "", "preview-report", "Prefix for the report name. The tool appends a timestamp to this in the format \"YYYYMMDD-HHMMSS.milliseconds\".")
+	previewCmd.Flags().BoolVarP(&previewParams.FullReport, "full-report", "f", false, "Write a full report with all the captured objects. Disabled by default.")
+	previewCmd.Flags().Float64VarP(&previewParams.GCPQPS, "gcpQPS", "q", 5.0, "Maximum qps for GCP API requests, per service. Default to 5.0. Set gcpQPS to 0 to disable rate limiting.")
```
gemmahou: AGENT: The flag name `gcpQPS` uses lowerCamelCase. Standard CLI convention and other flags in this file (e.g., `report-prefix`) use kebab-case. Consider renaming to `gcp-qps`.

pkg/cli/cmd/preview.go
```diff
@@ -0,0 +1,62 @@
...
+	previewCmd.Flags().IntVarP(&previewParams.GCPBurst, "gcpBurst", "b", 5, "Maximum burst for GCP API requests, per service. Default to 5. Set gcpQPS to 0 to disable rate limiting.")
```
gemmahou: AGENT: Similar to `gcpQPS`, the flag `gcpBurst` should likely be renamed to `gcp-burst` to maintain consistency with standard CLI conventions.

pkg/cli/cmd/preview/execute.go
```diff
@@ -0,0 +1,144 @@
...
+		UpstreamGCPAuthorization: authorization,
+		UpstreamGCPHTTPClient:    nil,
+		UpstreamGCPQPS:           opts.GCPQPS,
+		UpstreamGCPBurst:         opts.GCPBurst,
+		Namespace:                opts.Namespace,
+	})
+	if err != nil {
+		return fmt.Errorf("building preview instance: %w", err)
+	}
+	ctx, cancel := context.WithTimeout(ctx, time.Duration(opts.Timeout)*time.Minute)
```
gemmahou, line 81: AGENT: `context.WithTimeout` is called with `opts.Timeout`. If the user provides `0` (or if it's the default and not set), the timeout will be 0, causing the context to expire immediately. Consider validating that `opts.Timeout` is positive or handling the 0 case (e.g., as "no timeout").

There's a default timeout of 15, so this won't be an issue.

pkg/cli/cmd/preview/execute.go
```diff
@@ -0,0 +1,144 @@
...
+	} else {
+		loadingRules = clientcmd.NewDefaultClientConfigLoadingRules()
+	}
+	return clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		loadingRules,
+		&clientcmd.ConfigOverrides{}).ClientConfig()
+}
+
+func getGCPAuthorization(ctx context.Context, opts *parameters.Parameters) (oauth2.TokenSource, error) {
+	// TODO: Add scope
```
gemmahou: `// TODO: Add scope` - Is the hardcoded `defaultScope` sufficient for all preview use cases? If not, this should be prioritized or the limitation documented.

pkg/cli/cmd/preview/execute.go
```diff
@@ -0,0 +1,144 @@
...
+		return nil, err
+	}
+	return ts, nil
+}
+
+func printCapturedObjects(recorder *preview.Recorder, opts *parameters.Parameters) error {
+	now := time.Now()
+	timestamp := now.Format("20060102-150405.000")
+	summaryFile := fmt.Sprintf("%s-%s", opts.ReportNamePrefix, timestamp)
+	if err := recorder.SummaryReport(summaryFile); err != nil {
```
gemmahou, line 136: AGENT: `recorder.SummaryReport(summaryFile)` likely attempts to create a file at `summaryFile`. If `opts.ReportNamePrefix` contains directory components (e.g., `reports/preview`), this might fail if the directory does not exist. Consider ensuring the output directory exists before writing.

+1

pkg/cli/cmd/preview/execute.go
```diff
@@ -0,0 +1,144 @@
...
+)
+
+const (
+	defaultScope = "https://www.googleapis.com/auth/cloud-platform"
+)
+
+func Execute(ctx context.Context, opts *parameters.Parameters) error {
+	// Use a custom FlagSet for klog to avoid conflicts with global flag.CommandLine
+	klogFlags := flag.NewFlagSet("klog", flag.ExitOnError)
+	klog.InitFlags(klogFlags)
```
gemmahou, line 42: AGENT: `klog.InitFlags(klogFlags)` modifies the global `klog` state. If this function is called multiple times (e.g., in unit tests) or in parallel with other components using klog, it might panic or cause unexpected behavior. Consider if there's a way to configure logging locally or ensure this is safe.

I won't worry too much about the logger setup at this time... feel free to investigate this and address in a followup PR.

pkg/cli/cmd/preview/execute.go
```diff
@@ -0,0 +1,144 @@
...
+	if err := klogFlags.Set("v", fmt.Sprintf("%d", opts.Verbose)); err != nil {
+		fmt.Printf("Failed to set -v flag: %v\n", err)
+	}
+	if err := klogFlags.Set("alsologtostderr", "true"); err != nil {
+		fmt.Printf("Failed to set -alsologtostderr flag: %v\n", err)
+	}
+	log.SetLogger(klogr.New())
```
gemmahou, line 49: AGENT: log.SetLogger(klogr.New())` sets the global logger for the controller-runtime `log` package. This is a global side effect that persists after the command finishes. This could affect other parts of the application if they rely on a different logger configuration.

pkg/cli/cmd/preview/execute.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+	case <-ctx.Done():
+		return fmt.Errorf("timeout reached: %w", ctx.Err())
+	}
+	klog.V(0).Info("Preview finished successfully")
+	return nil
+}
+
+func getRESTConfig(ctx context.Context, opts *parameters.Parameters) (*rest.Config, error) {
+	// TODO: Add rate limiting.
```
gemmahou: AGENT: `// TODO: Add rate limiting.` suggests missing functionality. Please ensure that the default client configuration is safe enough for general use, or if this TODO needs to be addressed before merging.

Feel free to address in a followup PR

pkg/cli/cmd/preview/execute.go
gemmahou, line 123: AGENT: `// TODO: Add scope` - Is the hardcoded `defaultScope` sufficient for all preview use cases? If not, this should be prioritized or the limitation documented.

Feel free to address this in a followup PR

pkg/cli/cmd/preview.go
anhdle-sso: done

pkg/cli/cmd/preview.go
anhdle-sso: Done

pkg/cli/cmd/preview.go
anhdle-sso: done

pkg/cli/cmd/preview/execute.go
anhdle-sso: For now, we only create file with current folder. If users want a non-existing folder, we already print out the error. Will adjust if user's feedback differently.

pkg/cli/cmd/preview/execute.go
anhdle-sso: the CLI is only run as binary or in an isolated job. The global setting is intentional as we do not want logger down the line to print too much log.

pkg/cli/cmd/preview/execute.go
anhdle-sso: ack

pkg/cli/cmd/preview.go
anhdle-sso: done

pkg/cli/cmd/preview/execute.go
```diff
@@ -0,0 +1,144 @@
...
+package preview
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"time"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/cli/cmd/preview/parameters"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/cli/preview"
```
cheftako: AGENT: The file is in package `preview`, and it imports `github.com/.../pkg/cli/preview`. While valid in Go, having the import name match the package name can be confusing. Consider aliasing the import (e.g., `import corepreview "..."`) to make it clear which `preview` is being referenced.

TAKO: +1

pkg/cli/cmd/preview/execute.go
```diff
@@ -0,0 +1,144 @@
...
+const (
+	defaultScope = "https://www.googleapis.com/auth/cloud-platform"
+)
+
+func Execute(ctx context.Context, opts *parameters.Parameters) error {
+	// Use a custom FlagSet for klog to avoid conflicts with global flag.CommandLine
+	klogFlags := flag.NewFlagSet("klog", flag.ExitOnError)
+	klog.InitFlags(klogFlags)
+	if err := klogFlags.Set("v", fmt.Sprintf("%d", opts.Verbose)); err != nil {
+		fmt.Printf("Failed to set -v flag: %v\n", err)
```
cheftako, line 44: AGENT: Consider writing to `os.Stderr` or using a logger for this error message. Writing to `stdout` (`fmt.Printf`) might interfere if the user intends to pipe the command output, although for this specific error it might be fine.

TAKO: +1. Not blocking

pkg/cli/cmd/preview/execute.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	if err := klogFlags.Set("alsologtostderr", "true"); err != nil {
+		fmt.Printf("Failed to set -alsologtostderr flag: %v\n", err)
+	}
+	log.SetLogger(klogr.New())
+	defer klog.Flush()
+
```
cheftako, line 51: AGENT: This log line uses `klog.V(0)`, which is the default verbosity. This means it will always print "Starting preview tool." when the command runs. CLI tools typically remain silent on success or use higher verbosity for status messages unless this is intended as a permanent banner.

TAKO: Given how long the tool can run for, some signal that it start is nice. Eventually it would be nice to have a swirly/throbber to indicate that something is happening. Having said that it is important to not generate TB logs. Not blocking.

pkg/cli/cmd/preview/execute.go
```diff
@@ -0,0 +1,144 @@
...
+	klog.V(0).Info("Starting preview tool.")
+	upstreamRESTConfig, err := getRESTConfig(ctx, opts)
+	if err != nil {
+		return fmt.Errorf("error building kubeconfig: %w", err)
+	}
+	recorder := preview.NewRecorder()
+	if err := recorder.PreloadGKNN(ctx, upstreamRESTConfig, opts.Namespace); err != nil {
+		return fmt.Errorf("error preload the list of resources to reconcile: %w", err)
```
cheftako: AGENT: Nit: Grammar "error preload" -> "error preloading".

TAKO: Not blocking.

pkg/cli/cmd/preview/execute.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	authorization, err := getGCPAuthorization(ctx, opts)
+	if err != nil {
+		return fmt.Errorf("error building GCP authorization: %w", err)
+	}
+	previewInstance, err := preview.NewPreviewInstance(recorder, preview.PreviewInstanceOptions{
+		UpstreamRESTConfig:       upstreamRESTConfig,
+		UpstreamGCPAuthorization: authorization,
+		UpstreamGCPHTTPClient:    nil,
```
cheftako, line 69: AGENT: Explicitly passing `nil` for `UpstreamGCPHTTPClient`. Please ensure `NewPreviewInstance` handles a nil client gracefully (e.g., by initializing a default one), or if this is intended to rely on `UpstreamGCPAuthorization` to create the client.

TAKO: +1

pkg/cli/cmd/preview/execute.go
cheftako, line 81: AGENT: If `opts.Timeout` is set to 0 (which might be expected to mean "no timeout"), `context.WithTimeout` will create a context that expires immediately. Consider handling 0 as a special case or ensuring a minimum timeout.

TAKO: +1. I don't think there is value to an immediate timeout. Interesting question if there is value to suspending the timeout. I'm assuming it will eventually error out on things like a broken connection.

pkg/cli/cmd/preview/execute.go
```diff
@@ -0,0 +1,144 @@
...
+	defer func() {
+		if err := printCapturedObjects(recorder, opts); err != nil {
+			klog.Error(err, "error printing captured objects")
+		}
+	}()
+
+	errChan := make(chan error, 1)
+	go func() {
+		klog.V(0).Info("Starting preview")
+		errChan <- previewInstance.Start(ctx)
```
cheftako, line 91: AGENT: This log message "Starting preview" seems redundant given the earlier message "Starting preview tool." at line 50.

TAKO: +1. Printing once may be ok. Twice seems unprofessional.

pkg/cli/cmd/preview/execute.go
```diff
@@ -0,0 +1,144 @@
...
+	case err := <-errChan:
+		if err != nil {
+			return fmt.Errorf("error starting preview: %w", err)
+		}
+	case <-ctx.Done():
+		return fmt.Errorf("timeout reached: %w", ctx.Err())
+	}
+	klog.V(0).Info("Preview finished successfully")
+	return nil
+}
```
cheftako, line 104: AGENT: `TODO: Add rate limiting.` - Is this something that should be addressed before merging, or is it tracked elsewhere?

TAKO: I see you resolved this without comment below. Do we have rate limiting at the moment? If the flags above don't currently do anything it would be good to comment them out till they do something.

pkg/cli/cmd/preview/execute.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return ts, nil
+}
+
+func printCapturedObjects(recorder *preview.Recorder, opts *parameters.Parameters) error {
+	now := time.Now()
+	timestamp := now.Format("20060102-150405.000")
+	summaryFile := fmt.Sprintf("%s-%s", opts.ReportNamePrefix, timestamp)
```
cheftako, line 135: AGENT: Consider defining the timestamp format `20060102-150405.000` as a constant, as it might be useful to ensure consistency if used elsewhere.

TAKO: Meh

pkg/cli/cmd/preview.go
```diff
@@ -0,0 +1,61 @@
...
+		},
+		Args: cobra.ExactArgs(0),
+	}
+
+	cmd.Flags().StringVarP(&params.Kubeconfig, previewKubeconfigFlag, "", params.Kubeconfig, "path to the kubeconfig file.")
+	cmd.Flags().IntVarP(&params.Timeout, previewTimeoutFlag, "", 15, "timeout in minutes. Default to 15 minutes.")
+	cmd.Flags().StringVarP(&params.ReportNamePrefix, previewReportNamePrefixFlag, "", "preview-report", "Prefix for the report name. The tool appends a timestamp to this in the format \"YYYYMMDD-HHMMSS.milliseconds\".")
+	cmd.Flags().BoolVarP(&params.FullReport, "full-report", "f", false, "Write a full report with all the captured objects. Disabled by default.")
+	cmd.Flags().Float64VarP(&params.GCPQPS, "gcp-qps", "q", 5.0, "Maximum qps for GCP API requests, per service. Default to 5.0. Set gcp-qps to 0 to disable rate limiting.")
+	cmd.Flags().IntVarP(&params.GCPBurst, "gcp-burst", "b", 5, "Maximum burst for GCP API requests, per service. Default to 5. Set gcp-qps to 0 to disable rate limiting.")
```
cheftako: AGENT: The help text "Set gcp-qps to 0 to disable rate limiting." is helpful, but since this is the `gcp-burst` flag, it might be slightly confusing to reference the other flag without context. Maybe rephrase to "Burst size... (Note: set gcp-qps to 0 to disable limiting)".

TAKO: Assuming it should say ```Set gcp-burst to 0 to disable rate limiting.``` Either way not blocking.

cheftako (commented):
AGENT: Overall, the implementation is clean. The move to a `NewPreviewCmd` factory function is a good improvement for re-entrancy.
I have a few comments regarding the `preview` package import within the `preview` command package, some logging adjustments, and minor nits.

TAKO: Beware declaring interfaces you are not yet using. Setting up QPS flags locks you into a interface. This is problematic. 1) It may not match what you end up wanting to do. 2) It sets an expectation from customers that you are not (yet) meeting.

pkg/cli/cmd/preview/execute.go
anhdle-sso: I added the rate limiter to GCP client. The TODO here is for the kube client. I adjusted the TODO comment to be more specific.

pkg/cli/cmd/preview/execute.go
anhdle-sso: ack

pkg/cli/cmd/preview/execute.go
anhdle-sso: NewPreviewInstance has a check if the UpstreamGCPHTTPClient is nil. In that case, the gcp client is initialized with our custom GCP client.

pkg/cli/cmd/preview/execute.go
anhdle-sso: 0 value does mean immediate cancel. The CLI option default the value to 15. I added the logic if 0 is set, we default back to 15. I added a TODO to potentially make 0 mean running indefinitely.

pkg/cli/cmd/preview/execute.go
anhdle-sso: done

pkg/cli/cmd/preview/execute.go
anhdle-sso: It is a good suggestion. Create issue 6433 to track this.

pkg/cli/cmd/preview/execute.go
anhdle-sso: done

pkg/cli/cmd/preview.go
anhdle-sso: Setting `gcp-qps` to 0 is all it takes to disable rate limtting. It is somewhat confusing for users. I will address this later.

pkg/cli/cmd/preview/execute.go
anhdle-sso: This is a halucinated comment.

pkg/cli/cmd/preview/parameters/parameters.go
```diff
@@ -0,0 +1,27 @@
+// Copyright 2026 Google LLC
```
cheftako: These options/parameters are normally put in an options.go file (eg https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/experiments/kompanion/cmd/summary/options.go or https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-apiserver/app/options/options.go) Please switch.

cheftako:
Mostly good. Lets switch to the standard struct name and file name and good to go.

Issue #2756, opened 2024-09-20
RunJob: allow volumes to be created using Cloud Storage FUSE


roccolangeweg:
### Checklist

- [X] I did not find a related open enhancement request.
- [X] I understand that enhancement requests filed in the GitHub repository are by default low priority.
- [X] If this request is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Describe the feature or resource

Add support for defining RunJob to define volumes using Google Cloud Storage buckets. This is already supported in both the UI and the gcloud CLI.

### Additional information

API Docs: https://cloud.google.com/run/docs/configuring/services/cloud-storage-volume-mounts#yaml

Through YAML
```
 volumes:
      - name: VOLUME_NAME
        csi:
          driver: gcsfuse.run.googleapis.com
          readOnly: IS_READ_ONLY
          volumeAttributes:
            bucketName: BUCKET_NAME
```

Through gcloud
```
gcloud run services update SERVICE \
--add-volume name=VOLUME_NAME,type=cloud-storage,bucket=BUCKET_NAME \
--add-volume-mount volume=VOLUME_NAME,mount-path=MOUNT_PATH
```

### Importance

We use Flux to automatically deploy new versions of Cloud Run Services/Jobs.

One of our Jobs is mounting a GCS Bucket to read and write files too, we now have to exempt this container from automatic deployment and manually update the container configuration, causing drift because all other containers are able to run the latest version.

katrielt:
Supported as of v1.145.0 - https://github.com/GoogleCloudPlatform/k8s-config-connector/releases/tag/v1.145.0

PR #6417, opened 2026-02-05
feat: Runjob - Support NFS backed Volumes

katrielt:
Add support for NFS backed Volumes in Cloud Run Jobs

Addresses https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/2756 and internal issue b/339911170

### DIFF
--- a/pkg/controller/direct/run/mapper.go
+++ b/pkg/controller/direct/run/mapper.go
@@ -1,3 +1,17 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package run
 
 import (


apis/run/v1beta1/job_types.go
```diff
@@ -597,7 +597,7 @@ type Volume struct {
 
 	// For NFS Voumes, contains the path to the nfs Volume
 	// +kcc:proto:field=google.cloud.run.v2.Volume.nfs
-	// Nfs *NfsVolumeSource `json:"nfs,omitempty"`
+	NFS *NFSVolumeSource `json:"nfs,omitempty"`
```
yuwenma: Could you change "NFS" to "Nfs"? Then I think the mapper generator should be able to update the mapper.generated.go.

katrielt: It was auto-generated as Nfs, but I believe it should be NFS (Like GCS, GKE, etc). The list of fields under the type (NFSVolumeSource) are very likely not going to change, so I don't see this as a problem.

If you still think I should rename, LMK.

yuwenma: The way mapper generator works is that it requires the go field name (with capital letter in upper case) to match the proto name (nfs). So if you can change the go field to "Nfs", I believe the right function can be added to mapper.automated.go. 

The reason we prefer mapper generator is to avoid the review on manual mappers (which could lack corner case handlings)

katrielt: Ack. I'll move it back to the generated mapper and rename the reset to Nfs.

katrielt: Done. Both real and mock E2E tests showed no changes.

PR #6478, opened 2026-02-11
Add ReportReconcileStart/End to iamauditconfig controller

anhdle-sso:
### BRIEF Change description

Fixes #

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/controller/iam/auditconfig/iamauditconfig_controller.go
+++ b/pkg/controller/iam/auditconfig/iamauditconfig_controller.go
@@ -161,10 +161,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 	// after r.Get().
 	auditConfig.SetGroupVersionKind(iamv1beta1.IAMAuditConfigGVK)
 	uObj := &unstructured.Unstructured{}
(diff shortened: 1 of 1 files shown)


pkg/controller/iam/auditconfig/iamauditconfig_controller.go
```diff
@@ -159,6 +160,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 	// r.Get() overrides the TypeMeta to empty value, so need to configure it
 	// after r.Get().
 	auditConfig.SetGroupVersionKind(iamv1beta1.IAMAuditConfigGVK)
+	uObj := &unstructured.Unstructured{}
+	uObj.Object, err = runtime.DefaultUnstructuredConverter.ToUnstructured(&auditConfig)
+	if err != nil {
+		return reconcile.Result{}, err
```
cheftako: A little concerned this could be a breaking change. This means if we have an error we don't reconcile where we used to. What about doing everything through the defer inside of a if err == nil ?

anhdle-sso: we don't actually need the full object to restart start/stop. I removed the conversion to unstructured and error exit early. PTAL @cheftako

Issue #5678, opened 2025-11-19
Add resource CertificateManagerCertificateIssuanceConfig


yuwenma:

fkc1e100:
@katrielt Since the resolving PR #6506 has been merged, could you please close this issue?

katrielt:
@fkc1e100 PR #6506 was for gcloud CRUD tests, it's still missing a controller.

fkc1e100:
Hello! For anyone tracking this feature:
The `CertificateManagerCertificateIssuanceConfig` resource is actually **available** in KCC, but it is currently in the **Alpha (`v1alpha1`)** release stage. The merged PRs (#6506, #7086, #6881) have established `mockgcp` test coverage for it.
This issue remains open to track its promotion to **Beta/Stable (`v1beta1`)**.

anfernee:
Since the resource is already added. I'll suggest close this one and have a different issue to track the beta promotion. WDYT @fkc1e100 ?

anfernee:
fixed by https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/9639

PR #6506, opened 2026-02-15
Generate the mockgcp server for CertificateManager CertificateIssuanceConfig

katrielt:
### BRIEF Change description                                                                                                                                                                                                           
Generate the mockgcp server for CertificateManager CertificateIssuanceConfig                                                                                                                                                           
                                                                                                                                                                                                                                       
- Fix an error in the mapper and regenerate                                                                                                                                                                                            
- NOTE: This resource does't have an Update or Patch method (See [https://github.com/googleapis/googleapis/blob/master/google/cloud/certificatemanager/v1/certificate_issuance_config.proto](vscode-file://vscode-app/home/katriel/bin/Jetski-1.26.1/resources/app/out/vs/code/electron-browser/workbench/workbench.html)), which is why it's missing from the mock implementation                                                                                            
               
Addresses #5678
                                                                                                                                                                                                                        
### DIFF
--- a/mockgcp/mockcertificatemanager/certificateissuanceconfig.go
+++ b/mockgcp/mockcertificatemanager/certificateissuanceconfig.go
@@ -17,9 +17,12 @@ package mockcertificatemanager
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"strings"
 
 	"google.golang.org/genproto/googleapis/longrunning"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/emptypb"
 	"google.golang.org/protobuf/types/known/timestamppb"
@@ -73,7 +76,6 @@ func (s *CertificateManagerV1) CreateCertificateIssuanceConfig(ctx context.Conte
 
 	return s.operations.StartLRO(ctx, req.Parent, lroMetadata, func() (proto.Message, error) {
 		result := proto.Clone(obj).(*pb.CertificateIssuanceConfig)
-		result.Labels = nil
 		return result, nil
 	})
 }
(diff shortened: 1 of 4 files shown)


apis/certificatemanager/v1alpha1/certificateissuanceconfig_types.go
```diff
@@ -40,7 +40,7 @@ type CertificateManagerCertificateIssuanceConfigSpec struct {
 	// Required. The CA that issues the workload certificate. It includes the CA
 	//  address, type, authentication to CA service, etc.
 	// +kcc:proto:field=google.cloud.certificatemanager.v1.CertificateIssuanceConfig.certificate_authority_config
-	CertificateAuthorityConfig *CertificateIssuanceConfig_CertificateAuthorityConfig `json:"certificateAuthorityConfigRef,omitempty"`
+	CertificateAuthorityConfig *CertificateIssuanceConfig_CertificateAuthorityConfig `json:"certificateAuthorityConfig,omitempty"`
```
cheftako, line 43: Generally not ok to modify a field name. Given this resource is still in alpha we will provide an exception but it still needs to be release noted in the PR description.

katrielt: As discussed offline. This field has Ref erroneously appended. The resource itself is new and doesn't have a controller, so this will not affect resource.

mockgcp/mockcertificatemanager/certificateissuanceconfig.go
```diff
@@ -0,0 +1,133 @@
...
+	objs := []*pb.CertificateIssuanceConfig{}
+	kind := (&pb.CertificateIssuanceConfig{}).ProtoReflect().Descriptor()
+	if err := s.storage.List(ctx, kind, storage.ListOptions{}, func(obj proto.Message) error {
+		objs = append(objs, obj.(*pb.CertificateIssuanceConfig))
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+
+	// Filter by location
```
maqiuyujoyce, line 93: If `req.Parent` does not match the expected `projects/*/locations/*` format (e.g., if it's malformed or just `projects/p`), this check fails, and the code proceeds to return `objs`, which contains *all* certificate issuance configs across *all* projects and locations.

Consider adding an `else` block to return an error (e.g., `InvalidArgument`) or ensuring that we default to an empty list/error if the parent is not understood.

mockgcp/mockcertificatemanager/certificateissuanceconfig.go
```diff
@@ -0,0 +1,133 @@
...
+	lroMetadata := &pb.OperationMetadata{
+		ApiVersion: "v1",
+		CreateTime: now,
+		Target:     fqn,
+		Verb:       "create",
+	}
+
+	return s.operations.StartLRO(ctx, req.Parent, lroMetadata, func() (proto.Message, error) {
+		result := proto.Clone(obj).(*pb.CertificateIssuanceConfig)
+		result.Labels = nil
```
maqiuyujoyce: Is explicitly setting `result.Labels = nil` intended to match the specific behavior of the real `CertificateManager` API?

Typically, LRO results return the full resource state, including labels. If this is a copy-paste artifact, consider removing it.

mockgcp/mockcertificatemanager/certificateissuanceconfig.go
```diff
@@ -0,0 +1,133 @@
...
+		return result, nil
+	})
+}
+
+func (s *CertificateManagerV1) ListCertificateIssuanceConfigs(ctx context.Context, req *pb.ListCertificateIssuanceConfigsRequest) (*pb.ListCertificateIssuanceConfigsResponse, error) {
+	objs := []*pb.CertificateIssuanceConfig{}
+	kind := (&pb.CertificateIssuanceConfig{}).ProtoReflect().Descriptor()
+	if err := s.storage.List(ctx, kind, storage.ListOptions{}, func(obj proto.Message) error {
```
maqiuyujoyce, line 86: Note: This implementation ignores pagination parameters (`PageSize`, `PageToken`) by passing empty `storage.ListOptions{}`.

This is usually acceptable for mocks with small datasets, but if tests rely on pagination, this will need to be addressed.

mockgcp/mockcertificatemanager/certificateissuanceconfig.go
katrielt: Removed.

mockgcp/mockcertificatemanager/certificateissuanceconfig.go
katrielt: Good catch. Fixed.

mockgcp/mockcertificatemanager/certificateissuanceconfig.go
katrielt: Not sure if this would be needed or not, but non-the-less, asked Gemini to implement pagination.

PR #6516, opened 2026-02-16
Standardized Backup and Restore for Config Connector

fkc1e100:
This PR implements the requested backup and restore functionality for the Config Connector CLI.

Key features:
- **backup configure**: Deploys the necessary infrastructure (CronJob, Service Accounts, GCS Bucket via KCC resources).
- **backup create**: Manually triggers or runs as a CronJob to export all KCC resources in the cluster to GCS, with metadata sanitization.
- **backup status**: Provides status of recent backup jobs and lists artifacts in GCS.
- **backup restore**: Downloads resources from GCS, validates GVK compatibility, sanitizes live fields, injects acquisition annotations, and applies them in a priority-based order.

This PR was generated by the gemini-3-flash-preview model.

### DIFF
--- a/run-22071132339-failed.log
+++ b/run-22071132339-failed.log
@@ -1,0 +1,52 @@
+lint	golangci-lint	﻿2026-02-16T16:51:17.7871870Z ##[group]Run golangci/golangci-lint-action@v9
+lint	golangci-lint	2026-02-16T16:51:17.7872166Z with:
+lint	golangci-lint	2026-02-16T16:51:17.7872331Z   version: v2.7.1
+lint	golangci-lint	2026-02-16T16:51:17.7872529Z   args: --timeout=10m
+lint	golangci-lint	2026-02-16T16:51:17.7872728Z   install-mode: binary
+lint	golangci-lint	2026-02-16T16:51:17.7872911Z   install-only: false
+lint	golangci-lint	2026-02-16T16:51:17.7873249Z   github-token: ***
+lint	golangci-lint	2026-02-16T16:51:17.7873441Z   verify: true
+lint	golangci-lint	2026-02-16T16:51:17.7873618Z   only-new-issues: false
+lint	golangci-lint	2026-02-16T16:51:17.7873820Z   skip-cache: false
+lint	golangci-lint	2026-02-16T16:51:17.7873994Z   skip-save-cache: false
+lint	golangci-lint	2026-02-16T16:51:17.7874396Z   cache-invalidation-interval: 7
+lint	golangci-lint	2026-02-16T16:51:17.7874620Z   problem-matchers: false
+lint	golangci-lint	2026-02-16T16:51:17.7874816Z env:
+lint	golangci-lint	2026-02-16T16:51:17.7874962Z   GOTOOLCHAIN: local
+lint	golangci-lint	2026-02-16T16:51:17.7875141Z ##[endgroup]
+lint	golangci-lint	2026-02-16T16:51:17.9258510Z ##[group]Restore cache
+lint	golangci-lint	2026-02-16T16:51:17.9262703Z Checking for go.mod: go.mod
+lint	golangci-lint	2026-02-16T16:51:17.9267671Z (node:1980) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
+lint	golangci-lint	2026-02-16T16:51:17.9268562Z (Use `node --trace-deprecation ...` to show where the warning was created)
+lint	golangci-lint	2026-02-16T16:51:18.1844706Z Cache hit for: golangci-lint.cache-Linux-2928-536882fbb927c7cfe3fb04bfb6b25f5b20945e5f
+lint	golangci-lint	2026-02-16T16:51:18.1952307Z (node:1980) [DEP0169] DeprecationWarning: `url.parse()` behavior is not standardized and prone to errors that have security implications. Use the WHATWG URL API instead. CVEs are not issued for `url.parse()` vulnerabilities.
+lint	golangci-lint	2026-02-16T16:51:19.2979611Z Received 1370504 of 1370504 (100.0%), 1.6 MBs/sec
+lint	golangci-lint	2026-02-16T16:51:19.2980276Z Cache Size: ~1 MB (1370504 B)
+lint	golangci-lint	2026-02-16T16:51:19.3007000Z [command]/usr/bin/tar -xf /home/runner/work/_temp/91bfbb97-27de-435b-a23b-8a8202073f19/cache.tzst -P -C /home/runner/work/k8s-config-connector/k8s-config-connector --use-compress-program unzstd
+lint	golangci-lint	2026-02-16T16:51:19.6953065Z Cache restored successfully
+lint	golangci-lint	2026-02-16T16:51:19.6960229Z Restored cache for golangci-lint from key 'golangci-lint.cache-Linux-2928-536882fbb927c7cfe3fb04bfb6b25f5b20945e5f' in 1769ms
+lint	golangci-lint	2026-02-16T16:51:19.6962055Z ##[endgroup]
+lint	golangci-lint	2026-02-16T16:51:19.6962405Z ##[group]Install
+lint	golangci-lint	2026-02-16T16:51:19.6964251Z Finding needed golangci-lint version...
+lint	golangci-lint	2026-02-16T16:51:19.6968410Z Installation mode: binary
+lint	golangci-lint	2026-02-16T16:51:19.6969302Z Installing golangci-lint binary v2.7.1...
+lint	golangci-lint	2026-02-16T16:51:19.6970817Z Downloading binary https://github.com/golangci/golangci-lint/releases/download/v2.7.1/golangci-lint-2.7.1-linux-amd64.tar.gz ...
+lint	golangci-lint	2026-02-16T16:51:20.0191184Z [command]/usr/bin/tar xz --overwrite --warning=no-unknown-keyword --overwrite -C /home/runner -f /home/runner/work/_temp/a9ce5d03-e19c-4cf7-b19b-c5fba7606e4e
+lint	golangci-lint	2026-02-16T16:51:20.2614761Z Installed golangci-lint into /home/runner/golangci-lint-2.7.1-linux-amd64/golangci-lint in 565ms
+lint	golangci-lint	2026-02-16T16:51:20.2618093Z ##[endgroup]
+lint	golangci-lint	2026-02-16T16:51:20.2621435Z ##[group]run golangci-lint
+lint	golangci-lint	2026-02-16T16:51:20.2626416Z Running [/home/runner/golangci-lint-2.7.1-linux-amd64/golangci-lint config path] in [/home/runner/work/k8s-config-connector/k8s-config-connector] ...
+lint	golangci-lint	2026-02-16T16:51:20.3687018Z Running [/home/runner/golangci-lint-2.7.1-linux-amd64/golangci-lint config verify] in [/home/runner/work/k8s-config-connector/k8s-config-connector] ...
+lint	golangci-lint	2026-02-16T16:51:20.5129953Z Running [/home/runner/golangci-lint-2.7.1-linux-amd64/golangci-lint run  --timeout=10m] in [/home/runner/work/k8s-config-connector/k8s-config-connector] ...
+lint	golangci-lint	2026-02-16T16:59:22.4249508Z ##[error]pkg/cli/cmd/backup/restore.go:92:6: comparing with == will fail on wrapped errors. Use errors.Is to check for a specific error (errorlint)
+lint	golangci-lint	2026-02-16T16:59:22.4257100Z 		if err == iterator.Done {
+lint	golangci-lint	2026-02-16T16:59:22.4257372Z 		   ^
+lint	golangci-lint	2026-02-16T16:59:22.4258478Z ##[error]pkg/cli/cmd/backup/status.go:114:7: comparing with == will fail on wrapped errors. Use errors.Is to check for a specific error (errorlint)
+lint	golangci-lint	2026-02-16T16:59:22.4259538Z 			if err == iterator.Done {
+lint	golangci-lint	2026-02-16T16:59:22.4259768Z 			   ^
+lint	golangci-lint	2026-02-16T16:59:22.4259926Z 2 issues:
+lint	golangci-lint	2026-02-16T16:59:22.4260092Z * errorlint: 2
+lint	golangci-lint	2026-02-16T16:59:22.4260669Z 
+lint	golangci-lint	2026-02-16T16:59:22.4269187Z ##[error]issues found
+lint	golangci-lint	2026-02-16T16:59:22.4269923Z Ran golangci-lint in 481909ms
+lint	golangci-lint	2026-02-16T16:59:22.4270324Z ##[endgroup]
--- a/run-22071132340-failed.log
+++ b/run-22071132340-failed.log
@@ -1,0 +1,147 @@
+validations	Run validations	﻿2026-02-16T16:50:49.6730561Z ##[group]Run ./scripts/github-actions/ga-validation.sh
+validations	Run validations	2026-02-16T16:50:49.6731036Z [36;1m./scripts/github-actions/ga-validation.sh[0m
+validations	Run validations	2026-02-16T16:50:49.6787324Z shell: /usr/bin/bash -e {0}
+validations	Run validations	2026-02-16T16:50:49.6787764Z env:
+validations	Run validations	2026-02-16T16:50:49.6787947Z   GOTOOLCHAIN: local
+validations	Run validations	2026-02-16T16:50:49.6788167Z   EVENT_NAME: pull_request
+validations	Run validations	2026-02-16T16:50:49.6788438Z   COMMIT_HEAD: 5c656f0c6c40dbce79715d04f17cb9bf883d7116
+validations	Run validations	2026-02-16T16:50:49.6788739Z   COMMIT_CNT: 2
+validations	Run validations	2026-02-16T16:50:49.6788976Z   BASE_COMMIT: 235e7dbc564259fcc1616ec3dcbba621347793bf
+validations	Run validations	2026-02-16T16:50:49.6789264Z ##[endgroup]
+validations	Run validations	2026-02-16T16:50:49.6931177Z Downloading envtest assets...
+validations	Run validations	2026-02-16T16:50:51.8881137Z Running validations...
+validations	Run validations	2026-02-16T16:50:51.8985408Z mockgcp/dev/tasks/fix-gofmt
+validations	Run validations	2026-02-16T16:50:52.0220266Z go: downloading golang.org/x/tools v0.42.0
+validations	Run validations	2026-02-16T16:50:52.3542355Z go: downloading golang.org/x/telemetry v0.0.0-20260209163413-e7419c687ee4
+validations	Run validations	2026-02-16T16:50:52.3720710Z go: downloading golang.org/x/mod v0.33.0
+validations	Run validations	2026-02-16T16:51:01.2861585Z make -C operator fmt
+validations	Run validations	2026-02-16T16:51:01.2873613Z make[1]: Entering directory '/home/runner/work/k8s-config-connector/k8s-config-connector/operator'
+validations	Run validations	2026-02-16T16:51:01.2934904Z go fmt ./...
+validations	Run validations	2026-02-16T16:51:01.3541701Z go: downloading sigs.k8s.io/kubebuilder-declarative-pattern v0.20.0-beta.1.0.20250514194322-871029137730
+validations	Run validations	2026-02-16T16:51:01.4840103Z go: downloading github.com/blang/semver v3.5.1+incompatible
+validations	Run validations	2026-02-16T16:51:01.4840997Z go: downloading github.com/ghodss/yaml v1.0.0
+validations	Run validations	2026-02-16T16:51:01.4849049Z go: downloading github.com/blang/semver/v4 v4.0.0
+validations	Run validations	2026-02-16T16:51:01.4965654Z go: downloading gopkg.in/yaml.v2 v2.4.0
+validations	Run validations	2026-02-16T16:51:01.5434104Z go: downloading gomodules.xyz/jsonpatch/v2 v2.4.0
+validations	Run validations	2026-02-16T16:51:01.5436795Z go: downloading github.com/cenkalti/backoff v2.2.1+incompatible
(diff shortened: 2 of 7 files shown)


fkc1e100:
--- INVESTIGATION REPORT ---
Run ID: 22074909168
Name: Presubmit
Cause: Code Error
Details: The 'ga-validation.sh' script failed because 'go.mod' and 'go.sum' files were out of date after dependency updates in the PR.
Action Taken: Ran 'go mod tidy' in all workspace modules and committed the updated 'go.sum' and 'go.mod' files.

Run ID: 22072616429
Name: Presubmit
Cause: Code Error
Details: The 'validate-generated-files' check failed because 'pkg/controller/direct/iam/mapper.generated.go' was out of date (incorrect version string).
Action Taken: None. This was already fixed in the current branch (commit e34d337b23).

Run ID: 22072616408
Name: ci-presubmit
Cause: Code Error
Details: The check failed due to out-of-date generated files, similar to Run ID 22072616429.
Action Taken: None. Already addressed in the current branch.

Run ID: 22071132339
Name: linter
Cause: Code Error
Details: The linter reported 'errorlint' violations in 'pkg/cli/cmd/backup/restore.go' and 'pkg/cli/cmd/backup/status.go' (direct comparison of error with 'iterator.Done').
Action Taken: None. This was already fixed in the current branch (commit 05dc77603f).
----------------------------

run-22071132339-failed.log
```diff
@@ -0,0 +1,52 @@
+lint	golangci-lint	﻿2026-02-16T16:51:17.7871870Z ##[group]Run golangci/golangci-lint-action@v9
```
fkc1e100, line 1: Please remove this log file and other `run-*.log` files from the PR.

pkg/cli/cmd/backup/configure.go
```diff
@@ -0,0 +1,227 @@
...
+    name: cnrm-backup
+---
+apiVersion: iam.cnrm.cloud.google.com/v1beta1
+kind: IAMPolicyMember
+metadata:
+  name: cnrm-backup-bucket-admin
+  namespace: cnrm-system
+spec:
+  member: serviceAccount:cnrm-backup@{{.ProjectID}}.iam.gserviceaccount.com
+  role: roles/storage.admin
```
fkc1e100, line 168: `roles/storage.admin` grants full control over the bucket (including ACLs and deletion). The backup job only needs to read/write objects. Please downgrade to `roles/storage.objectAdmin` or `roles/storage.objectCreator` + `roles/storage.objectViewer` to follow least privilege principle.

pkg/cli/cmd/backup/configure.go
```diff
@@ -0,0 +1,227 @@
...
+      template:
+        metadata:
+          labels:
+            app: cnrm-backup
+        spec:
+          serviceAccountName: cnrm-backup-manager
+          containers:
+          - name: backup
+            image: gcr.io/gke-release/cnrm/controller:1.143.0
+            command: ["config-connector", "backup", "create", "--bucket", "{{.Bucket}}", "--project", "{{.ProjectID}}"]
```
fkc1e100, line 225: This image tag `1.143.0` is hardcoded and will become stale. Consider determining the version dynamically (e.g. from the CLI version if possible) or using a mechanism to keep this in sync with releases.

pkg/cli/cmd/backup/create.go
```diff
@@ -0,0 +1,180 @@
...
+					supportsList = true
+					break
+				}
+			}
+			if !supportsList {
+				continue
+			}
+
+			gvk := gv.WithKind(resource.Kind)
+			if err := backupResource(ctx, kubeClient, gcsClient, options.bucket, timestamp, gvk); err != nil {
```
fkc1e100, line 111: Scalability Warning: `backupResource` lists all items of a GVK into memory (`list.Items`). For clusters with a large number of resources (e.g. thousands of `IAMPolicyMember`s), this can consume significant memory. Consider using pagination (Continue token) and streaming the results to GCS to avoid loading everything at once.

pkg/cli/cmd/backup/restore.go
```diff
@@ -0,0 +1,258 @@
...
+		if !strings.HasSuffix(attrs.Name, ".yaml") {
+			continue
+		}
+
+		obj, err := loadObject(ctx, gcsClient, options.sourceBucket, attrs.Name)
+		if err != nil {
+			fmt.Printf("Warning: failed to load %s: %v\n", attrs.Name, err)
+			continue
+		}
+		objects = append(objects, obj)
```
fkc1e100, line 109: Scalability Warning: Similarly, loading all backup objects into memory (`objects` slice) before processing can be memory-intensive for large backups. Consider processing (validating and sorting) in a more streaming fashion if possible, although sorting requires having all items. If strict sorting is only needed for dependencies, maybe a multi-pass approach or topological sort on a graph would be better, but the current priority sort is acceptable for a first version. Just be aware of the memory limit.

pkg/cli/cmd/backup/create.go
```diff
@@ -0,0 +1,180 @@
...
+
+	// Remove status
+	unstructured.RemoveNestedField(obj.Object, "status")
+
+	// Remove common system annotations
+	annotations := obj.GetAnnotations()
+	if annotations != nil {
+		delete(annotations, "kubectl.kubernetes.io/last-applied-configuration")
+		delete(annotations, "deployment.kubernetes.io/revision")
+		delete(annotations, "cnrm.cloud.google.com/state-into-spec") // Sometimes added by KCC
```
fkc1e100, line 153: Removing `cnrm.cloud.google.com/state-into-spec` might change behavior upon restore if the user intentionally set this annotation to something other than the default (e.g. `merge`). Please verify if this removal is safe for all use cases.

pkg/cli/cmd/backup/configure.go
```diff
@@ -0,0 +1,227 @@
...
+spec:
+  displayName: Config Connector Backup Service Account
+---
+apiVersion: iam.cnrm.cloud.google.com/v1beta1
+kind: IAMPolicyMember
+metadata:
+  name: cnrm-backup-wi
+  namespace: cnrm-system
+spec:
+  member: serviceAccount:{{.ProjectID}}.svc.id.goog[cnrm-system/cnrm-backup-manager]
```
fkc1e100, line 154: The Workload Identity member binding uses `{{.ProjectID}}` which comes from the `--project` flag. This assumes the Kubernetes cluster is running in the same GCP project as the one specified for the bucket/SA. If the cluster is in a different project, this binding will be incorrect (it should point to the cluster's project). Please clarify this in the help text or allow specifying a separate cluster project.

pkg/cli/cmd/backup/create.go
```diff
@@ -0,0 +1,180 @@
...
+		return fmt.Errorf("creating kubernetes client: %w", err)
+	}
+
+	gcsClient, err := storage.NewClient(ctx)
+	if err != nil {
+		return fmt.Errorf("creating GCS client: %w", err)
+	}
+	defer gcsClient.Close()
+
+	timestamp := time.Now().Format("2006-01-02-15-04-05")
```
fkc1e100, line 74: The backup path `timestamp/...` does not include the cluster name. If multiple clusters are configured to back up to the same bucket, this could be disorganized or lead to potential collisions (though unlikely with seconds). Consider prefixing with the cluster name: `gs://<bucket>/<cluster-name>/<timestamp>/`.

pkg/cli/cmd/backup/create.go
```diff
@@ -0,0 +1,180 @@
...
+	if namespace == "" {
+		namespace = "_cluster_scoped"
+	}
+	kind := strings.ToLower(obj.GetKind())
+	name := obj.GetName()
+
+	objectName := fmt.Sprintf("%s/%s/%s/%s.yaml", timestamp, namespace, kind, name)
+
+	wc := gcsClient.Bucket(bucket).Object(objectName).NewWriter(ctx)
+	if _, err := wc.Write(data); err != nil {
```
fkc1e100, line 176: If `wc.Write` fails, `wc.Close` is not called. `storage.Writer` should be closed to clean up resources/connection even on error. Consider:
```go
if _, err := wc.Write(data); err != nil {
    _ = wc.Close()
    return err
}
return wc.Close()
```

pkg/cli/cmd/backup/restore.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+func sortResources(objects []*unstructured.Unstructured) {
+	// Priority-based sort to handle dependencies
+	// Priority 1: High-level containers (Project, Folder, Organization)
+	// Priority 2: Networking infrastructure (Network, Subnetwork)
+	// Priority 3: General resources
+	// Priority 4: IAM and dependent resources (IAMPolicy, IAMPolicyMember, etc.)
+	getPriority := func(obj *unstructured.Unstructured) int {
+		kind := obj.GetKind()
```
fkc1e100, line 187: `ServerPreferredResources` might not return the GVK present in the backup if the cluster prefers a different version (e.g. backup has v1beta1, cluster prefers v1). `validateResources` might incorrectly flag a supported resource as unsupported. Consider using `ServerGroupsAndResources` and checking if the specific Group-Version exists.

pkg/cli/cmd/backup/configure.go
```diff
@@ -0,0 +1,227 @@
...
+
+	fmt.Println("\nBackup configuration successful.")
+	return nil
+}
+
+const configureTemplate = `
+apiVersion: storage.cnrm.cloud.google.com/v1beta1
+kind: StorageBucket
+metadata:
+  name: {{.Bucket}}
```
fkc1e100, line 135: Resources are created in `cnrm-system`. This assumes `cnrm-system` is a namespace where KCC observes resources (i.e. ConfigConnectorContext exists or Cluster mode is on). If KCC is in namespaced mode and `cnrm-system` is not configured, the `StorageBucket` and `IAMServiceAccount` will not be reconciled.

pkg/cli/cmd/backup/configure.go
```diff
@@ -0,0 +1,227 @@
...
+      labels:
+        app: cnrm-backup
+    spec:
+      template:
+        metadata:
+          labels:
+            app: cnrm-backup
+        spec:
+          serviceAccountName: cnrm-backup-manager
+          containers:
```
fkc1e100, line 222: `cnrm-backup-manager` is the ServiceAccount name. Ensure this aligns with the one created in the YAML template (line 177). It does match.

pkg/cli/cmd/backup/restore.go
```diff
@@ -0,0 +1,258 @@
...
+		// Within the same priority, sort by kind then name for consistency
+		if objects[i].GetKind() != objects[j].GetKind() {
+			return objects[i].GetKind() < objects[j].GetKind()
+		}
+		return objects[i].GetName() < objects[j].GetName()
+	})
+}
+
+func applyObject(ctx context.Context, kubeClient *kubecli.Client, obj *unstructured.Unstructured, dryRun bool) error {
+	// Sanitization
```
fkc1e100, line 233: The priority sorting is a good heuristic. However, `strings.HasSuffix(kind, "Policy")` is quite broad. While it targets IAMPolicy, it might match other KCC resources (e.g. `AccessContextManagerAccessPolicy`, `NetworkSecurityAuthorizationPolicy`). Ensure this doesn't inadvertently push non-IAM resources to the end if they don't need to be. For now, it's likely safe as IAM is the main dependency.

pkg/cli/cmd/backup/status.go
```diff
@@ -0,0 +1,136 @@
...
+
+	if options.bucket != "" {
+		fmt.Printf("\nRecent Backups in GCS (gs://%s):\n", options.bucket)
+		gcsClient, err := storage.NewClient(ctx)
+		if err != nil {
+			return fmt.Errorf("creating GCS client: %w", err)
+		}
+		defer gcsClient.Close()
+
+		it := gcsClient.Bucket(options.bucket).Objects(ctx, &storage.Query{Delimiter: "/"})
```
fkc1e100, line 111: Wait, earlier linter logs indicated an issue with `errors.Is` here. The code looks correct now (`errors.Is(err, iterator.Done)`), but please ensure the `run-*.log` files (which showed the error) are stale and removed.

pkg/cli/cmd/backup/configure.go
```diff
@@ -0,0 +1,227 @@
...
+		Short: "Configure scheduled backups for Config Connector",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runConfigure(cmd.Context(), options)
+		},
+	}
+
+	cmd.Flags().StringVar(&options.cluster, "cluster", "", "Name of the cluster")
+	cmd.Flags().StringVar(&options.location, "location", "", "Region of the cluster")
+	cmd.Flags().StringVar(&options.bucket, "bucket", "", "GCS bucket name for backups")
+	cmd.Flags().StringVar(&options.bucketLocation, "bucket-location", "us-central1", "GCS bucket location")
```
fkc1e100, line 53: The default bucket location is hardcoded to "us-central1". While overrideable, consider if this should default to the cluster's region if available/detected, or keep as is.

pkg/cli/cmd/backup/configure.go
```diff
@@ -0,0 +1,227 @@
...
+  namespace: cnrm-system
+  annotations:
+    iam.gke.io/gcp-service-account: cnrm-backup@{{.ProjectID}}.iam.gserviceaccount.com
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: cnrm-backup-reader
+rules:
+- apiGroups: ["*.cnrm.cloud.google.com"]
```
fkc1e100, line 187: The `ClusterRole` grants `get, list, watch` on `*` resources in `*.cnrm.cloud.google.com`. This is necessary for backup but quite broad. It's acceptable for a backup tool, but worth noting for security review.

pkg/cli/cmd/backup/restore.go
```diff
@@ -0,0 +1,258 @@
...
+		return fmt.Errorf("creating GCS client: %w", err)
+	}
+	defer gcsClient.Close()
+
+	fmt.Printf("Loading backup from gs://%s/%s/...\n", options.sourceBucket, options.backupTimestamp)
+
+	prefix := options.backupTimestamp + "/"
+	it := gcsClient.Bucket(options.sourceBucket).Objects(ctx, &storage.Query{Prefix: prefix})
+
+	var objects []*unstructured.Unstructured
```
fkc1e100, line 90: Checking `strings.HasSuffix(attrs.Name, ".yaml")` is good. You might also want to filter out directories/prefixes if the iterator returns them (though `iterator` usually returns objects).

pkg/cli/cmd/backup/status.go
```diff
@@ -0,0 +1,136 @@
...
+
+func runStatus(ctx context.Context, options *statusOptions) error {
+	kubeClient, err := kubecli.NewClient(ctx, kubecli.ClusterOptions{})
+	if err != nil {
+		return fmt.Errorf("creating kubernetes client: %w", err)
+	}
+
+	fmt.Println("Recent Backup Jobs (CronJob):")
+	var jobs batchv1.JobList
+	if err := kubeClient.List(ctx, &jobs, client.InNamespace("cnrm-system")); err != nil {
```
fkc1e100, line 66: Hardcoded namespace `cnrm-system` for listing jobs. If the user installed KCC or the backup job in a different namespace (if that becomes supported), this will fail to find jobs.

pkg/cli/cmd/backup/configure.go
```diff
@@ -0,0 +1,227 @@
...
+			continue
+		}
+
+		obj := &unstructured.Unstructured{}
+		if err := yaml.Unmarshal(part, &obj.Object); err != nil {
+			return fmt.Errorf("unmarshaling YAML part: %w", err)
+		}
+
+		if err := kubeClient.Patch(ctx, obj, client.Apply, client.FieldOwner("config-connector-backup"), client.ForceOwnership); err != nil {
+			return fmt.Errorf("applying resource %s/%s (%s): %w", obj.GetNamespace(), obj.GetName(), obj.GetKind(), err)
```
fkc1e100, line 122: `client.FieldOwner("config-connector-backup")` is a good practice.

pkg/cli/cmd/backup/create.go
```diff
@@ -0,0 +1,180 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func backupObject(ctx context.Context, gcsClient *storage.Client, bucket, timestamp string, obj unstructured.Unstructured) error {
+	// Sanitization - Remove live fields
+	unstructured.RemoveNestedField(obj.Object, "metadata", "uid")
```
fkc1e100, line 139: **CRITICAL**: `ownerReferences` are not removed from the backup. If a resource is restored into a new cluster (or even the same cluster if the parent resource was recreated and has a new UID), the Kubernetes Garbage Collector will immediately delete the restored resource because the `ownerReference` points to a non-existent UID.
Please remove `metadata.ownerReferences`.

pkg/cli/cmd/backup/restore.go
```diff
@@ -0,0 +1,258 @@
...
+			return objects[i].GetKind() < objects[j].GetKind()
+		}
+		return objects[i].GetName() < objects[j].GetName()
+	})
+}
+
+func applyObject(ctx context.Context, kubeClient *kubecli.Client, obj *unstructured.Unstructured, dryRun bool) error {
+	// Sanitization
+	unstructured.RemoveNestedField(obj.Object, "metadata", "uid")
+	unstructured.RemoveNestedField(obj.Object, "metadata", "resourceVersion")
```
fkc1e100, line 235: **CRITICAL**: Similar to `create.go`, `ownerReferences` are not stripped here. Ensure they are removed before applying to prevent immediate garbage collection.

pkg/cli/cmd/backup/configure.go
fkc1e100, line 225: **CRITICAL**: The image `gcr.io/gke-release/cnrm/controller:1.143.0` likely contains the Config Connector controller binary, not the `config-connector` CLI binary. Unless the CLI is explicitly added to this image, the `command: ["config-connector", ...]` will fail with "executable file not found".
Ensure the image contains the CLI tool or use a dedicated image.

pkg/cli/cmd/backup/restore.go
fkc1e100, line 187: **Ordering Issue**: `Project`, `Folder`, and `Organization` are all assigned priority 1. Since `sort.SliceStable` is used on an undefined input order (map iteration), this may attempt to restore a `Project` before its parent `Folder` or `Organization`.
Refine the priority logic: Organization (0) -> Folder (1) -> Project (2).

pkg/cli/cmd/backup/restore.go
```diff
@@ -0,0 +1,258 @@
...
+			switch kind {
+			case "Project", "Folder", "Organization":
+				return 1
+			}
+		}
+
+		// Priority 2: Networking
+		if group == "compute.cnrm.cloud.google.com" {
+			switch kind {
+			case "ComputeNetwork", "ComputeSubnetwork":
```
fkc1e100, line 201: **Ordering Issue**: `ComputeNetwork` and `ComputeSubnetwork` are both priority 2. A Subnetwork depends on a Network. If sorted incorrectly, the restore will fail.
Refine the priority logic: Network (2) -> Subnetwork (3).

pkg/cli/cmd/backup/create.go
```diff
@@ -0,0 +1,180 @@
...
+	list := &unstructured.UnstructuredList{}
+	list.SetGroupVersionKind(gvk)
+
+	if err := kubeClient.List(ctx, list); err != nil {
+		return err
+	}
+
+	for _, item := range list.Items {
+		if err := backupObject(ctx, gcsClient, bucket, timestamp, item); err != nil {
+			return err
```
fkc1e100, line 130: **Robustness**: If `backupObject` fails for a single item (e.g., transient network error to GCS), the entire list backup for that GVK is aborted (`return err`).
It is better to log the error and `continue` to the next item so a partial backup is possible.

pkg/cli/cmd/backup/create.go
fkc1e100, line 74: **Collision Risk**: The backup path relies solely on `timestamp` (with seconds resolution). If automated backups run on multiple clusters pointing to the same bucket, collisions are possible if clocks align.
Include the cluster name in the path: `gs://<bucket>/<cluster-name>/<timestamp>/`.

pkg/cli/cmd/backup/configure.go
fkc1e100, line 154: **Correctness**: The Workload Identity binding assumes the cluster is in the project `{{.ProjectID}}` (which comes from the `--project` flag). However, users might specify a different project for the backup bucket/SA than where the cluster resides.
The cluster's project ID should ideally be detected from the current context or provided as a separate argument if different.

pkg/cli/cmd/backup/configure.go
fkc1e100, line 135: **Hardcoded Namespace**: The namespace `cnrm-system` is hardcoded in multiple places. If Config Connector is installed in namespaced mode (or a custom namespace), these resources will be created in the wrong place or fail if the namespace doesn't exist.
Consider discovering the installation namespace or allowing it to be configured.

run-22071132339-failed.log
fkc1e100, line 1: Please remove this log file from the PR.

run-22071132340-failed.log
```diff
@@ -0,0 +1,147 @@
+validations	Run validations	﻿2026-02-16T16:50:49.6730561Z ##[group]Run ./scripts/github-actions/ga-validation.sh
```
fkc1e100, line 1: Please remove this log file from the PR.

run-22072616408-failed.log
```diff
@@ -0,0 +1,2683 @@
+validate-generated-files	UNKNOWN STEP	﻿2026-02-16T17:43:36.7275403Z Current runner version: '2.331.0'
```
fkc1e100, line 1: Please remove this log file from the PR.

pkg/cli/cmd/backup/configure.go
```diff
@@ -0,0 +1,227 @@
...
+metadata:
+  name: cnrm-backup
+  namespace: cnrm-system
+spec:
+  displayName: Config Connector Backup Service Account
+---
+apiVersion: iam.cnrm.cloud.google.com/v1beta1
+kind: IAMPolicyMember
+metadata:
+  name: cnrm-backup-wi
```
fkc1e100, line 151: **Dependency**: The `IAMPolicyMember` `cnrm-backup-wi` is created in `cnrm-system`. It grants permissions to a Service Account in `cnrm-system`. This works only if `cnrm-system` exists. If the user installs KCC in a custom namespace, this manifest will fail to apply.

PR #6524, opened 2026-02-17
Add crd filtering for preview recorder

anhdle-sso:
### BRIEF Change description
Added CRD filtering for the preview recorder in pkg/cli/preview/recorder.go to skip non-CNRM objects and resources listed in IgnoredCRDList.

  Fixes #

The preview recorder currently attempts to preload all resources available in the discovery client. This includes non-CNRM objects and CRDs that should be ignored, which can lead to performance overhead or potential errors when attempting to list resources that KCC is not intended to manage.

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/cli/preview/recorder.go
+++ b/pkg/cli/preview/recorder.go
@@ -369,7 +369,7 @@ func (r *Recorder) PreloadGKNN(ctx context.Context, config *rest.Config, namespa
 		return fmt.Errorf("failed to get preferred resources: %w", err)
 	}
 	for _, apiResourceList := range apiResourceLists {
-		if !strings.Contains(apiResourceList.GroupVersion, ".cnrm.cloud.google.com/") {
+		if !strings.Contains(apiResourceList.GroupVersion, "."+constants.CNRMDomain+"/") {
 			continue
 		}
 
@@ -452,12 +452,12 @@ func toTrackedGVR(apiResource metav1.APIResource, apiResourceListGroupVersion sc
 		gvr.Version = apiResourceListGroupVersion.Version
 	}
 	// Not tracking CC and CCC objects.
-	if strings.HasSuffix(gvr.Group, "core.cnrm.cloud.google.com") {
+	if strings.HasSuffix(gvr.Group, constants.CoreCNRMGroup) {
 		return gvr, false
 	}
 
 	// Not tracking non-CNRM objects.
-	if !strings.Contains(gvr.Group, "cnrm.cloud.google.com") {
+	if !(strings.HasSuffix(gvr.Group, "."+constants.CNRMDomain) || gvr.Group == constants.CNRMDomain) {
 		return gvr, false
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/cli/preview/recorder.go
```diff
@@ -445,3 +436,34 @@ func contains(slice []string, str string) bool {
...
+		Version:  apiResource.Version,
+		Resource: apiResource.Name,
+	}
+	if gvr.Group == "" {
+		gvr.Group = apiResourceListGroupVersion.Group
+	}
+	if gvr.Version == "" {
+		gvr.Version = apiResourceListGroupVersion.Version
+	}
+	// Not tracking CC and CCC objects.
```
cheftako, line 454: AGENT: Consider using `constants.CoreCNRMGroup` instead of the hardcoded `"core.cnrm.cloud.google.com"` string, since the constant is already available in the imported `constants` package.

TAKO: +1

pkg/cli/preview/recorder.go
```diff
@@ -22,6 +22,8 @@ import (
 
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/structuredreporting"
+
+	constants "github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/k8s"
```
cheftako, line 26: AGENT: Importing `operator/pkg/k8s` inside `pkg/cli` tightly couples the CLI tool to the operator component's logic, which might not be desirable for architectural hygiene. Consider extracting `IgnoredCRDList` to a shared foundational package like `pkg/k8s/crds` or `pkg/k8s` so that it can be reused without creating a dependency on the operator lifecycle logic.

TAKO: I don't think we can avoid coupling the preview tool to the operator's logic. The preview tool is testing a particular operators logic.

pkg/cli/preview/recorder.go
```diff
@@ -445,3 +436,34 @@ func contains(slice []string, str string) bool {
...
+	if gvr.Group == "" {
+		gvr.Group = apiResourceListGroupVersion.Group
+	}
+	if gvr.Version == "" {
+		gvr.Version = apiResourceListGroupVersion.Version
+	}
+	// Not tracking CC and CCC objects.
+	if strings.HasSuffix(gvr.Group, "core.cnrm.cloud.google.com") {
+		return gvr, false
+	}
```
cheftako, line 457: AGENT: Using `strings.Contains` here is very permissive and could inadvertently match groups like `foo-cnrm.cloud.google.com` or `cnrm.cloud.google.com.example.org`. For a stricter match that ensures this truly is a CNRM object, consider using `strings.HasSuffix(gvr.Group, ".cnrm.cloud.google.com") || gvr.Group == "cnrm.cloud.google.com"`.

TAKO: +1

pkg/cli/preview/recorder.go
anhdle-sso: done

cheftako (approved):
Looks good thanks.

PR #6692, opened 2026-02-23
Added a MCP server for CRD change checks.

cheftako:
### BRIEF Change description

Please write a MCP server under dev/mcp. The MCP server should be about CRD changes. 

One operation should be if the CRD change is equivalent. https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5998/changes#diff-e71579c0b6550d0a4dedf6e5bf1875ab4305ceaf9070ef70a7744fa04c97f896 is an example of a equivalent CRD change. These are equivalent because no fields are added or deleted. It can be ok to add a field under status. Additionally field names or types should not change. It is fine for descriptions to change. Adding a listKind is fine.

Another operation should be if the CRD change of backward compatible. https://github.com/GoogleCloudPlatform/k8s-config-connector/commit/350b8986ff3947bb6c55829d372f03aed4107321#diff-a0c327ec2e0eccf01f7e8c780a52e09cc971a05ecbb4e99e28e0033d3b0bae71 is an example of a backward compatible change. These are backward compatible because no fields are removed or renamed. It is ok to add new fields. Additionally field types should not change Its fine for descriptions to change. 

If you have an questions or unclear decisions to make, please ask me for input. Preferably ask for input in a multiple choice manner.

```release-note
NONE
```

```docs
NONE
```

#### Intended Milestone

### Tests you have done


dev/mcp/crd-mcp-server
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
+# Copyright 2025 Google LLC
```
cheftako: 2026

dev/tools/crd-mcp-server/compare.go
```diff
@@ -0,0 +1,314 @@
+// Copyright 2025 Google LLC
```
cheftako: 2026

dev/mcp/crd-mcp-server
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
```
cheftako, line 1: AGENT: The file mode is `100644`. Since this is a script intended to be executed directly, please change the mode to `100755`.

dev/tools/crd-mcp-server/compare.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+
+	relPath, err := filepath.Rel(root, absFile)
+	if err != nil {
+		return nil, false, fmt.Errorf("computing relative path of %q from repo root %q: %w", absFile, root, err)
+	}
+
+	showCmd := exec.Command("git", "show", fmt.Sprintf("%s:%s", ref, relPath))
+	output, err := showCmd.Output()
+	if err != nil {
```
cheftako, line 72: AGENT: This swallows all errors from `git show`, including "invalid object name" (if `ref` is typoed) or other git failures.
It would be better to check if the error specifically indicates the file is missing from the revision. If it's another error (like an invalid ref), the tool should probably fail rather than reporting "File is new".

dev/tools/crd-mcp-server/compare.go
```diff
@@ -0,0 +1,314 @@
...
+		return nil, false, fmt.Errorf("getting absolute path of %q: %w", file, err)
+	}
+
+	relPath, err := filepath.Rel(root, absFile)
+	if err != nil {
+		return nil, false, fmt.Errorf("computing relative path of %q from repo root %q: %w", absFile, root, err)
+	}
+
+	showCmd := exec.Command("git", "show", fmt.Sprintf("%s:%s", ref, relPath))
```
cheftako, line 70: AGENT: Potential argument injection. If `ref` is user-controlled and starts with `-`, it could be interpreted as a flag by `git show`.
Consider verifying that `ref` is a valid revision format or strictly treating it as an argument (though `git show` argument parsing can be tricky).

dev/tools/crd-mcp-server/compare.go
```diff
@@ -0,0 +1,314 @@
...
+		return nil, false, fmt.Errorf("getting git root: %w", err)
+	}
+	root := strings.TrimSpace(string(rootOut))
+
+	absFile, err := filepath.Abs(file)
+	if err != nil {
+		return nil, false, fmt.Errorf("getting absolute path of %q: %w", file, err)
+	}
+
+	relPath, err := filepath.Rel(root, absFile)
```
cheftako, line 60: AGENT: `filepath.Rel` can return paths starting with `..` if the file is outside the repo.
If `relPath` starts with `..`, `git show` might fail or behave unexpectedly. It would be safer to return an error if the file is not within the repository root.

dev/tools/crd-mcp-server/compare.go
```diff
@@ -0,0 +1,314 @@
...
+	if err != nil {
+		// File likely does not exist at this ref (new file).
+		return nil, true, nil
+	}
+	return output, false, nil
+}
+
+// walk converts a JSONSchemaProps into a simplified nested structure for comparison.
+// Descriptions are intentionally discarded.
+func walk(s *apiextensionsv1.JSONSchemaProps) any {
```
cheftako, line 90: AGENT: `walk` is recursive and does not track visited nodes. If a schema contains a cycle (recursive definition), this will cause a stack overflow.
While rare in CRDs, adding a depth limit or cycle detection would make this robust.

dev/tools/crd-mcp-server/compare.go
```diff
@@ -0,0 +1,314 @@
...
+		m := make(map[string]any)
+		for k, v := range s.Properties {
+			val := v
+			m[k] = walk(&val)
+		}
+		return m
+	}
+
+	if s.Type == "array" {
+		if s.Items != nil && s.Items.Schema != nil {
```
cheftako, line 118: AGENT: This only handles `items.schema` (single schema for all items).
JSON Schema also supports `items` as an array of schemas (tuple validation), although rare in K8s CRDs.
It might be worth noting or adding a TODO if tuple support is ever needed.

dev/tools/crd-mcp-server/compare.go
```diff
@@ -0,0 +1,314 @@
...
+			return []any{walk(s.Items.Schema)}
+		}
+		return []any{"any"}
+	}
+
+	if s.AdditionalProperties != nil && s.AdditionalProperties.Schema != nil {
```
cheftako: AGENT: This checks `AdditionalProperties.Schema`. `AdditionalProperties` can also be a boolean (e.g., `false` to forbid unknown fields).
Changes to the boolean value (e.g. `true` -> `false`) would be missed by this logic.

dev/tools/crd-mcp-server/compare.go
```diff
@@ -0,0 +1,314 @@
...
+	"path/filepath"
+	"slices"
+	"strings"
+
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"sigs.k8s.io/yaml"
+)
+
+// parseCRD parses a CRD from YAML bytes.
+func parseCRD(data []byte) (*apiextensionsv1.CustomResourceDefinition, error) {
```
cheftako, line 32: AGENT: `yaml.Unmarshal` typically unmarshals only the first document in the stream.
If a file contains multiple YAML documents (e.g. multiple CRDs), only the first one will be checked.

dev/tools/crd-mcp-server/compare.go
```diff
@@ -0,0 +1,314 @@
...
+			diffs = append(diffs, fmt.Sprintf("%sfield type changed: %s (%s -> %s)", prefix, path, oldType, newType))
+		}
+	}
+
+	for _, path := range slices.Sorted(maps.Keys(newPaths)) {
+		if _, ok := oldPaths[path]; ok {
+			continue
+		}
+		if isUnderStatus(path) {
+			notes = append(notes, fmt.Sprintf("%sfield added under status: %s (type: %s, allowed)", prefix, path, newPaths[path]))
```
cheftako, line 296: AGENT: `isUnderStatus` logic verification:
Ensure that a field named `status` nested inside `spec` (e.g. `spec.status`) is not treated as "under status".
`flatten` uses dot notation. `spec` properties will have keys like `spec.status`.
`strings.HasPrefix("spec.status", "status.")` is false.
So this logic correctly identifies only the top-level status.

dev/tools/crd-mcp-server/compare.go
```diff
@@ -0,0 +1,314 @@
...
+		if len(s.OneOf) > 0 {
+			return "oneOf"
+		}
+		return "any"
+	}
+
+	if s.Format != "" {
+		switch s.Format {
+		case "int32", "int64":
+			return s.Format
```
cheftako, line 170: AGENT: Logic for integer formats:
If `format` is removed or changed (e.g. `int64` -> empty), `oldType` will be "int64" and `newType` will be "integer".
This will be flagged as a change. This seems strict and correct according to "Field types should not change".

dev/tools/crd-mcp-server/compare_test.go
```diff
@@ -0,0 +1,638 @@
...
+// limitations under the License.
+
+package main
+
+import (
+	"testing"
+)
+
+// baseCRD is a minimal CRD for testing.
+const baseCRD = `
```
cheftako, line 22: AGENT: Consider adding a test case where the git ref is invalid (e.g. "nonexistent-ref").
Currently, this would result in "File is new", which is misleading.

dev/tools/crd-mcp-server/compare.go
```diff
@@ -0,0 +1,314 @@
...
+	if s.AdditionalProperties != nil && s.AdditionalProperties.Schema != nil {
+		val := walk(s.AdditionalProperties.Schema)
+		if str, ok := val.(string); ok {
+			return fmt.Sprintf("map[string]%s", str)
+		}
+		return map[string]any{"KEY": val}
+	}
+
+	t := s.Type
+	if t == "" {
```
cheftako, line 141: AGENT: The `anyOf`, `allOf`, `oneOf` handling returns a string type.
If the nested schemas inside `anyOf` change, will this detect it?
`flatten` is NOT called recursively here?
Wait, `flatten` iterates over `walk`'s output.
`walk` returns `"anyOf"`. It does NOT recurse into `AnyOf` schemas.
If `AnyOf` options change, `walk` still returns `"anyOf"`.
**This is a potential issue.** If the schema definition inside `anyOf` changes, it might be missed.
Please check if `AnyOf/AllOf/OneOf` children should be walked.

dev/mcp/crd-mcp-server
```diff
@@ -0,0 +1,41 @@
...
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+
+SOURCE_DIR="${REPO_ROOT}/dev/tools/crd-mcp-server"
+BINDIR="${REPO_ROOT}/.build/mcp/bin/crd-mcp-server"
+mkdir -p "${BINDIR}"
+
+MCP_SERVER_PATH="${BINDIR}/crd-mcp-server"
+
+# Rebuild if the binary is missing or any source file is newer than the binary.
```
justinsb, line 28: Nit: you can probably just `go run`; it's fast now

PR #6694, opened 2026-02-23
Migrate lint command from kompanion to config-connector CLI

anhdle-sso:
### BRIEF Change description

Migrate the `lint` command from the `kompanion` experiment to the official Config Connector CLI (`config-connector`).

The `lint` tool is useful for identifying resources that should have the `abandon` deletion policy, helping users manage their resources more safely. Moving it to the main CLI makes it more accessible.

NONE

```release-note
Add `lint` command to `config-connector` CLI to identify resources without `abandon` deletion policy.
```

NONE

#### Intended Milestone

### Tests you have done

- [x] Run `make lint` to ensure this PR is ready for review.
- [x] Verified build success and `config-connector lint --help` output.

### DIFF
--- a/pkg/cli/cmd/lint/options/options.go
+++ b/pkg/cli/cmd/lint/options/options.go
@@ -41,7 +41,7 @@ var abandonOnDeleteCheckKind = map[string]bool{
 	"BigtableInstance":  true,
 	"ContainerCluster":  true,
 	"ContainerNodePool": true,
-	"KMSKeyRing":         true,
+	"KMSKeyRing":        true,
 	"KMSCryptoKey":      true,
 	"RedisCluster":      true,
 	"SpannerDatabase":   true,


pkg/cli/cmd/lint/execute.go
```diff
@@ -0,0 +1,345 @@
...
+
+			q.AddTask(&lintResourcesTask{
+				Result:        r,
+				Namespace:     ns.Name,
+				Resources:     resources,
+				DynamicClient: dynamicClient,
+			})
+		}
+	} else {
+		for _, resource := range resources {
```
gemmahou, line 106: When `TargetNamespaces` is empty (`perNamespace` is false), the code enters this `else` block and lists resources cluster-wide. In this path, `shouldExclude` is never called. This means the default `exclude-namespaces=kube` (and any user-provided excludes) are ignored, and resources in ignored namespaces will be reported. You should filter the `resources.Items` in `lintResourcesTask.Run` by namespace if `t.Namespace` is empty.

+1

pkg/cli/cmd/lint/options/options.go
```diff
@@ -0,0 +1,119 @@
...
+
+var abandonOnDeleteCheckKind = map[string]bool{
+	"AlloyDBCluster":    true,
+	"AlloyDBInstance":   true,
+	"BigQueryDataset":   true,
+	"BigQueryTable":     true,
+	"BigtableInstance":  true,
+	"ContainerCluster":  true,
+	"ContainerNodePool": true,
+	"KMKeyRing":         true,
```
gemmahou: Typo: This should likely be `KMSKeyRing`. Please verify the Kind name.

+1

pkg/cli/cmd/lint/execute.go
```diff
@@ -0,0 +1,345 @@
...
+}
+
+type Result struct {
+	lock sync.Mutex
+	// Map[namespace, Map[gk, []resourceID]]
+	resources map[string]map[string][]string
+}
+
+func (r *Result) Print() {
+	log.Println("Following Resources should include `cnrm.cloud.google.com/deletion-policy: abandon` annotation")
```
gemmahou: `r.Print()` uses `log.Printf`, which will include timestamps in the output. This makes the lint report hard to read and parse. Consider accepting an `io.Writer` (or using `cmd.OutOrStdout()`) and using `fmt.Fprintf` for the actual report output.

pkg/cli/cmd/lint/execute.go
```diff
@@ -0,0 +1,345 @@
...
+					errsMutex.Unlock()
+				}
+			}
+		}()
+	}
+	wg.Wait()
+
+	if len(errs) > 0 {
+		for _, err := range errs {
+			log.Printf("Error: %v", err)
```
gemmahou: Similar to the report output, using `log.Printf` for errors here will mix log metadata with the error message. Ensure this is the intended behavior for the CLI.

+1

pkg/cli/cmd/lint/options/options.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	if o.Verbose > 0 {
+		// Collect the keys to be linted so we can sort them for stable output.
+		var kindsToLint []string
+		for kind := range kindsToLintMap {
+			kindsToLint = append(kindsToLint, kind)
+		}
+		sort.Strings(kindsToLint)
+		fmt.Println("Running lint on the following resource kinds:")
```
gemmahou: `Validate` has a side effect of printing to stdout (`fmt.Println`) if verbose. It is generally better to separate validation logic from output logic. Consider moving the printing of "Running lint on..." to the `Execute` function.

+1 for better logging, but it's copied from kompanion so I'm fine to leave it for now.

pkg/cli/cmd/lint/execute.go
```diff
@@ -0,0 +1,345 @@
...
+	n.mu.Lock()
+	defer n.mu.Unlock()
+
+	if len(n.tasks) == 0 {
+		return nil
+	}
+
+	workItem := n.tasks[0]
+	n.tasks = n.tasks[1:]
+
```
gemmahou, line 392: In `GetWork`, `n.tasks[0]` is not nilled out before slicing `n.tasks = n.tasks[1:]`. While likely not an issue for this CLI tool's lifespan, it holds a reference to the completed task preventing GC. `n.tasks[0] = nil` before slicing is cleaner.

pkg/cli/cmd/lint/execute.go
```diff
@@ -0,0 +1,345 @@
...
+		var resources *unstructured.UnstructuredList
+		if t.Namespace != "" {
+			r, err := t.DynamicClient.Resource(gvr).Namespace(t.Namespace).List(ctx, metav1.ListOptions{})
+			if err != nil {
+				return fmt.Errorf("fetching gvr %s resources in namespace %s: %w", gvr, t.Namespace, err)
+			}
+			resources = r
+
+		} else {
+			r, err := t.DynamicClient.Resource(gvr).List(ctx, metav1.ListOptions{})
```
gemmahou, line 348: You are listing resources without pagination (`Limit` is 0). If there are a huge number of resources, this might consume a lot of memory as `UnstructuredList` holds all items. For a lint command, this is often acceptable, but worth noting for scalability.

pkg/cli/cmd/lint/execute.go
```diff
@@ -0,0 +1,345 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog/v2"
+)
+
+func Execute(ctx context.Context, opts *options.Options) error {
```
gemmahou: Please add unit tests for the logic in this package, particularly `shouldExclude`, the `taskQueue`, and the `Result` aggregation.

Feel free to let code-bot handle this.

pkg/cli/cmd/lint/options/options.go
```diff
@@ -0,0 +1,119 @@
...
+	ignoreNamespacesFlag = "exclude-namespaces"
+
+	includeResourcesFlag = "include-resources"
+	ignoreResourcesFlag  = "ignore-resources"
+
+	workerRoutinesFlag = "worker-routines"
+	verboseFlag        = "verbose"
+)
+
+var abandonOnDeleteCheckKind = map[string]bool{
```
gemmahou, line 36: This list of resources is hardcoded. As new stateful resources are added to Config Connector, this list will need to be updated. Consider if there's a scalable way to maintain this (e.g. tagging resources) or if manual updates are the intended strategy.

This is copied from kompanion so it's fine to leave it as is. Shall we remove the dup code under kompanion?

pkg/cli/cmd/lint/execute.go
```diff
@@ -0,0 +1,345 @@
...
+	if err != nil {
+		return err
+	}
+
+	config, err := getRESTConfig(ctx, opts)
+	if err != nil {
+		return fmt.Errorf("error building kubeconfig: %w", err)
+	}
+
+	if config.QPS == 0 {
```
gemmahou, line 61: Setting `QPS` to 100 and `Burst` to 20 is significantly higher than the defaults. This is fine for a CLI tool, but be aware it might put load on the API server if run frequently or in parallel.

Let ensure we set the reasonable default values.

pkg/cli/cmd/lint/execute.go
```diff
@@ -0,0 +1,345 @@
...
+			}
+
+			if gvr.Version == "" {
+				gvr.Version = apiResourceListGroupVersion.Version
+			}
+
+			resources = append(resources, gvr)
+		}
+	}
+	sort.Slice(resources, func(i, j int) bool {
```
gemmahou, line 241: Sorting by `resources[i].String()` works, but explicit sorting by Group, Version, and Resource might be clearer and more stable if `String()` representation changes.

pkg/cli/cmd/lint/execute.go
```diff
@@ -0,0 +1,345 @@
...
+		return resources[i].String() < resources[j].String()
+	})
+	return resources, nil
+}
+
+func contains(slice []string, str string) bool {
```
gemmahou, line 253: Consider using `slices.Contains` (if Go version allows) or ensuring this helper is necessary. The case-insensitive comparison is a nice touch though.

pkg/cli/cmd/lint.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/cli/cmd/lint/options"
+	"github.com/spf13/cobra"
+)
+
+const (
+	lintExamples = `
+	# lint Config Connector resources across all namespaces, excludes "kube" namespaces by default
+	config-connector lint
+
+	# exclude certain namespace prefixes
```
gemmahou, line 28: The flag description says "namespace prefix", but the implementation uses `strings.Contains`. Please align the description or the implementation.

+1

pkg/cli/cmd/lint/execute.go
```diff
@@ -0,0 +1,345 @@
...
+		} else {
+			r, err := t.DynamicClient.Resource(gvr).List(ctx, metav1.ListOptions{})
+			if err != nil {
+				return fmt.Errorf("fetching gvr %s resources: %w", gvr, err)
+			}
+			resources = r
+		}
+
+		for _, r := range resources.Items {
+			if !(r.GetAnnotations()["cnrm.cloud.google.com/deletion-policy"] == "abandon") {
```
gemmahou: The check `!(... == "abandon")` is correct, but be aware that `GetAnnotations()` can return nil (though accessing a key on a nil map is safe in Go).

anhdle-sso: fixed

PR #6758, opened 2026-02-24
Add structured reporting diff to ComputeForwardingRule

### LINKED ISSUE #6556: Add structured reporting diff to ComputeForwardingRule (pkg/controller/direct/compute/forwardingrule_controller.go)
Part of #5982. Tracked in the list in #6527.

Add structured reporting diff to the controller in `pkg/controller/direct/compute/forwardingrule_controller.go`.
The `structuredreporting.ReportDiff` should be used in the `Update` method of the adapter to report which fields are being updated.
This helps in debugging reconciliation loops and provides better visibility into what changed.

anhdle-sso:
### BRIEF Change description

Fixes #6556

Add structured reporting diff to the controller in `pkg/controller/direct/compute/forwardingrule_controller.go`.
The `structuredreporting.ReportDiff` should be used in the `Update` method of the adapter to report which fields are being updated.
This helps in debugging reconciliation loops and provides better visibility into what changed.

```release-note
NONE
```

```docs
NONE
```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/controller/direct/compute/forwardingrule_controller.go
+++ b/pkg/controller/direct/compute/forwardingrule_controller.go
@@ -30,6 +30,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/label"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/structuredreporting"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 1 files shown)


pkg/controller/direct/compute/forwardingrule_controller.go
```diff
@@ -335,6 +341,9 @@ func (a *forwardingRuleAdapter) Update(ctx context.Context, updateOp *directbase
 		}
 		log.V(2).Info("successfully updated ComputeForwardingRule target", "name", a.id)
 	}
+
+	structuredreporting.ReportDiff(ctx, report)
```
cheftako, line 345: There are returns before this after we try to write. Please consider how to switch the ReportDiff to a deferred call so that we don't lose  the reason for an update.

cheftako (approved):
Thanks Anh! Looks good!
Please look at the comment for future work.

PR #6760, opened 2026-02-24
Add structured reporting diff to ColabRuntimeTemplate

### LINKED ISSUE #6554: Add structured reporting diff to ColabRuntimeTemplate (pkg/controller/direct/colab/runtimetemplate_controller.go)
Part of #5982. Tracked in the list in #6527.

Add structured reporting diff to the controller in `pkg/controller/direct/colab/runtimetemplate_controller.go`.
The `structuredreporting.ReportDiff` should be used in the `Update` method of the adapter to report which fields are being updated.
This helps in debugging reconciliation loops and provides better visibility into what changed.

anhdle-sso:
### BRIEF Change description

Fixes #6554

Add structured reporting diff to the controller in `pkg/controller/direct/colab/runtimetemplate_controller.go`.
The `structuredreporting.ReportDiff` should be used in the `Update` method of the adapter to report which fields are being updated.
This helps in debugging reconciliation loops and provides better visibility into what changed.

```release-note
NONE
```

```docs
NONE
```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/controller/direct/colab/runtimetemplate_controller.go
+++ b/pkg/controller/direct/colab/runtimetemplate_controller.go
@@ -42,6 +42,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/common"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/directbase"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct/registry"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/structuredreporting"
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


pkg/controller/direct/colab/runtimetemplate_controller.go
```diff
@@ -269,7 +270,14 @@ func (a *runtimeTemplateAdapter) Update(ctx context.Context, updateOp *directbas
 
 			return updateOp.UpdateStatus(ctx, a.desired.Status, nil)
```
cheftako, line 271: Oddly I think we should consider covering this case as well. Safer upgrade is *unlikely* to catch the small window between KRM creation and updating status. So if safer upgrade sees this, it may be a case of catching an actual issue.

cheftako (approved):
Thanks Anh! Looks good!

Please see comment for future.

PR #6810, opened 2026-02-25
Added the ability to test crd equivalence call from the CLI.

cheftako:
### BRIEF Change description

Added the ability to test crd equivalence call from the CLI.

Fixes #

```release-note
NONE
```

```docs
NONE
```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/dev/tools/crd-mcp-server/main.go
+++ b/dev/tools/crd-mcp-server/main.go
@@ -16,12 +16,14 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"strings"
 
 	"github.com/metoro-io/mcp-golang"
 	"github.com/metoro-io/mcp-golang/transport/stdio"
+	"github.com/spf13/cobra"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


dev/tools/crd-mcp-server/main.go
```diff
@@ -55,7 +57,63 @@ type BackwardCompatArguments struct {
...
+	var opts EquivalentOptions
+	cmd := &cobra.Command{
+		Use:     "equivalent",
+		Short:   "equivalent",
+		Example: `crd-mcp-server equivalent`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return RunEquivalent(cmd.Context(), &opts)
+		},
+		Args: cobra.ExactArgs(0),
+	}
```
maqiuyujoyce, line 115: Agent:
> If the file argument is mandatory, consider adding cmd.MarkFlagRequired("file") to let Cobra handle the validation.

+1 to required field validation.

dev/tools/crd-mcp-server/main.go
```diff
@@ -55,7 +57,63 @@ type BackwardCompatArguments struct {
...
+	cmd.Flags().StringVarP(&opts.File, "file", "", opts.File, "Path to the CRD YAML file to check.")
```
maqiuyujoyce, line 116: Agent:
> The usage text mentions `default: HEAD`, but the default value passed to `StringVarP` is `""`. It is recommended to set `"HEAD"` as the default value in the function call to ensure the variable is correctly populated if the user omits the flag.

+1 to the defaulting. Also when `ref` is an empty string, the result varies when calling `gitShow()`, so I'd probably also do an empty check.
   * If the file is tracked and modified but not staged: git show :file returns
     the HEAD version (because the index matches HEAD until you stage).
   * If the file is staged: git show :file returns the staged content.
   * If the file is untracked: git show :file will fail (error: "path ... is not
     in the index").

dev/tools/crd-mcp-server/main.go
```diff
@@ -71,6 +129,18 @@ func main() {
 	}
 
 	<-done
+
+	return nil
+}
+
+func RunEquivalent(ctx context.Context, opts *EquivalentOptions) error {
+	result, err := runEquivalenceCheck(opts.File, opts.Ref)
```
maqiuyujoyce, line 162: Agent:
> Printing the error here and returning it will likely result in the error being printed twice (once here and once in `main` or by Cobra). It's cleaner to simply return the error and let the entry point handle the logging.

Good point though non-blocking.

dev/tools/crd-mcp-server/main.go
```diff
@@ -71,6 +129,18 @@ func main() {
...
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error: %v", err)
+		return err
```
maqiuyujoyce, line 165: Agent:
>`fmt.Fprintf(os.Stdout, result)` treats `result` as a format string. If `result` happens to contain `%`, it could lead to formatting errors. Please use `fmt.Fprint(os.Stdout, result)` or `fmt.Fprintf(os.Stdout, "%s", result)` for safety.

Good point though non-blocking.

dev/tools/crd-mcp-server/main.go
```diff
@@ -55,7 +57,63 @@ type BackwardCompatArguments struct {
...
+	cmd := &cobra.Command{
+		Use:     "stdio",
+		Short:   "stdio",
+		Example: `crd-mcp-server stdio`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return RunStdio(cmd.Context(), &opts)
+		},
+		Args: cobra.ExactArgs(0),
+	}
+
```
maqiuyujoyce, line 101: Agen:
>The description "equivalent" is quite brief. Consider expanding it to something like "Check CRD equivalence" for better clarity in the help output.

I think it's nice to have but non-blocking.

dev/tools/crd-mcp-server/main.go
```diff
@@ -55,7 +57,63 @@ type BackwardCompatArguments struct {
...
+		fmt.Fprintf(os.Stderr, "%v\n", err)
+		os.Exit(1)
+	}
+}
+
+func BuildStdioCmd() *cobra.Command {
+	var opts StdioOptions
+	cmd := &cobra.Command{
+		Use:     "stdio",
+		Short:   "stdio",
```
maqiuyujoyce, line 94: Agent:
>Similarly, the description "stdio" is brief. Consider "Run the MCP server over stdio".

I think it's nice to have but non-blocking.

dev/tools/crd-mcp-server/main.go
cheftako: Defaulting was handled deeper but made it more consistent.

dev/tools/crd-mcp-server/main.go
cheftako: Fixed but it doesn't actually change anything. Cobra does not handle the validation. You still just get a no such file error.

dev/tools/crd-mcp-server/main.go
cheftako: Fixed

PR #6836, opened 2026-02-27
Create patch to sync resource reference doc

xiaoweim:
### BRIEF Change description

This PR sync the open source resource reference docs with the google cloud public resource reference docs. This PR also create a patch to maintain the alignment of these documents. 

The differences between the open source reference docs and the google cloud public ones mostly comes from the CRD field description that was inherited from different resources' Go packages/libraries. We DO NOT modify these fields (if will override any manual changes), but Config Connector resource reference docs are edited by Google before publishing, which creates the differences. Thus, instead of directly changing the CRD description, we create a patch to reduce the noise as we continue to update our resource reference documents.

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/scripts/generate-google3-docs/resource-reference/main.go
+++ b/scripts/generate-google3-docs/resource-reference/main.go
@@ -20,6 +20,7 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


xiaoweim:
As per @maqiuyujoyce's comment, we have discussed offline and decided to take a different approach.

PR #6844, opened 2026-02-28
fix: port contention

acpana:
I've noticed that some of our tests' webhooks block on a port that is already in use. I suspect there'a race between the time that we get a port from the OS and actually send it to envtest to be used.

### DIFF
--- a/config/tests/samples/create/harness.go
+++ b/config/tests/samples/create/harness.go
@@ -152,6 +152,10 @@ func WithKubeTarget(kubeTarget string) HarnessOption {
 	}
 }
 
+// envtestWebhookMutex prevents port collision races when running envtest in parallel.
+// It locks the window between envtest allocating a free port and the manager binding it.
+var envtestWebhookMutex sync.Mutex
+
 type GCPTargetMode string
 
 const (
@@ -221,12 +225,23 @@ func NewHarness(ctx context.Context, t *testing.T, opts ...HarnessOption) *Harne
 	}
 
 	var webhooks []cnrmwebhook.Config
+	var envtestLocked bool
 
 	loadCRDs := true
 	if h.KubeTarget == "" {
 		h.KubeTarget = os.Getenv("E2E_KUBE_TARGET")
 	}
 	if h.KubeTarget == "envtest" {
+		// Lock the mutex to prevent parallel tests from stealing the port envtest is about to pick
+		envtestWebhookMutex.Lock()
+		envtestLocked = true
+		defer func() {
+			// Ensure we always unlock if something panics or t.Fatalf is called early
+			if envtestLocked {
+				envtestWebhookMutex.Unlock()
+				envtestLocked = false
+			}
+		}()
 		whCfgs, err := testwebhook.GetTestCommonWebhookConfigs()
 		if err != nil {
 			h.Fatalf("error getting common wehbook configs: %v", err)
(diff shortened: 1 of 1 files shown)


config/tests/samples/create/harness.go
```diff
@@ -152,6 +152,10 @@ func WithKubeTarget(kubeTarget string) HarnessOption {
 	}
 }
 
+// envtestWebhookMutex prevents port collision races when running envtest in parallel.
+// It locks the window between envtest allocating a free port and the manager binding it.
+var envtestWebhookMutex sync.Mutex
```
maqiuyujoyce, line 157: AGENT:
>Consider documenting that `sync.Mutex` only provides mutual exclusion for parallel tests running within the same test binary (i.e. the same package). If multiple packages (e.g. `tests/e2e` and `pkg/cli/preview`) run concurrently via `go test ./...`, each will have its own independent instance of `envtestWebhookMutex`, and port collisions could still occur across those packages.

I think it's good information, but as most of our test jobs only run the integration test within the same package, this shouldn't be a concern.

config/tests/samples/create/harness.go
```diff
@@ -221,12 +225,23 @@ func NewHarness(ctx context.Context, t *testing.T, opts ...HarnessOption) *Harne
...
 	var webhooks []cnrmwebhook.Config
+	var envtestLocked bool
 
 	loadCRDs := true
 	if h.KubeTarget == "" {
 		h.KubeTarget = os.Getenv("E2E_KUBE_TARGET")
 	}
 	if h.KubeTarget == "envtest" {
+		// Lock the mutex to prevent parallel tests from stealing the port envtest is about to pick
+		envtestWebhookMutex.Lock()
```
maqiuyujoyce, line 236: AGENT:
> If cross-process port collisions are a concern (see earlier comment), you might eventually need a file-based lock (e.g., `github.com/gofrs/flock`) or a mechanism to allocate non-overlapping port ranges per package. For intra-process parallelism, this mutex works well.

Again, just keep this as a good-to-know comment.

config/tests/samples/create/harness.go
```diff
@@ -221,12 +225,23 @@ func NewHarness(ctx context.Context, t *testing.T, opts ...HarnessOption) *Harne
...
+		envtestLocked = true
+		defer func() {
```
maqiuyujoyce, line 238: AGENT:
> A minor readability point: `defer` statements defined inside conditional blocks still evaluate their deferred functions when the *enclosing function* (`NewHarness`) returns, not when the block exits. Since the mutex must be held until near the end of the function, this works perfectly as intended. Adding a brief comment might help clarify this behavior for future readers.

I'm not concerned.

config/tests/samples/create/harness.go
```diff
@@ -763,6 +778,13 @@ func NewHarness(ctx context.Context, t *testing.T, opts ...HarnessOption) *Harne
 		time.Sleep(100 * time.Millisecond)
 	}
 
+	// The port is now successfully bound by the manager. We can safely release
+	// the lock so other parallel tests can provision their own environments.
+	if envtestLocked {
+		envtestWebhookMutex.Unlock()
+		envtestLocked = false
```
maqiuyujoyce, line 785: AGENT:
> The manual unlock combined with the state variable `envtestLocked` handles all paths correctly. Alternatively, you could extract the `envtest` setup and manager binding into a helper function and unconditionally use `defer envtestWebhookMutex.Unlock()` within that helper to simplify the state tracking.

It's nice if we extract it out to be a helper function but not a blocker.

PR #6906, opened 2026-03-04
ContainerCluster: fix deadlock when removing default node pool

eugenenuke:
Strips nodeVersion and nodeConfig from ContainerCluster when remove-default-node-pool is true, using a non-intrusive opt-in pattern to prevent Terraform validation deadlocks and permanent reconciliation loops.

### BRIEF Change description

This PR fixes a deadlock where `ContainerCluster` reconciliation fails with the error:
`error calculating diff: node_version can only be specified if remove_default_node_pool is not true`.

This occurs when the `cnrm.cloud.google.com/remove-default-node-pool: "true"` directive is used, as KCC's automated "GCP-managed fields" logic incorrectly overlays the `nodeVersion` (fetched from GCP) onto the desired state, triggering a validation conflict in the underlying Terraform provider.

Additionally, this PR unblocks cluster creation and acquisition for users with restrictive OrgPolicies (e.g., `disallowDefaultComputeServiceAccount`) who must provide `nodeConfig` (Tags/SA) for the temporary default pool, while avoiding the permanent reconciliation loops that normally follow once GKE deletes that pool.

Fixes # (Internal Bug b/489359646, b/491324027)

1. **The Deadlock:** Terraform explicitly forbids sending `node_version` (which only applies to the default pool) if `remove_default_node_pool` is enabled. Since GKE always returns a version, KCC auto-populates it into the desired state, causing a permanent `UpdateFailed` error.
2. **The Birth Path:** Users often need to specify `nodeConfig` (network tags or custom service accounts) to allow the temporary default pool to connect to the master during provisioning. Without this, the cluster creation fails.
3. **The Loop:** Once GKE deletes the default pool, any `nodeConfig` in the manifest becomes "missing" in GCP. KCC's default behavior tries to "re-sync" these missing fields forever, leading to infinite reconciliation cycles.

The fix is entirely contained within `pkg/krmtotf/legacygcpmanagedfields.go`

1. **`resolveContainerClusterNodeVersion`**: Combined with the existing `releaseChannel` logic. It uses `removeFromConfigIfNotApplied` to strip auto-filled versions from GCP while respecting any versions explicitly set by the user in their manifest.
2. **`resolveContainerClusterNodeConfig`**: 
    - **Creation Path:** Preserves `nodeConfig` during initial creation (`liveState.ID == ""`) to unblock OrgPolicies/custom requirements.
    - **Sync Path (Opt-in):** Introduced a new annotation `cnrm.cloud.google.com/remove-default-node-pool-allow-node-config: "true"`. When set, KCC will strip `nodeConfig` during reconciliation once the pool is confirmed deleted in GCP, breaking the infinite loop.

```release-note
Fixed a deadlock and infinite reconciliation loop in ContainerCluster when using the "remove-default-node-pool" directive alongside "nodeConfig" or "nodeVersion". Introduced "cnrm.cloud.google.com/remove-default-node-pool-allow-node-config" annotation to unblock cluster creation and acquisition with restricted security policies.
```

```docs
NONE
```

#### Intended Milestone

### Tests you have done

**Validation Details:**
1. Added a new comprehensive combined E2E fixture:
   - `containercluster-rm-default-np-with-node-config`: Verifies initial creation with network tags + custom Service Account Ref, and subsequent stable sync.
2. Added a dedicated **Acquisition Scenario** test:
   - `tests/e2e/testdata/scenarios/acquisition/containercluster`: Proves that KCC can successfully re-acquire existing clusters without falling into the validation trap.
3. Verified all tests pass against `mockgcp`.
4. Validated the behaviour in a real cluster - a cluster manifest with the new annotation and net tags was created successfully, whereas the same manifest, but without the new annotation demonstrated the old behaviour (UpdateFailed)

### DIFF
--- a/pkg/krmtotf/legacygcpmanagedfields.go
+++ b/pkg/krmtotf/legacygcpmanagedfields.go
@@ -113,12 +113,17 @@ func resolveContainerClusterNodeVersion(r *Resource, config map[string]interface
 	if err != nil {
 		return fmt.Errorf("error determining if release channel is set: %w", err)
 	}
-	if !found || releaseChannel == nil {
-		// Release channel is not specified, so no special behavior required.
-		return nil
-	}
-	if err := removeFromConfigIfNotApplied(r, config, "nodeVersion"); err != nil {
-		return fmt.Errorf("error resolving node version in config: %w", err)
+
+	// If the user opts to remove the default node pool, specifying a node version
+	// is also restricted by Terraform.
+	removeDefaultNodePoolDirective := "remove-default-node-pool"
+	removeDefaultNodePoolKey := k8s.FormatAnnotation(removeDefaultNodePoolDirective)
+	removeDefaultNodePool, _ := k8s.GetAnnotation(removeDefaultNodePoolKey, r)
+
+	if removeDefaultNodePool == "true" || (found && releaseChannel != nil) {
+		if err := removeFromConfigIfNotApplied(r, config, "nodeVersion"); err != nil {
+			return fmt.Errorf("error resolving node version in config: %w", err)
+		}
 	}
 	return nil
 }
@@ -150,6 +155,7 @@ func resolveContainerNodePoolVersion(r *Resource, config map[string]interface{})
 // resource. So in this case, we need to manually clean up `nodeConfig` field.
 func resolveContainerClusterNodeConfig(r *Resource, liveState *terraform.InstanceState, config map[string]interface{}) error {
 	removeDefaultNodePoolDirective := "remove-default-node-pool"
+	allowNodeConfigOverrideAnnotation := "remove-default-node-pool-allow-node-config"
 	nodeConfigFieldInTFState := "node_config"
 	nodeConfigFieldInKRMConfig := text.SnakeCaseToLowerCamelCase(nodeConfigFieldInTFState)
 
@@ -165,6 +171,22 @@ func resolveContainerClusterNodeConfig(r *Resource, liveState *terraform.Instanc
 		return fmt.Errorf("error resolving field '%v' in 'ContainerCluster': %w", nodeConfigFieldInKRMConfig, err)
 	}
 	if exists {
+		return nil
+	}
+
+	// If the resource is being created, we MUST NOT strip node_config, because the user may have specified
+	// important settings like network tags that are required for the temporary default pool to connect
+	// to the master during provisioning.
+	if liveState.ID == "" {
+		return nil
+	}
+
+	// Opt-in logic to allow specifying nodeConfig for the temporary default pool
+	// (unblocking OrgPolicies/custom requirements) without causing a permanent
+	// reconciliation loop once GKE deletes the pool.
+	allowNodeConfigOverrideKey := k8s.FormatAnnotation(allowNodeConfigOverrideAnnotation)
+	if override, ok := k8s.GetAnnotation(allowNodeConfigOverrideKey, r); ok && override == "true" {
+		unstructured.RemoveNestedField(config, nodeConfigFieldInKRMConfig)
 		return nil
 	}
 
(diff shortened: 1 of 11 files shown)


pkg/krmtotf/managedfields.go
```diff
@@ -57,6 +57,11 @@ func resolveUnmanagedFields(spec map[string]interface{}, r *Resource, liveState
 		if err = RemoveFieldsFromStateThatConflictWithSpec(stateAsKRM, spec, r.ResourceConfig, []string{}, r.TFResource.Schema); err != nil {
 			return nil, fmt.Errorf("error stripping fields from state that conflict with fields already in spec: %w", err)
 		}
+	case "ContainerCluster":
```
gemmahou: I doubt we still need this logic, as we have the new `resolveContainerClusterOverrides` function below to remove specific fields.

pkg/krmtotf/legacygcpmanagedfields.go
gemmahou: It is a bit concerning to have the additional logic above while also removing nodeVersion here. I believe the existing `removeFromConfigIfNotApplied` approach is more accurate; if a field is specified by the user, we should not remove it silently. It would be better to keep the field and log an error instead.

pkg/krmtotf/legacygcpmanagedfields.go
```diff
@@ -168,27 +176,28 @@ func resolveContainerClusterNodeConfig(r *Resource, liveState *terraform.Instanc
 		return nil
 	}
 
-	if err := removeFromConfigIfNotApplied(r, config, nodeConfigFieldInKRMConfig); err != nil {
-		return fmt.Errorf("error removing field '%v' in config: %w", nodeConfigFieldInKRMConfig, err)
-	}
+	// If the default node pool is already removed, we MUST NOT specify nodeConfig in the desired config,
+	// even if it was originally in the manifest, because it will conflict with remove_default_node_pool=true in TF.
+	unstructured.RemoveNestedField(config, nodeConfigFieldInKRMConfig)
```
gemmahou: I feel we should keep the original `removeFromConfigIfNotApplied` instead of blindly removing nodeConfig.

pkg/krmtotf/legacygcpmanagedfields.go
```diff
@@ -168,27 +176,28 @@ func resolveContainerClusterNodeConfig(r *Resource, liveState *terraform.Instanc
...
 	return nil
 }
 
 func topLevelObjectFieldExistsInStateMap(state map[string]interface{}, field string) (bool, error) {
```
gemmahou, line 199: The code change within this function seems to be a good improvement. However, since this function is used in multiple other places unrelated to ContainerCluster, I'd prefer keeping the original code to avoid any potential regressions.

pkg/test/resourcefixture/testdata/basic/container/v1beta1/containercluster/containercluster-rm-default-np-with-node-config/_generated_object_containercluster-rm-default-np-with-node-config.golden.yaml
gemmahou, line 1: The two fixtures test may not be enough, as our major test case would be re-reconcile and acquire. I feel it's better to add a scenarios test case under https://github.com/GoogleCloudPlatform/k8s-config-connector/tree/master/tests/e2e/testdata/scenarios/acquisition

pkg/krmtotf/legacygcpmanagedfields.go
eugenenuke: ack, done

pkg/krmtotf/legacygcpmanagedfields.go
eugenenuke: ack, done

pkg/test/resourcefixture/testdata/basic/container/v1beta1/containercluster/containercluster-rm-default-np-with-node-config/_generated_object_containercluster-rm-default-np-with-node-config.golden.yaml
eugenenuke: ack, done

acquisition was also tested manually on a real cluster

pkg/krmtotf/managedfields.go
eugenenuke: ack

eugenenuke:
superseded by #7235

Issue #6523, opened 2026-02-17
Add the `ipCollection` field from the `google.compute.v1.Address` to the `ComputeAddress` resource


sshcherbakov:
### Checklist

- [x] I did not find a related open enhancement request.
- [x] I understand that enhancement requests filed in the GitHub repository are by default low priority.
- [x] If this request is time-sensitive, I have submitted a corresponding issue with [GCP support](https://cloud.google.com/support-hub).

### Describe the feature or resource

Add the `ipCollection` field from the [google.compule.v1.Address](https://docs.cloud.google.com/compute/docs/reference/rest/v1/addresses) to the [ComputeAddress](https://docs.cloud.google.com/config-connector/docs/reference/resource-docs/compute/computeaddress) resource in the `computeaddresses.compute.cnrm.cloud.google.com/v1` API.

PR #6907, opened 2026-03-04
Migrate ComputeAddress to the direct controller mode and add ipCollection field to the ComputeAddress resource definition

sshcherbakov:
Implements #6523 

Migrates ComputeAddress to the direct controller mode and adds `ipCollection` field to the `ComputeAddress` resource definition.

The `ipCollection` field is a string yet. Will be updated to the reference in the next PR iteration.


sshcherbakov:
Trying to update this PR with making ipCollection field a reference instead of a string.

sshcherbakov:
This PR contains several things at the moment (I've been testing them out): ComputeAddress migration to direct controller; ipCollectionRef as reference to PDP, new PDP resource and tests for these elements.

I am not sure what the other PR does yet, will take a look.

ldanielmadariaga:
This PR is old and we have some new requirements for migration so best to close. This is also unnecessary now for the task we're trying to fix

sshcherbakov:
Hello @ldanielmadariaga 
Thank you for looking into this.
May I please ask for more details.
This is still a needed feature and the only reason it is not updated and not rebased over main is that it was waiting for [internal] review.
I see that another [PR 6896](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/6896) is also closed.

May I please ask what is the plan of getting this API support in KCC and what is the timeline for it?

I'm happy to update the PR if the review takes place and the approach is valid.

PR #6919, opened 2026-03-05
feat: syncer integ

acpana:
Adds syncer integration in operator and kccmanager

### BRIEF Change description

Fixes #

```release-note
`spec.experiments.MultiClusterLeaseSpec ` now supports integration with a syncer for KRM objects. This will help KCC take ownership of resources with service generated IDs.
```

```docs

```

#### Intended Milestone

### Tests you have done


pkg/controller/kccmanager/syncer_integration.go
```diff
@@ -0,0 +1,178 @@
...
+	syncer.SetGroupVersionKind(KRMSyncerGVK)
+
+	if err := si.client.Get(ctx, si.name, syncer); err != nil {
+		klog.Errorf("error getting KRMSyncer %s: %v", si.name, err)
+		return fmt.Errorf("error getting KRMSyncer %s: %w", si.name, err)
+	}
+
+	changed := false
+
+	dest, _, _ := unstructured.NestedString(syncer.Object, "spec", "destinationKubeConfig")
```
gemmahou: We do not have  `spec.destinationKubeConfig` in Syncer object. 
It should be `spec.remoteConfig.ClusterConfig.KubeConfigSecretRef.name`: https://github.com/gke-labs/kube-etl/blob/main/syncer/api/v1alpha1/krmsyncer_types.go#L80

pkg/controller/kccmanager/syncer_integration.go
```diff
@@ -0,0 +1,178 @@
...
+	if dest != leaderIdentity {
```
gemmahou: This is more of a design discussion: we cannot use l`eaderIdentity` directly as the kubeconfi (or kubeconfigsecret). If users can manually retrieve the secret for their clusters, we could modify the mcl CRD to include an additional field, such as `HolderKubeConfigSecret`. However, the user would still need to manually update the lease CR created by the KCC manager to set this field, which isn't an ideal UX.

pkg/controller/kccmanager/syncer_integration.go
acpana: good catch! that's why we want a good e2e test 🙏🏼

pkg/controller/kccmanager/syncer_integration.go
acpana: I think, for now, we are going to REQUIRE that the cluster lease identity and the `destinationKubeConfig` secret have the same name 😛 . We can improve and this afterwards!

gemmahou: Sounds good, this is a necessary information to be added to the setup guide! 

nit: If we are using "pull" model, it should be "remote" not "dest"

cmd/manager/main.go
```diff
@@ -72,6 +73,7 @@ func main() {
 	flag.Float32Var(&rateLimitQps, "qps", 20.0, "The client-side token bucket rate limit qps.")
 	flag.IntVar(&rateLimitBurst, "burst", 30, "The client-side token bucket rate limit burst.")
 	flag.StringVar(&leaderElectionMode, "leader-election-type", "disabled", "Leader election mode. One of: default, multicluster.")
+	flag.StringVar(&syncerIntegration, "syncer-integration", "disabled", "Enable integration with the KRMSyncer for suspending sync operations. One of: disabled, pull. Must be used with multi-cluster leader election.")
```
gemmahou: nit: I think "syncingMode" might be clearer description of the disabled and pull options.

I’m wondering if we can simplify the UX for now, by leveraging the leader election flag instead of adding a new syncer flag. If the syncer is intended to always operate alongside leader election, using the existing le flag might simplify the configuration(and below validation logic). 

There may be use cases in the future where we want to sync to a GCS bucket, but perhaps we can treat that as a separate feature and add a dedicated flag for it later?

pkg/controller/kccmanager/kccmanager.go
```diff
@@ -101,6 +102,9 @@ type Config struct {
 	// Configure manager to participate in leader election if MultiClusterLease is enabled.
 	MultiClusterLease bool
 
+	// Enable integration with KRMSyncer for multi-cluster leader election.
+	SyncerIntegration bool
```
gemmahou, line 110: Same here—if SyncerIntegration is strictly dependent on MultiClusterLease, can we consolidate these to simplify the configuration?

pkg/controller/kccmanager/syncer_integration.go
```diff
@@ -0,0 +1,301 @@
...
+
+	var rules []interface{}
+	for _, gvk := range syncerRulesGVKs {
+		var syncFields []interface{}
+		if fullReplication {
+			syncFields = []interface{}{"spec", "status"}
+		} else {
+			syncFields = []interface{}{"status"}
+			if syncerGVKsWithServiceGeneratedIDs[gvk] {
+				syncFields = append(syncFields, "spec.resourceID")
```
gemmahou: I want to highlight the generatedID lifecycle. 

For direct resources, since generatedID always saved in object's "status.externalRef", it effectively supports re-reconciliation when status is synced (either alone or with spec). Only during 'Spec-only' synchronization, that actually requires us to retrieve "spec.resourceID" from the "status". Current KCC-integrated syncer only covers the first scenario, so we don’t need to worry about "spec.resourceID".

For legacy resources, we will need to sync "spec.resourceID"(?)

pkg/controller/kccmanager/kccmanager.go
acpana: Not a bad idea but one could have leader election without syncing 🤔 . From the perspective of the manager, that allows allows us to piece-meal test some of these components.

pkg/controller/kccmanager/syncer_integration.go
acpana: I think we are ok with always syncing the spec.resourceID for both legacy and direct resources.

cmd/manager/main.go
cheftako: +1

operator/pkg/controllers/configconnector/experiments.go
```diff
@@ -76,6 +76,7 @@ func (r *Reconciler) applyMultiClusterLeaderElection(ctx context.Context, lease
 			// Add --leader-election-type=multicluster flag
 			args, _, _ := unstructured.NestedStringSlice(container, "args")
 			args = append(args, "--leader-election-type=multicluster")
+			args = append(args, "--syncer-integration=pull")
```
cheftako: Unconditionally appending `--syncer-integration=pull` whenever `MultiClusterLease` is enabled will break existing users who upgrade but do not have the `KRMSyncer` CRD (`syncer.gkelabs.io`) installed in their cluster. The KCC Manager will crash-loop during startup because `apiReader.Get` returns a `meta.NoKindMatchError` (which is not handled by `errors.IsNotFound` in `EnsureSuspended`). Consider adding an opt-in toggle or properly handling `NoKindMatchError` to gracefully degrade if the CRD is missing.

operator/pkg/apis/core/v1beta1/configconnector_types.go
```diff
@@ -88,6 +88,15 @@ type MultiClusterLeaseSpec struct {
 	//  This name must be unique across all the clusters that you are configuring the multi cluster set up for.
 	// +required
 	ClusterCandidateIdentity string `json:"clusterCandidateIdentity"`
+
+	// ResourceReplicationMode specifies how the resources will be synced to the passive clusters.
+	// Can be either 'Status' or 'Full'.
+	// 'Status' (default) means only the status field (and occasionally essential fields like spec.resourceID) are synced.
+	// 'Full' means both the spec and status are synced.
+	// +optional
+	// +kubebuilder:validation:Enum=Status;Full
```
cheftako: AGENT: Consider adding a `Disabled` option to the enum. Since `MultiClusterLease` now implicitly requires the Syncer Integration according to the operator, providing an explicit mechanism to opt-out without disabling the lease feature itself would improve backward compatibility.

pkg/controller/kccmanager/kccmanager.go
```diff
@@ -220,10 +226,42 @@ func (m *leaderElectionManager) Start(ctx context.Context) error {
...
+			if err := m.syncerInteg.EnsureSuspended(ctx); err != nil {
+				m.fatal("error ensuring syncer is suspended during startup: %v", err)
+			}
+		}
+
 		if err := m.Manager.Start(ctx); err != nil {
 			m.fatal("error running manager: %v", err)
 		}
 	}
+	m.leConfig.Callbacks.OnStoppedLeading = func() {
```
cheftako, line 258: AGENT: Configuring the syncer to pull from the new leader during `OnStoppedLeading` will likely fail and leave the syncer permanently suspended. When `OnStoppedLeading` executes, the new leader may not have acquired the lease yet, so `lease.Status.GlobalHolderIdentity` could still be empty or point to the old leader, causing `EnsurePullingFromLeader` to exit early. Furthermore, if `exitOnLeadershipLoss` is true, the manager restarts, blocks in `RunOrDie` waiting for leadership, and NEVER updates the syncer.

You MUST implement `OnNewLeader(identity string)` from `leaderelection.LeaderCallbacks` and use it to reconfigure the syncer whenever the identity changes to a remote cluster.

TAKO: We should look for the identity to 'change'. There are some utility methods which are good at looking for a state change like this.

pkg/controller/kccmanager/kccmanager.go
```diff
@@ -220,10 +226,42 @@ func (m *leaderElectionManager) Start(ctx context.Context) error {
 			m.fatal("inconsistent state: started leading but MultiClusterLease status.globalHolderIdentity is %q (expected %q)", globalHolderIdentity, m.leConfig.Lock.Identity())
 		}
 
+		if m.syncerInteg != nil {
+			if err := m.syncerInteg.EnsureSuspended(ctx); err != nil {
+				m.fatal("error ensuring syncer is suspended during startup: %v", err)
```
cheftako: If `EnsureSuspended` fails due to a transient API error (e.g., temporary network blip or API server unavailability), `m.fatal` will exit the manager entirely, causing an unnecessary pod restart. Consider implementing a retry loop here (similar to the one added in `OnStoppedLeading`) before resorting to `m.fatal`.

pkg/controller/kccmanager/syncer_integration.go
```diff
@@ -0,0 +1,327 @@
...
+
+	rules := si.getDesiredRules()
+
+	if isCreate {
+		syncerToCreate := &unstructured.Unstructured{}
+		syncerToCreate.SetGroupVersionKind(KRMSyncerGVK)
+		syncerToCreate.SetName(si.name.Name)
+		syncerToCreate.SetNamespace(si.name.Namespace)
+
+		_ = unstructured.SetNestedField(syncerToCreate.Object, leaderIdentity, "spec", "remote", "clusterConfig", "kubeConfigSecretRef", "name")
```
cheftako: Errors from `unstructured.SetNestedField` are ignored during creation, but checked during update checks (e.g., line 215). While creation uses an empty object and shouldn't realistically fail, it's safer and more consistent to handle these errors, or at least log them, to prevent silently invalid configurations if the CRD structure changes.

Also, based on a previous review comment on this PR, verify if this field should be `remoteConfig` instead of `remote`.

pkg/controller/kccmanager/syncer_integration.go
```diff
@@ -0,0 +1,327 @@
...
+		klog.Infof("successfully created KRMSyncer %s to pull from leader %s", si.name, leaderIdentity)
+		return nil
+	}
+
+	changed := false
+
+	remote, _, _ := unstructured.NestedString(syncer.Object, "spec", "remote", "clusterConfig", "kubeConfigSecretRef", "name")
+	if remote != leaderIdentity {
+		if err := unstructured.SetNestedField(syncer.Object, leaderIdentity, "spec", "remote", "clusterConfig", "kubeConfigSecretRef", "name"); err != nil {
+			return fmt.Errorf("error setting remote kubeconfig name: %w", err)
```
cheftako: You are mutating `syncer.Object` here purely as a dry-run check before setting `changed = true`. The object mutated here is the one retrieved from `apiReader` and is ultimately discarded because `RetryOnConflict` does a fresh `Get`. You can remove these `SetNestedField` calls and just set `changed = true` if the string values differ, avoiding unnecessary mutations of cached objects.

pkg/controller/kccmanager/syncer_integration.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		syncerToUpdate := &unstructured.Unstructured{}
+		syncerToUpdate.SetGroupVersionKind(KRMSyncerGVK)
+		if err := si.apiReader.Get(ctx, si.name, syncerToUpdate); err != nil {
+			return err
+		}
+
+		_ = unstructured.SetNestedField(syncerToUpdate.Object, leaderIdentity, "spec", "remote", "clusterConfig", "kubeConfigSecretRef", "name")
```
cheftako: Inside the `RetryOnConflict` block, errors from `SetNestedField` are ignored. If they were to fail, the loop would blindly call `si.client.Update` with an unmodified object. Please handle the errors and return them so the retry loop can behave correctly.

pkg/controller/kccmanager/syncer_integration.go
```diff
@@ -0,0 +1,327 @@
...
+	if err := si.apiReader.Get(ctx, si.name, syncer); err != nil {
+		if errors.IsNotFound(err) {
+			klog.Infof("KRMSyncer %s not found. Creating it in suspended state.", si.name)
+			syncerToCreate := &unstructured.Unstructured{}
+			syncerToCreate.SetGroupVersionKind(KRMSyncerGVK)
+			syncerToCreate.SetName(si.name.Name)
+			syncerToCreate.SetNamespace(si.name.Namespace)
+
+			_ = unstructured.SetNestedField(syncerToCreate.Object, true, "spec", "suspend")
+			_ = unstructured.SetNestedField(syncerToCreate.Object, "pull", "spec", "mode")
```
cheftako: When creating the syncer in `EnsureSuspended`, `mode` is set to `pull` but `remote` (or `remoteConfig`) is omitted. The `KRMSyncer` admission webhook might reject `mode: pull` if it's missing a valid remote configuration, even if `suspend` is true. Consider explicitly setting it to a default or omitting `mode: pull` if it's not strictly needed for the suspended state.

pkg/controller/kccmanager/syncer_integration.go
```diff
@@ -0,0 +1,327 @@
...
+		var fieldStrs []string
+		for _, f := range fields {
+			if s, ok := f.(string); ok {
+				fieldStrs = append(fieldStrs, s)
+			}
+		}
+		existingMap[ruleKey{group, version, kind}] = fieldStrs
+	}
+
+	for _, d := range desired {
```
cheftako, line 133: AGENT: If `existing` rules have multiple items with the exact same GVK, this map assignment overwrites previous ones. While the generated rules don't duplicate GVKs, if a user manually modified the syncer to have duplicate rules (which the CRD doesn't strictly prevent), this logic might incorrectly flag them as a match or mismatch. Not a blocker, but worth noting.

TAKO: +1

pkg/controller/kccmanager/syncer_integration_test.go
```diff
@@ -0,0 +1,306 @@
...
+
+func newKRMSyncer(name types.NamespacedName, mode string, dest string, suspend bool) *unstructured.Unstructured {
+	u := &unstructured.Unstructured{}
+	u.SetGroupVersionKind(KRMSyncerGVK)
+	u.SetName(name.Name)
+	u.SetNamespace(name.Namespace)
+
+	spec := map[string]interface{}{}
+	if mode != "" {
+		spec["mode"] = mode
```
cheftako: This uses `dest` to map to `spec.remote.clusterConfig.kubeConfigSecretRef.name`. Ensure this is kept consistent with whether the actual CRD field uses `remote` or `remoteConfig`.

pkg/controller/kccmanager/kccmanager.go
```diff
@@ -380,7 +419,29 @@ func New(ctx context.Context, restConfig *rest.Config, cfg Config) (manager.Mana
 	}
 
 	if leConfig != nil {
-		return &leaderElectionManager{Manager: mgr, leConfig: leConfig, mclConfig: mclConfig}, nil
+		var syncerInteg *SyncerIntegration
+		if mclConfig != nil && cfg.SyncerIntegration {
```
cheftako, line 451: AGENT: **Architectural Issue:** The `SyncerIntegration` relies strictly on leader election callbacks (`OnStartedLeading` and `OnStoppedLeading`) to reconcile the `KRMSyncer` state. If a manager pod starts up as a follower, it waits in `leaderelection.RunOrDie` and never invokes `EnsurePullingFromLeader`. Consequently, its `KRMSyncer` will remain in a suspended or stale state (not pointing to the current leader) until this specific pod happens to acquire and subsequently lose leadership. 

Followers should also reconcile their syncer state on startup or by continuously watching the `MultiClusterLease` to ensure they are properly configured to pull from the active leader.

TAKO: I like my comment on this above in OnStoppedLeading.

pkg/controller/kccmanager/kccmanager.go
```diff
@@ -153,7 +157,7 @@ func setUpMultiClusterLease(ctx context.Context, restConfig *rest.Config, scheme
 		// Create a new client for the lock to ensure it uses the correct configuration
```
cheftako, line 164: AGENT: If Config Connector is running in `namespace` mode (multiple controller instances), they will all retrieve the exact same cluster-scoped `ConfigConnector` object (`configconnector.core.cnrm.cloud.google.com`) and attempt to acquire the exact same `MultiClusterLease` lock (`LeaseName` and `Namespace`). 

This will cause all but one namespace controller across the entire cluster to be blocked as followers, effectively breaking namespace mode. Leader election leases must be distinct per KCC manager instance (e.g., by incorporating the manager's target namespace into the lease name).

TAKO: Ideally (at lest for V1) the operator would acquire/renew the lock and the individual managers would validate the lock was held by "them" (same ID). Having said that as long as all the managers in a cluster are using the same ID they should just be updating the lease time and renewing without a problem. One thought on this is given 10K namespaces, we get 10K manager. Renewing the lease every minute, means we are sending 170 QPS writes at the KAS, just for lease renewals...

We should also make sure our testing covers the multiple manager case. We should have a test for namespace mode, 2+ namespaces configured to reconcile and MCL and sync and ensure that 1 cluster is actually active and that the other cluster is seeing syncing in both namespaces.

pkg/controller/kccmanager/syncer_integration.go
```diff
@@ -0,0 +1,327 @@
...
+			syncerToCreate := &unstructured.Unstructured{}
+			syncerToCreate.SetGroupVersionKind(KRMSyncerGVK)
+			syncerToCreate.SetName(si.name.Name)
+			syncerToCreate.SetNamespace(si.name.Namespace)
+
+			_ = unstructured.SetNestedField(syncerToCreate.Object, true, "spec", "suspend")
+			_ = unstructured.SetNestedField(syncerToCreate.Object, "pull", "spec", "mode")
+			_ = unstructured.SetNestedSlice(syncerToCreate.Object, si.getDesiredRules(), "spec", "rules")
+
+			if err := si.client.Create(ctx, syncerToCreate); err != nil {
```
cheftako: If the `KRMSyncer` object is created by an external system (like a GitOps tool or a user) between the `Get` and `Create` calls, `client.Create` will fail with an `AlreadyExists` error. 

Because `EnsureSuspended` lacks a retry loop for creation, and the caller (`OnStartedLeading`) invokes `m.fatal` on any error, this edge case will cause the KCC manager to crash loop. Consider handling `errors.IsAlreadyExists(err)` gracefully or using a standard retry mechanism here.

cmd/manager/main.go
cheftako: This field is part of our API. We should be careful to get it "right" because we don't want to change it later. I think "sync-mode" or "syncing-mode" is a more descriptive name than "syncer-integration". Our other fields appear to be kebab case; so lets be consistent with that.

pkg/controller/kccmanager/syncer_integration.go
gemmahou: I took another look at the generated ID logic. Documenting my findings here for future reference. No action needed in this PR.

1. Direct and TF both support a fallback to status if spec.resourceID is missing([direct](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/essentialcontacts/v1beta1/contact_identity.go#L119), [TF](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/krmtotf/resource.go#L267)); **whereas DCL strictly depends on spec.resourceID for re-reconciliation.** 
2. Direct controller no longer writes back spec.resourceID. I’ve checked the Syncer code that if a target field is missing, the Syncer simply skips it rather than failing.

So I have no objection to always syncing spec.resourceID for all resources with generated IDs. This approach minimizes the complexity of distinguishing between Direct, TF, and DCL logic.

(My initial point was that we can simplify the code by not syncing spec.resourceID as long as we have the status. Like those unadoptable but reconcilable resources, they don't use spec.resourceID, but the Direct/TF controllers can still reconcile them using their status.)

cmd/manager/main.go
acpana: renamed!

pkg/controller/kccmanager/syncer_integration.go
```diff
@@ -0,0 +1,366 @@
...
+		changed = true
+	}
+
+	if !changed {
+		klog.Infof("KRMSyncer %s is already configured correctly (suspend=%v)", name, suspend)
+		return nil
+	}
+
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		syncerToUpdate := &unstructured.Unstructured{}
```
acpana: When `suspend` is true, the check to update `dest` is skipped, meaning `changed` evaluates to `false` and the function returns early. 

However, inside the `RetryOnConflict` block below, `dest` might be forcefully set to `"dummy-secret"`. Because of the early return, this dummy secret update won't actually be applied until another field (like `rules` or `mode`) triggers an update. This appears harmless since the syncer is suspended, but the logic is slightly disjointed.

pkg/controller/kccmanager/kccmanager.go
```diff
@@ -223,10 +233,76 @@ func (m *leaderElectionManager) Start(ctx context.Context) error {
...
+			// Check if we are the leader. If so, OnStartedLeading handles suspension.
+			myIdentity := m.leConfig.Lock.Identity()
+			if identity == myIdentity {
+				return
+			}
+
+			// Retry up to 5 times with a 2-second delay to handle transient API issues
+			retries := 5
+			for i := 0; i < retries; i++ {
+				if err := m.syncerInteg.EnsurePullingFromLeader(updateCtx, identity); err != nil {
```
cheftako, line 282: TAKO: This seems off. If we lost the election (We are in OnStoppedLeading()) then we should be trying to exit quickly, NOT sleeping for up to 10 seconds, while still reconciling. I think we should rely on the new manager coming up to verify the sync behavior.

pkg/controller/kccmanager/kccmanager.go
acpana: We've already added `OnNewLeader`. Are you arguing for a runnable that runs to check if the kccmanager's identity is the leader outside of the client-go guarantees? I think an earlier version of the MCL integration had that and I could bring it back if so!

pkg/controller/kccmanager/syncer_integration_test.go
acpana: it's `remote` :  https://github.com/gke-labs/kube-etl/blob/11816ec1e858022c071e53ba1e4eb9ec464ae915/syncer/api/v1alpha1/krmsyncer_types.go#L80

pkg/controller/kccmanager/kccmanager.go
acpana: I think with the changes in `OnNewLeader`, we cover this scenario, right?

pkg/controller/kccmanager/kccmanager.go
acpana: we've moved this logic away from OnStoppedLeading into OnNewLeader -- I think this is fine now, right?

pkg/controller/kccmanager/kccmanager.go
acpana: > Renewing the lease every minute, means we are sending 170 QPS writes at the KAS, just for lease renewals...

I'm curious if this would be something we could live with for V1? I think having each manager do this call is not ideal but keeping the call in the manager layer let's us keep the door open for namespace sharding too but YAGNI!

Another trick I was thinking we could implement was to designate on of the kccmanagers as the "lease leader" (maybe first one to wake up and announce itself to the operator).

> We should also make sure our testing covers the multiple manager case. We should have a test for namespace mode, 2+ namespaces configured to reconcile and MCL and sync and ensure that 1 cluster is actually active and that the other cluster is seeing syncing in both namespaces.

I think this is covered in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/7053

cheftako: > I'm curious if this would be something we could live with for V1?

Probably but I would suggest you should have a plan for monitoring and for mitigating. They should probably go in your feature doc. I would say mitigate is probably change the lease period. Ten minutes rather than one gets you down to 17 QPS. Do you have a way to tell?

pkg/controller/kccmanager/kccmanager.go
cheftako: If I understand this code correctly, its cleaning up on the way out. The method OnStoppedLeading makes me believe that. The various timeouts below make me believe that we are not going to try to immediately exit. However there are good reasons why leader election is built to exit quickly. If there is any chance reconciliation is still ongoing on the manager which has lost the election, that indicates that it would be fighting with the new active manager. We want to  avoid that.

pkg/controller/kccmanager/kccmanager.go
cheftako: `m.leConfig.Callbacks.OnStoppedLeading = func` makes me think we are still blocking shutdown.

pkg/controller/kccmanager/kccmanager.go
cheftako: OnNewLeader is good blocking shutdown is not.

pkg/controller/kccmanager/kccmanager.go
acpana: The retries and call to EnsureSuspend have been taken out!

pkg/controller/kccmanager/kccmanager.go
acpana: > Do you have a way to tell?

I think we can leverage KAS audit logs and I can add some more metrics as a follow on PR to look specifically for the syncer_integration.

cheftako: Audit isn't always turned on. Also requires a fair amount of parsing code to determine. Metrics seem like a good option. Tricky part with metrics, is that from the client side you need to aggrege a lot of different processes metrics together to get the answer.

Issue #6513, opened 2026-02-16
feat(networkservices): support lbRouteExtension


katrielt:
Support NetworkServices lbRouteExtension in v1alpha1 with a direct controller.

PR #6957, opened 2026-03-09
Implement Direct Controller for NetworkServices LBRouteExtension

katrielt:
### Change description
Implemented the direct controller, identity logic, and fixture tests for the NetworkServicesLBRouteExtension resource.

Addresses #6513

### WHY do we need this change?
To support NetworkServicesLBRouteExtension using the modern "direct" controller approach, replacing the need for Terraform-based reconciliation for this resource.

Special notes for your reviewer:
  * Manual Mappers: Implemented manual ToProto and FromProto mappers in lbrouteextension_mapper.go to handle the polymorphic service field (which can reference either a BackendService or a WasmPlugin).
  * Reference Normalization: Updated the controller to ensure both ForwardingRuleRef and BackendServiceRef are normalized to full HTTPS URLs before reconciliation.
  * Re-reconciliation Stability: Added logic to the controller to normalize GCP-returned Project Numbers back to Project IDs in URLs. This was necessary to match the KRM "desired" state and prevent infinite re-reconciliation
     loops.
  * Deterministic FieldMask: Explicitly sorted FieldMask paths in the Update method to ensure deterministic updateMask parameters in the recorded HTTP logs.

###  Tests you have done
  * Verified CRUD cycle against real GCP using hack/record-gcp.
  * Verified parity and stability against MockGCP using hack/compare-mock.
  * All fixture tests under pkg/test/resourcefixture/testdata/basic/networkservices/v1alpha1/networkserviceslbrouteextension/ are 

### DIFF
--- a/pkg/controller/direct/networkservices/lbrouteextension_controller.go
+++ b/pkg/controller/direct/networkservices/lbrouteextension_controller.go
@@ -234,8 +234,6 @@ func (a *LBRouteExtensionAdapter) resolve(ctx context.Context) (*networkservices
 	return desiredProto, nil
 }
 
-
-
 func (a *LBRouteExtensionAdapter) Create(ctx context.Context, createOp *directbase.CreateOperation) error {
 	log := klog.FromContext(ctx)
 	log.V(2).Info("creating LBRouteExtension", "name", a.id)


pkg/controller/direct/networkservices/lbrouteextension_controller.go
```diff
@@ -0,0 +1,322 @@
...
+		LbRouteExtensionId: a.id.ID(),
+		LbRouteExtension:   resource,
+	}
+	op, err := a.gcpClient.CreateLbRouteExtension(ctx, req)
+	if err != nil {
+		return fmt.Errorf("creating LBRouteExtension %s: %w", a.id, err)
+	}
+	created, err := op.Wait(ctx)
+	if err != nil {
+		return fmt.Errorf("LBRouteExtension %s waiting creation: %w", a.id, err)
```
maqiuyujoyce: Replacing the project value with `projectID` assumes that all referenced `ForwardingRule` and `BackendService` resources reside in the SAME project as the `LBRouteExtension`. This introduces severe bugs for references:

1. **Breaks Cross-Project References:** If a user configures an `External` reference pointing to `project-b`, `actual` will correctly reflect `project-b`, but this logic forcefully rewrites it to `project-a`. This breaks the comparison against `desired` (which is still `project-b`), causing an infinite update loop.

2. **Breaks Internal References via `name`:** When a user references a `BackendService` by `name`, `NormalizedExternal()` uses its `status.selfLink`, which GCP often populates with the project **number** instead of the ID. `desired` will contain this number, and GCP will return the number in `actual`. However, this function forcefully overwrites `actual` to use the project **ID**, creating a permanent mismatch between `desired` (number) and `actual` (ID).

To safely resolve diff loops caused by Project ID vs Number, you should avoid modifying `actual` directly in this manner. Either resolve numbers to IDs during the `AdapterForObject` phase if feasible, or implement a custom comparison logic in `Adapter` (or update `Normalize`) that treats matching IDs and Numbers equivalently without overwriting valid cross-project references.

mockgcp/mocknetworkservices/normalize.go
```diff
@@ -42,6 +44,25 @@ func (s *MockService) Previsit(event mockgcpregistry.Event, replacements mockgcp
...
+	u, err := url.Parse(rawURL)
+	if err == nil {
+		q := u.Query()
+		mask := q.Get("updateMask")
+		if mask != "" {
+			// The mask might be comma-separated or encoded
+			paths := strings.Split(mask, ",")
+			sort.Strings(paths)
+			q.Set("updateMask", strings.Join(paths, ","))
+			u.RawQuery = q.Encode()
```
maqiuyujoyce: Calling `q.Encode()` parses and re-encodes the entire query string. This behaves destructively by URL-encoding commas (e.g., turning `updateMask=a,b` into `updateMask=a%2Cb`) and reordering all query parameters alphabetically. This can create massive, unnecessary diffs against golden logs for other valid query parameters.

More importantly, the direct controller logic *already* explicitly sorts the `FieldMask` paths (`sort.Strings(sortedPaths)`) before assembling the `UpdateLbRouteExtensionRequest`. Because of this, the `updateMask` parameter serialized in the HTTP logs is *already* perfectly stable! You can likely remove this entire `updateMask` normalization block to avoid the encoding issues.

pkg/controller/direct/networkservices/lbrouteextension_controller.go
```diff
@@ -0,0 +1,322 @@
...
+			ref.External = "https://www.googleapis.com/compute/v1/" + ref.External
+		}
+	}
+	for i := range obj.Spec.ExtensionChains {
+		chain := &obj.Spec.ExtensionChains[i]
+		for j := range chain.Extensions {
+			extension := &chain.Extensions[j]
+			if extension.BackendServiceRef != nil {
+				external, err := extension.BackendServiceRef.NormalizedExternal(ctx, reader, obj.GetNamespace())
+				if err != nil {
```
maqiuyujoyce: If a user submits a malformed KRM object containing a `nil` pointer in the `forwardingRuleRefs` array, `ref.Normalize` will panic and crash the controller. It's best practice to add a quick `if ref == nil { continue }` safety check before calling methods on pointers from the Spec.

pkg/controller/direct/networkservices/lbrouteextension_controller.go
```diff
@@ -0,0 +1,322 @@
...
+	// Format: https://[hostname]/compute/v1/projects/[projectID_or_number]/...
+	tokens := strings.Split(url, "/")
+	if len(tokens) < 7 {
+		return url
+	}
+	// If it's a number (or just not the ID), replace it.
+	// Since we know the project ID from the identity, we can safely substitute it.
+	if tokens[6] != projectID {
+		tokens[6] = projectID
+	}
```
maqiuyujoyce, line 169: Replacing `tokens[6]` with the `LBRouteExtension`'s `projectID` assumes all referenced `ForwardingRule` and `BackendService` resources reside in the same project. Can we guarantee this is always the truth?

If a user provides a cross-project reference, this will overwrite the project ID/number of the target resource. Since `actual` will then be permanently rewritten to the wrong project, it will constantly mismatch `desired`, leading to an infinite update loop. If cross-project resources will be used, we probably need to discuss further for a proper solution to handle this issue safely.

pkg/controller/direct/networkservices/lbrouteextension_controller.go
```diff
@@ -0,0 +1,322 @@
...
+		return nil, err
+	}
+
+	// Get networkservices GCP client
+	gcpClient, err := m.client(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// Resolve references
```
maqiuyujoyce: Reference resolution should happen in the beginning of the Create() and Update() instead of being here. AdapterForObject() is called before deletion and at that time these references may be removed already.

mockgcp/mocknetworkservices/normalize.go
```diff
@@ -42,6 +44,25 @@ func (s *MockService) Previsit(event mockgcpregistry.Event, replacements mockgcp
 		return
 	}
 
+	// Normalize updateMask in query parameters to ensure stable field order in logs.
```
maqiuyujoyce: Looks like the direct controller logic *already* explicitly sorts the `FieldMask` paths (`sort.Strings(sortedPaths)`) before assembling the `UpdateLbRouteExtensionRequest`. Is this normalization needed?

pkg/controller/direct/networkservices/lbrouteextension_controller.go
```diff
@@ -0,0 +1,322 @@
...
+	lbrouteextensionpb, err := a.gcpClient.GetLbRouteExtension(ctx, req)
+	if err != nil {
+		if direct.IsNotFound(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("getting LBRouteExtension %q: %w", a.id, err)
+	}
+
+	a.actual = lbrouteextensionpb
+	a.normalizeActual(a.actual)
```
maqiuyujoyce, line 135: In principle, we shouldn't make any changes to the actual state. If the project is always the same as the project for NetworkServicesLBRouteExtension (discussed in a separate comment), then we can take it as a temporary solution for v1alpha1; if not, then we should discuss further.

pkg/controller/direct/networkservices/lbrouteextension_controller.go
```diff
@@ -0,0 +1,322 @@
...
+	resource := proto.Clone(a.desired).(*networkservicespb.LbRouteExtension)
+	resource.Name = a.id.String()
+
+	diff, err := common.CompareProtoMessage(a.actual, resource, common.BasicDiff)
+	if err != nil {
+		return fmt.Errorf("comparing LBRouteExtension %s: %w", a.id, err)
+	}
+
+	if diff.Len() == 0 {
+		log.V(2).Info("no changes detected for LBRouteExtension", "name", a.id)
```
maqiuyujoyce, line 294: We still need to update externalRef and observedState for acquisition use cases. I'd suggest you do something like:

```
if diff.Len() == 0 {
    log.V(2).Info("no changes detected for LBRouteExtension", "name", a.id)
} else {
    sortedPaths := diff.UnsortedList()
    ...
    log.V(2).Info("successfully updated LBRouteExtension", "name", a.id)
}

status := &krm.NetworkServicesLBRouteExtensionStatus{}
...
```

pkg/controller/direct/networkservices/lbrouteextension_controller.go
katrielt: Forwarding Rule and LBRouteExtension are required to be in the same project, so this would not be an issue (https://docs.cloud.google.com/service-extensions/docs/lb-extensions-overview#limitations)

pkg/controller/direct/networkservices/lbrouteextension_controller.go
katrielt: Done.

mockgcp/mocknetworkservices/normalize.go
katrielt: Good catch. Removed.

pkg/controller/direct/networkservices/lbrouteextension_controller.go
katrielt: Modified to update state even if there are no changes.

pkg/controller/direct/networkservices/lbrouteextension_controller.go
katrielt: Added a resolve method and moved from Adapter to Update/Create methods

pkg/controller/direct/networkservices/lbrouteextension_controller.go
katrielt: As I mentioned in the other comment, The forwarding-rule, lbrouteextension and backend-service all need to be in the same project, so there is no risk that normalizing will change anything.

pkg/controller/direct/networkservices/lbrouteextension_controller.go
```diff
@@ -139,7 +102,8 @@ func (m *modelLBRouteExtension) AdapterForURL(ctx context.Context, url string) (
 type LBRouteExtensionAdapter struct {
 	id        *krm.LBRouteExtensionIdentity
 	gcpClient *gcp.DepClient
-	desired   *networkservicespb.LbRouteExtension
+	obj       *krm.NetworkServicesLBRouteExtension
```
maqiuyujoyce: Nit: should `obj` be `desired`?

pkg/controller/direct/networkservices/lbrouteextension_controller.go
maqiuyujoyce: Thanks for the clarification, @katrielt ! While this (dependent Compute resources and the NetworkServicesLBRouteExtension must be under the same project) is now the truth, it's always possible that the API team extends the support to allow cross project reference.

I'm okay with how this works for now (in v1alpha1). In v1beta1, if we keep this assumption, we need to do the project ID equivalence check in first to ensure all the resources are under the same project in the beginning of the reconciliation.

pkg/controller/direct/networkservices/lbrouteextension_controller.go
```diff
@@ -197,12 +161,70 @@ func (a *LBRouteExtensionAdapter) normalizeURL(url string, projectID string) str
...
+		}
+	}
+
+	mapCtx := &direct.MapContext{}
+	desired := NetworkServicesLBRouteExtensionSpec_ToProto(mapCtx, &obj.Spec)
+	if mapCtx.Err() != nil {
+		return nil, mapCtx.Err()
+	}
+
+	// Handle GCP Labels
```
maqiuyujoyce: Nit: GCP labels should be handled in AdapterForObject() function for consistency.

maqiuyujoyce (dismissed):
/lgtm

The comments are not blockers and feel free to address the important ones in v1beta1 promotion.

pkg/controller/direct/networkservices/lbrouteextension_controller.go
katrielt: Fixed.

pkg/controller/direct/networkservices/lbrouteextension_controller.go
katrielt: Moved to Adapter.

pkg/controller/direct/networkservices/lbrouteextension_controller.go
katrielt: Addressed this now by adding a project equivalency check.

pkg/controller/direct/networkservices/lbrouteextension_controller.go
```diff
@@ -163,68 +165,108 @@ func (a *LBRouteExtensionAdapter) normalizeURL(url string, projectID string) str
...
-	}
-	desired.Labels = label.NewGCPLabelsFromK8sLabels(labels)
+	// Set GCP Labels
+	desiredProto.Labels = a.labels
+
+	return desiredProto, nil
+}
 
-	return desired, nil
+func extractProjectID(resourceName string) string {
```
maqiuyujoyce: Nit: Could you move it to the common/util package and add unit test for it?

Not a blocker.

pkg/controller/direct/networkservices/lbrouteextension_controller.go
```diff
@@ -163,68 +165,108 @@ func (a *LBRouteExtensionAdapter) normalizeURL(url string, projectID string) str
...
 			return nil, fmt.Errorf("resolving forwardingRuleRef: %w", err)
 		}
 		// GCP LBRouteExtension returns full URLs for forwarding rules.
 		// ForwardingRuleRef.Normalize strips the prefix, so we must add it back.
 		if ref.External != "" && !strings.HasPrefix(ref.External, "https://") {
 			ref.External = "https://www.googleapis.com/compute/v1/" + ref.External
 		}
+
+		// Ensure the forwarding rule is in the same project as the LBRouteExtension.
+		if refProject := extractProjectID(ref.External); refProject != "" && refProject != projectID {
```
maqiuyujoyce: I think the error cases should be: `refProject == "" || refProject != projectID` here or extractProjectID should return an error when the extracted out projectID is `""`? Same for the other checks.

pkg/controller/direct/networkservices/lbrouteextension_controller.go
```diff
@@ -163,68 +165,108 @@ func (a *LBRouteExtensionAdapter) normalizeURL(url string, projectID string) str
...
 		return nil, mapCtx.Err()
 	}
 
-	// Handle GCP Labels
-	labels := make(map[string]string)
-	for k, v := range obj.GetLabels() {
-		labels[k] = v
-	}
-	desired.Labels = label.NewGCPLabelsFromK8sLabels(labels)
+	// Set GCP Labels
```
maqiuyujoyce, line 231: Looks like you still want to move the label handling to the AdapterForObject()?

katrielt: Done

pkg/controller/direct/networkservices/lbrouteextension_controller.go
katrielt: Fixed.

pkg/controller/direct/networkservices/lbrouteextension_controller.go
katrielt: Moved to common/resource_name.go

PR #7075, opened 2026-03-17
Introduce skip-name-validation flag and consolidate tests

anhdle-sso:
### BRIEF Change description

Introduce `skip-name-validation` flag to bypass duplicate controller name checks during registration and consolidate related tests.

In integration tests and certain multi-manager scenarios, creating multiple managers in the same process can lead to conflicts in `controller-runtime`'s global controller name registry. This flag allows for opting out of that check when such conflicts are expected or managed.

Merged `pkg/controller/kccmanager/kccmanager_skipnamevalidation_test.go` into `pkg/controller/kccmanager/kccmanager_test.go` to maintain a cleaner package structure.
Ran `make fmt` to ensure proper code formatting.

```release-note
Added a `--skip-name-validation` flag to Config Connector controllers to bypass duplicate controller name checks during registration.
```

```docs
NONE
```

#### Intended Milestone

### Tests you have done

Ran all tests in `pkg/controller/kccmanager` package:
`go test -v ./pkg/controller/kccmanager -tags=integration`

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -73,7 +73,7 @@ func main() {
 	flag.Float32Var(&rateLimitQps, "qps", 20.0, "The client-side token bucket rate limit qps.")
 	flag.IntVar(&rateLimitBurst, "burst", 30, "The client-side token bucket rate limit burst.")
 	flag.StringVar(&leaderElectionMode, "leader-election-type", "disabled", "Leader election mode. One of: default, multicluster.")
-	flag.BoolVar(&skipNameValidation, "skip-name-validation", false, "option to bypass the duplicate controller name check during registration; false by default")
+	flag.BoolVar(&skipNameValidation, "skip-name-validation", false, "option to bypass the global controller name registry in controller-runtime; false by default")
 	profiler.AddFlag(flag.CommandLine)
 	flag.CommandLine.AddGoFlagSet(goflag.CommandLine)
 	flag.Parse()
--- a/pkg/controller/config.go
+++ b/pkg/controller/config.go
@@ -34,6 +34,6 @@ type Deps struct {
 	JitterGen         jitter.Generator
 	DependencyTracker *gcpwatch.DependencyTracker
 	// SkipNameValidation maps to the SkipNameValidation field in controller-runtime's controller.Options.
-	// It avoids a controller-runtime bug when running multiple managers in the same process (even if not running them concurrently).
+	// It bypasses the global controller name registry in controller-runtime which causes conflicts when running multiple managers in the same process.
 	SkipNameValidation bool
 }
--- a/pkg/controller/direct/directbase/directbase_controller.go
+++ b/pkg/controller/direct/directbase/directbase_controller.go
@@ -45,6 +45,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	crcontroller "sigs.k8s.io/controller-runtime/pkg/controller"
@@ -116,7 +117,7 @@ func add(mgr manager.Manager, r *DirectReconciler) error {
 	controllerBuilder := builder.
 		ControllerManagedBy(mgr).
 		Named(r.controllerName).
-		WithOptions(crcontroller.Options{MaxConcurrentReconciles: k8s.ControllerMaxConcurrentReconciles, SkipNameValidation: &r.SkipNameValidation, RateLimiter: ratelimiter.NewRateLimiter()}).
+		WithOptions(crcontroller.Options{MaxConcurrentReconciles: k8s.ControllerMaxConcurrentReconciles, SkipNameValidation: ptr.To(r.SkipNameValidation), RateLimiter: ratelimiter.NewRateLimiter()}).
 		WatchesRawSource(
 			source.TypedChannel(r.immediateReconcileRequests, &handler.EnqueueRequestForObject{})).
 		For(obj, builder.OnlyMetadata, builder.WithPredicates(predicateList...))
--- a/pkg/controller/gsakeysecretgenerator/gsakey_secret_generator.go
+++ b/pkg/controller/gsakeysecretgenerator/gsakey_secret_generator.go
@@ -31,6 +31,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -66,7 +67,7 @@ func Add(mgr manager.Manager, crd *apiextensions.CustomResourceDefinition, deps 
 	_, err := builder.
 		ControllerManagedBy(mgr).
 		Named(controllerName).
-		WithOptions(controller.Options{MaxConcurrentReconciles: k8s.ControllerMaxConcurrentReconciles, RateLimiter: ratelimiter.NewRateLimiter(), SkipNameValidation: &deps.SkipNameValidation}).
+		WithOptions(controller.Options{MaxConcurrentReconciles: k8s.ControllerMaxConcurrentReconciles, RateLimiter: ratelimiter.NewRateLimiter(), SkipNameValidation: ptr.To(deps.SkipNameValidation)}).
 		For(obj, builder.OnlyMetadata).
 		Build(r)
 	if err != nil {
--- a/pkg/controller/iam/auditconfig/iamauditconfig_controller.go
+++ b/pkg/controller/iam/auditconfig/iamauditconfig_controller.go
@@ -49,6 +49,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -80,7 +81,7 @@ func Add(mgr manager.Manager, deps *kontroller.Deps) error {
 		return err
 	}
 	opt := controller.Options{
-		SkipNameValidation: &deps.SkipNameValidation,
+		SkipNameValidation: ptr.To(deps.SkipNameValidation),
 	}
 	return add(mgr, reconciler, opt)
 }
--- a/pkg/controller/iam/policy/iampolicy_controller.go
+++ b/pkg/controller/iam/policy/iampolicy_controller.go
@@ -50,6 +50,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -83,7 +84,7 @@ func Add(mgr manager.Manager, deps *kontroller.Deps) error {
 		return err
 	}
 	opt := controller.Options{
-		SkipNameValidation: &deps.SkipNameValidation,
+		SkipNameValidation: ptr.To(deps.SkipNameValidation),
 	}
 	return add(mgr, reconciler, opt)
 }
--- a/pkg/controller/iam/policymember/iampolicymember_controller.go
+++ b/pkg/controller/iam/policymember/iampolicymember_controller.go
@@ -51,6 +51,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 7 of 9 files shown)


pkg/controller/kccmanager/kccmanager_test.go
```diff
@@ -79,9 +80,41 @@ func TestSchemeIsUniqueAcrossManagers(t *testing.T) {
 	}
 }
 
+func TestSkipNameValidation(t *testing.T) {
```
anhdle-sso, line 83: Note: Because `controller-runtime` uses a global map to register controller names, this map retains state across the entire lifetime of the test process. If this test is executed multiple times in the same process (e.g., using `go test -count=2`), the first manager creation will fail on the second run because the map was dirtied by the first run.

This isn't a blocker, but you might want to leave a small comment here for developers who might encounter confusing failures when testing with `-count>1`.

pkg/controller/config.go
```diff
@@ -33,4 +33,7 @@ type Deps struct {
 	Defaulters        []k8s.Defaulter
 	JitterGen         jitter.Generator
 	DependencyTracker *gcpwatch.DependencyTracker
+	// SkipNameValidation maps to the SkipNameValidation field in controller-runtime's controller.Options.
+	// It avoids a controller-runtime bug when running multiple managers in the same process (even if not running them concurrently).
```
anhdle-sso: Calling this a "bug" in `controller-runtime` is debatable. `controller-runtime` intentionally enforces unique controller names because Prometheus metrics registration will panic if multiple controllers attempt to register identical metrics for the same controller name.

Consider rephrasing to: "It bypasses the global controller name registry in controller-runtime which causes conflicts when running multiple managers in the same process."

pkg/controller/direct/directbase/directbase_controller.go
```diff
@@ -116,7 +116,7 @@ func add(mgr manager.Manager, r *DirectReconciler) error {
 	controllerBuilder := builder.
 		ControllerManagedBy(mgr).
 		Named(r.controllerName).
-		WithOptions(crcontroller.Options{MaxConcurrentReconciles: k8s.ControllerMaxConcurrentReconciles, SkipNameValidation: ptr.To(true), RateLimiter: ratelimiter.NewRateLimiter()}).
+		WithOptions(crcontroller.Options{MaxConcurrentReconciles: k8s.ControllerMaxConcurrentReconciles, SkipNameValidation: &r.SkipNameValidation, RateLimiter: ratelimiter.NewRateLimiter()}).
```
anhdle-sso: The previous code used `ptr.To(true)`. By using `&r.SkipNameValidation`, you are taking the address of the boolean field within the heap-allocated `DirectReconciler`. 

While this is safe memory-wise since the reconciler lives as long as the controller, using `ptr.To(r.SkipNameValidation)` is slightly more idiomatic and isolates the controller option from the memory of the struct.

pkg/controller/registration/registration_controller.go
```diff
@@ -352,7 +362,10 @@ func registerDefaultController(ctx context.Context, r *ReconcileRegistration, co
 				}
 			}
 			r.reconcilers[gvk] = reconcilers
-			if err := parent.Add(r.mgr, gvk, reconcilers); err != nil {
+			opt := crcontroller.Options{
+				SkipNameValidation: &r.SkipNameValidation,
```
anhdle-sso: Same as above: `opt` takes the address of `r.SkipNameValidation`. While safe because `r` (the `ReconcileRegistration`) is pinned by the manager, consider using `k8s.io/utils/ptr.To(r.SkipNameValidation)` to be consistent with standard Kubernetes pointer helpers and avoid unintentionally pinning or coupling struct fields.

config/tests/samples/create/harness.go
```diff
@@ -731,6 +731,7 @@ func NewHarness(ctx context.Context, t *testing.T, opts ...HarnessOption) *Harne
 	})
 	kccConfig.ManagerOptions.Logger = filterLogs(log)
 	kccConfig.ManagerOptions.Controller.SkipNameValidation = ptr.To(true)
+	kccConfig.SkipNameValidation = true
```
anhdle-sso, line 734: Good catch! Setting this on `kccConfig` is indeed necessary because calling `builder.WithOptions(opt)` on individual controllers will completely overwrite the manager-level defaults originally set in `ManagerOptions.Controller.SkipNameValidation`.

pkg/controller/deletiondefender/controller.go
```diff
@@ -49,7 +49,11 @@ type Reconciler struct {
 	logger    logr.Logger
 }
 
-func Add(mgr manager.Manager, crd *apiextensions.CustomResourceDefinition) error {
+func Add(mgr manager.Manager, crd *apiextensions.CustomResourceDefinition, opt controller.Options) error {
```
anhdle-sso, line 52: Initializing `MaxConcurrentReconciles` here safely preserves the original behavior without overriding any values intentionally passed in `opt`. Looks good!

pkg/controller/iam/auditconfig/iamauditconfig_controller.go
```diff
@@ -101,12 +104,19 @@ func NewReconciler(mgr manager.Manager, provider *tfschema.Provider, smLoader *s
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler.
-func add(mgr manager.Manager, r *Reconciler) error {
+func add(mgr manager.Manager, r *Reconciler, opt controller.Options) error {
+	if opt.MaxConcurrentReconciles == 0 {
+		opt.MaxConcurrentReconciles = k8s.ControllerMaxConcurrentReconciles
+	}
```
anhdle-sso, line 111: Great pattern to set default limits only if they aren't provided by the caller via `opt`. This ensures backwards compatibility with the previous `controller.Options` literal while supporting the new parameter efficiently.

cmd/manager/main.go
```diff
@@ -72,6 +73,7 @@ func main() {
 	flag.Float32Var(&rateLimitQps, "qps", 20.0, "The client-side token bucket rate limit qps.")
 	flag.IntVar(&rateLimitBurst, "burst", 30, "The client-side token bucket rate limit burst.")
 	flag.StringVar(&leaderElectionMode, "leader-election-type", "disabled", "Leader election mode. One of: default, multicluster.")
+	flag.BoolVar(&skipNameValidation, "skip-name-validation", false, "option to bypass the duplicate controller name check during registration; false by default")
```
anhdle-sso: Just a minor nit: most descriptions in this block follow either lowercase action ("configure...", "option to...") or sentence case. "option to bypass..." is consistent with `resource-name-label` above it, so it fits perfectly fine.

pkg/controller/gsakeysecretgenerator/gsakey_secret_generator.go
```diff
@@ -66,7 +66,7 @@ func Add(mgr manager.Manager, crd *apiextensions.CustomResourceDefinition, deps
 	_, err := builder.
 		ControllerManagedBy(mgr).
 		Named(controllerName).
-		WithOptions(controller.Options{MaxConcurrentReconciles: k8s.ControllerMaxConcurrentReconciles, RateLimiter: ratelimiter.NewRateLimiter()}).
+		WithOptions(controller.Options{MaxConcurrentReconciles: k8s.ControllerMaxConcurrentReconciles, RateLimiter: ratelimiter.NewRateLimiter(), SkipNameValidation: &deps.SkipNameValidation}).
```
anhdle-sso: `&deps.SkipNameValidation` takes the address of a field inside `controller.Deps`. Since `deps` is passed around and not stored directly by the `gsakeysecretgenerator` reconciler, the `controller-runtime` options struct will silently pin the `deps` object in memory.

Since `deps` contains shared manager resources (like `TFProvider`), pinning it doesn't leak memory per se, but it's a bit unclean. Using `ptr.To(deps.SkipNameValidation)` would avoid this coupling.

pkg/controller/registration/registration_controller.go
```diff
@@ -294,7 +302,7 @@ func registerDefaultController(ctx context.Context, r *ReconcileRegistration, co
 		// register the controller to automatically create secrets for GSA keys
 		if isServiceAccountKeyCRD(crd) {
 			logger.Info("registering the GSA-Key-to-Secret generation controller")
-			if err := gsakeysecretgenerator.Add(r.mgr, crd, &controller.Deps{JitterGen: r.jitterGenerator}); err != nil {
+			if err := gsakeysecretgenerator.Add(r.mgr, crd, &controller.Deps{JitterGen: r.jitterGenerator, SkipNameValidation: r.SkipNameValidation}); err != nil {
```
anhdle-sso, line 306: Propagating `SkipNameValidation: r.SkipNameValidation` here correctly cascades the bypass configuration down to the generated GSA key controllers.

cmd/deletiondefender/main.go
```diff
@@ -49,11 +49,13 @@ func main() {
 
 	var enablePprof bool
 	var pprofPort int
+	var skipNameValidation bool
 
 	profiler.AddFlag(flag.CommandLine)
 	flag.CommandLine.AddGoFlagSet(goflag.CommandLine)
 	flag.BoolVar(&enablePprof, "enable-pprof", false, "Enable the pprof server.")
 	flag.IntVar(&pprofPort, "pprof-port", 6060, "The port that the pprof server binds to if enabled.")
+	flag.BoolVar(&skipNameValidation, "skip-name-validation", false, "option to bypass the duplicate controller name check during registration; false by default")
```
cheftako, line 58: Nit: Dislike booleans in APIs. Also not sure why we need this control in the deletion defender.

cmd/deletiondefender/main.go
```diff
@@ -99,7 +101,7 @@ func main() {
 
 	// Register the registration controller, which will dynamically create controllers for
 	// all our resources.
-	if err := registration.AddDeletionDefender(mgr, &controller.Deps{}); err != nil {
+	if err := registration.AddDeletionDefender(mgr, &controller.Deps{SkipNameValidation: skipNameValidation}); err != nil {
```
cheftako, line 104: Nit: Seems like we can just pass false here.

cmd/manager/main.go
```diff
@@ -72,6 +73,7 @@ func main() {
 	flag.Float32Var(&rateLimitQps, "qps", 20.0, "The client-side token bucket rate limit qps.")
 	flag.IntVar(&rateLimitBurst, "burst", 30, "The client-side token bucket rate limit burst.")
 	flag.StringVar(&leaderElectionMode, "leader-election-type", "disabled", "Leader election mode. One of: default, multicluster.")
+	flag.BoolVar(&skipNameValidation, "skip-name-validation", false, "option to bypass the global controller name registry in controller-runtime; false by default")
```
cheftako, line 76: Nit: Dislike booleans in APIs. Also not sure why we need this control in the manager? Seems like we are running the manager from cli tool. We should chat about this. In a 10K namespace cluster, I don't think its safe to spin up 10K pods. I would rather have this run in process and take longer.

cmd/unmanageddetector/main.go
```diff
@@ -50,12 +50,14 @@ func main() {
...
 	var pprofPort int
 	var managerNamespaceIsolation string
+	var skipNameValidation bool
 
 	profiler.AddFlag(flag.CommandLine)
 	flag.CommandLine.AddGoFlagSet(goflag.CommandLine)
 	flag.BoolVar(&enablePprof, "enable-pprof", false, "Enable the pprof server.")
 	flag.IntVar(&pprofPort, "pprof-port", 6060, "The port that the pprof server binds to if enabled.")
 	flag.StringVar(&managerNamespaceIsolation, k8s.ManagerNamespaceIsolationFlag, k8s.ManagerNamespaceIsolationShared, fmt.Sprintf("'%s' if all controller managers run in shared 'cnrm-system' namespace, '%s' if controller managers run in dedicated namespace. Default is '%s'", k8s.ManagerNamespaceIsolationShared, k8s.ManagerNamespaceIsolationDedicated, k8s.ManagerNamespaceIsolationShared))
+	flag.BoolVar(&skipNameValidation, "skip-name-validation", false, "option to bypass the duplicate controller name check during registration; false by default")
```
cheftako, line 60: Nit: Dislike booleans in APIs. Also not sure why we need this control in the unmanaged detector? Lets get a bug going and clean this up before 1.48

pkg/controller/deletiondefender/controller.go
```diff
@@ -49,7 +49,11 @@ type Reconciler struct {
...
+	if opt.MaxConcurrentReconciles == 0 {
```
cheftako, line 53: NIT: Should probably be <= 0. Assuming we not support negatives.

pkg/controller/iam/auditconfig/iamauditconfig_controller.go
```diff
@@ -101,12 +105,19 @@ func NewReconciler(mgr manager.Manager, provider *tfschema.Provider, smLoader *s
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler.
-func add(mgr manager.Manager, r *Reconciler) error {
+func add(mgr manager.Manager, r *Reconciler, opt controller.Options) error {
+	if opt.MaxConcurrentReconciles == 0 {
```
cheftako, line 109: NIT: <= 0

pkg/controller/iam/policy/iampolicy_controller.go
```diff
@@ -108,12 +112,19 @@ func NewReconciler(mgr manager.Manager, provider *tfschema.Provider, smLoader *s
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler.
-func add(mgr manager.Manager, r *ReconcileIAMPolicy) error {
+func add(mgr manager.Manager, r *ReconcileIAMPolicy, opt controller.Options) error {
+	if opt.MaxConcurrentReconciles == 0 {
```
cheftako, line 116: NIT: <= 0

pkg/controller/iam/policymember/iampolicymember_controller.go
```diff
@@ -108,12 +112,19 @@ func NewReconciler(mgr manager.Manager, provider *tfschema.Provider, smLoader *s
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler.
-func add(mgr manager.Manager, r *Reconciler) error {
+func add(mgr manager.Manager, r *Reconciler, opt controller.Options) error {
+	if opt.MaxConcurrentReconciles == 0 {
```
cheftako, line 116: NIT: <= 0

pkg/controller/unmanageddetector/controller.go
```diff
@@ -47,9 +47,13 @@ type Reconciler struct {
 	gvk schema.GroupVersionKind
 }
 
-func Add(ctx context.Context, mgr manager.Manager, gvk schema.GroupVersionKind) error {
+func Add(ctx context.Context, mgr manager.Manager, gvk schema.GroupVersionKind, opt controller.Options) error {
 	logger := crlog.FromContext(ctx)
 
+	if opt.MaxConcurrentReconciles == 0 {
```
cheftako, line 53: <= 0

PR #7083, opened 2026-03-18
refactor(preview): support alternative controller comparison in preview mode

anhdle-sso:
### BRIEF Change description
Refactors the `config-connector preview` command to execute reconciliation twice: first using the default controller for each resource, and second using an alternative controller (if supported). The preview report is enhanced to provide a side-by-side comparison of the results, including controller types, reconciliation status, and field diffs.

Key changes:
- **Dual-Run Execution**: `pkg/cli/cmd/preview/execute.go` now triggers two consecutive KCC manager preview runs, isolating default and alternative controller behaviors.
- **Combined Summary Report**: Updated `pkg/cli/preview/result.go` to generate a unified summary table showing results from both runs. Resources are matched by `GKNN` (Group, Kind, Namespace, Name).
- **Schema Refactoring**: Optimized `RecorderReconciledResults` to use a flat map keyed by `GKNN`, improving lookup performance and simplifying result aggregation.
- **Timeout Resilience**: Enhanced timeout handling to report the number of resources that failed to reconcile within the allotted time instead of failing the entire command.
- **DeepCopy Support**: Added `DeepCopy` methods to `Recorder` and its components to safely branch the recorder state between the dual runs.
- **Support for Alternative Controller Identification**: Added logic to automatically identify which resources have alternative controllers defined in `static_config.go`.

Fixes #

To enable developers and users to compare how different controllers (e.g., Terraform vs. Direct) reconcile the same KRM resources. This helps in validating migrations to direct controllers by highlighting discrepancies in generated GCP actions and field mappings.

The `CombinedSummaryReport` now outputs a wider table with `DEFAULT-CONTROLLER`, `DEFAULT-RESULT`, `DEFAULT-DIFFS`, `ALTERNATIVE-CONTROLLER`, `ALTERNATIVE-RESULT`, and `ALTERNATIVE-DIFFS` columns.

```release-note
Enhanced `config-connector preview` to support side-by-side comparison between default and alternative controllers.
```

```docs
NONE
```

#### Intended Milestone

Please indicate water intended milestone. 

### Tests you have done

- [x] Added unit tests for alternative controller mapping logic in `pkg/cli/preview/result_test.go`.

### DIFF
--- a/pkg/cli/cmd/preview/execute.go
+++ b/pkg/cli/cmd/preview/execute.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	corepreview "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/cli/preview"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/resourceconfig"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 	"k8s.io/client-go/rest"
@@ -76,6 +77,10 @@ func Execute(ctx context.Context, opts *Options) error {
 		UpstreamGCPBurst:         opts.GCPBurst,
 		Namespace:                opts.Namespace,
 	}
+	if opts.Timeout == 0 {
+		opts.Timeout = defaultTimeout
+	}
+
 	// First run: execute the KCC Manager configuring resources strictly under their default controller type
 	// This generates baseline preview results reflecting standard reconciliation behavior.
 	defaultRunRecorder, err := runKCCManagerPreview(ctx, recorder, instanceOptions, opts.Timeout)
@@ -83,7 +88,7 @@ func Execute(ctx context.Context, opts *Options) error {
 		var timeoutErr *TimeoutError
 		if errors.As(err, &timeoutErr) {
 			// Log out the timeout error and the number of resources not fully reconciled, then continue the run.
-			klog.Errorf("Timeout reached during default controller preview run. Number of resources not fully reconciled: %d. Error: %v", defaultRunRecorder.RemainResourcesCount, err)
+			klog.Errorf("Timeout reached during default controller preview run. Number of resources not fully reconciled: %d. Error: %v", defaultRunRecorder.GetRemainResourcesCount(), err)
 		} else {
 			return fmt.Errorf("error running KCC manager preview with default controller type: %w", err)
 		}
@@ -99,7 +104,7 @@ func Execute(ctx context.Context, opts *Options) error {
 		var timeoutErr *TimeoutError
 		if errors.As(err, &timeoutErr) {
 			// Log out the timeout error and the number of resources not fully reconciled, then continue the run.
-			klog.Errorf("Timeout reached during alternative controller preview run. Number of resources not fully reconciled: %d. Error: %v", alternativeRunRecorder.RemainResourcesCount, err)
+			klog.Errorf("Timeout reached during alternative controller preview run. Number of resources not fully reconciled: %d. Error: %v", alternativeRunRecorder.GetRemainResourcesCount(), err)
 		} else {
 			return fmt.Errorf("error running KCC manager preview with alternative controller type: %w", err)
 		}
@@ -139,7 +144,8 @@ func printCapturedObjects(defaultRunRecorder, alternativeRunRecorder *coreprevie
 	now := time.Now()
 	timestamp := now.Format("20060102-150405.000")
 	summaryFile := fmt.Sprintf("%s-%s", opts.ReportNamePrefix, timestamp)
-	if err := defaultRunResult.CombinedSummaryReport(summaryFile, alternativeRunResult); err != nil {
+	altExpectedMap := corepreview.GetAlternativeControllerExpectedMap(resourceconfig.ControllerConfigStatic)
+	if err := defaultRunResult.CombinedSummaryReport(summaryFile, alternativeRunResult, altExpectedMap); err != nil {
 		return fmt.Errorf("error writing summary: %w", err)
 	}
 
@@ -168,6 +174,7 @@ func runKCCManagerPreview(ctx context.Context, recorder *corepreview.Recorder, i
 
 	errChan := make(chan error, 1)
 	go func() {
+		// The manager runs in a background goroutine and continuously mutates the recorder.
 		errChan <- instance.Start(ctx)
 	}()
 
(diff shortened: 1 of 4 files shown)


pkg/cli/cmd/preview/execute.go
```diff
@@ -133,20 +135,61 @@ func getGCPAuthorization(ctx context.Context, opts *Options) (oauth2.TokenSource
...
+		errChan <- instance.Start(ctx)
+	}()
+
+	select {
+	case err := <-errChan:
+		if err != nil {
+			return nil, fmt.Errorf("error starting preview: %w", err)
+		}
+	case <-ctx.Done():
+		return recorder, &TimeoutError{Err: ctx.Err()}
```
anhdle-sso, line 193: When `<-ctx.Done()` fires, `runKCCManagerPreview` returns immediately. However, the `instance.Start(ctx)` goroutine is still running in the background while it gracefully shuts down. 

Since `Execute` proceeds to immediately read `recorder` state (such as `RemainResourcesCount` and the `events` slice), these reads will silently race against the controllers still mutating them, causing a `panic: data race` or corrupted memory out-of-bounds slice access.

To safely return the partial results, you must wait for the manager to finish shutting down before returning the `recorder`. For example:

pkg/cli/cmd/preview/execute.go
```diff
@@ -67,44 +68,45 @@ func Execute(ctx context.Context, opts *Options) error {
...
 		UpstreamGCPBurst:         opts.GCPBurst,
 		Namespace:                opts.Namespace,
-	})
-	if err != nil {
-		return fmt.Errorf("building preview instance: %w", err)
-	}
-	// TODO: Consider if 0 means no timeout.
-	if opts.Timeout == 0 {
-		opts.Timeout = defaultTimeout
 	}
```
anhdle-sso, line 79: Directly accessing `defaultRunRecorder.RemainResourcesCount` is a data race. The background `PreviewInstance` is still running when the timeout hits and may concurrently mutate this field. You should add a thread-safe getter (e.g. `recorder.GetRemainResourcesCount()`) that explicitly locks `reconcileTrackerMutex`.

pkg/cli/cmd/preview/execute.go
```diff
@@ -67,44 +68,45 @@ func Execute(ctx context.Context, opts *Options) error {
...
+			return fmt.Errorf("error running KCC manager preview with default controller type: %w", err)
 		}
-	}()
-
-	errChan := make(chan error, 1)
-	go func() {
-		errChan <- previewInstance.Start(ctx)
-	}()
+	}
+	defaultRunResult := defaultRunRecorder.GetOrCreateReconciledResults()
```
anhdle-sso, line 96: Similar to the default run, accessing `alternativeRunRecorder.RemainResourcesCount` without the mutex is a data race while the background manager finishes shutting down.

pkg/cli/cmd/preview/execute.go
```diff
@@ -67,44 +68,45 @@ func Execute(ctx context.Context, opts *Options) error {
 	if err != nil {
 		return fmt.Errorf("error building GCP authorization: %w", err)
 	}
-	previewInstance, err := corepreview.NewPreviewInstance(recorder, corepreview.PreviewInstanceOptions{
+	instanceOptions := corepreview.PreviewInstanceOptions{
 		UpstreamRESTConfig:       upstreamRESTConfig,
 		UpstreamGCPAuthorization: authorization,
 		UpstreamGCPHTTPClient:    nil,
 		UpstreamGCPQPS:           opts.GCPQPS,
```
anhdle-sso, line 74: The check `if opts.Timeout == 0 { opts.Timeout = defaultTimeout }` was removed here. If a user specifies `--timeout=0` (often used to disable a timeout), `runKCCManagerPreview` will receive `0` and create an instantly canceled context, causing the preview to fail immediately. If `0` means no timeout, handle it by using `context.Background()`.

pkg/cli/preview/recorder.go
```diff
@@ -416,11 +416,7 @@ func (r *Recorder) PreloadGKNN(ctx context.Context, config *rest.Config, namespa
 	return nil
 }
 
-func (recorder *Recorder) SummaryReport(summaryFile string) error {
-	return recorder.getOrCreateSummary().SummaryReport(summaryFile)
-}
-
-func (recorder *Recorder) getOrCreateSummary() *RecorderReconciledResults {
+func (recorder *Recorder) GetOrCreateReconciledResults() *RecorderReconciledResults {
```
anhdle-sso, line 428: `GenerateRecorderReconciledResults()` iterates over `r.objects` without acquiring `r.mutex`. Since `Execute` calls this immediately after a timeout while the manager is still running, it will result in a `panic: concurrent map iteration and map write`. You must add `r.mutex.Lock()` before iterating inside that method.

pkg/cli/preview/result.go
```diff
@@ -126,45 +124,122 @@ func (r *Recorder) GenerateRecorderReconciledResults() *RecorderReconciledResult
...
-				for _, result := range r.results[group][kind][namespace] {
-					reconcileStatus := formatReconciledStatus(result)
-					if result.ReconcileStatus == ReconcileStatusUnhealthy {
-						badResult = append(badResult, result)
+	fmt.Fprintln(w, "GROUP\tKIND\tNAME\tDEFAULT-CONTROLLER\tDEFAULT-RESULT\tDEFAULT-DIFFS\tALTERNATIVE-CONTROLLER\tALTERNATIVE-RESULT\tALTERNATIVE-DIFFS")
+	type resultPair struct {
+		def  *GKNNReconciledResult
+		alt  *GKNNReconciledResult
+		gknn GKNN
+	}
```
anhdle-sso, line 156: This 4-level nested map (`map[string]map[string]map[string]map[string]*resultPair`) adds unnecessary complexity and runtime allocation overhead. Since `GKNN` is natively comparable, you can drastically simplify this to a flat `map[GKNN]*resultPair`, eliminating the need to continuously allocate nested map layers in `addPair`.

anhdle-sso:
Refactored the implementation to address review feedback regarding concurrency and performance:

1. **Fixed Data Races**: Updated `runKCCManagerPreview` to wait for the manager's background goroutine to stop after a timeout, and added mutex protection to `Recorder` fields.
2. **Simplified Result Mapping**: Refactored the nested map in `CombinedSummaryReport` to a flat map keyed by `GKNN`, improving performance and simplifying the sorting logic.
3. **Improved Timeout Resilience**: Ensured `opts.Timeout` is correctly defaulted and added nil-safety to reporting methods.
4. **Added Unit Tests**: Included `TestCombinedSummaryReport` to verify the side-by-side report generation and deterministic result ordering.

PR #7088, opened 2026-03-18
Add CRUD testdata for MemorystoreInstanceBackup mocks

suwandim:
### BRIEF Change description

This implements the mocks related to backups for Memorystore Instances.

Needed for mocking MemorystoreInstanceBackup types.

```release-note
NONE
```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/mockgcp/mockmemorystore/instance.go
+++ b/mockgcp/mockmemorystore/instance.go
@@ -31,6 +31,7 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/common/projects"
 	pb "github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/generated/mockgcp/cloud/memorystore/v1"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/mocks"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
 )
 
 type instanceServer struct {
@@ -330,6 +331,140 @@ func (r *instanceServer) DeleteInstance(ctx context.Context, req *pb.DeleteInsta
 	})
 }
 
+func (r *instanceServer) GetBackup(ctx context.Context, req *pb.GetBackupRequest) (*pb.Backup, error) {
+	name, err := r.parseBackupName(req.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	fqn := name.String()
+
+	obj := &pb.Backup{}
+	if err := r.storage.Get(ctx, fqn, obj); err != nil {
+		if status.Code(err) == codes.NotFound {
+			return nil, status.Errorf(codes.NotFound, "Resource '%s' was not found", fqn)
+		}
+		return nil, err
+	}
+
+	retObj := proto.Clone(obj).(*pb.Backup)
+	return retObj, nil
+}
+
+func (r *instanceServer) BackupInstance(ctx context.Context, req *pb.BackupInstanceRequest) (*longrunning.Operation, error) {
+	instanceName, err := r.parseInstanceName(req.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	instanceFqn := instanceName.String()
+	instanceObj := &pb.Instance{}
+	if err := r.storage.Get(ctx, instanceFqn, instanceObj); err != nil {
+		if status.Code(err) == codes.NotFound {
+			return nil, status.Errorf(codes.NotFound, "Resource '%s' was not found", instanceFqn)
+		}
+		return nil, err
+	}
+
+	backupCollectionName := ""
+	if instanceObj.BackupCollection != nil {
+		backupCollectionName = *instanceObj.BackupCollection
+	} else {
+		backupCollectionName = fmt.Sprintf("projects/%s/locations/%s/backupCollections/backupCollection-%s", instanceName.Project.ID, instanceName.Location, instanceName.Name)
+		instanceObj.BackupCollection = direct.LazyPtr(backupCollectionName)
+		if err := r.storage.Update(ctx, instanceFqn, instanceObj); err != nil {
+			return nil, err
+		}
+	}
+
+	reqName := fmt.Sprintf("%s/backups/%s", backupCollectionName, req.GetBackupId())
+	name, err := r.parseBackupName(reqName)
+	if err != nil {
+		return nil, err
+	}
+
+	fqn := name.String()
+	now := time.Now()
+
+	obj := &pb.Backup{
+		BackupType: pb.Backup_ON_DEMAND,
+		BackupFiles: []*pb.BackupFile{
+			&pb.BackupFile{
+				FileName:   fmt.Sprintf("file-%s.rdb", req.GetBackupId()),
+				SizeBytes:  141,
+				CreateTime: timestamppb.New(now),
+			},
+		},
+		CreateTime:     timestamppb.New(now),
+		ExpireTime:     timestamppb.New(now),
+		EngineVersion:  instanceObj.EngineVersion,
+		Instance:       instanceName.String(),
+		InstanceUid:    instanceObj.Uid,
+		Name:           fqn,
+		NodeType:       instanceObj.NodeType,
+		ShardCount:     instanceObj.ShardCount,
+		State:          pb.Backup_ACTIVE,
+		TotalSizeBytes: 141,
+		Uid:            fmt.Sprintf("backup-%s", req.GetBackupId()),
+	}
+	if req.GetTtl() != nil {
+		ttl := now.Add(req.GetTtl().AsDuration())
+		obj.ExpireTime = timestamppb.New(ttl)
+	}
+	if err := r.storage.Create(ctx, fqn, obj); err != nil {
+		return nil, err
+	}
+
+	prefix := name.OperationPrefix()
+	metadata := &pb.OperationMetadata{
+		ApiVersion: "v1",
+		CreateTime: timestamppb.New(now),
+		Target:     instanceName.String(),
+		Verb:       "backup",
+	}
+
+	return r.operations.StartLRO(ctx, prefix, metadata, func() (proto.Message, error) {
+		metadata.EndTime = timestamppb.Now()
+		return proto.Clone(instanceObj).(*pb.Instance), nil
+	})
+}
+
+func (r *instanceServer) DeleteBackup(ctx context.Context, req *pb.DeleteBackupRequest) (*longrunning.Operation, error) {
+	name, err := r.parseBackupName(req.Name)
+	if err != nil {
+		return nil, err
+	}
+	fqn := name.String()
+
+	now := time.Now()
+
+	obj := &pb.Backup{}
+
+	if err := r.storage.Get(ctx, fqn, obj); err != nil {
+		if status.Code(err) == codes.NotFound {
+			return nil, status.Errorf(codes.NotFound, "Resource '%s' was not found", fqn)
+		}
+		return nil, err
+	}
+
+	deletedObj := &pb.Backup{}
+	if err := r.storage.Delete(ctx, fqn, deletedObj); err != nil {
+		return nil, err
+	}
+
+	metadata := &pb.OperationMetadata{
+		ApiVersion: "v1",
+		CreateTime: timestamppb.New(now),
+		Target:     fqn,
+		Verb:       "delete",
+	}
+	prefix := fmt.Sprintf("projects/%s/locations/%s", name.Project.ID, name.Location)
+	return r.operations.StartLRO(ctx, prefix, metadata, func() (proto.Message, error) {
+		metadata.EndTime = timestamppb.Now()
+		return &emptypb.Empty{}, nil
+	})
+}
+
 type instanceName struct {
 	Project  *projects.ProjectData
 	Location string
@@ -388,5 +523,48 @@ func (s *instanceServer) parseNetworkName(name string) (*networkName, error) {
 			Name:    tokens[4],
 		}, nil
 	}
+	return nil, status.Errorf(codes.InvalidArgument, "name %q is not valid", name)
+}
+
+type backupName struct {
+	Project          *projects.ProjectData
+	Location         string
+	BackupCollection string
+	Name             string
+}
+
+func (n *backupName) String() string {
+	return fmt.Sprintf("%s/backups/%s", n.Parent(), n.Name)
+}
+
+func (n *backupName) Parent() string {
+	return fmt.Sprintf("%s/backupCollections/%s", n.OperationPrefix(), n.BackupCollection)
+}
+
+func (n *backupName) OperationPrefix() string {
+	return fmt.Sprintf("projects/%s/locations/%s", n.Project.ID, n.Location)
+}
+
+// parseBackupName parses a string into a backup name.
+// The expected form is `projects/*/locations/*/backupCollections/*/backups/*`.
+func (r *instanceServer) parseBackupName(name string) (*backupName, error) {
+	tokens := strings.Split(name, "/")
+
+	if len(tokens) == 8 && tokens[0] == "projects" && tokens[2] == "locations" && tokens[4] == "backupCollections" && tokens[6] == "backups" {
+		project, err := r.Projects.GetProjectByID(tokens[1])
+		if err != nil {
+			return nil, err
+		}
+
+		name := &backupName{
+			Project:          project,
+			Location:         tokens[3],
+			BackupCollection: tokens[5],
+			Name:             tokens[7],
+		}
+
+		return name, nil
+	}
+
 	return nil, status.Errorf(codes.InvalidArgument, "name %q is not valid", name)
 }
--- a/mockgcp/mockmemorystore/instance.go
+++ b/mockgcp/mockmemorystore/instance.go
@@ -17,7 +17,6 @@ package mockmemorystore
 import (
 	"context"
 	"fmt"
-	"math/rand/v2"
 	"strings"
 	"time"
 
@@ -73,8 +72,8 @@ func (r *instanceServer) CreateInstance(ctx context.Context, req *pb.CreateInsta
 	obj := proto.Clone(req.GetInstance()).(*pb.Instance)
 	obj.Name = fqn
 	obj.CreateTime = timestamppb.New(now)
-
-	obj.State = pb.Instance_CREATING
+	obj.UpdateTime = timestamppb.New(now)
+	obj.State = pb.Instance_ACTIVE
 
 	if err := r.populateDefaultsForInstance(name, obj); err != nil {
 		return nil, err
@@ -86,7 +85,7 @@ func (r *instanceServer) CreateInstance(ctx context.Context, req *pb.CreateInsta
 
 	updatedObj := proto.Clone(obj).(*pb.Instance)
 	updatedObj.CreateTime = nil
-	prefix := fmt.Sprintf("projects/%d/locations/%s", name.Project.Number, name.Location)
+	prefix := fmt.Sprintf("projects/%s/locations/%s", name.Project.ID, name.Location)
 
 	metadata := &pb.OperationMetadata{
 		ApiVersion: "v1",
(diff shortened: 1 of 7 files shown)


mockgcp/mockmemorystore/instance.go
```diff
@@ -330,6 +437,140 @@ func (r *instanceServer) DeleteInstance(ctx context.Context, req *pb.DeleteInsta
...
+		}
+		return nil, err
+	}
+
+	backupCollectionName := ""
+	if instanceObj.BackupCollection != nil {
+		backupCollectionName = *instanceObj.BackupCollection
+	} else {
+		backupCollectionName = fmt.Sprintf("projects/%s/locations/%s/backupCollections/backupCollection-%s", instanceName.Project.ID, instanceName.Location, instanceName.Name)
+		instanceObj.BackupCollection = direct.LazyPtr(backupCollectionName)
```
anhdle-sso: Instead of importing `github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct` into MockGCP, you can use the `mocks.PtrTo` function to create the pointer.
It is already imported in this file and used elsewhere:
`instanceObj.BackupCollection = mocks.PtrTo(backupCollectionName)`

mockgcp/mockmemorystore/instance.go
```diff
@@ -330,6 +437,140 @@ func (r *instanceServer) DeleteInstance(ctx context.Context, req *pb.DeleteInsta
...
+	metadata := &pb.OperationMetadata{
+		ApiVersion: "v1",
+		CreateTime: timestamppb.New(now),
+		Target:     instanceName.String(),
+		Verb:       "backup",
+	}
+
+	return r.operations.StartLRO(ctx, prefix, metadata, func() (proto.Message, error) {
+		metadata.EndTime = timestamppb.Now()
+		return proto.Clone(instanceObj).(*pb.Instance), nil
```
anhdle-sso, line 559: The `BackupInstance` operation creates a `Backup` object, but the LRO response here returns the cloned `Instance` object (`proto.Clone(instanceObj).(*pb.Instance)`).
According to the Memorystore API, the operation response should contain the newly created `Backup` object. Please verify if this should return `proto.Clone(obj).(*pb.Backup)` instead.

mockgcp/mockmemorystore/instance.go
```diff
@@ -241,22 +334,34 @@ func (r *instanceServer) UpdateInstance(ctx context.Context, req *pb.UpdateInsta
...
-			obj.Labels = req.Instance.Labels
-		case "nodeType":
-			obj.NodeType = req.Instance.NodeType
+		case "maintenancePolicy":
+			obj.MaintenancePolicy = req.Instance.MaintenancePolicy
+		case "automatedBackupConfig":
+			obj.AutomatedBackupConfig = req.Instance.AutomatedBackupConfig
+		case "crossInstanceReplicationConfig":
+			obj.CrossInstanceReplicationConfig = req.Instance.CrossInstanceReplicationConfig
+		case "gcsBucket":
```
anhdle-sso: The API protobuf field for GCS bucket imports is `gcs_source` (which uses the JSON tag `gcsSource`). The update mask path mapped by MockGCP should correspond to the API schema and use `gcsSource` (camelCase) rather than `gcsBucket`.

mockgcp/mockmemorystore/instance.go
```diff
@@ -72,8 +72,8 @@ func (r *instanceServer) CreateInstance(ctx context.Context, req *pb.CreateInsta
 	obj := proto.Clone(req.GetInstance()).(*pb.Instance)
 	obj.Name = fqn
 	obj.CreateTime = timestamppb.New(now)
-
-	obj.State = pb.Instance_CREATING
+	obj.UpdateTime = timestamppb.New(now)
+	obj.State = pb.Instance_ACTIVE
```
anhdle-sso: Changing the initial state from `CREATING` to `ACTIVE` directly may cause MockGCP to diverge from real GCP behavior.
The real Memorystore API typically returns the `CREATING` state in the initial response and transitions to `ACTIVE` only after the long-running operation completes.
If tests rely on the exact initial state (or to avoid breaking golden test matches with real GCP traffic), consider keeping `CREATING` here. The `StartLRO` callback immediately below already handles setting the final state to `ACTIVE`.

mockgcp/mockmemorystore/instance.go
```diff
@@ -267,6 +372,8 @@ func (r *instanceServer) UpdateInstance(ctx context.Context, req *pb.UpdateInsta
 		return nil, err
 	}
 
+	obj.State = pb.Instance_ACTIVE
```
anhdle-sso: Similar to `CreateInstance`, setting `obj.State = pb.Instance_ACTIVE` here overrides the `pb.Instance_UPDATING` state set earlier in this function. This means the object is marked `ACTIVE` synchronously before the LRO is executed. Consider removing this line to let the LRO callback handle the state transition to `ACTIVE`, which better simulates the real API.

mockgcp/mockmemorystore/normalize.go
```diff
@@ -21,8 +21,12 @@ import (
 var _ mockgcpregistry.SupportsNormalization = &MockService{}
 
 func (s *MockService) ConfigureVisitor(url string, replacements mockgcpregistry.NormalizingVisitor) {
-	replacements.ReplacePath(".pscAttachmentDetails[].serviceAttachment", "projects/1234567890/regions/us-central1/serviceAttachments/sampleServiceAttachment")
-	replacements.ReplacePath(".response.pscAttachmentDetails[].serviceAttachment", "projects/1234567890/regions/us-central1/serviceAttachments/sampleServiceAttachment")
+	replacements.ReplacePath(".backupFiles[].createTime", "2024-04-01T12:34:56.123456Z")
+	replacements.ReplacePath(".crossInstanceReplicationConfig.updateTime", "2024-04-01T12:34:56.123456Z")
+	replacements.ReplacePath(".response.backupFiles[].createTime", "2024-04-01T12:34:56.123456Z")
+	replacements.ReplacePath(".response.crossInstanceReplicationConfig.updateTime", "2024-04-01T12:34:56.123456Z")
+	replacements.ReplacePath(".status.observedState.backupFiles[].createTime", "2024-04-01T12:34:56.123456Z")
```
anhdle-sso: The `.status.observedState` paths are part of the KRM object schema but do not exist in the raw GCP API HTTP responses. The `ConfigureVisitor` substitutions operate exclusively on the raw HTTP JSON traffic (`_http.log`). Therefore, this replacement (and the one on the next line) will not match anything and can be safely removed.

mockgcp/mockmemorystore/instance.go
```diff
@@ -330,6 +437,140 @@ func (r *instanceServer) DeleteInstance(ctx context.Context, req *pb.DeleteInsta
...
+		backupCollectionName = *instanceObj.BackupCollection
+	} else {
+		backupCollectionName = fmt.Sprintf("projects/%s/locations/%s/backupCollections/backupCollection-%s", instanceName.Project.ID, instanceName.Location, instanceName.Name)
+		instanceObj.BackupCollection = direct.LazyPtr(backupCollectionName)
+		if err := r.storage.Update(ctx, instanceFqn, instanceObj); err != nil {
+			return nil, err
+		}
+	}
+
+	reqName := fmt.Sprintf("%s/backups/%s", backupCollectionName, req.GetBackupId())
```
anhdle-sso: The API definition notes that if `backup_id` is omitted in the request, the API generates a default ID (e.g., `[YYYYMMDDHHMMSS]_[Shortened Instance UID]`). 
Since `req.GetBackupId()` might be empty, you should check and generate a default ID here. Otherwise, `parseBackupName` might receive an invalid string with a trailing slash or parse an empty string as the name.

mockgcp/mockmemorystore/instance.go
```diff
@@ -118,21 +118,68 @@ func (s *instanceServer) populateDefaultsForInstance(name *instanceName, obj *pb
 		obj.NodeType = pb.Instance_HIGHMEM_MEDIUM
 	}
 
-	if obj.Endpoints == nil {
-		pscConnectionID := time.Now().UnixNano()
-
+	if obj.Mode == pb.Instance_MODE_UNSPECIFIED {
+		obj.Mode = pb.Instance_CLUSTER
+	}
+	if obj.PscAttachmentDetails == nil {
```
anhdle-sso: If `obj.PscAttachmentDetails` is sent as an empty slice (e.g., `[]` in the JSON payload), this `nil` check will be bypassed and the subsequent array access `attachmentDetails := obj.PscAttachmentDetails[i%2]` will panic. Consider checking `if len(obj.PscAttachmentDetails) == 0` instead to ensure the slice has elements before assuming it is safe to skip default generation.

mockgcp/mockmemorystore/instance.go
```diff
@@ -118,21 +118,68 @@ func (s *instanceServer) populateDefaultsForInstance(name *instanceName, obj *pb
...
+						},
+					})
+				}
+			}
+		}
+		pscConnectionID := int64(1234567890123456789)
 		for _, endpoint := range obj.Endpoints {
-			for _, connections := range endpoint.Connections {
+			for i, connections := range endpoint.Connections {
+				attachmentDetails := obj.PscAttachmentDetails[i%2]
```
anhdle-sso, line 157: If `len(obj.PscAttachmentDetails)` is exactly 1, `obj.PscAttachmentDetails[i%2]` will panic with an out-of-bounds index when `i` evaluates to an odd number (such as `1`). It is safer to use `i % len(obj.PscAttachmentDetails)` after verifying the slice length is greater than 0.

mockgcp/mockmemorystore/instance.go
```diff
@@ -199,6 +239,59 @@ func (s *instanceServer) populateDefaultsForInstance(name *instanceName, obj *pb
...
+					return err
+				}
+				secondaryInstance.Uid = fmt.Sprintf("instance-%s", secondaryName.Name)
+				crr.Membership.SecondaryInstances = append(crr.Membership.SecondaryInstances, &pb.CrossInstanceReplicationConfig_RemoteInstance{
+					Instance: secondaryInstance.Instance,
+					Uid:      secondaryInstance.Uid,
+				})
+			}
+		case pb.CrossInstanceReplicationConfig_SECONDARY:
+			primaryName, err := s.parseInstanceName(crr.PrimaryInstance.Instance)
```
anhdle-sso, line 277: If `crr.PrimaryInstance` is nil when `InstanceRole` is `SECONDARY` (e.g., if it is omitted from the request), accessing `crr.PrimaryInstance.Instance` will cause a panic. Please add a nil check for `crr.PrimaryInstance` before accessing its fields and return a `status.Errorf(codes.InvalidArgument, ...)` error if it is missing.

mockgcp/mockmemorystore/instance.go
```diff
@@ -330,6 +437,140 @@ func (r *instanceServer) DeleteInstance(ctx context.Context, req *pb.DeleteInsta
...
+		BackupType: pb.Backup_ON_DEMAND,
+		BackupFiles: []*pb.BackupFile{
+			&pb.BackupFile{
+				FileName:   fmt.Sprintf("file-%s.rdb", req.GetBackupId()),
+				SizeBytes:  141,
+				CreateTime: timestamppb.New(now),
+			},
+		},
+		CreateTime:     timestamppb.New(now),
+		ExpireTime:     timestamppb.New(now),
```
anhdle-sso: Setting `ExpireTime` to `now` by default implies that the backup expires immediately upon creation if a TTL is not provided. It is safer to omit `ExpireTime` entirely (or use a default standard expiration, if one exists) unless the `req.GetTtl() != nil` condition is satisfied below.

mockgcp/mockmemorystore/instance.go
suwandim: Will fix.

mockgcp/mockmemorystore/instance.go
suwandim: Will fix.

mockgcp/mockmemorystore/instance.go
suwandim: Will fix.

mockgcp/mockmemorystore/instance.go
suwandim: If you looked at the actual GCP responses recorded while executing the `scripts.yaml` (see https://github.com/suwandim/k8s-config-connector/blob/efe876431f2225e97efa300b06880013f23299d6/mockgcp/mockmemorystore/testdata/backup/crud/_http.log#L37-L109), the Operation returned an Instance type instead of Backup (see https://github.com/suwandim/k8s-config-connector/blob/efe876431f2225e97efa300b06880013f23299d6/mockgcp/mockmemorystore/testdata/backup/crud/_http.log#L63-L65). So the mocks implementation matches the real server.

mockgcp/mockmemorystore/instance.go
suwandim: Will fix.

mockgcp/mockmemorystore/normalize.go
suwandim: WIll fix.

mockgcp/mockmemorystore/instance.go
suwandim: Will fix.

mockgcp/mockmemorystore/instance.go
suwandim: Will fix.

mockgcp/mockmemorystore/instance.go
suwandim: `PscAttachmentDetails` is an output only field (see https://github.com/googleapis/googleapis/blob/c063f49852ce7fc08c722c87f3a6f676209985a2/google/cloud/memorystore/v1/memorystore.proto#L528-L530) so the value is always an empty slice.

The mock is always making sure that the field will be set to a slice containing two items.

mockgcp/mockmemorystore/instance.go
suwandim: `PscAttachmentDetails` is an output only field (see https://github.com/googleapis/googleapis/blob/c063f49852ce7fc08c722c87f3a6f676209985a2/google/cloud/memorystore/v1/memorystore.proto#L528-L530) so the value is always an empty slice.

The mock is always making sure that the field will be set to a slice containing two items.

mockgcp/mockmemorystore/instance.go
suwandim: Will fix.

maqiuyujoyce:
Just to double check, does commit efe876431f2225e97efa300b06880013f23299d6 contain the output of running the test against the real GCP?

mockgcp/mockmemorystore/testdata/backup/crud/_http.log
```diff
@@ -67,13 +64,9 @@ X-Xss-Protection: 0
     "automatedBackupConfig": {
       "automatedBackupMode": "DISABLED"
     },
-    "backupCollection": "projects/${projectId}/locations/us-central1/backupCollections/ee37999e-2b25-44b8-a151-b994a309b58f",
+    "backupCollection": "projects/${projectId}/locations/us-central1/backupCollections/backupCollection-test-instance-${uniqueId}",
     "createTime": "2024-04-01T12:34:56.123456Z",
     "deletionProtectionEnabled": false,
-    "effectiveMaintenanceVersion": "MEMORYSTORE_20260217_02_00",
```
maqiuyujoyce: Can we add these two fields to the mock? Same for other similar diffs.

mockgcp/mockmemorystore/testdata/backup/crud/_http.log
```diff
@@ -87,17 +80,16 @@ X-Xss-Protection: 0
...
+        "serviceAttachment": "projects/tp-test-instance-${uniqueId}/regions/us-central1/serviceAttachments/sa-test-instance-${uniqueId}"
       },
       {
         "connectionType": "CONNECTION_TYPE_READER",
-        "serviceAttachment": "projects/1234567890/regions/us-central1/serviceAttachments/sampleServiceAttachment"
+        "serviceAttachment": "projects/tp-test-instance-${uniqueId}/regions/us-central1/serviceAttachments/sa-test-instance-${uniqueId}-2"
       }
     ],
     "replicaCount": 0,
-    "serverCaMode": "SERVER_CA_MODE_UNSPECIFIED",
```
maqiuyujoyce: ditto

mockgcp/mockmemorystore/testdata/backup/crud/_http.log
```diff
@@ -87,17 +80,16 @@ X-Xss-Protection: 0
...
     "shardCount": 1,
-    "state": "ACTIVE",
```
maqiuyujoyce: Could you align the state value in mock?

pkg/test/resourcefixture/testdata/basic/memorystore/v1beta1/memorystoreinstance/fullmemorystoreinstance/_generated_object_fullmemorystoreinstance.golden.yaml
```diff
@@ -58,7 +58,19 @@ status:
     createTime: "1970-01-01T00:00:00Z"
     endpoints:
     - connections:
-      - pscAutoConnection: {}
```
maqiuyujoyce, line 61: As the mock change update the golden object, could you run the fullmemeorystoreinstance test against real GCP and validate the golden object change here?

mockgcp/mockmemorystore/testdata/backup/crud/_http.log
suwandim: `effectiveMaintenanceVersion` field is not available at the googleapis repo imported (see [`731d7f2`](https://github.com/googleapis/googleapis/blob/731d7f2ab6e4e2ea15030c95039e2cb66174d4fb/google/cloud/memorystore/v1/memorystore.proto)) by the mocks

mockgcp/mockmemorystore/testdata/backup/crud/_http.log
suwandim: `serverCaMode` field is not available at the googleapis repo imported (see [`731d7f2`](https://github.com/googleapis/googleapis/blob/731d7f2ab6e4e2ea15030c95039e2cb66174d4fb/google/cloud/memorystore/v1/memorystore.proto)) by the mocks.

It is not available even in the googleapis repo [master at HEAD](https://github.com/googleapis/googleapis/blob/master/google/cloud/memorystore/v1/memorystore.proto).

mockgcp/mockmemorystore/testdata/backup/crud/_http.log
suwandim: Will do.

pkg/test/resourcefixture/testdata/basic/memorystore/v1beta1/memorystoreinstance/fullmemorystoreinstance/_generated_object_fullmemorystoreinstance.golden.yaml
suwandim: Done.

See

-  [`./hack/record-gcp fixtures/memorystore`](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/7088/changes/4421946a1271b99932cfad6c6c526d20d52bd698)
-  [`./hack/compare-mock fixtures/memorystore`](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/7088/changes/c98be66305153d59e28fad4989f23c42d31e400a)

mockgcp/mockmemorystore/instance.go
suwandim: Also, the [BackupInstance() definition](https://github.com/googleapis/googleapis/blob/731d7f2ab6e4e2ea15030c95039e2cb66174d4fb/google/cloud/memorystore/v1/memorystore.proto#L216-L228) defines the LRO response type to be `Instance`.

```proto
  rpc BackupInstance(BackupInstanceRequest)
      returns (google.longrunning.Operation) {
    option (google.api.http) = {
      post: "/v1/{name=projects/*/locations/*/instances/*}:backup"
      body: "*"
    };
    option (google.api.method_signature) = "name";
    option (google.longrunning.operation_info) = {
      response_type: "Instance"
      metadata_type: "OperationMetadata"
    };
  }
}
```

mockgcp/mockmemorystore/testdata/backup/crud/_http.log
```diff
@@ -231,48 +214,12 @@ X-Frame-Options: SAMEORIGIN
...
-    "target": "projects/${projectId}/locations/us-central1/backupCollections/ee37999e-2b25-44b8-a151-b994a309b58f/backups/test-backup-${uniqueId}",
+    "target": "projects/${projectId}/locations/us-central1/backupCollections/backupCollection-test-instance-${uniqueId}/backups/test-backup-${uniqueId}",
     "verb": "delete"
   },
   "name": "projects/${projectId}/locations/us-central1/operations/${operationID}"
 }
-
----
-
-GET https://memorystore.googleapis.com/v1/projects/${projectId}/locations/us-central1/operations/${operationID}?alt=json
```
maqiuyujoyce: Could you ensure the delete operation also exist in mockgcp?

mockgcp/mockmemorystore/testdata/backup/crud/_http.log
```diff
@@ -67,13 +64,9 @@ X-Xss-Protection: 0
     "automatedBackupConfig": {
       "automatedBackupMode": "DISABLED"
     },
-    "backupCollection": "projects/${projectId}/locations/us-central1/backupCollections/ee37999e-2b25-44b8-a151-b994a309b58f",
```
maqiuyujoyce: Nit: backupCollection ID should be normalized for cleaner comparison.

Not a blocker for this PR, but will be a blocker when MemorystoreBackup is promoted to beta.

suwandim: Can you point out how and where to do the normalization?

The value of the backupCollection is randomly generated in the real service.

Just performing blind string replacement for `"response.backupCollection"` does not work because the locations won't always be `us-central1`

pkg/test/resourcefixture/testdata/basic/memorystore/v1beta1/memorystoreinstance/fullmemorystoreinstance/_http.log
```diff
@@ -674,6 +656,10 @@ X-Xss-Protection: 0
       "maxmemory-policy": "volatile-ttl"
     },
     "engineVersion": "VALKEY_7_2",
+    "labels": {
```
maqiuyujoyce, line 659: Interesting that labels is not returned in the "response" subfield of the response to GET operation.

Not a blocker. Just an interesting diff to note and fix separately.

maqiuyujoyce (approved):
/lgtm

PR #7140, opened 2026-03-22
Added controller and fixture tests for Parameter Manager Parameter Version

### LINKED ISSUE #5786: Add support of new resource ParameterManagerParameterVersion in KCC
### Checklist

The goal of this ticket is to add the support of parameter versions for parameter created in google cloud using KCC

### Additional information

Docs: https://docs.cloud.google.com/secret-manager/parameter-manager/docs/view-parameter-version-details

Go reference: https://pkg.go.dev/cloud.google.com/go/parametermanager/apiv1/parametermanagerpb#ParameterVersion

Google APIs proto: https://github.com/googleapis/googleapis/tree/master/google/cloud/parametermanager/v1

### Importance

_No response_

dhavalbhensdadiya-crest:
### BRIEF Change description

Added the controller and fixture tests.
Fixes #5786 

This change adds the support of new resource parameter manager version.

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/apis/parametermanager/v1alpha1/parameterversion_types.go
+++ b/apis/parametermanager/v1alpha1/parameterversion_types.go
@@ -23,8 +23,9 @@ var ParameterManagerParameterVersionGVK = GroupVersion.WithKind("ParameterManage
 
 // +kcc:proto=google.cloud.parametermanager.v1.ParameterVersionPayload
 type ParameterVersionPayload struct {
-	// Required. bytes data for storing payload.
+	// Required. bytes data for storing payload. Provide base64-encoded value
 	// +kcc:proto:field=google.cloud.parametermanager.v1.ParameterVersionPayload.data
+	// +kubebuilder:validation:Format=byte
 	// +required
 	Data []byte `json:"data,omitempty"`
 }
--- a/pkg/controller/direct/parametermanager/parameterversion_controller.go
+++ b/pkg/controller/direct/parametermanager/parameterversion_controller.go
@@ -65,9 +65,9 @@ func (m *modelParameterVersion) client(ctx context.Context, location string) (*g
 
 	gcpClient, err := gcp.NewRESTClient(ctx, opts...)
 	if err != nil {
-		return nil, fmt.Errorf("building Parameter client: %w", err)
+		return nil, fmt.Errorf("building Parameter Version client: %w", err)
 	}
-	return gcpClient, err
+	return gcpClient, nil
 }
 
 func (m *modelParameterVersion) AdapterForObject(ctx context.Context, op *directbase.AdapterForObjectOperation) (directbase.Adapter, error) {
@@ -119,7 +119,7 @@ var _ directbase.Adapter = &ParameterVersionAdapter{}
 // Return a non-nil error requeues the requests.
 func (a *ParameterVersionAdapter) Find(ctx context.Context) (bool, error) {
 	log := klog.FromContext(ctx)
-	log.V(2).Info("getting Parameter Version", "name", a.id)
+	log.V(2).Info("getting Parameter Version", "name", a.id.String())
 
 	req := &parametermanagerpb.GetParameterVersionRequest{Name: a.id.String()}
 	parameterVersionPb, err := a.gcpClient.GetParameterVersion(ctx, req)
@@ -127,17 +127,17 @@ func (a *ParameterVersionAdapter) Find(ctx context.Context) (bool, error) {
 		if direct.IsNotFound(err) {
 			return false, nil
 		}
-		return false, fmt.Errorf("getting Parameter Version %q: %w", a.id, err)
+		return false, fmt.Errorf("getting Parameter Version %q: %w", a.id.String(), err)
 	}
 
 	a.actual = parameterVersionPb
 	return true, nil
 }
 
-// Create creates the resource in GCP based on `spec` and update the Config Connector object `status` based on the GCP response.
+// Create creates the resource in GCP based on `spec` and updates the Config Connector object `status` based on the GCP response.
 func (a *ParameterVersionAdapter) Create(ctx context.Context, createOp *directbase.CreateOperation) error {
 	log := klog.FromContext(ctx)
-	log.V(2).Info("creating Parameter Version", "name", a.id)
+	log.V(2).Info("creating Parameter Version", "name", a.id.String())
 	mapCtx := &direct.MapContext{}
 
 	desired := a.desired.DeepCopy()
@@ -153,9 +153,9 @@ func (a *ParameterVersionAdapter) Create(ctx context.Context, createOp *directba
 	}
 	created, err := a.gcpClient.CreateParameterVersion(ctx, req)
 	if err != nil {
-		return fmt.Errorf("creating Parameter Version %s: %w", a.id, err)
+		return fmt.Errorf("creating Parameter Version %s: %w", a.id.String(), err)
 	}
-	log.V(2).Info("successfully created Parameter Version", "name", a.id)
+	log.V(2).Info("successfully created Parameter Version", "name", a.id.String())
 
 	status := &krm.ParameterManagerParameterVersionStatus{}
 	status.ObservedState = ParameterManagerParameterVersionObservedState_FromProto(mapCtx, created)
@@ -166,10 +166,10 @@ func (a *ParameterVersionAdapter) Create(ctx context.Context, createOp *directba
 	return createOp.UpdateStatus(ctx, status, nil)
 }
 
-// Update updates the resource in GCP based on `spec` and update the Config Connector object `status` based on the GCP response.
+// Update updates the resource in GCP based on `spec` and updates the Config Connector object `status` based on the GCP response.
 func (a *ParameterVersionAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOperation) error {
 	log := klog.FromContext(ctx)
-	log.V(2).Info("updating Parameter Version", "name", a.id)
+	log.V(2).Info("updating Parameter Version", "name", a.id.String())
 	mapCtx := &direct.MapContext{}
 	desired := a.desired.DeepCopy()
 	resource := ParameterManagerParameterVersionSpec_ToProto(mapCtx, &desired.Spec)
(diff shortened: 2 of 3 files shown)


apis/parametermanager/v1alpha1/parameterversion_types.go
```diff
@@ -21,6 +21,14 @@ import (
 
 var ParameterManagerParameterVersionGVK = GroupVersion.WithKind("ParameterManagerParameterVersion")
 
+// +kcc:proto=google.cloud.parametermanager.v1.ParameterVersionPayload
+type ParameterVersionPayload struct {
+	// Required. bytes data for storing payload.
+	// +kcc:proto:field=google.cloud.parametermanager.v1.ParameterVersionPayload.data
+	// +required
```
maqiuyujoyce: Nit: As the codebot suggested, it'd be helpful if we add the comment instructing users to provide a base64-encoded value here. Otherwise the parse will fail.

Not a blocker.

pkg/controller/direct/parametermanager/parameterversion_controller.go
```diff
@@ -0,0 +1,283 @@
...
+
+	// Add regional endpoint if region is specified
+	if location != "" && location != "global" {
+		endpoint := fmt.Sprintf("parametermanager.%s.rep.googleapis.com:443", location)
+		opts = append(opts, option.WithEndpoint(endpoint))
+	}
+
+	gcpClient, err := gcp.NewRESTClient(ctx, opts...)
+	if err != nil {
+		return nil, fmt.Errorf("building Parameter client: %w", err)
```
maqiuyujoyce: Nit: The error message says 'building Parameter client', but this is for `ParameterManagerParameterVersion`. Consider changing to 'building Parameter Version client' or for better clarity.

pkg/controller/direct/parametermanager/parameterversion_controller.go
```diff
@@ -0,0 +1,283 @@
...
+
+var _ directbase.Adapter = &ParameterVersionAdapter{}
+
+// Find retrieves the GCP resource.
+// Return true means the object is found. This triggers Adapter `Update` call.
+// Return false means the object is not found. This triggers Adapter `Create` call.
+// Return a non-nil error requeues the requests.
+func (a *ParameterVersionAdapter) Find(ctx context.Context) (bool, error) {
+	log := klog.FromContext(ctx)
+	log.V(2).Info("getting Parameter Version", "name", a.id)
```
maqiuyujoyce: Logging `a.id` (which is a pointer to a struct) will print its Go-syntax representation instead of a clean string. Use `a.id.String()` to ensure the log message clearly prints the fully-qualified resource name.

Same for all the other `a.id` usage in the logs.

pkg/controller/direct/parametermanager/parameterversion_controller.go
```diff
@@ -0,0 +1,283 @@
...
+			return false, nil
+		}
+		return false, fmt.Errorf("getting Parameter Version %q: %w", a.id, err)
+	}
+
+	a.actual = parameterVersionPb
+	return true, nil
+}
+
+// Create creates the resource in GCP based on `spec` and update the Config Connector object `status` based on the GCP response.
```
maqiuyujoyce: Nit: Typo in the docstring: 'update' should be 'updates' to match the tense of 'creates'.

pkg/controller/direct/parametermanager/parameterversion_controller.go
```diff
@@ -0,0 +1,283 @@
...
+	status := &krm.ParameterManagerParameterVersionStatus{}
+	status.ObservedState = ParameterManagerParameterVersionObservedState_FromProto(mapCtx, created)
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+	status.ExternalRef = direct.LazyPtr(a.id.String())
+	return createOp.UpdateStatus(ctx, status, nil)
+}
+
+// Update updates the resource in GCP based on `spec` and update the Config Connector object `status` based on the GCP response.
```
maqiuyujoyce: Nit: Typo in the docstring: 'update' should be 'updates'.

pkg/controller/direct/parametermanager/parameterversion_controller.go
```diff
@@ -0,0 +1,283 @@
...
+func (a *ParameterVersionAdapter) Delete(ctx context.Context, deleteOp *directbase.DeleteOperation) (bool, error) {
+	log := klog.FromContext(ctx)
+	log.V(2).Info("deleting Parameter Version", "name", a.id)
+
+	req := &parametermanagerpb.DeleteParameterVersionRequest{Name: a.id.String()}
+	err := a.gcpClient.DeleteParameterVersion(ctx, req)
+	if err != nil {
+		if direct.IsNotFound(err) {
+			// Return success if not found (assume it was already deleted).
+			log.V(2).Info("skipping delete for non-existent Parameter Version, assuming it was already deleted", "name", a.id)
```
maqiuyujoyce: Use `a.id.String()` to avoid logging the pointer directly in the 'name' field.

apis/parametermanager/v1alpha1/parameterversion_types.go
maqiuyujoyce, line 29: Nit: As the codebot suggested, it'd be helpful if we add the comment instructing users to provide a base64-encoded value here. Otherwise the parse will fail.

Not a blocker.

pkg/controller/direct/parametermanager/parameterversion_controller.go
```diff
@@ -0,0 +1,283 @@
...
+	if location != "" && location != "global" {
+		endpoint := fmt.Sprintf("parametermanager.%s.rep.googleapis.com:443", location)
+		opts = append(opts, option.WithEndpoint(endpoint))
+	}
+
+	gcpClient, err := gcp.NewRESTClient(ctx, opts...)
+	if err != nil {
+		return nil, fmt.Errorf("building Parameter client: %w", err)
+	}
+	return gcpClient, err
```
maqiuyujoyce: Nit: Please return `gcpClient, nil` here for better readability.

pkg/controller/direct/parametermanager/parameterversion_controller.go
```diff
@@ -0,0 +1,283 @@
...
+	id, err := obj.GetIdentity(ctx, reader)
+	if err != nil {
+		return nil, err
+	}
+
+	parameter := id.(*krm.ParameterVersionIdentity).Parent()
+
+	location := parameter.Parent().Location
+
+	// Get parmetermanager GCP client
```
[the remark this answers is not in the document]
maqiuyujoyce: I think this is a good point.

pkg/controller/direct/parametermanager/parameterversion_controller.go
```diff
@@ -0,0 +1,283 @@
...
+
+	resource.Name = a.id.String()
+
+	paths, err := common.CompareProtoMessage(resource, a.actual, common.BasicDiff)
+	if err != nil {
+		return err
+	}
+
+	// If parameterversion is disabled, payload is not available for retrieval leading to mismatch
+	// For disabled parameterversion, ignore difference in payload
```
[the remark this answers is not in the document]
maqiuyujoyce: I think this is a good point.

Could a user update the payload even if it is disabled?

pkg/controller/direct/parametermanager/parameterversion_controller.go
```diff
@@ -0,0 +1,283 @@
...
+	if a.actual.GetDisabled() {
+		paths = paths.Delete("payload.data")
+	}
+
+	if len(paths) == 0 {
+		log.V(2).Info("no field needs update", "name", a.id)
+		if a.desired.Status.ExternalRef == nil {
```
maqiuyujoyce, line 195: I think the potentially cleaner / more readable code would be something like:
```
if len(paths) == 0 {
} else {
    ...
}
status := &krm.CloudDeployTargetStatus{}
status.ObservedState = CloudDeployTargetObservedState_FromProto(mapCtx, updated)
...
```
This is because observedState may be updated by the service.

Not blocking.

maqiuyujoyce (commented):
Thank you for supporting the controller! Overall lgtm. The only thing I want to double check is the update behavior of updating payload when the version is disabled but is about to be enabled.

Others can be addressed separately.

pkg/controller/direct/parametermanager/parameterversion_controller.go
dhavalbhensdadiya-crest: This is to handle the acquisition scenarios as mentioned here - https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/5893#discussion_r2785280021

dhavalbhensdadiya-crest:
> Thank you for supporting the controller! Overall lgtm. The only thing I want to double check is the update behavior of updating payload when the version is disabled but is about to be enabled.
> 
> Others can be addressed separately.

Hi @maqiuyujoyce ,
Please find the behaviour below:
1. Payload is an immutable field. If user tries to update payload when parameter version is enabled, it will fail and they will get an error that you are trying to update an immutable field
2. The reason for the check is that when parameter version is disabled, gcp does not return payload. Hence, payload would show up in the diff and needs to be removed to avoid the immutability error.

Hope this covers your question. If not, could you please give a step by step explanation of the user behaviour. Thanks

apis/parametermanager/v1alpha1/parameterversion_types.go
dhavalbhensdadiya-crest: Done

pkg/controller/direct/parametermanager/parameterversion_controller.go
maqiuyujoyce: Thanks for pointing back to my previous comment! So over time, we try to improve our code, and the current preferred way to handle the `len(path) == 0` case is what I shared in this comment.

As I said, this is not blocking, but more like a FYI for future similar work.

maqiuyujoyce (commented):
> * Payload is an immutable field. If user tries to update payload when parameter version is enabled, it will fail and they will get an error that you are trying to update an immutable field
> * The reason for the check is that when parameter version is disabled, gcp does not return payload. Hence, payload would show up in the diff and needs to be removed to avoid the immutability error.

Hi @dhavalbhensdadiya-crest thank you for the explanation. And I think the current implementation makes sense as it reflects user error if user updates the payload when parameter version is enabled.

Just want to get further clarification: If parameter version is enabled in the UpdateRequest, will the parameter version uses:
1. the passed in payload during the creation of the parameter version, or
2. the passed in payload in this particular UpdateRequest?

/lgtm

maqiuyujoyce (approved):
/lgtm

PR #7163, opened 2026-03-25
CloudDeployTarget: add direct controller 

katrielt:
This PR adds the direct controller for CloudDeployTarget. 

   - Implements TargetAdapter in pkg/controller/direct/clouddeploy/.
   - Refactors CloudDeployTargetIdentity to implement the identity.Identity interface and use the standardized parent.ProjectAndLocationParent.
   - Refactors CloudDeployTargetRef to implement the refs.Ref and refs.ExternalRef interfaces.
   - Enables the direct reconciler for CloudDeployTarget in pkg/controller/resourceconfig/static_config.go.
   - Adds a new fixture test target-basic with golden logs recorded from both real GCP and MockGCP.

  Addresses #6318

 ### WHY do we need this change?

To migrate CloudDeployTarget to the "direct" controller pattern, improving maintainability, reliability, and alignment with modern KCC architectural standards.

###  Special notes for your reviewer:

The ANALYSIS enum value in the usages field is currently present in the upstream GCP API but missing from our local generated protos and the CloudDeployTarget CRD. As a result, it appears in the real GCP logs but is omitted in  the mock logs. This will be addressed in a follow-up PR focused on type and proto regeneration.

  Tests you have done

   - [x] Run make ready-pr to ensure this PR is ready for review.
     - Verified the controller via the target-basic fixture test against both real GCP and MockGCP.


pkg/controller/direct/clouddeploy/target_controller.go
```diff
@@ -0,0 +1,285 @@
...
+
+	a.desiredPb.Name = a.id.String()
+
+	paths := make(sets.Set[string])
+	var err error
+
+	// etag and executionConfigs are server-generated and not in KCC spec.
+	// We zero them out to avoid unnecessary drift.
+	// TODO: We should probably handle executionConfigs better if they are actually in the spec
+	a.actual.Etag = ""
```
maqiuyujoyce: Instead of updating actual, we should either fill desired with the same values of these two fields or skip the diff when they show up in path.

During the Update() function, actual is SoT that should not be touched when possible.

pkg/controller/direct/clouddeploy/target_controller.go
```diff
@@ -0,0 +1,285 @@
...
+	a.actual.ExecutionConfigs = nil
+
+	paths, err = common.CompareProtoMessage(a.desiredPb, a.actual, common.BasicDiff)
+	if err != nil {
+		return err
+	}
+	if len(paths) == 0 {
+		log.V(2).Info("no field needs update", "name", a.id)
+		return nil
```
maqiuyujoyce: We need to make sure we handle acquisition use cases properly, so instead of directly return, we'll also want to set externalRef and observedState. Consider do something like:
```
if len(paths) == 0 {
} else {
    ...
}
status := &krm.CloudDeployTargetStatus{}
status.ObservedState = CloudDeployTargetObservedState_FromProto(mapCtx, updated)
...
```

pkg/controller/direct/clouddeploy/target_controller.go
```diff
@@ -0,0 +1,285 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	mapCtx := &direct.MapContext{}
+	desiredPb := CloudDeployTargetSpec_ToProto(mapCtx, &obj.Spec)
+	if mapCtx.Err() != nil {
+		return nil, mapCtx.Err()
+	}
+	desiredPb.Labels = label.NewGCPLabelsFromK8sLabels(u.GetLabels())
```
maqiuyujoyce: I checked the mapper and found that you also didn't support annotations field: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/clouddeploy/target_mapper.go#L30. Was that intentional?

pkg/controller/direct/clouddeploy/target_controller.go
```diff
@@ -102,7 +102,7 @@ func (m *modelTarget) AdapterForObject(ctx context.Context, op *directbase.Adapt
 func (m *modelTarget) AdapterForURL(ctx context.Context, url string) (directbase.Adapter, error) {
 	id := &krm.CloudDeployTargetIdentity{}
 	if err := id.FromExternal(url); err != nil {
-		return nil, err
+		return nil, nil
```
maqiuyujoyce: Right now most of the resource just do `return nil, nil` by default for AdapterForURL(). This will be refined separately in the future.

pkg/test/resourcefixture/testdata/basic/clouddeploy/v1alpha1/clouddeploytarget/target-basic/_http.log
```diff
@@ -16,7 +16,14 @@ X-Xss-Protection: 0
 {
   "error": {
     "code": 404,
-    "message": "Resource 'projects/${projectId}/locations/us-central1/targets/clouddeploytarget-${uniqueId}' was not found",
+    "errors": [
```
maqiuyujoyce: Could we match this?

pkg/test/resourcefixture/testdata/basic/clouddeploy/v1alpha1/clouddeploytarget/target-basic/_http.log
```diff
@@ -110,8 +114,7 @@ X-Xss-Protection: 0
           "DEPLOY",
           "VERIFY",
           "PREDEPLOY",
-          "POSTDEPLOY",
-          "ANALYSIS"
+          "POSTDEPLOY"
```
maqiuyujoyce, line 110: Similarly, can we match the value here?

pkg/controller/direct/clouddeploy/target_controller.go
```diff
@@ -0,0 +1,285 @@
...
+
+var _ directbase.Adapter = &TargetAdapter{}
+
+// Find retrieves the GCP resource.
+// Return true means the object is found. This triggers Adapter `Update` call.
+// Return false means the object is not found. This triggers Adapter `Create` call.
+// Return a non-nil error requeues the requests.
+func (a *TargetAdapter) Find(ctx context.Context) (bool, error) {
+	log := klog.FromContext(ctx)
+	log.V(2).Info("getting Target", "name", a.id)
```
maqiuyujoyce: Logging a.id (which is a pointer to a struct) will print its Go-syntax representation instead of a clean string. Use `a.id.String()` to ensure the log message clearly prints the fully-qualified resource name.

Same for all the other `a.id` usage in the logs.

pkg/controller/direct/clouddeploy/target_controller.go
katrielt: Fixed.

pkg/controller/direct/clouddeploy/target_controller.go
katrielt: Done

pkg/controller/direct/clouddeploy/target_controller.go
katrielt: Done.

pkg/test/resourcefixture/testdata/basic/clouddeploy/v1alpha1/clouddeploytarget/target-basic/_http.log
katrielt: Done.

pkg/controller/direct/clouddeploy/target_controller.go
katrielt: That was not intentional. Re-added.

pkg/controller/direct/clouddeploy/target_controller.go
katrielt: Fixed across the board.

pkg/test/resourcefixture/testdata/basic/clouddeploy/v1alpha1/clouddeploytarget/target-basic/_http.log
katrielt: ANALYSIS is missing from the generated protos, so MockGCP doesn't support it yet. I'll add it in a follow-up PR with re-generated real/mock logs.

pkg/test/resourcefixture/testdata/basic/clouddeploy/v1alpha1/clouddeploytarget/target-basic/_http.log
```diff
@@ -16,6 +16,13 @@ X-Xss-Protection: 0
 {
   "error": {
     "code": 404,
+    "errors": [
```
maqiuyujoyce: To remove the additional errors list, you'll probably want to do something similar to this: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/mockgcp/mockalloydb/service.go#L71

maqiuyujoyce (commented):
Mostly lgtm, just a comment about improvements in mock.

pkg/controller/direct/clouddeploy/target_controller.go
```diff
@@ -0,0 +1,277 @@
...
+	// This also ensures we don't send them in the PATCH request body if they are not in our spec.
+	paths, err := common.CompareProtoMessage(a.desiredPb, a.actual, func(fieldName protoreflect.Name, a, b proto.Message) (bool, error) {
+		if fieldName == "etag" || fieldName == "execution_configs" {
+			return false, nil
+		}
+		return common.BasicDiff(fieldName, a, b)
+	})
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
katrielt: Done

pkg/controller/direct/clouddeploy/target_controller.go
```diff
@@ -0,0 +1,277 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	mapCtx := &direct.MapContext{}
+	desiredPb := CloudDeployTargetSpec_ToProto(mapCtx, &obj.Spec)
+	if mapCtx.Err() != nil {
+		return nil, mapCtx.Err()
+	}
+	desiredPb.Labels = label.NewGCPLabelsFromK8sLabels(u.GetLabels())
```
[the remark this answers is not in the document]
katrielt: Done

pkg/controller/direct/clouddeploy/target_controller.go
```diff
@@ -0,0 +1,277 @@
...
+
+	return &TargetAdapter{
+		id:        id.(*krm.CloudDeployTargetIdentity),
+		gcpClient: gcpClient,
+		desiredPb: desiredPb,
```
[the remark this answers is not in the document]
katrielt: Done

pkg/controller/direct/clouddeploy/target_controller.go
```diff
@@ -0,0 +1,277 @@
...
+	status.ExternalRef = direct.LazyPtr(a.id.String())
+	return createOp.UpdateStatus(ctx, status, nil)
+}
+
+// Update updates the resource in GCP based on `spec` and update the Config Connector object `status` based on the GCP response.
+func (a *TargetAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOperation) error {
+	log := klog.FromContext(ctx)
+	log.V(2).Info("updating Target", "name", a.id.String())
+	mapCtx := &direct.MapContext{}
+
```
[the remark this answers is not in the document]
katrielt: Done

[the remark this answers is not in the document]
katrielt: Done

pkg/controller/direct/clouddeploy/target_controller.go
```diff
@@ -0,0 +1,277 @@
...
+	if a.actual == nil {
+		return nil, fmt.Errorf("Find() not called")
+	}
+	u := &unstructured.Unstructured{}
+
+	obj := &krm.CloudDeployTarget{}
+	mapCtx := &direct.MapContext{}
+	obj.Spec = direct.ValueOf(CloudDeployTargetSpec_FromProto(mapCtx, a.actual))
+	if mapCtx.Err() != nil {
+		return nil, mapCtx.Err()
```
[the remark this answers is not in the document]
katrielt: Done

apis/clouddeploy/v1alpha1/target_reference.go
```diff
@@ -42,42 +42,51 @@ type CloudDeployTargetRef struct {
...
 	if err != nil {
-		return "", fmt.Errorf("reading status.externalRef: %w", err)
+		return err
 	}
-	if actualExternalRef == "" {
-		return "", k8s.NewReferenceNotReadyError(u.GroupVersionKind(), key)
+	return identity.FromExternal(external)
+}
+
+func (r *CloudDeployTargetRef) ParseExternalToIdentity() (identity.Identity, error) {
```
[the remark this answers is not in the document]
katrielt: Done

pkg/controller/direct/clouddeploy/target_controller.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+	created, err := op.Wait(ctx)
+	if err != nil {
+		return fmt.Errorf("Target %s waiting creation: %w", a.id.String(), err)
+	}
+	log.V(2).Info("successfully created Target", "name", a.id.String())
+
+	status := &krm.CloudDeployTargetStatus{}
+	status.ObservedState = CloudDeployTargetObservedState_FromProto(mapCtx, created)
+	if mapCtx.Err() != nil {
```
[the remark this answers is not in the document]
katrielt: Done

pkg/controller/direct/clouddeploy/target_controller.go
```diff
@@ -0,0 +1,277 @@
...
+		for path := range paths {
+			report.AddField(path, nil, nil)
+		}
+		structuredreporting.ReportDiff(ctx, report)
+
+		updateMask := &fieldmaskpb.FieldMask{
+			Paths: sets.List(paths),
+		}
+
+		req := &clouddeploypb.UpdateTargetRequest{
```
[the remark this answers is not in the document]
katrielt: Done

pkg/controller/direct/clouddeploy/target_controller.go
```diff
@@ -0,0 +1,277 @@
...
+			// Return success if not found (assume it was already deleted).
+			log.V(2).Info("skipping delete for non-existent Target, assuming it was already deleted", "name", a.id.String())
+			return true, nil
+		}
+		return false, fmt.Errorf("deleting Target %s: %w", a.id.String(), err)
+	}
+
+	err = op.Wait(ctx)
+	if err != nil {
+		return false, fmt.Errorf("waiting delete Target %s: %w", a.id.String(), err)
```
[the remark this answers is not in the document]
katrielt: Done

pkg/controller/direct/clouddeploy/target_controller.go
```diff
@@ -0,0 +1,277 @@
...
+		return nil, mapCtx.Err()
+	}
+	// TODO: Handle Labels in Export if needed (it is usually handled by the generic exporter)
+
+	uObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
+	if err != nil {
+		return nil, err
+	}
+
+	u.SetName(a.id.ID())
```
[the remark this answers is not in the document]
katrielt: Done

pkg/controller/direct/clouddeploy/target_mapper.go
```diff
@@ -27,7 +27,7 @@ func CloudDeployTargetSpec_FromProto(mapCtx *direct.MapContext, in *pb.Target) *
 	}
 	out := &krm.CloudDeployTargetSpec{}
 	out.Description = direct.LazyPtr(in.GetDescription())
-	//out.Annotations = in.Annotations
+	out.Annotations = in.Annotations
```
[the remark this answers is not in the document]
katrielt: Done

pkg/controller/direct/clouddeploy/target_mapper.go
```diff
@@ -64,7 +64,7 @@ func CloudDeployTargetSpec_ToProto(mapCtx *direct.MapContext, in *krm.CloudDeplo
 	}
 	out := &pb.Target{}
 	out.Description = direct.ValueOf(in.Description)
-	//out.Annotations = in.Annotations
+	out.Annotations = in.Annotations
```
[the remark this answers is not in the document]
katrielt: Done

pkg/controller/direct/clouddeploy/target_controller.go
```diff
@@ -0,0 +1,277 @@
...
+func (m *modelTarget) client(ctx context.Context) (*gcp.CloudDeployClient, error) {
+	var opts []option.ClientOption
+	opts, err := m.config.RESTClientOptions()
+	if err != nil {
+		return nil, err
+	}
+	gcpClient, err := gcp.NewCloudDeployRESTClient(ctx, opts...)
+	if err != nil {
+		return nil, fmt.Errorf("building Target client: %w", err)
+	}
```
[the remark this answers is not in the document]
katrielt: Done

pkg/test/resourcefixture/testdata/basic/clouddeploy/v1alpha1/clouddeploytarget/target-basic/_http.log
katrielt: Done

pkg/controller/direct/clouddeploy/target_controller.go
```diff
@@ -0,0 +1,367 @@
...
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &obj); err != nil {
+		return nil, fmt.Errorf("error converting to %T: %w", obj, err)
+	}
+
+	id, err := obj.GetIdentity(ctx, reader)
+	if err != nil {
+		return nil, err
+	}
+
+	// Normalize all reference fields
```
maqiuyujoyce, line 123: Reference fields (non-parent ones) resolution should be extracted out into a separate function and should be done in beginning of Create() and Update(). AdapterForObject() is called in the beginning of resource deletion and at that time the references may be deleted already and couldn't be resolved.

katrielt: Done

pkg/controller/direct/clouddeploy/target_controller.go
```diff
@@ -89,90 +90,102 @@ func (m *modelTarget) AdapterForObject(ctx context.Context, op *directbase.Adapt
 		return nil, err
 	}
 
+	// Get clouddeploy GCP client
+	gcpClient, err := m.client(ctx)
+	if err != nil {
+		return nil, err
+	}
+
```
maqiuyujoyce: You'll still want to handle labels here.

pkg/controller/direct/clouddeploy/target_controller.go
```diff
@@ -0,0 +1,390 @@
...
+func NewTargetModel(ctx context.Context, config *config.ControllerConfig) (directbase.Model, error) {
+	return &modelTarget{config: *config}, nil
+}
+
+var _ directbase.Model = &modelTarget{}
+
+type modelTarget struct {
+	config    config.ControllerConfig
+	gcpClient *gcp.CloudDeployClient
```
maqiuyujoyce: I think we don't want to preserve gcpClient in model.

pkg/controller/direct/clouddeploy/target_controller.go
```diff
@@ -0,0 +1,390 @@
...
+			if strings.HasPrefix(k, "goog-") || strings.HasPrefix(k, "go-") {
+				a.desiredPb.Labels[k] = v
+			}
+		}
+	}
+
+	// etag, execution_configs, uid, create_time, update_time, and target_id are server-generated.
+	// We skip the diff when they show up in path to avoid unnecessary drift.
+	paths, err := common.CompareProtoMessage(a.desiredPb, a.actual, func(fieldName protoreflect.Name, a, b proto.Message) (bool, error) {
+		if fieldName == "etag" || fieldName == "uid" || fieldName == "create_time" || fieldName == "update_time" || fieldName == "target_id" {
```
maqiuyujoyce: These output only field are covered by the BasicDiff I think. You don't need to explicitly list them here.

pkg/controller/direct/clouddeploy/target_controller.go
```diff
@@ -0,0 +1,390 @@
...
+			return false, nil
+		}
+		// Compare execution_configs if it's specified in the desired state
+		if fieldName == "execution_configs" {
+			desired := a.(*clouddeploypb.Target)
+			if len(desired.ExecutionConfigs) == 0 {
```
maqiuyujoyce, line 280: We should use the exact desired state under spec to represent the user intention when we decide if we want to update the resource or not.

For example, if user set executionConfigs before but unset it now, it means they want executionConfigs to be empty. We can handle default values if there are any, but we shouldn't treat "unsetting a field" as "doing nothing".

pkg/controller/direct/clouddeploy/target_controller.go
katrielt: Moved back to the Adapter.

pkg/controller/direct/clouddeploy/target_controller.go
katrielt: Removed specific fields.

pkg/controller/direct/clouddeploy/target_controller.go
katrielt: I've updated the diff to respect the intention to unset fields. I also verified the behavior against both real and mock GCP to make sure it handles the server defaults without looping.

maqiuyujoyce (approved):
/lgtm

PR #7172, opened 2026-03-25
Ensure PreviewResult logging in CombinedSummaryReport

anhdle-sso:
### BRIEF Change description

This pull request ensures that `PreviewResult` is logged in `CombinedSummaryReport` even if subsequent file operations fail, by moving the logging logic into a `defer` block. It also fixes a logic where `altResult.badResult` was only added to `combinedBadResult` after some initial reporting.

Fixes #

To guarantee that we capture bad results in logs for easier debugging, especially when report generation encounters errors.

```release-note
NONE
```

```docs

```

#### Intended Milestone

### Tests you have done

Manual verification that logs are produced correctly.

### DIFF
--- a/pkg/cli/preview/result.go
+++ b/pkg/cli/preview/result.go
@@ -139,33 +139,21 @@ func (r *RecorderReconciledResults) CombinedSummaryReport(summaryFile string, al
 	if altResult != nil && len(altResult.badResult) > 0 {
 		combinedBadResult = append(combinedBadResult, altResult.badResult...)
 	}
-
-	defer func() {
-		for _, result := range combinedBadResult {
-			klog.V(0).Info("\"PreviewResult\" ", result.FormatGKNNReconciledResult())
-		}
-	}()
+	for _, result := range combinedBadResult {
+		klog.V(0).Info("\"PreviewResult\" ", result.FormatGKNNReconciledResult())
+	}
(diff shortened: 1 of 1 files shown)


pkg/cli/preview/result.go
```diff
@@ -132,20 +132,27 @@ func (r *Recorder) GenerateRecorderReconciledResults() *RecorderReconciledResult
...
-	}
 	var combinedBadResult []*GKNNReconciledResult
 	if len(r.badResult) > 0 {
 		combinedBadResult = append(combinedBadResult, r.badResult...)
 	}
+	if altResult != nil && len(altResult.badResult) > 0 {
+		combinedBadResult = append(combinedBadResult, altResult.badResult...)
+	}
+
+	defer func() {
```
cheftako: While we are putting defers in the following from the end of this method should also be in defers
```
defer func() {
	if err := f.Close(); err != nil {
        klog.ErrorS("failed to close %s: %w", summaryFile, err)
    }
}()
defer func() {
    if err := w.Flush(); err != nil {
		klog.ErrorS("error flushing summary report %s: %w", summaryFile, err)
	}
}()
```

anhdle-sso: clean up the defer logics.

PR #7213, opened 2026-03-26
⚡ optimize string manipulation in pkg/text

cheftako:
💡 **What:**
Optimized multiple string manipulation functions in `pkg/text/text.go` that were using inefficient string concatenation inside loops.
- `CamelCaseToKebabCase`: Refactored to use `strings.Builder` with pre-allocation (`Grow`) and `b.WriteRune(unicode.ToLower(r))` to avoid extra string allocations.
- `snakeCaseToUpperCamelCase`: Refactored to use `strings.Builder` with pre-allocation.
- `SnakeCaseToLowerCase`: Replaced manual split-and-loop concatenation with a single `strings.ReplaceAll(s, "_", "")`.

🎯 **Why:**
String concatenation in Go creates a new string on every iteration, leading to O(N^2) time complexity and many allocations. Using `strings.Builder` (and `ReplaceAll` where applicable) improves both CPU and memory efficiency.

📊 **Measured Improvement:**
Benchmark results on a representative long input string:
- `BenchmarkCamelCaseToKebabCase`: 7402 ns/op -> 720.3 ns/op (**~90% reduction**)
- `BenchmarksnakeCaseToUpperCamelCase`: 3008 ns/op -> 2049 ns/op (**~32% reduction**)
- `BenchmarkSnakeCaseToLowerCase`: 1389 ns/op -> 371.4 ns/op (**~73% reduction**)

All existing tests passed successfully.

---
*PR created automatically by Jules for task [3521777165363716854](https://jules.google.com/task/3521777165363716854) started by @cheftako*

### DIFF
--- a/pkg/text/text.go
+++ b/pkg/text/text.go
@@ -81,8 +81,8 @@ func snakeCaseToUpperCamelCase(s string) string {
 	return b.String()
 }
 
-func SnakeCaseToLowerCase(s string) string {
-	return strings.ReplaceAll(s, "_", "")
+func SnakeCaseToFlatCase(s string) string {
+	return strings.ToLower(strings.ReplaceAll(s, "_", ""))
 }
 
 // AsSnakeCase returns the given string converted to lowercase snake_case. If the input is already snake_case, no
--- a/pkg/text/text_test.go
+++ b/pkg/text/text_test.go
@@ -44,7 +44,7 @@ func TestAsSnakeCase(t *testing.T) {
 	}
 }
 
-func TestSnakeCaseToLowerCase(t *testing.T) {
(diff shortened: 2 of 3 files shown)


pkg/text/text.go
```diff
@@ -69,24 +69,20 @@ func SnakeCaseToUpperCamelCase(s string) string {
...
 		} else {
-			ret += strings.Title(v)
+			b.WriteString(strings.Title(v))
 		}
 	}
-	return ret
+	return b.String()
 }
 
 func SnakeCaseToLowerCase(s string) string {
```
cheftako: Not sure this should be called "ToLowerCase" as it does not change the case. ToCamel or ToPascal would seem more correct.

PR #7228, opened 2026-03-27
Fix BigQueryDataTransferConfig resource duplication loop

eugenenuke:
Ensures `externalRef` is preserved in `BigQueryDataTransferConfig` status updates, preventing a "permanent amnesia" loop that leads to multiple duplicate scheduled queries in GCP.

### BRIEF Change description

This PR fixes a bug where `BigQueryDataTransferConfig` reconciliation fails to persist the link to the underlying GCP resource, leading to the creation of multiple duplicate scheduled queries.

The issue occurs because the `setStatus` helper function unconditionally overwrites the newly generated `externalRef` (from a successful `Create` call) with the potentially empty `externalRef` from the old Kubernetes object status. Since `BigQueryDataTransferConfig` uses service-generated IDs, the loss of this reference causes KCC to lose track of the resource, leading it to create a new instance on every subsequent reconciliation loop.

Fixes # (Internal Bug b/493812753)

1. **Identity Loss:** `BigQueryDataTransferConfig` relies entirely on `status.externalRef` to identify the existing resource in GCP. If this field is lost, KCC's `Find()` method returns `false`, and the controller concludes the resource does not exist.
2. **Duplication Loop:** When a resource is successfully created but the `externalRef` is immediately wiped from the status, KCC enters a state of "permanent amnesia." On the next sync (which is triggered immediately by the status update), KCC calls `Create()` again, resulting in an infinite loop of duplicate resources in GCP.
3. **Dependency Sensitivity:** This behavior is particularly triggered when dependencies (like Service Accounts or Datasets) are temporarily unhealthy, as the error-reporting path for those failures initializes an empty status that then "infects" the successful creation path via unconditional inheritance.
4. **Orphaned objects:** On the parent CR deletion, the corresponding GCP objects won't be deleted due to the lost link between the CR and the GCP resource.

The fix is surgical and targeted at the `setStatus` function in the `BigQueryDataTransferConfig` direct reconciler.

```release-note
Fixed a bug in BigQueryDataTransferConfig where the link to the GCP resource (externalRef) could be lost during status updates, causing KCC to create multiple duplicate scheduled queries in GCP.
```

```docs
NONE
```

#### Intended Milestone

### Tests you have done

- [x] Verified the fix in a live cluster using custom container images (with the fix).
- [x] Confirmed that `externalRef` is correctly persisted and duplication stops after the first successful reconciliation.
- [x] Verified that unhealthy dependency states (missing SA/Dataset) no longer trigger a duplication loop once the dependencies become healthy.

### DIFF
--- a/pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
+++ b/pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
@@ -227,8 +227,6 @@ func (a *Adapter) Find(ctx context.Context) (bool, error) {
 }
 
 func (a *Adapter) Create(ctx context.Context, createOp *directbase.CreateOperation) error {
-	u := createOp.GetUnstructured()
-
 	log := klog.FromContext(ctx)
 	log.V(2).Info("creating BigQueryDataTransferConfig", "name", a.id.FullyQualifiedName())
 	mapCtx := &direct.MapContext{}
@@ -270,7 +268,7 @@ func (a *Adapter) Create(ctx context.Context, createOp *directbase.CreateOperati
 	a.id.transferConfigID = serviceGeneratedID
 	status.ExternalRef = a.id.AsExternalRef()
 
-	return setStatus(u, status)
+	return createOp.UpdateStatus(ctx, status, nil)
 }
 
 func (a *Adapter) Update(ctx context.Context, updateOp *directbase.UpdateOperation) error {
@@ -372,7 +370,7 @@ func (a *Adapter) Update(ctx context.Context, updateOp *directbase.UpdateOperati
 	if mapCtx.Err() != nil {
 		return mapCtx.Err()
 	}
-	return setStatus(u, status)
+	return updateOp.UpdateStatus(ctx, status, nil)
 }
 
 func (a *Adapter) Export(ctx context.Context) (*unstructured.Unstructured, error) {
@@ -414,22 +412,4 @@ func (a *Adapter) Delete(ctx context.Context, deleteOp *directbase.DeleteOperati
 	log.V(2).Info("successfully deleted BigQueryDataTransferConfig", "name", a.id.FullyQualifiedName())
 
 	return true, nil
-}
-
-func setStatus(u *unstructured.Unstructured, typedStatus any) error {
-	status, err := runtime.DefaultUnstructuredConverter.ToUnstructured(typedStatus)
-	if err != nil {
-		return fmt.Errorf("error converting status to unstructured: %w", err)
-	}
-
-	old, _, _ := unstructured.NestedMap(u.Object, "status")
-	if old != nil {
-		status["conditions"] = old["conditions"]
-		status["observedGeneration"] = old["observedGeneration"]
-		status["externalRef"] = old["externalRef"]
-	}
-
-	u.Object["status"] = status
-
-	return nil
 }
--- a/tests/e2e/testdata/scenarios/bigquerydatatransferconfig_duplication/script.yaml
+++ b/tests/e2e/testdata/scenarios/bigquerydatatransferconfig_duplication/script.yaml
@@ -1,0 +1,56 @@
+# Step 0: Create BQDT with missing dependency
+# Use APPLY-10-SEC to record the expected failure and missing externalRef
+# without waiting for the full 20-minute timeout.
+TEST: APPLY-10-SEC
+apiVersion: bigquerydatatransfer.cnrm.cloud.google.com/v1beta1
+kind: BigQueryDataTransferConfig
+metadata:
+  name: bqdt-duplication-test
+spec:
+  projectRef:
+    external: ${projectId}
+  location: us
+  displayName: "duplication test"
+  dataSourceID: "scheduled_query"
+  params:
+    destination_table_name_template: "my_table"
+    write_disposition: "WRITE_APPEND"
+    query:  "SELECT 1"
+  schedule: "every day 05:00"
+  serviceAccountRef:
+    name: dependency-sa
+---
+# Step 1: Create the dependency
+apiVersion: iam.cnrm.cloud.google.com/v1beta1
+kind: IAMServiceAccount
+metadata:
+  name: dependency-sa
+spec:
+  displayName: "Dependency SA"
+---
+# Step 2: Touch BQDT to trigger reconciliation
+TEST: TOUCH
+apiVersion: bigquerydatatransfer.cnrm.cloud.google.com/v1beta1
+kind: BigQueryDataTransferConfig
+metadata:
+  name: bqdt-duplication-test
+---
+# Step 3: Ensure BQDT is Ready before cleanup to avoid long deletion timeouts
+# Default APPLY will wait for BQDT to be Ready
+apiVersion: bigquerydatatransfer.cnrm.cloud.google.com/v1beta1
+kind: BigQueryDataTransferConfig
+metadata:
+  name: bqdt-duplication-test
+spec:
+  projectRef:
+    external: ${projectId}
+  location: us
+  displayName: "duplication test"
+  dataSourceID: "scheduled_query"
+  params:
+    destination_table_name_template: "my_table"
+    write_disposition: "WRITE_APPEND"
+    query:  "SELECT 1"
+  schedule: "every day 05:00"
+  serviceAccountRef:
+    name: dependency-sa
--- a/apis/bigquerydatatransfer/v1alpha1/config_types.go
+++ b/apis/bigquerydatatransfer/v1alpha1/config_types.go
@@ -83,7 +83,7 @@ type BigQueryDataTransferConfigSpec struct {
 
 	Parent `json:",inline"`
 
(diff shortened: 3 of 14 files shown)


pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
```diff
@@ -426,7 +426,9 @@ func setStatus(u *unstructured.Unstructured, typedStatus any) error {
 	if old != nil {
 		status["conditions"] = old["conditions"]
 		status["observedGeneration"] = old["observedGeneration"]
-		status["externalRef"] = old["externalRef"]
+		if status["externalRef"] == nil {
+			status["externalRef"] = old["externalRef"]
+		}
 	}
 
 	u.Object["status"] = status
```
acpana: I need to take a closer look at the implications of this change. For context, this code (as most of the _controller codes for direct resources)  is generated and it follows established patterns.

This change would point to an issue in the current pattern with respect to dependency ordering. To be clear, the order in which dependencies come up should not have side effects [citation needed]. Further, following k8s, the main object should eventually be in the desired state automatically once all the dependencies are resolved.

acpana: cc @gemmahou

gemmahou: This resource was introduced in 2024 so it did not follow our standard identity&reference pattern. 

Unlike other `status` fields, `status.externalRef` should only be set during initial creation and it's immutable once set. It must not be populated by copying the old value. Thus, `setStatus` should update all other fields to reflect the current state of the GCP resource, except for `status.externalRef`. `status.externalRef` is handled separately [here](https://github.com/eugenenuke/k8s-config-connector/blob/master/pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go#L271).

gemmahou:
Since BigQueryDataTransferConfig contains service-generated id, can we also correct it's `resourceID` field description to `Immutable. Optional. The service-generated name of the resource. Used for acquisition only. Leave unset to create a new resource.`? We also need to include this resource into [function](https://github.com/gemmahou/k8s-config-connector/blob/eab262a1af32d3ca1e777fc3abf330a0c7d0e323/scripts/generate-google3-docs/resource-lists/main.go#L111) and use this function to re-gen this [list](https://github.com/gemmahou/k8s-config-connector/blob/eab262a1af32d3ca1e777fc3abf330a0c7d0e323/scripts/generate-google3-docs/resource-lists/generated/_resources-with-service-generated-resource-id.html), so it can be included in our [public documentation](https://docs.cloud.google.com/config-connector/docs/how-to/managing-deleting-resources#service-generated-resource-id).

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
maqiuyujoyce: @gemmahou thank you for the details. Did you suggest that we should just remove the original line 429 and add `status.ExternalRef = a.id.AsExternalRef()` in Update()?

gemmahou: Removing line 429 is sufficient. 

We only set `status.ExternalRef = a.id.AsExternalRef()` in Create(), which is already handled by the linked code above, and should **not** re-assign its value in Update(). See [controller template](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/ec0e4462476e543a76110abdf310a3336b5506af/dev/tools/controllerbuilder/template/controller/controller.go#L195) for reference; Though this file was not match the template, we can still see we only set `externalRef` during initial creation

jrmbrgs:
Hello, Any update on this issue ? 🙏

pkg/controller/direct/bigquerydatatransfer/bigquerydatatransferconfig_controller.go
eugenenuke: If we just remove line 429, the issue won't be solved at all as ExternalRef will be emptied by the next line: `u.Object["status"] = status`

I'm going to replace the cusom setStatus with a shared/common [createOp.UpdateStatus()](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/ec0e4462476e543a76110abdf310a3336b5506af/pkg/controller/direct/directbase/operations.go#L161) which follows the same logic as I was suggesting in my initial commit.

eugenenuke:
@gemmahou could you take another look please?

tests/e2e/testdata/scenarios/bigquerydatatransferconfig_duplication/script.yaml
```diff
@@ -0,0 +1,42 @@
+# Step 0: Create BQDT with missing SA dependency
```
gemmahou: kudos to @maqiuyujoyce 

This test does not seem to reproduce the issue. IIUC, the scenario is as follows:
- Step 00: Create a dependency IAMServiceAccount
- Step 01: Create a healthy BQDTC resource and verify that externalRef is populated correctly.
- Step 02: Delete the dependency IAMServiceAccount. 
- Step 03: Touch the BQDTC resource, now the resource should fall into a failure status; verify that externalRef persists and is not overridden by the status updates.

eugenenuke: done, PTAL

gemmahou: Thank you! The test results look good to me. object03 shows the expected failure, while externalRef remains persisted. No POST log in http03 verifies that no new BQDTC resource was created. @maqiuyujoyce would you mind approving this?

/lgtm

gemmahou: Sorry @eugenenuke, I double-checked the original issue, correct me if I'm wrong, I think the reproduce steps should be as follows:
- Step 00: Create an "unhealthy" BQDTC resource(missing the dependency IAMServiceAccount). The resource should fail, and externalRef should not be populated. This is the same as your previous implementation, however, please use APPLY instead of APPLY-NO-WAIT and SLEEP, as the latter won't record golden object([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/tests/e2e/script_test.go#L241-L242)). We need the golden log _object00.yaml to verify the failure status and missing externalRef.
- Step 01: Create a dependency IAMServiceAccount.
- Step 02: Touch the BQDTC resource to trigger reconciliation. 
   - Pre-fix behavior: reproducing the bug - The resource becomes healthy, but externalRef is missing. Investigation - externalRef should always be generated with successful reconciliation; but here it's overridden by the previous failed object without externalRef. Why the Condition field is not overridden by the previous failed object? I believe it comes from [handleUpToDate](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/60676c7f7f95e7e7e3645444c92529a2b245017c/pkg/controller/direct/directbase/directbase_controller.go#L460-L462), where we set Condition back to [UpToDate](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/lifecyclehandler/handler.go#L226) after the successful reconciliation.
     Would you mind verifying if you can get the 'Healthy but missing externalRef' state before your code change? Attaching those golden logs in this comment would be very helpful!
   - Post-fix behavior: The resource becomes healthy with externalRef populated. Let's record the golden logs.
   
@maqiuyujoyce would you mind taking a look at this test plan and see if that makes sense? Thanks!

eugenenuke: No problem at all

I've updated the test script so it follows the logic from the above, but I had to use `APPLY-10-SEC` instead of `APPLY` for the Step 0 and added Step 3 (to make the object "Ready") to avoid 20-30 mins waiting times (timeouts) as it seems that the testing framework waits till the Object is in Ready for the "APPLY".

PTAL

eugenenuke: Here is the diff between the BEFORE and AFTER states: 
`bigquerydatatransferconfig_duplication_bad`: golden logs generated by the original code (without the fix)

`bigquerydatatransferconfig_duplication`: golden logs generated by the code from PR #7228

``` $ pwd
/bigquerydatatransferconfig_duplication_bad

$ ls
_http01.log  _http02.log  _object00.yaml  _object01.yaml  _object02.yaml  _object03.yaml  script.yaml

$ diff _object00.yaml ../bigquerydatatransferconfig_duplication/_object00.yaml 
$ diff _object01.yaml ../bigquerydatatransferconfig_duplication/_object01.yaml 
$ diff _object02.yaml ../bigquerydatatransferconfig_duplication/_object02.yaml 
32a33
>   externalRef: //bigquerydatatransfer.googleapis.com/projects/${projectId}/locations/us/transferConfigs/${transferConfigID}
$ diff _object03.yaml _object02.yaml 
```
 As you can see the only diff is that externalRef added to the logs with changes from PR #7228.

maqiuyujoyce (approved):
/lgtm

PR #7235, opened 2026-03-27
feat: Allow setting nodeConfig while remove default-pool

gemmahou:
### BRIEF Change description

When the `cnrm.cloud.google.com/remove-default-node-pool: "true"` directive is used, as KCC's automated "GCP-unmanaged fields" logic overlays the nodeVersion (fetched from GCP) onto the desired state, triggering a validation conflict in the underlying Terraform provider. The validation should only happen during new creation([code](https://github.com/gemmahou/terraform-provider-google-beta/blob/main/google-beta/services/container/resource_container_cluster.go#L8321-L8322)).

Setting `nodeConfig` caused Terraform to interpret the reconciliation as a resource recreation. This **likely** occurs because `node_config` is marked as `ForceNew` in the TF schema. Once the default pool is deleted, `node_config` is absent from the live state; the resulting diff between the desired state (containing `node_config`) and the live state (missing it) triggers a replacement rather than an update, leading to the above error.

To avoid unintended re-creation in Terraform, we should ensure `nodeConfig` is removed from the desired state after the initial cluster provisioning. We eliminate the drift that causes the `ForceNew` logic to trigger.

This change is necessary because customers may require `nodeConfig` on the default node pool to satisfy system requirements during the initial cluster creation process. They also need to remove the default node pool later and manage their own nodes using the `ContainerNodePool` resource.

```release-note
Introduce "cnrm.cloud.google.com/remove-default-node-pool-allow-node-config" annotation to unblock cluster creation and reconciliation with restricted security policies that requires "nodeConfig".
```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/pkg/krmtotf/legacygcpmanagedfields.go
+++ b/pkg/krmtotf/legacygcpmanagedfields.go
@@ -113,12 +113,17 @@ func resolveContainerClusterNodeVersion(r *Resource, config map[string]interface
 	if err != nil {
 		return fmt.Errorf("error determining if release channel is set: %w", err)
 	}
-	if !found || releaseChannel == nil {
-		// Release channel is not specified, so no special behavior required.
-		return nil
-	}
-	if err := removeFromConfigIfNotApplied(r, config, "nodeVersion"); err != nil {
-		return fmt.Errorf("error resolving node version in config: %w", err)
+
+	// If the user opts to remove the default node pool, specifying a node version
+	// is also restricted by Terraform.
+	removeDefaultNodePoolDirective := "remove-default-node-pool"
+	removeDefaultNodePoolKey := k8s.FormatAnnotation(removeDefaultNodePoolDirective)
+	removeDefaultNodePool, _ := k8s.GetAnnotation(removeDefaultNodePoolKey, r)
+
+	if removeDefaultNodePool == "true" || (found && releaseChannel != nil) {
+		if err := removeFromConfigIfNotApplied(r, config, "nodeVersion"); err != nil {
+			return fmt.Errorf("error resolving node version in config: %w", err)
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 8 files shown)


pkg/krmtotf/legacygcpmanagedfields_test.go
```diff
@@ -496,7 +540,7 @@ func TestResolveGCPManagedFields(t *testing.T) {
 			},
 		},
 		{
-			name: "BigtableInstance removes an existing numNodes, if the value is removed",
+			name: "ContainerCluster removes an existing numNodes, if the value is removed",
 			kind: "BigtableInstance",
```
ldanielmadariaga, line 544: ContainerCluster?

pkg/krmtotf/legacygcpmanagedfields_test.go
```diff
@@ -539,17 +653,25 @@ func TestResolveGCPManagedFields(t *testing.T) {
...
+			}
+			annotations[k8s.LastAppliedConfigurationAnnotation] = string(lastAppliedConfigJSON)
+			r.SetAnnotations(annotations)
 			var liveState *terraform.InstanceState
 			if tc.resourceExists {
 				// The content of the instance state does not matter here,
 				// as the diff function later handles the merge. We just
 				// need to supply *something* to mark this resource
 				// as already existing.
+				attr := tc.liveStateAttr
```
ldanielmadariaga, line 668: Very small nit: when possible prefer to use sane defaults for your vars and only overwrite them when you know the new data is valid, this generally helps in preventing issues where we fail to check the validity of our input data. For this partcular case the recommendation doesn't lead to any functional changes but it's value compounds the more code you write.

In this case this could be rewritten as: 

```
attr := make(map[string]string)
if tc.liveStateAttr != nil {
  attr = tc.liveStateAttr
}
```

pkg/krmtotf/legacygcpmanagedfields.go
```diff
@@ -168,6 +182,22 @@ func resolveContainerClusterNodeConfig(r *Resource, liveState *terraform.Instanc
 		return nil
 	}
 
+	// If the resource is being created, we MUST NOT strip node_config, because the user may have specified
```
ldanielmadariaga, line 185: Tip for the future: Whenever possible try to express these comments in code / extract methods rather than comments, see: https://martinfowler.com/bliki/FunctionLength.html

I do understand that some of these comments are hard to translate to code so I definitely prefer having said comments to none at all :)

ldanielmadariaga (approved):

gemmahou:
Let's use https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/7664 for the fix

PR #7296, opened 2026-04-04
chore: apis/workstations should follow generate.sh pattern

justinsb:
Consolidated v1alpha1 into v1beta1 and introduced generate.sh to workstations to generate types and mapper.


pkg/controller/direct/workstations/mapper.generated.go
```diff
@@ -12,16 +12,15 @@
...
-/*
-import (
-	pb "cloud.google.com/go/workstations/apiv1/workstationspb"
-	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
-	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/workstations/v1alpha1"
-	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/controller/direct"
-)
+package workstations
 
+/* found existing non-generated mapping function "WorkstationClusterObservedState_FromProto", skipping
```
ldanielmadariaga, line 23: These seem like AI worklog comments, these should be removed. 
Applies to this comment and similar ones below

pkg/controller/direct/workstations/mapper.generated.go
```diff
@@ -84,66 +92,89 @@ func WorkstationClusterSpec_ToProto(mapCtx *direct.MapContext, in *krm.Workstati
...
 	if in == nil {
 		return nil
 	}
 	out := &krm.WorkstationCluster_PrivateClusterConfig{}
 	out.EnablePrivateEndpoint = direct.LazyPtr(in.GetEnablePrivateEndpoint())
 	// MISSING: ClusterHostname
-	// MISSING: ServiceAttachmentUri
-	out.AllowedProjects = in.AllowedProjects
+	// MISSING: ServiceAttachmentURI
+	out.AllowedProjects = WorkstationCluster_PrivateClusterConfig_AllowedProjects_FromProto(mapCtx, in.AllowedProjects)
```
ldanielmadariaga, line 116: I don't see the methods 'WorkstationCluster_PrivateClusterConfig_AllowedProjects_*' defined are we sure this function call is correct?

pkg/controller/direct/workstations/mapper.generated.go
```diff
@@ -12,16 +12,15 @@
...
 
-package workstations
+// Code generated by dev/tasks/generate-all. DO NOT EDIT.
+// +generated:mapper
+// krm.group: workstations.cnrm.cloud.google.com
+// krm.version: v1beta1
+// proto.service: google.cloud.workstations.v1
 
-/*
-import (
```
ldanielmadariaga, line 18: We should keep these imports

ldanielmadariaga (dismissed):

pkg/controller/direct/workstations/mapper.generated.go
justinsb: Actually this is controlled by `--include-skipped-output `.  I think it can be useful because then both humans and LLMs have the generated form for each type/function, even when there is a hand-coded type/function.  That is super-handy because when a field is added we can look at the diff and know we have to apply the same change to the hand-code type/function.

But we can always drop the flag.  But these are not "train of thought" comments from the AI - all the .generated.go files are generated deterministically (I believe)

pkg/controller/direct/workstations/mapper.generated.go
justinsb: The code compiles, so I suspect it is a commented-out method.  We will sometimes generate calls to methods that we cannot implement automatically, the requirement is that they be hand-coded.

(FYI, we use the function name as the key to allow replacement of a generated function)

pkg/controller/direct/workstations/mapper.generated.go
justinsb: It's tricky to do so, because this file is generated, and we only generate the imports we need.  Why do you want to keep them?

pkg/controller/direct/workstations/mapper.generated.go
ldanielmadariaga: Yeah this is all commented out let's keep as is

pkg/controller/direct/workstations/mapper.generated.go
ldanielmadariaga: Ah cool thanks for the context I was missing this

pkg/controller/direct/workstations/mapper.generated.go
ldanielmadariaga: We're commenting out the rest of the code this is fine. Was just comparing against other generated resources but this case is mostly overridden by manual implementations.

PR #7297, opened 2026-04-04
chore: create skill for backfilling generate.sh

justinsb:
Prompt:

   I want to repeat a change I made across a codebase, specifically https://github.com/GoogleCloudPlatform/k8s-config-connector. Specifically
   https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/7293 . I want to author a skill that will periodically look for subdirectories of apis/
   that do not have a generate.sh script, and whenever I find one I want to create a PR similar to #7293 to add generate.sh

### DIFF
--- a/.gemini/skills/generate-sh-checker/SKILL.md
+++ b/.gemini/skills/generate-sh-checker/SKILL.md
@@ -1,0 +1,85 @@
+---
+name: generate-sh-checker
+description: Checks apis/ subdirectories for missing generate.sh and helps create PRs to add them, following the pattern in #7293. Use this when you want to ensure all KCC resources have a standardized generation script.
+---
+
+# Generate.sh Checker
+
+This skill helps maintain the `generate.sh` pattern across all `apis/` subdirectories in the Config Connector codebase.
+
+## Workflow
+
+1.  **Scan for missing scripts**: Find subdirectories in `apis/` (usually `v1beta1` or `v1alpha1`) that do not have a `generate.sh` file.
+    ```bash
+    find apis -maxdepth 2 -type d \( -name "v1beta1" -o -name "v1alpha1" \) | while read dir; do [ ! -f "$dir/generate.sh" ] && echo "$dir"; done
+    ```
+
+2.  **Gather Resource Information**: For each identified directory, read `api_types.go` and `groupversion_info.go` to extract:
+    -   `PROTO_SERVICE`: Look for `// +kcc:spec:proto=` or `// +kcc:proto=` markers in `api_types.go`.
+    -   `GROUP`: Look for `// +groupName=` in `groupversion_info.go`.
+    -   `VERSION`: The directory name (e.g., `v1beta1`).
+    -   `RESOURCE_MAPPINGS`: Mapping of `Kind:ProtoMessage` from `// +kcc:spec:proto=` markers.
+    -   `SERVICE_NAME`: The parent directory name in `apis/` (e.g., `apigateway`).
+
+3.  **Create generate.sh**: Create a `generate.sh` file in the directory. Ensure the year in the copyright header is current (2026).
+    
+    Template:
+    ```bash
+    #!/bin/bash
+    # Copyright 2026 Google LLC
+    #
+    # Licensed under the Apache License, Version 2.0 (the "License");
+    # you may not use this file except in compliance with the License.
+    # You may obtain a copy of the License at
+    #
+    #      http://www.apache.org/licenses/LICENSE-2.0
+    #
+    # Unless required by applicable law or agreed to in writing, software
+    # distributed under the License is distributed on an "AS IS" BASIS,
+    # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    # See the License for the specific language governing permissions and
+    # limitations under the License.
+
+    set -o errexit
+    set -o nounset
+    set -o pipefail
+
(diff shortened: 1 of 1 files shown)


.gemini/skills/generate-sh-checker/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+
+# Generate.sh Checker
+
+This skill helps maintain the `generate.sh` pattern across all `apis/` subdirectories in the Config Connector codebase.
+
+## Workflow
+
+1.  **Scan for missing scripts**: Find subdirectories in `apis/` (usually `v1beta1` or `v1alpha1`) that do not have a `generate.sh` file.
+    ```bash
+    find apis -maxdepth 2 -type d \( -name "v1beta1" -o -name "v1alpha1" \) | while read dir; do [ ! -f "$dir/generate.sh" ] && echo "$dir"; done
```
ldanielmadariaga, line 14: Consider adding `v1` to the find command, as there are resources in `v1` (like `apis/connector/v1`). For example: `\( -name "v1beta1" -o -name "v1alpha1" -o -name "v1" \)`.

.gemini/skills/generate-sh-checker/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+    ```
+
+2.  **Gather Resource Information**: For each identified directory, read `api_types.go` and `groupversion_info.go` to extract:
```
ldanielmadariaga, line 17: Type definitions are typically located in a file named after the resource, like `*_types.go` (e.g., `billingaccount_types.go`). They are rarely named exactly `api_types.go` (which is specific to APIGateway). It would be safer to advise the LLM to read `*_types.go`.

.gemini/skills/generate-sh-checker/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+    -   `PROTO_SERVICE`: Look for `// +kcc:spec:proto=` or `// +kcc:proto=` markers in `api_types.go`.
```
ldanielmadariaga, line 18: Similarly, this should advise looking for markers in `*_types.go` rather than strictly `api_types.go`.

.gemini/skills/generate-sh-checker/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+    cd ${REPO_ROOT}/dev/tools/controllerbuilder
+
+    ./generate-proto.sh
+
+    go run . generate-types \
+      --service <PROTO_SERVICE> \
+      --api-version <GROUP>/<VERSION> \
+      --include-skipped-output \
+      --resource <KIND1>:<PROTO_MESSAGE1> \
+      [--resource <KIND2>:<PROTO_MESSAGE2> ...]
```
ldanielmadariaga, line 58: You might want to add a comment indicating this is pseudo-code for 'add as many --resource flags as needed'. Otherwise, an LLM might literally insert `[--resource <KIND2>:<PROTO_MESSAGE2> ...]` into the bash script.

.gemini/skills/generate-sh-checker/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+    dev/tasks/generate-crds
+
+    go run -mod=readonly golang.org/x/tools/cmd/goimports@${GOLANG_X_TOOLS_VERSION} -w  pkg/controller/direct/<SERVICE_NAME>/
+    ```
+
+4.  **Special Handling (Promotion/Consolidation)**:
+    -   If a `v1beta1` directory is being updated and a `v1alpha1` directory exists for the same service, check if `v1alpha1` should be consolidated.
+    -   Following #7293, this involves:
+        -   Removing the `v1alpha1` directory.
+        -   Adding `// +kubebuilder:metadata:labels="internal.cloud.google.com/additional-versions=v1alpha1"` to the `v1beta1` `api_types.go` file (near the Kind struct).
```
ldanielmadariaga, line 75: This should also reference `*_types.go` instead of `api_types.go` to match the file naming convention across most services.

.gemini/skills/generate-sh-checker/SKILL.md
justinsb: So I propose the policy here should be that you check in the skill as you used it, and you check in improvements to the skills when you need to make them.

aka the proof of the pudding is in the eating :-)

ldanielmadariaga: Sounds good let's go with that approach

.gemini/skills/generate-sh-checker/SKILL.md
ldanielmadariaga: Closing per the above comment

.gemini/skills/generate-sh-checker/SKILL.md
ldanielmadariaga: Closing per the above comment

.gemini/skills/generate-sh-checker/SKILL.md
ldanielmadariaga: Closing per the above comment

.gemini/skills/generate-sh-checker/SKILL.md
ldanielmadariaga: Closing per the above comment

PR #7298, opened 2026-04-04
chore: run diff-crds in presubmit

justinsb:
We're not blocking on it yet, but we want to see the differences in CI.

### DIFF
--- a/dev/tasks/diff-crds
+++ b/dev/tasks/diff-crds
@@ -23,10 +23,21 @@ def main():
     parser = argparse.ArgumentParser(description="Diff CRDs against a base revision.")
     parser.add_argument(
         "--base",
-        default="upstream/master",
-        help="The base revision to compare against (default: upstream/master)",
+        default="",
+        help="The base revision to compare against",
     )
     args = parser.parse_args()
+
+    base = args.base
+    if not base:
+        # Try upstream/master, then origin/master, then master, then HEAD~1
+        for candidate in ["upstream/master", "origin/master", "master", "HEAD~1"]:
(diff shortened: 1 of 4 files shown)


dev/tasks/diff-crds
```diff
@@ -23,11 +23,22 @@ def main():
...
-        help="The base revision to compare against (default: upstream/master)",
+        default="",
+        help="The base revision to compare against",
     )
     args = parser.parse_args()
 
+    base = args.base
+    if not base:
+        # Try upstream/master, then origin/master, then master, then HEAD~1
+        for candidate in ["upstream/master", "origin/master", "master", "HEAD~1"]:
```
ldanielmadariaga, line 34: Since this runs in a CI environment (like GitHub Actions), consider checking environment variables such as `GITHUB_BASE_REF` or `PULL_BASE_REF` first before falling back to `master` variants. This ensures the script still compares against the correct base if a PR targets a non-default branch (e.g., a `release-*` branch).

dev/tasks/diff-crds
```diff
@@ -107,8 +124,11 @@ def main():
 
             if stdout_data:
                 print("***")
-                print(f"Diffs for CRD: {crd_file}")
+                print(f"schema-diffs for CRD: {crd_file}")
```
ldanielmadariaga, line 127: Preserve the old naming scheme "Diffs for CRD"

dev/tasks/diff-crds
```diff
@@ -107,8 +124,11 @@ def main():
...
                 print(stdout_data, end="")
+            else:
+                print("***")
+                print(f"No schema-diffs for CRD: {crd_file}")
```
ldanielmadariaga, line 132: Preserve the old naming scheme for diffs. Replace with "No diffs for CRD: "

PR #7311, opened 2026-04-06
Add conditional CRD equivalence check to presubmit

maqiuyujoyce:
### BRIEF Change description

Add the CRD equivalence presubmit test for PRs with label `step/gen-types`.

Test PRs are #7311, #7312, #7313, #7314.

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/.github/workflows/crd-equivalence-check.yaml
+++ b/.github/workflows/crd-equivalence-check.yaml
@@ -32,7 +32,7 @@ on:
 jobs:
   crd-equivalence-check:
     runs-on: ubuntu-latest
-    if: ${{ contains(github.event.pull_request.labels.*.name, 'step/gen-types') }}
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'direct-migration') }}
     timeout-minutes: 60
     steps:
       - uses: actions/checkout@v4
--- a/dev/tasks/generate-github-actions
+++ b/dev/tasks/generate-github-actions
@@ -226,7 +226,7 @@ on:
 jobs:
   crd-equivalence-check:
     runs-on: ubuntu-latest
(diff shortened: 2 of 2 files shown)


.github/workflows/crd-equivalence-check.yaml
```diff
@@ -0,0 +1,58 @@
...
+  merge_group:
+    types: [checks_requested]
+    branches: ["master", "release-*"]
+    paths:
+      - 'config/crds/resources/**'
+
+jobs:
+  crd-equivalence-check:
+    runs-on: ubuntu-latest
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'step/gen-types') }}
```
ldanielmadariaga: Let's use a super obvious label like: 'tf2direct-check-equivalence. 

Alternatively, have you considered tweaking the script to not depend on a label but rather check for diffs in the CRD? It may seem a bit overkill right now but we should eventually also add checks to ensure changes to a CRD are compatible. So by tackling this now we'd be closer to also having the scaffolding for that check

maqiuyujoyce: Thank you for the feedback, @ldanielmadariaga !

Discussed with @cheftako offline that we want a obvious label for direct migration in general, so I updated to call it `direct-migration`. For either DCL to direct or TF to direct, and requirement should be the same. And for any step of the direct migration, the requirement of keep the CRD equivalent should apply. And we should ensure all the direct migration PRs have `direct-migration` label, and all direct migration specific checks should be triggered by this label.

And yes, for CRD backwards compatibility (looser compared with the equivalence check), we should have a presubmit check for it as well! Logged a issue for it: https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/7340

PR #7332, opened 2026-04-07
Support key project resolution mode in KMS Autokey configurations

### LINKED ISSUE #7691: Feature Request: Support keyProjectResolutionMode in KMS Autokey configurations
### Checklist

### Description
Currently, there is a need to support `keyProjectResolutionMode` in KMS Autokey configurations to manage project-level autokey settings. 

This feature request is to add the `keyProjectResolutionMode` field to the `KMSAutokeyConfig` resource and introduce a `KMSProjectAutokeyConfig` resource to handle identity resolution and project-level configurations. 

These changes will bring Config Connector in parity with the updates made in magic-modules (specifically PR 16401).

### Proposed Solution
* Add `KeyProjectResolutionMode` to `KMSAutokeyConfigSpec`.
* Create the `KMSProjectAutokeyConfig` resource to support project-level autokey configurations.
* Create the `KMSProjectAutokeyConfigIdentity` resource for identity resolution.
* Update the underlying `cloud.google.com/go/kms` dependency to pull in the necessary field definitions (v1.27.0+).

### API Documentation
The `AutokeyConfig` resource has been updated in the Cloud KMS API to include the `keyProjectResolutionMode` field and to support project-level autokey configurations. 
* **AutokeyConfig REST Resource**: [https://cloud.google.com/kms/docs/reference/rest/v1/folders/autokeyConfig](https://cloud.google.com/kms/docs/reference/rest/v1/folders/autokeyConfig) (also applicable to `projects/{PROJECT_NUMBER}/autokeyConfig`)
* **KeyProjectResolutionMode Enum**: Determines the mechanism by which AutokeyConfig identifies a key project (e.g., `RESOURCE_PROJECT`, `DEDICATED_KEY_PROJECT`, `DISABLED`).

### `gcloud` Commands
The Google Cloud CLI (`gcloud alpha`) supports updating and describing this configuration at both the folder and project levels:
* **Update Config:** `gcloud alpha kms autokey-config update` (accepts `--project` or `--folder` arguments and configurations containing `keyProjectResolutionMode`).
* **Describe Config:** `gcloud alpha kms autokey-config describe` (accepts `--project` or `--folder` arguments and outputs the effective `keyProjectResolutionMode`).
* **Show Effective Config:** `gcloud alpha kms autokey-config show-effective-config` (reflects the effective key-project and resolution mode for a given project).

### Importance

_No response_

iamkonohamaru:
Support key project resolution mode in KMS Autokey configurations

Fixes #7691

This change replicates the changes from magic-modules PR 16401 in Config Connector to support `keyProjectResolutionMode` in KMS Autokey configurations.

Changes:
- Added `KeyProjectResolutionMode` field to `KMSAutokeyConfigSpec` in `apis/kms/v1beta1/autokeyconfig_types.go`.
- Created `KMSProjectAutokeyConfig` resource in `apis/kms/v1beta1/projectautokeyconfig_types.go` to support project-level autokey configurations.
- Created `KMSProjectAutokeyConfigIdentity` in `apis/kms/v1beta1/projectautokeyconfig_identity.go` for identity resolution.
- Updated `pkg/controller/direct/kms/autokeyconfig/autokeyconfig_controller.go` to include `key_project_resolution_mode` in the update mask.
- Created controller for `ProjectAutokeyConfig` in `pkg/controller/direct/kms/projectautokeyconfig/projectautokeyconfig_controller.go`.
- Updated `cloud.google.com/go/kms` dependency to `v1.27.0` to pull in the required field definitions.

Generated CRDs and deepcopy code were also updated as part of the process.

Verified by successful compilation of both `autokeyconfig` and `projectautokeyconfig` controllers.


iamkonohamaru:
Local Testing results
(Cluster created via gcloud container clusters create my-cluster --zone us-central1-a --project key-project-442206)
[Setting Project Config to RESOURCE_PROJECT]
https://screenshot.googleplex.com/95DBZPkBS4A5zE2
https://screenshot.googleplex.com/idRWHL7evc5vTzA

[Setting PROJECT CONFIG TO DISABLED]
https://screenshot.googleplex.com/D8az246PthTAyQV
https://screenshot.googleplex.com/Bmh5EeAZYt7sPqX

maqiuyujoyce:
@gemmahou do you mind taking a look when you get a chance? I recall you are most familiar with KMS Autokey, right?
@ldanielmadariaga FYI.

apis/kms/v1beta1/projectautokeyconfig_types.go
gemmahou, line 1: I may not have context of this feature, why we introduce a new "KMSProjectAutokeyConfig" with fields "ProjectRef" and "KeyProjectResolutionMode" fields, I can see these same fields in "AutokeyConfig" resource. Is this a separate GCP resource?

apis/bigquerydatatransfer/v1beta1/types.generated.go
gemmahou, line 1: There're many files that irrelevant to KMS, they might be generated by a command that is needed to generate KMS, can we remove them from this PR?

apis/kms/v1beta1/projectautokeyconfig_types.go
iamkonohamaru: yes the underline resource is same, little differences are there around how we perform update opeation on folder vs newly supported project,
I was following the TF implementation we did for it.
https://github.com/GoogleCloudPlatform/magic-modules/pull/16401 

But no issues in making the support for both under one resource! updating

apis/bigquerydatatransfer/v1beta1/types.generated.go
iamkonohamaru: Sure, updating

iamkonohamaru:
I see below validation errors asking me to run 
https://github.com/GoogleCloudPlatform/k8s-config-connector/actions/runs/25487261383/job/74785994226?pr=7332
ERROR: Resource Go Clients must be regenerated. Please run 'make ready-pr' or 'make generate-go-client ensure fmt' and update your PR.
<hint_for_agent>Run 'make ready-pr' to regenerate Go clients.</hint_for_agent>
that might create unnecessary files on the PR like earlier

pkg/test/resourcefixture/testdata/basic/kms/v1beta1/kmsautokeyconfig/kmsautokeyconfig-project/update.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: kms.cnrm.cloud.google.com/v1beta1
+kind: KMSAutokeyConfig
+metadata:
+  name: autokeyconfig-project-${uniqueId}
+spec:
+  projectRef:
+    external: projects/${projectId}
+  keyProject:
+    name: project-${uniqueId}
```
gemmahou: Should be project2-${uniqueId} to test Update operation?

iamkonohamaru: Good catch here,
Attempting to update key_project for a project config is not supported
http://docs.google.com/spreadsheets/d/1TYLbb65tbZ-v-ydFLbWwdbXa3Q55QJlKHTJtroSedsk?content_ref=updating+key+project+in+autokey+config+for+a+project+is+not+supported#gid=0

We only support KPRM field for the config update.
Updating other snippets accordingly.

pkg/test/resourcefixture/testdata/basic/kms/v1beta1/kmsautokeyconfig/kmsautokeyconfig-project/create.yaml
```diff
@@ -4,8 +4,5 @@ metadata:
   name: autokeyconfig-project-${uniqueId}
 spec:
   projectRef:
-    external: projects/${projectId}
-  keyProject:
```
gemmahou: Could you confirm if keyProject is no longer needed here?

iamkonohamaru: yes, its redundant. removing.

gemmahou: We should remove "dependencies.yaml" as well. Just to double check: the autokeyconfig can be created and updated successfully against realGCP even there's no "keyProject" specified. Asking because in my previous test I remember keyProject is required to exist to manage the autokeyconfig.

iamkonohamaru: Done! Removed dependencies.yaml and regenerated the golden HTTP logs.

To confirm your check: yes, KMSAutokeyConfig successfully creates and updates against real GCP without specifying keyProject, provided the keyProjectResolutionMode is set to RESOURCE_PROJECT or DISABLED. When resolving keys via RESOURCE_PROJECT, GCP manages the key ring inside the resource's own project, making the separate keyProject reference completely redundant.

The e2e tests pass cleanly against both real GCP and our mock layer with this streamlined setup. Latest changes are pushed!

Used below
```
GCP_PROJECT_ID="konoha-apigee" \
E2E_GCP_TARGET="real" \
WRITE_GOLDEN_OUTPUT=1 \
RUN_TESTS="TestAllInSeries/fixtures/kmsautokeyconfig-project" \
./dev/tasks/run-e2e

```

tests/apichecks/testdata/exceptions/multi_version_crd_diff/KMSAutokeyConfig.diff
```diff
@@ -0,0 +1,51 @@
...
+  	Not:   nil,
+  	Properties: map[string]v1.JSONSchemaProps{
+  		"apiVersion": {Description: "APIVersion defines the versioned schema of this representation o"..., Type: "string"},
+  		"kind":       {Description: "Kind is a string value representing the REST resource this objec"..., Type: "string"},
+  		"metadata":   {Type: "object"},
+  		"spec": {
+  			... // 20 identical fields
+  			MaxProperties: nil,
+  			MinProperties: nil,
+- 			Required:      []string{"folderRef"},
```
maqiuyujoyce, line 15: Not a blocker for this PR.

We should update v1alpha1 version to match the v1beta1 version.

pkg/controller/direct/kms/autokeyconfig/autokeyconfig_controller.go
```diff
@@ -174,9 +182,12 @@ func (a *Adapter) updateAutokeyConfig(ctx context.Context, resource *kmspb.Autok
 		return nil, err
 	}
```
maqiuyujoyce: If for any reason we want to keep this a.Find() call here, these checks need to be updated to match the logic update in Find().

pkg/controller/direct/kms/autokeyconfig/autokeyconfig_controller.go
maqiuyujoyce: I don't think this is needed as a.Find() is already called in the directbase controller logic here: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/directbase/directbase_controller.go#L357 before a.Update() is called here: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/directbase/directbase_controller.go#L449

apis/kms/v1beta1/autokeyconfig_identity.go
```diff
@@ -37,44 +37,72 @@ func (r *KMSAutokeyConfigIdentity) Parent() *KMSAutokeyConfigParent {
...
 
 func (p *KMSAutokeyConfigParent) String() string {
-	return "folders/" + p.FolderID
+	if p.FolderID != "" {
+		return "folders/" + p.FolderID
+	}
+	return "projects/" + p.ProjectID
 }
 
 func NewAutokeyConfigIdentity(ctx context.Context, reader client.Reader, obj *KMSAutokeyConfig) (*KMSAutokeyConfigIdentity, error) {
```
maqiuyujoyce, line 51: Not a blocker for this PR.

The way we resolve the identity is error-prone, we should check status.externalRef first and then resolve the folderRef and projectRef.

pkg/controller/direct/kms/autokeyconfig/autokeyconfig_controller.go
maqiuyujoyce, line 85: Not a blocker for this PR.

We should resolve the references in the beginning of Update().

pkg/controller/direct/kms/autokeyconfig/mapper.generated.go
```diff
@@ -73,15 +79,21 @@ func KMSAutokeyConfigSpec_FromProto(mapCtx *direct.MapContext, in *pb.AutokeyCon
...
 		}
 	}
+	if in.GetKeyProjectResolutionMode() != pb.AutokeyConfig_KEY_PROJECT_RESOLUTION_MODE_UNSPECIFIED {
+		out.KeyProjectResolutionMode = direct.Enum_FromProto(mapCtx, in.GetKeyProjectResolutionMode())
+	}
 	return out
 }
 
-func KMSAutokeyConfig_FromFields(mapCtx *direct.MapContext, id *krm.KMSAutokeyConfigIdentity, keyProject *refs.ProjectIdentity) *pb.AutokeyConfig {
+func KMSAutokeyConfig_FromFields(mapCtx *direct.MapContext, id *krm.KMSAutokeyConfigIdentity, keyProject *refs.ProjectIdentity, keyProjectResolutionMode *string) *pb.AutokeyConfig {
```
maqiuyujoyce: Is this function updated manually? If so, please move it to a separate file called `autokey_mapper.go` under this folder. `mapper.generated.go` should contain generated code only.

pkg/controller/direct/kms/autokeyconfig/autokeyconfig_controller.go
maqiuyujoyce: KMSAutokeyConfigSpec_FromProto() set FolderRef by default. Please either update KMSAutokeyConfigSpec_FromProto() to distinguish between a folder level autokey config or a project level autokey config, or unset `obj.Spec.FolderRef` when projectId is not empty.

pkg/test/resourcefixture/testdata/basic/kms/v1beta1/kmsautokeyconfig/kmsautokeyconfig-folder/create.yaml
```diff
@@ -20,5 +20,5 @@ spec:
   folderRef:
     external: folders/${TEST_FOLDER_ID}
   keyProject:
```
maqiuyujoyce, line 22: Is there a reason you updated this test case? The existing test case covers different spec fields, and we don't want to remove the field coverage.

pkg/controller/direct/kms/autokeyconfig/autokeyconfig_controller.go
```diff
@@ -167,18 +175,16 @@ func (a *Adapter) Update(ctx context.Context, updateOp *directbase.UpdateOperati
...
 func (a *Adapter) updateAutokeyConfig(ctx context.Context, resource *kmspb.AutokeyConfig) (*kmspb.AutokeyConfig, error) {
 	log := klog.FromContext(ctx)
-	// To populate a.actual calling a.Find()
-	isExist, err := a.Find(ctx)
-	if !isExist {
-		return nil, fmt.Errorf("updateAutokeyConfig failed as AutokeyConfig does not exist, name: %s", a.id)
-	}
-	if err != nil {
-		return nil, err
+	if a.actual == nil {
```
maqiuyujoyce, line 178: Nit: This check is not necessary.

Not a blocker for this PR.

pkg/controller/direct/kms/autokeyconfig/autokeyconfig_controller.go
```diff
@@ -227,13 +240,16 @@ func (a *Adapter) Export(ctx context.Context) (*unstructured.Unstructured, error
 func (a *Adapter) Delete(ctx context.Context, deleteOp *directbase.DeleteOperation) (bool, error) {
 	log := klog.FromContext(ctx)
 	log.V(2).Info("deleting AutokeyConfig", "name", a.id)
-	_, err := a.Find(ctx)
-	if err != nil {
-		return false, err
+	if a.actual == nil {
```
maqiuyujoyce, line 243: We should redesign this deletion workflow.

It is a valid scenario that the AutokeyConfig doesn't exist any more in GCP when users delete this resource, i.e. the corresponding project or folder doesn't exist any more.

So for deletion, it should be something like:

```
	// Parent is gone so this resource no longer exists.
	if a.actual == nil {
    	return true, nil 
	}
	...
	updated, err := a.updateAutokeyConfig(ctx, resource)
	if err != nil {
		// Error message indicate the parent is gone.
		if NOT_FOUND {
			return true, nil
		}
		return false, fmt.Errorf("updating AutokeyConfig %s: %w", a.id, err)
	}
```

Not a blocker for this PR, but please follow up later.

maqiuyujoyce (approved):
/lgtm

pkg/controller/direct/kms/autokeyconfig/autokeyconfig_controller.go
gemmahou: Let's remove line178-179, this is already handled in directbase controller: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/d02ccb9422957a2c456c6d3f66ea0f1d59d1f268/pkg/controller/direct/directbase/directbase_controller.go#L357

pkg/controller/direct/kms/autokeyconfig/autokeyconfig_controller.go
gemmahou, line 98: This is no longer needed.

pkg/controller/direct/kms/autokeyconfig/autokeyconfig_controller.go
```diff
@@ -144,7 +152,7 @@ func (a *Adapter) Update(ctx context.Context, updateOp *directbase.UpdateOperati
 	log.V(2).Info("updating AutokeyConfig", "name", a.id)
 	mapCtx := &direct.MapContext{}
 
-	resource := KMSAutokeyConfig_FromFields(mapCtx, a.id, a.desiredKeyProject)
+	resource := KMSAutokeyConfig_FromFields(mapCtx, a.id, a.desiredKeyProject, a.desired.Spec.KeyProjectResolutionMode)
```
gemmahou, line 155: If we resolve keyproject under this Update function, we can replace a.desiredKeyProject in this mapper function to the resolved keyProject.

pkg/controller/direct/kms/autokeyconfig/autokeyconfig_controller.go
```diff
@@ -227,13 +240,16 @@ func (a *Adapter) Export(ctx context.Context) (*unstructured.Unstructured, error
 func (a *Adapter) Delete(ctx context.Context, deleteOp *directbase.DeleteOperation) (bool, error) {
 	log := klog.FromContext(ctx)
 	log.V(2).Info("deleting AutokeyConfig", "name", a.id)
-	_, err := a.Find(ctx)
-	if err != nil {
-		return false, err
+	if a.actual == nil {
+		return false, fmt.Errorf("delete AutokeyConfig failed: Find() was not called or returned false")
```
gemmahou, line 244: Per the example Joyce shared, should be "return true, nil". Because the autoKey is gone we can skip the rest code. But we consider the "deletion" succeed so we should not return error here.

pkg/controller/direct/kms/autokeyconfig/autokeyconfig_controller.go
```diff
@@ -227,13 +240,16 @@ func (a *Adapter) Export(ctx context.Context) (*unstructured.Unstructured, error
...
-	// Keep only spec fields, by round-tripping through KRM
+	// Keep only spec fields, by round-tripping through KRM.
+	// Make a copy of the a.actual, disable AutokeyConfig, and clear the project reference.
 	tempKrmAutokeyResource := AutokeyConfig_FromProto(mapCtx, a.actual)
+	tempKrmAutokeyResource.KeyProject = nil
+	disabledMode := "DISABLED"
+	tempKrmAutokeyResource.KeyProjectResolutionMode = &disabledMode
 	resource := AutokeyConfig_ToProto(mapCtx, tempKrmAutokeyResource)
 	updated, err := a.updateAutokeyConfig(ctx, resource)
 	if err != nil {
```
gemmahou, line 255: Per the example Joyce shared above, we should double-check if the err here is "not found" error, check how it's handled here for reference: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/d02ccb9422957a2c456c6d3f66ea0f1d59d1f268/pkg/controller/direct/workflowexecutions/execution_controller.go#L239-L245

apis/kms/v1beta1/autokeyconfig_identity.go
gemmahou: I feel the current code is sufficient. 
1. When we check status.externalRef, we need to resolve folderRef or projectRef first and later compare with externalRef to validate parent in externalRef is not changed.
2. We want to make sure the existence of one of the parent, though this might be handled by the CRD.

We have the similar logic in other NewIdentity function: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/d02ccb9422957a2c456c6d3f66ea0f1d59d1f268/apis/asset/v1beta1/feed_identity.go#L72

pkg/controller/direct/kms/autokeyconfig/autokeyconfig_controller.go
gemmahou: AdapterForObject is called at the beginning of every reconciliation. If it's a Delete, and the dependency keyProject is already gone, the reconcile will fail at resolve reference stage here and block the future deletion of the autokey resource. We only need to resolve the dependency keyProject  if it's Update so let's let's move line85-91 to Update().

Also we no longer need the "desiredKeyProject" field in Adapter, see the comment below.

tests/apichecks/testdata/exceptions/multi_version_crd_diff/KMSAutokeyConfig.diff
gemmahou: Since this resource is already in beta, we can safely remove the v1alpha1 code.
Please check this for reference: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/350d55443a00f2b3acb98a8785d28b8525fac391/.gemini/skills/generate-sh-checker/SKILL.md?plain=1#L81-L84

PR #7570, opened 2026-04-13
Allow read-only POST requests in preview mode

### LINKED ISSUE #7569: Allow read-only POST requests in KCC preview mode
The KCC `preview` command currently blocks all POST requests because they are assumed to be write operations. However, some read-only operations in GCP APIs use POST, such as `getIamPolicy` and `getOrgPolicy`.

We need to allow these specific custom methods to pass through in `interceptingGCPClient`.

Proposed solution:
Extract the custom method name from the URL path (using the `:` separator) and allow it if it matches a predefined list of read-only methods.

Current allowed methods:
- `getIamPolicy`
- `getOrgPolicy`

anhdle-sso:
This PR allows specific read-only custom methods that use POST in preview mode. Fixes #7569.

### DIFF
--- a/pkg/cli/preview/intercepting_gcp_client.go
+++ b/pkg/cli/preview/intercepting_gcp_client.go
@@ -193,21 +193,29 @@ func (c *interceptingGCPClient) RoundTrip(req *http.Request) (*http.Response, er
 	return c.blockedHTTPMethod(req)
 }
 
+// checkGCPRequestIsAllowed determines if an outgoing HTTP request to the GCP API is permitted.
+// In preview mode, we want to allow read-only operations and block any modifications.
 func (c *interceptingGCPClient) checkGCPRequestIsAllowed(req *http.Request) bool {
+	// Allow all GET requests as they are read-only.
 	if req.Method == "GET" {
 		return true
 	}
 
+	// POST requests are blocked by default, unless they are specific read-only custom methods.
 	if req.Method == "POST" {
 		allowedCustomMethods := map[string]bool{
 			"getIamPolicy": true,
 			"getOrgPolicy": true,
 		}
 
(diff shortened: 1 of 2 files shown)


pkg/cli/preview/intercepting_gcp_client.go
```diff
@@ -196,6 +193,27 @@ func (c *interceptingGCPClient) RoundTrip(req *http.Request) (*http.Response, er
...
+		return true
+	}
+
+	if req.Method == "POST" {
+		allowedCustomMethods := map[string]bool{
+			"getIamPolicy": true,
+			"getOrgPolicy": true,
+		}
+
+		if idx := strings.LastIndex(req.URL.Path, ":"); idx != -1 {
```
cheftako, line 212: Can we either use regex or make sure there is only 1 ':'.

pkg/cli/preview/intercepting_gcp_client.go
```diff
@@ -196,6 +193,35 @@ func (c *interceptingGCPClient) RoundTrip(req *http.Request) (*http.Response, er
...
+		// Check if the URL path ends with a custom method (e.g., ":getIamPolicy").
+		if idx := strings.LastIndex(req.URL.Path, ":"); idx != -1 {
+			// Ensure there is only one colon to avoid ambiguity in parsing.
+			if strings.Count(req.URL.Path, ":") == 1 {
+				customMethod := req.URL.Path[idx+1:]
+				if allowedCustomMethods[customMethod] {
+					return true
+				}
+			}
+		}
```
cheftako, line 220: Do we want an else with print for non 1 counts?

cheftako (approved):
Looks good

PR #7571, opened 2026-04-14
mockgcp: update memorystore fidelity

justinsb:
- **mockgcp: update memorystorage fidelity**
- **tests: update golden output for memorystore gcloud tests**
- **mockgcp: stop generating memorystore from protos**


mockgcp/mockmemorystore/instance.go
```diff
@@ -106,7 +107,17 @@ func (s *instanceServer) populateDefaultsForInstance(name *instanceName, obj *pb
 	}
 
 	if obj.DeletionProtectionEnabled == nil {
-		obj.DeletionProtectionEnabled = mocks.PtrTo(false)
+		obj.DeletionProtectionEnabled = new(false)
```
fedebongio, line 110: new(false) and new("MEMORYSTORE_20260313_01_02") on line 114 are compile errors — Go's built-in new() takes a type, not a value. This looks like an attempted refactor away from mocks.PtrTo, but the replacement isn't valid Go. The "…/mockgcp/mocks" import is still present (line 33), so the simplest fix is to keep the helper:


```suggestion
obj.DeletionProtectionEnabled = mocks.PtrTo(false)
obj.EffectiveMaintenanceVersion = mocks.PtrTo("MEMORYSTORE_20260313_01_02")
```
CI confirms: test-mockgcp and tests-e2e-fixtures-memorystore checks are failing on the head commit — likely for this reason.

mockgcp/mockmemorystore/instance.go
```diff
@@ -106,7 +107,17 @@ func (s *instanceServer) populateDefaultsForInstance(name *instanceName, obj *pb
...
+	}
+
+	if obj.EffectiveMaintenanceVersion == nil {
+		obj.EffectiveMaintenanceVersion = new("MEMORYSTORE_20260313_01_02")
+	}
+
+	if obj.EncryptionInfo == nil {
+		obj.EncryptionInfo = &pb.EncryptionInfo{
```
fedebongio, line 118: On the Instance default path you set EncryptionInfo without a LastUpdateTime, but on the Backup path at line 560 you set both EncryptionType and LastUpdateTime: timestamppb.New(now). normalize.go line 49 adds a normalizer for .encryptionInfo.lastUpdateTime. Three possibilities:

Real GCP omits lastUpdateTime on Instance responses — then this is correct (please confirm with a real-GCP log).
Real GCP returns it on both — then Instance also needs LastUpdateTime set, or goldens will miss a field.
It's set server-side lazily and we're fine either way.
The normalizer path suggests you expected it somewhere; which path is it covering?

mockgcp/mockmemorystore/instance.go
```diff
@@ -126,17 +138,20 @@ func (s *instanceServer) populateDefaultsForInstance(name *instanceName, obj *pb
 			types = append(types, pb.ConnectionType_CONNECTION_TYPE_PRIMARY)
 			types = append(types, pb.ConnectionType_CONNECTION_TYPE_READER)
 		}
+		pscProjectNumber := 11199043661                                 // This is (presumably) a project owned by memorystore
```
fedebongio, line 141: - The comment // This is (presumably) a project owned by memorystore is doing a lot of work. "Presumably" is exactly the kind of thing that needs a real-GCP log for confirmation. If a real recording shows this project number, great — cite the commit. If it's guessed, the tests are aspirational rather than verified.
- pscUID is randomized per-run, which makes the mock more realistic, but I don't see a normalizer entry for the raw pscUID token — only for the full serviceAttachment string in normalize.go's Previsit. Since the attachment-name suffix is what's normalized there, this should work out — but worth confirming that no other path leaks the pscUID substring somewhere Previsit doesn't visit.

mockgcp/mockmemorystore/instance.go
```diff
@@ -224,8 +239,9 @@ func (s *instanceServer) populateDefaultsForInstance(name *instanceName, obj *pb
 		obj.TransitEncryptionMode = pb.Instance_TRANSIT_ENCRYPTION_DISABLED
 	}
 	if obj.Uid == "" {
-		obj.Uid = fmt.Sprintf("instance-%s", name.Name)
+		obj.Uid = uuid.New().String()
```
fedebongio, line 246: Replacing the deterministic "instance-<name>" UID with a random UUID is more realistic, but I don't see .uid (on Instance) in normalize.go's ConfigureVisitor replacements — only .instanceUid (on Backup responses). If real GCP returns uid on GET instance / list responses, the goldens will flake once that path is exercised. Is the Instance .uid already covered elsewhere, or should a .uid → PlaceholderUID replacement be added?

mockgcp/mockmemorystore/normalize.go
```diff
@@ -22,22 +22,60 @@ import (
...
+	replacements.ReplacePath(".backupFiles[].fileName", "fakefilename.rdb")
+	replacements.ReplacePath(".response.backupFiles[].createTime", PlaceholderTimestamp)
+	replacements.ReplacePath(".instanceUid", PlaceholderUID)
+	replacements.ReplacePath(".encryptionInfo.lastUpdateTime", PlaceholderTimestamp)
+
 }
 
 func (s *MockService) Previsit(event mockgcpregistry.Event, replacements mockgcpregistry.NormalizingVisitor) {
-	// No-op for now
+	if strings.Contains(event.URL(), "https://memorystore.googleapis.com") || strings.HasPrefix(event.URL(), "https://memorystore.mtls.googleapis.com") {
```
fedebongio, line 54: URL-predicate inconsistency with ConfigureVisitor at line 31 of the same file: ConfigureVisitor uses HasPrefix && HasPrefix, but this uses Contains || HasPrefix. Recommend switching to HasPrefix for both checks here — Contains will match any URL that mentions memorystore.googleapis.com anywhere (e.g. as a query param or in a response body URL), which isn't what you want, and it diverges from the sibling function's intent.

mockgcp/mockmemorystore/normalize.go
```diff
@@ -22,22 +22,60 @@ import (
...
+					replacements.ReplaceStringValue(tokens[5], "${backupCollectionID}")
+					replacements.ReplaceStringValue(tokens[7], "${backup}")
+				}
+
+			case ".pscAttachmentDetails[].serviceAttachment":
+				tokens := strings.Split(value, "/")
+
+				if len(tokens) == 6 && tokens[0] == "projects" && tokens[2] == "regions" && tokens[4] == "serviceAttachments" {
+					replacements.ReplaceStringValue(tokens[1], "${pscProjectNumber}")
+					//replacements.ReplaceStringValue(tokens[3], "${region}")
```
fedebongio, line 71: Dead code — either delete this line, or add a one-line // Don't normalize region because … reason so the next reader knows why it's left in.

fedebongio:
Left some specific comments, but more in general
- CI is red
- PR description under-sells the change (14k deletion, I understand that retires the local-proto generation for memorystore in favor of upstream cloud.google.com/go/memorystore/apiv1/memorystorepb but is probably worth calling out)
- Good call on retiring local proto generation: Strategically this is the right direction: upstream-proto-as-truth will keep fidelity up without a manual regeneration step each time memorystore's proto ships. Once the CI is green and there's a real-GCP anchor, this is a solid cleanup. Worth mentioning in the PR body that it reduces ~14k lines of mockgcp maintenance burden.

justinsb:
We're going to do this using codebot-robot

/close

PR #7581, opened 2026-04-14
tests: add endpoints testing for memorystore

justinsb:
- **tests: add endpoints testing for memorystore**
  

- **autogen: update golden output for memorystore tests**


justinsb:
@suwandim I am trying to create a test case for multiple PSC attachments (I think that is the "multi VPC" scenario?)

Two questions:

1) It looks like I can't create multiple PSC attachments in the initial create, but can then patch in a second one?  Is that right?

2) It looks like this works (I do get 2 PSC attachments showing up in status), but I do see 11 `PATCH` requests to the memory store instance.  It looks like they are on different fields though, so maybe that is just expected?

suwandim:
> @suwandim I am trying to create a test case for multiple PSC attachments (I think that is the "multi VPC" scenario?)

The multi-vpc scenario is described in details: https://docs.cloud.google.com/memorystore/docs/valkey/about-multiple-vpc-networking#connection_types.

Essentially, there are two ways to establish connection to the Valkey instance, either through automatically created connections (`PscAutoConnection`) or through manually user created connections (`PscConnection`). Automatically created connections (always two of them, one per each `PscAttachmentDetails`) are created during
the creation of the `MemorystoreInstance` and these connections belong to 1 VPC. To connect from more than 1 VPCs, users will need to create their own connections and attaching them to the `PscAttachmentDetails`. 

Due to the declarative nature of KCC, there is an issue of the object ordering for creation (`MemorystoreInstance`, `PscAttachmentDetails`, and `ForwardingRule`). Memorystore Terraform provider has the same issue. Their solution is to create  a proxy object solely to manage these `PscConnection`. The proxy object is called [`google_memorystore_instance_desired_user_created_endpoints`](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/memorystore_instance_desired_user_created_endpoints). The KCC proxy object is [`MemorystoreInstanceEndpoint`](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/7119/changes#diff-6030680eaa1c1266375f7e954420cd64daa0cc3817ac1d8db3711aecb288cbbb)

> 
> Two questions:
> 
> 1. It looks like I can't create multiple PSC attachments in the initial create, but can then patch in a second one?  Is that right?

Creating a MemorystoreInstance object always results in 2 PSC attachments.
You can observe this in the [_http.log#L1613-L1621](https://github.com/justinsb/k8s-config-connector/blob/f683b858cc77ec67b60f10b7a321abdf2a832790/pkg/test/resourcefixture/testdata/basic/memorystore/v1beta1/memorystoreinstance/fullmemorystoreinstance/_http.log#L1613-L1621)

```
    "pscAttachmentDetails": [
      {
        "connectionType": "CONNECTION_TYPE_DISCOVERY",
        "serviceAttachment": "projects/835388146573/regions/us-central1/serviceAttachments/gcp-memorystore-auto-e7049f56305c8d50-psc-sa"
      },
      {
        "serviceAttachment": "projects/835388146573/regions/us-central1/serviceAttachments/gcp-memorystore-auto-e7049f56305c8d50-psc-sa-2"
      }
    ],
```


> 2. It looks like this works (I do get 2 PSC attachments showing up in status), but I do see 11 `PATCH` requests to the memory store instance.  It looks like they are on different fields though, so maybe that is just expected?






You see 11 `PATCH` requests because MemorystoreInstance API only allows updating one field at a time, so the controller is sending serialized update requests, one per detected changed field (see https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/0614b986bfcf3d47aef4f9c374325cc1b692ea95/pkg/controller/direct/memorystore/instance_controller.go#L271-L290).

PR #7582, opened 2026-04-14
memorystore: add crossInstanceReplicationConfig

justinsb:
- **chore: create skill for backfilling generate.sh**
- **chore: apis/memorystore should follow generate.sh pattern**
- **memorystore: add crossInstanceReplicationConfig field**


pkg/controller/direct/memorystore/instance_controller.go
```diff
@@ -189,115 +203,117 @@ func (a *InstanceAdapter) Create(ctx context.Context, createOp *directbase.Creat
...
-		paths = append(paths, "endpoints")
-	}
-	if a.desired.Spec.Labels != nil && !reflect.DeepEqual(desiredPb.Labels, a.actual.Labels) {
-		report.AddField("labels", a.actual.Labels, desiredPb.Labels)
-		paths = append(paths, "labels")
-	}
-	if a.desired.Spec.EngineVersion != nil && !reflect.DeepEqual(desiredPb.EngineVersion, a.actual.EngineVersion) {
-		report.AddField("engine_version", a.actual.EngineVersion, desiredPb.EngineVersion)
-		paths = append(paths, "engine_version")
+	req.Instance.Name = a.id.String()
```
justinsb: @suwandim when do we have to break updates into separate steps, vs which can we do all at once?

suwandim: The MemorystoreInstance API does not allow updating more than 1 field at a time. Modifications are quite expensive and getting them applied can take multi-minutes to multi-hours

justinsb:
We did this in another PR.

/close

PR #7588, opened 2026-04-14
tests: iam introspections

acpana:
This adds a test to check whether a direct resource supports IAM and if we are supporting the IAMAdapter.

Note I am taking an "ai chore" approach to catching up w the missing IAMAdapters for direct -- https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/.agents/direct-iam-implementation.md 

This should help the bots.


tests/apichecks/crds_test.go
```diff
@@ -1144,3 +1152,80 @@ func validateCRDProps(props *apiextensions.JSONSchemaProps, path string) error {
...
+	var errs []string
+	for _, crd := range crds {
+		gvk := schema.GroupVersionKind{
+			Group:   crd.Spec.Group,
+			Version: crd.Spec.Versions[0].Name,
+			Kind:    crd.Spec.Names.Kind,
+		}
+
+		// Skip IAM resources themselves
+		if gvk.Group == "iam.cnrm.cloud.google.com" && (gvk.Kind == "IAMPolicy" || gvk.Kind == "IAMPolicyMember" || gvk.Kind == "IAMPartialPolicy" || gvk.Kind == "IAMAuditConfig") {
```
maqiuyujoyce: Nit: A switch case clause might be more readable here.

Not a blocker.

tests/apichecks/crds_test.go
```diff
@@ -1144,3 +1152,80 @@ func validateCRDProps(props *apiextensions.JSONSchemaProps, path string) error {
...
+		discovered := supportedgvks.SupportsIAMByGVK(gvk)
+		actual := isIAMSupportedInKCC(gvk, smLoader, dclLoader, dclSchemaLoader)
+
+		if discovered && !actual {
+			errs = append(errs, fmt.Sprintf("[iam_gap_direct] crd=%s: supports IAM in REST but not in KCC direct", crd.Name))
+		}
+		if !discovered && actual {
+			// Report resources that support IAM in KCC but REST discovery didn't find it.
+			// This helps identify resources where IAM functions might be under a different name or path.
+			errs = append(errs, fmt.Sprintf("[iam_unexpected] crd=%s: supports IAM in KCC but REST discovery didn't find it", crd.Name))
```
maqiuyujoyce, line 1197: I think instead of an exception, this means we should fix the gvks_generated.go, right? I'm okay with an exception for now, but this looks like a data problem we should fix.

scripts/generate-gvks/main.go
```diff
@@ -115,3 +129,96 @@ var legacyGVKs = map[schema.GroupVersionKind]legacyGVKData{`
 		os.Exit(1)
 	}
 }
+
+func findIAMResources() []IAMResource {
+	var iamResources []IAMResource
+	baseDir := ".build/third_party/googleapis"
```
maqiuyujoyce, line 135: Just curious, do we have the googleapis files under .build when we run the GitHub actions?

maqiuyujoyce (commented):
Overall lgtm with a couple of questions.

scripts/generate-gvks/main.go
acpana: yes I believe we generate it as part of `validate-generated-files`

tests/apichecks/crds_test.go
acpana: +1 however I was happy to see that there where no such instances (see the exceptions file). But yes this scenario shouldn't happen and I'd rather we are loud about if we hit it.

maqiuyujoyce (approved):
/lgtm

PR #7590, opened 2026-04-14
Fix preview mode to use upstream REST config for manager

### LINKED ISSUE #7589: Preview tool fails to read ConfigConnector due to empty restConfig
In preview mode, the manager is initialized with an empty restConfig, causing GetAPIReader() to default to localhost:80. This causes failures when AddDefaultControllers attempts to read ConfigConnector from the cluster. We should use upstreamRestConfig instead.

anhdle-sso:
Fixes #7589.

In preview mode, pass the upstreamRestConfig to kccmanager.New so that the APIReader can correctly query the cluster for ConfigConnector resources during startup.

### Why this change is safe:
1. **Writes are still blocked**: The manager's default client (`mgr.GetClient()`) remains hooked to an intercepting client that denies all mutations.
2. **Targeted Read-Only Access**: Only the `apiReader` (which bypasses the cache) will use this `upstreamRestConfig` to perform live GET requests for cluster configuration.
3. **Failsafe Intact**: The rest of the manager's operations still rely on the hooked clients, ensuring no requests "leak" to the real API server unexpectedly.

### DIFF
--- a/pkg/cli/preview/preview.go
+++ b/pkg/cli/preview/preview.go
@@ -176,15 +176,15 @@ func (i *PreviewInstance) Start(ctx context.Context) error {
 
 	// turn off caching (otherwise we get partial object metadata)
 	nocache.OnlyCacheCCAndCCC(&kccConfig.ManagerOptions)
-	// Use an empty restConfig as a failsafe against requests "leaking" to real kube-apiserver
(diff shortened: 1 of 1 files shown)


xiaoweim (approved):

anhdle-sso:
No need

PR #7592, opened 2026-04-14
Fix preview tool connection error and enforce read-only mode

### LINKED ISSUE #7589: Preview tool fails to read ConfigConnector due to empty restConfig

anhdle-sso:
### Description
This PR fixes a connection error in the KCC preview tool and enforces read-only access to the cluster for security.

### Key Changes
- **Manager Config**: Modified `pkg/cli/preview/preview.go` to pass `i.hookKube.upstreamRestConfig` to `kccmanager.New`. Previously, it was using an empty or invalid config in some code paths, which caused `apiReader` to default to `localhost:80` and fail with connection refused when reading `ConfigConnector` resources.
- **Read-Only Transport**: Implemented a custom `readOnlyTransport` that wraps the `rest.Config` transport. This ensures that any non-GET/HEAD/OPTIONS request made via this config is immediately rejected with a 403 Forbidden at the network level.
- **Isolation**: Used `rest.CopyConfig` to create a dedicated read-only configuration, ensuring the original `upstreamRESTConfig` remains untouched for other components.
- **Debug Logging**: Added a log line to print the host of the REST config before creating the manager to help diagnose any future connection issues.

Fixes #7589

### DIFF
--- a/pkg/cli/preview/preview.go
+++ b/pkg/cli/preview/preview.go
@@ -17,8 +17,10 @@ package preview
 import (
 	"context"
 	"fmt"
+	"io"
 	"net/http"
 	"slices"
+	"strings"
 	"time"
 
 	"golang.org/x/oauth2"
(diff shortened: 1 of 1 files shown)


justinsb (dismissed):
/approve
/lgtm

cheftako (approved):
Looks good.

Issue #194, opened 2020-06-01
Container Cluster stuck in non-ready state because channel update rejected 


jlewi:
**Describe the bug**

**ConfigConnector Version**
Run the following command to get the current ConfigConnector version
```bash 
kubectl get ns cnrm-system -o jsonpath='{.metadata.annotations.cnrm\.cloud\.google\.com/version}' 
```

cnrm.cloud.google.com/version: 1.9.1

**To Reproduce**
*Steps to reproduce the behavior:*

1. Create a ContainerCluster CNRM resource setting the release channel for the cluster to get
the status

1. Apply the resource to create the cluster

1. Reapply the resource 

1. Container cluster reports

    ```
      - lastTransitionTime: "2020-05-29T00:15:38Z"
       message: 'Update call failed: the desired mutation for the following field(s)
         is invalid: [releaseChannel.0.Channel]'
       reason: UpdateFailed
       status: "False"
       type: Ready
    ```

The release channel shouldn't be changing. 

I suspect this an issue in the update logic since there are some restrictions about mutations to
release channel
https://cloud.google.com/kubernetes-engine/docs/concepts/release-channels

*YAML snippets:*

```yaml
apiVersion: container.cnrm.cloud.google.com/v1beta1
kind: ContainerCluster
metadata:
  clusterName: gcp-private-dev/us-central1/jl-0601
  labels:
    kf-name: gcp-private-0527
    mesh_id: gcp-private-dev_us-central1_jl-0601
  name: jl-0601
  namespace: gcp-private-dev
spec:
  clusterAutoscaling:
    autoProvisioningDefaults:
      oauthScopes:
      - https://www.googleapis.com/auth/logging.write
      - https://www.googleapis.com/auth/monitoring
      - https://www.googleapis.com/auth/devstorage.read_only
      serviceAccountRef:
        name: jl-0601-vm
    enabled: true
    resourceLimits:
    - maximum: 128
      resourceType: cpu
    - maximum: 2000
      resourceType: memory
    - maximum: 16
      resourceType: nvidia-tesla-k80
  initialNodeCount: 2
  ipAllocationPolicy:
    clusterSecondaryRangeName: pods
    createSubnetwork: false
    servicesSecondaryRangeName: services
    useIpAliases: true
  location: us-central1
  loggingService: logging.googleapis.com/kubernetes
  monitoringService: monitoring.googleapis.com/kubernetes
  networkRef:
    name: jl-0601
[...]

PR #7598, opened 2026-04-15
memorystore: update controller to use change cookies

justinsb:


pkg/controller/direct/memorystore/instance_controller.go
```diff
@@ -166,17 +152,10 @@ func (a *InstanceAdapter) Find(ctx context.Context) (bool, error) {
...
-	desired := a.desired.DeepCopy()
-	resource := MemorystoreInstanceSpec_ToProto(mapCtx, &desired.Spec)
-	if mapCtx.Err() != nil {
-		return mapCtx.Err()
-	}
 
 	req := &memorystorepb.CreateInstanceRequest{
 		Parent:     a.id.Parent().String(),
-		Instance:   resource,
+		Instance:   a.desired,
```
cheftako: [Non blocking] Any concern we are no longer working with a copy but the "original"? Create does not (yet) seem to modify the object but I notice for isntance that Update does in a similar code path.

pkg/controller/direct/memorystore/instance_controller.go
```diff
@@ -189,113 +168,156 @@ func (a *InstanceAdapter) Create(ctx context.Context, createOp *directbase.Creat
...
+		for _, field := range diffs.Fields {
+			if field.ProtoFieldDescriptor == nil {
+				return fmt.Errorf("unexpected diff field without proto descriptor: %s", field.ID)
+			}
+
+			path := string(field.ProtoFieldDescriptor.Name())
+
+			desired := direct.ProtoClone(a.desired)
+
+			switch path {
```
cheftako: [Non blocking] Switch with 1 case seems verbose.

cheftako:
```
if path == "engine_version" && desired.EngineVersion == "" {
     continue
}
```
Just seems more clear.

pkg/controller/direct/memorystore/instance_controller.go
```diff
@@ -189,113 +168,156 @@ func (a *InstanceAdapter) Create(ctx context.Context, createOp *directbase.Creat
...
-	}
+		// The memorystore UpdateInstance API allows exactly one field to be updated per request.
+		for _, field := range diffs.Fields {
+			if field.ProtoFieldDescriptor == nil {
+				return fmt.Errorf("unexpected diff field without proto descriptor: %s", field.ID)
+			}
+
+			path := string(field.ProtoFieldDescriptor.Name())
+
+			desired := direct.ProtoClone(a.desired)
```
cheftako, line 220: [Non blocking] Seems like we could/should move this out of the for loop.

pkg/controller/direct/memorystore/instance_controller.go
```diff
@@ -189,113 +168,156 @@ func (a *InstanceAdapter) Create(ctx context.Context, createOp *directbase.Creat
...
 			if err != nil {
 				return fmt.Errorf("instance %s waiting update: %w", a.id, err)
 			}
-			log.V(2).Info("successfully updated Instance", "name", a.id, "updateMask", path)
+			latest = updated
 		}
+
 		log.V(2).Info("successfully updated Instance", "name", a.id)
+
+		if err := updateOp.SetLastModifiedCookie(ctx, a.desired, latest); err != nil {
```
cheftako: Is `latest` correct here? I would have thought it should be `a.actual`?

pkg/controller/direct/memorystore/instance_controller.go
justinsb: Yeah, we copy it when we think it will be changed.  I don't think CreateInstanceRequest will change the object, but we could defensively copy it anyway (?)

pkg/controller/direct/memorystore/instance_controller.go
justinsb: So we want the state after we applied the change.  Going forwards, if the desired state hash matches, and the GCP state hash matches, we will fast-path skip applies.

There is a potential race here, if we apply a change, and someone else applies another change, we will "miss" the second change.  In this case the more likely scenario is that someone "squeeze" a change in between our per-field changes.

We _could_ not use change-cookies, but if we do we will likely encounter continuous reconciliation.  We have a particularly problematic field here (`engine_version`), but regardless the change-cookie is an insurance policy.  It does mean we _sometimes_ will miss drift, but drift is the rare case; we could always force a full invalidation (much more) infrequently.

pkg/controller/direct/memorystore/instance_controller.go
```diff
@@ -79,54 +79,40 @@ func (m *modelInstance) AdapterForObject(ctx context.Context, op *directbase.Ada
...
 		return nil, err
 	}
+
+	mapCtx := &direct.MapContext{}
+	desired := MemorystoreInstanceSpec_ToProto(mapCtx, &obj.Spec)
+	if mapCtx.Err() != nil {
+		return nil, mapCtx.Err()
+	}
+
+	// Mask out some fields in the persistenceConfig, to accommodate KRM "discriminated union" semantics
```
ldanielmadariaga, line 98: Please add a link to driscriminated union semantics in the comment

pkg/controller/direct/memorystore/instance_controller.go
```diff
@@ -189,113 +168,156 @@ func (a *InstanceAdapter) Create(ctx context.Context, createOp *directbase.Creat
...
+	if err != nil {
+		return fmt.Errorf("comparing actual and desired Instance: %w", err)
 	}
 
-	report := &structuredreporting.Diff{Object: updateOp.GetUnstructured()}
+	latest := a.actual
 
-	var paths []string
+	if !diffs.HasDiff() {
+		log.Info("resource is up to date (no diff)", "name", a.id)
```
ldanielmadariaga: I would just call updateStatus and exit here. It follows the same short-circuit pattern as the cookie check in #194 and simplifies indentation levels which are already quite gnarly in the else branch.

pkg/controller/direct/memorystore/instance_controller.go
ldanielmadariaga: +1

ldanielmadariaga (dismissed):

pkg/controller/direct/memorystore/instance_controller.go
justinsb: Added a defensive copy here, but _not_ in the SetLastModifiedCookie.  I think that's the right "pattern", because it's probably not unusual that we will tweak desired for "API reasons", but we want the cookie to be based on what we compute before those modifications.

pkg/controller/direct/memorystore/instance_controller.go
justinsb: So we're defensively copying it before passing it to UpdateRequest.  For example, I half expected the API to only allow us to pass the one field we are updating.  But ... it doesn't seem to mind.  So I think copying inside the loop, before each request, is _logically_ correct.

That said, I do agree it's inefficient, and probably needless.  But I also think that we're establishing a pattern where before each API call, we copy the object, and then mutate for any API weirdnesses.  That feels like a good pattern (IMO)

pkg/controller/direct/memorystore/instance_controller.go
justinsb: Done!

pkg/controller/direct/memorystore/instance_controller.go
justinsb: You make a good point.  I realized we _do_ want to update the cookie here (I think) because the cookie did not match, but we applied all the changes we are going to apply.  If we don't update the cookie, we'll keep going through the "full diff" comparison, which is (marginally) less efficient and more noisy.

I would like to refactor the inner update loop, but I don't really know how to.  It would be nice if we didn't "hold" the reconcile goroutine while we make a dozen update calls, but I think early exiting would be particularly fragile here.

I did reorder the if statement so that the more normal case comes first (i.e. when there are diffs), I think it reads better.

pkg/controller/direct/memorystore/instance_controller.go
justinsb: Good call, cleaned it up.  It is a very special case, I thought there would be more like this, but I think the others are more straightforward fixed defaulting, this one is a (potentially) variable default.  (A bit like the maintenance version in SQL)

ldanielmadariaga (approved):

pkg/controller/direct/memorystore/instance_controller.go
cheftako: It sounds like it would be good to have an immutable version of or wrapper for the object for cases like this.

pkg/controller/direct/memorystore/instance_controller.go
cheftako: Not concerned about the race condition, because the cookie should get correctly set on the 2nd reconciliation, which still prevents infinite reconciliation, which is the goal. I was more concerned that Wait() (which returns latest), might not have the complete state of the resource. If we think it has all the field values for the resource then I'm good with it.

I think our goal with safer-upgrade to be to make it so we no longer need the cookie insurance policy. We're not there yet though.

PR #7621, opened 2026-04-16
memorystore: update identity & reference; also add skill

justinsb:


apis/memorystore/v1beta1/instance_identity.go
```diff
@@ -17,101 +17,92 @@ package v1beta1
...
+)
 
-func (i *InstanceIdentity) String() string {
-	return i.parent.String() + "/instances/" + i.id
-}
+var MemorystoreInstanceIdentityFormat = gcpurls.Template[MemorystoreInstanceIdentity]("memorystore.googleapis.com", "projects/{project}/locations/{location}/instances/{instance}")
 
-func (i *InstanceIdentity) ID() string {
-	return i.id
+// +k8s:deepcopy-gen=false
```
justinsb, line 35: FYI: We don't need deepcopy for this type, and this is why it "drops out" of the generated code.

apis/memorystore/v1beta1/instance_reference.go
```diff
@@ -42,42 +41,52 @@ type InstanceRef struct {
...
+
+func (r *InstanceRef) GetExternal() string {
+	return r.External
+}
+
+func (r *InstanceRef) SetExternal(ref string) {
+	r.External = ref
+}
+
+func (r *InstanceRef) ValidateExternal(ref string) error {
```
justinsb, line 67: @ldanielmadariaga :  We know that the function signatures on ValidateExternal and Normalize are not quite right (we can probably combine them into one).  My strategy has been to get everything onto the pattern we have today, and then fix the pattern.  You are free to make a different choice (or to make a different choice in a few weeks!)

pkg/controller/direct/memorystore/instance_controller.go
```diff
@@ -108,7 +108,7 @@ func (m *modelInstance) AdapterForObject(ctx context.Context, op *directbase.Ada
 	}
 
 	return &InstanceAdapter{
-		id:        id,
+		id:        id.(*krm.MemorystoreInstanceIdentity),
```
justinsb, line 111: FYI: It would be nice if we could have `type Foo[T] interface {  Get() T }` but every time I've tried it doesn't work.  Hence the cast.  Annoying, but .. not top of the list?

pkg/controller/direct/memorystore/instance_controller.go
```diff
@@ -120,7 +120,7 @@ func (m *modelInstance) AdapterForURL(ctx context.Context, url string) (directba
 }
 
 type InstanceAdapter struct {
-	id        *krm.InstanceIdentity
+	id        *krm.MemorystoreInstanceIdentity
```
justinsb, line 123: FYI: My _feeling_ is that we should err on the side of longer, more precise type names, even if they "stutter".  Same with the file names.  The reason is that it's actually easier to find as a human in the IDE, and the LLM actually seems to do better with them also.  The LLM definitely prefers a consistent pattern though, so having both while we change is a downside.

pkg/controller/direct/memorystore/instance_controller.go
```diff
@@ -157,9 +157,9 @@ func (a *InstanceAdapter) Create(ctx context.Context, createOp *directbase.Creat
 	desired := direct.ProtoClone(a.desired)
 
 	req := &memorystorepb.CreateInstanceRequest{
-		Parent:     a.id.Parent().String(),
+		Parent:     fmt.Sprintf("projects/%s/locations/%s", a.id.Project, a.id.Location),
```
justinsb, line 160: FYI: Having the method Parent() and defining the Parent type feels like overkill, given we only use it here.

.gemini/skills/kcc-identity-reference/SKILL.md
```diff
@@ -0,0 +1,59 @@
+---
+name: kcc-identity-reference
+description: Creates or updates the _identity.go and _reference.go files for a Config Connector resource, ensuring they follow the canonical gcpurls.Template pattern. Use this when you need to make sure the identity and reference is up to date for a KCC resource or when implementing IdentityV2 and refs.Ref for a resource.
+---
+
+# Kcc Identity Reference
+
+## Overview
+
+This skill guides you through creating or updating the `_identity.go` and `_reference.go` files for a given Config Connector resource (e.g. `MemoryStoreInstance`). The goal is to implement `identity.IdentityV2` and `refs.Ref` using the canonical `gcpurls.Template` pattern, identical to the pattern used in `apis/artifactregistry/v1beta1/artifactregistryrepository_identity.go`.
```
cheftako, line 10: Not sure if we need to say it explicitly but it would be nice if the change included the appropriate type assertion. I've noticed a few recent changes for these types not including the type assertion.

cheftako: NM I see you have it later.

apis/memorystore/v1beta1/instance_identity.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2025 Google LLC
+// Copyright 2026 Google LLC
```
cheftako: Shouldn't change the copyright year.

apis/memorystore/v1beta1/instance_reference.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2025 Google LLC
+// Copyright 2026 Google LLC
```
cheftako: Ditto

apis/memorystore/v1beta1/instance_reference.go
cheftako: +1. Agents work better if things are standardized. They also lower the cost of getting/keeping things standard.

pkg/controller/direct/memorystore/instance_controller.go
cheftako: This seems nice to have. Your call but if we get everything standardized, then when we solve this it should be easy to roll out.

pkg/controller/direct/memorystore/instance_controller.go
cheftako: +1. (But then I come from a Java background)

pkg/controller/direct/memorystore/instance_controller.go
cheftako: If we're going to do parent I think it should be actual composition.

cheftako (commented):
Looks pretty good. Our standard copyright date problem.

apis/memorystore/v1beta1/instance_identity.go
```diff
@@ -17,101 +17,92 @@ package v1beta1
...
-func (i *InstanceIdentity) String() string {
-	return i.parent.String() + "/instances/" + i.id
-}
+var MemorystoreInstanceIdentityFormat = gcpurls.Template[MemorystoreInstanceIdentity]("memorystore.googleapis.com", "projects/{project}/locations/{location}/instances/{instance}")
 
-func (i *InstanceIdentity) ID() string {
-	return i.id
+// +k8s:deepcopy-gen=false
+type MemorystoreInstanceIdentity struct {
+	Project  string
```
ldanielmadariaga, line 37: Do we need to export these?

ldanielmadariaga (dismissed):

justinsb:
Rebased on #7635 - cc @cheftako and @ldanielmadariaga

ldanielmadariaga (dismissed):

justinsb:
> justinsb dismissed ldanielmadariaga’s stale review
> The merge-base changed after approval.

Just for the record, this was github, not me.  I'm not sure this setting is going to speed us up.

justinsb:
Rebased now that base PR has merged.

ldanielmadariaga (approved):

PR #7624, opened 2026-04-16
mockgcp: GKEHub v1

acpana:
mockgcp GKEHub v1 introduction & update controller to use v1 protos!


mockgcp/mockgkehub/gkehubmembershipbinding.go
```diff
@@ -0,0 +1,140 @@
...
+	obj.State = &pb.MembershipBindingLifecycleState{Code: pb.MembershipBindingLifecycleState_READY}
+	obj.Uid = "111111111111111111111" // Stable UID for testing
+
+	if err := s.storage.Create(ctx, fqn, obj); err != nil {
+		return nil, err
+	}
+	metadata := &pb.OperationMetadata{
+		Target:     fqn,
+		CreateTime: now,
+		EndTime:    now,
```
ldanielmadariaga, line 70: QQ do we usually use the same value for start and end? I know we're gonna override the timstamps anyway but curious if we should advance to make this a bit more real. 

Ditto other resources

mockgcp/apply-proto-patches.sh
```diff
@@ -100,13 +100,6 @@ EOF
...
 sed -i '/message NodeNetworkConfig {/,/}/ s/string subnetwork = 19 \[(google.api.field_behavior) = OUTPUT_ONLY\];/string subnetwork = 19;/' ${REPO_ROOT}/mockgcp/third_party/googleapis/google/container/v1beta1/cluster_service.proto
 
-go run . --file ${REPO_ROOT}/mockgcp/third_party/googleapis/google/container/v1beta1/cluster_service.proto --message AdditionalIPRangesConfig --mode append <<EOF
-
-  // Output only. The status of the additional pod ranges.
-  string status = 3;
-
-EOF
-
 go run . --file ${REPO_ROOT}/mockgcp/third_party/googleapis/google/container/v1beta1/cluster_service.proto --message DNSEndpointConfig --mode append <<EOF
```
ldanielmadariaga, line 111: Is it really ok to remove these lines? We don't seem to be modifying any of this in the PR

mockgcp/mock_http_roundtrip.go
```diff
@@ -145,7 +144,6 @@ func NewMockRoundTripper(ctx context.Context, k8sClient client.Client, storage s
 	services = append(services, mocksecuresourcemanager.New(env, storage))
 	services = append(services, mockcloudfunctions.New(env, storage))
 	services = append(services, mockedgenetwork.New(env, storage))
-	services = append(services, mockgkehub.New(env, storage))
```
ldanielmadariaga, line 148: Not super familiar with this but seems like we shouldn't be removing Hub from the list of registered services?

pkg/controller/direct/gkehub/scope_controller.go
```diff
@@ -47,14 +45,14 @@ type gkeHubScopeModel struct {
 // model implements the Model interface.
 var _ directbase.Model = &gkeHubScopeModel{}
 
-type gkeHubScopeAdapter struct {
+type gkehubScopeAdapter struct {
```
ldanielmadariaga, line 48: nit: Naming should be preserverd as gkeHubScopeAdapter

pkg/controller/direct/gkehub/scope_mapper.go
```diff
@@ -15,47 +15,116 @@
 package gkehub
 
 import (
-	gkehubapi "google.golang.org/api/gkehub/v1beta"
+	pb "github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/generated/mockgcp/cloud/gkehub/v1"
```
ldanielmadariaga, line 18: Why are we doing conversions to the mock proto in controller code?

pkg/controller/direct/gkehub/utils.go
```diff
@@ -0,0 +1,90 @@
...
+func waitForLRO(ctx context.Context, client *gkehubapi.ProjectsLocationsOperationsService, op *gkehubapi.Operation) error {
+	if op.Done {
+		if op.Error != nil {
+			return fmt.Errorf("operation %q completed with error: %v", op.Name, op.Error)
+		}
+		return nil
+	}
+
+	retryPeriod := 5 * time.Second
+	timeoutDuration := 20 * time.Minute
```
ldanielmadariaga, line 37: We generally seem to timeout on context expiration (not saying it's better). We also don't seem to have a  common way to handle a common operation such as LRO waits. Not something to address in this PR but may be worth keeping consistency with how timeouts are handled.


https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/networkconnectivity/serviceconnectionpolicy_controller.go#L178

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/common/lro.go#L27

pkg/controller/direct/gkehub/utils.go
```diff
@@ -0,0 +1,90 @@
...
+
+func waitForLRO(ctx context.Context, client *gkehubapi.ProjectsLocationsOperationsService, op *gkehubapi.Operation) error {
+	if op.Done {
+		if op.Error != nil {
+			return fmt.Errorf("operation %q completed with error: %v", op.Name, op.Error)
+		}
+		return nil
+	}
+
+	retryPeriod := 5 * time.Second
```
ldanielmadariaga, line 36: This is almost a no-op you'll get to the 30 second upper bound in 3 attempts (35 seconds). Whereas if you just statically default to 30seconds you'll get 2 attempts in the same amount of time. I'd rather simplify the logic or receive the timeout value as a parameter (seems commonly used in our codebase)

PR #7625, opened 2026-04-16
Enable projectID to projectNumber transform in fields in Container LROs

maqiuyujoyce:
### BRIEF Change description

* Aligned `targetField` in LROs in mock http responses in Container tests with the real ones so that it uses `projectNumber` instead of `projectID`.
* Supported new handy functions in NormalizingVisitor:
  * TransformString transforms the string field based on the existing value.
  * ReplaceProjectIDWithProjectNumber replaces the projectID section in the selfLink/relative resource name with projectNumber.
  * TransformLRO transforms subfields when the response is an LRO.

Example diff between real and mock:

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/mockgcp/common/projects/projects.go
+++ b/mockgcp/common/projects/projects.go
@@ -15,6 +15,7 @@
 package projects
 
 import (
+	"fmt"
 	"strconv"
 	"strings"
 
@@ -77,22 +78,20 @@ func ParseProjectIDOrNumber(s string) (*ProjectName, error) {
 	return name, nil
 }
 
-// ReplaceProjectIDWithProjectNumber replaces the project identifier (ID or Number)
+// ReplaceProjectWithProjectNumberTemplate replaces the project identifier (ID or Number)
 // inside a GCP resource link with the literal string "${projectNumber}".
 // For example: "projects/my-project/zones/z" -> "projects/${projectNumber}/zones/z"
-func ReplaceProjectIDWithProjectNumber(s string) string {
+func ReplaceProjectWithProjectNumberTemplate(s string) (string, error) {
 	prefix := "projects/"
-	idx := strings.Index(s, prefix)
-	if idx == -1 {
-		return s
+	before, after, found := strings.Cut(s, prefix)
+	if !found {
+		return "", fmt.Errorf("project identifier prefix %q not found in string: %q", prefix, s)
 	}
 
-	start := idx + len(prefix)
-	nextSlash := strings.Index(s[start:], "/")
-
-	if nextSlash == -1 {
-		return s[:start] + "${projectNumber}"
+	_, suffix, foundSuffix := strings.Cut(after, "/")
+	if foundSuffix {
+		suffix = "/" + suffix
 	}
 
-	return s[:start] + "${projectNumber}" + s[start+nextSlash:]
+	return fmt.Sprintf("%s%s${projectNumber}%s", before, prefix, suffix), nil
 }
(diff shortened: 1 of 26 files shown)


maqiuyujoyce:
@ldanielmadariaga I added a couple of transform functions to help bridge the gap between mock behaviors and real behaviors. I also configured it in ContainerNodePool to make sure it works. PTAL!

mockgcp/common/projects/projects.go
```diff
@@ -76,3 +76,23 @@ func ParseProjectIDOrNumber(s string) (*ProjectName, error) {
...
+// inside a GCP resource link with the literal string "${projectNumber}".
+// For example: "projects/my-project/zones/z" -> "projects/${projectNumber}/zones/z"
+func ReplaceProjectIDWithProjectNumber(s string) string {
+	prefix := "projects/"
+	idx := strings.Index(s, prefix)
+	if idx == -1 {
+		return s
+	}
+
+	start := idx + len(prefix)
```
ldanielmadariaga: len(string) returns the number of bytes in a string I don't believe we want this. You can just do "start := idx + 1"

I do think we can be much less low level by using strings.CutPrefix() and strings.Cut() see:

```
prefix := "projects/"
sub, found := strings.CutPrefix(s, prefix)
if !found {
  return errors.New()
}

_, suffix, _ := strings.Cut(sub, "/") 

return fmt.sprintF("%s%s%s", prefix, "${projectNumber}", suffix)
```

mockgcp/common/projects/projects.go
```diff
@@ -76,3 +76,23 @@ func ParseProjectIDOrNumber(s string) (*ProjectName, error) {
...
 }
+
+// ReplaceProjectIDWithProjectNumber replaces the project identifier (ID or Number)
+// inside a GCP resource link with the literal string "${projectNumber}".
+// For example: "projects/my-project/zones/z" -> "projects/${projectNumber}/zones/z"
+func ReplaceProjectIDWithProjectNumber(s string) string {
+	prefix := "projects/"
+	idx := strings.Index(s, prefix)
+	if idx == -1 {
+		return s
```
ldanielmadariaga: Not throwing an error here can lead to subtle failures where we think we're passing the right data to the function but it's not doing anything in actuality.

mockgcp/common/projects/projects.go
```diff
@@ -76,3 +76,23 @@ func ParseProjectIDOrNumber(s string) (*ProjectName, error) {
 
 	return name, nil
 }
+
+// ReplaceProjectIDWithProjectNumber replaces the project identifier (ID or Number)
+// inside a GCP resource link with the literal string "${projectNumber}".
+// For example: "projects/my-project/zones/z" -> "projects/${projectNumber}/zones/z"
+func ReplaceProjectIDWithProjectNumber(s string) string {
```
ldanielmadariaga: I'd rename this to ReplaceProjectWithProjectNumberTemplate the existing name is a tad misleading I'd expect it to just swap a projectID with a projectNumber. Also note that I removed ID from the name as projectID is a very specific concept for GCP projects.

maqiuyujoyce: Makes sense. Done. Also renamed the function in the ObjectWalker.

mockgcp/common/projects/projects.go
maqiuyujoyce: Done. Though the callers will currently panic when getting an error as I don't want to make big refactoring of the ObjectWalker framework.

mockgcp/common/projects/projects.go
maqiuyujoyce: Done. Also added unit tests to verify it indeed works.

ldanielmadariaga (approved):

PR #7627, opened 2026-04-17
tests: strict testing for memorystore

justinsb:


justinsb:
FYI @ldanielmadariaga , strict testing means two things:

1) We enable server-side apply in our tests.  Without server-side apply, if we remove fields in `update.yaml`, they are not actually removed.
2) We turn on re-reconciliation tests, meaning we force the reconciler to run again without a change, and verify that we don't write to GCP again.  (A read is fine / expected).


Ideally we would opt-in all the kinds into "strict testing".  It often means looking carefully at the test output, though, so it would be a pretty big PR if we did them in "one shot"

tests/e2e/ratcheting.go
```diff
@@ -367,7 +367,6 @@ func ShouldTestRereconiliation(t *testing.T, testName string, primaryResource *u
 	case schema.GroupKind{Group: "managedkafka.cnrm.cloud.google.com", Kind: "ManagedKafkaConsumerGroup"}:
 	case schema.GroupKind{Group: "managedkafka.cnrm.cloud.google.com", Kind: "ManagedKafkaTopic"}:
 	case schema.GroupKind{Group: "memcache.cnrm.cloud.google.com", Kind: "MemcacheInstance"}:
-	case schema.GroupKind{Group: "memorystore.cnrm.cloud.google.com", Kind: "MemorystoreInstance"}:
```
fedebongio, line 370: Nice — this moves MemorystoreInstance toward the "no exemptions" goal called out in the file docstring (L26–L34). One check: sibling fixture memorystoreinstance/basicmemorystoreinstance/ is now also under SSA + re-reconciliation but isn't touched in this PR. Did it pass locally without needing a golden update? If so, worth a one-liner in the PR description; if not, its goldens should probably land in the same PR so the ratchet flip is atomic.

pkg/test/resourcefixture/testdata/basic/memorystore/v1beta1/memorystoreinstance/fullmemorystoreinstance/_generated_object_fullmemorystoreinstance.golden.yaml
```diff
@@ -26,15 +26,11 @@ spec:
...
   labels:
-    key1: val1
-    key2: val2
     keyx: valx
   location: us-central1
   mode: CLUSTER
   nodeType: STANDARD_SMALL
   persistenceConfig:
-    aofConfig:
-      appendFsync: EVERY_SEC
```
fedebongio, line 37: Useful cross-check: aofConfig is removed here, but _http.log shows no corresponding removal of an aofConfig field in GCP requests. That's the right outcome — the controller was already correctly omitting aofConfig from the GCP call when mode: RDB; client-side merge was only leaving the stale field on the Kubernetes object. Enabling SSA cleans up the KRM object without changing GCP traffic, so no mock/real divergence is introduced by this hunk.

pkg/test/resourcefixture/testdata/basic/memorystore/v1beta1/memorystoreinstance/fullmemorystoreinstance/_http.log
```diff
@@ -815,8 +815,6 @@ X-Goog-Request-Params: instance.name=projects%2F${projectId}%2Flocations%2Fus-ce
   },
   "engineVersion": "VALKEY_7_2",
   "labels": {
-    "key1": "val1",
-    "key2": "val2",
```
fedebongio, line 819: All 11 _http.log hunks follow the same pattern: only key1/key2 are removed from labels blocks; keyx is retained. create.yaml has {key1, key2} and update.yaml has {keyx}, so under SSA the post-update state correctly drops the keys the manifest no longer declares. No aofConfig/rdbConfig lines change in the traffic log, and no request or response bodies outside labels change — the diff is entirely explained by SSA field-removal semantics and matches the object golden update exactly. No unexpected discrepancies.

fedebongio:
None of my comments are blocking. 

Just a nit, could you fill in the description? The commit title + your follow-up comment explain it well — pulling that into the PR body would help future reviewers (and the release-notes process) understand the intent without reading the thread.

tests/e2e/ratcheting.go
justinsb: > Did it pass locally without needing a golden update
Yes - and we know this because the tests passed.  So I'd push back on whether we need a comment, the changes in the PR _are_ the changes needed.

It's also weird because it should be surprising to everyone that our test was removing a field but we were not actually applying it.  So I'm wary of slowing down the bug fixes here....

pkg/test/resourcefixture/testdata/basic/memorystore/v1beta1/memorystoreinstance/fullmemorystoreinstance/_generated_object_fullmemorystoreinstance.golden.yaml
justinsb: Spot on - that is exactly the problem we're fixing, and your review here is exactly what a reviewer should be looking for: are the golden output changes expected?

justinsb:
> Just a nit, could you fill in the description? 

Good call - done!

Also rebased on #7621 to set up a "pipeline" of PRs

ldanielmadariaga (dismissed):

ldanielmadariaga (dismissed):

ldanielmadariaga (dismissed):

justinsb:
> The merge-base changed after approval.
@ldanielmadariaga did you turn this on?  It seems like it is reducing velocity, and I'm not sure what problem it is solving now we have the merge queue.

I rebased this one, as it needs "approve" from the PR review screen.

justinsb:
We'll do this in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/7978 instead

PR #7635, opened 2026-04-19
chore: tweak schema for forwardingRule -> memorystore service attachment

justinsb:
Also we finally hit a circular dependency between compute and memorystore,
so we need to move the instance identity and ref into its own package,
that both compute and memorystore can use without creating a circular dependency.


apis/compute/v1beta1/forwardingrule_types.go
```diff
@@ -111,6 +112,22 @@ type Target struct {
 	TargetVPNGatewayRef *refs.ComputeTargetVPNGatewayRef `json:"targetVPNGatewayRef,omitempty"`
 }
 
+// MemorystoreInstanceServiceAttachment defines the resource reference to the GCP identifier
+// for the ServiceAttachment managed by the MemorystoreInstance pointed by the MemorystoreInstanceRef.
+// +k8s:deepcopy-gen=true
+type MemorystoreInstanceServiceAttachment struct {
+	// A reference to a MemorystoreInstance resource.
+	// +required
+	MemorystoreInstanceRef *memorystorerefs.InstanceRef `json:"memorystoreInstanceRef,omitempty"`
```
justinsb: Not sure if this should be instanceRef or memorystoreInstanceRef.

The parent is memorystoreInstanceServiceAttachment (not ...Ref;  it isn't a "normal" ref because of the extra field)

justinsb:
cc @suwandim 

/assign @cheftako 

/assign @ldanielmadariaga 

Looks like we have to create a separate refs package; it's kind of surprising we got away without it for this long.  The identity & reference are just moved into a new package memorystorerefs, except that NewInstanceIdentity has a dependency on `type Instance`, so has to stay in the v1beta1 package.

apis/memorystore/v1alpha1/endpoint_types.go
```diff
@@ -16,7 +16,7 @@ package v1alpha1
 
 import (
 	computev1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/apis/compute/v1beta1"
-	refsv1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/apis/memorystore/memorystorerefs"
```
justinsb: Yes, we repeat "memorystore".  I think it makes it much more ergonomic though; even if we called it "memorystore/refs" we'd probably still have to alias it to `memorystorerefs` (just as we alias the refs package currently)

apis/memorystore/memorystorerefs/instance_identity.go
```diff
@@ -0,0 +1,61 @@
...
+// limitations under the License.
+
+package memorystorerefs
+
+import (
+	"fmt"
+	"strings"
+)
+
+// InstanceIdentity defines the resource reference to MemorystoreInstance, which "External" field
```
justinsb: Note that we clean up the identity and reference types to follow our latest patterns in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/7621

apis/memorystore/memorystorerefs/instance_identity.go
```diff
@@ -0,0 +1,61 @@
+// Copyright 2025 Google LLC
```
ldanielmadariaga: Guessing this still needs to say 2026 even if the contents of this file are technically a branch

apis/compute/v1beta1/forwardingrule_types.go
ldanielmadariaga: Seems that ServiceAttachments are a part of compute engine and an external ref.
https://docs.cloud.google.com/compute/docs/reference/rest/v1/serviceAttachments
Resource format: projects/{project}/regions/{region}/serviceAttachments/{service_attachment}

This is made all the more confusing as this struct seems to just output data for an attached Private Service Connect:
https://docs.cloud.google.com/vpc/docs/about-vpc-hosted-services

apis/memorystore/v1beta1/instance_types.go
```diff
@@ -87,6 +94,57 @@ type MemorystoreInstanceSpec struct {
 	Mode *string `json:"mode,omitempty"`
 }
 
+// New builds a InstanceIdentity from the Config Connector Instance object.
+func NewInstanceIdentity(ctx context.Context, reader client.Reader, obj *MemorystoreInstance) (*memorystorerefs.InstanceIdentity, error) {
```
ldanielmadariaga: Are we forced to keep this builder here instead of moving it to memorystorerefs? Even if that's the case seems preferable to just have a passthrough here and avoid the exported fields InstanceIdentity. 

Passthrough: 
```
NewInstance(...) {
  return memorystorerefs.NewInstance(...)
}
```

apis/memorystore/memorystorerefs/instance_identity.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+// InstanceIdentity defines the resource reference to MemorystoreInstance, which "External" field
+// holds the GCP identifier for the KRM object.
+type InstanceIdentity struct {
+	Parent_ *InstanceParent
```
ldanielmadariaga: Weird naming scheme and having exported fields collides a bit with having getters. See my comment in apis/memorystore/v1beta1/instance_types.go

apis/memorystore/generate.sh
ldanielmadariaga, line 1: Not sure if this is intended, it seems in most cases we prefer to have a generate.sh per API version.

apis/memorystore/memorystorerefs/instance_identity.go
justinsb: Will update

apis/memorystore/v1beta1/instance_types.go
justinsb: If we put it back in the same package we end up with circular dependencies again.

In #7621 we're going to export all the fields anyway.  There's no reason to "hide" the Identity schema (IMO).

apis/memorystore/memorystorerefs/instance_identity.go
justinsb: Right - we could merge #7621 first if you'd prefer

apis/memorystore/generate.sh
justinsb: It was intended: once we have refs separately it becomes more obvious that we only need one generate.sh per API version.

However, if we put the refs into `apis/refs` then I can put this back as it was before.  We can always clean this up later.

go.mod
```diff
@@ -336,6 +336,5 @@ replace (
 	k8s.io/klog/v2 => k8s.io/klog/v2 v2.130.1
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20250318190949-c8a335a9a2ff
 	k8s.io/kubectl => k8s.io/kubectl v0.32.1
-	sigs.k8s.io/structured-merge-diff/v6 => sigs.k8s.io/structured-merge-diff/v4 v4.6.0
```
justinsb, line 339: I don't know why we're pinning so much stuff here, but this alias is very wrong.

ldanielmadariaga (approved):

PR #7644, opened 2026-04-21
dcl: report structured diff on resource creation

anhdle-sso:
This PR adds structured diff reporting for DCL resources during creation.
### Changes
- In `pkg/controller/dcl/controller.go`, added a check for `liveLite == nil` just before calling `dclunstruct.Apply`.
- If `liveLite` is nil (indicating the resource does not exist in GCP), it marshals the resource as unstructured and reports it as a new object diff using `structuredreporting.ReportDiff`.
### Rationale
This improves observability by ensuring that creation events are also logged with structured diffs, consistent with recent changes in direct controllers.


cheftako (approved):
Looks good.

PR #7647, opened 2026-04-21
Implement mockGCP for ComputeFutureReservation and add crud test

gemmahou:
### BRIEF Change description

- Update mockgcp's googleapis dependency for compute to support ComputeFutureReservation in compute/v1
Relevant PR: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/7494
Relevant issue: https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/7426
- Add crud test, collect realGCP and mockGCP logs.
- Improve mock service fidelity by comparing with realGCP log.

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/mockgcp/mockcompute/futurereservationsv1.go
+++ b/mockgcp/mockcompute/futurereservationsv1.go
@@ -1,0 +1,265 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mockcompute
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/proto"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/common/projects"
+	pb "github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/generated/mockgcp/cloud/compute/v1"
+	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/pkg/storage"
+)
+
+type FutureReservationsV1 struct {
+	*MockService
+	pb.UnimplementedFutureReservationsServer
+}
+
+func (s *FutureReservationsV1) Get(ctx context.Context, req *pb.GetFutureReservationRequest) (*pb.FutureReservation, error) {
+	reqName := "projects/" + req.GetProject() + "/zones/" + req.GetZone() + "/futureReservations/" + req.GetFutureReservation()
+	name, err := s.parseFutureReservationName(reqName)
+	if err != nil {
+		return nil, err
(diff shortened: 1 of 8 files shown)


mockgcp/mockcompute/futurereservationsv1.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+func (n *futureReservationName) String() string {
+	return "projects/" + n.Project.ID + "/zones/" + n.Zone + "/futureReservations/" + n.Name
+}
+
+func (s *MockService) parseFutureReservationName(name string) (*futureReservationName, error) {
+	tokens := strings.Split(name, "/")
+
+	if len(tokens) == 6 && tokens[0] == "projects" && tokens[2] == "zones" && tokens[4] == "futureReservations" {
```
ldanielmadariaga, line 249: QQ do we usually hardcode this to zones? Most API resources should support either zones or regions

mockgcp/mockcompute/futurereservationsv1.go
```diff
@@ -0,0 +1,265 @@
...
+	"github.com/GoogleCloudPlatform/k8s-config-connector/mockgcp/pkg/storage"
+)
+
+type FutureReservationsV1 struct {
+	*MockService
+	pb.UnimplementedFutureReservationsServer
+}
+
+func (s *FutureReservationsV1) Get(ctx context.Context, req *pb.GetFutureReservationRequest) (*pb.FutureReservation, error) {
+	reqName := "projects/" + req.GetProject() + "/zones/" + req.GetZone() + "/futureReservations/" + req.GetFutureReservation()
```
ldanielmadariaga, line 37: I'm a bit confused by this pattern of roundtripping the resource name in these methods. Do we usually do this just to regrab the string we had in the first place? 

It seems cleaner to just validate the project (only real thing we validate given we're creating the format internally). We could avoid all this parsing and validation by just templating against regex string and returning an error if we fail to apply

mockgcp/mockcompute/futurereservationsv1.go
```diff
@@ -0,0 +1,265 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	fqn := name.String()
+
+	obj := &pb.FutureReservation{}
+	if err := s.storage.Get(ctx, fqn, obj); err != nil {
+		return nil, err
+	}
```
ldanielmadariaga, line 112: nit: inconsistent use of newlines after code blocks(sometimes there's a newline sometimes there's not)

ldanielmadariaga (approved):

mockgcp/mockcompute/futurereservationsv1.go
gemmahou: ComputeFutureReservation seems to be a zonal resource([REST doc](https://docs.cloud.google.com/compute/docs/reference/rest/v1/futureReservations/get)), while there are other resources like [BackendService](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/mockgcp/mockcompute/regionalbackendservicev1.go#L160) is a regional resource so we validate its name by hard-coding to "regions". This is to restrict the name format of the resource.

mockgcp/mockcompute/futurereservationsv1.go
gemmahou: I could be wrong, but I think the parseXxxName function was originally intended to validate the full resource URL in the request. For example, as seen in https://screenshot.googleplex.com/ApKudVhVx6WNn9Z.png, req.Name is a full URL string (e.g., "projects/projectid/locations/us/resources/resourceName"). 

For compute resources, I agree that it may not be necessary to reconstruct the string since we can already extract the partial values from the request. Given that this code was generated, I suspect Gemini simply followed the existing patterns found in other resources.

PR #7651, opened 2026-04-21
Add types, mappers, fuzzer for ComputeFutureReservation and generate v1alpha1 CRD

### LINKED ISSUE #7403: Create generate.sh and types.go files for ComputeFutureReservation
As part of creating a resource with direct controller for a new GCP service (Epic #7276), we need to create the Go types for `ComputeFutureReservation`.

Currently, `ComputeFutureReservation` is not a pre-existing KCC resource. The goal is to create the Go types in `apis/compute/v1alpha1/` so that we can test the the "direct" controller before we promote it to production ready.

### Instructions

1.  **Add to generate.sh**:
    Create `apis/compute/v1alpha1/generate.sh` to include `ComputeFutureReservation`.
    It likely maps to something like `google.cloud.compute.v1beta`.
    Example:
    ```bash
    go run . generate-types \
      --service google.cloud.compute.v1beta \
      --api-version compute.cnrm.cloud.google.com/v1alpha1 \
      --resource ComputeFutureReservation:FutureReservation
    ```

2.  **Generate Scaffolding**:
    Run `apis/compute/v1alpha1/generate.sh`. This should create `apis/compute/v1alpha1/futurereservation_types.go`.

3.  **Copyright Headers**:
    Ensure that new files have the correct copyright header:
    ```go
    // Copyright 2026 Google LLC
    ```
    Please do not change the copyright on existing files.

4.  **Labels**:
    Ensure the controller-runtime annotations adds the following CRD labels:
    ```go
    // +kubebuilder:metadata:labels="cnrm.cloud.google.com/managed-by-kcc=true"
    // +kubebuilder:metadata:labels="cnrm.cloud.google.com/system=true"
    // +kubebuilder:metadata:labels="cnrm.cloud.google.com/stability-level=alpha"
    ```

5. **References**:
   Ensure fields which store references to other KCC or GCP resources use the appropriate Reference type.
   The GCP proto field corresponding to a reference would contain can be found at https://docs.cloud.google.com/asset-inventory/docs/asset-names.
   KCC reference types are generally found in files like `apis/<group>/<version>/<name>_reference.go`.
   A sample of reference type would then be something like:
   ```go 
   // +kcc:proto=<proto-type>
   type <type> struct {
	// +optional
	<field-name>Ref *refs.<ref-type>Ref 
   ```

6.  **Status**:
    `status.observedGeneration` should be an `*int64`.

7. **Generate Mappers**:
   - Running `dev/tasks/generate-types-and-mappers` will generate the mapper code once the `apis/compute/v1alpha1/futurereservation_types.go` file is generating an equivalent CRD.
   - Run `make all-binary` to ensure the generated mapper code compiles. Please fix any issue discovered by this compilation.

This issue is part of Epic #7276.

gemmahou:
### BRIEF Change description

Cherry-picked from #7444 to bypass CLA check.

This PR creates the Go types for ComputeFutureReservation in apis/compute/v1alpha1/

Note: Our dependency [version](https://github.com/codebot-robot/k8s-config-connector/blob/master/apis/git.versions) need to be updated to support ComputeFutureReservation v1 proto. However, bumping the dependency to the latest [googleapis](https://github.com/googleapis/googleapis) SHA introduced irrelevant updates across several other resource types (e.g., "generate.sh" introduced new fields to existing resources).

I manually updated my local googleapis dependency and generated the ComputeFutureReservation types and mappers using compute v1 proto. This approach prevents the "generate.sh" script from inadvertently modifying existing resources.

Moving forward, we'll need to update our "apis/git.versions" and then using "generate.sh" to ensure all resources are properly updated.

Fixes #7403

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


apis/compute/v1alpha1/types.generated.go
```diff
@@ -120,3 +545,22 @@ type NetworkAttachmentConnectedEndpoint struct {
 	// +kcc:proto:field=google.cloud.compute.v1.NetworkAttachmentConnectedEndpoint.subnetwork_cidr_range
 	SubnetworkCIDRRange *string `json:"subnetworkCIDRRange,omitempty"`
 }
+
+// +kcc:proto=google.cloud.compute.v1.ShareSettings
+type ShareSettings struct {
+
+	// TODO: unsupported map type with key string and value message
```
gemmahou, line 530: `shareSettings.projectMap` is a `map<string, ShareSettingsProjectConfig>`, which is currently unsupported by our generation tool: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/7c1d4184ccfe6d863b1d32896b7221708897e746/dev/tools/controllerbuilder/pkg/codegen/typegenerator.go#L283

Related issue: https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/6944

`message` type could potentially cause infinite loop so we'll need to discuss this case-by case and implement the field manually. Also in a previous resource ComputeReservation, the key of the project map is a `reference` rather than `string`([keyRef](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_computereservations.compute.cnrm.cloud.google.com.yaml#L77)). I think it's better to leave projectMap field as a TODO and implement it in a separate PR.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_computefuturereservations.compute.cnrm.cloud.google.com.yaml
```diff
@@ -0,0 +1,668 @@
...
+                      properties:
+                        accelerator:
+                          description: Properties of accelerator resources in this
+                            reservation.
+                          properties:
+                            acceleratorCount:
+                              description: Number of accelerators of specified type.
+                              format: int32
+                              type: integer
+                            acceleratorType:
```
ldanielmadariaga: Optional / possible follow-up: it'd be interesting to consider whether this should be a ref. I'm fine keeping as is given these resources don't really have an UID.

Ditto to other usages of acceleratorType

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_computefuturereservations.compute.cnrm.cloud.google.com.yaml
```diff
@@ -0,0 +1,668 @@
...
+                type: string
+              resourceID:
+                description: The ComputeFutureReservation name. If not given, the
+                  metadata.name will be used.
+                type: string
+              schedulingType:
+                description: Maintenance information for this reservation Check the
+                  SchedulingType enum for the list of possible values.
+                type: string
+              shareSettings:
```
ldanielmadariaga, line 233: We seem to be missing the project map and project list for shareSettings

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_computefuturereservations.compute.cnrm.cloud.google.com.yaml
```diff
@@ -0,0 +1,668 @@
...
+                description: Indicates whether the auto-created reservation can be
+                  consumed by VMs with affinity for "any" reservation. If the field
+                  is set, then only VMs that target the reservation by name can consume
+                  from the delivered reservation.
+                type: boolean
+              specificSkuProperties:
+                description: Future Reservation configuration to indicate instance
+                  properties and total count.
+                properties:
+                  instanceProperties:
```
ldanielmadariaga, line 251: We seem to be missing the following fields from `instanceProperties`: `maintenanceFreezeDurationHours, maintenanceInterval`
Ref: https://docs.cloud.google.com/compute/docs/reference/rest/beta/futureReservations/get

apis/compute/v1alpha1/futurereservation_types.go
```diff
@@ -0,0 +1,224 @@
...
+	ProjectRef *refsv1beta1.ProjectRef `json:"projectRef"`
+	// +required
+	Location string `json:"location"`
+}
+
+// ComputeFutureReservationSpec defines the desired state of ComputeFutureReservation
+// +kcc:spec:proto=google.cloud.compute.v1.FutureReservation
+type ComputeFutureReservationSpec struct {
+	// Parent reference.
+	ComputeFutureReservationParent `json:",inline"`
```
ldanielmadariaga, line 37: I don't think I've seen this parent before / in other _types.go files do we wanto to replace this with `ProjectRef refs.ProjectRef` and `Location *string`?

pkg/controller/direct/compute/futurereservation_mapper.go
```diff
@@ -0,0 +1,336 @@
...
+
+func ComputeFutureReservationObservedState_v1alpha1_FromProto(mapCtx *direct.MapContext, in *pb.FutureReservation) *krmcomputev1alpha1.ComputeFutureReservationObservedState {
+	if in == nil {
+		return nil
+	}
+	out := &krmcomputev1alpha1.ComputeFutureReservationObservedState{}
+	out.CreationTimestamp = in.CreationTimestamp
+	out.ID = in.Id
+	out.Kind = in.Kind
+	// MISSING: Name
```
ldanielmadariaga, line 32: QQ do we need to set Name to report back the UID?  Ditto below (multiple instances)

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_computefuturereservations.compute.cnrm.cloud.google.com.yaml
gemmahou: We have a crds_test to detect if a field should be a reference field: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/b5de9042c8675cf07f81629053bbd5ea7236e02f/tests/apichecks/crds_test.go#L53

This field is excluded becasue: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/b5de9042c8675cf07f81629053bbd5ea7236e02f/tests/apichecks/crds_test.go#L126

apis/compute/v1alpha1/futurereservation_types.go
gemmahou: We do have this recommended pattern for parent: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/b5de9042c8675cf07f81629053bbd5ea7236e02f/tests/apichecks/crds_test.go#L126, it's not applied to all resources, but new resources are following this, for example: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/compute/v1alpha1/networkedgesecurityservice_types.go#L26-L31

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_computefuturereservations.compute.cnrm.cloud.google.com.yaml
gemmahou: These fields are available in compute/beta not compute/v1: https://docs.cloud.google.com/compute/docs/reference/rest/v1/futureReservations/get

I'm generating the KRM types using compute/v1 proto because 
1. Existing compute resources are all using compute/v1, this resource is available in v1
2. Only adding this resource using compute/beta would introduce unnecessary complexity

I know we should only support GA/stable API, but since we are adding this resource as v1alpha1, and if user requires these specific fields, we may want to discuss if we should use compute/beta proto. There's another resource using beta API([dataflow v1beta3](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/dataflow/v1beta1/generate.sh#L28)), but seems v1beta3 is the only public version.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_computefuturereservations.compute.cnrm.cloud.google.com.yaml
gemmahou: I want to address "project map" as a follow-up, because it's a map, both key and value seem to be a reference(projectRef).

Regarding project list, it's a field in compute/beta and not in compute/v1. Depends on the proto version we want to use.

pkg/controller/direct/compute/futurereservation_mapper.go
gemmahou: Context: We use spec.resourceID or status.externalRef as the name or identifier of the resource. So we usually do not have Name field in our types and CRD.

I want to reference to this PR https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4136, the conductor has been deprecated but you can see that during types and CRD generation, we remove Name field.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_computefuturereservations.compute.cnrm.cloud.google.com.yaml
maqiuyujoyce: I'm in general advocating we support Beta unless we know the service team will definitely graduate features to GA (guaranteed by hard policy, etc). I understand that some of our Compute resources are already supported based on Compute v1, but I think we may want to be flexible about the exact GCP API versions we use per resource (not per Service).

ldanielmadariaga (dismissed):

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_computefuturereservations.compute.cnrm.cloud.google.com.yaml
```diff
@@ -0,0 +1,668 @@
...
+            type: object
+          spec:
+            description: ComputeFutureReservationSpec defines the desired state of
+              ComputeFutureReservation
+            properties:
+              aggregateReservation:
+                description: Aggregate reservation details for the future reservation.
+                properties:
+                  inUseResources:
+                    description: Output only. [Output only] List of resources currently
```
maqiuyujoyce: This field is output only. Could you move it to status.observedState?

maqiuyujoyce (commented):
Could you fix the output only spec field? Otherwise the PR looks good.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_computefuturereservations.compute.cnrm.cloud.google.com.yaml
gemmahou: Done.

maqiuyujoyce (dismissed):
/lgtm

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_computefuturereservations.compute.cnrm.cloud.google.com.yaml
gemmahou: Discussed offline, we'll stick with compute/v1 for this resource to keep version consistency with other existing direct compute resources.

maqiuyujoyce (approved):
/lgtm

PR #7657, opened 2026-04-22
Fix CertificateManagerDNSAuthorization diff issue

himanigulati01:
Suppress unexpected label updates in DNSAuthorization controller to prevent operational reconciliation storms during upgrades. Fixes b/500747761.


anhdle-sso:
question: is the `managed-by-cnrm` label allowed by this API? If that is the case, we do not need the compare ignoring logic. One time reconcile will add this label to the gcp resource. Then the next time, the diff would be gone.

anhdle-sso (changes requested):
Please check the comment

PR #7664, opened 2026-04-23
Fix infinite ForceNew diff loops when remove-default-node-pool is true

himanigulati01:
### Problem
When using the annotation `cnrm.cloud.google.com/remove-default-node-pool: "true"` on a `ContainerCluster`, if the user's YAML manifest still contains properties mapping to the initial pool (like `nodeConfig`), it can experience unresolvable reconciliation loops or errors on subsequent reconciliations. The typical error seen is: `node_version can only be specified if remove_default_node_pool is not true`.

### Root Cause
1.  The user specifies `nodeConfig` in YAML (often to satisfy organizational policies during initial creation).
2.  GKE creates the cluster and then KCC removes the default node pool as requested.
3.  On subsequent reconciliations, Terraform sees `nodeConfig` in the desired config (from YAML) but absent in the live state (since the pool was removed).
4.  Since `nodeConfig` is marked as `ForceNew` (immutable) in the Terraform schema, Terraform flags this as a change requiring recreation (`RequiresNew`).
5.  Terraform clears the state (`d.state = nil`) before invoking `CustomizeDiff`.
6.  The empty state tricks the validation logic into thinking this is a brand-new cluster creation, which then throws the error because both `node_version` (computed and written back) and `remove_default_node_pool` (true) appear to be set for a new cluster.

### Solution
To user, where they must specify `nodeConfig` during initial creation but want to avoid diff loops later, we introduce a new annotation:
`cnrm.cloud.google.com/remove-default-node-pool-allow-node-config: "true"`

When this annotation is set to `"true"` and the resource already exists (live state is not empty) but `nodeConfig` is missing from the live state (meaning the default pool was removed):
KCC will programmatically remove `nodeConfig` from the desired configuration before invoking Terraform in `legacygcpmanagedfields.go`.

This effectively suppresses the diff and allows the resource to reconcile successfully.

Fixes internal ticket b/500804917.

### DIFF
--- a/pkg/krmtotf/krmtotf.go
+++ b/pkg/krmtotf/krmtotf.go
@@ -61,12 +61,19 @@ func KRMResourceToTFResourceConfigFull(r *Resource, c client.Client, smLoader *s
 		config = make(map[string]interface{})
 	}
 	if jsonSchema != nil {
-		if err := ResolveLegacyGCPManagedFields(r, liveState, config); err != nil {
-			return nil, nil, fmt.Errorf("error resolving legacy GCP-managed fields: %w", err)
+		if r.GroupVersionKind().Kind != "ContainerCluster" {
+			if err := ResolveLegacyGCPManagedFields(r, liveState, config); err != nil {
+				return nil, nil, fmt.Errorf("error resolving legacy GCP-managed fields: %w", err)
+			}
 		}
 		config, err = resolveUnmanagedFields(config, r, liveState, jsonSchema)
 		if err != nil {
 			return nil, nil, fmt.Errorf("error resolving externally-managed fields: %w", err)
+		}
+		if r.GroupVersionKind().Kind == "ContainerCluster" {
+			if err := ResolveLegacyGCPManagedFields(r, liveState, config); err != nil {
+				return nil, nil, fmt.Errorf("error resolving legacy GCP-managed fields: %w", err)
+			}
 		}
 	}
 	if err := handleUserSpecifiedID(config, r, smLoader, c); err != nil {
--- a/pkg/krmtotf/legacygcpmanagedfields.go
+++ b/pkg/krmtotf/legacygcpmanagedfields.go
@@ -118,12 +118,15 @@ func resolveContainerClusterNodeVersion(r *Resource, config map[string]interface
 	releaseChannelSet := found && releaseChannel != nil
 	removeDefaultNodePoolSet := removeDefaultNodePoolFound && removeDefaultNodePoolVal == "true"
 
+	fmt.Printf("DEBUG: resolveContainerClusterNodeVersion called for %s, removeDefaultNodePoolSet: %v, config has nodeVersion: %v\n", r.GetName(), removeDefaultNodePoolSet, config["nodeVersion"])
+
 	if !releaseChannelSet && !removeDefaultNodePoolSet {
 		return nil
 	}
 	if err := removeFromConfigIfNotApplied(r, config, "nodeVersion"); err != nil {
 		return fmt.Errorf("error resolving node version in config: %w", err)
 	}
+	fmt.Printf("DEBUG: after removeFromConfigIfNotApplied, config has nodeVersion: %v\n", config["nodeVersion"])
 	return nil
 }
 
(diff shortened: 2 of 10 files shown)


pkg/krmtotf/legacygcpmanagedfields.go
```diff
@@ -168,6 +176,13 @@ func resolveContainerClusterNodeConfig(r *Resource, liveState *terraform.Instanc
 		return nil
 	}
 
+	allowKey := k8s.FormatAnnotation("remove-default-node-pool-allow-node-config")
+	if allowVal, ok := k8s.GetAnnotation(allowKey, r); ok && allowVal == "true" {
+		unstructured.RemoveNestedField(config, nodeConfigFieldInKRMConfig)
+		unstructured.RemoveNestedField(config, "nodeVersion")
```
gemmahou: I feel remove "nodeConfigFieldInKRMConfig" is sufficient. Having "nodeVersion" is expected and it will only trigger an error during the initial resource creation not update.

pkg/krmtotf/krmtotf.go
```diff
@@ -69,7 +69,14 @@ func KRMResourceToTFResourceConfigFull(r *Resource, c client.Client, smLoader *s
 			return nil, nil, fmt.Errorf("error resolving externally-managed fields: %w", err)
 		}
 	}
+	allowKey := k8s.FormatAnnotation("remove-default-node-pool-allow-node-config")
+
+	// Bridge between the K8s annotation and the Terraform schema
+	if allowVal, ok := k8s.GetAnnotation(allowKey, r); ok && allowVal == "true" {
+		config["remove_default_node_pool_allow_node_config"] = true
+	}
```
gemmahou: Is this bridge necessary? in "legacygcpmanagedfields.go" we already checked for certain annotation before converting to TF.

pkg/test/resourcefixture/testdata/directives/removedefaultnodepool/create.yaml
```diff
@@ -25,4 +25,7 @@ spec:
   location: us-central1-a
   initialNodeCount: 1
   loggingService: logging.googleapis.com/kubernetes
-  monitoringService: monitoring.googleapis.com/kubernetes
\ No newline at end of file
+  monitoringService: monitoring.googleapis.com/kubernetes
+  nodeConfig:
```
gemmahou: This test may not help us verify the fix. I think it's better to leverage the scenarios test, feel free to revert this.

tests/e2e/testdata/scenarios/containercluster/removedefaultnodepool/script.yaml
```diff
@@ -0,0 +1,57 @@
...
+# Step 2: Trigger a reconcile.
+# Without the allow-node-config annotation, this should fail with the diff error.
+# We use APPLY-60-SEC to apply it and wait a bit, then it will be captured in golden file.
+TEST: APPLY-60-SEC
+apiVersion: container.cnrm.cloud.google.com/v1beta1
+kind: ContainerCluster
+metadata:
+  annotations:
+    cnrm.cloud.google.com/remove-default-node-pool: "true"
+    kubectl.kubernetes.io/last-applied-configuration: |
```
gemmahou, line 26: Great find on using the "last-applied-configuration"! It worked perfectly to simulate the failure, thanks for digging into that.

tests/e2e/testdata/scenarios/containercluster/removedefaultnodepool/script.yaml
```diff
@@ -0,0 +1,57 @@
+# Step 1: Create ContainerCluster with remove-default-node-pool: "true"
+# We use APPLY to wait for the cluster to become Ready in real GCP
+TEST: APPLY
```
gemmahou: This step seems redundant with Step 2. Since this is the initial resource creation, can we remove 'last-applied-configuration' here, and verify the resource is created successfully in the beginning?

tests/e2e/testdata/scenarios/containercluster/removedefaultnodepool/script.yaml
```diff
@@ -0,0 +1,57 @@
...
+  location: us-central1-a
+  initialNodeCount: 1
+  nodeConfig:
+    metadata:
+      dev: "true"
+---
+# Step 2: Trigger a reconcile.
+# Without the allow-node-config annotation, this should fail with the diff error.
+# We use APPLY-60-SEC to apply it and wait a bit, then it will be captured in golden file.
+TEST: APPLY-60-SEC
```
gemmahou: nit: I know we have "APPLY-10-SEC" for failed resources, is 10 sec too short to have the error message populated for this resource?

tests/e2e/testdata/scenarios/containercluster/removedefaultnodepool/script.yaml
```diff
@@ -0,0 +1,57 @@
+# Step 1: Create ContainerCluster with remove-default-node-pool: "true"
```
gemmahou: super nit: I usually use step00, step01, etc in comment so it matches our log number!

pkg/krmtotf/legacygcpmanagedfields.go
gemmahou, line 1: Can we update the comment of the functions with the code change, and add unit tests in pkg/krmtotf/legacygcpmanagedfields_test.go?

Feel free to check #7235 for reference.

himanigulati01:
More Progress is being tracked here: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/8651

FYI, And currently we are trying to fix the mockgcp to be able to run the test successfully, though it is running fine with the real gcp.

PR #7677, opened 2026-04-24
Add controllers/test cases for MemorystoreInstanceEndpoint

suwandim:
### BRIEF Change description

This implements the types/fuzzers/mappers related to endpoint connections for Memorystore Instances.

Needed for controllers of the MemorystoreInstanceEndpoint types.

```release-note
NONE
```

```docs

```

#### Intended Milestone

### Tests you have done


apis/compute/v1beta1/forwardingrule_types.go
```diff
@@ -116,6 +116,10 @@ type Target struct {
 // for the ServiceAttachment managed by the MemorystoreInstance pointed by the MemorystoreInstanceRef.
 // +k8s:deepcopy-gen=true
 type MemorystoreInstanceServiceAttachment struct {
+	// A reference to a ServiceAttachment managed by a MemorystoreInstance resource.
+	// +optional
+	ServiceAttachmentExternal string `json:"serviceAttachmentExternal,omitempty"`
```
justinsb: Can we connect to a serviceAttachment that is _not_ from memorystore?

pkg/controller/direct/compute/forwardingrule_resolverefs.go
```diff
@@ -428,6 +429,65 @@ func ResolveComputeTargetVPNGateway(ctx context.Context, reader client.Reader, s
...
+	}
+
+	key := types.NamespacedName{
+		Namespace: ref.MemorystoreInstanceRef.Namespace,
+		Name:      ref.MemorystoreInstanceRef.Name,
+	}
+	if key.Namespace == "" {
+		key.Namespace = src.GetNamespace()
+	}
+	instance, err := resolveResourceName(ctx, reader, key, schema.GroupVersionKind{
```
justinsb: I think we can read the typed object - should be easier than using unstructured

pkg/controller/direct/memorystore/endpoint_controller.go
```diff
@@ -0,0 +1,339 @@
...
+					return err
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func (m *modelEndpoint) AdapterForURL(ctx context.Context, url string) (directbase.Adapter, error) {
+	// TODO: Support URLs
```
justinsb: I actually don't think there is a TODO here, we can't export these as they aren't a "real" GCP resource.  (That's what AdapterForURL is about)

pkg/controller/direct/memorystore/endpoint_controller.go
```diff
@@ -0,0 +1,339 @@
...
+	instancepb, err := a.gcpClient.GetInstance(ctx, req)
+	if err != nil {
+		if direct.IsNotFound(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("getting Instance %q: %w", a.id, err)
+	}
+
+	a.actual = instancepb
+	return true, nil
```
justinsb: I think we need to check if the _endpoint_ is found on the object.  We normally return true/false based on whether the object exists or not.

In this case, we might just always return false and always go through the Create method (or always return true and always go through the Update method).

pkg/controller/direct/memorystore/endpoint_controller.go
```diff
@@ -0,0 +1,339 @@
...
+}
+
+// Export maps the GCP object to a Config Connector resource `spec`.
+func (a *EndpointAdapter) Export(ctx context.Context) (*unstructured.Unstructured, error) {
+	if a.actual == nil {
+		return nil, fmt.Errorf("Find() not called")
+	}
+	u := &unstructured.Unstructured{}
+
+	obj := &krm.MemorystoreInstanceEndpoint{}
```
justinsb: I think if we implemented export, it would probably be on the MemorystoreInstance.  Today Export can only return one object, I think in this case a MemorystoreInstance maps to multiple objects.

I don't think Export can be called today, because I think Export is only called when AdapterForURL returns non-nil

pkg/controller/direct/memorystore/endpoint_controller.go
```diff
@@ -0,0 +1,339 @@
...
+	u.Object = uObj
+	return u, nil
+}
+
+// Delete the resource from GCP service when the corresponding Config Connector resource is deleted.
+func (a *EndpointAdapter) Delete(ctx context.Context, deleteOp *directbase.DeleteOperation) (bool, error) {
+	log := klog.FromContext(ctx)
+	log.V(2).Info("deleting user created instance endoints", "name", a.id)
+
+	if _, err := a.updateConnections(ctx, nil, deleteOp.GetUnstructured()); err != nil {
```
justinsb, line 220: Nit: might be easier to implement deleteConnection separately from adding connections .. up to you

pkg/test/resourcefixture/testdata/basic/memorystore/v1alpha1/memorystoreinstanceendpoint/basicmemorystoreinstanceendpoint/update.yaml
```diff
@@ -0,0 +1,36 @@
...
+apiVersion: memorystore.cnrm.cloud.google.com/v1alpha1
+kind: MemorystoreInstanceEndpoint
+metadata:
+  name: memorystoreinstanceendpoint-1-${uniqueId}
+spec:
+  endpoints:
+    - connections:
+      - pscConnection:
+            forwardingRuleRef:
+              name: computeforwardingrule-1a-${uniqueId}
```
justinsb, line 24: Dumb question: we create the forwarding rule and point it at the memorystore instance.  AIUI the forwarding rule allows access across projects, from a networking perspective.  Why do I also need to register the forwarding rule with the MemorystoreInstance ... what does this actually do beyond creating the forwarding rule?

apis/compute/v1beta1/forwardingrule_types.go
justinsb: Sorry, I now understand better what this is.  If I want to connect to an arbitrary serviceAttachment from a ComputeForwardingRule, I think I would set target.serviceAttachmentRef.  So I don't think we need this field?

suwandim: How do I keep a cache of this value? Do I have to retrieve it from the MemorystoreInstance object every single time?

pkg/controller/direct/memorystore/endpoint_controller.go
suwandim: Ah ok. I will fix the comment.

pkg/controller/direct/memorystore/endpoint_controller.go
suwandim: Got it. Will update the implementation.

apis/compute/v1beta1/forwardingrule_types.go
justinsb: You should just retrieve it, yes.  The client _might_ actually do some magic caching behind the scenes; we've turned it off for scalability reasons (memory usage) but the key point is that you don't need to worry about whether it's cached or not.

pkg/controller/direct/memorystore/endpoint_controller.go
suwandim: I will check if the endpoint is found on the object. I feel that this choice will make it consistent with the Create/Update/Delete semantics.

pkg/controller/direct/memorystore/endpoint_controller.go
suwandim: I will leave it as is.

pkg/test/resourcefixture/testdata/basic/memorystore/v1alpha1/memorystoreinstanceendpoint/basicmemorystoreinstanceendpoint/update.yaml
suwandim: This is to assert ownership of the `ForwardingRule` resources; once they are registered to a `MemorystoreInstance` resource, they must first be deleted before the `MemorystoreInstance` resource.

justinsb (approved):

PR #7680, opened 2026-04-27
Revert "feat: Add support for Shared Reservations in ComputeReservation"

gemmahou:
Reverts GoogleCloudPlatform/k8s-config-connector#7278

`shareSettings` implementation in TF is problematic, instead of fixing in TF, it's better to migrate this resource to direct controller.

Keep the changes that unrelated to `shareSettings`:
- mockGCP for basic ComputeReservation
- test case and golden logs for basic ComputeReservation

Quote Joyce's investigation for future reference:
```
1. Diffs in re-reconciliation without any user changes
a. The configured values in shareSettings.projectMap field can be either projectIDs or projectNumbers, but the returned values are all projectNumbers. This means there could always be diffs found in re-reconciliation.
b. In the TF-based controller, we usually can handle this projectID / projectNumber diff by marking the field mutable but unreadable, so that we can filter the noisy, constant, quota-consuming diffs. However, projectMap is a list and our exiting mutable but unreadable logic doesn't handle list fields.

2. Can't handle entry replacement in the projectMap list
After some testing, I found that when an entry is replaced in the projectMap list, it is treated as one addition and one removal. However, a single PATCH doesn't handle both addition and removal at the same time, thus returned this error: 
Only adding or removing from project list is allowed per each update command.

3. Update logic incorrect
shareSettings.shareType is literally hardcoded in the Update function: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/v1.149.1/third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/compute/resource_compute_reservation.go#L1109, this means as long as projectMap is updated, we assume shareType must be SPECIFIC_PROJECTS. While the call may still error out, it may success when it is an mismatching acquisition, and will hide user errors.

4. The map schema
We don't want keyRef and subfields in value object to be at the same level.
```


gemmahou:
cc @cheftako @ldanielmadariaga, after investigation we feel it's better to support `shareSettings` field in direct controller instead of patching TF code. So I'm reverting this PR.

maqiuyujoyce (approved):
/lgtm

Issue #6828, opened 2026-02-26
[New Resource] Support for Compute Engine Future Reservations


fkc1e100:
### Description
We would like to request support for Google Compute Engine Future Reservations in Config Connector. Future Reservations allow users to request and secure compute capacity for a specific time window in the future, ensuring that the necessary resources are guaranteed to be available when needed.

This feature is highly beneficial for workloads with predictable capacity needs (such as batch processing, ML training runs, or scheduled events) and helps in managing long-term cloud spend efficiently.

Currently, this resource is not supported in the KCC provider. We would like a new CRD (e.g., `ComputeFutureReservation`) to be implemented using the direct controller approach.

### References
*   **GCP API Reference:** [Compute Engine Future Reservations REST API](https://cloud.google.com/compute/docs/reference/rest/v1/futureReservations)
*   **GCP Product Documentation:** [About future reservations](https://cloud.google.com/compute/docs/instances/reserving-zonal-resources#future)

### Resource Mapping
*   **Proposed KCC CRD:** `ComputeFutureReservation`
*   **GCP Service:** `compute.googleapis.com`
*   **API Version:** `v1beta1` (as seen in recent implementations)

---
*Note: This feature request is being addressed by the implementation proposed in PR [#4801](https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/4801).*

gemmahou:
Please don't generate PRs for this issue. We’re using this issue for high-level FR tracking. Actual implementation and code reviews will take place in separate dedicated issues.

gemmahou:
This is code complete. New direct controller for ComputeFutureReservation(v1alpha1) will be available in next KCC release.

fkc1e100:
Merged into the 1.151 release per https://github.com/GoogleCloudPlatform/k8s-config-connector/releases

PR #7723, opened 2026-04-29
feat: Add direct controller for ComputeFutureReservation

gemmahou:
### BRIEF Change description

Addresses #6828 

Add direct controller for ComputeFutureReservation and fixtures tests

- Extended reservation(Calendar mode or AI Hypercomputer) requires TAM approval, more details can be found at: https://docs.cloud.google.com/compute/docs/instances/choose-reservation-type. Excluded fields that will trigger an extended reservation in test for now. Those fields are all primitive types so I don't expect issues during creation.
- If "duration" is specified, GCP API will convert it into a timestamp. Added a similar conversion logic in direct controller Update function when comparing desired and live state.
- Some boolean fields will be cleared by the GCP API when they are "false". Added a similar logic to clear these fields  in direct controller Update function when comparing desired and live state.

```release-note
New Alpha Resources (Direct Reconciler): ComputeFutureReservation
```

```docs

```

#### Intended Milestone

### Tests you have done


pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -0,0 +1,401 @@
...
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+	resource.Name = direct.LazyPtr(a.id.ID())
+
+	// Assign API output-only values
+	resource.CreationTimestamp = a.actual.CreationTimestamp
+	resource.Id = a.actual.Id
+	resource.Kind = a.actual.Kind
+	resource.Name = direct.LazyPtr(a.id.ID())
```
gemmahou: Redundant assignment. `resource.Name` was already assigned slightly above on line 183.

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -0,0 +1,401 @@
...
+	resource.SelfLink = a.actual.SelfLink
+	resource.SelfLinkWithId = a.actual.SelfLinkWithId
+	resource.Zone = a.actual.Zone
+	resource.Status = a.actual.Status
+	if resource.PlanningStatus == nil {
+		resource.PlanningStatus = a.actual.PlanningStatus
+	}
+
+	// Assign API default(nil) values
+	if resource.GetAutoDeleteAutoCreatedReservations() == true {
```
gemmahou: The PR description notes that some boolean fields are cleared when they are 'false', but this logic clears `AutoDeleteAutoCreatedReservations` when it is `true`. Is this intended? If the goal is to clear it when it is false, the condition should be `if resource.AutoDeleteAutoCreatedReservations != nil && !*resource.AutoDeleteAutoCreatedReservations {`.

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -0,0 +1,401 @@
...
+		resource.AutoDeleteAutoCreatedReservations = nil
+	}
+	if resource.ShareSettings != nil && resource.ShareSettings.GetShareType() == "LOCAL" {
+		resource.ShareSettings = nil
+	}
+	if resource.GetSchedulingType() == "FLEXIBLE" {
```
gemmahou: The mock HTTP handler (`futurereservationsv1.go`) clears `EnableEmergentMaintenance` when it is `false`. If the actual GCP API behaves this way, you should also clear `EnableEmergentMaintenance` here in the controller's `Update` function. Otherwise, if a user explicitly sets it to `false` in their YAML, it will cause continuous false diffs and unnecessary API updates.

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -0,0 +1,401 @@
...
+		resource.SchedulingType = nil
+	}
+
+	// handle endtime calculated by duration
+	timeWindowDuration := &computepb.Duration{}
```
gemmahou: If the user configures `Duration` with both `Seconds` and `Nanos` as `nil` (meaning zero duration), the `if timeWindowDuration.Seconds != nil || timeWindowDuration.Nanos != nil` check later will fail, and `resource.TimeWindow.Duration` won't be restored. Consider initializing as a pointer `var timeWindowDuration *computepb.Duration` instead of an empty struct.

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -0,0 +1,401 @@
...
+	if resource.TimeWindow != nil && resource.TimeWindow.Duration != nil {
+		timeWindowDuration = resource.TimeWindow.Duration
+		start, err := time.Parse(time.RFC3339Nano, direct.ValueOf(resource.TimeWindow.StartTime))
+		if err != nil {
+			panic(err)
```
gemmahou: CRITICAL: Do not use `panic` in a Kubernetes controller as it will crash the entire controller manager. Also, since `StartTime` is an optional field (`omitempty`), if it is not set by the user, `time.Parse` will fail on the empty string and trigger this panic. Please return an error instead (e.g., `return fmt.Errorf("parsing start time: %w", err)`).

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -0,0 +1,401 @@
...
+		}
+		duration := time.Duration(0)
+		if resource.TimeWindow.Duration.Seconds != nil {
+			duration = duration + time.Duration(direct.ValueOf(resource.TimeWindow.Duration.Seconds))*time.Second
+		}
+		if resource.TimeWindow.Duration.Nanos != nil {
+			duration = duration + time.Duration(direct.ValueOf(resource.TimeWindow.Duration.Nanos))*time.Nanosecond
+		}
+		end := start.Add(duration)
```
gemmahou: Formatting the calculated `end` time as a string and relying on `CompareProtoMessage` for string comparison is risky. If the format produced by `time.RFC3339Nano` differs slightly from what the GCP API returns (e.g., number of fractional digits), it will cause continuous false diffs and unnecessary API updates. Consider parsing `a.actual.TimeWindow.EndTime` and comparing the `time.Time` values directly using `time.Equal()`, and removing `time_window.end_time` from the paths if they match.

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -0,0 +1,401 @@
...
+		resource.TimeWindow.EndTime = direct.PtrTo(end.Format(time.RFC3339Nano))
+		resource.TimeWindow.Duration = nil
+	}
+
+	// handle AutoCreatedReservationsDeleteTime calculated by duration
+	autoCreatedReservationsDuration := &computepb.Duration{}
```
gemmahou: Consider using `var autoCreatedReservationsDuration *computepb.Duration` to correctly handle zero-duration edge cases when restoring the duration later.

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -0,0 +1,401 @@
...
+	if resource.AutoCreatedReservationsDuration != nil {
+		autoCreatedReservationsDuration = resource.AutoCreatedReservationsDuration
+		start, err := time.Parse(time.RFC3339Nano, direct.ValueOf(resource.TimeWindow.StartTime))
+		if err != nil {
+			panic(err)
```
gemmahou: CRITICAL: Do not use `panic`. Return `fmt.Errorf(...)` instead. Panicking will crash the controller manager if parsing fails or if the optional `StartTime` is omitted.

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -0,0 +1,401 @@
...
+
+	paths, err := common.CompareProtoMessage(resource, a.actual, common.BasicDiff)
+	if err != nil {
+		return err
+	}
+
+	// Change back to duration values after proto message comparison
+	// Error 400: Invalid value for field 'resource.timeWindow.endTime': '2026-05-11T00:00:01.000000101Z'.
+	// Must be a string representation of a full-date or date-time in valid RFC3339 format with either 0 or 3 digits for fractional seconds
+	if timeWindowDuration.Seconds != nil || timeWindowDuration.Nanos != nil {
```
gemmahou: If you change the initialization to a nil pointer (as suggested), update this check to `if timeWindowDuration != nil {`.

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -0,0 +1,401 @@
...
+		resource.TimeWindow.Duration = timeWindowDuration
+		resource.TimeWindow.EndTime = nil
+		if paths.Has("time_window.end_time") {
+			paths.Delete("time_window.end_time")
+			paths.Insert("time_window.duration")
+		}
+	}
+	if autoCreatedReservationsDuration.Seconds != nil || autoCreatedReservationsDuration.Nanos != nil {
```
gemmahou: If you change the initialization to a nil pointer, update this check to `if autoCreatedReservationsDuration != nil {`.

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -0,0 +1,401 @@
...
+	}
+
+	if len(paths) == 0 {
+		log.V(2).Info("no field needs update", "name", a.id)
+		status := &krm.ComputeFutureReservationStatus{}
+		status.ObservedState = ComputeFutureReservationObservedState_v1alpha1_FromProto(mapCtx, a.actual)
+		if mapCtx.Err() != nil {
+			return mapCtx.Err()
+		}
+		if a.desired.Status.ExternalRef == nil {
```
gemmahou, line 308: If `a.desired.Status.ExternalRef` is already set, `status.ExternalRef` will remain `nil` in the new `status` object. This could cause `ExternalRef` to be cleared during the status update. You should set `status.ExternalRef = direct.LazyPtr(a.id.String())` unconditionally.

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -0,0 +1,401 @@
...
+	if err != nil {
+		return fmt.Errorf("getting ComputeFutureReservation %s: %w", a.id, err)
+	}
+
+	status := &krm.ComputeFutureReservationStatus{}
+	status.ObservedState = ComputeFutureReservationObservedState_v1alpha1_FromProto(mapCtx, updated)
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+	if a.desired.Status.ExternalRef == nil {
```
gemmahou, line 347: Same issue here as on line 281: if `a.desired.Status.ExternalRef` is already set, it won't be assigned to `status.ExternalRef`, potentially clearing it from the status. Set it unconditionally.

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -0,0 +1,401 @@
...
+	}
+	u := &unstructured.Unstructured{}
+
+	obj := &krm.ComputeFutureReservation{}
+	mapCtx := &direct.MapContext{}
+	obj.Spec = direct.ValueOf(ComputeFutureReservationSpec_v1alpha1_FromProto(mapCtx, a.actual))
+	if mapCtx.Err() != nil {
+		return nil, mapCtx.Err()
+	}
+	// obj.Spec.ProjectRef = &refsv1beta1.ProjectRef{External: a.id.Parent().ProjectID}
```
gemmahou: This line is commented out, likely because `refsv1beta1` is not imported. Please import `"github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"` and uncomment this line to ensure the `ProjectRef` is properly exported and present in the generated YAML.

mockgcp/mockcompute/futurereservationsv1.go
```diff
@@ -93,6 +98,48 @@ func (s *FutureReservationsV1) Insert(ctx context.Context, req *pb.InsertFutureR
...
 	return s.startZonalLRO(ctx, name.Project.ID, name.Zone, op, func() (proto.Message, error) {
+		obj.Status.ExistingMatchingUsageInfo = &pb.FutureReservationStatusExistingMatchingUsageInfo{
+			Count:     PtrTo(int64(0)),
+			Timestamp: PtrTo(s.nowString()),
+		}
+		// handle endtime calculated by duration
+		if obj.TimeWindow != nil && obj.TimeWindow.Duration != nil {
+			start, err := time.Parse(time.RFC3339Nano, direct.ValueOf(obj.TimeWindow.StartTime))
+			if err != nil {
+				panic(err)
```
gemmahou: Do not use `panic` here. Return a gRPC error instead (e.g., `return nil, status.Errorf(codes.InvalidArgument, "invalid start_time: %v", err)`).

mockgcp/mockcompute/futurereservationsv1.go
```diff
@@ -93,6 +98,48 @@ func (s *FutureReservationsV1) Insert(ctx context.Context, req *pb.InsertFutureR
...
+			end := start.Add(duration)
+			obj.TimeWindow.EndTime = direct.PtrTo(end.Format(time.RFC3339Nano))
+			obj.TimeWindow.Duration = nil
+		}
+
+		// handle AutoCreatedReservationsDeleteTime calculated by duration
+		if obj.AutoCreatedReservationsDuration != nil {
+			start, err := time.Parse(time.RFC3339Nano, direct.ValueOf(obj.TimeWindow.StartTime))
+			if err != nil {
+				panic(err)
```
gemmahou: Do not use `panic` here. Return a gRPC error instead.

mockgcp/mockcompute/futurereservationsv1.go
```diff
@@ -110,17 +157,69 @@ func (s *FutureReservationsV1) Update(ctx context.Context, req *pb.UpdateFutureR
...
+
+	if obj.Description == nil {
+		obj.Description = direct.PtrTo("")
+	}
+
+	// handle endtime calculated by duration
 	if obj.TimeWindow != nil && obj.TimeWindow.Duration != nil {
+		start, err := time.Parse(time.RFC3339Nano, direct.ValueOf(obj.TimeWindow.StartTime))
+		if err != nil {
+			panic(err)
```
gemmahou: Do not use `panic` here. Return a gRPC error instead.

mockgcp/mockcompute/futurereservationsv1.go
```diff
@@ -110,17 +157,69 @@ func (s *FutureReservationsV1) Update(ctx context.Context, req *pb.UpdateFutureR
...
+		end := start.Add(duration)
+		obj.TimeWindow.EndTime = direct.PtrTo(end.Format(time.RFC3339Nano))
 		obj.TimeWindow.Duration = nil
 	}
 
+	// handle AutoCreatedReservationsDeleteTime calculated by duration
+	if obj.AutoCreatedReservationsDuration != nil {
+		start, err := time.Parse(time.RFC3339Nano, direct.ValueOf(obj.TimeWindow.StartTime))
+		if err != nil {
+			panic(err)
```
gemmahou: Do not use `panic` here. Return a gRPC error instead.

mockgcp/mockcompute/futurereservationsv1.go
```diff
@@ -263,3 +367,21 @@ func (s *MockService) parseFutureReservationName(name string) (*futureReservatio
 		return nil, status.Errorf(codes.InvalidArgument, "name %q is not valid", name)
 	}
 }
+
+func (s *FutureReservationsV1) clearValues(obj *pb.FutureReservation) {
+	if obj.GetAutoDeleteAutoCreatedReservations() == false {
```
gemmahou: In `clearValues()`, `AutoDeleteAutoCreatedReservations` is being cleared if it is `false` (here) and also if it is `true` (on line 378). This unconditionally clears the value. Please fix this logic to correctly reflect the actual GCP API behavior.

mockgcp/mockcompute/futurereservationsv1.go
```diff
@@ -263,3 +367,21 @@ func (s *MockService) parseFutureReservationName(name string) (*futureReservatio
...
+		obj.AutoDeleteAutoCreatedReservations = nil
+	}
+	if obj.GetEnableEmergentMaintenance() == false {
+		obj.EnableEmergentMaintenance = nil
+	}
+	if obj.GetAutoDeleteAutoCreatedReservations() == true {
```
gemmahou: Redundant check that unconditionally clears `AutoDeleteAutoCreatedReservations` when combined with line 372. Remove this or correct the logic to match the intended API behavior.

mockgcp/mockcompute/futurereservationsv1.go
```diff
@@ -110,17 +157,69 @@ func (s *FutureReservationsV1) Update(ctx context.Context, req *pb.UpdateFutureR
...
+			obj.TimeWindow.EndTime = req.GetFutureReservationResource().TimeWindow.EndTime
+		case "time_window.duration", "time_window.duration.seconds", "time_window.duration.nanos":
+			obj.TimeWindow.Duration = req.GetFutureReservationResource().TimeWindow.Duration
+		default:
+			return nil, status.Errorf(codes.InvalidArgument, "update_mask path %q not valid", path)
 		}
 	}
+
+	if obj.Description == nil {
+		obj.Description = direct.PtrTo("")
```
gemmahou: If the `Description` field is explicitly cleared by setting it to `""` during updates, is this mimicking GCP's exact behavior? Make sure `direct.PtrTo("")` accurately represents how GCP normalizes absent descriptions.

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -0,0 +1,401 @@
...
+	log.Info("successfully created compute FutureReservation in gcp", "name", a.id)
+
+	// Get the created resource
+	created, err := a.get(ctx)
+	if err != nil {
+		return fmt.Errorf("getting FutureReservation %s: %w", a.id, err)
+	}
+
+	status := &krm.ComputeFutureReservationStatus{}
+	status.ObservedState = ComputeFutureReservationObservedState_v1alpha1_FromProto(mapCtx, created)
```
gemmahou, line 173: The result of `ComputeFutureReservationObservedState_v1alpha1_FromProto` is assigned to `status.ObservedState`, but you are not checking `mapCtx.Err()` afterward. If the conversion accumulates mapping errors, they will be silently ignored. Please add an error check here.

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -0,0 +1,401 @@
...
+
+	// Change back to duration values after proto message comparison
+	// Error 400: Invalid value for field 'resource.timeWindow.endTime': '2026-05-11T00:00:01.000000101Z'.
+	// Must be a string representation of a full-date or date-time in valid RFC3339 format with either 0 or 3 digits for fractional seconds
+	if timeWindowDuration.Seconds != nil || timeWindowDuration.Nanos != nil {
+		resource.TimeWindow.Duration = timeWindowDuration
+		resource.TimeWindow.EndTime = nil
+		if paths.Has("time_window.end_time") {
+			paths.Delete("time_window.end_time")
+			paths.Insert("time_window.duration")
```
gemmahou: When removing `time_window.end_time` from the update mask and inserting `time_window.duration`, this assumes that the update mask accurately understands `time_window.duration`. Ensure the GCP API accepts `time_window.duration` in the update mask, since some APIs only accept scalar field paths.

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -0,0 +1,401 @@
...
+		}
+	}
+	if autoCreatedReservationsDuration.Seconds != nil || autoCreatedReservationsDuration.Nanos != nil {
+		resource.AutoCreatedReservationsDuration = autoCreatedReservationsDuration
+		resource.AutoCreatedReservationsDeleteTime = nil
+		if paths.Has("auto_created_reservations_delete_time") {
+			paths.Delete("auto_created_reservations_delete_time")
+			paths.Insert("auto_created_reservations_duration")
```
gemmahou: Same as above, ensure that `auto_created_reservations_duration` is a valid update mask path accepted by the GCP API.

mockgcp/mockcompute/futurereservationsv1.go
gemmahou: Yes, when comparing realGCP and mockGCP logs, if "description" is not set in the configuration, it will populated as empty string in the GET response. Added the code to simulate the realGCP.

pkg/controller/direct/compute/futurereservation_controller.go
gemmahou: Confirmed by running test against realGCP.

The actual updated field is "time_window.duration". GCP API converts duration into timestamp, so we have the same logic to prevent perm diff when comparing proto message. Here we change everything back before sending the update request to GCP.

pkg/controller/direct/compute/futurereservation_controller.go
gemmahou: Confirmed by running test against realGCP.

The actual updated field is "auto_created_reservations_duration". GCP API converts duration into timestamp, so we have the same logic to prevent perm diff when comparing proto message. Here we change everything back before sending the update request to GCP.

pkg/controller/direct/compute/futurereservation_controller.go
gemmahou: Updated  the code to handle AutoDeleteAutoCreatedReservations field

pkg/controller/direct/compute/futurereservation_controller.go
gemmahou: Updated the code to temporarily exclude these fields in "updateMask", as they are currently immutable. i.e. Changing SchedulingType from "FLEXIBLE" to another value will trigger "extended reservation"; ShareSettings feature will be followed up in another PR, so the only allowed ShareType is "LOCAL".

pkg/controller/direct/compute/futurereservation_controller.go
gemmahou: Updated.

pkg/controller/direct/compute/futurereservation_controller.go
gemmahou: ExternalRef will not be overridden because we keep its old value in [UpdateStatus()](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/directbase/operations.go#L173-L175)

pkg/controller/direct/compute/futurereservation_controller.go
gemmahou: ExternalRef will not be overridden because we keep its old value in [UpdateStatus()](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/directbase/operations.go#L173-L175)

apis/compute/v1alpha1/futurereservation_identity.go
ldanielmadariaga, line 1: QQ how do we plan to merge this with #7730? We seem to have very similar logic with different method names and slight deviations in code.

mockgcp/mockcompute/futurereservationsv1.go
```diff
@@ -110,17 +170,71 @@ func (s *FutureReservationsV1) Update(ctx context.Context, req *pb.UpdateFutureR
 	if err := s.storage.Get(ctx, fqn, obj); err != nil {
 		return nil, err
 	}
-	proto.Merge(obj, req.GetFutureReservationResource())
-	if obj.Status != nil {
-		obj.Status.ExistingMatchingUsageInfo = &pb.FutureReservationStatusExistingMatchingUsageInfo{
-			Count:     PtrTo(int64(0)),
-			Timestamp: PtrTo(s.nowString()),
+
+	paths := strings.Split(req.GetUpdateMask(), ",")
```
ldanielmadariaga, line 159: Is it common for us to hardcode field masks in mock gcp? Since this is testing we may have made the choice that it is acceptable but it's very brittle and very easy to get out of sync with real GCP. I think I'd rather err on the side of accepting most values.

mockgcp/mockcompute/futurereservationsv1.go
```diff
@@ -93,6 +95,64 @@ func (s *FutureReservationsV1) Insert(ctx context.Context, req *pb.InsertFutureR
...
+		if obj.GetSchedulingType() == "FLEXIBLE" {
+			obj.SchedulingType = nil
+		}
+
+		obj.Status.ExistingMatchingUsageInfo = &pb.FutureReservationStatusExistingMatchingUsageInfo{
+			Count:     PtrTo(int64(0)),
+			Timestamp: PtrTo(s.nowString()),
+		}
+		// handle endtime calculated by duration
+		if obj.TimeWindow != nil && obj.TimeWindow.Duration != nil {
```
ldanielmadariaga, line 120: There are 4 blocks parsing durations with roughly the same logic, in fact the logic seems to be identical between the 2 locations where this code is implemented. Let's extract a method here similar to:

```
func parseDuration(startTime, duration) (endTime, error) {
	start, err := time.Parse(time.RFC3339Nano, direct.ValueOf(startTime))
			if err != nil {
				return nil, status.Errorf(codes.InvalidArgument, "invalid start_time: %v", err)
			}

			duration := time.Duration(0)
			if duration.Seconds != nil {
				duration = duration + time.Duration(direct.ValueOf(duration.Seconds))*time.Second
			}
			if duration.Nanos != nil {
				duration = duration + time.Duration(direct.ValueOf(duration.Nanos))*time.Nanosecond
			}

			end := start.Add(duration)
			return end.Format(time.RFC3339Nano), nil
}
```

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -0,0 +1,449 @@
...
+	resource := ComputeFutureReservationSpec_v1alpha1_ToProto(mapCtx, &desired.Spec)
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+	resource.Name = direct.LazyPtr(a.id.ID())
+
+	updateMask := fieldmaskpb.FieldMask{}
+	report := &structuredreporting.Diff{Object: updateOp.GetUnstructured()}
+
+	// GCP API does not have AutoDeleteAutoCreatedReservations field in the response body.
```
ldanielmadariaga: The following reads easier and more closely resembles our standard normal `if desired != actual` checks  while also eliminating nesting / branching. It's fine to remove if/else for two unnested ifs as both conditions are mutually exclusive.

```
// We need to check if AutoCreatedReservationsDeleteTime is set in actual as the GCP API 
// read method does not populate AutoDeleteAutoCreatedReservations.
if resource.GetAutoDeleteAutoCreatedReservations() == false  && a.actual.AutoCreatedReservationsDeleteTime != nil {
		report.AddField("auto_delete_auto_created_reservations", "true",  resource.GetAutoDeleteAutoCreatedReservations())
			updateMask.Paths = append(updateMask.Paths, "auto_delete_auto_created_reservations")
}

if resource.GetAutoDeleteAutoCreatedReservations() == true  && if a.actual.AutoCreatedReservationsDeleteTime == nil {
		report.AddField("auto_delete_auto_created_reservations", "false", resource.GetAutoDeleteAutoCreatedReservations())
			updateMask.Paths = append(updateMask.Paths, "auto_delete_auto_created_reservations")
}
```

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -0,0 +1,449 @@
...
+		}
+	} else {
+		if a.actual.AutoCreatedReservationsDeleteTime == nil {
+			report.AddField("auto_delete_auto_created_reservations", "false", resource.GetAutoDeleteAutoCreatedReservations())
+			updateMask.Paths = append(updateMask.Paths, "auto_delete_auto_created_reservations")
+		}
+	}
+
+	// handle AutoCreatedReservationsDeleteTime calculated by duration
+	if resource.AutoCreatedReservationsDuration != nil {
```
ldanielmadariaga, line 210: There's probably some simplification that we could do to this code similar to my comment in the mock GCP file

apis/compute/v1alpha1/futurereservation_identity.go
gemmahou: Plan to rebase and align with #7730 after it's merged. I have these files earlier but the implementation in #7730 leverages a newer interface and includes better unit test coverage.

pkg/test/resourcefixture/testdata/basic/compute/v1alpha1/computefuturereservation/basicreservation-duration/_generated_object_basicreservation-duration.golden.yaml
```diff
@@ -38,5 +38,8 @@ status:
     selfLink: https://www.googleapis.com/compute/v1/projects/${projectId}/zones/us-central1-a/futureReservations/${futureReservationID}
     selfLinkWithID: https://www.googleapis.com/compute/v1/projects/${projectId}/zones/us-central1-a/futureReservations/${futureReservationID}
     status:
+      existingMatchingUsageInfo:
```
maqiuyujoyce, line 41: Looks like this value should not exist here?

tests/e2e/normalize.go
```diff
@@ -97,6 +97,7 @@ func buildKRMNormalizer(t *testing.T, u *unstructured.Unstructured, project test
 	visitor.replacePaths[".status.etag"] = "abcdef123456"
 	visitor.replacePaths[".status.observedState.etag"] = "abcdef123456"
 	visitor.replacePaths[".status.observedState.creationTimestamp"] = "1970-01-01T00:00:00Z"
+	visitor.replacePaths[".status.observedState.status.existingMatchingUsageInfo.timestamp"] = "1970-01-01T00:00:00Z"
```
maqiuyujoyce: Can we normalize it in the each service's individual normalize.go?

pkg/test/resourcefixture/testdata/basic/compute/v1alpha1/computefuturereservation/basicreservation-duration/update.yaml
```diff
@@ -0,0 +1,35 @@
...
+metadata:
+  name: computefuturereservation-${uniqueId}
+spec:
+  projectRef:
+    external: ${projectId}
+  location: us-central1-a
+  # Required field
+  autoDeleteAutoCreatedReservations: false
+  specificSkuProperties:
+    totalCount: 2
```
maqiuyujoyce, line 26: Can we have a test case to update the count to 0 and ensure setting to 0 to supported?

pkg/test/resourcefixture/testdata/basic/compute/v1alpha1/computefuturereservation/basicreservation-endtime/create.yaml
```diff
@@ -0,0 +1,34 @@
...
+  specificSkuProperties:
+    totalCount: 1
+    instanceProperties:
+      machineType: n1-standard-1
+  timeWindow:
+    # Future reservation must specify a future start time.
+    startTime: 2026-11-10T00:00:00Z
+    # Future reservation end time must be at least 24 hours later than start time.
+    # One of duration or endTime must be set.
+    endTime: 2026-11-11T00:00:00Z
```
maqiuyujoyce, line 34: Can we unset this field in the update to ensure we support unset behavior properly in the Update logic?

mockgcp/mockcompute/futurereservationsv1.go
gemmahou: Updated.

pkg/controller/direct/compute/futurereservation_controller.go
gemmahou: Updated

pkg/controller/direct/compute/futurereservation_controller.go
gemmahou: Updated

tests/e2e/normalize.go
gemmahou: Moved to mockcompute/normalize.go

pkg/test/resourcefixture/testdata/basic/compute/v1alpha1/computefuturereservation/basicreservation-duration/_generated_object_basicreservation-duration.golden.yaml
gemmahou: "existingMatchingUsageInfo" is output-only fields under "observedState", and it takes a random time(several seconds) for this field to populate after successful creation of the resource, that's why we sometimes see the field is missing in http and object real logs. In mockGCP log we always have a "fake" output. I think this field won't affect reconciliation, so the diff is probably acceptable.

pkg/test/resourcefixture/testdata/basic/compute/v1alpha1/computefuturereservation/basicreservation-duration/update.yaml
gemmahou: "0" is not supported by the API: googleapi: Error 400: Invalid value for field 'resource.specificSkuProperties.totalCount': '0'. Future reservation must specify a nonzero total count.

pkg/test/resourcefixture/testdata/basic/compute/v1alpha1/computefuturereservation/basicreservation-endtime/create.yaml
gemmahou: One of "duration" or "endTime" must be set, so I don't think we can "unset" this field: googleapi: Error 400: Invalid value for field 'resource.timeWindow.endTime': ''. Must be a string representation of a full-date or date-time in valid RFC3339 format with either 0 or 3 digits for fractional seconds

mockgcp/mockcompute/futurereservationsv1.go
gemmahou: We have this implementation in many mock services: https://github.com/search?q=repo%3AGoogleCloudPlatform%2Fk8s-config-connector+%22update_mask+path+%25q+not+valid%22&type=code, but that could be legacy.

I updated to use "fields.UpdateByFieldMask" to accept and update field values that direct controller accepts, see commit https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/7723/changes/a301df57228854b50bf6682d3402d97944e1e3f7

pkg/test/resourcefixture/testdata/basic/compute/v1alpha1/computefuturereservation/basicreservation-endtime/create.yaml
maqiuyujoyce: Can we have a test case that updates from using `endTime` to using `duration`?

gemmahou: Added two tests in the latest two commit, updated endtime to duration, and update duration to endtime. You can check the GET log after PATCH, the endtime(it always shows endtime in GET response) has been updated to a day plus one second.

ldanielmadariaga (dismissed):

ldanielmadariaga (approved):

PR #7724, opened 2026-04-29
chore: bump googleapis protos

acpana:
Using the `dev/tasks/update-gcp-dependencies` to bump the googleapis protos. I chose to "defer" fixing up the issues that arose from our generators not being able to generate types and mappers correctly for the following resources:

```
       * networkmanagement
       * gkehub
       * sql
       * firestore (v1alpha1 and v1beta1)
```  

I or an agent will follow up on it!


apis/firestore/v1alpha1/generate.sh
```diff
@@ -23,18 +23,20 @@ cd ${REPO_ROOT}/dev/tools/controllerbuilder
 
 ./generate-proto.sh
 
-go run . generate-types \
-  --service google.firestore.admin.v1 \
-  --api-version firestore.cnrm.cloud.google.com/v1alpha1  \
-  --resource FirestoreDocument:google.firestore.v1.Document \
-  --resource FirestoreField:Field \
-  --resource FirestoreBackupSchedule:BackupSchedule
+# Temporarily disabled: controllerbuilder is not yet generating all transitive dependencies,
```
justinsb, line 26: This feels like a regression, so I don't think we can merge as-is.

I suggest either fixing controllerbuilder, or hand-coding some of the missing types (or the type that needs a missing type)

apis/gkehub/v1beta1/generate.sh
```diff
@@ -24,16 +24,17 @@ cd ${REPO_ROOT}/dev/tools/controllerbuilder
 
 ./generate-proto.sh
 
-go run . generate-types \
-  --service google.cloud.gkehub.v1beta --api-version gkehub.cnrm.cloud.google.com/v1beta1 \
-  --resource GKEHubFeatureMembership:MembershipFeatureSpec
+# Temporarily disabled: requires manual intervention to complete the API definition.
```
justinsb, line 27: Same here.  You are very allowed to override types.  But let's not turn off the generation.

acpana:
abandoning in favor of #7745 which only bumps the gkehub protos!

PR #7728, opened 2026-04-30
feat: Add ShareSettings.ProjectMap in ComputeFutureReservation CRD

gemmahou:
### BRIEF Change description

Addresses #6828

Introduce "shareSettings.projectMap" field and resource reference in ComputeFutureReservation

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_computefuturereservations.compute.cnrm.cloud.google.com.yaml
```diff
@@ -233,6 +233,87 @@ spec:
               shareSettings:
                 description: List of Projects/Folders to share with.
                 properties:
+                  projectMap:
+                    description: A map of project id and project config. This is only
```
ldanielmadariaga: I thought we had landed that the key would not be a project ID . I think the external ref here is our scape hatch but we probably don't want make folks think project id is the only valid key.

gemmahou: Updated the comment so it is no longer limited to 'project id,' which was originally copy-pasted from the GCP proto. The implementation of the 'keyRef' field follows our discussion - either use "external" or use both "name" and "kind" to support multi-kind references.

maqiuyujoyce (commented):
Could we cover this newly added field in the test case?

gemmahou:
> Could we cover this newly added field in the test case?

I updated the fuzzer test with this field. I will cover them in fixtures test in another PR when we support ShareSettings feature in direct controller. Is there a specific test you are looking for?

maqiuyujoyce:
> > Could we cover this newly added field in the test case?
> 
> I updated the fuzzer test with this field. I will cover them in fixtures test in another PR when we support ShareSettings feature in direct controller. Is there a specific test you are looking for?

Oh I meant the fixture test and yes, let's address that separately.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_computefuturereservations.compute.cnrm.cloud.google.com.yaml
```diff
@@ -233,6 +233,88 @@ spec:
...
+                    items:
+                      properties:
+                        keyRef:
+                          oneOf:
+                          - not:
+                              required:
+                              - external
+                            required:
+                            - name
+                            - kind
```
maqiuyujoyce, line 251: I'm leaning against using Project as the default kind (instead of requiring this field) because Project is the only supported kind right now and will probably be the only supported kind for some time (depending on the API team's timeline of supporting more resource as keys, which is out of our control).

I'm okay with how it is right now and decide later at the Beta promotion time.

maqiuyujoyce (commented):
/lgtm

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_computefuturereservations.compute.cnrm.cloud.google.com.yaml
gemmahou: Updated in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/7728/changes/d8eda47acd7e15bb9068ea0b5c9a42fe05e58760

ldanielmadariaga (approved):

PR #7743, opened 2026-04-30
preview: fix crash on typed resources and hang on defaulting

### LINKED ISSUE #7742: preview: tool crashes on typed resources and hangs on defaulting
### Description
The Config Connector `preview` tool encountered two main issues:

1. **Crash on Typed Resources**: The `recorder.go` crashed with `unhandled arg type *v1beta1.IAMPartialPolicy` when intercepting typed Kubernetes client calls, as it previously only handled `*unstructured.Unstructured`.
2. **Hang on Defaulting**: The tool infinite-looped because it blocked `Update` and `Patch` operations. Many controllers (e.g., `IAMPartialPolicy`) apply defaults in-memory and immediately attempt an `Update`, which failed when blocked and triggered immediate reconciler requeues.

### Resolution
- Updated `recorder.go` to support `client.Object` and dynamically resolve Group and Kind via reflection when missing.
- Modified `intercepting_kube_client.go` to ignore `Update` and `Patch` operations (similar to `status.update`), allowing controllers to safely proceed with in-memory mutations without deadlocking the reconciler.

anhdle-sso:
Fixes #7742

### DIFF
--- a/pkg/cli/preview/recorder.go
+++ b/pkg/cli/preview/recorder.go
@@ -21,9 +21,9 @@ import (
 	"strings"
 	"sync"
 
+	kccscheme "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/clients/generated/client/clientset/versioned/scheme"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/k8s"
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/structuredreporting"
-	kccscheme "github.com/GoogleCloudPlatform/k8s-config-connector/pkg/clients/generated/client/clientset/versioned/scheme"
 
 	constants "github.com/GoogleCloudPlatform/k8s-config-connector/operator/pkg/k8s"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
--- a/pkg/cli/preview/recorder.go
+++ b/pkg/cli/preview/recorder.go
@@ -17,6 +17,7 @@ package preview
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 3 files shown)


pkg/cli/preview/recorder.go
```diff
@@ -267,6 +268,32 @@ func (r *Recorder) recordKubeAction(ctx context.Context, method string, args []a
 			}
 			// We could capture the object here: kubeAction.object = arg.DeepCopy()
 
+		case client.Object:
+			gvk := arg.GetObjectKind().GroupVersionKind()
```
justinsb: Ugh ... good point.  I would just use a runtime.Scheme.  You probably want https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client/apiutil#GVKForObject

anhdle-sso: updated the PR. Please take a look

pkg/cli/preview/recorder.go
```diff
@@ -267,6 +270,38 @@ func (r *Recorder) recordKubeAction(ctx context.Context, method string, args []a
 			}
 			// We could capture the object here: kubeAction.object = arg.DeepCopy()
 
+		case client.Object:
+			gvk, err := apiutil.GVKForObject(arg, kccscheme.Scheme)
+			if err != nil {
+				klog.V(2).Infof("apiutil.GVKForObject failed: %v", err)
+			}
+			if gvk.Kind == "" {
+				gvk = arg.GetObjectKind().GroupVersionKind()
```
justinsb: Do we still need this branch - will it ever "work"?

pkg/cli/preview/recorder.go
```diff
@@ -267,6 +270,38 @@ func (r *Recorder) recordKubeAction(ctx context.Context, method string, args []a
...
+			}
+			if gvk.Kind == "" {
+				t := reflect.TypeOf(arg)
```
justinsb: Same here - does this ever get us a Kind?  Which types are not in the scheme?

anhdle-sso:
I have addressed the review comments by:
1. Updating the recorder to accept and use a *runtime.Scheme* for GVK resolution.
2. Using apiutil.GVKForObject instead of the custom reflection-based logic.
3. Removing the redundant GVK resolution branches that were questioned.
4. Passing the scheme from the client to the recorder in intercepting_kube_client.go.

ldanielmadariaga (approved):

PR #7971, opened 2026-05-09
feat: Support SharedSettings in ComputeFutureReservation

gemmahou:
### BRIEF Change description

Addresses https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/6828, support ShareSettings in ComputeFutureReservation

In the GCP API, the ShareSettings.ProjectMap field can be specified as a projectID or projectNumber during creation or update, but the API only returns the projectNumber when the resource is fetched. KCC needs to convert a projectID to a projectNumber prior to comparison by reading the resource from GCP.

```release-note
Support SharedSettings in ComputeFutureReservation
```

```docs

```

#### Intended Milestone

### Tests you have done

### DIFF
--- a/apis/refs/v1beta1/projectref.go
+++ b/apis/refs/v1beta1/projectref.go
@@ -98,6 +98,14 @@ type ExtendedProjectRef struct {
 	External string `json:"external,omitempty"`
 }
 
+func (r *ExtendedProjectRef) GetGVK() schema.GroupVersionKind {
+	// If Kind is not specified, default to "Project"
+	if r.Kind == "" {
+		return ProjectGVK
+	}
+	return schema.FromAPIVersionAndKind(r.APIVersion, r.Kind)
+}
+
 func (r *ExtendedProjectRef) GetNamespacedName() types.NamespacedName {
 	return types.NamespacedName{
 		Name:      r.Name,
@@ -106,11 +114,48 @@ func (r *ExtendedProjectRef) GetNamespacedName() types.NamespacedName {
 }
 
 func (r *ExtendedProjectRef) GroupVersionKind() schema.GroupVersionKind {
-	return schema.FromAPIVersionAndKind(r.APIVersion, r.Kind)
+	return r.GetGVK()
 }
 
 func (r *ExtendedProjectRef) SetGroupVersionKind(gvk schema.GroupVersionKind) {
 	r.APIVersion, r.Kind = gvk.ToAPIVersionAndKind()
+}
+
+func (r *ExtendedProjectRef) GetExternal() string {
+	return r.External
+}
+
+func (r *ExtendedProjectRef) SetExternal(ref string) {
+	r.External = ref
+}
+
+func (r *ExtendedProjectRef) ValidateExternal(ref string) error {
+	// If Kind is not specified, default to "Project"
+	if r.Kind == "" || r.Kind == "Project" {
+		id := &ProjectIdentity{}
+		if err := id.FromExternal(ref); err != nil {
+			return err
+		}
+		return nil
+	}
+	return nil
+}
+
+func (r *ExtendedProjectRef) Normalize(ctx context.Context, reader client.Reader, defaultNamespace string) error {
+	// If Kind is not specified, default to "Project"
+	if r.Kind == "" || r.Kind == "Project" {
+		projectRef := &ProjectRef{
+			External:  r.External,
+			Name:      r.Name,
+			Namespace: r.Namespace,
+		}
+		if err := projectRef.Normalize(ctx, reader, defaultNamespace); err != nil {
+			return err
+		}
+		r.External = projectRef.External
+		return nil
+	}
+	return Normalize(ctx, reader, r, defaultNamespace)
 }
 
 type ProjectIdentity struct {
--- a/apis/refs/v1beta1/projectref.go
+++ b/apis/refs/v1beta1/projectref.go
@@ -89,7 +89,6 @@ func AsProjectRef(in *deprecatedrefs.ResourceRef) *ProjectRef {
 // The resource reference that defaults to Project if Kind is not specified.
 type ExtendedProjectRef struct {
 	// Kind of the referenced resource
-	// +kubebuilder:default=Project
 	Kind      string `json:"kind,omitempty"`
 	Namespace string `json:"namespace,omitempty"`
 	Name      string `json:"name,omitempty"`
--- a/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_computefuturereservations.compute.cnrm.cloud.google.com.yaml
+++ b/config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_computefuturereservations.compute.cnrm.cloud.google.com.yaml
@@ -248,15 +248,12 @@ spec:
                               - external
                             required:
                             - name
-                            - kind
                           - not:
                               anyOf:
                               - required:
                                 - name
                               - required:
                                 - namespace
-                              - required:
-                                - kind
                             required:
                             - external
                           properties:
@@ -267,7 +264,6 @@ spec:
                               description: The external name of the referenced resource
                               type: string
                             kind:
-                              default: Project
                               description: Kind of the referenced resource
                               type: string
                             name:
--- a/mockgcp/common/fields/updatemask.go
+++ b/mockgcp/common/fields/updatemask.go
@@ -97,6 +97,11 @@ func replace(original, update protoreflect.Message, fieldName string) error {
 	originalFd := findField(original, fieldName)
 	originalVal := original.Get(originalFd)
 	updateFd := findField(update, fieldName)
+	if !update.Has(updateFd) {
+		// If the field is in the mask but not in the update message, it should be cleared (AIP-134).
+		original.Clear(originalFd)
+		return nil
+	}
 	updateVal := update.Get(updateFd)
 
 	// Update Map
--- a/mockgcp/mockcompute/futurereservationsv1.go
+++ b/mockgcp/mockcompute/futurereservationsv1.go
@@ -17,6 +17,7 @@ package mockcompute
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
@@ -119,6 +120,16 @@ func (s *FutureReservationsV1) Insert(ctx context.Context, req *pb.InsertFutureR
 		if obj.ShareSettings != nil && obj.ShareSettings.GetShareType() == "LOCAL" {
 			obj.ShareSettings = nil
 		}
+		if obj.ShareSettings != nil && obj.ShareSettings.GetShareType() == "SPECIFIC_PROJECTS" {
+			if obj.GetShareSettings().GetProjectMap() == nil {
+				return nil, status.Error(codes.InvalidArgument, "Invalid value for field 'resource.shareSettings.projects': ''. Shared future reservation must be shared with at least one project.")
+			}
+			newMap, err := s.convertProjectMap(ctx, obj.GetShareSettings().GetProjectMap())
+			if err != nil {
+				return nil, err
+			}
+			obj.ShareSettings.ProjectMap = newMap
+		}
 		if obj.GetSchedulingType() == "FLEXIBLE" {
 			obj.SchedulingType = nil
 		}
@@ -219,6 +230,16 @@ func (s *FutureReservationsV1) Update(ctx context.Context, req *pb.UpdateFutureR
 		// Simulate realGCP
 		if obj.Description == nil {
 			obj.Description = direct.PtrTo("")
+		}
+		if obj.ShareSettings != nil && obj.ShareSettings.GetShareType() == "LOCAL" {
+			obj.ShareSettings = nil
+		}
+		if obj.ShareSettings != nil && obj.ShareSettings.GetShareType() == "SPECIFIC_PROJECTS" {
+			newMap, err := s.convertProjectMap(ctx, obj.GetShareSettings().GetProjectMap())
+			if err != nil {
+				return nil, err
+			}
+			obj.ShareSettings.ProjectMap = newMap
 		}
 
 		if err := s.storage.Update(ctx, fqn, obj); err != nil {
@@ -354,6 +375,24 @@ func (s *MockService) parseFutureReservationName(name string) (*futureReservatio
 	} else {
 		return nil, status.Errorf(codes.InvalidArgument, "name %q is not valid", name)
 	}
+}
+
+func (s *MockService) convertProjectMap(ctx context.Context, projectMap map[string]*pb.ShareSettingsProjectConfig) (map[string]*pb.ShareSettingsProjectConfig, error) {
+	if projectMap == nil {
+		return nil, nil
+	}
+	newMap := make(map[string]*pb.ShareSettingsProjectConfig)
+	for idOrNumber, config := range projectMap {
+		project, err := s.Projects.GetProjectByIDOrNumber(idOrNumber)
+		if err != nil {
+			return nil, err
+		}
+		projectNumber := strconv.FormatInt(project.Number, 10)
+		newConfig := proto.Clone(config).(*pb.ShareSettingsProjectConfig)
+		newConfig.ProjectId = &projectNumber
+		newMap[projectNumber] = newConfig
+	}
+	return newMap, nil
 }
 
 func parseDuration(startTime *string, duration *pb.Duration) (time.Time, error) {
--- a/pkg/controller/direct/compute/futurereservation_controller.go
+++ b/pkg/controller/direct/compute/futurereservation_controller.go
@@ -25,6 +25,10 @@ import (
 	"sort"
 	"strings"
 
+	"google.golang.org/protobuf/proto"
+
+	"github.com/GoogleCloudPlatform/k8s-config-connector/apis/common/projects"
+
 	refsv1beta1 "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -87,10 +91,11 @@ func (m *futureReservationModel) AdapterForObject(ctx context.Context, op *direc
 	}
 
 	return &FutureReservationAdapter{
-		gcpClient: futureReservationClient,
-		id:        id.(*krm.ComputeFutureReservationIdentity),
-		desired:   obj,
-		reader:    reader,
+		gcpClient:     futureReservationClient,
+		projectMapper: m.config.ProjectMapper,
+		id:            id.(*krm.ComputeFutureReservationIdentity),
+		desired:       obj,
+		reader:        reader,
 	}, nil
 }
 
@@ -100,11 +105,12 @@ func (m *futureReservationModel) AdapterForURL(ctx context.Context, url string) 
 }
 
 type FutureReservationAdapter struct {
-	gcpClient *compute.FutureReservationsClient
-	id        *v1alpha1.ComputeFutureReservationIdentity
-	desired   *krm.ComputeFutureReservation
-	actual    *computepb.FutureReservation
-	reader    client.Reader
+	gcpClient     *compute.FutureReservationsClient
+	projectMapper *projects.ProjectMapper
+	id            *v1alpha1.ComputeFutureReservationIdentity
+	desired       *krm.ComputeFutureReservation
+	actual        *computepb.FutureReservation
+	reader        client.Reader
 }
 
 var _ directbase.Adapter = &FutureReservationAdapter{}
@@ -141,6 +147,9 @@ func (a *FutureReservationAdapter) Create(ctx context.Context, createOp *directb
 	mapCtx := &direct.MapContext{}
 
 	desired := a.desired.DeepCopy()
+	if err := ResolveComputeFutureReservationRefs(ctx, a.reader, a.projectMapper, desired); err != nil {
+		return err
+	}
 	resource := ComputeFutureReservationSpec_v1alpha1_ToProto(mapCtx, &desired.Spec)
 	if mapCtx.Err() != nil {
 		return mapCtx.Err()
@@ -185,6 +194,9 @@ func (a *FutureReservationAdapter) Update(ctx context.Context, updateOp *directb
 	mapCtx := &direct.MapContext{}
 
 	desired := a.desired.DeepCopy()
+	if err := ResolveComputeFutureReservationRefs(ctx, a.reader, a.projectMapper, desired); err != nil {
+		return err
+	}
 	resource := ComputeFutureReservationSpec_v1alpha1_ToProto(mapCtx, &desired.Spec)
 	if mapCtx.Err() != nil {
 		return mapCtx.Err()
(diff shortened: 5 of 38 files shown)


apis/refs/v1beta1/projectref.go
```diff
@@ -114,6 +115,21 @@ func (r *ExtendedProjectRef) SetGroupVersionKind(gvk schema.GroupVersionKind) {
 	r.APIVersion, r.Kind = gvk.ToAPIVersionAndKind()
 }
 
+func (r *ExtendedProjectRef) ConvertToProjectNumber(ctx context.Context, projectMapper *projects.ProjectMapper) error {
```
gemmahou: This method appears to be entirely unused in this PR. Consider either utilizing it in `ResolveComputeFutureReservationRefs` (by passing the `ProjectMapper` to it) to normalize the references early—which is a common pattern in direct controllers to prevent drift—or removing this dead code.

apis/refs/v1beta1/projectref.go
```diff
@@ -114,6 +115,21 @@ func (r *ExtendedProjectRef) SetGroupVersionKind(gvk schema.GroupVersionKind) {
...
+	if r.External == "" {
+		return nil
+	}
+	// If it's a project, convert ID to number
+	if r.Kind == "Project" || r.Kind == "" {
```
gemmahou: Handling the empty string here ensures backward compatibility since the `+kubebuilder:default=Project` marker was removed from `ExtendedProjectRef`. Good defensive programming!

apis/refs/v1beta1/projectref.go
```diff
@@ -303,3 +319,26 @@ func (r *ProjectRef) ParseExternalToIdentity() (identity.Identity, error) {
 	}
 	return id, nil
 }
+
+func (r *ProjectRef) ConvertToProjectNumber(ctx context.Context, projectMapper *projects.ProjectMapper) error {
```
gemmahou: Similar to `ExtendedProjectRef.ConvertToProjectNumber`, this method is added but never used. Consider using it for early reference normalization or removing it to avoid introducing unused code.

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -140,6 +146,10 @@ func (a *FutureReservationAdapter) Create(ctx context.Context, createOp *directb
 	log.V(2).Info("creating FutureReservation", "name", a.id)
 	mapCtx := &direct.MapContext{}
 
+	if err := ResolveComputeFutureReservationRefs(ctx, a.reader, a.desired); err != nil {
```
gemmahou: Modifying `a.desired` directly here alters the state of the adapter. Since you immediately deep-copy the object on line 153 anyway, it is cleaner and safer to deep-copy first (`desired := a.desired.DeepCopy()`), and then resolve references on the local `desired` copy. This ensures `a.desired` strictly reflects the original input.

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -184,6 +194,10 @@ func (a *FutureReservationAdapter) Update(ctx context.Context, updateOp *directb
 	log.V(2).Info("updating FutureReservation", "name", a.id)
 	mapCtx := &direct.MapContext{}
 
+	if err := ResolveComputeFutureReservationRefs(ctx, a.reader, a.desired); err != nil {
```
gemmahou: Similar to the `Create` function, mutating `a.desired` directly here could be misleading. Consider performing the deep-copy first and resolving references on the local copy instead.

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -448,3 +467,57 @@ func compareTimestamp(actual, desired string) (bool, error) {
...
+	}
+	if desired == nil || actual == nil {
+		return false
+	}
+
+	newDesired := convertShareSettingsProjectMap(ctx, desired, projectMapper)
+	return proto.Equal(newDesired, actual)
+}
+
+func convertShareSettingsProjectMap(ctx context.Context, in *computepb.ShareSettings, projectMapper *projects.ProjectMapper) *computepb.ShareSettings {
```
gemmahou: Since this function iterates over map keys and makes external cache/API lookups, doing this during diffing is slightly inefficient. If the inputs were already normalized to project numbers by the mapper early on, this function could be entirely removed and replaced with a standard `proto.Equal`.

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -448,3 +467,57 @@ func compareTimestamp(actual, desired string) (bool, error) {
...
+	out := &computepb.ShareSettings{
+		ShareType:  in.ShareType,
+		ProjectMap: make(map[string]*computepb.ShareSettingsProjectConfig),
+	}
+
+	keys := make([]string, 0, len(in.ProjectMap))
+	for k := range in.ProjectMap {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
gemmahou: Sorting the keys ensures deterministic iteration which is good practice. However, since the output map `out.ProjectMap` is fundamentally unordered in Go, the order of insertion doesn't guarantee a deterministic output for `proto.Equal`. Fortunately, `proto.Equal` handles map comparisons gracefully regardless of insertion order.

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -448,3 +467,57 @@ func compareTimestamp(actual, desired string) (bool, error) {
...
+
+	for _, k := range keys {
+		v := in.ProjectMap[k]
+		projectNumber, err := projectMapper.LookupProjectNumber(ctx, k)
+		key := k
+		if err == nil {
```
gemmahou: Errors from `LookupProjectNumber` are silently swallowed here, and the unmapped `k` value is used as a fallback. While this might be acceptable for a diffing function (`isShareSettingsEqual`), it masks potential resolution failures. Normalizing references to project numbers early during the resolve phase would allow you to avoid this conversion and return errors gracefully.

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -448,3 +467,57 @@ func compareTimestamp(actual, desired string) (bool, error) {
...
+			key = fmt.Sprintf("%d", projectNumber)
+		}
+
+		config := &computepb.ShareSettingsProjectConfig{}
+		if v.ProjectId != nil {
+			projectNumber, err := projectMapper.LookupProjectNumber(ctx, *v.ProjectId)
+			if err == nil {
```
gemmahou: Same here. Swallowing the error hides potential resolution failures from the user. Normalizing the reference early in the reconcile loop would allow proper error propagation.

pkg/controller/direct/compute/futurereservation_resolverefs.go
```diff
@@ -0,0 +1,78 @@
...
+	"context"
+	"fmt"
+	"sort"
+
+	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/compute/v1alpha1"
+	refs "github.com/GoogleCloudPlatform/k8s-config-connector/apis/refs/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func ResolveComputeFutureReservationRefs(ctx context.Context, reader client.Reader, obj *krm.ComputeFutureReservation) error {
```
gemmahou: Consider passing `projectMapper *projects.ProjectMapper` to this function and calling `entry.KeyRef.ConvertToProjectNumber` and `entry.Value.ProjectIDRef.ConvertToProjectNumber` here. This would normalize the references to project numbers early, removing the need for custom conversion logic in `isShareSettingsEqual` and preventing drift when the GCP API returns project numbers.

pkg/controller/direct/compute/futurereservation_resolverefs.go
```diff
@@ -0,0 +1,78 @@
...
+	if obj.Spec.ShareSettings == nil {
+		return nil
+	}
+	for i := range obj.Spec.ShareSettings.ProjectMap {
+		entry := &obj.Spec.ShareSettings.ProjectMap[i]
+		if entry.KeyRef != nil {
+			kind := entry.KeyRef.Kind
+			switch kind {
+			// If Kind is not specified, default to "Project"
+			case "Project", "":
```
gemmahou: Defaulting to `Project` when `Kind` is empty gracefully covers any existing KRM resources that didn't have this field explicitly set when the kubebuilder default was removed.

pkg/controller/direct/compute/futurereservation_resolverefs.go
```diff
@@ -0,0 +1,78 @@
...
+	// Sort by KeyRef.External to ensure log stability
+	sort.Slice(obj.Spec.ShareSettings.ProjectMap, func(i, j int) bool {
+		return obj.Spec.ShareSettings.ProjectMap[i].KeyRef.External < obj.Spec.ShareSettings.ProjectMap[j].KeyRef.External
+	})
+
+	// Deduplicate
+	if len(obj.Spec.ShareSettings.ProjectMap) > 1 {
+		unique := make([]krm.ShareSettingsProjectMap, 0, len(obj.Spec.ShareSettings.ProjectMap))
+		for _, entry := range obj.Spec.ShareSettings.ProjectMap {
+			if len(unique) == 0 || unique[len(unique)-1].KeyRef.External != entry.KeyRef.External {
```
gemmahou: If a user mistakenly provides duplicate `KeyRef` entries with conflicting values, this logic will silently discard the subsequent ones. Consider returning an error to inform the user of the misconfiguration rather than silently dropping it.

mockgcp/mockcompute/futurereservationsv1.go
```diff
@@ -119,6 +121,13 @@ func (s *FutureReservationsV1) Insert(ctx context.Context, req *pb.InsertFutureR
 		if obj.ShareSettings != nil && obj.ShareSettings.GetShareType() == "LOCAL" {
 			obj.ShareSettings = nil
```
gemmahou, line 121: This successfully mimics the GCP API behavior of clearing `ShareSettings` when the `ShareType` is set to `LOCAL`.

mockgcp/mockcompute/futurereservationsv1.go
```diff
@@ -119,6 +121,13 @@ func (s *FutureReservationsV1) Insert(ctx context.Context, req *pb.InsertFutureR
...
 		}
+		if obj.ShareSettings != nil && obj.ShareSettings.GetShareType() == "SPECIFIC_PROJECTS" {
```
gemmahou, line 123: Is it possible for the `ShareType` to be `SPECIFIC_PROJECTS` but `ProjectMap` to be nil or empty? If the GCP API enforces validation that `ProjectMap` must contain at least one entry when `SPECIFIC_PROJECTS` is used, we might want to add that validation here or in the API checking webhook.

mockgcp/mockcompute/futurereservationsv1.go
```diff
@@ -356,6 +375,24 @@ func (s *MockService) parseFutureReservationName(name string) (*futureReservatio
 	}
 }
 
+func (s *MockService) convertProjectMap(ctx context.Context, projectMap map[string]*pb.ShareSettingsProjectConfig) (map[string]*pb.ShareSettingsProjectConfig, error) {
+	if projectMap == nil {
+		return nil, nil
+	}
+	newMap := make(map[string]*pb.ShareSettingsProjectConfig)
+	for idOrNumber, config := range projectMap {
```
gemmahou, line 385: Iterating over `projectMap` here and fetching project numbers correctly models how the GCP API transforms input project IDs to project numbers internally.

mockgcp/mockcompute/futurereservationsv1.go
```diff
@@ -356,6 +375,24 @@ func (s *MockService) parseFutureReservationName(name string) (*futureReservatio
...
+		project, err := s.Projects.GetProjectByIDOrNumber(idOrNumber)
+		if err != nil {
+			return nil, err
+		}
+		projectNumber := strconv.FormatInt(project.Number, 10)
+		newConfig := proto.Clone(config).(*pb.ShareSettingsProjectConfig)
+		newConfig.ProjectId = &projectNumber
```
gemmahou, line 392: This unconditionally sets `ProjectId` to the project number, even if it was originally `nil` in the request. If the real GCP API auto-populates this field when omitted, this accurately models the API. If not, you might want to preserve the `nil` state.

mockgcp/common/fields/updatemask.go
```diff
@@ -97,6 +97,11 @@ func replace(original, update protoreflect.Message, fieldName string) error {
 	originalFd := findField(original, fieldName)
 	originalVal := original.Get(originalFd)
 	updateFd := findField(update, fieldName)
+	if !update.Has(updateFd) {
+		// If the field is in the mask but not in the update message, it should be cleared (AIP-134).
```
gemmahou, line 101: This is a great fix and correctly implements AIP-134 for handling fields that are in the update mask but absent from the update message.

pkg/controller/direct/compute/futurereservation_controller.go
```diff
@@ -448,3 +467,57 @@ func compareTimestamp(actual, desired string) (bool, error) {
...
+		return false
+	}
+
+	newDesired := convertShareSettingsProjectMap(ctx, desired, projectMapper)
+	return proto.Equal(newDesired, actual)
+}
+
+func convertShareSettingsProjectMap(ctx context.Context, in *computepb.ShareSettings, projectMapper *projects.ProjectMapper) *computepb.ShareSettings {
+	if in == nil {
+		return nil
```
gemmahou: Instead of manually instantiating a new `computepb.ShareSettings` and copying fields, consider using `proto.Clone(in).(*computepb.ShareSettings)` and then modifying its `ProjectMap`. This prevents future fields added to `ShareSettings` from being inadvertently dropped during comparison, which could cause false diffs and continuous reconciliation loops if the GCP API evolves.

pkg/controller/direct/compute/futurereservation_controller.go
gemmahou: Similarly, instantiating a new `computepb.ShareSettingsProjectConfig` here will drop any additional fields that might be added to this message in the future. Using `proto.Clone(v).(*computepb.ShareSettingsProjectConfig)` is a safer approach to ensure the configuration remains intact.

pkg/controller/direct/compute/futurereservation_resolverefs.go
```diff
@@ -0,0 +1,78 @@
...
+			project, err := refs.ResolveProject(ctx, reader, obj.Namespace, entry.Value.ProjectIDRef)
+			if err != nil {
+				return err
+			}
+			entry.Value.ProjectIDRef.External = project.ProjectID
+		}
+	}
+
+	// Sort by KeyRef.External to ensure log stability
+	sort.Slice(obj.Spec.ShareSettings.ProjectMap, func(i, j int) bool {
```
gemmahou, line 54: If `obj.Spec.ShareSettings.ProjectMap[i].KeyRef` or `obj.Spec.ShareSettings.ProjectMap[j].KeyRef` is `nil`, accessing `.External` will cause a panic (nil pointer dereference). While KRM validation (`required: - keyRef`) usually prevents this on admitted resources, the controller should defensively handle `nil` references.

pkg/controller/direct/compute/futurereservation_resolverefs.go
```diff
@@ -0,0 +1,78 @@
...
+		return obj.Spec.ShareSettings.ProjectMap[i].KeyRef.External < obj.Spec.ShareSettings.ProjectMap[j].KeyRef.External
+	})
+
+	// Deduplicate
+	if len(obj.Spec.ShareSettings.ProjectMap) > 1 {
```
gemmahou: Similar to the sorting logic, if `unique[len(unique)-1].KeyRef` or `entry.KeyRef` is `nil`, this comparison will panic. Please add `nil` safety checks before accessing `.External`.

pkg/controller/direct/compute/futurereservation_controller.go
gemmahou: Removed this whole function.

pkg/controller/direct/compute/futurereservation_resolverefs.go
gemmahou: Updated. Moved the projectID and Number conversion to this function.

pkg/controller/direct/compute/futurereservation_resolverefs.go
gemmahou: "Removed the deduplication logic to allow duplicate inputs. Any duplicates will be caught by the GCP API, which will reject them with an InvalidArgument error: Unknown name \"xxx\" at 'resource.share_settings.project_map[0]': Repeated map key: 'xxx' is already set.

mockgcp/mockcompute/futurereservationsv1.go
gemmahou: Updated with realGCP error message

mockgcp/mockcompute/futurereservationsv1.go
gemmahou: Tested against realGCP, if ProjectId is nil, GCP API will assign the key's value(projectNumber) to it. So this simulates the realGCP correctly.

apis/refs/v1beta1/projectref.go
```diff
@@ -89,7 +89,6 @@ func AsProjectRef(in *deprecatedrefs.ResourceRef) *ProjectRef {
 // The resource reference that defaults to Project if Kind is not specified.
 type ExtendedProjectRef struct {
 	// Kind of the referenced resource
-	// +kubebuilder:default=Project
```
ldanielmadariaga, line 92: For my learning, what would be the implication of removing the kubebuilder tag in this case?

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_computefuturereservations.compute.cnrm.cloud.google.com.yaml
```diff
@@ -267,7 +264,6 @@ spec:
                               description: The external name of the referenced resource
                               type: string
                             kind:
-                              default: Project
```
ldanielmadariaga, line 270: We're still defaulting to project in the controller, should we keep the behavior here?

pkg/controller/direct/compute/futurereservation_resolverefs.go
```diff
@@ -0,0 +1,92 @@
...
+		return nil
+	}
+	for i := range obj.Spec.ShareSettings.ProjectMap {
+		entry := &obj.Spec.ShareSettings.ProjectMap[i]
+		if entry.KeyRef != nil {
+			kind := entry.KeyRef.Kind
+			switch kind {
+			// If Kind is not specified, default to "Project"
+			case "Project", "":
+				projectRef := &refs.ProjectRef{
```
ldanielmadariaga: Please extract method here

pkg/controller/direct/compute/futurereservation_resolverefs.go
```diff
@@ -0,0 +1,92 @@
...
+					return err
+				}
+				projectID := project.ProjectID
+				projectNumber, err := projectMapper.LookupProjectNumber(ctx, projectID)
+				if err != nil {
+					return err
+				}
+				entry.KeyRef.External = fmt.Sprintf("%d", projectNumber)
+
+			default:
```
ldanielmadariaga: We are making kind optional in the CRD I'm not sure this'll be evidently clear to users. Another valid avenue here would be ignoring any values that are not project. 

Do we usually enforce these kinds of failures in CRDs seems a bit overzealous for a behavior that will not be evidently clear.

apis/refs/v1beta1/projectref.go
gemmahou: This default will automatically fill in the "Kind" field. From a user's perspective, they didn't set it, but it appears when they describe the KRM object. So I handled the "default to Project" logic inside the Ref's Normalize function instead: https://github.com/gemmahou/k8s-config-connector/blob/e82235f52b65b54eb73506165500676a2c655412/apis/refs/v1beta1/projectref.go#L146

scripts/add-validation-to-crds/parse-crds.go
```diff
@@ -316,6 +316,9 @@ oneOf:
 			// hack for IAMPolicy.spec.resourceRef for backwards compat
 			if fieldPath == ".spec.resourceRef" {
 				ruleYAML = resourceRefRuleWithOnlyKind
+				// hack for Compute spec.shareSettings.projectMap[].keyRef, support multi-kind reference and default kind is "Project"
+			} else if strings.HasSuffix(fieldPath, ".projectMap[].keyRef") {
+				ruleYAML = refRuleWithOptionalKind
```
gemmahou, line 321: I encounter an issue with fixtures test so I have to modify the CRD, here's the Gemini summary:


1. The CRD's Validation Logic:
Branch A (Internal): Requires name and kind. Specifically forbids external.
Branch B (External): Requires external. Specifically forbids name, namespace, and kind.
2. The Conflict with the Default Value:
When you provided external: ${projectId} in your YAML, you (correctly) omitted the kind field.
However, because the Go type had +kubebuilder:default=Project, the Kubernetes API server automatically "filled in" the missing field, effectively changing your input to:
```
 keyRef:
   external: my-project
   kind: Project # Added automatically by the default
```
3. The Failure:
Now, when the validator checks Branch B (External), it sees that kind is present. Since Branch B strictly forbids kind, the validation fails.
It also fails Branch A (Internal) because name is still missing.

4. The fix I applied:
Updated CRD generation code to make "Kind" truly optional for both internal and external reference. When an external reference is used, it now satisfies the CRD validation.

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_computefuturereservations.compute.cnrm.cloud.google.com.yaml
gemmahou: This is removed because I removed the "// +kubebuilder:default=Project" tag in `ExtendedProjectRef`. Replied to the previous comment.

pkg/controller/direct/compute/futurereservation_resolverefs.go
gemmahou: Done: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/7971/changes/b5d29c425fbff9e2961cb46cebb3d1fa15c46962#diff-2507b14031e410ef7fa34374f59f7f9c12a2d4a345519ac04be372c11bc288a6R73-R84

pkg/controller/direct/compute/futurereservation_resolverefs.go
gemmahou: I updated the reference resolution logic in https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/7971/changes/b5d29c425fbff9e2961cb46cebb3d1fa15c46962.

ExtendedProjectRef now implements the centralized "[Ref](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/5d343c9bda81a6edc27e6b49607dc1f1f6203c0d/apis/refs/v1beta1/interface.go#L40)" interface. The `ExtendedProjectRef.Normalize` method will call the generic `refs.Normalize` function. So it will be able to resolve both Project Kind and all other existing resource kinds. For example, if providing `kind: AlloyDBCluster`, the `Normalize` function will successfully resolve the referenced resource to its `externalRef`(e.g., projects/.../clusters/...). Then we will let the GCP API determines if that value is valid, so as suggested, we can remove the check here. WDYT?

(Notes from Gemini: For FutureReservation.share_settings, GCP typically only accepts Project, Folder or ContainerCluster. Referencing to AlloyDBCluster will likely result in a Bad Request/Invalid Argument from the GCP API, even though the reference was resolved correctly by KCC.)

ldanielmadariaga: That sounds great thanks!

ldanielmadariaga (dismissed):

ldanielmadariaga:
Looks good from my side, I think some feedback remains to be addressed. Please let me know if you need me to approve again once revisions are finished.

acpana (approved):
/approve

PR #8025, opened 2026-05-11
Fix case sensitivity in SQLInstance availabilityType

himanigulati01:
This PR fixes a regression in SQLInstance where case sensitivity in .settings.availabilityType caused spurious drift and restarts. It uses strings.EqualFold for comparison and adds a unit test.

Internal bug ID: b/510919254


anhdle-sso (approved):
Looks good. Thank you

PR #8117, opened 2026-05-13
fix(preview): filter redundant logs and ensure sorted reporting

### LINKED ISSUE #8116: preview: filter redundant logs from alternative controller runs and ensure sorted reporting
### Description

When running the KCC `preview` CLI tool with alternative controller comparisons, the reporting mechanism logs reconciled results as they are processed. This results in two issues:
1. Unstable/unsorted logging order across multiple runs.
2. Redundant log entries when the alternative controller run produces the exact same result/output as the default controller run.
3. Missing the confirmation of resource with healthy result during default run.

### Expected Behavior

- All reconciled results (healthy and unhealth) should be logged in a stable, sorted order.
- Redundant logs from alternative controller runs that duplicate the default controller's output should be filtered out to prevent log clutter.

anhdle-sso:
Fixes #8116

### DIFF
--- a/pkg/cli/preview/result.go
+++ b/pkg/cli/preview/result.go
@@ -139,9 +139,6 @@ func (r *RecorderReconciledResults) CombinedSummaryReport(summaryFile string, al
 	if altResult != nil && len(altResult.badResult) > 0 {
 		combinedBadResult = append(combinedBadResult, altResult.badResult...)
 	}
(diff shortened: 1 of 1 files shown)


acpana (approved):
/lgtm
/approve

but I think we should have some cheap tests here or at least show me how this diff changes things 😛

PR #8132, opened 2026-05-14
NetworkServicesWasmPlugin: direct controller and fixtures

katrielt:
### BRIEF Change description
 
This PR implements the direct controller for `NetworkServicesWasmPlugin` and adds end-to-end fixture tests, completing the third and final PR in the new direct resource onboarding workflow.

Fixes #7391
  
To enable Config Connector users to declaratively manage Google Cloud Network Services WasmPlugins (`NetworkServicesWasmPlugin`) using Kubernetes manifests via the direct controller architecture.
  
#### Special notes for your reviewer
  -  **Child Resource Versioning**: `WasmPlugin` versions exist as separate child resources (`WasmPluginVersion`) in GCP. The controller correctly iterates over `ListWasmPluginVersions` during `Find(ctx)` to populate `a.actual.
Versions`, ensuring accurate diffing and `Export()` fidelity.
  -  **Labels Handling**: Configured `Labels` to be managed via `metadata.labels` and implemented system label preservation (`goog-`, `go-`) during updates.
  -  **MockGCP Alignment**: Configured `MockGCP` normalization for `wasmPluginVersions` and aligned mock digest fields to ensure golden log stability.
  
Add direct controller support for NetworkServicesWasmPlugin
  
### Tests you have done
  
[ x] Run  make ready-pr  to ensure this PR is ready for review.
[x] Perform necessary E2E testing for changed resources.
  
Additional E2E Test Details:
  - Fixture tests ( wasmplugin-basic ) executed successfully against real GCP ( E2E_GCP_TARGET=real ).
  - Fixture tests executed successfully against MockGCP ( hack/compare-mock fixtures/wasmplugin-basic ) with zero re-reconciliation diffs.
  - Verified  Export()  correctly populates  ProjectAndLocationRef  (both  projectRef  and  location ).
  - Audited golden HTTP log discrepancies between  real_http.log  and  _http.log  to ensure complete MockGCP fidelity.


pkg/controller/direct/networkservices/wasmplugin_controller.go
```diff
@@ -0,0 +1,310 @@
...
+
+func (a *WasmPluginAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOperation) error {
+	log := klog.FromContext(ctx)
+	log.V(2).Info("updating WasmPlugin", "name", a.id)
+	mapCtx := &direct.MapContext{}
+
+	resource := proto.Clone(a.desiredProto).(*networkservicespb.WasmPlugin)
+	resource.Name = a.id.String()
+
+	// Preserve system labels (goog- or go-)
```
maqiuyujoyce: Just to confirm, does NetworkServices API update the labels field and add system labels?

pkg/controller/direct/networkservices/wasmplugin_controller.go
```diff
@@ -0,0 +1,310 @@
...
+	status := &krm.NetworkServicesWasmPluginStatus{}
+	status.ObservedState = NetworkServicesWasmPluginObservedState_FromProto(mapCtx, updated)
+	if mapCtx.Err() != nil {
+		return mapCtx.Err()
+	}
+	status.ExternalRef = direct.LazyPtr(a.id.String())
+	return updateOp.UpdateStatus(ctx, status, nil)
+}
+
+func (a *WasmPluginAdapter) Export(ctx context.Context) (*unstructured.Unstructured, error) {
```
maqiuyujoyce, line 249: I think you'll also want to export labels properly here?

katrielt: Fixed. Added u.SetLabels and verified with hack/compare-mock.

pkg/controller/direct/networkservices/wasmplugin_controller.go
katrielt: Doesn't appear to, but keeping this to avoid potential reconciliation loops.

maqiuyujoyce (dismissed):
/lgtm

pkg/controller/direct/networkservices/wasmplugin_controller.go
```diff
@@ -0,0 +1,313 @@
...
+	resource.Name = a.id.String()
+
+	// Preserve system labels (goog- or go-)
+	if a.actual.Labels != nil {
+		if resource.Labels == nil {
+			resource.Labels = make(map[string]string)
+		}
+		for k, v := range a.actual.Labels {
+			if strings.HasPrefix(k, "goog-") || strings.HasPrefix(k, "go-") {
+				resource.Labels[k] = v
```
maqiuyujoyce: If we want to keep this logic, we should keep it correct. However, the current logic could overwrite the desired labels.

E.g. actual.Labels is `{"go-x": "x-value", "goog-y": "y-value", "z": "z-value"}`, and resource.Labels is (potentially incorrect): `{"go-x": "x", "z": "z-value"}`, then the resource.Labels becomes `{"go-x": "x-value", "goog-y": "y-value", "z": "z-value"}` after this piece of code, which is identical to the actual labels, but it shouldn't be.

If the API team doesn't have service-generated labels, then I suggest we don't support this feature in the controller.

katrielt: Good catch. I've removed the merging logic after verifying the api indeed doesn't set any system labals.

maqiuyujoyce (dismissed):
/lgtm

maqiuyujoyce:
Looks like there are some failures during the merge.

maqiuyujoyce (approved):

PR #8161, opened 2026-05-15
Deleting the deprecated `presubmit.yaml` file as this was noisy and confusing to have 2 presubmit workflows.

ldanielmadariaga:
Deleting the deprecated `presubmit.yaml` file as this was noisy and confusing to have 2 presubmit workflows.

Migrated validation checks from `presubmit.yaml` to `ci-presubmit.yaml` as were a more complete superset.


ldanielmadariaga:
@acpana @maqiuyujoyce failure seems like a static analysis error. There's no reason why this should trigger make/generate AFAIK. We may need to disable ci-presubmit before submitting this, but I was under the impression that's done by just removing the file.

maqiuyujoyce:
@ldanielmadariaga looks like this PR is outdated. Do we want to close it?

Issue #7636, opened 2026-04-19
Support CloudDeploy Automation in v1alpha1 direct


katrielt:
Support CloudDeploy Automation in v1alpha1 direct

PR #8255, opened 2026-05-17
CloudDeployAutomation: direct controller and fixtures

katrielt:
### BRIEF Change description
  
Add direct controller and resource fixture tests for `CloudDeployAutomation`.
  
Addresses #7636 
 
To enable Config Connector management of GCP Cloud Deploy Automation resources using the modern direct controller architecture.
  
* Implements custom `CompareProtoMessage` handling in `automation_controller.go` to clone `actual` and dynamically strip output-only `Condition` structs from rules via protobuf reflection (`protoreflect`), preventing false  
  drift and infinite update loops in a 100% future-proof manner.
* Updates `automation_mapper.go` to extract short names (last segment) for target IDs in rules and selectors to comply with GCP API validation requirements.
* Updates `mockgcp/mockclouddeploy/automation.go` to preserve rule condition defaults during updates, maintaining perfect fidelity with real GCP `GET` responses.
  
Add direct controller support for `CloudDeployAutomation`.
  
  [ ] Reviewer reviewed release note.

### Tests you have done
  
  [x] Run  make ready-pr  to ensure this PR is ready for review.
  [x] Perform necessary E2E testing for changed resources.


pkg/controller/direct/clouddeploy/clouddeployautomation_controller.go
```diff
@@ -0,0 +1,388 @@
...
+			actualClone := proto.Clone(b.(*clouddeploypb.Automation)).(*clouddeploypb.Automation)
+			for _, rule := range actualClone.Rules {
+				clearRuleCondition(rule)
+			}
+			desired := a.(*clouddeploypb.Automation)
+			if len(desired.Rules) != len(actualClone.Rules) {
+				return true, nil
+			}
+			for i := range desired.Rules {
+				if !proto.Equal(desired.Rules[i], actualClone.Rules[i]) {
```
maqiuyujoyce, line 298: Looks like we assumed the desired list sorted the same as the actual list here? Should we actually sort them before comparison?

pkg/controller/direct/clouddeploy/clouddeployautomation_controller.go
```diff
@@ -0,0 +1,388 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &AutomationAdapter{
+		id:        id.(*krm.CloudDeployAutomationIdentity),
+		gcpClient: gcpClient,
+		desired:   obj,
+		reader:    reader,
+		labels:    u.GetLabels(),
```
maqiuyujoyce, line 109: From this line, I assume the GCP resource has a top-level labels field that maps to metadata.labels in the KRM resource, however, we don't seem to pass in the labels in Create(), Update(), or Export(). Could you make sure the labels field handled properly and is tested?

pkg/controller/direct/clouddeploy/automation_controller.go
```diff
@@ -0,0 +1,388 @@
...
+			// Return success if not found (assume it was already deleted).
+			log.V(2).Info("skipping delete for non-existent Automation, assuming it was already deleted", "name", a.id.String())
+			return true, nil
+		}
+		return false, fmt.Errorf("deleting Automation %s: %w", a.id.String(), err)
+	}
+
+	err = op.Wait(ctx)
+	if err != nil {
+		if direct.IsNotFound(err) {
```
maqiuyujoyce: If a previously returned operation is NotFound, I assume something wrong happens in the GCP API, so I'd keep this error and return as is.

pkg/controller/direct/clouddeploy/clouddeployautomation_controller.go
```diff
@@ -0,0 +1,388 @@
...
+
+	if err := a.resolveReferences(ctx); err != nil {
+		return err
+	}
+
+	mapCtx := &direct.MapContext{}
+
+	a.desiredPb.Name = a.id.String()
+
+	// Preserve system labels (goog- or go-)
```
maqiuyujoyce, line 261: Just to confirm, does CloudDeploy API also update the labels field and add system labels?

pkg/controller/direct/clouddeploy/clouddeployautomation_controller.go
katrielt: Done. Added a sortRules helper to sort rules by ID before comparison in Update, ensuring a deterministic order and avoiding spurious drifts.

pkg/controller/direct/clouddeploy/clouddeployautomation_controller.go
katrielt: Done. Updated Create, Update, and Export to map metadata.labels using common.ComputeGCPLabels. I also added a label to the automation-basic test fixture for verification.

pkg/controller/direct/clouddeploy/automation_controller.go
katrielt: Done. Removed the IsNotFound check during the Wait phase of the Delete operation so that unexpected API errors are properly reported.

pkg/controller/direct/clouddeploy/clouddeployautomation_controller.go
katrielt: Confirmed. I’ve kept the preservation logic for goog- and go- prefixes to align with KCC standards and prevent "drift-fighting" with server-side generated labels.

pkg/controller/direct/clouddeploy/clouddeployautomation_controller.go
```diff
@@ -0,0 +1,420 @@
...
+
+	a.desiredPb.Name = a.id.String()
+	a.desiredPb.Labels = label.NewGCPLabelsFromK8sLabels(a.labels)
+
+	// Preserve system labels (goog- or go-)
+	if a.actual.Labels != nil {
+		if a.desiredPb.Labels == nil {
+			a.desiredPb.Labels = make(map[string]string)
+		}
+		for k, v := range a.actual.Labels {
```
maqiuyujoyce, line 266: Are the system labels not allowed in the desired labels field? If so, we probably want to check system labels don't exist in the desired labels field first to avoid hiding the issue? (See more explanation here: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/8132/changes#r3462274498)

pkg/controller/direct/clouddeploy/clouddeployautomation_controller.go
```diff
@@ -0,0 +1,420 @@
...
+	uObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
+	if err != nil {
+		return nil, err
+	}
+
+	u.SetName(k8s.ValueToDNSSubdomainName(a.id.ID()))
+	u.SetGroupVersionKind(krm.CloudDeployAutomationGVK)
+
+	u.Object = uObj
+	u.SetLabels(a.actual.Labels)
```
maqiuyujoyce: I think we want to remove the system labels from the labels list, right?

pkg/controller/direct/clouddeploy/clouddeployautomation_controller.go
katrielt: Done.

To avoid the silent overwrite issue you described, I added a validation step at the beginning of Update. It now returns an error if any GCP-reserved labels (goog-, google-, go-) are found in metadata.labels. This ensures we never hide a user's incorrect label configuration while still preserving service-generated labels to prevent reconciliation loops.

pkg/controller/direct/clouddeploy/clouddeployautomation_controller.go
katrielt: Done. Updated the Export method to filter out all GCP-reserved labels and the managed-by-cnrm label. This keeps the exported YAML clean and prevents users from accidentally including system-managed metadata in their configurations.

comments
maqiuyujoyce, line 1: Maybe delete this file?

pr-8255.json
maqiuyujoyce, line 1: ditto

pkg/controller/direct/clouddeploy/clouddeployautomation_controller.go
```diff
@@ -247,6 +247,14 @@ func (a *AutomationAdapter) Update(ctx context.Context, updateOp *directbase.Upd
 	mapCtx := &direct.MapContext{}
 
 	a.desiredPb.Name = a.id.String()
+
+	// Validate that users are not trying to manage system labels (goog- or go-)
```
maqiuyujoyce: I think this validation should be done in AdapterForObject function?

comments
katrielt: Done.

pr-8255.json
katrielt: Done

pkg/controller/direct/clouddeploy/clouddeployautomation_controller.go
katrielt: Done

maqiuyujoyce (approved):
/lgtm

PR #8484, opened 2026-05-20
Fix ComposerEnvironment reconciliation loop failure with 400 error

himanigulati01:
### Description
  This PR fixes a reconciliation loop for ComposerEnvironment that causes a the error:- 400: update_mask path not valid error. 

  The Cloud Composer API assigns certain internal, immutable fields (like composerInternalIpv4CidrBlock and composerNetworkAttachment) upon resource update/reconciliation. Previously, the composer controller used a check for the NodeConfig object. If a user specified any field inside NodeConfig (like a network reference), KCC would skip copying these server-assigned fields from the actual GCP state. This led to a
  false drift detection where KCC incorrectly included these immutable fields in the updateMask, resulting in API rejection.

  ### Changes
   * Updated pkg/controller/direct/composer/environment_controller.go to implement granular merging for NodeConfig. It now explicitly preserves server-assigned fields from the actual state if
     they are missing from the manifest.
   * Updated mockgcp/mockcomposer/environment.go to simulate the behavior of the real Composer API assigning internal IPs and network attachments.
   * Added a new permanent test fixture composerenvironmentnodeconfig that mimics the user's reported configuration to verify the fix and prevent regressions.

  ### Tests
   * Local reproduced and confirmed the 400 error occurs in MockGCP without the fix.
   * Verified that with the fix, the reproducebug (now composerenvironmentnodeconfig) fixture reconciles successfully to a Ready state.

Internal bugID: b/514547354


himanigulati01:
I have updated the branch with two final fixes to ensure all CI presubmit checks pass:
1. **`tests/apichecks/testdata/exceptions/missingfields.txt`**:
   * **Reason**: Removed `networkRef`, `serviceAccountRef`, `subnetworkRef`, and `imageVersion` from the missing-field exceptions list. Since these fields are now actively covered by our new `composerenvironmentnodeconfig` E2E test fixture, they are no longer "missing" from test coverage. This resolves the `unit-tests` failure.
2. **`mockgcp/mockcomposer/testdata/environment/crud/_http.log`**:
   * **Reason**: Updated the golden HTTP log for the existing `mockcomposer` test case to reflect the new mock response containing the default values for `composerInternalIpv4CidrBlock` and `composerNetworkAttachment`. This resolves the `test-mockgcp` failure.

mockgcp/mockcomposer/environment.go
```diff
@@ -257,6 +257,12 @@ func (s *ComposerV1) populateDefaultsForEnvironmentConfig(config *pb.Environment
 	if config.NodeConfig.ServiceAccount == "" {
 		config.NodeConfig.ServiceAccount = "${projectNumber}-compute@developer.gserviceaccount.com"
 	}
+	if config.NodeConfig.ComposerInternalIpv4CidrBlock == "" {
+		config.NodeConfig.ComposerInternalIpv4CidrBlock = "172.31.251.0/24"
+	}
```
xiaoweim, line 269: Is this the GCP defaults? If so, can we make it a constant?

mockgcp/mockcomposer/environment.go
```diff
@@ -257,6 +257,12 @@ func (s *ComposerV1) populateDefaultsForEnvironmentConfig(config *pb.Environment
...
+	if config.NodeConfig.ComposerNetworkAttachment == "" {
+		config.NodeConfig.ComposerNetworkAttachment = "projects/${projectId}/regions/us-central1/networkAttachments/test-attachment"
```
xiaoweim: Same here, if this is the GCP defaults if the user leave the field empty, can we make it a constant and add comment to explain that the value is GCP default?

pkg/controller/direct/composer/environment_controller.go
```diff
@@ -351,8 +351,19 @@ func populateDefaultsForEnvironmentConfig(desired, actual *composerpb.Environmen
...
+		// Preserve immutable, server-assigned fields.
+		// These fields are not sent in update requests (see NodeConfig_ToProto),
+		// but we need to match them in the desired state so CompareProtoMessage
+		// doesn't flag them as changed (drift).
+		if desired.NodeConfig.ComposerInternalIpv4CidrBlock == "" {
+			desired.NodeConfig.ComposerInternalIpv4CidrBlock = actual.NodeConfig.ComposerInternalIpv4CidrBlock
+		}
+		if desired.NodeConfig.ComposerNetworkAttachment == "" {
+			desired.NodeConfig.ComposerNetworkAttachment = actual.NodeConfig.ComposerNetworkAttachment
+		}
```
xiaoweim, line 367: Are these two the only two fields that have GCP defaults?

xiaoweim (commented):
Could you also make sure that the presubmit tests pass?

pkg/controller/direct/composer/environment_controller.go
himanigulati01: Thanks for pointing it out. I check the API documentation and none of the fields under nodeconfig seems mutable: https://docs.cloud.google.com/composer/docs/reference/rest/v1beta1/projects.locations.environments/patch

I tested in the lab and below 6 fields are causing the error:
composerInternalIpv4CidrBlock
composerNetworkAttachment
network
subnetwork
serviceAccount
ipAllocationPolicy

As these are auto-default by GCP. 

However I noticed a few fields like machine_type etc are not causing the error. For input fields like machineType, location, or diskSizeGb, even if we are not specifying them in the yaml they are not causing an issue maybe be because composer 3 is using autopilot under the hood. Or maybe because our local mockgcp server is a simplified version of real GCP, and does not yet default machineType, diskSizeGb, or location (they are left empty as "" or 0). Since KCC's desired state and mock GCP's actual state were both empty, they matched, and KCC saw no drift.

Nevertheless I am going to cover all the fields in next commit.

mockgcp/mockcomposer/environment.go
himanigulati01: thanks for suggestion, Added

acpana (dismissed):
/lgtm
/approve

anhdle-sso (changes requested):
The coding pattern of this controller is unconventional. Please update the funciton `populateDefaultsForEnvironmentConfig` which populate the empty value from actual gcp object.

himanigulati01:
Ack on the unconventional method, refactored populateDefaultsForEnvironmentConfig and populateDefaultsForEnvironment to deeply copy nested protobuf sub-messages using proto.Clone.

However to keep the scope of this PR strictly focused on nodeConfig changes, I have only applied it here for now. I will raise a follow-up PR dedicated to cleaning up the remaining direct pointer assignments across the rest of populateDefaultsForEnvironmentConfig.

anhdle-sso (approved):
lgtm. Thanks for the change

PR #8488, opened 2026-05-20
ai:agents: greenfield tracker, iterating

acpana:
Iterating on the greenfield step
* controller
* skills improvements
* offline tracker


maqiuyujoyce (approved):
I like how you have resource status tracked here. Do you think it'll be helpful if we (eventually) have a periodical job that updates these status automatically?

/lgtm

PR #8533, opened 2026-05-21
fix(gsakeysecretgenerator): refuse to overwrite non-KCC-managed Secrets

evilgensec:
## Summary

`pkg/controller/gsakeysecretgenerator/gsakey_secret_generator.go` unconditionally `Update`s any Secret whose name collides with the IAMServiceAccountKey being reconciled. The pre-existing Secret's `Data`, `Labels`, and `OwnerReferences` are replaced with the SA key material; the Secret is re-parented to the attacker's CRD via `OwnerReferences`; subsequent deletion of the CRD garbage-collects the Secret entirely.

Any principal with rights to create `IAMServiceAccountKey` in a namespace can therefore destroy and replace any other Secret in that namespace whose name they can predict — including Secrets created by sibling tenants, cluster-wide controllers (cert-manager, external-secrets, Helm), or CI systems. The bug has been present since the controller was added in 2022.

### Verification that the Update path is reachable

`r.Update` is called on the freshly-constructed `secret` object (no `ResourceVersion`). I confirmed against `kube-apiserver` v1.29 via envtest that this Update succeeds — the API server treats an empty `metadata.resourceVersion` as an unconditional PUT and overwrites the existing object's full spec.

### Fix

- Before overwriting, verify the observed Secret carries the `managed-by-cnrm: true` label that Config Connector sets on Secrets it creates (`pkg/label/const.go`). Foreign Secrets are left untouched and a `Warning` event is recorded.
- Bind the Update to the observed `ResourceVersion` so a concurrent label-removal cannot race the check.

### Tests

Two new unit tests in `pkg/controller/gsakeysecretgenerator/gsakey_secret_generator_security_test.go`:

- `TestReconcile_RefusesToOverwriteForeignSecret` — drives the reconciler against a pre-existing Secret that lacks the KCC label, with an attacker-controlled `IAMServiceAccountKey` whose name collides. Asserts the reconcile returns an error and the victim Secret's `Data` / `Labels` / `OwnerReferences` are unchanged.
- `TestReconcile_UpdatesOwnSecret` — drives the reconciler against a Secret that was created by the controller (carries `managed-by-cnrm: true`) and asserts that rotation correctly replaces the key material.

Both tests pass on this branch. The first test would fail on `master`.

## Test plan

- [x] `go test ./pkg/controller/gsakeysecretgenerator/` — all tests pass
- [x] `go vet ./pkg/controller/gsakeysecretgenerator/` — clean
- [x] `gofmt -l pkg/controller/gsakeysecretgenerator/` — clean
- [x] Empirical envtest confirmation that an Update with empty `ResourceVersion` overwrites the existing Secret on kube-apiserver v1.29

## Reproduction without this fix

```yaml
# Victim Secret (created earlier by anything other than KCC):
apiVersion: v1
kind: Secret
metadata:
  name: production-db-password
  namespace: tenant-a
type: Opaque
data:
  password: aHVudGVyMg==
---
# Attacker objects in tenant-a:
apiVersion: iam.cnrm.cloud.google.com/v1beta1
kind: IAMServiceAccount
metadata: { name: pwn-sa, namespace: tenant-a }
---
apiVersion: iam.cnrm.cloud.google.com/v1beta1
kind: IAMServiceAccountKey
metadata:
  name: production-db-password   # collides with victim
  namespace: tenant-a
spec:
  serviceAccountRef: { name: pwn-sa }
```

After the reconciler runs, `kubectl get secret -n tenant-a production-db-password -o yaml` shows `data.password` replaced with `data."key.json"` (the attacker's GCP SA key), `ownerReferences` pointing to `IAMServiceAccountKey/production-db-password`, and the `managed-by-cnrm: true` label. With this PR, the reconciler returns an error and the victim Secret is left intact.

### DIFF
--- a/pkg/controller/gsakeysecretgenerator/gsakey_secret_generator.go
+++ b/pkg/controller/gsakeysecretgenerator/gsakey_secret_generator.go
@@ -157,6 +157,21 @@ func (r *ReconcileSecret) Reconcile(ctx context.Context, request reconcile.Reque
 
 	originalSecret := &corev1.Secret{}
 	if err = r.Get(ctx, request.NamespacedName, originalSecret); err == nil {
+		// Refuse to overwrite a pre-existing Secret that was not previously
+		// created by this controller. Otherwise an IAMServiceAccountKey CRD
+		// with a name that collides with any existing Secret in the same
+		// namespace would replace that Secret's contents with the SA key
+		// material and re-parent it via OwnerReferences.
+		if originalSecret.Labels[label.CnrmManagedKey] != "true" {
+			msg := fmt.Sprintf("refusing to overwrite Secret %v/%v: it is not managed by Config Connector (missing label %s=true); rename the IAMServiceAccountKey or remove the conflicting Secret",
+				request.Namespace, request.Name, label.CnrmManagedKey)
+			r.recorder.Eventf(u, corev1.EventTypeWarning, k8s.UpdateFailed, eventMessageTemplate, u.GetName(), u.GetNamespace(), msg)
+			return reconcile.Result{}, fmt.Errorf("%s", msg)
+		}
+		// Bind to the observed ResourceVersion so the Update fails-closed
+		// if the Secret changed between Get and Update (e.g. an attacker
+		// trying to race a label removal in).
+		secret.ResourceVersion = originalSecret.ResourceVersion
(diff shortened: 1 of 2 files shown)


evilgensec:
Gentle bump on this one. Has anyone had a chance to look yet? Happy to rebase or adjust if that helps.

PR #8592, opened 2026-05-22
Add brownfield-reviewgen-new-types.md agent

ldanielmadariaga:
Adding brownfield reviewgen instructions for step 1. These are very similar to the greenfield instructions with further restrictions for backwards compatibility. Greenfield instrucitons: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/.agents/greenfield-reviewgen-new-types.md


.agents/brownfield-reviewgen-new-types.md
acpana, line 1: let's try it out! Maybe we will get a chance to combine these two (greenfield, brownfield) in the future!

acpana (approved):
/lgtm
/approve

PR #8651, opened 2026-05-26
feat: Allow setting nodeConfig with "remove-default-node-pool" annotation in ContainerCluster

### LINKED ISSUE #7274: Feature Request: Support Network Tags for Default Node Pools in Private Clusters when using remove-default-node-pool directive
# Feature Request: Support Network Tags for Default Node Pools in Private Clusters when using `remove-default-node-pool` directive

**Proposed by:** `fkc1e100`

## Description

Currently, when creating a private GKE cluster via Config Connector and utilizing the `cnrm.cloud.google.com/remove-default-node-pool: "true"` annotation, the operator attempts to clean up the `nodeConfig` field to prevent recreation loops (as implemented in `pkg/krmtotf/legacygcpmanagedfields.go`).

However, for private clusters deployed in networks require specific **Network Tags** (e.g., for egress firewalls or master-to-node communication), stripping `nodeConfig` strips these tags. This causes the default node pool creation (triggered by GCP automatically during cluster creation) to fail or hang because the nodes lack the required tags to join the cluster or communicate properly.

This is a request to support passing network tags to the default node pool creation to ensure it succeeds, before it is subsequently deleted by the directive.

## Relation to Existing Issues

This issue is a direct follow-up to **#441 (ContainerCluster nodeConfig immutable/recreation loop)**. While KCC introduced a fix to strip `nodeConfig` to prevent recreation loops, it left unaddressed the side-effect for private clusters that rely on tags within `nodeConfig` to allow nodes to route traffic during initial provisioning. Issue #441 remains **Open** on GitHub.

## Use Case

- Deploying GKE using Config Connector `ContainerCluster` in a shared VPC or restricted network.
- The network security policy requires all VM instances (including GKE nodes in the default pool) to have specific tags for internet access or internal routing.
- The user wants the default pool deleted immediately after cluster creation via `remove-default-node-pool: "true"`.

## Current Behavior

KCC strips `nodeConfig` (and thus the tags) to prevent state conflicts, but this causes the default pool to fail creation in tagged networks.

## Requested Feature

Introduce a way to pass network tags to the default node pool creation *even when it is slated for deletion*, or ensure that tags specified in `nodeConfig` are preserved for the initial default pool creation phase before the pool is wiped.

Alternatively, allow a subset of `nodeConfig` (specifically `tags`) to be preserved during the initial creation even if the rest of the pool is managed by GCP/KCC.

gemmahou:
### BRIEF Change description

Fixes: #7274 

A Terraform validation error is triggered([code](https://github.com/gemmahou/terraform-provider-google-beta/blob/main/google-beta/services/container/resource_container_cluster.go#L8321-L8322)) when all three conditions are met: **1) the `cnrm.cloud.google.com/remove-default-node-pool: true` directive exists in the desired state, 2) `nodeVersion` exists in the desired state, and 3) TF treats the resource as a new resource.**

Based on a recent customer issue, we discovered that when both `remove-default-node-pool` and `nodeConfig` are specified, this same validation error occurs after the default node pool has been deleted. The breakdown of why this happens is as follows:

1. **the `cnrm.cloud.google.com/remove-default-node-pool: true` directive exists**: This is defined by the user in their configuration.
2. **`nodeVersion` exists**: KCC's automated GCP unmanaged fields logic([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/k8s/managedfields.go#L201-L205)) fetches the `nodeVersion` from GCP and overlays it onto the desired state during reconciliation.
3. **The resource is treated as "new"**: Once the default pool is deleted in GCP, `nodeConfig` also removed from the live state. resulting a diff between the desired state (still contains nodeConfig set by user) and the live state. Because `nodeConfig` is marked as `ForceNew` in the Terraform schema, the diff forces a resource replacement in TF, tricking it into thinking it is a brand-new resource.

KCC recommends against setting `remove-default-nodepool` alongside fields like `nodeVersion` and `nodeConfig`. These fields are only relevant to the default pool which is intended for deletion, specifying them is invalid and will likely result in errors. 

However, customers may require setting `nodeConfig` to satisfy system or policy requirements during initial cluster provisioning. To support this, we introduced the "remove-default-node-pool-allow-node-config" annotation. This allows customers to set a `nodeConfig` to successfully pass their requirements during cluster creation, while allowing KCC to subsequently remove it after removing the default pool to align with the recommended practice.

```release-note
Introduce "cnrm.cloud.google.com/remove-default-node-pool-allow-node-config" annotation in ContainerCluster to allow setting "nodeConfig" with "cnrm.cloud.google.com/remove-default-node-pool".
```

```docs

```

#### Intended Milestone

### Tests you have done

Tried to apply a resource with "cnrm.cloud.google.com/remove-default-node-pool" and "nodeConfig" .
Before(Without new annotation):
```
  Normal   UpToDate      13m                   containercluster-controller  Update in progress
  Normal   UpToDate      8m20s (x2 over 13m)   containercluster-controller  The resource is up to date
  Warning  UpdateFailed  115s (x6 over 3m58s)  containercluster-controller  Update call failed: error calculating diff: 1 error occurred:
           * node_version can only be specified if remove_default_node_pool is not true
```
After(With new annotation):
```
Events:
  Type    Reason    Age               From                         Message
  ----    ------    ----              ----                         -------
  Normal  Updating  11m               containercluster-controller  Update in progress
  Normal  UpToDate  10m               containercluster-controller  The resource is up to date
  Normal  UpToDate  5m (x2 over 10m)  containercluster-controller  The resource is up to date
```

### DIFF
--- a/mockgcp/mockcontainer/cluster.go
+++ b/mockgcp/mockcontainer/cluster.go
@@ -120,7 +120,7 @@ func (s *ClusterManagerV1) CreateCluster(ctx context.Context, req *pb.CreateClus
 
 	obj.ServicesIpv4Cidr = "34.118.224.0/20"
 
-	if err := s.populateClusterDefaults(name.Project, obj); err != nil {
+	if err := s.populateClusterDefaults(name.Project, obj, true); err != nil {
 		return nil, err
 	}
 
@@ -362,7 +362,8 @@ func (s *ClusterManagerV1) UpdateCluster(ctx context.Context, req *pb.UpdateClus
 		return nil, status.Errorf(codes.InvalidArgument, "update was not fully implemented ClusterUpdate=%v", prototext.Format(update))
 	}
 
-	if err := s.populateClusterDefaults(name.Project, obj); err != nil {
+	if err := s.populateClusterDefaults(name.Project, obj, false); err != nil {
+
 		return nil, err
 	}
 
@@ -473,12 +474,27 @@ func (s *ClusterManagerV1) DeleteCluster(ctx context.Context, req *pb.DeleteClus
 	})
 }
 
-func (s *ClusterManagerV1) populateClusterDefaults(project *projects.ProjectData, obj *pb.Cluster) error {
-	if obj.NodeConfig == nil {
-		obj.NodeConfig = &pb.NodeConfig{}
+func (s *ClusterManagerV1) populateClusterDefaults(project *projects.ProjectData, obj *pb.Cluster, isCreate bool) error {
+	hasDefaultPool := false
+	for _, np := range obj.NodePools {
+		if np.Name == "default-pool" {
+			hasDefaultPool = true
+			break
+		}
 	}
-	if err := s.populateNodeConfig(obj.NodeConfig); err != nil {
-		return err
+	if isCreate && len(obj.NodePools) == 0 {
+		hasDefaultPool = true
+	}
+
+	if hasDefaultPool {
+		if obj.NodeConfig == nil {
+			obj.NodeConfig = &pb.NodeConfig{}
+		}
+		if err := s.populateNodeConfig(obj.NodeConfig); err != nil {
+			return err
+		}
+	} else {
+		obj.NodeConfig = nil
 	}
 
 	// Populate new fields based on deprecated fields
--- a/pkg/krmtotf/krmtotf.go
+++ b/pkg/krmtotf/krmtotf.go
@@ -56,12 +56,10 @@ func KRMResourceToTFResourceConfig(r *Resource, c client.Client, smLoader *servi
 //   - defaultLabels: if set, these labels will be added to tfConfig.
 func KRMResourceToTFResourceConfigFull(r *Resource, c client.Client, smLoader *servicemappingloader.ServiceMappingLoader,
 	liveState *terraform.InstanceState, jsonSchema *apiextensions.JSONSchemaProps, mustResolveSensitiveFields bool) (tfConfig *terraform.ResourceConfig, secretVersions map[string]string, err error) {
-	fmt.Printf("DEBUG: KRMResourceToTFResourceConfigFull start for %v, generation %v, jsonSchema is nil: %v\n", r.GetName(), r.GetGeneration(), jsonSchema == nil)
 	config := deepcopy.MapStringInterface(r.Spec)
 	if config == nil {
 		config = make(map[string]interface{})
 	}
-	fmt.Printf("DEBUG: KRMResourceToTFResourceConfigFull jsonSchema is nil: %v\n", jsonSchema == nil)
 	if jsonSchema != nil {
 		if err := ResolveLegacyGCPManagedFields(r, liveState, config); err != nil {
 			return nil, nil, fmt.Errorf("error resolving legacy GCP-managed fields: %w", err)
@@ -173,10 +171,8 @@ func KRMResourceToTFResourceConfigFull(r *Resource, c client.Client, smLoader *s
 	state := InstanceStateToMap(r.TFResource, liveState)
 	config, err = withResourceCustomResolvers(config, state, r.Kind, r.TFResource)
 	if err != nil {
-		fmt.Printf("DEBUG: KRMResourceToTFResourceConfigFull custom resolver error: %v\n", err)
 		return nil, nil, fmt.Errorf("error running resource custom resolver: %w", err)
 	}
-	fmt.Printf("DEBUG: KRMResourceToTFResourceConfigFull success for %v, generation %v\n", r.GetName(), r.GetGeneration())
 	return MapToResourceConfig(r.TFResource, config), secretVersions, nil
 }
 
--- a/pkg/krmtotf/krmtotf.go
+++ b/pkg/krmtotf/krmtotf.go
@@ -56,10 +56,12 @@ func KRMResourceToTFResourceConfig(r *Resource, c client.Client, smLoader *servi
 //   - defaultLabels: if set, these labels will be added to tfConfig.
 func KRMResourceToTFResourceConfigFull(r *Resource, c client.Client, smLoader *servicemappingloader.ServiceMappingLoader,
 	liveState *terraform.InstanceState, jsonSchema *apiextensions.JSONSchemaProps, mustResolveSensitiveFields bool) (tfConfig *terraform.ResourceConfig, secretVersions map[string]string, err error) {
+	fmt.Printf("DEBUG: KRMResourceToTFResourceConfigFull start for %v, generation %v, jsonSchema is nil: %v\n", r.GetName(), r.GetGeneration(), jsonSchema == nil)
 	config := deepcopy.MapStringInterface(r.Spec)
 	if config == nil {
 		config = make(map[string]interface{})
 	}
+	fmt.Printf("DEBUG: KRMResourceToTFResourceConfigFull jsonSchema is nil: %v\n", jsonSchema == nil)
 	if jsonSchema != nil {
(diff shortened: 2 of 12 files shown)


pkg/test/resourcefixture/testdata/directives/removedefaultnodepool/_http.log
```diff
@@ -379,1800 +379,6 @@ X-Xss-Protection: 0
     "serviceExternalIpsConfig": {},
     "subnetwork": "projects/${projectId}/regions/us-central1/subnetworks/default"
   },
-  "nodeConfig": {
```
gemmahou, line 382: This is expected: nodeConfig field no longer exists after default pool is deleted

pkg/test/resourcefixture/testdata/directives/removedefaultnodepool/_http.log
```diff
@@ -379,1800 +379,6 @@ X-Xss-Protection: 0
...
-        "variantConfig": {
-          "variant": "DEFAULT"
-        }
-      },
-      "nodeKubeletConfig": {
-        "insecureKubeletReadonlyPortEnabled": false
-      }
-    }
-  },
-  "nodePools": [
```
gemmahou, line 433: This is expected: nodePools should nor longer exist as the only default pool is deleted

tests/e2e/testdata/scenarios/containercluster/removedefaultnodepool/_http01.log
```diff
@@ -138,244 +129,16 @@ X-Xss-Protection: 0
...
-  },
-  "status": "RUNNING",
-  "subnetwork": "default",
-  "userManagedKeysConfig": {},
-  "zone": "us-central1-a"
-}
-
----
-
-GET https://container.googleapis.com/v1beta1/projects/${projectId}/locations/us-central1-a/clusters/cluster-${uniqueId}?alt=json&prettyPrint=false
```
gemmahou: Comparing with realGCP, the diff comes from extra GET cluster call. Duplicated GET calls should not affect anything so we can probably accept this diff.

tests/e2e/testdata/scenarios/containercluster/removedefaultnodepool/script.yaml
```diff
@@ -0,0 +1,51 @@
...
+# Without the allow-node-config annotation, this should fail with the diff error.
+# We use APPLY-10-SEC to apply it and wait a bit, then it will be captured in golden file.
+TEST: APPLY-10-SEC
+apiVersion: container.cnrm.cloud.google.com/v1beta1
+kind: ContainerCluster
+metadata:
+  annotations:
+    cnrm.cloud.google.com/remove-default-node-pool: "true"
+    test.cnrm.cloud.google.com/reconcile-cookie: "v1"
+    kubectl.kubernetes.io/last-applied-configuration: |
```
gemmahou, line 25: Our test framework uses "server-side apply"([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/350d55443a00f2b3acb98a8785d28b8525fac391/tests/e2e/script_test.go#L509)) and customer uses "kubectl apply". This [code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/17071f12991532a7783f4c40abbf832c45a340cd/pkg/krmtotf/legacygcpmanagedfields.go#L315-L318) will directly remove problematic field "nodeConfig" if it's not considered "last-applied"([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/17071f12991532a7783f4c40abbf832c45a340cd/pkg/krmtotf/legacygcpmanagedfields.go#L297-L300)) so we are not able to reproduce.

To accurately simulate this scenario, we need to include the "last-applied-configuration" annotation in the test yaml.

pkg/krmtotf/krmtotf.go
```diff
@@ -56,10 +56,12 @@ func KRMResourceToTFResourceConfig(r *Resource, c client.Client, smLoader *servi
 //   - defaultLabels: if set, these labels will be added to tfConfig.
 func KRMResourceToTFResourceConfigFull(r *Resource, c client.Client, smLoader *servicemappingloader.ServiceMappingLoader,
 	liveState *terraform.InstanceState, jsonSchema *apiextensions.JSONSchemaProps, mustResolveSensitiveFields bool) (tfConfig *terraform.ResourceConfig, secretVersions map[string]string, err error) {
+	fmt.Printf("DEBUG: KRMResourceToTFResourceConfigFull start for %v, generation %v, jsonSchema is nil: %v\n", r.GetName(), r.GetGeneration(), jsonSchema == nil)
```
maqiuyujoyce: Please clean up.

pkg/krmtotf/krmtotf.go
```diff
@@ -56,10 +56,12 @@ func KRMResourceToTFResourceConfig(r *Resource, c client.Client, smLoader *servi
...
 	config := deepcopy.MapStringInterface(r.Spec)
 	if config == nil {
 		config = make(map[string]interface{})
 	}
+	fmt.Printf("DEBUG: KRMResourceToTFResourceConfigFull jsonSchema is nil: %v\n", jsonSchema == nil)
```
maqiuyujoyce: ditto

pkg/krmtotf/krmtotf.go
```diff
@@ -146,8 +173,10 @@ func KRMResourceToTFResourceConfigFull(r *Resource, c client.Client, smLoader *s
 	state := InstanceStateToMap(r.TFResource, liveState)
 	config, err = withResourceCustomResolvers(config, state, r.Kind, r.TFResource)
 	if err != nil {
+		fmt.Printf("DEBUG: KRMResourceToTFResourceConfigFull custom resolver error: %v\n", err)
```
maqiuyujoyce: ditto

pkg/krmtotf/krmtotf.go
```diff
@@ -146,8 +173,10 @@ func KRMResourceToTFResourceConfigFull(r *Resource, c client.Client, smLoader *s
...
 		return nil, nil, fmt.Errorf("error running resource custom resolver: %w", err)
 	}
+	fmt.Printf("DEBUG: KRMResourceToTFResourceConfigFull success for %v, generation %v\n", r.GetName(), r.GetGeneration())
```
maqiuyujoyce: ditto

mockgcp/mockcontainer/cluster.go
```diff
@@ -473,12 +474,27 @@ func (s *ClusterManagerV1) DeleteCluster(ctx context.Context, req *pb.DeleteClus
 	})
 }
 
-func (s *ClusterManagerV1) populateClusterDefaults(project *projects.ProjectData, obj *pb.Cluster) error {
-	if obj.NodeConfig == nil {
-		obj.NodeConfig = &pb.NodeConfig{}
+func (s *ClusterManagerV1) populateClusterDefaults(project *projects.ProjectData, obj *pb.Cluster, isCreate bool) error {
+	hasDefaultPool := false
```
gemmahou, line 478: To match realGCP, mockGCP should not populate nodeConfig field in Cluster resource, if the default pool is gone.

tests/e2e/testdata/scenarios/containercluster/removedefaultnodepool/script.yaml
maqiuyujoyce: Thank you for the explanation. It's interesting to find that `kubectl` still doesn't support SSA by default (https://github.com/kubernetes/enhancements/issues/3805).

maqiuyujoyce (approved):
/lgtm

PR #8860, opened 2026-05-29
chore: Improve types generator to handle output-only fields properly

gemmahou:
### BRIEF Change description

Improve types generator to handle output-only fields properly.
Generated by Gemini CLI.

There are some output-only fields in the types.generated.go that should be use "ObeservedState" version struct but not, or mistakenly pointing to an empty struct.

Example:
[Membership](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/redis/v1beta1/types.generated.go#L579) is output-only, is should use the [MembershipObservedState](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/redis/v1beta1/types.generated.go#L585) struct, not [Membership](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/redis/v1beta1/types.generated.go#L301) struct which is empty.

The types generator depends on the structure of the GCP proto, so it cannot be perfect. Our goal is to cover the general use cases, and manual intervention will still be necessary.

The rule for output-only fields and ObservedState version are:
1.  A message is identified as ObservedState version if 1) itself is marked as output-only, 2) it contains output-only primitive field, 3) (recursive)it contains nested fields that are marked as output-only or the nested fields contain output-only fields
2. After the 3 rules in the first step, if an ObservedState version message results in an empty struct, remove it and fallback to the regular struct. 

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_datacatalogentries.datacatalog.cnrm.cloud.google.com.yaml
```diff
@@ -929,9 +929,21 @@ spec:
                           here.
                         items:
                           properties:
-                            filePath:
-                              description: 'Required. Full file path. Example: `gs://bucket_name/a/b.txt`.'
-                              type: string
+                            gcsTimestamps:
```
gemmahou, line 935: The CRD change makes sense, we identified missing output-only fields here.
Also it's v1alpha1 so we should accept the change.

apis/dataproc/v1alpha1/types.generated.go
```diff
@@ -1561,10 +1565,9 @@ type Batch_StateHistoryObservedState struct {
 type InstanceFlexibilityPolicyObservedState struct {
 	// Output only. A list of instance selection results in the group.
 	// +kcc:proto:field=google.cloud.dataproc.v1.InstanceFlexibilityPolicy.instance_selection_results
-	InstanceSelectionResults []InstanceFlexibilityPolicy_InstanceSelectionResult `json:"instanceSelectionResults,omitempty"`
+	InstanceSelectionResults []InstanceFlexibilityPolicy_InstanceSelectionResultObservedState `json:"instanceSelectionResults,omitempty"`
```
acpana, line 1568: if this is the type we are taking as is in the ObservedState for this resource (haven't double checked), we should insist that the casing be camel! We can follow up

acpana (commented):
a bit on the fence about this but if Lucas is okay happy to have this merged in!

justinsb (dismissed):
/approve
/lgtm

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_datacatalogentries.datacatalog.cnrm.cloud.google.com.yaml
ldanielmadariaga: Filepath is a part of the output only field sampleGCSFileSpec, it should not be removed. How can we make sure we avoid this issue for other cases?


Definition: https://docs.cloud.google.com/data-catalog/docs/reference/rpc/google.cloud.datacatalog.v1#google.cloud.datacatalog.v1.GcsFileSpec

Declared as output only: https://docs.cloud.google.com/data-catalog/docs/reference/rpc/google.cloud.datacatalog.v1#google.cloud.datacatalog.v1.GcsFilesetSpec

dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
ldanielmadariaga, line 1: These changes should really have unit tests

ldanielmadariaga (changes requested):

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_datacatalogentries.datacatalog.cnrm.cloud.google.com.yaml
gemmahou: Good catch. Filepath is excluded because it does not has "[(google.api.field_behavior) = OUTPUT_ONLY]"([ref](https://github.com/googleapis/googleapis/blob/5481332007e57ad3d9cb81e83ff6bf9c449476b6/google/cloud/datacatalog/v1/gcs_fileset_spec.proto#L69-L80)), and our generator really relies on the proto annotation to determine output-only fields.

Regarding this field, it's marked as `REQUIRED`, I'm a bit confused because `REQUIRED` and `OUTPUT_ONLY` seem to conflict. I do see there's another case(mentioned in PR description as well), [SystemTimestamps](https://github.com/googleapis/googleapis/blob/master/google/cloud/datacatalog/v1/timestamps.proto#L30) contains three fields but only one is marked output-only.

Above are edge cases where a parent field is `OUTPUT_ONLY` but its child fields are inconsistently annotated. To address this, I updated the generator to include all nested fields regardless of their individual annotations if the parent field is `OUTPUT_ONLY`: https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/8860/changes/38d45a0ae34c57444773b62d96b8118cf6c333ab

dev/tools/controllerbuilder/pkg/codegen/typegenerator.go
gemmahou: Added unit test

config/crds/resources/apiextensions.k8s.io_v1_customresourcedefinition_datacatalogentries.datacatalog.cnrm.cloud.google.com.yaml
gemmahou: cc @justinsb in this PR, even the field is `OUTPUT_ONLY`, I continue using the regular structs when the content is identical with the ObservedState version([example](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/apis/dataproc/v1alpha1/types.generated.go#L1589)) to reduce code duplication and minimize code changes, see point2 in PR description. But I think yours #8143 is correct we should change it to ObservedState and remove the regular one. Please let me know if you suggest a followup on this.

ldanielmadariaga: Sounds good agree that marking as observed state if the parent is output only makes sense. I was also a bit miffed about the 'required' annotation I think it may mean that the field should always be part of the response or it may betray that this proto was also used for input at some point. Thanks for the fix!

ldanielmadariaga (approved):

PR #8967, opened 2026-06-01
Terraform Codegen: Create 2 skills to update and test TF resources. Validate with ContainerNodePool

ldanielmadariaga:
Terraform Codegen: Create 2 skills to update and test TF resources. Validate with ContainerNodePool which hasn't yet been migrated to direct.

ContainerNodePool tests `Case A of Section 2` of the generate skill.


gemmahou:
The validation test failed due to https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/0358fb8c6b984ffd6f9ee7c6cdf9445d35fa2bff/scripts/github-actions/ga-validation.sh#L39

Previously, we required TF code changes to be in a separate commit. @maqiuyujoyce , do you know the exact reason? Correct me if I’m wrong, it was helpful for us to track all TF patches and to keep sync'd with upstream TF repo when we relied on Terraform. If that's no longer necessary, I feel we could remove this check now.

.gemini/skills/test-terraform-fields/SKILL.md
```diff
@@ -0,0 +1,146 @@
...
+
+## 3. Recording Ground Truth against Real GCP
+
+To establish a baseline, run the tests against real GCP to produce golden logs.
+
+1. **GCP Project & Credentials Check**:
+   - Ensure the terminal is authenticated (e.g. `gcloud auth application-default login`).
+   - KCC uses the Kubernetes namespace as the GCP project name by default. Ensure the test namespace maps to an active GCP project or is configured with appropriate annotations.
+2. **Run E2E Recorder**:
+   - Run `hack/record-gcp <testname>`.
```
acpana, line 42: we should give an example for the testname as it often confuses the LLMs

acpana: there's also ways to run ALL tests (tests-e2e-sql for instance should run the SQL tests only).

.gemini/skills/test-terraform-fields/SKILL.md
acpana, line 1: nit: This is fine for now but we may want to break the SKILL of actually running tests away from this skill of testing terraform fields.

.gemini/skills/update-terraform-fields/SKILL.md
```diff
@@ -0,0 +1,142 @@
...
+4. **Verify Compilation**:
+   - Run the compiler check locally:
+     ```bash
+     go vet ./third_party/github.com/hashicorp/terraform-provider-google-beta/...
+     ```
+5. **Stage Provider Changes**:
+   - Stage the modified third-party files. You can commit them together with KCC changes as the repository no longer enforces subtree change isolation:
+     ```bash
+     git add third_party/
+     ```
```
acpana, line 63: In keeping with previous conventions we may want to commit third_party changes separately.

acpana (dismissed):
I like this! A few nits/ followups but LGTM

.gemini/skills/update-terraform-fields/SKILL.md
```diff
@@ -0,0 +1,142 @@
...
+   - Run `git diff config/crds/resources/` to verify that the new field is properly generated in the CRD schema.
+3. **Update Go Client**:
+   - Run:
+     ```bash
+     make generate-go-client
+     ```
+   - Verify if any generated files under `pkg/clients/generated/` are modified using `git status` or `git diff`. If there are modifications, stage and commit them.
+
+### Case B: The resource HAS Go types in `apis/`
```
gemmahou, line 101: The TF controller relies on servicemappings to translate special KRM object fields especially reference fields into TF. When implementing reference fields, besides updating the Go types and CRD, we have to ensure the references are also configured within servicemappings.

Just a note and not a blocker for this PR, we can add a step to handle reference fields when we have an use case.

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/node_config.go
```diff
@@ -666,6 +666,23 @@ func schemaNodeConfig() *schema.Schema {
 						},
 					},
 				},
+				"windows_node_config": {
```
gemmahou, line 669: I'm checking the [TF provider code](https://github.com/gemmahou/terraform-provider-google-beta/blob/main/google-beta/services/container/node_config.go#L1059-L1076), and this field does not have "ForceNew", but it's "Computed"(serve-generated). Wondering why it's marked as "ForceNew" here?

We can improve the prompt to copy the schema directly from TPG, once the TF provider file is found. We can also have a check to compare with the TF provider file, to double-check if the diff is expected.

.gemini/skills/test-terraform-fields/SKILL.md
ldanielmadariaga: hmmm interesting point, I think I'll keep as is for now because I do prefer case specific test instructions but do agree those can also live in a separate skill

.gemini/skills/test-terraform-fields/SKILL.md
ldanielmadariaga: Done

.gemini/skills/update-terraform-fields/SKILL.md
ldanielmadariaga: I think this doesn't that much sense in the new agentic workflow hence why I removed the validation check.

.gemini/skills/update-terraform-fields/SKILL.md
ldanielmadariaga: Updated the instructions

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/node_config.go
ldanielmadariaga: You're right this was made mutable, ingested the new changes from TF and updated

acpana (approved):

PR #9085, opened 2026-06-03
Update migration tracker to auto-detect Direct controller progress steps

### LINKED ISSUE #9386: Migration tracker: Test fixtures existence incorrectly flags resource migration as In Progress
Existence of test fixtures is not evidence that the migration is in progress for a v1beta1 resource (and for some v1alpha1 resources). The fixtures tests were added whenever the original resources were supported. This causes the migration tracker to incorrectly flag many resources as 'In Progress'. The tracker should only flag the 'tests' step as complete or the resource as 'In Progress' if direct migration has actually started (e.g. types generated, controller implemented) or is completed.

anhdle-sso:
Closes #9386. Fixes the logic in generate_data.py to not automatically set the tests step to True just because test fixtures exist. It will now only be auto-set if the resource is fully completed (i.e. 'Direct' in supportedControllers).

### DIFF
--- a/dev/migration-tracker/data.json
+++ b/dev/migration-tracker/data.json
@@ -1592,7 +1592,7 @@
     },
     "mocksLastRefreshed": "Never",
     "notes": "",
-    "sortOrder": 485,
+    "sortOrder": 487,
     "downstreamCount": 0
   },
   {
@@ -1882,7 +1882,7 @@
     },
     "mocksLastRefreshed": "Never",
     "notes": "",
-    "sortOrder": 486,
+    "sortOrder": 485,
     "downstreamCount": 4
   },
   {
@@ -2565,7 +2565,7 @@
     },
     "mocksLastRefreshed": "Never",
     "notes": "",
-    "sortOrder": 487,
+    "sortOrder": 486,
     "downstreamCount": 0
   },
   {
--- a/dev/migration-tracker/data.json
+++ b/dev/migration-tracker/data.json
@@ -1,32 +1,6 @@
 [
   {
     "group": "apigateway",
-    "kind": "APIGatewayAPI",
-    "version": "v1beta1",
-    "controllerType": "Direct",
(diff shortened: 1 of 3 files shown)


dev/migration-tracker/data.json
```diff
@@ -1,4 +1,30 @@
 [
+  {
+    "group": "aistreams",
+    "kind": "AIStreamsCluster",
+    "version": "v1beta1",
+    "controllerType": "Direct",
+    "defaultController": "Direct",
+    "supportedControllers": [
+      "Direct"
+    ],
```
acpana: this is a direct resource not a "migrated" resource

dev/migration-tracker/generate_data.py
```diff
@@ -53,6 +54,43 @@ def get_implemented_types(apis_dir="../../apis"):
...
+            
+    return controllers, fuzzers
+
+def find_test_fixtures(repo_root):
+    test_kinds = set()
+    basic_dir = os.path.join(repo_root, 'pkg', 'test', 'resourcefixture', 'testdata', 'basic')
+    if not os.path.exists(basic_dir):
+        return test_kinds
+    for root, dirs, files in os.walk(basic_dir):
+        if 'create.yaml' in files:
```
maqiuyujoyce: Existence of test fixtures is not an evidence that the migration is in progress for a v1beta1 resource (and for some v1alpha1 resources). The fixtures tests were added whenever the original resources were supported.

maqiuyujoyce (commented):
Not a blocker for this PR, but can we run this script as either a periodical job or as part of presubmit so that we can have the updated info?

dev/migration-tracker/generate_data.py
anhdle-sso: sg, I remove the detection on test for now

maqiuyujoyce (approved):
/lgtm

PR #9087, opened 2026-06-03
Verify compilation of integration-tagged tests in unit-test presubmit

anhdle-sso:
This pull request adds a new presubmit job `compile-integration-tests` to check that all integration test files compile successfully during PR checks. This prevents syntax or signature breakages in integration-tagged tests from going undetected.

Also fixes pre-existing compilation issues in:
- `pkg/controller/gsakeysecretgenerator/service_account_key_integration_test.go`
- `scripts/resource-autogen/servicemapping/embed/generated/main/assets_generate.go`

### DIFF
--- a/.github/workflows/ci-presubmit.yaml
+++ b/.github/workflows/ci-presubmit.yaml
@@ -56,6 +56,25 @@ jobs:
           path: /tmp/artifacts/
 
 
+  compile-integration-tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v6
+      - uses: ./dev/actions/cache-go
+      - name: "Run dev/ci/presubmits/compile-integration-tests"
+        run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/ci-presubmit.yaml
acpana, line 1: ah we've started generating this file a while back. We should probably have another check to see if "regen"-ing this file comes out clean or with a diff to catch issues like this

https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/76d8bd89531477b538b200296623124a1d0f3e54/dev/tasks/generate-github-actions#L26

anhdle-sso: good call, let me run the regen script to see if there is diff. Adding a check for another check might be over engineer.

anhdle-sso: Confirm running `dev/tasks/generate-github-actions` produce no diff

PR #9091, opened 2026-06-04
Refactor RedisCluster direct controller to follow up-to-date pattern

gemmahou:
### BRIEF Change description

Refactor RedisCluster direct controller to follow up-to-date pattern
1. use identity and reference
2. use shared methods to resolve refs and update status
3. ensure externalRef is assigned
4. use DiffForTopLevelFields to determine diffs for update

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


pkg/test/resourcefixture/testdata/basic/redis/v1beta1/rediscluster/fullrediscluster/_http.log
gemmahou, line 1: Order of the Patch calls changed, this is acceptable. The API requires update one field at a time.

pkg/test/resourcefixture/testdata/basic/redis/v1beta1/rediscluster/basicrediscluster/_http.log
```diff
@@ -528,9 +528,6 @@ User-Agent: kcc/${kccVersion} (+https://github.com/GoogleCloudPlatform/k8s-confi
 X-Goog-Request-Params: parent=projects%2F${projectId}%2Flocations%2Fus-central1
 
 {
-  "persistenceConfig": {
```
gemmahou, line 531: Previously we [assigned default value in the Adaptor](https://github.com/justinsb/k8s-config-connector/blob/master/pkg/controller/direct/redis/cluster/rediscluster_controller.go#L120-L122) before creating and updating the resource, which is probably incorrect. Since this field is not specified by the user, it should be omitted from the request body. The default value will be assigned during the comparison with the actual state.

pkg/controller/direct/redis/cluster/rediscluster_controller.go
```diff
@@ -322,14 +284,72 @@ func (a *redisClusterAdapter) Update(ctx context.Context, updateOp *directbase.U
...
+		maskedActual = RedisClusterSpec_ToProto(mapCtx, spec)
+		if mapCtx.Err() != nil {
+			return nil, mapCtx.Err()
+		}
+	}
+
+	maskedActual = populateDefaults(maskedActual)
+	desired = populateDefaults(proto.CloneOf(desired))
+
+	diffs, _, err := tags.DiffForTopLevelFields(ctx, desired.ProtoReflect(), maskedActual.ProtoReflect())
```
gemmahou, line 319: @suwandim updated to use DiffForTopLevelFields. New fields are handled without requiring modifications to the controller code.

ldanielmadariaga: Any reason why we can't use `CompareProtoMessageStructuredDiff` to get the modified paths? 


Usage: `paths, report, err := common.CompareProtoMessageStructuredDiff(resource, a.actual, common.BasicDiff)`

pkg/controller/direct/redis/cluster/rediscluster_controller.go
```diff
@@ -322,14 +284,72 @@ func (a *redisClusterAdapter) Update(ctx context.Context, updateOp *directbase.U
...
+		}
+		if cluster.TransitEncryptionMode == pb.TransitEncryptionMode_TRANSIT_ENCRYPTION_MODE_UNSPECIFIED {
+			cluster.TransitEncryptionMode = pb.TransitEncryptionMode_TRANSIT_ENCRYPTION_MODE_DISABLED
+		}
+		if cluster.ZoneDistributionConfig == nil {
+			cluster.ZoneDistributionConfig = &pb.ZoneDistributionConfig{}
+		}
+		if cluster.ZoneDistributionConfig.Mode == pb.ZoneDistributionConfig_ZONE_DISTRIBUTION_MODE_UNSPECIFIED {
+			cluster.ZoneDistributionConfig.Mode = pb.ZoneDistributionConfig_MULTI_ZONE
+		}
```
anhdle-sso: Would it make sense to move the defaulting to mapping? or make this a separate function?

pkg/controller/direct/redis/cluster/rediscluster_controller.go
anhdle-sso: does the API allow nested field update or just the top level?

gemmahou: Had the same question before and was told "CompareProtoMessageStructuredDiff" was not working well for this resource so we used "DiffForTopLevelFields" in [MemorystoreInstance](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/memorystore/instance_controller.go#L327) too. I haven't tested the real API here myself and I am assuming the existing http log is accurate - updating the any of the subfields should result in only the top-level path being included in the updateMask(See Patch persistenceConfig in the http log). While "CompareProtoMessageStructuredDiff" treats each subfields individually.

The API allows nested field update.

pkg/controller/direct/redis/cluster/rediscluster_controller.go
gemmahou: Our mapper contains generated part, so we may want to minimize the customization to mappers. I moved this to its own helper function.

pkg/controller/direct/redis/cluster/rediscluster_controller.go
anhdle-sso: from my previous encounter with this API, some sub field fieldmasks are accepted but not all of them. Using top field mask should be sufficient for KCC. A few nits:

- We still want to use `CompareProtoMessageStructuredDiff` for structured diff reporting so we know exact subfield (maybe a few level nested) trigger the update.
- Since the API requires multiple update call (one field at a time). It is good to sort the fields before processing through them to get consistent behavior.

ldanielmadariaga (approved):

pkg/controller/direct/redis/cluster/rediscluster_controller.go
gemmahou: - Let's follow up on this. I feel we can track both changed nested fields and top-level fields in a single compare function to satisfy both needs...
- "DiffForTopLevelFields" sorts fields: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/tags/diff.go#L48

PR #9111, opened 2026-06-04
Fix markdown table header separator formatting in CRD report

anfernee:
### BRIEF Change description

No bug for this.

The markdown table format is currently broken. It looks like [this](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/docs/reports/crd_report.md)

### Tests you have done

The fixed version is like [this](https://github.com/anfernee/k8s-config-connector/blob/fix/crd-report-markdown-formatting/docs/reports/crd_report.md)


anhdle-sso (approved):
looks good. 👍

PR #9388, opened 2026-06-05
Migrate RunJob direct controller to use last-changed-cookie annotation

### LINKED ISSUE #9387: Migrate RunJob direct controller to use last-changed-cookie annotation
The RunJob direct controller is currently the only controller using the legacy status-based LastModifiedCookie (via common.NewLegacyCookie). We should migrate it to use the standard 'cnrm.cloud.google.com/last-changed-cookie' annotation-based mechanism implemented in directbase.Operation. This will allow us to remove the legacy cookie implementation in pkg/controller/direct/common/hash.go.

anhdle-sso:
Closes #9387. Migrates RunJob from the legacy status-based cookie to the annotation-based cookie mechanism in directbase.Operation. This allowed us to delete the legacy cookie utility pkg/controller/direct/common/hash.go since RunJob was its last user.


pkg/controller/direct/run/job_controller.go
```diff
@@ -186,13 +183,9 @@ func (a *JobAdapter) Create(ctx context.Context, createOp *directbase.CreateOper
 	if mapCtx.Err() != nil {
 		return mapCtx.Err()
 	}
-	status.ExternalRef = direct.LazyPtr(a.id.String())
```
maqiuyujoyce, line 189: `status.externalRef` field should still be set, right?

pkg/controller/direct/run/job_controller.go
```diff
@@ -186,13 +183,9 @@ func (a *JobAdapter) Create(ctx context.Context, createOp *directbase.CreateOper
...
-	newCookie, err := common.NewLegacyCookie(a.desired, created)
-	if err != nil {
-		return fmt.Errorf("composing cookie: %w", err)
+	if err := createOp.SetLastModifiedCookie(ctx, a.desired, created); err != nil {
```
maqiuyujoyce, line 186: Nit: Maybe add the log?

pkg/controller/direct/run/job_controller.go
```diff
@@ -201,11 +194,7 @@ func (a *JobAdapter) Update(ctx context.Context, updateOp *directbase.UpdateOper
 	log := klog.FromContext(ctx)
 	log.V(2).Info("updating Job", "name", a.id)
 
-	currentCookie, err := common.NewLegacyCookie(a.desired, a.actual)
-	if err != nil {
-		return err
-	}
-	if currentCookie.Equal(a.lastModifiedCookie) {
+	if upToDate, err := updateOp.CompareLastModifiedCookie(a.desired, a.actual); err == nil && upToDate {
```
maqiuyujoyce, line 197: You didn't seem to return the error?

maqiuyujoyce (commented):
For the cookie migration, when an existing RunJob resource is reconciled for the first time after this update, the old cookie will only be in `.status.lastModifiedCookie` but the controller will expect it in the annotation. Does it mean:
1. we'll expect the force re-reconciliation at the upgrade time to fill out the cookie in the annotation?
2. the `.status.lastModifiedCookie` will left untouched (and then become stale)?

PR #9394, opened 2026-06-05
fix(compute): support sourceRegions in ComputeHealthCheck KCC resource

ldanielmadariaga:
fix(compute): support sourceRegions in ComputeHealthCheck KCC resource

This PR adds support for the `sourceRegions` field in the `ComputeHealthCheck` KCC resource.

Under the hood, this:
- Updates the KCC `ComputeHealthCheck` CRD and generated Go types to expose `sourceRegions`.
- Patches the vendored Terraform Google Beta provider to support `source_regions` in `google_compute_health_check`.
- Adds the necessary field mapper generation.
- Adds `sourceRegions` to `missingfields.txt` as a known missing field exception.

This is a prerequisite for external DNS routing policies (specifically Geolocation-based routing) to be able to refer to Health Checks with specific source regions.


pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computehealthcheck/globalcomputehealthchecksourceregions/update.yaml
```diff
@@ -0,0 +1,30 @@
...
+kind: ComputeHealthCheck
+metadata:
+  name: computehealthcheck-${uniqueId}
+spec:
+  checkIntervalSec: 40
+  timeoutSec: 10
+  healthyThreshold: 2
+  unhealthyThreshold: 3
+  location: global
+  sourceRegions:
```
gemmahou, line 25: Is sourceRegions mutable? If so, the update test should modify this field, i.e. change a region from the list.

gemmahou, line 25: super-nit: Can we rename this test to 'globalcomputehealthcheck-full'? If we add more fields in the future, we can put them into this full test case instead of creating a brand new test for every single field.

mockgcp/common/workflows/computenetwork_routes.go
```diff
@@ -69,13 +69,10 @@ func (e *Engine) CreateComputeNetworkRoutes(ctx context.Context, projectID strin
 			Project:       projectID,
 			RouteResource: route,
 		}
-		op, err := routesClient.Insert(ctx, req)
+		_, err := routesClient.Insert(ctx, req)
```
ldanielmadariaga, line 72: @acpana @gemmahou this seems fine to me since it's mockGCP but curious if it can have some hidden pitfalls.

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computehealthcheck/globalcomputehealthchecksourceregions/update.yaml
ldanielmadariaga: The API only seems to enforce that 3 regions are in place not that it's immutable let me give that a shot! https://docs.cloud.google.com/compute/docs/reference/rest/v1/healthChecks/get

ldanielmadariaga: I don't think this is necessarily full. Using source-regions per the API using forces some state to be enabled or disabled. Let's keep as is for now :)

ldanielmadariaga: Updated :)

ldanielmadariaga:
Superseded by https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/10259

Issue #9221, opened 2026-06-04
MissingField: AutomatedBackupConfig in RedisCluster


gemmahou:
In pkg/controller/direct/redis/cluster/cluster_fuzzer.go, we can see that we are missing support for [automated_backup_config](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/redis/cluster/cluster_fuzzer.go#L47)

This field should be surfaced as part of the spec.

Use the skill .gemini/skills/add-missing-field/SKILL.md to add support for the field. Ensure we have fixtures test coverage for the new field.

Please update the skill where you find it can be improved (and put observations that aren't necessarily as general in the journal.md file in the same folder)

PR #9602, opened 2026-06-09
fix: Fix permdiff caused by automatedbackupconfig default in RedisCluster

gemmahou:
### BRIEF Change description

Followup of #9221 

- Assign API default value to this field if unspecified. 
- Update mockGCP to include this field with default value in response body to match realGCP behavior .

By default, GCP sets the "automatedBackupConfig" mode to "disabled" if it is unspecified. During reconcile, comparing the desired(nil) and actual state(disabled) results in a diff, causing unnecessary Patch calls every time. We must assign the default value to this spec field to match actual state and avoid triggering a diff on this field.

```release-note

```

```docs

```

#### Intended Milestone

### Tests you have done


pkg/test/resourcefixture/testdata/basic/redis/v1beta1/rediscluster/basicrediscluster/_http.log
gemmahou, line 1: No diff from "automatedBackupConfig" field

ldanielmadariaga (changes requested):
Nit: Can we add a bit more context in the opening message as to why this was triggering a permadiff?

ldanielmadariaga (approved):

PR #9608, opened 2026-06-09
Exclude direct-only controllers from migration tracker

### LINKED ISSUE #9607: Exclude direct-only controllers from migration tracker
### Description

Direct-only controllers (like `AIStreamsCluster`, `AlloyDBInstance`, etc.) are currently being tracked in the migration tracker. 

Since these resources were either born direct or are already fully migrated (with legacy controllers removed), they do not need to be tracked as part of the active migration effort. Tracking them distorts the progress metrics and adds noise.

### Proposed Solution

Modify `dev/migration-tracker/generate_data.py` to automatically identify and skip resources that only support the `Direct` reconciler in `pkg/controller/resourceconfig/static_config.go`.

This will focus the tracker on active migrations (resources that still have TF/DCL versions, or are in transition).

anhdle-sso:
Closes #9607.

This PR implements a general rule to exclude resources that only support the `Direct` reconciler from the migration tracker. 

### Changes
*   Modified `dev/migration-tracker/generate_data.py` to skip resources where `SupportedControllers` in `static_config.go` contains only `Direct`.
*   Regenerated `dev/migration-tracker/data.json` to apply the new logic, reducing the tracked resources from 451 to 304.

This removes noise (like `AIStreamsCluster` and `AlloyDBInstance`) and focuses the tracker on active migrations.

### DIFF
--- a/dev/migration-tracker/data.json
+++ b/dev/migration-tracker/data.json
@@ -93,7 +93,7 @@
     ],
     "state": "Not Started",
     "steps": {
-      "gen-types": false,
+      "gen-types": true,
       "identity-reference": true,
       "mapper-fuzzer": false,
       "mocks": false,
@@ -143,7 +143,7 @@
     "dependencies": [],
     "state": "Not Started",
     "steps": {
-      "gen-types": false,
+      "gen-types": true,
       "identity-reference": true,
       "mapper-fuzzer": false,
       "mocks": false,
@@ -196,7 +196,7 @@
     ],
     "state": "Not Started",
     "steps": {
-      "gen-types": false,
+      "gen-types": true,
       "identity-reference": true,
       "mapper-fuzzer": false,
       "mocks": false,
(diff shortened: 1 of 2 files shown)


dev/migration-tracker/data.json
```diff
@@ -8581,49 +5734,50 @@
...
+      "Folder",
       "Project",
-      "Service"
+      "StorageBucket"
     ],
     "state": "Completed",
     "steps": {
       "gen-types": true,
-      "identity-reference": true,
+      "identity-reference": false,
```
maqiuyujoyce: I believe this step should be `true`.

dev/migration-tracker/data.json
```diff
@@ -7924,48 +5118,45 @@
...
     "dependencies": [
-      "MonitoringAlertPolicy",
-      "Project"
+      "IAMServiceAccount"
     ],
-    "state": "Completed",
+    "state": "Not Started",
     "steps": {
-      "gen-types": true,
+      "gen-types": false,
```
maqiuyujoyce: I think we have serviceaccountkey_types.go under iam/v1beta1.

maqiuyujoyce (commented):
I did a couple of spot checks and found that some steps incorrectly get reset to `false`. Could you fix them and verify that all the state reset is expected?

barney-s (approved):
/lgtm
/approve

PR #9610, opened 2026-06-09
gkehub: add structured reporting with new and old value for FeatureMembership

### LINKED ISSUE #9609: gkehub: add structured reporting with new and old value for FeatureMembership
Currently, the GKEHub FeatureMembership controller uses an incomplete structured reporting diff pattern, calling `report.AddField(path, nil, nil)` which only reports that a field changed but does not provide the actual old and new values.

This issue tracks refactoring the diffing logic in `gkehub/diffs.go` and `gkehub/featuremembership_controller.go` to directly compare the KRM desired spec with the Proto actual spec, and populate the structured diff report with both the old (actual) and new (desired) values. It also fixes a perpetual diff bug where `policyControllerHubConfig` was being compared using `reflect.DeepEqual` on two different struct types, always resulting in a mismatch.

anhdle-sso:
Closes #9609

### DIFF
--- a/pkg/controller/direct/gkehub/diffs.go
+++ b/pkg/controller/direct/gkehub/diffs.go
@@ -19,237 +19,443 @@ import (
 	"reflect"
 
 	krm "github.com/GoogleCloudPlatform/k8s-config-connector/apis/gkehub/v1beta1"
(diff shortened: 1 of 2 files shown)


maqiuyujoyce (approved):
/lgtm

PR #9658, opened 2026-06-10
kms: clean up and improve autokey config identity and deletion resolution

iamkonohamaru:
This PR cleans up and improves the identity resolution and deletion logic for the KMSAutokeyConfig resource to handle edge cases where referenced projects or folders have already been deleted.

Key Changes
Identity Resolution (apis/kms/v1beta1/autokeyconfig_identity.go):

Updated NewAutokeyConfigIdentity to parse Status.ExternalRef first if it is present.
Validates the Spec references (folderRef or projectRef) against the parsed identity without executing Kubernetes reference resolution. This prevents Delete reconciles from blocking when the parent project/folder has already been deleted.
Added comprehensive unit tests in apis/kms/v1beta1/autokeyconfig_identity_test.go covering all combinations of externalRef and Spec inputs.
Controller Cleanups (pkg/controller/direct/kms/autokeyconfig/autokeyconfig_controller.go):

Reference Resolution: Moved the resolution of keyProjectRef from AdapterForObject down to the Update method. This prevents resolution failures from blocking other operations like Delete.
Deletion Logic:
Updated Delete to return success immediately (return true, nil) if the resource is already UNINITIALIZED on GCP, skipping the rest of the code.
Handled NOT_FOUND errors during the update call in Delete (e.g., if the parent project/folder was already deleted in GCP), treating them as successful deletions.
Adapter Struct: Removed the unused desiredKeyProject field and stored the reader client.Reader to perform reference resolution inside Update.
Redundant Checks: Removed the unnecessary if a.actual == nil check in updateAutokeyConfig.
Schema Compatibility Cleanups:

Deleted the stale exception file tests/apichecks/testdata/exceptions/multi_version_crd_diff/KMSAutokeyConfig.diff. The schemas for v1alpha1 and v1beta1 are now fully identical, and the TestMultiVersionCRDNoDiff schema test passes.


apis/kms/v1beta1/autokeyconfig_identity.go
```diff
@@ -49,6 +49,44 @@ func (p *KMSAutokeyConfigParent) String() string {
 }
 
 func NewAutokeyConfigIdentity(ctx context.Context, reader client.Reader, obj *KMSAutokeyConfig) (*KMSAutokeyConfigIdentity, error) {
+	externalRef := common.ValueOf(obj.Status.ExternalRef)
```
gemmahou: I think the previous implementation is fine(please check our group chat). So the code change in this identity file is unnecessary.

apis/kms/v1beta1/autokeyconfig_identity_test.go
gemmahou, line 1: Let revert the identity file per my above comment and keep this test to ensure there's no issues.

iamkonohamaru: sure

gemmahou (approved):

PR #9683, opened 2026-06-10
ai: update skills, chores

acpana:
Per my greatest hits (aka most frequent comments) when reviewing PRs, this is what the bots think it will be useful

### DIFF
--- a/.gemini/skills/kcc-direct-greenfield-types-implementer/SKILL.md
+++ b/.gemini/skills/kcc-direct-greenfield-types-implementer/SKILL.md
@@ -41,6 +41,24 @@ Apply the baseline validations from `kcc-direct-base-types-implementer`, plus th
 - **Enums**: 
   - Use `*string` for the Go type of proto enum fields (do NOT use custom wrapped string types).
   - Use `// +kubebuilder:validation:Enum=VALUE1;VALUE2` to provide validation in the CRD while keeping the Go type simple.
+- **Naming Conventions & Acronyms**:
+  - Structs in `*_types.go` must be strictly CamelCased (no snake_case allowed).
+  - Ensure GCP acronyms are capitalized correctly in resource and field names (e.g., `APIHub`, `SIP`, `JSON`, `PDF`, `DER`, `CEL`).
+  - **No Exceptions**: Do **NOT** add new exceptions to `tests/apichecks/testdata/exceptions/acronyms.txt`. Fix the Go field names instead.
+- **Pointers**: Validate that optional or nullable fields (like `Location`, booleans, integers) are correctly typed as pointers (e.g., `*string`, `*bool`) rather than raw values.
+- **ETag Exclusion**: Remove `etag` fields from the KRM spec; they are generally disallowed. Do **NOT** add exceptions to `spec_dislike_etag.txt`.
+- **KCC References (`missingrefs.txt`)**:
+  - Ensure fields representing GCP resources are implemented using the KCC Ref pattern (e.g., `ProjectRef`, `NetworkRef`) rather than plain strings.
+  - **No Exceptions**: Do **NOT** add exceptions to `tests/apichecks/testdata/exceptions/missingrefs.txt`. Fix the underlying field types instead.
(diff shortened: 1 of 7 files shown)


.gemini/skills/kcc-direct-greenfield-types-implementer/SKILL.md
```diff
@@ -41,6 +41,24 @@ Apply the baseline validations from `kcc-direct-base-types-implementer`, plus th
 - **Enums**: 
   - Use `*string` for the Go type of proto enum fields (do NOT use custom wrapped string types).
   - Use `// +kubebuilder:validation:Enum=VALUE1;VALUE2` to provide validation in the CRD while keeping the Go type simple.
+- **Naming Conventions & Acronyms**:
+  - Structs in `*_types.go` must be strictly CamelCased (no snake_case allowed).
+  - Ensure GCP acronyms are capitalized correctly in resource and field names (e.g., `APIHub`, `SIP`, `JSON`, `PDF`, `DER`, `CEL`).
+  - **No Exceptions**: Do **NOT** add new exceptions to `tests/apichecks/testdata/exceptions/acronyms.txt`. Fix the Go field names instead.
+- **Pointers**: Validate that optional or nullable fields (like `Location`, booleans, integers) are correctly typed as pointers (e.g., `*string`, `*bool`) rather than raw values.
+- **ETag Exclusion**: Remove `etag` fields from the KRM spec; they are generally disallowed. Do **NOT** add exceptions to `spec_dislike_etag.txt`.
```
ldanielmadariaga, line 49: Sorry to pile on here but we just don't support etags? Is this based on some team decision made in the past? In principle I get with supporting etags is messy in KRM mostly trying to get context.

.gemini/skills/kcc-direct-greenfield-types-implementer/SKILL.md
```diff
@@ -41,6 +41,24 @@ Apply the baseline validations from `kcc-direct-base-types-implementer`, plus th
...
+- **KCC References (`missingrefs.txt`)**:
+  - Ensure fields representing GCP resources are implemented using the KCC Ref pattern (e.g., `ProjectRef`, `NetworkRef`) rather than plain strings.
+  - **No Exceptions**: Do **NOT** add exceptions to `tests/apichecks/testdata/exceptions/missingrefs.txt`. Fix the underlying field types instead.
+  - Reuse existing refs from `apis/refs/v1beta1/` (e.g., `Compute`, `BigQueryDataset`). If creating a new ref, put it in a service-specific file (e.g., `apis/dlp/v1alpha1/dlprefs.go`).
```
ldanielmadariaga: I've seen inconsistent use and recommendations here. Most of the resources are doing the contrary to this :)

Why are we settling on this pattern now? I also see that some of the refs in api/refs don't really follow the new standards. I think these mostly reside in apis/refs rather than apis/refs/v1beta1 we should remove these to avoid confusion.

ldanielmadariaga (changes requested):
Let's add some comments about copyright years as well :) I keep having to reject PRs because the agent will update copyright years for existing files

.gemini/skills/kcc-direct-greenfield-types-implementer/SKILL.md
acpana: > Why are we settling on this pattern now? 

IIUC, we are marching towards having every ref in its api package, i.e. `apis/dlp/v1alpha1/resource_reference.go`; and we wanted to move away from `apis/refs/v1beta1` as they continued old style refs but @gemmahou has more context here.

ldanielmadariaga: That's what I understood from talking to you. I fear that 'Reuse existing refs from `apis/refs/v1beta1/`' will lead the bot to prefer that folder rather than api packages may be we should scope it down to just very speicific refs like: project, org, folder? 

Anyways I'd reword the text a bit to block possible generalizations assumptions like:  

When creating a new ref you **must** put it in a service-specific file (e.g., `apis/dlp/v1alpha1/dlp_reference.go`). You can reuse some already existing refs from `apis/refs/v1beta1/` (e.g., `Compute`, `BigQueryDataset`). You must only reuse refs from the `apis/refs/v1beta1/` you must not use refs under `apis/refs/` as these use a deprecated style that is no longer correct. You must not create new reference files under `apis/refs` or any subfolder within.

PR #9696, opened 2026-06-11
compute: fix missing structured diff for ForwardingRule in Preview Mode

### LINKED ISSUE #9695: [ComputeForwardingRule] Structured diff is missing in Preview Mode
### Description
The `ComputeForwardingRule` direct controller does not report structured diffs in Preview Mode (or when dry-running).

### Root Cause
In `forwardingrule_controller.go`, the `structuredreporting.ReportDiff(ctx, report)` call is placed at the very end of the `Update` method, after all the GCP write requests (like `Patch`, `setLabels`, `SetTarget`) have been made.
In Preview Mode, these mutating requests are intercepted and blocked by the `interceptingGCPClient`, which returns a `BlockedGCPError`. The controller catches this error and immediately returns early, preventing the execution from reaching `ReportDiff`. Thus, no diff is ever reported for the resource.

### Solution
Use `defer structuredreporting.ReportDiff(ctx, report)` right after `report` is initialized at the beginning of the `Update` method. This guarantees that the diff is always reported when the function returns, even on early exits due to blocked requests.

anhdle-sso:
Fixes #9695

### DIFF
--- a/pkg/controller/direct/compute/forwardingrule_controller.go
+++ b/pkg/controller/direct/compute/forwardingrule_controller.go
@@ -260,6 +260,8 @@ func (a *forwardingRuleAdapter) Update(ctx context.Context, updateOp *directbase
 	forwardingRule.Labels = sanitizedLabels
 
 	report := &structuredreporting.Diff{Object: updateOp.GetUnstructured()}
+	// Use defer to ensure the diff is always reported, even if we return early due to errors (crucial for preview mode)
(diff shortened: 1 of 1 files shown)


anfernee:
/lgtm 

https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/9695 is a small change, but this includes a bit refactoring too.

anfernee (approved):

PR #9731, opened 2026-06-12
Implement IP filter for Storage Buckets TF resource

### LINKED ISSUE #6208: feature: support ipFilter in StorageBucket
We need to support the `ipFilter` field in the `StorageBucket` resource.

This feature allows users to restrict access to Cloud Storage buckets based on IP address ranges, which is a critical security requirement for many organizations.

To do that, I think we need to make the following changes:

1.  **We should** add the `ipFilter` field to the `StorageBucket` CRD (likely requiring migration to the KRM-native reconciler if not already supported there).
2.  **We should** map this to the `ipFilter` field in the Storage JSON API (note: check API maturity, as it might be in preview).

ldanielmadariaga:
### BRIEF Change description
Implement IP filter for Storage Buckets TF resource

Fixes #6208

Customer request

### Tests you have done
Local run against real GCP


anfernee (dismissed):
/lgtm

It's brownfield. what test do we normally do before merging this.

maqiuyujoyce:
Could you record the real GCP logs in one commit and then ensure the mockgcp logs match the real ones (for the newly added field)?

tests/apichecks/testdata/exceptions/acronyms.txt
```diff
@@ -743,6 +743,11 @@
...
 [acronyms] crd=sqlsslcerts.sql.cnrm.cloud.google.com version=v1beta1: field ".status.serverCaCert" should be ".status.serverCACert"
+[acronyms] crd=storagebuckets.storage.cnrm.cloud.google.com version=v1beta1: field ".spec.ipFilter.allowCrossOrgVpcs" should be ".spec.ipFilter.allowCrossOrgVPCs"
+[acronyms] crd=storagebuckets.storage.cnrm.cloud.google.com version=v1beta1: field ".spec.ipFilter.publicNetworkSource.allowedIpCidrRanges" should be ".spec.ipFilter.publicNetworkSource.allowedIPCIDRRanges"
+[acronyms] crd=storagebuckets.storage.cnrm.cloud.google.com version=v1beta1: field ".spec.ipFilter.publicNetworkSource.allowedIpCidrRanges[]" should be ".spec.ipFilter.publicNetworkSource.allowedIPCIDRRanges[]"
+[acronyms] crd=storagebuckets.storage.cnrm.cloud.google.com version=v1beta1: field ".spec.ipFilter.vpcNetworkSources[].allowedIpCidrRanges" should be ".spec.ipFilter.vpcNetworkSources[].allowedIPCIDRRanges"
+[acronyms] crd=storagebuckets.storage.cnrm.cloud.google.com version=v1beta1: field ".spec.ipFilter.vpcNetworkSources[].allowedIpCidrRanges[]" should be ".spec.ipFilter.vpcNetworkSources[].allowedIPCIDRRanges[]"
```
acpana, line 750: up to you if you'd like to fix those but ... I think it's low value at this point. One way to fix them is this I think:

https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/2178

third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/storage/resource_storage_bucket.go
```diff
@@ -458,6 +458,94 @@ func ResourceStorageBucket() *schema.Resource {
 					},
 				},
 			},
+			"ip_filter": {
```
acpana, line 461: Another way to fix the capitalization, IIRC, would be to name this field as such `i_p_filter` as every char after `_` gets capitalized (or used to at least)

pkg/test/resourcefixture/testdata/basic/storage/v1beta1/storagebucket/storagebucketipfilter/create.yaml
```diff
@@ -0,0 +1,42 @@
...
+    mode: Enabled
+    publicNetworkSource:
+      allowedIpCidrRanges:
+        - 8.8.8.8/32
+    vpcNetworkSources:
+      - networkRef:
+          name: computenetwork-${uniqueId}
+        allowedIpCidrRanges:
+          - 10.0.0.0/8
+    allowCrossOrgVpcs: false
```
acpana, line 41: I'd update these fields too but this can be done as a follow up

acpana (approved):
LGTM a few nits/ questions but otherwise 👍🏼 

let's also add a commit for real log, followed by mock log

* real log
* mock log

so we can see if there are any meaningful diffs (like server defaulting, etc.) as those may miss in our mocks

PR #10132, opened 2026-06-13
Add DefaultCustomErrorResponsePolicy to ComputeURLMap

### LINKED ISSUE #6206: feature: support defaultCustomErrorResponsePolicy in ComputeURLMap
We need to support the `defaultCustomErrorResponsePolicy` field in the `ComputeURLMap` resource.

This feature allows users to configure custom error response pages for their load balancers directly through Config Connector.

To do that, I think we need to make the following changes:

1.  **We should** add the `defaultCustomErrorResponsePolicy` field (and its associated rule fields) to the `ComputeURLMap` CRD.
2.  **We should** ensure the field is correctly mapped to the underlying Compute Engine API field `defaultCustomErrorResponsePolicy`.

ldanielmadariaga:
Tested with real and mock GCP


gemmahou:
/lgtm

CRD looks good to me, and the minor differences between real and mock are acceptable.

gemmahou (approved):

PR #10378, opened 2026-06-16
feat: Add ipCollection field to ComputeAddress

### LINKED ISSUE #6523: Add the `ipCollection` field from the `google.compute.v1.Address` to the `ComputeAddress` resource

ldanielmadariaga:
Add the `ipCollection` field to the ComputeAddress resource (in v1beta1 API) to allow reserving an external IPv4 address using BYOIP (Bring Your Own IP) from an IP Collection (e.g. PublicDelegatedPrefix).

This includes:
- Adding `IpCollection` spec field in KRM Go types.
- Patching the vendored copy of the Terraform Google Beta provider to support the `ip_collection` schema field, diff suppression, expander, and flattener.
- Regenerating CRD manifests, Go clients, and reference docs.
- Creating a resource fixture test case `computeaddressipcollection` and recording its MockGCP golden logs.

WARNING: PDPs can't be tested in realGCP as they need a real IP to be used by BYOP


apis/compute/v1beta1/address_types.go
```diff
@@ -51,6 +51,11 @@ type ComputeAddressSpec struct {
 	// +optional
 	IPV6EndpointType *string `json:"ipv6EndpointType,omitempty"`
 
+	/* Immutable. Reference to the source of external IPv4 addresses,
+	   like a PublicDelegatedPrefix(PDP) for BYOIP. */
+	// +optional
+	IpCollection *string `json:"ipCollection,omitempty"`
```
ldanielmadariaga: This should technically be a ref but we haven't implemented yet. For now the safest bet seems to just use a string a migrate to ref once we implement that. Open to opinions on this.

gemmahou: This a beta resource and future CRD changes are not permitted, changing string field to a reference field would break existing customer. 

We should implement the reference with minimal code and comment out both 'name' and 'namespace' fields, as only the 'external' value is accepted at this time, so we don't worry about the reference normalization etc. Example: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/610e951797ae2ce3f51f73e074d32cf5df4ae7d0/apis/dataplex/v1alpha1/aspecttype_reference.go#L32

However, since this is a TF-based resource, we define references in service mappings. I am afraid we will need to implement the CRD for `PublicDelegatedPrefix` before we can proceed...

ldanielmadariaga: QQ would we need the complete CRD for PDP or just the ref? The first step in the brownfield workflow is generating/updating how we implement references we may be able to piggyback off that?

I'm inclined in having a ref even if it doesn't work super well as we could just use the external field in the ref as needed.

gemmahou: If we are planning to implement this field in TF, then we'll need the CRD of the reference resource. We can make a comment in the servicemappings saying only `external` is allowed, for example: https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/96791593451f5aac98c5c4954129624ee16e626f/config/servicemappings/certificatemanager.yaml#L66, but we'll need the GVK of the reference resource in TF's servicemappings.

acpana (dismissed):
make sure to run against real too 🙏🏼 to make sure there are no surprises!

gemmahou (commented):
/lgtm

gemmahou (approved):

PR #10990, opened 2026-06-30
Split HTTP logs into _http.log and _http_mock.log, add alignment check

ldanielmadariaga:
## Description
This PR separates the HTTP/gRPC golden logs between real GCP and mock GCP E2E runs, preventing mock runs from overwriting real GCP traffic records. It also introduces a fast, semantic alignment check.

### Key Changes
1. **Split Log Files**: 
   * Mock GCP runs (`E2E_GCP_TARGET=mock`) now write to and verify against `_http_mock.log`.
   * Real GCP runs continue to use `_http.log`.
   * **Fallback**: If `_http_mock.log` is missing, the test falls back to `_http.log`, ensuring 100% backward compatibility for existing tests.
2. **Alignment Test (`TestGoldenLogAlignment`)**:
   * Added a new test in `pkg/test/resourcefixture` that compares `_http.log` and `_http_mock.log` when both exist.
   * To prevent flakiness from asynchronous timing or polling, it normalizes the logs by sorting the events and collapsing duplicate `GET` requests.
   * Since it is a file-only check, it runs in **<100ms** and is automatically included in the unit-test presubmits.
  
### Benefits of Splitting E2E Golden Logs & Adding Alignment Check
1. **Agentic Workflow Validation**: Since the real GCP traffic (`_http.log`) is no longer overwritten by mock runs, humans can easily inspect the git diff to verify that an agent actually ran its validation against the real GCP API rather than just using the mock.
2. **Prevents Log Overwriting**: Splitting logs into `_http.log` (real) and `_http_mock.log` (mock) allows developers and CI to run mock E2E tests without losing the recorded real GCP traffic.
3. **Automated Drift Detection**: The new `TestGoldenLogAlignment` test automatically compares the two logs to ensure that `mockgcp`'s behavior does not drift from the real GCP API.


pkg/test/resourcefixture/testdata/basic/kms/v1beta1/kmskeyring/kmskeyring/_final_object.diff
acpana, line 1: for my own learning, when did we start keeping a "_final_*" golden? Or what's the purpose?

pkg/test/resourcefixture/testdata/basic/kms/v1beta1/kmskeyring/kmskeyring/_http_mock.log
acpana, line 1: purely for my own UX, it would be nice if the other log file would be called `_http_real.log`. Either way, I would've expected to see changes to the current `_http.log` if we're keeping those as the `real` logs.

pkg/test/resourcefixture/golden_alignment_test.go
```diff
@@ -0,0 +1,267 @@
...
+		t.Fatalf("failed to get absolute path for %s: %v", rootDir, err)
+	}
+
+	err = filepath.WalkDir(absRootDir, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if d.IsDir() {
+			realLogPath := filepath.Join(path, "_http.log")
```
acpana, line 43: Yup just underscoring that unless we run and save the http requests against real,  existing http log files would have the mock run.

For the many disparities that would show a diff between real and mock, we probably want to have an exception list to "stop the bleeding" for now and chip at the diffs in separate PRs.

acpana (commented):
some questions/ thoughts

ldanielmadariaga:
Superseded by https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/11308

PR #10994, opened 2026-06-30
E2E Fail-Fast: Add Pre-flight API Checks and Terminal Error Polling

ldanielmadariaga:
### Problem
When prerequisite steps of an E2E test fail (such as a required GCP API not being enabled in the test project), the test framework continues to poll the resource status for the full `35-minute` timeout before finally failing. This results in extremely slow feedback loops for basic environment/project configuration issues.

### Solution
This PR introduces two mechanisms to fail-fast on configuration and API enablement issues:
1. **Pre-flight API Verification (`preflightCheckAPIs`):**
   Before deploying resources on the test cluster, the test runner now inspects the resources under test for any `Service` resources (which enable GCP APIs). Using the GCP `serviceusage` API client, it verifies if the Service Usage API is accessible and queries the status of those services. If the project's Service Usage API is disabled or credentials lack permissions, the test fails immediately (within seconds) before starting the test infrastructure.
2. **Fail-Fast Polling (`isTerminalError`):**
   During the wait/polling loop for resources to become ready, we now inspect the `Ready` condition. If the condition status is `False` due to a terminal error (e.g., `"Access Not Configured"`, `"API is not enabled"`, `"has not been used in project"`, etc.), we abort the polling immediately rather than waiting for the 35-minute timeout.

### Benefits
* **Faster Feedback Loops:** Reduces E2E test failure times from **35 minutes to under 5 seconds** for API enablement or permission-related misconfigurations.
* **Clearer Failures:** Fails immediately with the exact GCP API error message, making it easier to diagnose project setup issues.

### DIFF
--- a/config/tests/samples/create/samples.go
+++ b/config/tests/samples/create/samples.go
@@ -38,6 +38,8 @@ import (
 	"github.com/GoogleCloudPlatform/k8s-config-connector/pkg/util/repo"
 
 	"github.com/ghodss/yaml" //nolint:depguard
+	"google.golang.org/api/option"
+	serviceusage "google.golang.org/api/serviceusage/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -127,6 +129,12 @@ type CreateDeleteTestOptions struct { //nolint:revive
 
 func RunCreateDeleteTest(t *Harness, opt CreateDeleteTestOptions) {
 	ctx := t.Ctx
+
+	if t.GCPTarget == GCPTargetModeReal {
+		if err := preflightCheckAPIs(ctx, t, opt.Create); err != nil {
+			t.Fatalf("FAIL: pre-flight check failed: %v", err)
+		}
+	}
 
 	if t.Manager != nil {
 		var gvks []schema.GroupVersionKind
@@ -272,6 +280,11 @@ func waitForReadySingleResource(t *Harness, u *unstructured.Unstructured, timeou
 				logger.Info("resource is ready", "kind", u.GetKind(), "name", u.GetName())
 				return true, nil
 			}
+			if c.Type == "Ready" && c.Status == "False" {
+				if terminal, category := isTerminalError(c.Reason, c.Message); terminal {
+					return false, fmt.Errorf("terminal error (%s): reason: %s, message: %s", category, c.Reason, c.Message)
+				}
+			}
 		}
 		// This resource is not completely ready. Let's keep polling.
 		logger.Info("resource is not ready", "kind", u.GetKind(), "name", u.GetName(),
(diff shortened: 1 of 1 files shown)


config/tests/samples/create/samples.go
```diff
@@ -294,6 +307,57 @@ func waitForReadySingleResource(t *Harness, u *unstructured.Unstructured, timeou
...
+		return true, "Unauthorized (401)"
+	case "403":
+		return true, "Forbidden (403)"
+	case "501":
+		return true, "Not Implemented (501)"
+	default:
+		return false, ""
+	}
+}
+
```
acpana: nit: I think there some utility out there better for this

config/tests/samples/create/samples.go
```diff
@@ -294,6 +307,57 @@ func waitForReadySingleResource(t *Harness, u *unstructured.Unstructured, timeou
 	t.Errorf("%v, final status: %+v", baseMsg, objectStatus)
 }
 
+func preflightCheckAPIs(ctx context.Context, h *Harness, resources []*unstructured.Unstructured) error {
+	var services []string
+	for _, u := range resources {
+		if u.GroupVersionKind().Group == "serviceusage.cnrm.cloud.google.com" && u.GroupVersionKind().Kind == "Service" {
+			services = append(services, u.GetName())
+		}
+	}
```
acpana, line 317: so we'd be starting off a new pattern where we define the services that need to be enable for every fixture test, right? 

otherwise, I don't think we generally define a Service kind alongside the fixture tests today and we'd have to add it. 


Nit: Further, I'm thinking we only need to enable the service once per test project, do we want to pay the cost of running this Get operation every time for our many, many ever expanding fixture test list?

acpana (commented):
two nits, one real question about changing the fixture tests pattern; LGTM otherwise!

config/tests/samples/create/samples.go
ldanielmadariaga: This is read from the test manifests so although not the most thing effective ever... particularly dependencies.yaml usually denotes the uncommon dependencies that a test would have. 

Although not all encompassing it seems like a decent trade-off where we check for more exotic APIs that we're not using all the time (contrasting with compute/container) and that our local test projects are more likely to have never used. 

There are ways to achieve better coverage here but there are more heavyweight and you note will issue more requests to service usage for well known types. That could benefit agent runs but probably be overkill for human runs. 

I think I'd keep the scheme for now (I've added a comment to the code) and we can revisit based on how effective we notice this is. FWIW I don't think the get requests to service usage will be too problematic it's a fairly lightweight request in a previous world I think we'd be fine today we may get quota'ed.

config/tests/samples/create/samples.go
ldanielmadariaga: Probably not this a bit custom to how we wanna handle it particularly in K8S land, and there are some cases that we are skipping like unimplemented which should be terminal but would actually fail our tests.

acpana (approved):

PR #11081, opened 2026-06-30
Split logs mock logs

ldanielmadariaga:
Add new _http_mock.log and _http_old_controller_mock.logs for existing E2E fixture tests.
This is most of the list but we have some failing resources which I'll tackle separately.


acpana (approved):
additive changes working towards no diff b/w current `_http.log`s and new `_http_mock.log`s

PR #11084, opened 2026-06-30
Implement GKE Parallelstore CSI Driver Addon support for ContainerCluster

### LINKED ISSUE #6861: [Feature] Support Dynamic Workload Scheduler (DWS) and Parallelstore CSI in GKE Clusters/NodePools
**Description:**
Config Connector `v1beta1` schemas for `ContainerCluster` and `ContainerNodePool` are currently missing two critical fields required for provisioning modern AI Infrastructure architectures on GKE (such as A3 Ultra clusters).

1. **`ContainerNodePool.spec.queuedProvisioning`**: This field is mandatory for provisioning A3/A4 instances utilizing Dynamic Workload Scheduler (DWS). Without it, users cannot provision queued GPU capacity.
2. **`ContainerCluster.spec.addonsConfig.parallelstoreCsiDriverConfig`**: This field is required to mount the high-performance fully managed Parallelstore file system to the cluster.

**Steps to Reproduce:**
Attempt to apply a `ContainerNodePool` containing:
```yaml
  queuedProvisioning:
    enabled: true
```
Or a `ContainerCluster` with:
```yaml
  addonsConfig:
    parallelstoreCsiDriverConfig:
      enabled: true
```

**Result:**
Strict schema validation immediately rejects the apply:
```
.spec.queuedProvisioning: field not declared in schema
.spec.addonsConfig.parallelstoreCsiDriverConfig: field not declared in schema
```

**Proposed Workaround:**
Users must manually strip these fields to deploy via GitOps, resorting to imperative commands (`gcloud`) or raw Terraform to supplement the provisioning process until these API fields are ingested into KCC.

ldanielmadariaga:
# Description
Implement GKE Parallelstore CSI Driver Addon support for `ContainerCluster`.
This adds the schema mapping and validation logic for enabling the parallelstore CSI driver on container clusters.
### Details of Changes:
1. **API Schema & Type Mapping:**
   - **`ContainerCluster`:** Added `parallelstoreCsiDriverConfig` (with `enabled` boolean) mapped from `google_container_cluster.addons_config.parallelstore_csi_driver_config`.
2. **E2E Tests:**
   - **`containercluster-addon`:** Added `parallelstoreCsiDriverConfig` to basic addon test manifests, recorded traffic against real GCP, and aligned golden logs with MockGCP.


third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go
```diff
@@ -422,6 +423,23 @@ func ResourceContainerCluster() *schema.Resource {
 								},
 							},
 						},
+						"parallelstore_csi_driver_config": {
```
ldanielmadariaga, line 426: TF provider ref: https://github.com/hashicorp/terraform-provider-google/blob/main/google/services/container/resource_container_cluster.go#L497

pkg/test/resourcefixture/testdata/basic/container/v1beta1/containercluster/containercluster-addon/_http.log
acpana, line 1: could you please add two commits:

- one that shows the real log run
- one that shows the mock log  run

?

pkg/test/resourcefixture/testdata/basic/container/v1beta1/containercluster/containercluster-addon/_http.log
```diff
@@ -258,7 +261,8 @@ X-Xss-Protection: 0
     },
     "networkPolicyConfig": {
       "disabled": true
-    }
+    },
+    "parallelstoreCsiDriverConfig": {}
```
gemmahou, line 265: A bit concerned here: When parallelstoreCsiDriverConfig.enabled is false, the GET response returns empty "{}", which might trigger an unintended diff between desired and actual state. 

The "{}" values comes from mockGCP and may not be accurate. Can we verify the actual GCP log behavior to see what the live API returns when the field is disabled?

pkg/test/resourcefixture/testdata/basic/container/v1beta1/containercluster/containercluster-addon/_http.log
ldanielmadariaga: Done

pkg/test/resourcefixture/testdata/basic/container/v1beta1/containercluster/containercluster-addon/_http.log
gemmahou: I think this extra PUT call confirms that there's unintended diff between desired and actual state: https://github.com/ldanielmadariaga/k8s-config-connector/blob/b9cb90e11dd9aaf78eaa622765487165dcf9a74b/pkg/test/resourcefixture/testdata/basic/container/v1beta1/containercluster/containercluster-addon/_http.log#L1551-L1596

desired:
```
"parallelstoreCsiDriverConfig": {
    "enabled": false
}
```

actual:
```
"parallelstoreCsiDriverConfig": {}
```
But I also notice that this is an existing problem, not this new field that causes the diff, other configs, like `kalmConfig` when it's set to "false" in the configuration yaml, the returned values from API is also `{}`. 

So this is not a blocker of this PR at this moment. I think we need to handle all those configs to avoid the diff, maybe in https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/krmtotf/krmtotf.go#L64. I'm also curious why we don't get ticket on this continuous update...?

gemmahou (approved):

PR #11156, opened 2026-07-02
Fix unstable CRD schema for aiplatformmodels due to recursion cycle

ldanielmadariaga:
Map google.protobuf.ListValue to KRM JSON and fix dereference generation in mapper


ldanielmadariaga:
Potentially addressed by https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/11163

PR #11186, opened 2026-07-02
Implement GCP DNS Authorization Per Project Record (#2943)

### LINKED ISSUE #2943: GCP DNS Authorization Per Project Record
### Describe your question

The gcloud command for creating a dns authorization now supports using a `--type` flag and using a PER_PROJECT_RECORD option. This allows for creation the same google managed certificate in mutiple projects and having a unique DNS authorization cname so google can authorize and issue the certificate for each project. In the past this was not possible because the same acme challenge cname was created with different a different `data` value. I don't see this as a field in the CertificateManagerDNSAuthorizationSpec type yet. What information would be needed to expose this so it can be configured when creating a CertificateManagerDNSAuthorization resource? 

Citing documentation: https://cloud.google.com/certificate-manager/docs/dns-authorizations

```
The default DNS authorization type for a global Google-managed certificate is FIXED_RECORD. To use per-project DNS authorization, run the following command:

gcloud certificate-manager dns-authorizations create AUTHORIZATION_NAME \
    --domain="DOMAIN_NAME" \
    --type="PER_PROJECT_RECORD"
```

ldanielmadariaga:
This pull request implements the PER_PROJECT_RECORD type support for CertificateManagerDNSAuthorization (Issue #2943).


pkg/controller/direct/certificatemanager/certificatemanagerdnsauthorization_controller.go
```diff
@@ -285,9 +285,17 @@ func compareCertificateManagerDNSAuthorization(ctx context.Context, actual, desi
...
 	diffs, updateMask, err := tags.DiffForTopLevelFields(ctx, desired.ProtoReflect(), maskedActual.ProtoReflect())
 	if err != nil {
 		return nil, nil, err
 	}
+	var allowedPaths []string
+	for _, path := range updateMask.Paths {
+		if path == "description" || path == "labels" {
+			allowedPaths = append(allowedPaths, path)
+		}
+	}
```
maqiuyujoyce: Looks like this logic will hide changes made to the `type` field?

Also if the `type` field is unset in the desired resource then its proto value will be `DnsAuthorization_TYPE_UNSPECIFIED`, but if the actual resource is using its default value, `DnsAuthorization_FIXED_RECORD`, an unexpected diff will be generated. And we should avoid such cases.

tests/e2e/httplog.go
maqiuyujoyce, line 1: Could we move those normalization to mockcertificatemanager?

pkg/controller/direct/certificatemanager/certificatemanagerdnsauthorization_controller.go
```diff
@@ -285,9 +285,17 @@ func compareCertificateManagerDNSAuthorization(ctx context.Context, actual, desi
 		return nil, nil, err
 	}
 	maskedActual.Name = desired.Name
+	maskedActual.Labels = actual.Labels
 	diffs, updateMask, err := tags.DiffForTopLevelFields(ctx, desired.ProtoReflect(), maskedActual.ProtoReflect())
 	if err != nil {
 		return nil, nil, err
 	}
+	var allowedPaths []string
```
maqiuyujoyce: In the long term, we probably need to figure out a strategy to identify silently dropped changes. Not a blocker for this PR.

pkg/controller/direct/certificatemanager/certificatemanagerdnsauthorization_controller.go
ldanielmadariaga: Updated defaulting, but the type field is immutable so we're incorrectly ignoring any changes during update.

pkg/controller/direct/certificatemanager/certificatemanagerdnsauthorization_controller.go
ldanielmadariaga: I don't think this was incorrectly dropped in this case. The field is immutable it seems most fields in this resource are.

tests/e2e/httplog.go
ldanielmadariaga: Done

pkg/controller/direct/certificatemanager/certificatemanagerdnsauthorization_controller.go
maqiuyujoyce: Actually, we do need every single field to be compared properly in the Update() function even if the field is immutable.

A common edge case would be acquisition. For example, the underlying GCP resource has type A configured, but the KRM object has type B configured. Because this is an acquisition, it's creation for the KRM object (so immutability check is not triggered), but it goes through Update() to reconcile (because the actual resource exists). Then the difference of the `type` value will be hidden if we don't do an explicit comparison here. KCC should return an error clarifying that the user has configured the wrong type when they tried to acquire the resource.

ldanielmadariaga: By acquisition d'you mean adopting/importing an existing resource to KCC? I don't think we manage this very elegantly in general but sure I can change it. I've other controllers that drop immutable fields though and it seem common though none come to mind. 

Didn't even know we supported this and not sure we should anyways, at least not as anything better than a best effort capacity.

ldanielmadariaga: Reverted the behavior now all fields will be sent to update

pkg/controller/direct/certificatemanager/certificatemanagerdnsauthorization_controller.go
maqiuyujoyce, line 198: I think we shouldn't configure fields manually when creating a req (but instead use a parsed desired resource directly.

This may also hide the user error (if the immutability check is not handled via checking updateMask).

pkg/controller/direct/certificatemanager/certificatemanagerdnsauthorization_controller.go
```diff
@@ -285,9 +285,14 @@ func compareCertificateManagerDNSAuthorization(ctx context.Context, actual, desi
 		return nil, nil, err
 	}
 	maskedActual.Name = desired.Name
+	maskedActual.Labels = actual.Labels
+	if desired.Type == certificatemanagerpb.DnsAuthorization_TYPE_UNSPECIFIED {
+		desired.Type = certificatemanagerpb.DnsAuthorization_FIXED_RECORD
+	}
```
maqiuyujoyce, line 291: Not a blocker.

I wondered if we should handle default in a centralized place when we create the adapter. E.g. https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/sql/sqlinstance_defaults.go (not the best example though).

maqiuyujoyce (commented):
> By acquisition d'you mean adopting/importing an existing resource to KCC?

Yes. Please find the details at: https://docs.cloud.google.com/config-connector/docs/how-to/managing-deleting-resources#acquiring_an_existing_resource

> I don't think we manage this very elegantly in general but sure I can change it. I've other controllers that drop immutable fields though and it seem common though none come to mind.

It was handled properly in the legacy controllers but we don't have a consistent way to handle it in the direct controllers. This is a known problem that we should fix/improve in direct controllers.

>  Didn't even know we supported this and not sure we should anyways, at least not as anything better than a best effort capacity.

We support it since the first public version of KCC and this is part of how users like KCC. It's quite flexible as users can transition from other tools to KCC risk free (they don't need to delete any GCP resources).

maqiuyujoyce (approved):
/lgtm

I'll let you decide how you plan to follow up on the comments.

PR #11198, opened 2026-07-02
Add certificateManagerCertificates support to ComputeTargetHTTPSProxy regional resource

### LINKED ISSUE #4999: Unable to reference regional certificate manager certificates to a regional HTTPS proxy
### Checklist

Hey Guys, 

When trying to create a ComputeTargetHTTPSProxy, I am unable to attach a regional certificate manager certficate ie when certificateManagerCertificates.external property is set to "projects/project-id/locations/europe-west1/certificates/europe-west1-local-srinath-2-co-gcp-dhplatform-cloud", i get error that the certificate are not a global one. 

This is not the case when deploying through terraform as it works correctly,  

resource "google_compute_region_target_https_proxy" "alb_target_ssl_proxy_https" {
  certificate_manager_certificates = ["projects/project-id/locations/europe-west1/certificates/europe-west1-local-srinath-2-co-gcp-dhplatform-cloud"]
}

Upon checking the CC is stil using a old version of the gcp provider, which doesnt take into the new apis. 
https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/058b2061ed16c7b76ce196e2043e4d94aba66838/third_party/github.com/hashicorp/terraform-provider-google-beta/google-beta/services/compute/resource_compute_region_target_https_proxy.go#L149C2-L149C21

The terraform provider handles the certificateManagerCertificates differently 
https://github.com/hashicorp/terraform-provider-google/blob/37babef82028b61873a7c9bf3669b05722354c18/google/services/compute/resource_compute_region_target_https_proxy.go#L200C2-L200C36

### Additional information

_No response_

### Importance

_No response_

ldanielmadariaga:
This PR adds certificateManagerCertificates support to ComputeTargetHTTPSProxy regional resource (Issue #4999).


pkg/resourceskeleton/resourceskeleton.go
```diff
@@ -237,3 +238,16 @@ func trimRegionPrefix(host string) string {
 	parts := strings.Split(host, "-")
 	return parts[len(parts)-1]
 }
+
+func trimURLPathPrefix(path string) string {
+	cleanedPath := strings.TrimPrefix(path, "/")
+	if strings.HasPrefix(cleanedPath, "batch/compute/") {
```
gemmahou, line 247: I'm not super familiar with the path contains "batch/compute", can we double-check if that's intended, and add a short comment to this function(before URL, cleaned URL) for future reference?🙏

pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computetargethttpsproxy/regionaltargethttpsproxy-direct/_http.diff
gemmahou, line 1: Not a blocker. Please see this comment in my PR https://github.com/GoogleCloudPlatform/k8s-config-connector/pull/11681#discussion_r3599650869, I think it's unnecessary for us to keep "-direct" tests. Depends on you to clean it up in this PR or not.

gemmahou, line 1: I noticed we already have direct controller for ComputeTargetHTTPSProxy resource([code](https://github.com/GoogleCloudPlatform/k8s-config-connector/blob/master/pkg/controller/direct/compute/computetargethttpsproxy_controller.go)), can we add the new field in direct controller instead of doing TF patch?

ldanielmadariaga: Started this PR before there was a direct resource. WIll add as a follow-up but we should support in both.

ldanielmadariaga: Yeah I've cleaned it up in some. Let me follow-up in a different PR cleaning up the whole repo these are not needed. Removing the changes from the PR doesn't move the needle though since the files will remain there so that's why I'd rather keep do a real cleanup in a follow-up.

pkg/resourceskeleton/resourceskeleton.go
ldanielmadariaga: Function should have an explanatory comment now. 
TL/DR is: the compute batch API has a non-standard URL path instead of the regular compute.googleapis.com/projects it uses googleapis.com/compute/batch/project. This function normalizes the batch resource that to a standard resource name.

gemmahou (approved):

PR #11201, opened 2026-07-02
Add support for resourceLabels in ContainerCluster CRD

### LINKED ISSUE #5186: Missing resourceLabels support in ContainerCluster CRD
### Checklist

## What happened?

When trying to label a GKE cluster via the `ContainerCluster` resource in Config Connector, the field resourceLabels is not available.
In the generated Go client (`container.cnrm.cloud.google.com/v1beta1`) and in the CRD, there is no field for cluster-level labels.

 ## What did you expect to happen?

The `ContainerCluster` spec should expose the resourceLabels field (as supported by the [GKE API](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.locations.clusters#Cluster)) so that cluster-level labels can be declared and managed declaratively via Config Connector.

### Additional information

- Labels at the node level are supported via `spec.nodeConfig.labels`, but cluster-level resource labels are missing.
- This limits our ability to enforce consistent cost-allocation and governance tags across GKE clusters declaratively.

### Importance

_No response_

ldanielmadariaga:
Closes #5186

### DIFF
--- a/apis/container/v1beta1/containercluster_types.go
+++ b/apis/container/v1beta1/containercluster_types.go
@@ -1509,6 +1509,10 @@ type ContainerClusterSpec struct {
 	// +kcc:proto:field=google.container.v1.Cluster.locations
 	NodeLocations []string `json:"nodeLocations,omitempty"`
 
+	/* The GCE resource labels (a map of key/value pairs) to be applied to the cluster. */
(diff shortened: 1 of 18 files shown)


ldanielmadariaga:
This'll be a bit unnecessary and confusing given we already map to `metadata.labels`. So keeping the behavior as is for now.

PR #11292, opened 2026-07-03
Migrate ComputeNetworkEndpointGroup to Direct Reconciler

ldanielmadariaga:
Migrate zonal ComputeNetworkEndpointGroup to use a direct reconciler and implement mockgcp support.


ldanielmadariaga:
Not interested in implement direct controller yet

PR #11321, opened 2026-07-04
Support disableL4LbFirewallReconciliation in GKE ContainerCluster

### LINKED ISSUE #7605: ContainerCluster: Support for disableL4LbFirewallReconciliation
### Checklist

The `ContainerCluster` resource does not currently support the `networkConfig.disableL4LbFirewallReconciliation` field. This field is available in the [GKE API](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.locations.clusters#NetworkConfig) and the Terraform `google_container_cluster` resource.

When set to `true`, GKE stops automatically creating and managing VPC firewall rules for L4 (TCP/UDP) load balancer Services. This is important in shared VPC environments where firewall rules are managed centrally in the host project — typically via network firewall policies or resource manager tag-based rules — and the GKE service account in the service project does not have permission to create firewall rules in the host project.

Without this field, KCC-managed clusters default to `false` (reconciliation enabled), causing GKE to attempt firewall rule creation in the host project on every L4 Service. This doesn't block Service creation, but generates persistent permission errors on Service status and conflicts with centrally managed firewall policies.

### Additional information

GKE API field: [`cluster.networkConfig.disableL4LbFirewallReconciliation`](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.locations.clusters#NetworkConfig)

Terraform equivalent (top-level attribute on `google_container_cluster`):

```hcl
resource "google_container_cluster" "example" {
  # ...
  disable_l4_lb_firewall_reconciliation = true
}
```

gcloud equivalent:

```
gcloud container clusters create CLUSTER_NAME \
  --disable-l4-lb-firewall-reconciliation
```

### Importance

Not vital. The service can be created anyway it will just create uneccessary firewall rules in air-gapped environments which can be a security risk.

ldanielmadariaga:
Add support for disableL4LbFirewallReconciliation on GKE ContainerCluster resources.

Tested with MockGCP E2E test fixtures validating creation, update, and deletion behavior.


gemmahou (dismissed):
LGTM, considering all tests pass and this is a simple boolean field.

gemmahou (approved):

PR #11469, opened 2026-07-07
compute: suppress diff for delete_default_routes_on_create in TF controller

### LINKED ISSUE #6195: [Suggested Fix] ComputeNetwork flaps when deleteDefaultRoutesOnCreate: true is set
## Description
The `ComputeNetwork` resource exhibits "flapping" behavior (constant reconciliation/updates) when the `deleteDefaultRoutesOnCreate` field is set to `true`.

### Symptom
Users observe the resource alternating between "Update in progress" and "Up to date", or simply never stabilizing, even though the underlying GCP resource is created correctly.

### Root Cause
The `deleteDefaultRoutesOnCreate` field is a "Create-Only" (ForceNew) argument in the underlying Terraform provider. Crucially, the GCP API **does not return this field** in the Get/Read response because it is a one-time instruction during creation, not a persistent state property.

As a result:
1.  KCC reads the User's Spec: `deleteDefaultRoutesOnCreate: true`
2.  KCC reads the GCP State: `deleteDefaultRoutesOnCreate: null` (or `false`)
3.  KCC detects a diff and attempts to reconcile.
4.  The cycle repeats indefinitely.

## Suggested Fix
We recommend modifying the `ComputeNetwork` Service Mapping configuration to treat `delete_default_routes_on_create` as a **`mutableButUnreadableField`**.

This directive instructs the KCC status-matcher to:
*   Trust the value in the User's Spec when the field is missing or empty in the State.
*   Suppress the diff if the Spec value allows it.

### Proposed Configuration Change
In `config/servicemappings/compute.yaml`, under `google_compute_network`:

```yaml
    - name: google_compute_network
      kind: ComputeNetwork
      # ... existing configuration ...
      mutableButUnreadableFields:
        - delete_default_routes_on_create
```

This change effectively standardizes the handling of this stateless field, aligning it with other similar "create-time parameter" fields in Config Connector.

anhdle-sso:
### BRIEF Change description

Suppress perpetual diffs on the virtual/create-only `delete_default_routes_on_create` field in `ComputeNetwork` by configuring it under `mutableButUnreadableFields` in the service mapping.

### Key Enhancements:
1. **Declarative Diff Suppression**:
   - Adds `delete_default_routes_on_create` to `mutableButUnreadableFields` under `google_compute_network` in `config/servicemappings/compute.yaml`.
   - When fetching live state from GCP/Terraform (where GCP's API does not return this create-only field), KCC automatically injects the stored spec value from the resource's annotation into the live state before diff comparison occurs.
   - Eliminates perpetual diffs cleanly and declaratively without requiring custom Go overrides.

Fixes #6195

### DIFF
--- a/config/servicemappings/compute.yaml
+++ b/config/servicemappings/compute.yaml
@@ -1285,6 +1285,8 @@ spec:
             group: resourcemanager.cnrm.cloud.google.com
     - name: google_compute_network
       kind: ComputeNetwork
+      mutableButUnreadableFields:
+        - delete_default_routes_on_create
       metadataMapping:
         name: name
       resourceID:
--- a/pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computenetwork/computenetworkautocreatesubnets/_final_object.diff
+++ b/pkg/test/resourcefixture/testdata/basic/compute/v1beta1/computenetwork/computenetworkautocreatesubnets/_final_object.diff
@@ -1,12 +1,14 @@
-7d6
+6d5
+<     cnrm.cloud.google.com/mutable-but-unreadable-fields: '{}'
+8d6
 <     cnrm.cloud.google.com/state-into-spec: absent
-11c10
+12c10
 <   generation: 3
 ---
 >   generation: 2
-20d18
+21d18
 <   resourceID: ${networkID}
-29c27
+30c27
 <   observedGeneration: 3
 ---
 >   observedGeneration: 2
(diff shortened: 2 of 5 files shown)


maqiuyujoyce:
I think the way we usually handle this is to configure it as the mutable but unreadable field (which is a one line change). Have you tried it?

anhdle-sso:
> I think the way we usually handle this is to configure it as the mutable but unreadable field (which is a one line change). Have you tried it?

nice call. updated the PR

ldanielmadariaga (dismissed):

maqiuyujoyce (approved):
/lgtm

